PR #2815, opened 2017-12-13
compaction bash script

### LINKED ISSUE #2807: Sweep is not running on products that have been restored
issue: We have a fix #2778, but products will have to release with the fix and impls to upgrade to the latest/patch versions, this would be a very tight schedule before the break.

Proposed solution: We should have a CLI that can fix up the issue on the affected deployment.

**Option1**
There is practically no way of determining from the`_transactions` table alone as to which timestamps were deleted by the CLI once the CLI has done its job.. We might be able to guess using the restore timestamp if the impl remembers the restore timestamp for each product. The service will have to be offline.

```
for (i = restoreTimestamp; ;i++) {
       Long commitTs = read commitTs from "_transactions" where startTs = i;
       if (commitTs == null) {
                  write (startTs=i, commitTs = -1) to "_transactions";
       } else {
                  break;
       }
}
```

Con:
Lots of redundant values in the transactions table, which will never be cleaned up as a lot of the timestamps will not be startTs at all.

**Option2**
If we transactionally read entire tables, i.e. a rangescan.ALL for all tables, we can cause all uncommitted transactions to rollback.

Con:
1. This CLI will be slow.
2. It can cause heavy load on Cassandra.
3. This will have to be schema aware.

**Option3**:

Run the migration CLI and skip writes. The impl will just involve adding a configuration option to the migration CLI to skip writes?

hsaraogi:
Fix #2807.

Tested on internal stack setup.

@ cstokes (but cant find your git alias).

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2815)

### DIFF
--- a/scripts/compaction.sh
+++ b/scripts/compaction.sh
@@ -1,4 +1,5 @@
-lineOfKeyspaces=`./service/bin/cqlsh -e "describe keyspaces; "`
+#!/bin/bash
+lineOfKeyspaces=`./service/bin/cqlsh -e "describe keyspaces;"`
 echo $lineOfKeyspaces
 
 for keyspace in ${lineOfKeyspaces[@]}
@@ -11,7 +12,6 @@ do
                 continue;
         fi
         txTable=" _transactions"
-        ./service/bin/cqlsh -e "use $keyspace; alter table \"_transactions\" with gc_grace_seconds = 0;"
-        ./service/bin/nodetool flush && ./service/bin/nodetool compact $keyspaceWithoutQuotes
-        echo "Compacted the $txTable in keyspace $keyspace!"
+        ./service/bin/nodetool flush && ./service/bin/nodetool compact $keyspaceWithoutQuotes$txTable
+        echo "Compacted the $txTable table in keyspace $keyspace!"
 done
--- a/scripts/compaction.sh
+++ b/scripts/compaction.sh
@@ -1,0 +1,17 @@
+lineOfKeyspaces=`./service/bin/cqlsh -e "describe keyspaces; "`
+echo $lineOfKeyspaces
+
+for keyspace in ${lineOfKeyspaces[@]}
+do
+        echo "keyspace = '$keyspace'"
+        keyspaceWithoutQuotes="${keyspace//\"}"
+        echo "keyspaceWithoutQuotes = $keyspaceWithoutQuotes"
+        if [[ $keyspace == *"system"* || $keyspace == *"__simple_rf_test_keyspace__"* ]]; then
+                echo "Skipping $keyspace as this is a test or system keyspace!"
+                continue;
+        fi
+        txTable=" _transactions"
+        ./service/bin/cqlsh -e "use $keyspace; alter table \"_transactions\" with gc_grace_seconds = 0;"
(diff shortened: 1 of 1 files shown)


scripts/compaction.sh
```diff
@@ -0,0 +1,17 @@
...
+do
+        echo "keyspace = '$keyspace'"
+        keyspaceWithoutQuotes="${keyspace//\"}"
+        echo "keyspaceWithoutQuotes = $keyspaceWithoutQuotes"
+        if [[ $keyspace == *"system"* || $keyspace == *"__simple_rf_test_keyspace__"* ]]; then
+                echo "Skipping $keyspace as this is a test or system keyspace!"
+                continue;
+        fi
+        txTable=" _transactions"
+        ./service/bin/cqlsh -e "use $keyspace; alter table \"_transactions\" with gc_grace_seconds = 0;"
```
tboam: Should we reset the gc_grace_seconds once we're done?

scripts/compaction.sh
```diff
@@ -0,0 +1,17 @@
...
+        ./service/bin/nodetool flush && ./service/bin/nodetool compact $keyspaceWithoutQuotes
```
tboam: should `flush` also be passed the keyspace?

hsaraogi: No flush doesnt take a keyspace.

scripts/compaction.sh
hsaraogi: Doesnt really matter, we can also skip the alter gc_grace_seconds as the required time has passed, but it hadnt on my test stack and this is the script tested.

scripts/compaction.sh
tboam: http://docs.datastax.com/en/archived/cassandra/2.0/cassandra/tools/toolsFlush.html

scripts/compaction.sh
tboam: We think manual changes to the `gc_grace_seconds` is a possible reason for Atlas failing to start against a degraded cluster.  So I'd push for not changing it if we can avoid it.

hsaraogi: Lets skip it then.

scripts/compaction.sh
hsaraogi: We can flush the entire memtable here, since this is run on all keyspaces. We will be a bit slower for the first keyspace but will save time on subsequent ones.

scripts/compaction.sh
tpetracca: Yea I'd rather avoid altering `gc_grace_seconds` as well.

scripts/compaction.sh
tpetracca: we should only compact the `_transactions` table, not the whole keyspace. Doing the whole keyspace depending on the deployment and product can take over a whole day to run.

hsaraogi: I had transactions table in there but missed it when adding `kespaceWithoutQuotes`, fixing and testing again.

hsaraogi:
Closing as we have run the compactions on the necessary deployment. Do we want to still merge this for future deployments @tpetracca ?

Issue #2821, opened 2017-12-14
Nonexistent Config Types Can Be Deserialized


jeremyk-91:
```
type: xyz

cluster:
  uris:
    - svc-47775-2.svc-47775.namespace.svc.cluster.local
```

This deserializes correctly after #2794, but it shouldn't. We really only want `default`, `kubernetes` or nothing at all to be accepted (at least for now).

hsaraogi:
Only a problem if you specify a misspelt version of 'kubernetes' as the config type and the config can be deserialized as a default config.

tpetracca:
@jeremyk-91 not sure I understand. Is the issue that it currently deserializes even if you don't specify a `type`? I think Div did that intentionally by specifying a default impl of the jackson polymorphic interface.

PR #2794, opened 2017-12-07
Support for running within a k8s stateful set

darora:
**Goals (and why)**:

**Implementation Description (bullets)**:

**Concerns (what feedback would you like?)**:

**Where should we start reviewing?**:

**Priority (whenever / two weeks / yesterday)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2794)

PR #2822, opened 2017-12-14
Cluster Config Deserialisation Tests

jeremyk-91:
**Goals (and why)**:
- Add full deserialisation tests for polymorphic deserialisation of `ClusterConfiguration`
- Gain more confidence that `ClusterConfiguration` deserialisation works properly

- Add several yaml blobs which contain possible cluster configurations, including one without type information (because this is consistent with the legacy format)

- One of the tests exposes #2821 but I don't see a simple way to fix it so I left it as it is.
- The K8s test here is not ideal, not sure if there's a better way.

- It's probably easiest to look at each of the small yaml blobs.

**Priority (whenever / two weeks / yesterday)**: whenever; I did a manual test on the PR #2794 and am pretty confident it works.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2822)


timelock-agent/src/test/java/com/palantir/timelock/config/ClusterConfigurationDeserializationTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    public void throwsOnMalformedClusterConfiguration() throws IOException {
+        assertThatThrownBy(() -> deserializeClusterConfiguration(CLUSTER_CONFIG_MALFORMED))
+                .isInstanceOf(UnrecognizedPropertyException.class)
+                .hasMessageContaining("my-node-set");
+    }
+
+    @Test
+    public void doesNotDeserializeKubernetesClusterConfigurationAsDefaultClusterConfiguration() throws IOException {
```
gmaretic: Do we already have a test with a k8s config that deserializes correctly?

timelock-agent/src/test/java/com/palantir/timelock/config/ClusterConfigurationDeserializationTest.java
```diff
@@ -0,0 +1,92 @@
...
+        assertDefaultClusterConfigurationCorrect(deserializeClusterConfiguration(CLUSTER_CONFIG_NO_TYPE_INFO));
+    }
+
+    @Test
+    public void canDeserializeClusterConfigurationWithDefaultTypeInformation() throws IOException {
+        assertDefaultClusterConfigurationCorrect(deserializeClusterConfiguration(CLUSTER_CONFIG_DEFAULT_TYPE_INFO));
+    }
+
+    @Test
+    public void canDeserializeClusterConfigurationWithInvalidTypeInformation() throws IOException {
```
gmaretic: I'm a bit uneasy merging a test that tests that something is wrong :)

Maybe rather write the test you want to have and then @Ignore it?

jeremyk-91: Ah, okay. Yeah, the ignore-test makes sense.

timelock-agent/src/test/java/com/palantir/timelock/config/ClusterConfigurationDeserializationTest.java
jeremyk-91: There are tests that the already-deserialized k8s config is right. As mentioned in the comment, unfortunately we rely on a static instance that finds the local hostname and uses it. In terms of deserialisation unfortunately not 😞

PR #2826, opened 2017-12-18
Fix/new sweep oom

gsheasby:
**Goals (and why)**: Fix the sweep OOM that caused outages in PDS-61184.
 The query in CqlExecutorImpl.getTimestamps doesn't specify an exact set of row keys, meaning that the Thrift driver uses a range scan under the hood, causing OOM when the values are large.

* Use a query with an IN clause to specify row keys

**Concerns (what feedback would you like?)**: Does this have edge cases? (is the paging token correct?)
Under testing, there was still some memory pressure - running sweep on a stream table with a batch size of 2048 succeeded more than 70 times, but failed when running sweep on the stream table and concurrently writing files to the table. In practice, we will be using a lower batch size (the default is 128 currently).
Should I add more tests / remove the old ones?

Before merging, we should:
- [x] Remove the old getTimestamps method, assuming it isn't used elsewhere
- [x] Test with a more realistic batch size (and concurrent writing to the table)
- [x] Audit the sweep code path to ensure that there aren't more unbounded memory queries elsewhere

Edit: decided not to add a check that the batch size is at most 65536 (CQL only allows 2^16 parameters in the query). It's non-trivial to do this just for C*KVS, and it's extremely likely we'll hit this case anyway. Can revisit after winter break

**Where should we start reviewing?**: The CQL queries

@jeremyk-91 and @tboam for SA.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2826)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutor.java
```diff
@@ -41,4 +41,9 @@
             byte[] startRowInclusive,
             byte[] endRowInclusive,
             int limit);
+
+    List<CellWithTimestamp> getTimestamps(
```
gsheasby: Can be one line, and the above method declaration can be deleted.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
```diff
@@ -80,6 +80,28 @@
         return executeAndGetCells(query, startRowInclusive, CqlExecutorImpl::getCellFromRow);
     }
 
+    @Override
+    public List<CellWithTimestamp> getTimestamps(
+            TableReference tableRef,
+            List<byte[]> rowsAscending,
+            int limit) {
+        String selQuery = "SELECT key, column1, column2 FROM %s"
```
gsheasby: can be one line

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetCellTimestamps.java
```diff
@@ -96,28 +95,24 @@ private void fetchBatchOfTimestampsBeginningAtStartRow() {
...
+            if (rows.isEmpty()) {
                 return;
             }
 
             // Note that both ends of this range are *inclusive*
-            List<CellWithTimestamp> batch = cqlExecutor.getTimestamps(tableRef, rangeStart, rangeEnd.get(), batchHint);
+            List<CellWithTimestamp> batch = cqlExecutor.getTimestamps(tableRef, rows, batchHint);
             timestamps.addAll(batch);
-            rangeStart = RangeRequests.nextLexicographicName(rangeEnd.get());
+            rangeStart = RangeRequests.nextLexicographicName(Iterables.getLast(rows));
```
gsheasby: Shouldn't we use the last cell in the batch here, to avoid missing data?

gsheasby:
Tested with batch sizes 128, 512 and 1024 and concurrent file upload - sweep continued just fine.

tboam:
Since this performs better than the current implementation I'd like to merge this if we can verify that there are no issues with the implementation.  @tpetracca @ashrayjain

nziebart:
Do we understand why the OOM happened here? I'd like to understand why the multiget avoids it. My understanding is that the thrift query returns every row in the range. So, regardless of whether we do a range scan or a multiget (via CQL) over those rows, we scan over the same data.

It's also surprising that this query could cause an OOM, given that it does not return any values (and therefore should not be retaining any values on the heap). Perhaps we don't understand the codepath for CQL queries as well the thrift ones, and it's actually post-filtering the columns that were not part of the query after it gets enough rows. Do we have a stacktrace for the OOM?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetCellTimestamps.java
nziebart: this seems ok to me: if the query returned zero results, then the last cell in the query could not have had any results, so no need to include it in the next one

gsheasby: My thought here was that we now pass in 128 row keys (by default), but if we have wide rows then the results returned could be from the first 28 of these (for example), so by picking the 128th row key as the paging token, there's 100 row keys' worth of data that we're skipping over.

The tests are passing, though, and I'd expect this kind of case to be covered by the tests. Will do some more digging pre-merge.

gsheasby: This isn't an issue - I was conflating a couple of parts of our paging logic.

gsheasby:
The slice predicates are different for the Thrift query (to get the row keys) and the CQL query. Since in the CQL-turned-Thrift query, we need to fetch cells and timestamps, this also scans over the values, causing increased memory pressure in Cassandra.

The new query avoids this by providing specific row keys to the CQL query, causing a different code path (apparently) to be hit under the hood.

PR #2829, opened 2017-12-20
add logging for the values used to determine sweep priority scores

tboam:
**Goals (and why)**: We want to understand which tables are being chosen and why.  In cases where we discover a table hasn't been swept in a long time we want to understand what was being chosen instead and if the logic should be changed.

**Implementation Description (bullets)**: add a check so we only log every 10 minutes.  Add logging method.

**Concerns (what feedback would you like?)**: Have I got all the Args to the log in right order?  Is logging a line for all tables every 10 minutes too often?  Not often enough?

**Where should we start reviewing?**: SweepPriorityCalculator.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2829)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityCalculator.java
```diff
@@ -107,6 +117,53 @@ public SweepPriorityCalculator(KeyValueService kvs, SweepPriorityStore sweepPrio
         return scores;
     }
 
+    private boolean decideWhetherToLogAllPriorities() {
+        ZonedDateTime now = ZonedDateTime.now();
```
hsaraogi: We could use a rate-limiter that has `1 permit/(10 minutes)` or `0.1 permit/minute`.

tboam: Is this required for merging?

tboam:
@hsaraogi is review comment a blocker for you?  I think the logic is cleaner this way.

docs/source/release_notes/release-notes.rst
```diff
@@ -55,6 +55,10 @@ develop
            Stream store value tables are now more likely to be chosen because they contain lots of data per write.  We ensure we sweep index tables before value tables, and allow a gap after sweeping index tables and before sweeping value tables.  Wait 3 days between sweeps of a value table to prevent unnecessary work, allow other tables to be swept and tombstones to be compacted away. 
            (`Pull Request <https://github.com/palantir/atlasdb/pull/2793>`__)
 
+    *    - |improved| |logs|
+         - Added logging of the values used to determine which table to sweep, provides more insight into why tables are being swept and others aren't.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/2829>`__)
+
     *    -
```
hsaraogi: Extra list item.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityCalculator.java
hsaraogi: Not really required, but it will be cleaner if we use the Java rate-limiter, we have this style of rate-limiting logging at multiple places in the codebase and this one will be a snowflake, also we are implementing custom logic for exactly what a ratelimiter does.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityCalculator.java
```diff
@@ -107,6 +117,53 @@ public SweepPriorityCalculator(KeyValueService kvs, SweepPriorityStore sweepPrio
...
+                    LoggingArgs.tableRef(tableRef),
+                    SafeArg.of("priorityScore", priorityScore),
+
+                    SafeArg.of("oldPriority_lastSweepTimeMillis", oldPriority.lastSweepTimeMillis()),
+                    SafeArg.of("oldPriority_minimumSweptTimestamp", oldPriority.minimumSweptTimestamp()),
+                    SafeArg.of("oldPriority_cellTsPairsExamined", oldPriority.cellTsPairsExamined()),
+                    SafeArg.of("oldPriority_staleValuesDeleted", oldPriority.staleValuesDeleted()),
+                    SafeArg.of("oldPriority_writeCount", oldPriority.writeCount()),
+
+                    SafeArg.of("newPriority_lastSweepTimeMillis", newPriority.lastSweepTimeMillis()),
```
hsaraogi: nit: we can refactor this to generate logging for old and new priorities as this a bunch of duplicate code. Something like `getLogMesasageAndArgsForPriority("old"/"new", priority)` If we add a new parameter for priority calculation, we'll have to update both places in the future.

tboam: This is kind of messy because we have to generate the 2 fragments, concatenate the  strings and merge the arrays.  Given we've only duplicated this code once I don't think it is worth the effort and reduced readability to refactor this.  If we end up logging many historical versions then we can think more about how to do this.

tboam:
@hsaraogi ready for re-review,  I've pushed a new version which uses a RateLimiter.

However, would it be simpler if we didn't have this at all.  We were worried about excessive logging but when Sweep is progressing though a large table it will take more 10 minutes between calling this code.  If there's nothing to sweep we back off for 2 minutes.  So removing this check shouldn't result in much more logging really.  Thoughts?

hsaraogi:
I think 10 minutes is an okay rate to log as we are logging a lot of info, especially since we could have 10s of tables on products, I was wondering if we should have some config to disable this logging for the large internal product that can have 1000s of tables and find this log spammy?

gsheasby:
This PR has been hanging around for a while. Do we still want this? Any reason not to merge it in? @tboam

PR #2830, opened 2017-12-20
Fix/sweep results cells examined

mswintermeyer:
**Goals (and why)**:
Fix SweepResults.getCellTsPairsExamined over multiple batches

BatchOfCellsToSweep.numCellTsPairsExaminedSoFar() -> numCellTsPairsExamined(), as that's what it was returning anyways.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2830)

### DIFF
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
@@ -64,6 +64,7 @@ import com.palantir.atlasdb.transaction.service.TransactionService;
 import com.palantir.atlasdb.transaction.service.TransactionServices;
 import com.palantir.timestamp.InMemoryTimestampService;
 import com.palantir.timestamp.TimestampService;
+import com.palantir.util.Pair;
 
 public abstract class AbstractSweepTaskRunnerTest {
     private static final String FULL_TABLE_NAME = "test_table.xyz_atlasdb_sweeper_test";
@@ -73,12 +74,15 @@ public abstract class AbstractSweepTaskRunnerTest {
 
     private static final List<Cell> SMALL_LIST_OF_CELLS = Lists.newArrayList();
     private static final List<Cell> BIG_LIST_OF_CELLS = Lists.newArrayList();
+    private static final List<Cell> BIG_LIST_OF_CELLS_IN_DIFFERENT_ROWS = Lists.newArrayList();
 
     static {
         for (int i = 0; i < 10; i++) {
-            String colName = String.format("c%d", i);
+            String zeroPaddedIndex = String.format("%05d", i);
             BIG_LIST_OF_CELLS.add(
-                    Cell.create("row".getBytes(StandardCharsets.UTF_8), colName.getBytes(StandardCharsets.UTF_8)));
+                    Cell.create("row".getBytes(StandardCharsets.UTF_8), (COL + zeroPaddedIndex).getBytes(StandardCharsets.UTF_8)));
+            BIG_LIST_OF_CELLS_IN_DIFFERENT_ROWS.add(
+                    Cell.create(("row" + zeroPaddedIndex).getBytes(StandardCharsets.UTF_8), (COL + zeroPaddedIndex).getBytes(StandardCharsets.UTF_8)));
         }
         SMALL_LIST_OF_CELLS.addAll(BIG_LIST_OF_CELLS.subList(0, 4));
     }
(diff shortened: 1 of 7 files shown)


gsheasby:
I noticed this too this week - will look at it in the new year. Thanks for the contribution!

It would also be nice to report "cells examined so far" rather than "cells examined by last batch" when running a full sweep through the endpoint.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
```diff
@@ -513,6 +524,35 @@ public void testSweepBatches() {
...
+        putTwoValuesInEachCell(BIG_LIST_OF_CELLS_IN_DIFFERENT_ROWS);
+
+        int deleteBatchSize = 2;
+        Pair<List<List<Cell>>, SweepResults> sweptCellsAndSweepResults = runSweep(cellsSweeper, spiedSweepRunner,
+                10, 1, deleteBatchSize);
+        List<List<Cell>> sweptCells = sweptCellsAndSweepResults.getLhSide();
+        SweepResults sweepResults = sweptCellsAndSweepResults.getRhSide();
+        assertThat(Iterables.concat(sweptCells)).containsExactlyElementsOf(BIG_LIST_OF_CELLS_IN_DIFFERENT_ROWS);
+        for (List<Cell> sweptBatch : sweptCells.subList(0, sweptCells.size() - 1)) {
+            // We requested deleteBatchSize = 2, so we expect between 2 and 4 timestamps deleted at a time.
```
jeremyk-91: nit: if I'm reading it correctly, you want start-inclusive but end-exclusive, since each row has at most 2 (while this is both start and end inclusive).

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
```diff
@@ -488,7 +495,9 @@ public void testSweepBatchesUpToDeleteBatchSize() {
 
         putTwoValuesInEachCell(SMALL_LIST_OF_CELLS);
 
-        List<List<Cell>> sweptCells = runSweep(cellsSweeper, spiedSweepRunner, 8, 1, 4);
+        Pair<List<List<Cell>>, SweepResults> sweptCellsAndSweepResults =
```
gsheasby: nit: would be nice to have "swept cells and sweep results" encapsulated as a private class, but I can't immediately think of a good name for it.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
gsheasby: @jeremyk-91 What do you mean by 'this' in this case? Do you mean that the swept batch should have between 2 and 3 timestamps deleted?

docs/source/release_notes/release-notes.rst
```diff
@@ -69,6 +69,10 @@ develop
          - Messages to the `slow-lock-log` now log at `WARN` rather than `INFO`, these messages can indicate a problem so we should be sure they are visible.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/2828>`__)
 
+    *    - |fixed|
+         - SweepResults.getCellTsPairsExamined now returns the correct result when sweep affects multiple batches.
```
gsheasby: Let's explain the problem we fixed here.

"SweepResults.getCellTsPairsExamined now returns the correct result when sweep is run over multiple batches. Previously, the result would only count cell-ts pairs examined in the last batch."

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
jeremyk-91: @gsheasby that's what I meant - but disregard. I think I got confused, thinking that we had a max-cell-ts-pairs of 2 (it is 1000).

mswintermeyer:
Thanks everyone for reviewing this! What else do I need to do to get this merged? Also, I think before break I ran the verifications locally and they passed and I was wondering if the verify-fail was a flake. I don't have premissions to rebuild though.

hsaraogi:
Hey @mswintermeyer, our tests have been very flaky over the last month. One of the highest priorities this week is to reduce the flakiness. The build failures for this PR are all known flakes. I retriggered the build and the build number is 8631. Also, you should now have permissions to retrigger builds :).

PR #2832, opened 2018-01-02
Publish 0.39.x from internal circle

fsamuel-bs:
**Goals (and why)**: Publish 0.39.x from internal circle.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2832)


fsamuel-bs:
I've added docs on our internal release process, and also added docs on develop.

PR #2834, opened 2018-01-03
Docs for manual rate limiting

hsaraogi:
**Goals (and why)**: Docs for client level configuration of limits in the qos service.

**Implementation Description (bullets)**: Docs.

1. Should we include the future plans in the docs? (eg: we have a QoS service config but no QoS service which is awkward)
2. Should we skip the config option for ``qosService`` in the config?
3. Do we want to make all config param names kabab-case to prevent breaks in the future. (This will be a break at the moment but we can be pretty confident that no-one is using this yet.)

**Where should we start reviewing?**: any file.
 
**Priority (whenever / two weeks / yesterday)**: two weeks (as this is the break week).

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2834)

### DIFF
--- a/docs/source/configuration/qos_configs/qos_client_config.rst
+++ b/docs/source/configuration/qos_configs/qos_client_config.rst
@@ -1,0 +1,53 @@
+.. _qos-client-configuration:
+
+QoS Client Configuration
+========================
+
+.. warning::
+
+    If you are using Cassandra, then you can use the AtlasDB QoS to rate-limit requests to Cassandra. Note that this
+    service is under active development and testing, please contact the AtlasDB team before using this feature.
+
+
+You will need to update your AtlasDB configuration in order to have said clients limit the read and write requests to
+Cassandra.
+
+.. note::
+
+    QoS client configuration is a part of the AtlasDB Runtime configuration.
+
+Install-Time Configuration
+--------------------------
+
+An AtlasDB config can have an optional ``qos`` block if the service wants to rate limit the reads/writes to Cassandra. This is
+live-reloadable and hence the limits can be modified while the service is online. The limits will be loaded when the
+next request hits the server and knowledge of the previous rate-limiting will be lost.
+
+Optional parameters:
+
+.. list-table::
+    :widths: 5 40
+    :header-rows: 1
+
+    *    - Property
+         - Description
+
+    *    - maxBackoffSleepTime (HumanReadableDuration)
+         - The rate-limiter can cause the request processing thread to sleep for sometime if soft-limiting is being applied.
+           This parameter configures the maximum time a client request can be made to wait by the rate limiter.
+           This should definitely be less than the server idle timeout or jetty will cause the request to timeout and potentially retry.
+           The default value is 10 seconds.
+           This is of type `HumanReadableDuration <https://github.com/palantir/http-remoting-api/blob/develop/service-config/src/main/java/com/palantir/remoting/api/config/service/HumanReadableDuration.java>`__.
+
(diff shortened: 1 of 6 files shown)


docs/source/services/qos_service/index.rst
```diff
@@ -0,0 +1,17 @@
...
+
+   Rate limiting using the QoS service is currently an experimental feature.
+
+.. toctree::
+    :maxdepth: 1
+    :titlesonly:
+
+    enabling-rate-limiting
+
+The AtlasDB QoS Service is an attempt to prevent misbehaving or rogue clients from making Cassandra slow for everyone
```
gmaretic: "is an attempt to" -> aims to

docs/source/configuration/qos_configs/qos_client_config.rst
```diff
@@ -0,0 +1,53 @@
+.. _qos-client-configuration:
+
+QoS Client Configuration
+========================
+
+.. warning::
+
+    If you are using Cassandra, then you can use the AtlasDB QoS to rate-limit requests to Cassandra. Note that this
```
gmaretic: This is a bit unclear -- what if I don't have Cassandra, can I still use it? Also I think the warning should only be about the QoS service being under development and testing and the rest should be normal text.

docs/source/configuration/qos_configs/qos_client_config.rst
```diff
@@ -0,0 +1,53 @@
...
+Optional parameters:
+
+.. list-table::
+    :widths: 5 40
+    :header-rows: 1
+
+    *    - Property
+         - Description
+
+    *    - maxBackoffSleepTime (HumanReadableDuration)
```
gmaretic: Maybe unnecessary to say it's a `HumanReadableDuration ` twice. But OK if you're worried people will miss it.

docs/source/services/qos_service/enabling-rate-limiting.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Enabling Rate Limiting
+======================
+
+Currently the QoS service allows a client to configure the read and write limits and enforces these limits.
+Detailed documentation is :ref:`here <qos-client-configuration>`.
+
+QoS also publishes client level metrics for the following:
+
+        - Meter readRequestCount
```
gmaretic: Specify the common prefix (com.palantir...) and say that they are all meters instead of repeating it every time

docs/source/services/qos_service/enabling-rate-limiting.rst
```diff
@@ -0,0 +1,42 @@
...
+
+A client should ideally look at the historical values of the ``bytesRead`` and ``bytesWritten`` metrics to determine the limits.
+
+How does QoS rate limit?
+------------------------
+
+We are using the standard Java Rate limiter to enforce the rate limiting with some modifications to support adjustment.
+It is not possible to estimate the number of bytes a read request is going to consume beforehand, so we request a
+default number of permits from the ratelimiter (currently 100) and subsequently return some permits (if the request
+returned less than a 100 bytes) or steal more permits (if the request returned more than a 100 bytes).
```
gmaretic: steal -> consume

hsaraogi: Steal seems better her eas you are not consuming in the standard way to acquire permits. The rate-limiter has been modified to add new methods and this method is also named `steal`.

gmaretic: Makes sense!

PR #2836, opened 2018-01-08
99-Flake Part 1: Retry N Times + Cassandra One Node Down Get

jeremyk-91:
**Goals (and why)**:
- Incorporate some of my recent experimental work in automated test retrying into AtlasDB so we can benefit from it quickly, while I figure out how to move the project as a whole into OSS and bring up components like documentation and CI tool-chains to the required standard.

- Implement a JUnit `TestRule` that allows for things to be retried N times.
- Wider-reaching changes to AtlasDB to stop more of the flakes is intended to be part 2
- Non-core portions of work involving selective retry or more comprehensive reporting are intended to be later parts.

- Were there any unexpected things that I've run afoul of?

- `ShouldRetry.java` explains what's going on; then proceed to `FlakeRetryingRule`.

**Priority (whenever / two weeks / yesterday)**: this week - would be nice to start alleviating some of the flaky pain

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2836)


flake-rule/src/main/java/com/palantir/flake/FlakeRetryingRule.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static void runStatementWithRetry(ShouldRetry retryAnnotation, Statement base, Description description) {
+        Preconditions.checkArgument(retryAnnotation.numAttempts() > 0,
+                "Number of attempts should be positive, but found %s", retryAnnotation.numAttempts());
+
+        for (int attempt = 1; attempt <= retryAnnotation.numAttempts(); attempt++) {
+            try {
+                base.evaluate();
+                return;
+            } catch (Throwable t) {
```
tboam: Wonder if we should retry only Exceptions?  Errors would be OOMs or similarly fatal problems that are unlikely to succeed on retry and which we want to know if application is hitting, even just occasionally.

flake-rule/src/main/java/com/palantir/flake/FlakeRetryingRule.java
```diff
@@ -0,0 +1,81 @@
...
+                log.info("Test {}.{} failed on attempt {} of {}.",
+                        description.getClassName(),
+                        description.getMethodName(),
+                        attempt,
+                        retryAnnotation.numAttempts());
+                if (attempt == retryAnnotation.numAttempts()) {
```
tboam: nit: this could just be after the `for` loop

flake-rule/src/main/java/com/palantir/flake/ShouldRetry.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.palantir.flake;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * An annotation signalling that a given test class or test method should be retried in the event of a test failure.
+ * In the event that both a test class and a test method are annotated with this
```
tboam: Incomplete sentence

flake-rule/src/main/java/com/palantir/flake/ShouldRetry.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Note that this annotation will only actually cause a test to be retried if it is run with the
```
tboam: I wonder if there's a way around this?  Or if we could add a Findbugs rule that catches cases where you forget to add the rule.  Definitely for the future and not now.

flake-rule/src/main/java/com/palantir/flake/FlakeRetryingRule.java
```diff
@@ -0,0 +1,81 @@
...
+                : Optional.ofNullable(description.getTestClass().getAnnotation(ShouldRetry.class));
+    }
+
+    private static void runStatementWithRetry(ShouldRetry retryAnnotation, Statement base, Description description) {
+        Preconditions.checkArgument(retryAnnotation.numAttempts() > 0,
+                "Number of attempts should be positive, but found %s", retryAnnotation.numAttempts());
+
+        for (int attempt = 1; attempt <= retryAnnotation.numAttempts(); attempt++) {
+            try {
+                base.evaluate();
```
tboam: Should we also log that the test succeeded after a certain number of retries?  In the future we could analyse which tests are flaking most often and fix them (to, for example, reduce build times).

jeremyk-91: Yes, I think logging on success makes sense (the idea about extracting data was part of the "non-core portions of work" I mentioned, but this would give us the ability to do some ad-hoc analysis if needed).

Following your comment above - I think it makes sense to log even on the first attempt; I think the "pain" introduced by having a spurious `Test foo.bar succeeded on attempt 1 of 5` would be less than that from not seeing `foo.bar` mentioned anywhere, and then having to figure out that that's because it passed on the first try.

I'll add logging on the 1st round.

(Of course, this will only be printed if the test method is annotated with `@ShouldRetry`.)

flake-rule/src/main/java/com/palantir/flake/FlakeRetryingRule.java
jeremyk-91: It could, though we would lose the detail of the exception (of course we can separately track it). Though currently we just throw the exception returned on failure of the last attempt.

I think I'll leave as is for now - but can extract it if needed.

flake-rule/src/main/java/com/palantir/flake/ShouldRetry.java
jeremyk-91: oops! I've updated it to read

```
 * If specified on a test class, all test methods in the class will be considered retriable. In the event that both a 
 * test class and a test method have this annotation, the method-level annotation will take precedence.
```

flake-rule/src/main/java/com/palantir/flake/ShouldRetry.java
jeremyk-91: Yep, a Baseline rule sounds plausible. If the annotation is present but the rule is not, there's almost certainly something wrong.

jeremyk-91:
TODO on me:

- [ ] document how this interfaces with Before/After (and their Class versions) annotations
- [ ] add an example of a test that relies on the above behaviour

PR #2837, opened 2018-01-09
Allow overriding JDBC URL for Oracle

gbonik:
Also remove getDefaultUrl() from DBType

**Goals (and why)**:
See QA-102902 / PDS-60780

This will break the internal product but I have a fix ready

High

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2837)


commons-db/src/main/java/com/palantir/nexus/db/DBType.java
```diff
@@ -42,17 +41,12 @@ public boolean supportsGIS() {
...
 
-    private DBType(String defaultUrl, String driver, String testQuery, boolean hasGIS) {
-        this.defaultUrl = defaultUrl;
+    private DBType(String driver, String testQuery, boolean hasGIS) {
         this.driver = driver;
         this.testQuery = testQuery;
         this.hasGIS = hasGIS;
     }
 
-    public String getDefaultUrl() {
```
hsaraogi: Do we need to break this? I am not entirely aware of the usecase here, but looks like we should use the configured URL if its specified or default to the default url,

gbonik: This is legacy code that we used to format the URLs. The logic now lives in OracleConnectionConfig.getUrl() etc., so getDefaultUrl() is not used by Atlas. It is still used in our internal product, but I have a change which makes it go through the new code path instead.

hsaraogi:
Hey! can you put up a Pr with the fix on the large internal product side. Looks like the autobump is broken this week. We have an API break too and @jeremyk-91 is looking into the fix.

gbonik:
@hsaraogi The internal change has already merged 4 days ago. Search gerrit for QA-102902

PR #2838, opened 2018-01-09
Correct use of SafeArg and UnsafeArg

tboam:
Use different names for the safe and unsaf table names (matches spec), don't rewrap SafeArg and UnsafeArg

**Goals (and why)**: Use different names for the safe and unsafe table names (matches spec).   Don't rewrap SafeArg and UnsafeArg.

**Implementation Description (bullets)**:  Updated names, removed wrappers

**Concerns (what feedback would you like?)**: Is it ok to just change the names?  Don't think anything uses them.

**Where should we start reviewing?**: Only 2 classes

**Priority (whenever / two weeks / yesterday)**: Before release as currently logging table names as safe

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2838)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityCalculator.java
```diff
@@ -115,8 +114,8 @@ private void logUnsweptTables(Set<TableReference> unsweptTables) {
         LoggingArgs.SafeAndUnsafeTableReferences safeAndUnsafeTableReferences = LoggingArgs.tableRefs(unsweptTables);
 
         log.debug("Unswept tables: {} and {}",
-                SafeArg.of("tables", safeAndUnsafeTableReferences.safeTableRefs()),
-                SafeArg.of("unsafeTables", safeAndUnsafeTableReferences.unsafeTableRefs()));
+                safeAndUnsafeTableReferences.safeTableRefs(),
+                safeAndUnsafeTableReferences.unsafeTableRefs());
```
jeremyk-91: ok, confirmed these are already Args so double-wrapping isn't needed.

docs/source/release_notes/release-notes.rst
```diff
@@ -55,6 +55,10 @@ develop
            Stream store value tables are now more likely to be chosen because they contain lots of data per write.  We ensure we sweep index tables before value tables, and allow a gap after sweeping index tables and before sweeping value tables.  Wait 3 days between sweeps of a value table to prevent unnecessary work, allow other tables to be swept and tombstones to be compacted away. 
            (`Pull Request <https://github.com/palantir/atlasdb/pull/2793>`__)
 
+    *    - |fixed| |logs|
+         - Safe and Unsafe table name logging args are now different, fixed unreleased bug tables names were logged as Safe
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/2838>`__)
```
jeremyk-91: I guess we mean `fixed unreleased bug where unsafe table names were logged as Safe`?

tboam: fixed while merging in develop

PR #2840, opened 2018-01-09
Fix log message and add logging at the right place

hsaraogi:
**Goals (and why)**: Fix comfusing log message from `loadProgress`.

**Implementation Description (bullets)**: Saving the progress creates confusion as it says that it is choosing a table to sweep. 

**Concerns (what feedback would you like?)**: NA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2840)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/progress/SweepProgressStoreImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/progress/SweepProgressStoreImpl.java
@@ -149,15 +149,13 @@ public final class SweepProgressStoreImpl implements SweepProgressStore {
 
     private static Optional<SweepProgress> hydrateProgress(Map<Cell, Value> result) {
         if (result.isEmpty()) {
-            log.info("No persisted intermittent SweepProgress information found. "
-                    + "Sweep will choose a new table to sweep.");
+            log.info("No persisted SweepProgress information found.");
             return Optional.empty();
         }
         try {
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/progress/SweepProgressStoreImpl.java
```diff
@@ -149,15 +149,13 @@ private void tryInitialize() {
 
     private static Optional<SweepProgress> hydrateProgress(Map<Cell, Value> result) {
         if (result.isEmpty()) {
-            log.info("No persisted intermittent SweepProgress information found. "
-                    + "Sweep will choose a new table to sweep.");
```
hsaraogi: Dont guess what might be done next.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/progress/SweepProgressStoreImpl.java
```diff
@@ -149,15 +149,13 @@ private void tryInitialize() {
...
+            log.info("No persisted SweepProgress information found.");
             return Optional.empty();
         }
         try {
             return Optional.of(OBJECT_MAPPER.readValue(result.get(CELL).getContents(), SweepProgress.class));
         } catch (Exception e) {
-            log.warn("Error deserializing SweepProgress object while attempting to load intermediate result. "
-                    + "Sweep will choose a new table to sweep.", e);
```
hsaraogi: Dont guess the caller ('while attempting to load intermediate result') and what might be done next.

PR #2842, opened 2018-01-09
Make createTablesAutomatically final

mnazbro:
**Goals (and why)**:
InMemoryKeyValueService has a volatile variable that is never
modified. This variable can be made final.

* createTablesAutomatically was changed to final from volatile

whenever

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2842)


tboam:
@hsaraogi to check if test failures are flakes or not.

hsaraogi:
The last failure is a flake. Rebuilding now.

PR #2843, opened 2018-01-10
99-Flake Part 2: Applying the Retry Rule

jeremyk-91:
**Goals (and why)**:
- Reduce the frequency of build flakes by actually retrying tests; #2836 only retried _one_ of the tests that was flaky.

- Apply the `@ShouldRetry` annotation and the flake retrying rule where needed.
- Release the schema mutation lock after failed attempts for the Cassandra integration tests.
- We've had a success rate of 14/16 = 87.5% in preliminary testing, with both failures being outside of the tests (one was on `wget` and another was on the errorprone compiler OOMing).

Flakes I did NOT put the rule on:
- For the TimeLock integration test, I made nodes wait for the cluster to be healthy before firing requests / between requests in a rolling bounce. These are time-limited anyway (see `TestableTimelockCluster`) so we should be fine.
- For the Coalescing Supplier test, I made it a bit more permissive to account for scheduling quirks.

- Is releasing the schema mutation lock after an arbitrary failed attempt in the Cassandra integration tests safe?

- Might be easier to do commit by commit, as for the most part the flaky tests are handled one test class at a time.

**Priority (whenever / two weeks / yesterday)**: this/early next week to alleviate ~pain au chocolat~ flaky pain, but of course not super high priority.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2843)

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/HeartbeatServiceIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/HeartbeatServiceIntegrationTest.java
@@ -29,6 +29,7 @@ import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.junit.rules.TestRule;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,7 +38,10 @@ import com.palantir.atlasdb.config.LockLeader;
 import com.palantir.atlasdb.containers.CassandraContainer;
 import com.palantir.atlasdb.containers.Containers;
 import com.palantir.atlasdb.keyvalue.impl.TracingPrefsConfig;
+import com.palantir.flake.FlakeRetryingRule;
+import com.palantir.flake.ShouldRetry;
 
+@ShouldRetry // There are flakes with the heartbeat service throwing because it was unable to stop beating.
 public class HeartbeatServiceIntegrationTest {
     private static final Logger log = LoggerFactory.getLogger(HeartbeatServiceIntegrationTest.class);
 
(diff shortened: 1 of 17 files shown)


atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/HeartbeatServiceIntegrationTest.java
```diff
@@ -37,7 +38,10 @@
 import com.palantir.atlasdb.containers.CassandraContainer;
 import com.palantir.atlasdb.containers.Containers;
 import com.palantir.atlasdb.keyvalue.impl.TracingPrefsConfig;
+import com.palantir.flake.FlakeRetryingRule;
+import com.palantir.flake.ShouldRetry;
 
+@ShouldRetry // There are flakes with the heartbeat service throwing because it was unable to stop beating.
```
tboam: Adding comments to explain why the test is flaking seems like good practice.  Maybe we should enforce in the future by adding a required `reason` field to the annotation?

jeremyk-91: +1. This would be good to do before broader rollout of the flake-rule as well (so we don't have people shouting about API breaks!)

jeremyk-91: also, yes - retry on specific exception type is planned for the next part.

jeremyk-91:
Let me check the effect on build times.

OK, yep that sounds like a good addition for Part 3! Let's break the API as much as we can before this is released more generally 😄

jeremyk-91:
Build times seem broadly similar for successful builds, and seem to be faster than a small sample of flaky failures I looked at on develop (perhaps waiting for schema mutation lock tests to fail takes time?)

leader-election-impl/src/test/java/com/palantir/paxos/CoalescingSupplierTest.java
```diff
@@ -69,8 +71,10 @@ public void batchesConcurrentRequests() throws InterruptedException {
 
         batch.await();
 
-        // once for initial task; once for batch
-        verify(delegate, times(2)).get();
+        // At least some of these requests should be batched. We can't guarantee it will always be 2 though, if
+        // we get really unlucky with scheduling.
+        verify(delegate, atLeast(2)).get();
+        verify(delegate, atMost(5)).get();
```
felixdesouza: doesn't this defeat the point of the entire test? granted the test as written will always be flaky

PR #2844, opened 2018-01-10
[WIP] Retry on RateLimitedException

fsamuel-bs:
**Goals (and why)**: When doing StreamStore writes, we break the stream in 1MB blocks and call `kvs.multiPut`. MultiPut groups up blocks in groups of 10MBs and spawns a thread to write such group to CKVS.

If the stream is big, as the internal search product case, we will use all CKVS threads to attempt such write. This causes problems to the underlying DB. We want to throttle the write at the QoS level.

If one of the threads that attempt the write fails at the QoS layer (e.g. it takes more time than `maxBackoffSleepTime` to write), it's likely the stream write will fail, and further retries to write the stream completely will also fail.

We want to add a retry layer in AtlasDB per thread, to avoid the write from failing completely.

**Implementation Description (bullets)**: Special case RLE in `CassandraRequestExceptionHandler`.

**Concerns (what feedback would you like?)**: Is this the right approach for this problem?

**Priority (whenever / two weeks / yesterday)**: EOW.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2844)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
@@ -34,6 +34,7 @@ import org.slf4j.helpers.MessageFormatter;
 import com.google.common.annotations.VisibleForTesting;
 import com.palantir.atlasdb.keyvalue.api.InsufficientConsistencyException;
 import com.palantir.atlasdb.keyvalue.cassandra.CassandraClientFactory.ClientCreationFailedException;
+import com.palantir.atlasdb.qos.ratelimit.RateLimitExceededException;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.remoting.api.config.service.HumanReadableDuration;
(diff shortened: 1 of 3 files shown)


nziebart:
I don't think that globally retrying on QoS exceptions is the right approach. The intent is that we automatically back-pressure up to a reasonable threshold (by sleeping). If you exceed this threshold, we want to shed the load to prevent an unbounded backup up of requests.

The stream store is a client of the core atlas layer, and in this case it's misbehaving (or multiPut has a bug in that it allows an overly large request and attempts to shove it in the DB all in parallel). 

Would it work to limit the parallelism in multiPut or make the stream store break up writes into multiple transactions? Alternatively, having the stream store code catch qos exceptions and retry seems reasonable (but again this probably requires breaking up writes into transactions).

fsamuel-bs:
It turns out there's already a way of writing a StreamStore in multiple transactions. And this is the recommended way to avoid having a long-running transaction, according to the docs. I'll add the retry mechanism on RLE specifically for the StreamStore there.

I'm not sure how much this will impact the throughput of large writes though. I'll run some perf test tomorrow to confirm that this is a viable solution.

Note that `multiPut` does have the behavior you specified — it will just write, with all the available threads — whatever you ask it to write, be it from a StreamStore or from another table. I'm not sure if adding throttling at that level is sensible, giving we already have a layer of throttling at the CassandraClient level. If we really want, we could add another throttle there though, per kvs `multiPut` call (while QoS is global), to throttle large writes.

Closing this for now, will open a follow up PR soon.

PR #2846, opened 2018-01-10
Generalize pre-commit conditions on transactions

sandorw:
**Goals (and why)**:
Generalize the current advisory locks pre-commit checks to allow other types of pre-commit checks/conditions. Concretely, I want to use this to validate that application schema versions are still valid before committing writes and after fetching read timestamps for migrations.

I've added (Abstract)ConditionAwareTransactionManager, which runs transactions with generalized conditions attached. All internal LockAware methods have been refactored to use this. 

Existing tests haven't been updated yet (and won't compile) and new tests are still missing pending an initial broad overview.

Initially, I'm looking for broad feedback on the changes to the TransactionManager code and structure - there are some slight behavior changes that need to be vetted.

I expect some (slight?) performance hit for code that uses the LockAware methods to run transactions with advisory locks and additional locks. As written, this is now two separate calls to the lock service to verify the locks separately, but could be batched together again with some additional code complexity.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2846)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -116,24 +110,15 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
     }
 
     @Override
-    public <T, E extends Exception> T runTaskWithLocksThrowOnConflict(
-            Iterable<HeldLocksToken> lockTokens,
-            LockAwareTransactionTask<T, E> task)
+    public <T, C extends PreCommitCondition, E extends Exception> T runTaskWithConditionThrowOnConflict(
```
fsamuel-bs: I think this will break a lot of products. Instead of changing the interface, can we add one more method with the PreCommitCondition?

fsamuel-bs: Or maybe just add new method and make `runTaskWithLocksThrowOnConflict` use it.

fsamuel-bs: Oh, nvm, I misread this, you did exactly what I suggested on `AbstractLockAwareTransactionManager`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractLockAwareTransactionManager.java
```diff
@@ -51,66 +37,16 @@
...
-                log.info("[{}] Retrying transaction", runId, e);
-            } catch (RuntimeException e) {
-                log.warn("[{}] RuntimeException while processing transaction.", runId, e);
-                throw e;
+                return task.execute(transaction, condition.getLocks());
             } finally {
-                if (lockToken != null) {
-                    getLockService().unlock(lockToken.getLockRefreshToken());
-                }
+                condition.cleanup();
```
sandorw: finally block shouldn't be needed here, as runTaskWithConditionWithRetry should presumably also include it - there's no special logic here for the lock case.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AdvisoryLockConditionSuppliers.java
```diff
@@ -0,0 +1,105 @@
...
+                RuntimeException ex = new LockAcquisitionException(
+                        "Failed to lock using the provided lock request: " + lockRequest);
+                log.warn("Count not lock successfully", ex);
+                ++failureCount;
+                if (failureCount >= NUM_RETRIES) {
+                    log.warn("Failing after {} tries", failureCount, ex);
+                    throw ex;
+                }
+                AbstractTransactionManager.sleepForBackoff(failureCount);
+                continue;
```
jeremyk-91: nit: Doesn't seem like there's anything after this statement, so this probably isn't needed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CombinedLocksCondition.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package com.palantir.atlasdb.transaction.impl;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.palantir.lock.HeldLocksToken;
+import com.palantir.lock.LockService;
+
+public class CombinedLocksCondition implements AdvisoryLocksCondition {
```
jeremyk-91: We're splitting the conditions up so as you mention we now make two separate calls to timelock for requests with advisory locks. Probably fine though, especially since the external condition shortcircuits if empty.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractConditionAwareTransactionManager.java
```diff
@@ -0,0 +1,95 @@
...
+                        SafeArg.of("failureCount", failureCount), e);
+            } catch (NotInitializedException e) {
+                log.info("TransactionManager is not initialized. Aborting transaction with runTaskWithRetry", e);
+                throw e;
+            } catch (RuntimeException e) {
+                log.warn("[{}] RuntimeException while processing transaction.", SafeArg.of("runId", runId), e);
+                throw e;
+            }
+            sleepForBackoff(failureCount);
+        }
```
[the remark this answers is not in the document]
jeremyk-91: discussed offline - yes, should belong here

sandorw:
Behavior changes:

- Previously, if an external lock (acquired outside the current transaction) expires while trying to commit a transaction, a retriable TransactionLockTimeoutException is thrown, despite retries not being possible. This specific case now throws a TransactionLockTimeoutNonRetriableException. This is needed as part of restructuring the condition logic to not special-case lock behavior.
- For transactions with both external locks and a transaction specific lock, verification of the locks will happen separately rather than as a single batch. This will add a performance hit in these cases.
- External locks are no longer checked for validity again after the transaction fails, before potentially retrying. Instead, a non-retriable exception is thrown whenever external locks are checked and any are expired. They are checked before transaction-specific locks to avoid unnecessary retries.
- Failures in acquiring locks for a transaction are now tracked and counted separately from other transaction failures. Technically, this means a given transaction can now fail and retry more times than before.
- Snapshot transaction constructor signature has changed, which could affect anyone extending classes in the transaction/txnManager space.
- Missing setupRunTaskWithLocksThrowOnConflict on SnapshotTransactionManager, which was public, but this could easily be added back in for convenience and wrap setupRunTaskWithConditionThrowOnConflict.

jeremyk-91:
I'm picking this up for another detailed review. `defaultGetRangesConcurrency` was added in between the time we branched off and develop here so we have conflicts, but they look reasonably straightforward to resolve.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -750,20 +769,152 @@ public void partiallyFilledRowsShouldBeVisible() {
...
+
+        try {
+            serializableTxManager.runTaskReadOnlyWithCondition(failsCondition,
+                    (tx, condition) -> tx.get(TABLE, ImmutableSet.of(TEST_CELL)));
+            fail();
+        } catch (TransactionFailedRetriableException e) {
+            // expected
+        }
+        assertThat(counter.intValue(), is(2));
+    }
```
jeremyk-91: nit: Seems like this could be two tests (read-only/general case), as well as the one above?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CombinedLocksCondition.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void throwIfConditionInvalid(long timestamp) {
+        externalLocksCondition.throwIfConditionInvalid(timestamp);
+        transactionLocksCondition.throwIfConditionInvalid(timestamp);
+    }
+
+    @Override
+    public void cleanup() {
+        externalLocksCondition.cleanup();
+        transactionLocksCondition.cleanup();
```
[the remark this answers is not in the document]
sandorw: Good find. I'll specify that the contract is that cleanup should not throw and should handle its state as best possible. Added a utility method for this in a new `PreCommitConditions` class and moved the NO_OP condition there from the interface.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
sandorw: I'm not sure it's worth it. There's either code duplication of the conditions or I move them up to non-static class variables and bloat the setup. I think this is a bit cleaner despite combining test cases.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/PreCommitConditions.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private PreCommitConditions() {}
+
+    /**
+     * Runs a cleanup task as best-effort, since we want to avoid throwing during cleanup.
+     */
+    public static void runCleanupTask(Runnable task) {
+        try {
+            task.run();
+        } catch (Throwable t) {
```
jeremyk-91: Not 100% sure if we want to swallow `Error`s - should this just be `Exception`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionLocksCondition.java
```diff
@@ -50,7 +50,7 @@ public void throwIfConditionInvalid(long timestamp) {
 
     @Override
     public void cleanup() {
-        lockService.unlock(heldLock.getLockRefreshToken());
+        PreCommitConditions.runCleanupTask(() -> lockService.unlock(heldLock.getLockRefreshToken()));
```
jeremyk-91: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/PreCommitConditions.java
sandorw: `Exception` should be fine, I'll switch to that and merge.

PR #2748, opened 2017-11-28
Cleanup Requirements, Part 3: Schema Metadata Service

jeremyk-91:
**Goals (and why)**:
- Provide an interface for actually storing metadata about a schema, which will eventually be readable by an external service.
- Support async initialization, since this depends on the underlying KVS.

**Implementation Description (bullets)**:
- The "schema" for this service (which uses a table backed by check and set) uses the schema name as its row name, and stores a serialized protobuf of the metadata.
- The service delegates to the appropriate calls on the raw KVS.

**Concerns (what feedback would you like?)**:
- Pitfalls with concurrency / if nodes disagree on schemas would be the main concerns here.
- Is exposing `getName()` on a schema a bad thing?
- What if we have schemas that have name clashes being registered at the same time?

**Where should we start reviewing?**: `SchemaMetadataService` and its tests.

**Priority (whenever / two weeks / yesterday)**: tail end of this week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2748)

PR #2747, opened 2017-11-28
Cleanup Requirements, Part 2: Computation from Schemas

jeremyk-91:
**Goals (and why)**:
- Determine the cleanup requirements for tables automatically from a schema definition. Part of broader Sweep work.

**Implementation Description (bullets)**:
- Compute the cleanup requirements from schemas. We don't use the special-cased stream store yet. Basically, if a user puts an *unspecified* cleanup task (or one specified as SYNC) we return SYNC; if a user puts a known ASYNC task we return ASYNC; if a table is a StreamStore metadata or index table then we return ASYNC (unless the user also added a SYNC task); otherwise we return NONE.

**Concerns (what feedback would you like?)**:
- Don't really like relying on the ordering of the enum
- Is the logic for addition correct? 

**Where should we start reviewing?**: `Schema.java`, though really most of this not-large PR is tests.

**Priority (whenever / two weeks / yesterday)**: this week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2747)

PR #2847, opened 2018-01-10
Cleanup Requirements Part 1.5: Stream Store Cleanup Metadata

jeremyk-91:
**Goals (and why)**:
- Be able to provide enough information for an external sweeper service to sweep stream stores.
- Yet, in doing the above, ensure that there is a smooth way to add more information to sweep if needed.

- Add a tagged-union-type of metadata options, which allows us to tell which type is present.
- Provide a mechanism for specifying multiple versions of the stream store cleanup metadata, in case we need to pass more parameters to it (one possibility is if we want to make block size configurable and want sweep prioritisation to incorporate block size as an input).

- Is the `import` of `ValueType` from `TableMetadataPersistence` a good idea?
- Is the v1 metadata actually enough to allow us to sweep things?
- Have we designed ourselves into a corner? I don't think so as we can always destroy the `StreamStoreCleanupMetadata` protobuf and make a new one.
- The code for parsing is a bit nasty, though it is consistent with what we have e.g. for table metadata.
- I used a Visitor so that the classes eventually making decisions in Nimbus (along the lines of `shouldExternalSweeperSweepThisTable` and `createCleanupTaskForTable`) don't have to do this side by side with the data. Is this unnecessarily complex?

**Where should we start reviewing?**: `SchemaMetadataPersistence.proto` first. The `SchemaDependentTableMetadata` class and the `cleanup` package probably next.

**Priority (whenever / two weeks / yesterday)**: soon - next steps for sweeper are to re-work #2747 and #2748 for this schema.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2847)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
@@ -63,11 +63,11 @@ public abstract class SchemaDependentTableMetadata implements Persistable {
     private static CleanupMetadata parseCleanupMetadata(
             SchemaMetadataPersistence.SchemaDependentTableMetadata message) {
         switch (message.getCleanupMetadataCase()) {
-            case ARBITRARYCLEANUPMETADATA:
+            case ARBITRARY_CLEANUP_METADATA:
                 return ArbitraryCleanupMetadata.hydrateFromProto(message.getArbitraryCleanupMetadata());
-            case NULLCLEANUPMETADATA:
+            case NULL_CLEANUP_METADATA:
                 return NullCleanupMetadata.hydrateFromProto(message.getNullCleanupMetadata());
-            case STREAMSTORECLEANUPMETADATA:
+            case STREAM_STORE_CLEANUP_METADATA:
                 return StreamStoreCleanupMetadata.hydrateFromProto(message.getStreamStoreCleanupMetadata());
             default:
                 throw new IllegalStateException("Unexpected type of cleanup metadata found: "
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
@@ -20,13 +20,16 @@ import org.immutables.value.Value;
 
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.palantir.atlasdb.protos.generated.SchemaMetadataPersistence;
-import com.palantir.atlasdb.protos.generated.SchemaMetadataPersistence.CleanupRequirement;
+import com.palantir.atlasdb.schema.cleanup.ArbitraryCleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.CleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.NullCleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.StreamStoreCleanupMetadata;
 import com.palantir.common.base.Throwables;
 import com.palantir.common.persist.Persistable;
 
 @Value.Immutable
 public abstract class SchemaDependentTableMetadata implements Persistable {
-    public static final Hydrator<SchemaDependentTableMetadata> HYDRATOR = input -> {
+    public static final Hydrator<SchemaDependentTableMetadata> BYTES_HYDRATOR = input -> {
         try {
             SchemaMetadataPersistence.SchemaDependentTableMetadata message =
                     SchemaMetadataPersistence.SchemaDependentTableMetadata.parseFrom(input);
@@ -36,7 +39,7 @@ public abstract class SchemaDependentTableMetadata implements Persistable {
         }
     };
 
-    public abstract CleanupRequirement cleanupRequirement();
+    public abstract CleanupMetadata cleanupMetadata();
 
     @Override
     public byte[] persistToBytes() {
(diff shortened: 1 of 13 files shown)


jeremyk-91:
Hmm, it seems I need to update some of the classes too.

jeremyk-91:
Okay. I think this should be ready for review now.

I've updated the description above for how to start and bits I'm less certain of, so I think it's still worth a read before diving in. Most of this PR is actually the protobuf and tests.

tboam:
@gsheasby to review

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
```diff
@@ -46,18 +49,56 @@
...
-            cleanupRequirement = message.getCleanupRequirement();
+        ImmutableSchemaDependentTableMetadata.Builder builder = ImmutableSchemaDependentTableMetadata.builder();
+        builder.cleanupMetadata(parseCleanupMetadata(message));
+        return builder.build();
+    }
+
+    private static CleanupMetadata parseCleanupMetadata(
+            SchemaMetadataPersistence.SchemaDependentTableMetadata message) {
+        switch (message.getCleanupMetadataCase()) {
+            case ARBITRARYCLEANUPMETADATA:
```
gsheasby: itshardtoparsethingswithoutspaces.

this reads like a [Banjo-Kazooie sandcastle code](http://banjokazooie.wikia.com/wiki/Cheat_Codes) - let's use `SCREAMING_SNAKE_CASE` for readability.

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/ArbitraryCleanupMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            return hydrateFromProto(SchemaMetadataPersistence.ArbitraryCleanupMetadata.parseFrom(message));
+        } catch (InvalidProtocolBufferException e) {
+            throw Throwables.propagate(e);
+        }
+    };
+
+    public static ArbitraryCleanupMetadata hydrateFromProto(
+            SchemaMetadataPersistence.ArbitraryCleanupMetadata unused) {
+        // For now, this is correct!
```
gsheasby: ...when will it not be? When that happens, will this comment become amusingly out of date?

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/NullCleanupMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        return this == other || (other != null && this.getClass() == other.getClass());
+    }
+
+    @Override
+    public int hashCode() {
+        return ArbitraryCleanupMetadata.class.hashCode();
```
gsheasby: copy-paste error here

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadata.java
```diff
@@ -0,0 +1,86 @@
...
+        if (streamStoreCleanupMetadata.hasV1Metadata()) {
+            StreamStoreCleanupV1Metadata v1Metadata = streamStoreCleanupMetadata.getV1Metadata();
+            if (v1Metadata.hasNumHashedRowComponents()) {
+                builder.numHashedRowComponents(v1Metadata.getNumHashedRowComponents());
+            }
+            if (v1Metadata.hasStreamIdType()) {
+                builder.streamIdType(ValueType.hydrateFromProto(v1Metadata.getStreamIdType()));
+            }
+        } else {
+            log.warn("Encountered stream store cleanup metadata without v1 metadata, which is unexpected.");
```
gsheasby: Are we exposing ourselves to any badness by not throwing an exception here? What will happen if Nimbus tries to sweep this table?

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/CleanupMetadataTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Test
+    public void visitorThrowsIfMethodUnimplemented() {
+        assertThatThrownBy(() -> new HashComponentsVisitor().visit(new ArbitraryCleanupMetadata()))
+                .isInstanceOf(UnsupportedOperationException.class);
+        assertThatThrownBy(() -> new HashComponentsVisitor().visit(new NullCleanupMetadata()))
+                .isInstanceOf(UnsupportedOperationException.class);
+    }
+
+    @Test
+    public void canInvokeMethodsOnVisitor() {
```
gsheasby: This test may be non-trivial to debug if it fails. Does `isEqualTo` return a useful error message if it fails (expected + actual would be enough).

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/NullCleanupMetadataTest.java
```diff
@@ -0,0 +1,50 @@
...
+package com.palantir.atlasdb.schema.cleanup;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.Test;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+
+public class NullCleanupMetadataTest {
```
gsheasby: You've written the same few tests twice. I think this is OK - presumably the `Null` and `Arbitrary` cases will soon be different.

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadataTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.stream.IntStream;
+
+import org.junit.Test;
+
+import com.palantir.atlasdb.protos.generated.SchemaMetadataPersistence;
+import com.palantir.atlasdb.table.description.ValueType;
+
+public class StreamStoreCleanupMetadataTest {
+    @Test
+    public void canSerializeAndDeserializePreservingHashedComponentsAndValueTypes() {
```
gsheasby: Perhaps this test and `CleanupMetadataTest.canInvokeMethodsOnVisitor` should be parameterised?

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/SchemaDependentTableMetadata.java
jeremyk-91: These names aren't unfortunately as poetic though.
(I've seen the sandcastle grid before, but not what the cheats actually are!)

- This table is lean; no need to clean. (Null)
- This is a Stream Store; I'll need to clean more. (Stream-Store)
- I don't know what to do - maybe you do. (Arbitrary)

The cases here are actually generated by the protobuf compiler. This would involve making the proto field names snake case as [that's how the code gen works](https://developers.google.com/protocol-buffers/docs/reference/java-generated#oneof). 

Maybe it's worth the trade-off of having a few lines of snake case in the protobuf? It looks like we're bound to have a bit of inconsistency somewhere, though in the protobuf it's guaranteed to only occur once.

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/ArbitraryCleanupMetadata.java
jeremyk-91: Fair enough - I'll rewrite this as `this is correct while the ArbitraryCleanupMetadata message has no fields`, which is what it is

jeremyk-91: There's also probably the same message for the Null one, as well

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/NullCleanupMetadata.java
jeremyk-91: Updated. Thanks

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadata.java
jeremyk-91: We _do_ actually throw an exception in this case, on `builder.build()` since required attributes are not set. It is a little awkward though.

I've added an explicit `throw IllegalStateException(...)` here.

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/NullCleanupMetadataTest.java
jeremyk-91: Interesting. They'll definitely diverge in tests for the relevant visitors, but in terms of these tests (which mostly center around serialization and object-equality) I'm a bit less certain that they will in the short run.

@tboam and I have talked about possibly implementing the queue for the `Arbitrary` case at a later date (where we'll need to know where to write the cells to be cleaned-up to), but that's not immediately planned!

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/CleanupMetadataTest.java
jeremyk-91: Yep, can confirm it does. You'll get something of this form if things aren't equal:

```
org.junit.ComparisonFailure: 
Expected :7
Actual   :0
```

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadataTest.java
jeremyk-91: Hmm, yeah they could be. I'll look into this (`0-2` is a common range for checking things on numbers of hashed components!)

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadata.java
gsheasby: Thanks - without the exception here, it looks like we're trying to skate past the issue.

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/CleanupMetadataTest.java
gsheasby: That error is good enough for me!

atlasdb-client/src/test/java/com/palantir/atlasdb/schema/cleanup/StreamStoreCleanupMetadataTest.java
jeremyk-91: The main benefit of doing this would be that you get to see which type(s) and number(s) of hashed components failed, so could be worth doing. There is a little bit more complexity in the test code unfortunately but I think it's still pretty manageable.

jeremyk-91: Done. I didn't bother to do this for the cleanup metadata one as there are only three possible values and the complexity isn't so nice - but can imagine the serialization one could be a horrible mess as it goes over 33 (!) possibilities

PR #2848, opened 2018-01-10
Refactor Cassandra exception handler

fsamuel-bs:
**Goals (and why)**: This will make further changes easier, e.g. #2844.

**Implementation Description (bullets)**: Group exceptions in groups. Map groups to desired behavior.

**Concerns (what feedback would you like?)**: Did I break anything? My intention was to not change any behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2848)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
@@ -54,13 +54,13 @@ class CassandraRequestExceptionHandler {
         this.blacklist = blacklist;
     }
 
-    <K extends Exception> void handleRequest(
+    <K extends Exception> void handleExceptionFromRequest(
                 RetryableCassandraRequest<?, K> req,
                 InetSocketAddress hostTried,
                 Exception ex)
             throws K {
         req.triedOnHost(hostTried);
-        this.handleExceptionInternal(req, hostTried, ex);
+        this.handleNumberOfAttempts(req.getNumberOfAttempts(), hostTried, ex);
 
         if (isRetriableWithBackoffException(ex)) {
             // And value between -500 and +500ms to backoff to better spread load on failover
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -53,109 +54,162 @@
...
-                    log.info("Error occurred talking to cassandra. Attempt {} of {}.",
-                            SafeArg.of("numTries", req.getNumberOfAttempts()),
-                            SafeArg.of("maxTotalTries", maxTriesTotal),
-                            ex);
-                }
-                if (isConnectionException(ex) && req.getNumberOfAttempts() >= maxTriesSameHost.get()) {
-                    blacklist.add(host);
-                }
-            }
+    private <K extends Exception> K logAndReturnException(int numberOfAttempts, Exception ex) {
```
tboam: Can we change this to logAndThrowException and then actually throw the Exception.  This way callers don't have to remember to do the throwing (but the code is clearer if they do throw the result still).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -166,4 +220,16 @@ static boolean isFastFailoverException(Throwable ex) {
             + " Error writing to Cassandra socket."
             + " Likely cause: Exceeded maximum thrift frame size;"
             + " unlikely cause: network issues.";
+
+    @VisibleForTesting
+    enum Backoff {
```
tboam: Right now this could just be a `boolean`, what's the use case you foresee where we would want different Backoff types?

tboam:
@fsamuel-bs to look at review comments and check why builds are failing.

fsamuel-bs:
I've replaced the `Backoff` enum with a boolean and addressed the comments.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
fsamuel-bs: I also had a `LONG` behavior which was used for QoS exceptions, and would backoff for 10s. This is not needed anymore, since we're not going to retry QoS exceptions at this level, so converted back to a boolean.

PR #2849, opened 2018-01-11
Feature/prepared cql queries

gsheasby:
Use a more performant* CQL query in the read stage of sweep

The query that this replaced had a significant perf regression for tables with wide rows, which this query largely addresses. However, the increased number of round trips leads to a significant slowdown for tables with very narrow rows (e.g. sequential tables).

The suggestion is to merge this in and release a beta version that we can push to the places with OOM concerns, and to follow up with more performant queries.

Perf data (seconds per full table scan, batch size 1000):
Clean tables (1 column, 1 overwrite):
0.72: 25s
develop: 99s (298% worse)
prepared: 810s (3170% worse)

"Average" tables (10 columns, 3 overwrites)
0.72: 3s
develop: 30s (886% worse)
prepared: 11s (263% worse)

"Dirty" tables (1 column, 100 overwrites)
0.72: 8.8s
develop: 185s (2000% worse)
prepared: 15.5s (77% worse)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2849)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
@@ -98,12 +98,6 @@ public class CqlExecutorImpl implements CqlExecutor {
         return result;
     }
 
-    private Arg<String> keys(List<byte[]> rowsAscending) {
-        return UnsafeArg.of("keys",
-                rowsAscending.stream().map(CqlExecutorImpl::getKey)
-                        .collect(Collectors.joining(",")));
-    }
-
     /**
      * Returns a list of {@link CellWithTimestamp}s within the given {@code row}, starting at the (column, timestamp)
      * pair represented by ({@code startColumnInclusive}, {@code startTimestampExclusive}).
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
@@ -53,7 +53,7 @@ public class CqlExecutorImpl implements CqlExecutor {
 
     public interface QueryExecutor {
         CqlResult execute(CqlQuery cqlQuery, byte[] rowHintForHostSelection);
-        CqlPreparedResult prepare(ByteBuffer query, Compression compression);
+        CqlPreparedResult prepare(ByteBuffer query, byte[] rowHintForHostSelection, Compression compression);
         CqlResult executePrepared(int queryId, List<ByteBuffer> values);
     }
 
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorTest.java
```diff
@@ -80,6 +90,20 @@ public void getTimestampsWithinRow() {
         verify(queryExecutor).execute(argThat(cqlQueryMatcher(expected)), eq(ROW));
     }
 
+    private ArgumentMatcher<ByteBuffer> byteBufferMatcher(String expected) {
```
hsaraogi: 👍

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorTest.java
```diff
@@ -57,16 +60,23 @@ public void before() {
...
+        when(queryExecutor.executePrepared(eq(1), any())).thenReturn(result);
     }
 
     @Test
     public void getTimestampsForGivenRows() {
         String expected = "SELECT key, column1, column2 FROM \"foo__bar\""
-                + " WHERE key IN (0x0102,0x0509) LIMIT 100;";
+                + " WHERE key = ? LIMIT 100;";
 
         executor.getTimestamps(TABLE_REF, ImmutableList.of(ROW, END_ROW), LIMIT);
```
hsaraogi: We should sanity check the return values here too. I believe we have ETE tests for sweep that would ensure that the loading is done correctly. Maybe worth running the large internal product tests once as they are more comprehensive and compare bytes etc.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
```diff
@@ -63,9 +71,28 @@
...
             List<byte[]> rowsAscending,
             int limit) {
-        String selQuery = "SELECT key, column1, column2 FROM %s WHERE key IN (%s) LIMIT %s;";
-        CqlQuery query = new CqlQuery(selQuery, quotedTableName(tableRef), keys(rowsAscending), limit(limit));
-        return executeAndGetCells(query, rowsAscending.get(0), CqlExecutorImpl::getCellFromRow);
+        String preparedSelQuery = String.format("SELECT key, column1, column2 FROM %s WHERE key = ? LIMIT %d;",
+                quotedTableName(tableRef).getValue(),
+                limit);
+        ByteBuffer queryBytes = ByteBuffer.wrap(preparedSelQuery.getBytes(StandardCharsets.UTF_8));
+        CqlPreparedResult preparedResult = queryExecutor.prepare(queryBytes, Compression.NONE);
```
fsamuel-bs: Let's make the prepare and the execute use the same CassandraClient — Cassandra doesn't distribute the prepared statements by default.

If the Cassandra node restarts between the prepare and the execute statements, it will lose all prepared statements. On an extreme scenario, this might happen between the prepare and execute statements. If we tell a Cassandra node to execute a statement it doesn't know about, I think it will throw a `InvalidRequestException` (complete guess — we should probably test this). So we should probably surround the execute statement with try-catch. We should then retry to prepare and to execute.

There're some interesting knowledge about the prepare statements here:
https://docs.datastax.com/en/developer/java-driver/3.1/manual/statements/prepared/#preparing-on-multiple-nodes

gsheasby:
Stress-test results: 20 successive iterations of sweep completed successfully, taking 11 minutes, with about 2 seconds total GC time per C* node.

PR #2850, opened 2018-01-11
Fail Fast on Unclear TimeLock Install/Runtime Configuration Pairs

jeremyk-91:
**Goals (and why)**:
- Fail fast if we are presented with an install/runtime configuration pair that isn't very clear as far as timelock is concerned. Specifically, this is with Install configured _not_ to use TimeLock, but with Runtime configured with a timelock block.

Previously, we would have ignored the timelock block in runtime and started up anyway with leader/embedded/remote, based on what the install config said.

- When creating the timestamp and lock services, throw if you have (leader/remote) in install AND a timelock block in runtime.
- Accept embedded with a timelock block in runtime as timelock (see discussion in comments)
- Note: Even if you _don't_ have a timelock block in runtime, we still need to be able to run with a timelock in install (for back-compat reasons).

- I think the state of the world after this change is better than before. Do you agree?
- This PR does not handle the cases where the initial config makes sense (i.e. embedded/leader/remote in install with an empty timelock block in runtime) but a different runtime config with a timelock block is loaded in later on.

**Where should we start reviewing?**: `TransactionManagers.java`, the rest is tests and release notes.

**Priority (whenever / two weeks / yesterday)**: not too high. 2 weeks?

@chrisalice for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2850)


tboam:
@hsaraogi to review

hsaraogi:
The PR looks good overall, but I think we should use a different approach here.

Instead of forcing users to specify `timelock: {}` in the install config, we can assume timelock if the runtimeconfig has `timelock`. Of course, we need to have the checks in place to ensure no leader/timestamp/lock blocks are specified along with `timelock` in runtime. Will be happy to merge this in for the time-being if the suggested approach seems high-lift.

jeremyk-91:
Ack on the possibility of having nothing in Install *AND* timelock in the runtime config indicate a decision to use timelock. @tboam any thoughts/objections to this?

chrisalice:
This will break badly if someone has a script that passes default runtime config in (as Multipass used to). Instead of refusing to run, it will corrupt the database.

jeremyk-91:
@chrisalice The default runtime config does not have a timelock block, so actually I think we're safe then.

Here's a more detailed table with how the resolution works now:

```
+---------------+--------------+-------------+
|    INSTALL    | RT: TimeLock | RT: Nothing |
+---------------+--------------+-------------+
| Embedded (EM) | EM -> **     | EM -> EM    |
| Leader   (LD) | LD -> FAIL   | LD -> DL    |
| Remote   (RM) | RM -> FAIL   | RM -> RM    |
| TimeLock (TL) | TL -> TL     | TL -> TL    |
+---------------+--------------+-------------+
```
Table values have the form old behaviour -> new behaviour.

I think we're pretty safe in most cases: if the runtime config has no timelock block then this isn't intended to change behaviour.

Question is what `**` should be (original version has `FAIL` and @hsaraogi's suggestion is that it could be `TL`). I guess the main argument about `TL` then is if we subsequently change the runtime config to include a timelock block and aren't careful about that then there might be an issue?

Even then I think we're protected because of the shim that invalidates the timestamp bound before being willing to read anything from timelock though.

chrisalice:
I was actually talking about @hsaraogi's suggestion.

chrisalice:
Apparently Atlas won't let you access an external-timelock-controlled db without going via timelock so I retract my objection :)

jeremyk-91:
Another issue with the implementation is someone could write a bad config and have it pass for now, but have this config cause the service to fail to start when coming up again. Following conventions of internal service specification having a way to fail some kind of healthcheck would be useful.

jeremyk-91:
@hsaraogi I've updated the PR to accept not having `timelock: {}` and have smoke tested this locally, so should be ready for review again. Aware that support load is a bit high this week, so not too much of a rush here!

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -518,17 +519,36 @@ static LockAndTimestampServices createRawInstrumentedServices(
...
+        if (timestampOrLockOrLeaderBlockPresent(config) && initialRuntimeConfig.timelockRuntime().isPresent()) {
+            throw new IllegalStateException("Found a service configured not to use timelock, with a timelock"
+                    + " block in the runtime config! This is unexpected. If you wish to use non-timelock services,"
+                    + " please remove the timelock block from the runtime config; if you wish to use timelock,"
+                    + " please remove the leader, remote timestamp or remote lock configuration blocks.");
+        }
+    }
+
+    private static boolean timestampOrLockOrLeaderBlockPresent(AtlasDbConfig config) {
+        return config.timestamp().isPresent() || config.lock().isPresent() || config.leader().isPresent();
```
hsaraogi: nit: we should either check for only `timestamp` or `timestamp` && `lock`. This kinda conflicts with the top level config check.

docs/source/configuration/external_timelock_service_configs/timelock_client_config.rst
```diff
@@ -90,9 +90,10 @@ We support live reloading of the ``ServerListConfiguration`` for TimeLock. This
 Note that if this block is present, then the ``ServerListConfiguration`` in the install configuration will be ignored.
 
 Also, although we support live-reloading of the server configuration, AtlasDB needs to know at install time that it
-should talk to TimeLock - thus, the install configuration must contain a ``timelock`` block (even if said block is
-possibly empty). Specifying an empty block in the configuration is done in the same way as specifying an empty block in
-YAML i.e. ``{}``.
+should talk to TimeLock. In versions of Atlas before 0.74.0, this is done by specifying an empty YAML object
+for the timelock block in the install configuration (i.e. ``timelock: {}``); in more recent versions, the presence
+of a ``timelockRuntime`` block is enough - *however*, it must already be present at install time, or AtlasDB
+will attempt to start up using embedded services.
```
hsaraogi: 👍

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -426,6 +432,48 @@ public void timeLockMigrationReportsReadyIfMigrationNotNeeded() {
         assertTrue(TransactionManagers.timeLockMigrationCompleteIfNeeded(lockAndTimestampServices));
     }
 
+    @Test
+    public void usesTimeLockIfInstallConfigIsEmbeddedButInitialRuntimeConfigContainsTimeLockBlock() {
```
hsaraogi: `InstallConfigIsUnspecified`

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -447,15 +495,28 @@ private void setUpForLocalServices() throws IOException {
...
-        setupLeaderBlockInConfig();
+        setUpLeaderBlockInConfig();
+    }
+
+    private void setUpTimeLockBlockInInstallConfig() {
+        when(config.timelock()).thenReturn(Optional.of(mockClientConfig));
+    }
+
+    private void setUpTimeLockBlockInRuntimeConfig() {
+        when(runtimeConfig.timelockRuntime()).thenReturn(Optional.of(mockTimelockRuntimeConfig));
```
hsaraogi: nit: `mockTimelockRuntimeConfig` is not a mock. Also, might be a good idea to inline this variable.

docs/source/release_notes/release-notes.rst
```diff
@@ -105,6 +105,21 @@ develop
...
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/2850>`__)
+
+    *    - |improved| |userbreak|
+         - AtlasDB users can now specify the usage of TimeLock Server purely by including a TimeLock block in the initial runtime configuration.
+           Previously, AtlasDB users would need to specify that they were using TimeLock in the install configuration, possibly with an empty object (``timelock: {}``).
+           This is a change from previous behaviour in cases where users specified an embedded install configuration but a TimeLock block in the runtime configuration; previously, the embedded configuration would have been selected, while now the TimeLock configuration will be selected.
+
+           Also, users with scripts that depend on supplying a default runtime configuration may need to be careful to ensure that TimeLock configuration is preserved when such scripts are run.
+           That said, AtlasDB will fail to start if trying to access a database that uses TimeLock without going through TimeLock, so we don't think there is a risk of data corruption.
```
hsaraogi: "access a database that uses Timelock" is confusing as timelock does not technically use a database.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: Fair enough. I'd rather not check _only_ tor timestamp - think `timestamp && lock` is clearer in terms of intention, so will switch to that.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
jeremyk-91: Can do - indeed this is the case after this PR

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
jeremyk-91: Ack - renamed. I think this was a hangover from a previous version where I tried to use mocks.

docs/source/release_notes/release-notes.rst
jeremyk-91: I've rewritten this to be `That said, AtlasDB will fail to start if trying to access a key-value service where TimeLock has been used as a source of timestamps without going through TimeLock, so we don't think there is a risk of data corruption.` It's still a bit clunky - suggestions to improve this would be welcomed 😄

PR #2853, opened 2018-01-11
Retry on RLE on storeStream

fsamuel-bs:
**Goals (and why)**: When doing StreamStore writes, we break the stream in 1MB blocks and call kvs.multiPut. MultiPut groups up blocks in groups of 10MBs and spawns a thread to write such group to CKVS.

If the stream is big, as the internal search product case, we will use all CKVS threads to attempt such write. This causes problems to the underlying DB. We want to throttle the write at the QoS level.

If one of the threads that attempt the write fails at the QoS layer (e.g. it takes more time than maxBackoffSleepTime to write), it's likely the stream write will fail, and further retries to write the stream completely will also fail.

**Implementation Description (bullets)**: Add a retry mechanism at the StreamStore write level. We should direct consumers to write large stores by using `table.storeStream(InputStream)`, as the [docs suggest](https://palantir.github.io/atlasdb/html/schemas/streams.html#storing-streams). This will break down each block to be written by a separate transaction, being a separate `multiPut` call. 

**Concerns (what feedback would you like?)**: Is this a perf regression? I'll be running benchmarks to compare both options.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractPersistentStreamStore.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractPersistentStreamStore.java
@@ -20,6 +20,7 @@ import java.io.InputStream;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nullable;
 
@@ -31,10 +32,12 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.io.ByteStreams;
 import com.google.common.io.CountingInputStream;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.protobuf.ByteString;
 import com.palantir.atlasdb.encoding.PtBytes;
 import com.palantir.atlasdb.protos.generated.StreamPersistence.Status;
 import com.palantir.atlasdb.protos.generated.StreamPersistence.StreamMetadata;
+import com.palantir.atlasdb.qos.ratelimit.RateLimitExceededException;
 import com.palantir.atlasdb.transaction.api.Transaction;
 import com.palantir.atlasdb.transaction.api.TransactionManager;
 import com.palantir.atlasdb.transaction.api.TransactionTask;
@@ -45,6 +48,9 @@ import com.palantir.util.crypto.Sha256Hash;
 
 public abstract class AbstractPersistentStreamStore extends AbstractGenericStreamStore<Long>
         implements PersistentStreamStore {
+
+    public static final int NUMBER_OF_TIMES_TO_ATTEMPT_TO_STORE_BLOCK = 6;
+
     protected AbstractPersistentStreamStore(TransactionManager txManager) {
         super(txManager);
     }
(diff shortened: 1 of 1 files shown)


fsamuel-bs:
#2884

fsamuel-bs:
The benchmark results are out:
```
"benchmark":"StreamStorePutBenchmarks#storeStreamWithTransaction-CASSANDRA",
"p50":1481637.8879999998,
"p90":1632842.5472,
"p99":1809842.176
"units":"MICROSECONDS",

"benchmark":"StreamStorePutBenchmarks#storeStreamWithoutTransaction-CASSANDRA",
"p50":8359247.872,
"p90":9059696.64,
"p99":9059696.64
```

The benchmark consisted of writing 60MB through the two methods. We see a 8x slowdown when creating multiple transactions. This benchmark is not very representative of field results though, since this was ran with embedded timelock service and single-node Cassandra cluster.

Will run benchmarks more field-representative now.

tboam:
Doing all writes inside of the transaction gets more parallelisation benefits as we write the maximum amount at once.  Running each block in a separate transaction offers reliability guarantees.  When doing 2 separate transactions do they write to the same table in a way which will conflict?  If not then we can run multiple transactions in parallel.

fsamuel-bs:
Yep, I thought of parallelizing the writes, as I think they won't conflict. But I'm reluctant to add one more ExecutorService to the codebase, as it generates memory pressure on the client side, and wanted to benchmark it better with a real scenario. Should we just add an ExecutorService in this case?

tboam:
Memory is probably cheap enough that it is ok to add.  We certainly verify there won't be conflicts (otherwise the parallel transactions will just run in serial) and see if we get the expected perf gain.

We can always use the ExecutorService if users configure more that 1 thread for streamStoreParallelisation or similar config value

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractPersistentStreamStore.java
```diff
@@ -45,6 +48,9 @@
 
 public abstract class AbstractPersistentStreamStore extends AbstractGenericStreamStore<Long>
         implements PersistentStreamStore {
+
+    public static final int NUM_RETRIES_ON_QOS_EXCEPTION = 6;
```
tboam: For the first use case it may make sense to back off infinitely on RLE.  Certainly value in making this configurable.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractPersistentStreamStore.java
```diff
@@ -200,12 +206,24 @@ protected void storeBlockWithNonNullTransaction(@Nullable Transaction tx, final
...
+
+        for (int attempt = 1; attempt <= NUM_RETRIES_ON_QOS_EXCEPTION; attempt++) {
+            try {
+                txnMgr.runTaskThrowOnConflict(
+                        (TransactionTask<Void, RuntimeException>) t1 -> {
+                            storeBlock(t1, id, blockNumber, bytesToStore);
+                            return null;
+                        });
+            } catch (RateLimitExceededException ex) {
+                Uninterruptibles.sleepUninterruptibly(attempt, TimeUnit.SECONDS);
```
tboam: Would be good to add some logging and maybe a maximum sleep duration if we make the number configurable.

fsamuel-bs:
It seems it's not parallelizable. :/

On `touchMetadataWhileStoringForConflicts`, we do:
```
        Builder builder = StreamMetadata.newBuilder(metadata);
        builder.setLength(blockNumber * BLOCK_SIZE_IN_BYTES + 1);
        metaTable.putMetadata(row, builder.build());
```

This is, we update the size of the stream we're storing by doing `blockNumber * BLOCK_SIZE_IN_BYTES + 1`. Note that the metadata table is `id -> metadata`, where `id` is the Stream ID. This means that all writes would conflict as they're trying to store it on the Metadata table.

gsheasby:
Pinging an old PR. Seems from the previous comment that we're trying to parallelise a non-parallelisable thing. If so, why has this been left open?

fsamuel-bs:
I think the issue that generated this PR — our indexing service causing Cassandra perf issues — still remains. I think the solution is to throttle StreamStore writes in some way, which this PR attempted to do, but not in the way this PR is currently doing.

Probably worth talking offline, but one of the solutions would be to create a new stream store endpoint, which would write X blocks per transaction, instead of the current world, where we:
1. write all blocks in a single transaction, or
2. write 1 block per transaction.

PR #2857, opened 2018-01-12
Add metric for the throttling applied by the Qos service

hsaraogi:
**Goals (and why)**: Metrics for read and write limit multipliers

**Implementation Description (bullets)**: add gauges.

I have two separate metrics because I think going forward we might change the read and write multipliers to not be the same.

**Where should we start reviewing?**: one file.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2857)

### DIFF
--- a/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/QosResource.java
+++ b/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/QosResource.java
@@ -25,19 +25,20 @@ public class QosResource implements QosService {
     private final ClientLimitMultiplier clientLimitMultiplier;
     private static volatile double readLimitMultiplier = 1.0;
     private static volatile double writeLimitMultiplier = 1.0;
+    private static final MetricsManager metricsManager = new MetricsManager();
 
     public QosResource(QosClientConfigLoader qosClientConfigLoader, ClientLimitMultiplier clientLimitMultiplier) {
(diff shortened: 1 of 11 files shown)


qos-service-impl/src/main/java/com/palantir/atlasdb/qos/QosResource.java
```diff
@@ -18,27 +18,42 @@
...
+        new MetricsManager().registerMetric(QosResource.class, "readLimitMultiplier", this::getReadLimitMultiplier);
+        new MetricsManager().registerMetric(QosResource.class, "writeLimitMultiplier", this::getWriteLimitMultiplier);
     }
 
     @Override
     public long readLimit(String client) {
         QosClientLimitsConfig qosClientLimitsConfig = qosClientConfigLoader.getConfigForClient(client);
-        return (long) (clientLimitMultiplier.getClientLimitMultiplier(qosClientLimitsConfig.clientPriority())
-                * qosClientLimitsConfig.limits().readBytesPerSecond());
+        readLimitMultiplier = this.clientLimitMultiplier.getClientLimitMultiplier(
```
sandorw: This is awkward with the current `CassandraMetricsClientLimitMultiplier` implementation, since any high priority clients will update the limit back to 1.0. We'll likely see some sawtooth metrics if there are clients at different priority levels while throttling is ongoing.

PR #2744, opened 2017-11-27
Fix/avoid corner case

gsheasby:
**Goals (and why)**: 
- Avoid sweep getting stuck in an infinite loop (why? because infinite loops are bad)
- Actually run the test that would have caught this in circle again

**Implementation Description (bullets)**:
- Add memorySensitiveTest back to the least-stressed container
- Use RangeRequests.nextLexicographicName to avoid successive loop iterations containing the same row, which is a big problem if this is the only row remaining and it only contains tombstones, because we'll never get any farther.

**Concerns (what feedback would you like?)**:
- Do we need the extra tests? Are they at the right specificity?
- Do we need all the timeouts on the sweep tests? None of them take longer than 12 seconds (with batch size = false) on my IntelliJ.
- Is the release note messaging appropriate?

**Where should we start reviewing?**: GetCellTimestamps, which turns out to be the first file on the PR.

**Priority (whenever / two weeks / yesterday)**: ASAP

**The gory details (copied from the message of the commit that actually fixes the thing)
Fix an edge case that caused an infinite loop in
AbstractSweepTaskRunnerTest.testSweepLatestDeletedThrough.

If the range contains a row that has no columns (including if the row's
only column was deleted), the Thrift rowGetter query will return that
row, although the list of columns would be empty - the row's still
there, but it's tombstoned. Therefore, determineSafeRangeEndInclusive
would return that row's key as a safe range end (and not
Optional.Empty). So we run CqlExecutor.getTimestamps, get an empty
list, and iterate again through the loop.

Previously, we set rangeStart to the previous rangeEnd. Both ends of
the CqlExecutor.getTimestamps query are inclusive, so on the next
iteration we'd get the empty row again, and rangeEnd and rangeStart
would never change.

Using RangeRequests.nextLexicographicName avoids fetching the same
empty row again, so we can continue making forward progress.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2744)

PR #2732, opened 2017-11-21
avoid tombstoned rows

nziebart:
**Goals (and why)**:
Addresses part of https://github.com/palantir/atlasdb/issues/2691 by paging over tombstones, rather than making Cassandra do an unbounded range scan.

**Implementation Description (bullets)**:
- Impose an upper bound on the CQL range scan used for sweep.
- Use thrift to determine a safe upper bound (see comments in `GetCellTimestamps` for an explanation of why this works.

**Concerns (what feedback would you like?)**:

**Where should we start reviewing?**:

**Priority (whenever / two weeks / yesterday)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2732)

PR #2610, opened 2017-11-02
fix sweep batch size

nziebart:
**Goals (and why)**:
Remove the hackery around Cassandra's sweep batch size, now that we can safely increase it.

**Implementation Description (bullets)**:

**Concerns (what feedback would you like?)**:

**Where should we start reviewing?**:

**Priority (whenever / two weeks / yesterday)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2610)

PR #2858, opened 2018-01-12
Fix/sweep results cells examined (#2830)

mswintermeyer:
* Add test demonstrating SweepResults.getCellTsPairsExamined incorrect

* Fix SweepResults.getCellTsPairsExamined

* Fix CellsToSweepPartitioningIteratorTest use of cellsExaminedSoFar

* update BatchOfCellsToSweep.numCellTsPairsExamined javadoc

* Add release notes

* Fix checkstyle

* Release note: explain the issue that was fixed

**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2858)


tboam:
@gmaretic to review and look at test failures

gmaretic:
This PR is broken because the fix assumes the changes in https://github.com/palantir/atlasdb/pull/2546. Is backporting this high priortiy?

mswintermeyer:
Thanks @gmaretic for figuring out what commit I was missing! I attempted to fix the errors with cherry picking that too. The only conflicts were in CassandraKeyValueServiceImpl. I believe that fixed my test failures, and don't think it added new ones. The error in internal circleci passes for me locally. Regarding priority, it's a P2. No one clamoring for it yet, but a PG LTS uses this version of atlas, so it'd be nice to have sweep info be accurate before someone needs sweep critically.

gsheasby:
Is the problem that this fixes (inaccuracy of cells examined) even known to be an issue on this version of Atlas?

gsheasby:
"* Add test demonstrating SweepResults.getCellTsPairsExamined incorrect"

Did you check that this test also demonstrates `getCellTsPairsExamined` was incorrect to begin with on this version?

gmaretic:
@gsheasby I checked, `testSweepBatchesInDifferentRows` was failing on 0.61.0. The other failing tests were actually passing since the logic that broke them (before they were fixed again) was introduced in the missing commit so the fix actually broke them on the PR. 
Now looks good to me, @gsheasby since you have more SA on sweep, is there anything else that needs to be fixed on this version of sweep? Are we confident cherry picking this commit did not introduce a bug that we later fixed?

gsheasby:
So we've advanced in the chain of sweep PRs, but might still want to go further?

#2610 - ups the default batch size (might not be relevant for the internal product?)
#2732 - fixes an edge case that caused us to page through a lot of deleted data for thoroughly swept tables
#2744 - fixes a corner case introduced by the edge case fix, which results in sweep getting stuck in an infinite loop

Then we'd be down to the OOM fix PRs that are still in progress.

gmaretic:
That makes sense, but seems like we're going down a rabbit hole. Can you weigh in on how we should proceed @tboam ?

tboam:
Definitely want to address the counting issue on the LTS branch as it is broken and confusing.

I agree with cherry picking the commits Glenn mentions as important sweep fixes too.  The sweep OOM fixes should be backported, but I'm happy to wait until we've verified the fixes in the field (demand for OOM fixes is lower on the LTS branch as it is less likely to be affected by Cassandra issues).

Checking the git log for the main sweep classes should reveal any other major changes that should be backported.

gsheasby:
How about the demand for fixing the edge/corner cases? They also only affect Cassandra, so a smaller proportion of LTS consumers will care - but on the other hand, they affect thorough tables, which are more prevalent on the large internal product.

I'd be happy to merge this as-is, and we can continue the chain of backports in later PRs (tiny tasks?)

PR #2860, opened 2018-01-15
Feature/parallel sweep

gsheasby:
Tagging @tpetracca and @schlosna for Cassandra/parallel code feedback.

**Goals (and why)**: Follow-up to #2849, which itself was a follow-up to #2826. 
Speed up sweep by fetching rows from Cassandra in parallel. This improves the speed by 40-80% compared to develop.

Currently, develop exhibits a 24x regression for the `fullTableScanCleanConservative` benchmark compared to the 0.72 release; this PR reduces the regression to 3.4x.
For the `fullTableScanDirtyConservative` benchmark, this PR is 3% faster than 0.72, whereas develop was 68% slower.

When we consider that deployments should be able to run sweep with a batch size of 1024 (compared to the current default of 128), the read stage will be about 3x faster for regular tables, and only 19% slower for clean (i.e. very narrow/sequential) tables.

* Create an ExecutorService when creating a CqlExecutor
* Use 16 threads (currently hard-coded) to send off CQL queries to Cassandra in parallel

**Concerns (what feedback would you like?)**: This means sweep will use 16 clients from Cassandra's client pool instead of 1. Are we concerned by this increase in utilisation?
I believe that, if sweep exhausts the client pool, we will simply wait on the Atlas side for new clients; however, this could lead to regular queries becoming slower.

**Where should we start reviewing?**: CqlExecutorImpl

**Priority (whenever / two weeks / yesterday)**: ASAP - this is blocking a "proper" AtlasDB release.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2860)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
@@ -111,6 +111,7 @@ public class CqlExecutorImpl implements CqlExecutor {
                 result.addAll(CqlExecutorImpl.getCells(CqlExecutorImpl::getCellFromRow, cqlResult));
 
                 if (result.size() > limit) {
+                    cancelFutures(futures.subList(i + 1, futures.size()));
                     break;
                 }
 
@@ -131,6 +132,10 @@ public class CqlExecutorImpl implements CqlExecutor {
         Callable<CqlResult> task = () -> queryExecutor.executePrepared(queryId,
                 ImmutableList.of(ByteBuffer.wrap(row)));
         return executor.submit(task);
+    }
+
+    private void cancelFutures(List<Future<CqlResult>> futures) {
+        futures.forEach(f -> f.cancel(true));
     }
 
     /**
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
@@ -56,7 +56,8 @@ import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 
 public class CqlExecutorImpl implements CqlExecutor {
-    private QueryExecutor queryExecutor;
+    private final QueryExecutor queryExecutor;
+    private final ExecutorService executor;
 
     public interface QueryExecutor {
         CqlResult execute(CqlQuery cqlQuery, byte[] rowHintForHostSelection);
@@ -64,13 +65,15 @@ public class CqlExecutorImpl implements CqlExecutor {
         CqlResult executePrepared(int queryId, List<ByteBuffer> values);
     }
 
-    CqlExecutorImpl(CassandraClientPool clientPool, ConsistencyLevel consistency) {
+    CqlExecutorImpl(ExecutorService executor, CassandraClientPool clientPool, ConsistencyLevel consistency) {
         this.queryExecutor = new QueryExecutorImpl(clientPool, consistency);
+        this.executor = executor;
     }
 
     @VisibleForTesting
     CqlExecutorImpl(QueryExecutor queryExecutor) {
         this.queryExecutor = queryExecutor;
+        this.executor = PTExecutors.newFixedThreadPool(16);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
```diff
@@ -81,23 +89,43 @@
...
+        }
+
+        try {
+            for (int i = 0; i < futures.size(); i++) {
+                Future<CqlResult> future = futures.get(i);
+                CqlResult cqlResult = future.get();
+                result.addAll(CqlExecutorImpl.getCells(CqlExecutorImpl::getCellFromRow, cqlResult));
+
+                if (result.size() > limit) {
+                    break;
```
tboam: We should cancel any remaining Futures before returning.  This way we won't execute queries for which we're never going to look at the results.  Need to check if the `queryExecutor` or some other part of the future checks for cancellation.

gsheasby: Added the cancellations (pushing in a few mins). The `queryExecutor` pretty much delegates to `clientPool.runWithRetryOnHost`, so this may be interrupted - are there any concerns about that?

tboam: Concern was if nothing checked it then work would still happen.  But I imagine it happens before the task starts running and when the query is sent and received over the network, which is probably enough.

fsamuel-bs:
I ran benchmarks with #2869, and here are the results:

```
         OOM Sweep  #2860    #2869
Clean    21.239s    87.610s  18.421s
Average  2.460s     2.423s   1.705s
Dirty    6.313s     3.919s   4.605s
```

If we feel that 2869 doesn't change the behavior of this PR, I'd suggest merging that, as it seems to give us the non-OOMing perf with no possibility of OOMing Cassandra as far as we know.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
```diff
@@ -82,22 +95,73 @@
...
+        }
+        AtomicInteger nextRowToQuery = new AtomicInteger(0);
+        for (int i = 0; i < executorThreads; i++) {
+            scheduleSweepRowTask(futures, queryId, nextRowToQuery.getAndIncrement(), nextRowToQuery, rowsAscending,
+                    executor);
+        }
+
+        try {
+            for (int rowIndex = 0; result.size() < limit && rowIndex < rowsAscending.size(); rowIndex++) {
+                Future<CqlResult> future = futures.get(rowIndex);
```
tboam: I'd be happier if we checked for null here but don't think it should block the release.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
```diff
@@ -82,22 +95,73 @@
...
-            // 2. There are at most <limit> results
-            // 3. We return the first <limit> results
-            // We can ensure this by not applying the limit until after collection, and grouping by row
-            // I think this is ensured later on in the code path?
-            CqlResult cqlResult = queryExecutor.executePrepared(queryId,
-                    ImmutableList.of(ByteBuffer.wrap(rows.next())));
-            result.addAll(CqlExecutorImpl.getCells(CqlExecutorImpl::getCellFromRow, cqlResult));
+
+        List<Future<CqlResult>> futures = new ArrayList<>(rowsAscending.size());
+        for (int i = 0; i < rowsAscending.size(); i++) {
```
tboam: Do we need to initialize everything to null when we already create it to be the right size?

fsamuel-bs: Yep. Setting the `rowsAscending.size()` sets the capacity, but not the size. We rely on being able to `set(integer, future)` down, so we need to fill it up.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlExecutorImpl.java
fsamuel-bs: Our logic should prevent NPE. But this whole code feels horrible, and I do want to refactor it very soon.

PR #2862, opened 2018-01-15
[TTT] validate index name length with suffix

hsaraogi:
**Goals (and why)**: Index name should be reported as too long at schema gen time if it exceeds Cassandra or Postgres table name limit.

**Implementation Description (bullets)**: validateTableNameLength

**Where should we start reviewing?**: SchemaTest.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2862)

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,6 +50,11 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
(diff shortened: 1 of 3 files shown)


docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,11 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
```
jeremyk-91: nit: maybe `|fixed|`?

Issue #2803, opened 2017-12-11
Remove partitionStrategy if not used


gregakinman:
https://palantir.github.io/atlasdb/html/schemas/tables_and_indices.html#partitioners

@gmaretic @j-baker @gsheasby

gsheasby:
So partitionStrategy does indeed seem to be unused.

It predates the "partitioning" PR, #124, and is unrelated. Indeed, it predates the move to OSS AtlasDB, although it is used in a couple of our internal products (our internal auth product, and our internal citrus product)

hsaraogi:
Tagging @tboam for FLUP.

tboam:
@gregakinman for my understanding: looks like this is a clean up task to remove this functionality to simplify schema creation by users.  Is that correct?  The existence of this functionality adds some time to writing schemas but isn't blocking development work, correct?

gregakinman:
According to @j-baker specifying `partitionStrategy` is a no-op. It should be removed because it adds the illusion that it does something when it in fact does not. It adds time to writing schemas in the sense that it's additional characters being typed for no effect I guess?

tpetracca:
Is it actually always a no-op? If so, we should almost certainly be removing it. The time added is likely someone seeing it in the API and actually tryign to understand what it is and if/why they should use it.

gregakinman:
Yep, believe Jeremy has a PR up to remove it.

jeremyk-91:
We've released 0.73.0, so I'm going to merge the PR today. Sorry about the delay!

PR #2864, opened 2018-01-15
[MMM] Remove Unused partitionStrategy Parameter

jeremyk-91:
**Goals (and why)**:
- Address #2803 - @gregakinman for SA
- Remove the unused parameter - having something unused that actually does stuff

- Remove the partitionStrategy parameter from the protobuf, and propagate this through to Java code
- Remove mentions of partitionStrategy from the docs
- Release notes

- This *WILL* dev-break large internal product, internal auth product and internal citrus product, though resolution simply involves deleting one line of code. This parameter doesn't look like it's ever read (not even separately in large internal product).
- Am I wrong in saying that this parameter is never read?
- I had a janky way of stopping people from using tag number 5 by leaving a comment. Unfortunately `reserved` in protobuf is only available with protoc 3.0.0 which we don't use yet. I did think about upgrading our usage to protoc 3.0.0 but that looks like it might open a can of worms with (1) circle (admittedly this is probably not hard to resolve), and (2) large internal product.

`TableMetadataPersistence.proto`

**Priority (whenever / two weeks / yesterday)**: this week. Obviously not super high priority.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2864)


jeremyk-91:
This is breaking, so holding until after 0.73.0 release.

PR #2865, opened 2018-01-15
Upgrades gradle wrapper to the latest version

gsheasby:
Started as a copy of #2824... became a massive hunt for broken and outdated dependencies.

**Goals:** eliminate dependency conflicts across the repo

**Implementation description**
- Upgrade nebula and sls-packaging to their latest versions
- Apply the correct force strategy to clear out the duplicated dependencies
- Add timelock.props (and later qos.props) to make explicit the special dependencies used by Timelock and QoS
- Fine-tune the dependency bumps to correct test failures

The root cause of the test failures was that Timelock and QoS each have a special set of dependencies. QoS was simply using Timelock's old, kind of janky dependency specification logic. QoS needs some of the same custom dependencies as Timelock, but not all. This is now made explicit via `qos.props` and `timelock.props`

**Future work**
- There's still some special dependency logic in `versions.gradle` - we should unify this with the `props` files so that dependencies are specified in one place.
- @iamdanfox suggested that we can/should get rid of the troublesome *.lock files entirely, but this is beyond the scope of this PR
- ~Upgrade baseline/errorprone (we're on v13, v18 is available)~
- The baseline/errorprone upgrade flagged a couple of errors in `PTExecutors.java`; I'll open a separate PR for this after this merges, possibly as part of MMM next week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2865)


gsheasby:
@jeremyk-91 I'll own investigating what's going on here.

gsheasby:
There's a dependency conflict somewhere in the buildscript dependencies, relating to guava:
```
Caused by: java.lang.NoSuchMethodError: com.google.common.collect.ArrayListMultimap.create()Lcom/google/common/collect/ArrayListMultimap;
```

However, I can't run depencencyInsight because I hit the same roadblock!

gsheasby:
Reassigning to @jeremyk-91 for review. However, this PR has high potential for merge conflicts, and there's also the danger that someone will, in the meantime, introduce something that will make the new dependency-recommender version freak out.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
```diff
@@ -115,11 +115,13 @@ private CassandraVerifier() {
...
     }
 
+    @SuppressWarnings("ValidateConstantMessage") // https://github.com/palantir/gradle-baseline/pull/175
     static void sanityCheckTableName(TableReference tableRef) {
         String tableName = tableRef.getQualifiedName();
         Validate.isTrue(!(tableName.startsWith("_") && tableName.contains("."))
                 || AtlasDbConstants.hiddenTables.contains(tableRef)
-                || tableName.startsWith(AtlasDbConstants.NAMESPACE_PREFIX), "invalid tableName: " + tableName);
+                || tableName.startsWith(AtlasDbConstants.NAMESPACE_PREFIX),
+                "invalid tableName: %s", tableName);
```
jeremyk-91: nice, no idea we used commons-lang for these

Issue #2111, opened 2017-07-07
AtlasDB-Commons, Java 6 and Canids


jeremyk-91:
Internal canid product can't upgrade AtlasDB, because for other unrelated reasons they are forced to use Java 6. The main issue is `atlasdb-commons`; we should also check which other AtlasDB projects said product uses.

Ideally we would also maintain devs' ability to work without necessarily requiring a Java 6 compiler. Like what we do with ErrorProne, we could possibly implement this by having a separate compilation configuration for those projects with Java 6, and ensure that this is tested in CI. We would also need to publish artifacts with Java 6.

@gmaretic this is likely to mean that JAR publish might require a Java 6 compiler. Would that be okay? If not we can probably put in an additional publish config, though we'll need to be very careful that only the 'real' one is actually called when we really do publish e.g. to bintray.

@schlosna @twilson-palantir for SA

tpetracca:
I was under the impression as of a couple months ago we no longer had to support java6?

tjwilson90:
My understanding of why we still need java 6 is that we support patching branches on lts releases that are still on java 6.

fsamuel-bs:
We do support Java 6 on our LTS branches, 0.4.x and 0.39.x, and we backport fixes to them whenever needed. If backporting becomes painful, we can think of a better solution, but right now we'll continue with our current setup (0.41 onwards Java 8 only support).

hsaraogi:
Internal candid product needs Java 6 as they have client side applications in environments where only Java 6 in available. These depend only on AtlasDB commons. We can reintroduce the Java6 support by specifying multiple targets in the gradle config, just as we did before. This will help them take the latest sweep improvements and move to higher Atlas versions going further. @carrino for SA.

carrino:
this is all we need: https://github.com/palantir/atlasdb/pull/2122
it's not all of commons, but just executors

carrino:
we may have a workaround where the 1.6 projects depend on an old version and the 1.8 projects a newer version and gradle will just bump up the deps if you depend on 1.6 and 1.8 stuff

fsamuel-bs:
Fixed by #2122.

PR #2868, opened 2018-01-15
Clean up executor wrapping for tracing

sandorw:
**Goals (and why)**:
Enable http-remoting tracing by default on our executor services

- Modify PTExecutors to auto-wrap
- Find existing executor services not using PTExecutors and switch them (I did not update instances in test code)

PTExecutors.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2868)


sandorw:
@carrino - I removed the 1.6 compatibility from the project with PTExecutors. Looking at https://github.com/palantir/atlasdb/issues/2111, is this still an issue? Would like to figure out a work around if this change will break our older projects.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
```diff
@@ -501,7 +487,13 @@ public void run() {
...
+        });
+        if (runnable instanceof Future<?>) {
+            @SuppressWarnings("unchecked")
+            Future<Object> unsafeFuture = (Future<Object>) runnable;
+            return new ForwardingRunnableFuture<Object>(wrappedRunnable, unsafeFuture);
+        }
+        return wrappedRunnable;
     }
 
     public static <T> RunnableFuture<T> wrap(RunnableFuture<T> rf) {
```
fsamuel-bs: I don't think we use this method, but let's also wrap it just to be safe.

sandorw:
@carrino confirmed that it should be fine to drop the Java 6 compatibility here.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
sandorw: Thanks for catching that - fixed it and refactored the code a bit more.

PR #2869, opened 2018-01-15
Schedule future just after the previous is executed

fsamuel-bs:
---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2869)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetCellTimestamps.java
```diff
@@ -94,17 +100,21 @@ private void fetchBatchOfTimestamps() {
     private void fetchBatchOfTimestampsBeginningAtStartRow() {
         byte[] rangeStart = startRowInclusive;
 
+        ExecutorService executor = PTExecutors.newFixedThreadPool(config.sweepReadThreads());
```
gsheasby: Huh, this was actually the opposite of what I thought would happen - I thought the executor service would be created and destroyed within `CqlExecutor.getTimestamps`.

I think that this will almost always be equivalent, though: `timestamps.isEmpty` will be false after one iteration except in cases where we're sweeping a thorough table and have `batchHint` rows of tombstones in front of us.

fsamuel-bs: It seemed weird that we'd be potentially recreating the executor in each `getTimestamps` call at first glance, and we had a `while` loop above. But yeah, we just do that on the multiple tombstones scenario, which is pretty rare though. I think we're fine leaving like this is well — we already start to extract some of the logic from the CqlExecutor.

gsheasby:
An integration test is failing :(
```
java.lang.IndexOutOfBoundsException: Index: 11, Size: 10
	at java.util.ArrayList.rangeCheckForAdd(ArrayList.java:661)
	at java.util.ArrayList.add(ArrayList.java:473)
	at com.palantir.atlasdb.keyvalue.cassandra.CqlExecutorImpl.scheduleSweepRowTask(CqlExecutorImpl.java:145)
	at com.palantir.atlasdb.keyvalue.cassandra.CqlExecutorImpl.getTimestamps(CqlExecutorImpl.java:99)
```

You also have a set of checkstyle errors.

tboam:
`List.add(index, value)` fails if the index is greater than the List's size.  We should either create the `List` as long enough to hold all possible values.  Or just use a `Map<Integer, Future>` to achieve what we set out to - we could also remove the values as we go.

fsamuel-bs:
There were some issues with the previous version. I've fixed them (tests are now green locally) and added some docs. I'm quite unhappy with how this code turned out, as its complicated and error-prone (vide previous failures on circle). I think it is correct now, but plan to refactor it very soon.

PR #2871, opened 2018-01-16
 Remove unnecessary remote call on txn commit

sandorw:
**Goals (and why)**:
Remove unnecessary lock refresh _after_ a transaction has already been committed. 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2871)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1360,14 +1360,6 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             putCommitTimestamp(commitTimestamp, commitLocksToken, transactionService);
             long millisForCommitTs = TimeUnit.NANOSECONDS.toMillis(commitTsTimer.stop());
 
-            Set<LockToken> expiredLocks = refreshCommitAndImmutableTsLocks(commitLocksToken);
-            if (!expiredLocks.isEmpty()) {
-                final String baseMsg = "This isn't a bug but it should happen very infrequently. "
-                        + "Required locks are no longer valid but we have already committed successfully. ";
-                String expiredLocksErrorString = getExpiredLocksErrorString(commitLocksToken, expiredLocks);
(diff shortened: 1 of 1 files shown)


fsamuel-bs:
@dxiao for SA. Maybe this was used in the past, but doesn't seem useful now.

sandorw:
Also adding @mswintermeyer in case he has thoughts

dxiao:
We probably should all be looking at these logs.

My understanding is that the transaction commit protocol is thus:
1) transaction starts, does some reading and writing
2) transaction goes to commit
2a) grabs lock leases
2b) checks for conflicts
2c) check lock lease validity
2d) commit
2e) double-check lock lease validity, if invalid, throw error
3) return

Steps 2c and 2d cannot be done atomically, since they involve writes to separate services. The problem though, is if there's a 1minute long gc on the client between 2c and 2d, where the locks timeout, and another transaction from another client grabs the locks and commits its own transaction, which should have caused a conflict but didn't because it slid in under the radar. 

The firing of this log line presents the possibility that an execution of the transaction protocol did not actually ensure correctness of its operations. Since a long pause between 2c and 2d is not the only way this can happen (between 2d and 2e for example), there's a possibility of false positive, but seeing large numbers of this message in a run would make me very worried about possible database consistency violations.

sandorw:
I don't think the case you outlined has any consistency issues due to the rollback behavior in the read transaction protocol. I verified that the code in `SnapshotTransaction` agrees with the behavior described in the transaction protocol docs.

If the second transaction reads any values before writing to them, the reads would roll back the uncommitted transaction. If it writes them without reading, the conflict detection code still loads the latest version of each cell and fetches the commit timestamp, which would force the first transaction to either commit or be rolled back. If it's rolled back, no conflict. If it commits, then the conflict detection code can determine whether that's a problem or not based on conflict handler settings.

Fetching latest version of each cell: https://github.com/palantir/atlasdb/blob/develop/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java#L1597

Roll back uncommitted transactions: https://github.com/palantir/atlasdb/blob/develop/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java#L1636

dxiao:
Ah I see. Works for me then.

PR #2872, opened 2018-01-16
Hsaraogi/query the qos service less frequently

hsaraogi:
**Goals (and why)**:
1. Query the qos service for the rate every 2 seconds, instead of every read/write request to Cassandra.

**Implementation Description (bullets)**: rate limit the limit updating, handle exceptions

**Concerns (what feedback would you like?)**: Is 2 seconds okay? Is preserving the limit if we cannot contact QoS a good idea?
Should we add more metrics here to get a client side view of the limits? Concern is that we have two metrics tracking the same thing then.

**Where should we start reviewing?**: QosRateLimiters.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2872)

### DIFF
--- a/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/QosRateLimiter.java
+++ b/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/QosRateLimiter.java
@@ -41,6 +41,7 @@ import com.palantir.logsafe.SafeArg;
 public class QosRateLimiter {
 
     private static final Logger log = LoggerFactory.getLogger(QosRateLimiter.class);
+    private static final double ONCE_EVERY_TWO_SECONDS = 0.5;
 
     private static final long MAX_BURST_SECONDS = 5;
 
@@ -50,6 +51,7 @@ public class QosRateLimiter {
     private final RateLimiter.SleepingStopwatch stopwatch;
 
     private volatile RateLimiter rateLimiter;
+    private final com.google.common.util.concurrent.RateLimiter rateUpdateLimiter;
     private volatile long currentRate;
 
     public static QosRateLimiter create(Supplier<Long> maxBackoffTimeMillis, Supplier<Long> unitsPerSecond,
@@ -65,6 +67,7 @@ public class QosRateLimiter {
         this.unitsPerSecond = unitsPerSecond;
         this.maxBackoffTimeMillis = maxBackoffTimeMillis;
         this.rateLimiterName = rateLimiterName;
+        this.rateUpdateLimiter = com.google.common.util.concurrent.RateLimiter.create(ONCE_EVERY_TWO_SECONDS);
 
         createRateLimiterAtomically();
     }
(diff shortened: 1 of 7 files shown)


sandorw:
I don't think we need client side metrics if there's only a 2-second lag. 2 seconds seems pretty reasonable to help ensure that any throttling happens as quickly as possible once cassandra metrics show signal.

I think preserving last seen limits is fine for now. As we test and roll this out, if rate limiting becomes very commonplace as a measure to improve stability and we continue to invest, we can revisit and consider having clients start rate limiting themselves if QoS is unavailable.

qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/QosRateLimiter.java
```diff
@@ -95,8 +98,21 @@ public Duration consumeWithBackoff(long estimatedNumUnits) {
      * overhead and double comparisons, we maintain the current rate ourselves.
      */
     private void updateRateIfNeeded() {
-        if (currentRate != unitsPerSecond.get()) {
-            createRateLimiterAtomically();
+        if (rateUpdateLimiter.tryAcquire()) {
+            long updatedRate = getUpdatedRate();
```
sandorw: It might still be nice to not add any network overhead to a request, even if it's time to fetch a new limit. Could we just have a background scheduled executor that fetches new limits, then updates the rate limiter when there's a change?

sandorw:
I agree, especially since the cost should only be the network hop cost, since we're caching the last seen cassandra metrics on the QoS size.

qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/QosRateLimiter.java
```diff
@@ -112,7 +130,15 @@ private long getUpdatedRate() {
         } catch (Exception e) {
             log.error("Could not refresh the Qos rate. This can happen if the Qos Service is unreachable."
                     + " Extended periods of being unable to refresh will hinder QoS of all clients., ", e);
+            return returnPreviousOrDefaultRate();
```
sandorw: Use the static message template that you added here

qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/QosRateLimiters.java
```diff
@@ -20,16 +20,25 @@
...
 @Value.Immutable
 public interface QosRateLimiters {
 
     static QosRateLimiters create(Supplier<Long> maxBackoffSleepTimeMillis, Supplier<Long> readLimitSupplier,
             Supplier<Long> writeLimitSupplier) {
         QosRateLimiter readLimiter = QosRateLimiter.create(maxBackoffSleepTimeMillis,
-                readLimitSupplier, "read");
+                readLimitSupplier,
+                "read",
+                PTExecutors.newSingleThreadScheduledExecutor(),
```
sandorw: PTExecutors 🥇

PR #2804, opened 2017-12-11
Make refresh time configurable

abless:
The LockRefreshingLockService currently hard-codes the refresh period. This change makes it configurable while maintaining back-compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2804)

PR #2875, opened 2018-01-16
Make refresh time configurable (branch-version of #2804)

jeremyk-91:
**Goals (and why)**:
- The LockRefreshingLockService currently hard-codes the refresh period. This change makes it configurable while maintaining back-compatibility.

- Make the param configurable (programmatically - not from config)

**Where should we start reviewing?**: 9/3

@abless: I think we need to figure out the use case here too - see https://github.com/palantir/atlasdb/pull/2804#pullrequestreview-82525569

Fairly sure the change looks fine.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2875)


jeremyk-91:
@abless a ping on this - will this still be useful for your product?

abless:
Hey, this is no longer needed for our product. The original driver was that we wanted to be able to have shorter lock timeouts but the current implementation forced us to >=5 seconds.

jeremyk-91:
Ack - closing this out then. Thanks

Issue #2881, opened 2018-01-17
Sweep can try to delete cells with null startTs 


hsaraogi:
```
com.palantir.common.exception.AtlasDbDependencyException: {throwable0_message}
	at com.palantir.common.base.Throwables.createAtlasDbDependencyException(Throwables.java:100)
	at com.palantir.common.base.Throwables.unwrapAndThrowAtlasDbDependencyException(Throwables.java:92)
	at com.palantir.atlasdb.qos.ratelimit.QosAwareThrowables.unwrapAndThrowRateLimitExceededOrAtlasDbDependencyException(QosAwareThrowables.java:42)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl.deleteOnSingleHost(CassandraKeyValueServiceImpl.java:1403)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl.delete(CassandraKeyValueServiceImpl.java:1338)
	at com.palantir.atlasdb.keyvalue.impl.ProfilingKeyValueService.lambda$delete$33(ProfilingKeyValueService.java:156)
	at com.palantir.atlasdb.logging.KvsProfilingLogger.lambda$maybeLog$18(KvsProfilingLogger.java:96)
	at com.palantir.atlasdb.logging.KvsProfilingLogger.maybeLog(KvsProfilingLogger.java:119)
	at com.palantir.atlasdb.logging.KvsProfilingLogger.maybeLog(KvsProfilingLogger.java:102)
	at com.palantir.atlasdb.logging.KvsProfilingLogger.maybeLog(KvsProfilingLogger.java:95)
	at com.palantir.atlasdb.keyvalue.impl.ProfilingKeyValueService.maybeLog(ProfilingKeyValueService.java:398)
	at com.palantir.atlasdb.keyvalue.impl.ProfilingKeyValueService.delete(ProfilingKeyValueService.java:156)
	at com.palantir.atlasdb.keyvalue.impl.ForwardingKeyValueService.delete(ForwardingKeyValueService.java:70)
	at com.palantir.atlasdb.keyvalue.impl.TracingKeyValueService.delete(TracingKeyValueService.java:147)
	at sun.reflect.GeneratedMethodAccessor544.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #2882, opened 2018-01-17
Log if sweep cell startTs is null

hsaraogi:
**Goals (and why)**: To investigate #2881.

**Concerns (what feedback would you like?)**: Should we skip the delete part and just return as we know delete will throw an NPE and make sweep stuck. Danger is we never see this logging and assume sweep is running, but we can have monitors for the log line - its already error?

**Where should we start reviewing?**: CellsSweeper

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2882)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
@@ -16,6 +16,7 @@
 package com.palantir.atlasdb.sweep;
 
 import java.util.Collection;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -90,6 +94,13 @@ public void sweepCells(
                     sentinelsToAdd);
         }
 
+        if (cellTsPairsToSweep.entries().stream().anyMatch(entry -> entry.getValue() == null)) {
```
tboam: `cellTsPairsToSweep.containsValue(null)`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -90,6 +94,13 @@ public void sweepCells(
...
+            log.error("When sweeping table {} found cells to sweep with the start timestamp null."
+                            + " This is unexpected. The cellTs pairs to sweep were: {}.",
+                    SafeArg.of("tableRef", LoggingArgs.tableRef(tableRef)),
```
tboam: Don't rewrap the `Arg` from `tableRef` as this will potential convert an `UnsafeArg` into a `SafeArg`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -98,4 +109,10 @@ public void sweepCells(
             persistentLockManager.releasePersistentLock();
         }
     }
+
+    private String getLoggingMessageFromMap(Multimap<Cell, Long> cellTsPairsToSweep) {
```
tboam: I would either sort the entries by their key or use the `Multimap`'s own `toString` method.  Fr large maps this output may be hard to read if there are many values associated with the key

PR #2884, opened 2018-01-18
BackgroundSweeperImpl shutdown and fatal metrics

tboam:
Added 2 new sweep outcomes to the BackgroundSweeperImpl for permanent end states.

**Goals (and why)**: We want to be able to detect and immediately alarm if the Background Sweeper fails.

**Implementation Description (bullets)**:  Add 2 new SweepOutcome values to the enum.  These differ to previous values in that they are permanent end states so we record and publish the metric differently: rather than use a reservoir which evicts values we permanently switch modes for these metric types.

This is done in a bit of a brute force manner for now but there are only 2 of these permanent outcomes so don't think it warrants special handling (if we add more then I'd change the enum value to know if it is permanent or not and store non-permanent values in the reservoir and permanent values in a separate Map).

Set the shutdown flag either if the shutdown method is called on the sweeper or the thread is interrupted directly as either could technically happen.

**Concerns (what feedback would you like?)**: Have I introduced a bug?  There's no testing at the top level of the sweeper.

**Where should we start reviewing?**: Single class.

**Priority (whenever / two weeks / yesterday)**: 1 week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2884)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
@@ -36,6 +36,7 @@ import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.common.base.Throwables;
 import com.palantir.lock.LockService;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 
 public final class BackgroundSweeperImpl implements BackgroundSweeper {
     private static final Logger log = LoggerFactory.getLogger(BackgroundSweeperImpl.class);
@@ -133,7 +134,8 @@ public final class BackgroundSweeperImpl implements BackgroundSweeper {
             log.warn("Shutting down background sweeper. Please restart the service to rerun background sweep.");
             sweepOutcomeMetrics.shutdown();
         } catch (Throwable t) {
-            log.error("BackgroundSweeper failed fatally and will not rerun until restarted", t);
+            log.error("BackgroundSweeper failed fatally and will not rerun until restarted: {}",
+                    UnsafeArg.of("message", t.getMessage()), t);
             sweepOutcomeMetrics.fatal();
         }
     }
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
@@ -132,11 +132,11 @@ public final class BackgroundSweeperImpl implements BackgroundSweeper {
             }
         } catch (InterruptedException e) {
             log.warn("Shutting down background sweeper. Please restart the service to rerun background sweep.");
-            sweepOutcomeMetrics.shutdown();
+            sweepOutcomeMetrics.registerOccurrenceOf(SweepOutcome.SHUTDOWN);
         } catch (Throwable t) {
             log.error("BackgroundSweeper failed fatally and will not rerun until restarted: {}",
                     UnsafeArg.of("message", t.getMessage()), t);
-            sweepOutcomeMetrics.fatal();
+            sweepOutcomeMetrics.registerOccurrenceOf(SweepOutcome.FATAL);
         }
     }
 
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -271,6 +277,7 @@ SweepLocks createSweepLocks() {
 
     @Override
     public synchronized void shutdown() {
+        sweepOutcomeMetrics.shutdown();
```
hsaraogi: We should call `sweepOutcomeMetrics .registerOccurrenceOf(SHUTDOWN);`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -131,6 +132,11 @@ public void run() {
             }
         } catch (InterruptedException e) {
             log.warn("Shutting down background sweeper. Please restart the service to rerun background sweep.");
+            sweepOutcomeMetrics.shutdown();
```
hsaraogi: We should call `sweepOutcomeMetrics .registerOccurrenceOf(SHUTDOWN);`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -131,6 +132,11 @@ public void run() {
...
+        } catch (Throwable t) {
+            log.error("BackgroundSweeper failed fatally and will not rerun until restarted: {}",
+                    UnsafeArg.of("message", t.getMessage()), t);
+            sweepOutcomeMetrics.fatal();
```
hsaraogi: We should call `sweepOutcomeMetrics .registerOccurrenceOf(FATAL);`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -286,29 +293,59 @@ public synchronized void shutdown() {
...
+            }
+            if (outcome == SweepOutcome.FATAL) {
+                fatal();
+                return;
+            }
+
             reservoir.update(outcome.ordinal());
         }
+
+        public void shutdown() {
```
hsaraogi: We should inline `shutdown()` and `fatal()` as these new APIs to SweepOutcomeMetrics could be confusing. I see this as something I register my outcome with and then it reports the last outcome when I call `getOutcome`.

tboam: Yeah, I wrote this the other way round - `shutdown()` and `fatal()` were separate things so it made sense, but then I reused the SweepOutcome to make the metric creation easy and so had to guard against someone passing them to `registerOccurenceOf()`.

I'll inline everything as it should reduce API size.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -286,28 +293,50 @@ public synchronized void shutdown() {
...
+        NOT_ENOUGH_DB_NODES_ONLINE, TABLE_DROPPED_WHILE_SWEEPING, ERROR,
+        SHUTDOWN, FATAL
     }
 
     private class SweepOutcomeMetrics {
         private final MetricsManager metricsManager = new MetricsManager();
         private final SlidingTimeWindowReservoir reservoir;
 
+        private boolean shutdown;
+        private boolean fatal;
```
gmaretic: Should this be volatile?

tboam: As discussed I think we're ok with an update taking some amount of time before it is read by the metrics thread.  We only write in one driection and it is only a boolean so we won't get weird race conditions

PR #2887, opened 2018-01-19
move to proto3

berler:
**Goals (and why)**: Upgrade to the latest protobuf library, so that we can similarly upgrade internal projects that depend on atlas.  The old version we use now is vulnerable to https://github.com/google/protobuf/issues/760

* deletes defunct proto project/dir
* use the newest protobuf gradle plugin
* use protobuf 3.5.1 (currently the latest)
* correctly lock in protobuf-java version
* .proto files are now located in a place that makes more sense
* generated proto java sources are no longer checked into git

Note that all .proto files still use the proto2 spec. It is not trivial to change existing protobufs to use the proto3 spec in a way that is backwards compatible due to using features not present in proto3 (custom default values, enums don't have a zero value)

**Priority (whenever / two weeks / yesterday)**: Normal

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2887)


berler:
After looking at this more, I think actually changing GeneratedMessage -> AbstractMessage instead will be better, since both GeneratedMessage and GeneratedMessageV3 extend AbstractMessage.  Hopefully this will mean that schemas compiled with the proto2 compiler would still work (so people shouldn't even need to recompile).

PR #2889, opened 2018-01-19
Fix TokenRangeWriteLogger registering the same metric over and over when table names aren't marked as safe

tboam:
**Goals (and why)**: Register different tag values when table names are all unsafe and so can't be used directly.

**Implementation Description (bullets)**: Compute a hash of the table name when it is unsafe and then take the first few characters, this prevents us from recomputing the table name.  But if we know the table names from a schema then we can figure out which table is bad.

**Concerns (what feedback would you like?)**: Is this worth it?  We could just ignore the warnings in the logs and make products mark tables as safe.

**Where should we start reviewing?**: MetricsManager.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2889)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -95,8 +105,18 @@ public void registerMetric(Class clazz, String metricName, Gauge gauge, Map<Stri
...
+        String tableName = tableRef.getTablename();
+        if (!isSafeToLog.apply(tableRef)) {
+            tableName = "unsafeTableRef_" + obfuscate(tableRef);
+        }
+
+        return ImmutableMap.of("tableName", tableName);
+    }
+
+    private String obfuscate(TableReference tableRef) {
+        return Sha256Hash.computeHash(tableRef.getTablename().getBytes()).serializeToHexString().substring(0, 8);
```
hsaraogi: Was thinking about collisions here, https://stackoverflow.com/questions/19962424/probability-of-collision-with-truncated-sha-256-hash, theres about ~1% probability of collision in 10k tables, but we are fine I think (except for the large internal product that is creating and truncating tables). We can take a 16 character substring?

atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -95,8 +105,18 @@ public void registerMetric(Class clazz, String metricName, Gauge gauge, Map<Stri
         taggedMetricRegistry.gauge(metricToAdd, gauge);
     }
 
-    private Map<String, String> getTableNameTagFor(TableReference tableRef) {
-        return ImmutableMap.of("tableName", LoggingArgs.safeTableOrPlaceholder(tableRef).getQualifiedName());
+    @VisibleForTesting
+    Map<String, String> getTableNameTagFor(TableReference tableRef) {
+        String tableName = tableRef.getTablename();
+        if (!isSafeToLog.apply(tableRef)) {
+            tableName = "unsafeTableRef_" + obfuscate(tableRef);
```
hsaraogi: We can save some characters here with t_hash.

hsaraogi:
However, if we get products on this version to mark the table names as safe, we would leave hanging metric names in the internal metrics framework. Do we want to hold off until we send out comms around making table name safety?

tboam:
I'll increase the hash size slightly.  Once metrics stop appearing they are cleaned up pretty aggressively so not too concerned about this.  Also the same name is chosen for a table everywhere it is deployed so explosion is not too big (at most 2x the actual number of tables but in reality much less).

PR #2892, opened 2018-01-19
Cleanup Requirements, Part 4: SchemaMetadataService ETE Tests

jeremyk-91:
**Goals (and why)**:
- Add an ETE test to seek confidence that the schema metadata service works properly for actual key-value services.
- Fix the rather restrictive Visitor APIs

- Add `BlobSchema` which has two stream stores and a table with arbitrary cleanup tasks.
- Verify that things can be persisted correctly to both Postgres and Cassandra.
- Correct the visitor APIs to work for things other than voids.

Nothing in particular. This one is just a test!

- `BlobSchema` and `SchemaMetadataServiceEteTest`

**Priority (whenever / two weeks / yesterday)**: next week (likely this week when you read this!)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2892)


atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/blob/BlobSchema.java
```diff
@@ -0,0 +1,91 @@
...
+    private static Schema generateSchema() {
+        Schema schema = new Schema(
+                BlobSchema.class.getSimpleName(),
+                BlobSchema.class.getPackage().getName() + ".generated",
+                BLOB_NAMESPACE);
+
+        schema.addStreamStoreDefinition(
+                new StreamStoreDefinitionBuilder("data", "Data", ValueType.VAR_LONG)
+                .hashRowComponents()
+                        .tableNameLogSafety(TableMetadataPersistence.LogSafety.SAFE)
```
gsheasby: nit: whitespace looks out-of-whack here

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/schema/CleanupMetadataResourceImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public Optional<SerializableCleanupMetadata> get(
+            String schemaName,
+            String tableName) {
+        return schemaMetadataServiceSupplier.get().loadSchemaMetadata(schemaName)
+                .map(SchemaMetadata::schemaDependentTableMetadata)
+                .flatMap(metadataMap -> Optional.ofNullable(metadataMap.get(
+                        TableReference.createFromFullyQualifiedName(tableName))))
+                .map(SchemaDependentTableMetadata::cleanupMetadata)
```
gsheasby: what happens if the table is not present (and so the `flatMap` in the previous line resolves to `null`)?

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/schema/SerializableCleanupMetadata.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.atlasdb.schema.cleanup.ArbitraryCleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.CleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.NullCleanupMetadata;
+import com.palantir.atlasdb.schema.cleanup.StreamStoreCleanupMetadata;
+
+/**
+ * This class is somewhat hacky and should not be used outside of the atlasdb-ete-tests project.
```
gsheasby: Can it be given `package` visibility to ensure this, or would that not be sensible due to where this is used? It's somewhat stronger than project-only...

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/blob/BlobSchema.java
jeremyk-91: Ack, will fix this. Not sure what happened here!

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/schema/CleanupMetadataResourceImpl.java
jeremyk-91: `flatMap` shouldn't ever return `null` actually - if the table doesn't exist then `metadataMap.get(...)` does return `null`, but then `Optional.ofNullable(null)` is `Optional.empty()`, which is what we want for this line as I think in this case the whole method returning `Optional.empty()` is right.

jeremyk-91: Agree that unit tests would be good - will write some tomorrow morning.

jeremyk-91:
I've fixed the comments - sending back over to @gsheasby for a final look especially regarding the optional-chain logic.

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/schema/SerializableCleanupMetadata.java
gsheasby: I was just wondering if this would be easy - turns out not to be. I can live with that 👍

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/schema/CleanupMetadataResourceImpl.java
gsheasby: Thanks for adding the tests. Your `flatMap` explanation makes sense, too.

PR #2894, opened 2018-01-20
Fix/transaction manager tests

berler:
**Goals (and why)**: Fix a test that was broken because it was not being run

* Move the test into the correct test sources dir
* Revert the commit that broke the test

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2894)


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConstraintCheckable.java
```diff
@@ -23,7 +23,7 @@
 public interface ConstraintCheckable {
     /**
      * This method checks row, table, and foreign-key constraints on the values given in writes.
-     * This method may require additional reads from the database, depending on the value of
+     * This method may require additional reads from the database, dependong on the value of
```
clockfort: uhh

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -68,7 +68,7 @@ default boolean isInitialized() {
     /**
      * {@link #runTaskWithRetry(TransactionTask)} should be preferred over
      * {@link #runTaskThrowOnConflict(TransactionTask)}.
-     * This method should be used unless {@link #runTaskWithRetry(TransactionTask)} cannot be used because the arguments
+     * This method should be used unless {@link #runTaskWithRetry(TransactionTask)} cannot be used becuase the arguments
```
clockfort: ?

PR #2895, opened 2018-01-22
Generalize pre-commit conditions v2

sandorw:
**Goals (and why)**:
Re-merge https://github.com/palantir/atlasdb/pull/2846 and fix the test failures from https://github.com/palantir/atlasdb/pull/2894

`SnapshotTransaction` now tracks whether any reads have occurred and skips the pre-commit checks if there are no reads or writes.

**Where should we start reviewing?**: `SnapshotTransaction.java` on the second commit only (first commit is a cherry-pick from the revert)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2895)


sandorw:
@berler for SA - don't know of any other broken tests caused by the PR, otherwise I'm happy to work on them too

PR #2896, opened 2018-01-22
Live reload CKVS config

fsamuel-bs:
**Goals (and why)**: Introduce a notion of `KVSRuntimeConfig`. The only implemented is `CassandraKVSRuntimeConfig`, which makes some values of `CKVSConfig` live-reloadable, as well as makes some parameters of Cassandra retry logic.

**Concerns (what feedback would you like?)**: Do we want to do the same Jackson de-serialization we did with KeyValueServiceConfig and ask products to do register a custom ObjectMapper?

**Where should we start reviewing?**: `KVSRuntimeConfig`, then `CassandraKVSRuntimeConfig`, then `CassandraKVSReloadableConfig`.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2896)


fsamuel-bs:
Lacks tests and docs (mainly around CassandraKVSReloadableConfig), but wanted to get the discussion going around the current direction of the PR.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public CassandraKeyValueServiceConfig delegate() {
+        return config;
+    }
+
+    @Override
+    public int unresponsiveHostBackoffTimeSeconds() {
+        return MoreObjects.firstNonNull(
```
tboam: Can probably simplify all these methods by passing a single `function` and then something else can decide which config object to invoke it on.

atlasdb-jdbc/src/main/java/com/palantir/atlasdb/keyvalue/jdbc/JdbcAtlasDbFactory.java
```diff
@@ -46,7 +48,8 @@ public String getType() {
      * Creates a JdbcKeyValueService.
      *
      * @param config Configuration file.
-     * @param leaderConfig unused.
+     * @param runtimeConfig
+     *@param leaderConfig unused.
```
tboam: nit: spacing

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
+         - Make some values of ``CassandraKeyValueServiceConfig`` live-reloadable.
+           To check which parameters are live-reloadable, check the `CassandraKeyValueServiceRuntimeConfig` class.
+           Docs about this config can be found `here<insert docs url>`.
```
tboam: Still need docs

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -153,7 +154,7 @@
  * and these inactive nodes will be removed afterwards.
  */
 @AutoDelegate(typeToExtend = CassandraKeyValueService.class)
-public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implements CassandraKeyValueService {
+public final class CassandraKeyValueServiceImpl extends AbstractKeyValueService implements CassandraKeyValueService {
```
tboam: Is this  ok with the large internal product?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -168,7 +179,9 @@ private CassandraClientPoolImpl(
                 .build());
         this.blacklist = blacklist;
         this.exceptionHandler = new CassandraRequestExceptionHandler(
-                this::getMaxRetriesPerHost, this::getMaxTriesTotal, blacklist);
+                () -> runtimeConfig.get().numberOfRetriesOnSameHost(),
```
tboam: We could just pass the supplier to the `ExceptionHAndler` now.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
fsamuel-bs: Yep. We used to extend it for the TTL CassandraKVS, but we removed it in atlasdb recently.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
fsamuel-bs: The supplier here is of CKVSRuntimeConfig. This is how we make a supplier of Integer.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
fsamuel-bs: Done.

fsamuel-bs:
Should have addressed all the comments. Ready for review again.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
```diff
@@ -58,14 +64,34 @@ public KeyValueService createRawKeyValueService(
...
+        return new CassandraReloadableKvsConfig(configWithNamespace, runtimeConfig);
+    }
+
+    @VisibleForTesting
+    Supplier<CassandraKeyValueServiceRuntimeConfig> preprocessKvsRuntimeConfig(
+            Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig) {
+        return () -> {
+            Optional<KeyValueServiceRuntimeConfig> configOptional = runtimeConfig.get();
+
+            return configOptional.map(config -> {
```
tboam: Does this mean that if I change the runtime config to something invalid (or invalid but of a different type), that the system will fall back to the default config?  Can we log a very clear message and keep the previous version instead?  Seems safer than potentially changing everything.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
tboam: Yeah, I meant give it the `Supplier<Config>` so it can just access what it needs.  Not important right now, we can do it if it starts reading more config parameters.

fsamuel-bs:
Verified that if the runtime config is empty, we correctly use the install config. Merging on the next green build.

tboam:
@fsamuel-bs just want to confirm some behaviour here: if I have install config set up (overriding default values) and then I add some runtime config.  What I end up with default config + runtime config.  That is, the runtime config replaces the install config rather than overriding each field.  Is this correct?

tboam:
So for each parameter individually we look at runtime, then install and then default.  Cool

fsamuel-bs:
Yep. That's correct.

Issue #2897, opened 2018-01-22
Row Components Matching Local Variables or Fields Can Break V2 Schemas


jeremyk-91:
```
        schema.addTableDefinition("foo", new TableDefinition() {{
            enableV2Table();

            rowName();
            hashFirstRowComponent();
            rowComponent("t", ValueType.STRING);

            columns();
            column("column", "c", ValueType.BLOB);
        }});
```
compiles, but the generated code doesn't compile because `t` is used as a field (for the transaction being used). Specifically, if any row component is named to be the same as any variable used in the generated code, we don't currently check for this and the resultant V2 table won't compile.

Fixing may require some careful consideration. One possibility is to add one or more underscores to the beginning of the variable names.

PR #2899, opened 2018-01-22
[MMM] V2 Schemas should support non-Object types

### LINKED ISSUE #2890: [V2 table] Can't generate getter for BLOB column [0.67..0.73]
This table definition:

```java
schema.addTableDefinition("foo", new TableDefinition() {{
    enableV2Table();

    rowName();
    hashFirstRowComponent();
    rowComponent("row", ValueType.STRING);

    columns();
    column("column", "c", ValueType.BLOB);
}});
```

Yields this error:

```
Exception in thread "main" java.lang.IllegalArgumentException: array types cannot be represented as a ClassName
	at com.squareup.javapoet.Util.checkArgument(Util.java:64)
	at com.squareup.javapoet.ClassName.get(ClassName.java:141)
	at com.palantir.atlasdb.table.description.render.TableClassRendererV2.renderNamedGetColumn(TableClassRendererV2.java:264)
	at com.palantir.atlasdb.table.description.render.TableClassRendererV2.renderNamedGet(TableClassRendererV2.java:235)
	at com.palantir.atlasdb.table.description.render.TableClassRendererV2.getMethods(TableClassRendererV2.java:193)
	at com.palantir.atlasdb.table.description.render.TableClassRendererV2.buildTypeSpec(TableClassRendererV2.java:140)
	at com.palantir.atlasdb.table.description.render.TableClassRendererV2.render(TableClassRendererV2.java:120)
	at com.palantir.atlasdb.table.description.render.TableRendererV2.render(TableRendererV2.java:38)
	at com.palantir.atlasdb.table.description.Schema.renderTables(Schema.java:326)
```

Because it attempts to wrap the java type corresponding to BLOB (`byte[]`) in a javapoet `ClassName` (that only supports object types).
Presumably the code should use `TypeName.get()` rather than `ClassName.get()` to handle such cases.

jeremyk-91:
**Goals (and why)**:
- Fix #2890 

- Use `TypeName` (supports array types) instead of `ClassName` (only Object types)
- Add a test that creates a V2 table with row components of all value-types apart from STRING (can only have 1 ending one), and column components of all value-types

- I didn't find any other V2 renderers apart from `TableRendererV2` which doesn't actually deal directly with components. Were there any other places I missed?

**Where should we start reviewing?**: `TableClassRendererV2`

@dansanduleac for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2899)


jeremyk-91:
*Verification:* Add Dan's schema to the `TodoSchema` in `atlas-ete-tests`, though the row component must be changed because of #2897.

```
schema.addTableDefinition("foo", new TableDefinition() {{
    enableV2Table();

    rowName();
    hashFirstRowComponent();
    rowComponent("rowComp", ValueType.STRING); // just "row" breaks things

    columns();
    column("column", "c", ValueType.BLOB);
}});
```

This passes if the schema compiles and the resultant generated code file also compiles.

jeremyk-91:
@tboam I'm going to go ahead and merge this as this ticket's verification is simple - nonetheless, note the commentary above. Might need a bit more clarity around what expectations are:

1. Is verification to be performed on the develop branch post-merge, or pre-merge? There are pros and cons to each.
2. Is verification intended to be done by the PR author, the reviewer, or a third party?
3. Is a write-up on verification methodology required and/or desired?

On past internal product we've worked on together, the answers were

1. post-merge
2. a third party which could also be either the author or the reviewer, I think.
3. I always did this, though I don't think it was necessarily enforced.

PR #2900, opened 2018-01-22
Check if thread is interrupted and stop to Sweep

### LINKED ISSUE #2458: Interrupt BackgroundSweeper when TransactionsManager is closed
When we close the TransactionsManager, the BackgroundSweeper thread can continue to run. Since it will just fail, as we also close all Cassandra pools, we could interrupt it's thread earlier.

fsamuel-bs:
**Goals (and why)**: This should fix #2458. Even though it's unclear why we weren't throwing `InterruptedException` it might be the case that the thread was interrupted when on calls that do not throw such an exception (not sure if Thrift calls throw Interrupted Exception, even though they should), or maybe we're wrapping an InterruptedException wrongly — i.e. I think if we're wrapping Thrift's InterruptedException to AtlasDbDependencyException on the KVS layer.

In any case, this should do the trick.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
@@ -121,6 +121,10 @@ public final class BackgroundSweeperImpl implements BackgroundSweeper {
             Thread.sleep(getBackoffTimeWhenSweepHasNotRun());
             log.info("Starting background sweeper.");
             while (true) {
+                if (Thread.currentThread().isInterrupted()) {
+                    break;
+                }
+
                 SweepOutcome outcome = checkConfigAndRunSweep(locks);
 
                 log.info("Sweep iteration finished with outcome: {}", SafeArg.of("sweepOutcome", outcome));


fsamuel-bs:
@tboam for SA.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -121,6 +121,13 @@ public void run() {
             Thread.sleep(getBackoffTimeWhenSweepHasNotRun());
             log.info("Starting background sweeper.");
             while (true) {
+                // InterruptedException might be wrapped in RuntimeException (i.e. AtlasDbDependencyException),
+                // which would be caught downstream.
+                // Therefore we need to actively check if the thread has been interrupted here, and exit sweep if so.
+                if (Thread.currentThread().isInterrupted()) {
+                    break;
```
tboam: Can we rethrow the InterruptedException instead?  This way we hit the catch block below and move to SHUTDOWN state.  Useful for metrics and monitoring

fsamuel-bs: Done.

Issue #2902, opened 2018-01-22
Indexes cannot be marked safe for logging


hsaraogi:

hsaraogi:
The safety can be derived from the table name.

Issue #2901, opened 2018-01-22
Table metadata deserialiser/serializer do not work.


hsaraogi:
```
public class TableMetadataDeserializerTest {

    private File file = new File("output.json");


    @Test
    public void canSerializeAndDeserializeTableMetadata() throws Exception {

        TableMetadata genericTableMetadata = new TableMetadata();

        ObjectMapper mapper = new ObjectMapper();
        SimpleModule module = new SimpleModule();
        module.addSerializer(new TableMetadataSerializer());
        module.addDeserializer(TableMetadata.class, new TableMetadataDeserializer());
        mapper.registerModule(module);

        mapper.writeValue(file, genericTableMetadata);
        TableMetadata readValue = mapper.readValue(new BufferedReader(new FileReader(file)).readLine(), TableMetadata.class);

        assertThat(readValue, is(genericTableMetadata));

    }
}

```

This test fails.

hsaraogi:
We should remove this after checking that it is unused.

PR #2903, opened 2018-01-22
Internal Tables should be safe for logging

hsaraogi:
**Goals (and why)**: Mark safety for internal tables.

**Implementation Description (bullets)**: Add LogSafety.SAFE to table metadata where possible, manually add the DEFAULT_METADATA_TABLE. 

1. TableMetadata({}/De)serializer doesnt respect log safety. Its broken in other ways too #2901.
2. Indexes are not safe still. I think their safety should be derived from the table, that is filed as a separate issue though #2902.

**Where should we start reviewing?**: TableMetadata

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2903)


atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
```diff
@@ -62,7 +63,8 @@ private AtlasDbConstants() {
 
     public static final TableReference PARTITION_MAP_TABLE = TableReference.createWithEmptyNamespace("_partition_map");
     public static final byte[] EMPTY_TABLE_METADATA = {}; // use carefully
-    public static final byte[] GENERIC_TABLE_METADATA = new TableMetadata().persistToBytes();
+    public static final byte[] GENERIC_TABLE_METADATA = new TableMetadata(TableMetadataPersistence.LogSafety.SAFE)
+            .persistToBytes();
```
jeremyk-91: I've checked for uses of this and it is safe, but I think we should do one of

1. leave the old `GENERIC_TABLE_METADATA` as is (maybe with a comment that it exists to avoid breaking the API), and add a new constant `GENERIC_SAFE_TABLE_METADATA` which is this constant.
2. make a note that this is a dev break in the release notes (it is, because `GENERIC_TABLE_METADATA` used to be unsafe)

atlasdb-client/src/main/java/com/palantir/atlasdb/logging/SafeLoggableDataUtils.java
```diff
@@ -54,6 +55,8 @@ static void addLoggableNamesToBuilder(
         if (IS_SAFE.test(tableMetadata.getNameLogSafety())) {
             builder.addPermittedTableReferences(ref);
         }
+        // this is a system table with empty metadata, but safe for logging.
+        builder.addPermittedTableReferences(AtlasDbConstants.DEFAULT_METADATA_TABLE);
```
jeremyk-91: This is a little hacky but I don't see an easy alternative (and don't really fancy making `EMPTY_TABLE_METADATA` be safe), so this is fine.

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
hsaraogi: I dont think anyone uses `GENERIC_TABLE_METADATA` directly, checked the large internal product too, so going with option 2.

jeremyk-91: Yep, I should have mentioned that I checked internally too!

PR #2905, opened 2018-01-22
Make _sweep_progress2 table hidden

sandorw:
**Goals (and why)**: Hide the _sweep_progress2 table to resolve https://github.com/palantir/atlasdb/issues/2881

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2905)


atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
```diff
@@ -45,6 +45,9 @@ private AtlasDbConstants() {
     public static final TableReference DEFAULT_SCHEMA_METADATA_TABLE = TableReference.createWithEmptyNamespace(
             "_schema_metadata");
 
+    // Deprecated tables
+    public static final TableReference SWEEP_PROGRESS_V2 = TableReference.createWithEmptyNamespace("_sweep_progress2");
```
hsaraogi: There used to be `sweep_progress` (the version 1) too, but I think that was sweepable, so dont need to blacklist it :).

PR #2910, opened 2018-01-23
TokenRangeWriteLogging metrics are updated more often

tboam:
**Goals (and why)**: We want to see metrics on uneven writes appear quicker as few tables were publishing numbers.  I believe this is because few tables accumulate 1,000,000 writes between restarts.

**Implementation Description (bullets)**: The probability of uneven token writes is updated every 1000 writes, logs happen every 6 hours and these 2 things are independent.
I had to update a bunch of tests to get them passing and deleted a couple I think didn't make sense now.

**Concerns (what feedback would you like?)**: Is this approach valid?

**Where should we start reviewing?**: TokenRangeWritesLogger

**Priority (whenever / two weeks / yesterday)**: this week if possible

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2910)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TokenRangeWritesLogger.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TokenRangeWritesLogger.java
@@ -16,6 +16,7 @@
 
 package com.palantir.atlasdb.keyvalue.cassandra;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TokenRangeWritesLogger.java
```diff
@@ -43,9 +44,9 @@
     private static final Logger log = LoggerFactory.getLogger(TokenRangeWritesLogger.class);
 
     @VisibleForTesting
-    static final long THRESHOLD_WRITES_PER_TABLE = 1_000_000L;
+    static final long THRESHOLD_WRITES_PER_TABLE = 1_000L;
     @VisibleForTesting
-    static final long TIME_UNTIL_LOG_MILLIS = 24 * 60 * 60 * 1000L;
+    static final long TIME_UNTIL_LOG_MILLIS = Duration.ofHours(6).toMillis();
```
gmaretic: Nice

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TokenRangeWritesLogger.java
```diff
@@ -139,9 +149,9 @@ private void logNotUniform() {
         }
 
         private void logUniform() {
-            log.info("There were at least {} writes into the table {} since the last statistical analysis. The "
-                            + "distribution of writes over token ranges does not appear to be skewed.",
-                    SafeArg.of("numberOfWrites", writesSinceLastLog.get()),
+            log.info("The distribution of writes over token ranges does not appear to be skewed"
+                            + " based on {} writes to table {}.",
+                    SafeArg.of("numberOfWrites", numberOfWritesUsedToCalculateDistribution),
```
gmaretic: That's useful!

tboam:
I'm not going to add the other metric just now given its cardinality.  I don't think we're going to create a monitor on this value, more likely it will be occasionally reviewed by product owners and they can analyse the value it is tending towards.  Or we can use a smoothing function to flatten out restarts.

PR #2920, opened 2018-01-25
Add shared secret when getting a C* metric

hsaraogi:
**Goals (and why)**: The C*MetricService now has a shared secret, we should be compatible with it.
This is for short-term testing of the metrics service internally. Longer term we dont want to copy the classes. @gsheasby is looking at implementing the shim to be replaced internally.

**Implementation Description (bullets)**: copy up the C*metricsService, sharedSecret and other classes. Add the secret to the install config.

**Where should we start reviewing?**: CMS.java

**Priority (whenever / two weeks / yesterday)**: soon.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2920)

### DIFF
--- a/qos-service-impl/src/main/java/com/palantir/cassandra/sidecar/SharedSecretHeader.java
+++ b/qos-service-impl/src/main/java/com/palantir/cassandra/sidecar/SharedSecretHeader.java
@@ -1,12 +1,24 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


qos-service-impl/src/main/java/com/palantir/cassandra/sidecar/SharedSecretHeader.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
```
ashrayjain: Don't think these work outside.

qos-service-impl/src/main/java/com/palantir/cassandra/sidecar/metrics/CassandraMetricsService.java
```diff
@@ -1,36 +1,24 @@
...
- * You may obtain a copy of the License at
- *
- * http://opensource.org/licenses/BSD-3-Clause
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
```
ashrayjain: same

ashrayjain:
Generally not in favor of doing this since there are better ways to handle this :(

PR #2923, opened 2018-01-25
Fix/generify metrics service

gsheasby:
**Goals (and why)**: Give a couple of classes more generic names

**Concerns (what feedback would you like?)**: This PR couldn't take us very far, because some of the implementation rely on JaxRs stuff - if we removed the Jersey annotations from `MetricsService`, we'd need to move the implementation details internal, which would mean adapting `TransactionManagers`. This would satisfy the Dependency Inversion Principle, but could also be high lift.

**Where should we start reviewing?**: MetricsService / HealthMetric

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2923)

### DIFF
--- a/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricsClientLimitMultiplier.java
+++ b/qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricsClientLimitMultiplier.java
@@ -25,8 +25,6 @@ import com.palantir.atlasdb.qos.config.QosCassandraMetricsInstallConfig;
 import com.palantir.atlasdb.qos.config.QosCassandraMetricsRuntimeConfig;
 import com.palantir.atlasdb.qos.config.ThrottlingStrategy;
 import com.palantir.atlasdb.qos.metrics.MetricsService;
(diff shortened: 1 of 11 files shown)


qos-service-impl/src/main/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricsClientLimitMultiplier.java
```diff
@@ -42,8 +42,8 @@ private CassandraMetricsClientLimitMultiplier(
 
     public static ClientLimitMultiplier create(Supplier<QosCassandraMetricsRuntimeConfig> runtimeConfig,
             QosCassandraMetricsInstallConfig installConfig, ScheduledExecutorService metricsLoaderExecutor) {
-        CassandraMetricsService metricsService = JaxRsClient.create(
-                CassandraMetricsService.class,
+        MetricsService metricsService = JaxRsClient.create(
```
hsaraogi: This would mean that we try to run a getMetric RPC on the cassandraService, which will fail with a `NotFoundException` as the Cassandra service doesnt have the `MetricsService` endpoint. We need to create the client internally to the actual CassandraMetricsService by taking the required internal dependency.

PR #2924, opened 2018-01-25
CassandraClient slow log

fsamuel-bs:
**Goals (and why)**: Internal reference PDS-62553.

Add information about durations of calls and size of bytes fetched from and written to Cassandra.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2924)


fsamuel-bs:
Lacks tests for the new ThriftObjectSizeUtils method.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ProfilingCassandraClient.java
```diff
@@ -69,12 +70,18 @@ public ProfilingCassandraClient(CassandraClient client) {
...
                         LoggingArgs.tableRef(tableRef),
-                        SafeArg.of("number of keys", numberOfKeys),
-                        SafeArg.of("number of columns", numberOfColumns),
+                        LoggingArgs.rowCount(numberOfKeys),
+                        LoggingArgs.columnCount(numberOfColumns),
                         SafeArg.of("consistency", consistency_level.toString()),
-                        SafeArg.of("kvsMethodName", kvsMethodName)));
+                        SafeArg.of("kvsMethodName", kvsMethodName),
+                        LoggingArgs.durationMillis(timer)),
+                (logger, rowsToColumns) -> logger.log("and returned {} cells in {} rows with {} bytes",
```
sandorw: Why do we split up these log lines?

tboam: I don't know, but it is very annoying to read these logs.  If we can combine them during this PR it would be very helpful.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/qos/ThriftObjectSizeUtils.java
```diff
@@ -53,6 +54,28 @@ public static long getApproximateSizeOfMutationMap(Map<ByteBuffer, Map<String, L
...
+
+        mutation_map.forEach((key, tableToMutations) -> {
+            long keySize = ThriftObjectSizeUtils.getByteBufferSize(key);
+
+            tableToMutations.forEach((table, mutations) -> {
+                if (!tableToSize.containsKey(table)) {
+                    tableToSize.put(table, 0L);
+                }
+
+                Long size = tableToSize.get(table);
```
sandorw: could use `getOrDefault(..., 0L)` and drop the above loop

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -16,6 +16,8 @@
 package com.palantir.atlasdb.sweep;
 
 import java.util.Arrays;
+import java.util.HashMap;
```
sandorw: Are these used?

fsamuel-bs: Removed.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ProfilingCassandraClient.java
fsamuel-bs: 1. Note that this does not create two log lines. Our logger is a bit special here, and just concatenates these two log lines, printing just once, when the call to Cassandra returns with either a result or an exception.

2. We need to have two methods because the second one is called when the call to Cassandra returns and we have access to the result of the call. The call might fail, and in this case we concatenate an "and failed with exception {}" at the end of the log line.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/qos/ThriftObjectSizeUtils.java
fsamuel-bs: Done.

PR #2927, opened 2018-01-25
Mark sweep outcome as tags

fsamuel-bs:
**Goals (and why)**: Mark sweep outcome as tags. This will make it easier to plot graphs in our internal metric analyzer tool.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2927)


fsamuel-bs:
It's the same as creating metrics with the outcome in the name — which is what we're doing before. This just makes it easier to aggregate them.

PR #2932, opened 2018-01-29
SSSS Part 1: deleteStreams()

jeremyk-91:
// This is still a WIP in that I've finished coding up `deleteStreams()`, but need to test it to make sure that the row and cell generators agree with the code from a generated stream store. The high-level implementation is not likely to change dramatically (unless of course there turn out to be serious problems with this approach).

**Goals (and why)**:
- Implement programmatic deletion of streams from a stream store without needing to know a product's full schema, instead relying on the data from #2748. This is needed for Nimbus.

- Implement deleteStreams() without needing to know about an Atlas client's schema
- Completing the implementation of `FollowerGeneratingCleanupMetadataVisitor` and wiring it into Sweep is not the scope of this PR

- Should the deleter reference `V1`? Or would it be fine to, if/when we change the schema, declare `V2`?
- Does the wrapping of `GenericStreamIdentifier` make sense?

**Where should we start reviewing?**: `SchemalessStreamStoreDeleter` hopefully splits this nicely into enough pieces.

**Priority (whenever / two weeks / yesterday)**: this week? I realise this is large so happy to talk through things in person too.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2932)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    // TODO (jkong): Main method needs tests, both unit tests and an integration test!
+    public void deleteStreams(Transaction tx, Set<GenericStreamIdentifier> streamIds) {
+        if (streamIds.isEmpty()) {
+            log.debug("deleteStreams() was called with no identifiers, so we are returning.");
+            return;
+        }
+
+        // Safe, because these are longs.
```
gsheasby: You're somewhat at the mercy of the person who creates the `GenericStreamIdentifier` here; making that class (or at least its constructors) package-private would protect you here.

In general I've started biasing towards package-private where possible, to avoid exposing new classes/methods to our public API.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
```diff
@@ -0,0 +1,127 @@
...
+    private StreamStoreCellDeletion findCellsToPropagateDeletes(
+            Map<GenericStreamIdentifier, StreamMetadata> streamIdentifierToMetadata) {
+        ImmutableStreamStoreCellDeletion.Builder builder = ImmutableStreamStoreCellDeletion.builder();
+
+        for (Map.Entry<GenericStreamIdentifier, StreamMetadata> entry : streamIdentifierToMetadata.entrySet()) {
+            GenericStreamIdentifier streamId = entry.getKey();
+            StreamMetadata streamMetadata = entry.getValue();
+
+            builder.addAllValueTableCellsToDelete(cellCreator.constructValueTableCellSet(
+                    streamId, getNumberOfBlocksFromMetadata(entry.getValue())));
```
gsheasby: nit: use `streamMetadata` here

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
```diff
@@ -0,0 +1,127 @@
...
+                LoggingArgs.tableRef("metadataTable", getTableReference(StreamTableType.METADATA)),
+                UnsafeArg.of("hashCells", deletion.hashTableCellsToDelete()),
+                LoggingArgs.tableRef("hashTable", getTableReference(StreamTableType.HASH)));
+    }
+
+    private void transactionallyDeleteCells(Transaction tx, StreamTableType streamTableType, Set<Cell> cellsToDelete) {
+        tx.delete(getTableReference(streamTableType), cellsToDelete);
+    }
+
+    @VisibleForTesting
```
gsheasby: General tip: I guess you added the methods as `@VisibleForTesting` because it didn't feel necessary/appropriate for this class to expose the methods publicly, which is a hint that the logic belongs elsewhere.

Consider inverting control here and having a public method `StreamTableType#getTableName(namespace, streamStoreShortName)` (and similarly for the metadata)?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
jeremyk-91: Doh! Will fix this

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
jeremyk-91: Hmm, `GenericStreamIdentifier` could indeed be package private. In general, this pattern is already used in e.g. http-remoting. Will change this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/SchemalessStreamStoreDeleter.java
jeremyk-91: Yep, that could work. The methods are tricky enough (well, maybe not `getTableReference`...) that I want them tested as well (as opposed to simply leaving them private). I though they were quite small, but it does make sense for gTR to be in StreamTableType (and same for the stream metadata)

jeremyk-91:
To answer the general comments:

- we want to do the whole delete in a single transaction: This is consistent with the existing behaviour. Deletes should be roughly linear in the size of the streams being deleted but with a small coefficient (since you need to write `0x` in place of a 1MB blob). If this is bad, see the last comment; we might be able to split a batched delete up.
- retry logic: This is typically called as part of a cleanup task (e.g. see `DataMetadataCleanupTask` which also needs to be generically implemented). These go through a standard `runTaskWithRetry`; see `CleanupFollower` for where it's expected to be called from. My plan was to have Nimbus use the existing follower framework.
- `deleteStreams` with many streams: We could implement another level of batching somewhere, which splits the set into multiple components if possible (`streamDeleteBatchHint`)? 😅

gsheasby:
To answer your answer, that all seems fine. Deletes should be small enough that timeouts/failures won't be an issue, so it doesn't seem necessary to implement batching at this point. The retry logic is also sound.

jeremyk-91:
closing for now - we can reopen later if needed

PR #2939, opened 2018-01-29
Add TimeLockStatus.PENDING_ELECTION

fsamuel-bs:
**Goals (and why)**: Add TimeLockStatus.PENDING_ELECTION when no clients have been created.

**Concerns (what feedback would you like?)**: Is it too hacky to expose number of clients as a public method of the resource?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2939)

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,8 +50,10 @@ develop
     *    - Type
          - Change
 
-    *    -
-         -
+    *    - |improved|
+         - Added a TimeLock healthcheck for signalling that no leader election has been triggered.
+           This will allow TimeLock itself to broadcast a HEALTHY status even without a leader.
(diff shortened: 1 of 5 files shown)


fsamuel-bs:
Lacks tests.

fsamuel-bs:
This should be used by the timelock wrapper to report HEALTHY when NO_CLIENTS are present.

tpetracca:
Why is leader election predicated on the existence of a client. That to me seems to be the core issue here.

fsamuel-bs:
I think that currently, the leader election is a function of the client, i.e. each client creates a timestamp and lock services and wrap them in a AwaitingLeadershipProxy — a proxy that throws if you're not the leader. Changing this would be a much bigger change, and would possibly hinder future optimizations — multiple leaders per cluster (one per client), for example.

This is the minimal change which will bring us what we want, even though a little bit counterintuitive.

Issue #2852, opened 2018-01-11
clean-cass-locks-state throws "Multiple schema mutation lock tables have been created"


hsaraogi:
On running:
PCLOUD [palantir@xxxx service]$ ./bin/atlasdb-cli --offline -c var/conf/install.yml clean-cass-locks-state

We got:

```
19:36:14.149 [main calling cassandra host keppel-soa-cassandra-16.ops.palantir.local/10.3.41.149:9160 started at 2018-01-11T19:36:14.147Z - 5] DEBUG c.p.a.k.c.CassandraClientPoolingContainer - During com.palantir.atlasdb.keyvalue.cassandra.qos.QosCassandraClient@640f11a1 check-in, cleaned up a read buffer of 1277 bytes of host keppel-soa-cassandra-16.ops.palantir.local
19:36:14.153 [main] ERROR com.palantir.atlasdb.cli.AtlasCli - Fatal exception thrown during cli command execution.  Exiting.
java.lang.IllegalStateException: Multiple schema mutation lock tables have been created.
This happens when multiple nodes have themselves as lockLeader in the configuration.
Please ensure the lockLeader is the same for each node, stop all Atlas clients using this keyspace, restart your cassandra cluster and run the clean-cass-locks-state cli command.
The tables that clashed were: [_locks_5fe819e3_7ff0_4884_b3a7_b7f96091bc58, _locks_c146b83a_27ae_40b4_9498_2fe69959dbb8]
        at com.palantir.atlasdb.keyvalue.cassandra.UniqueSchemaMutationLockTable.getSingleTable(UniqueSchemaMutationLockTable.java:87) ~[atlasdb-cassandra-0.72.0.jar:0.72.0]
        at com.palantir.atlasdb.keyvalue.cassandra.UniqueSchemaMutationLockTable.ensureLockTableExists(UniqueSchemaMutationLockTable.java:80) ~[atlasdb-cassandra-0.72.0.jar:0.72.0]
        at com.palantir.atlasdb.keyvalue.cassandra.UniqueSchemaMutationLockTable.getOnlyTable(UniqueSchemaMutationLockTable.java:48) ~[atlasdb-cassandra-0.72.0.jar:0.72.0]
        at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl.tryInitialize(CassandraKeyValueServiceImpl.java:320) ~[atlasdb-cassandra-0.72.0.jar:0.72.0]
[...]

PR #2943, opened 2018-01-30
Fix clean cass locks state cli, Part 1: Extract all the things

gsheasby:
**Goals (and why)**: Extract objects so that `CassandraKeyValueServiceImpl.cleanUpSchemaMutationLockTablesState` does not depend on any other part of `CassandraKeyValueService`.
This is a preliminary of fixing #2852 - the remaining work is to extract an object for `cleanUpSchemaMutationLockTablesState` and then to use that in `CleanCassLocksStateCommand`. Each commit here is a refactor and should not change behaviour.
As a pleasant side-effect, more than 400 lines of code have been removed from `CassandraKeyValueServiceImpl` into classes with single, clearly-defined responsibilities. The diff looks quite big but the vast majority is simply moving code around.

- Remove the hard dependency on `UniqueSchemaMutationLockTable` from `SchemaMutationLock` (so that in the next PR, we can sidestep the constraint)
- Commits starting with "move" or "delegate" consist of moving methods, usually out of `CassandraKeyValueServiceImpl`, to new classes.
- Started by moving smaller parts with few dependencies, until I was able to move out larger parts.

- Most of the new classes are not directly tested, but are indirectly tested by the (admittedly extensive) suite of CassandraKVS tests. I did not take the time to write explicit/direct unit tests yet; is it OK to defer this to a "small tasks" slot?
- Other concerns, where relevant, I included in the commit messages.
**- Is it worth splitting up the commits into separate PRs, or should I just push the next parts onto this PR?**

**Where should we start reviewing?**: Go commit-by-commit. Please pay attention to the commit messages, as I have called out specific things there.

**Priority (whenever / two weeks / yesterday)**: Today - there's a fairly high potential for merge conflicts, and I will be building on top of this branch to actually fix the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2943)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/HostPartitioner.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimaps;
+import com.palantir.atlasdb.keyvalue.api.Cell;
+
+class HostPartitioner<V> {
+    private final CassandraClientPool cassandraClientPool;
+
+    HostPartitioner(CassandraClientPool cassandraClientPool) {
```
sandorw: nit: looks like these methods could be static instead - only one uses this variable and it could be passed in as an argument.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TimestampsLoader.java
```diff
@@ -0,0 +1,41 @@
...
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+
+class TimestampsLoader {
+    private CellLoader cellLoader;
+
+    TimestampsLoader(CellLoader cellLoader) {
+        this.cellLoader = cellLoader;
+    }
+
+    Multimap<Cell, Long> getAllTimestamps(TableReference tableRef, Set<Cell> cells, long ts,
```
sandorw: +1 to combining with CellLoader

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TableReference.java
```diff
@@ -79,6 +79,17 @@ private TableReference(
         this.tablename = tablename;
     }
 
+    /** @deprecated uses TableReference.createUnsafe, which is itself deprecated.
```
sandorw: nit: docs formatting

PR #2547, opened 2017-10-24
Do not shutdown the async initializer executor

fsamuel-bs:
**Goals (and why)**: Do not shutdown `AsyncInitializer` executor as there might have been a `cancelInitialization` call to the initializing object, and we need to be able to process its handler.

**Implementation Description (bullets)**: -

**Concerns (what feedback would you like?)**: -

**Where should we start reviewing?**: -

**Priority (whenever / two weeks / yesterday)**: -

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2547)

PR #2520, opened 2017-10-19
TimeLock + Async Initialization, Part 1

jeremyk-91:
**Goals (and why)**:
- Previously, for Cassandra we would attempt to immediately migrate the timestamp bound from Cassandra to TimeLock on startup, which would fail if either of them was unavailable. For DBKVS or other key-value services, we would attempt to ping TimeLock on startup, which would fail if TimeLock was available (though the KVS could be unavailable).
- Fix #2500
- Drive by fix of a thread leak in the async initializer (it doesn't shut down the executor service after successfully initializing the client)

**Implementation Description (bullets)**:
- Make TimeLockMigrator an AsyncInitializer
- Change SerializableTransactionManager to take a Supplier<Boolean> instead of an AsyncInitializer for its TransactionManagersInitializer check
- Require both TLM and TMI to be happy before that prerequisite passes.

**Concerns (what feedback would you like?)**:
- Is the TransactionManagers part too hacky?
- Are there any pitfalls I've missed out on?

**Where should we start reviewing?**: TimeLockMigrator

**Priority (whenever / two weeks / yesterday)**: next week, this would probably block GA of Async Initialization

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2520)

PR #2945, opened 2018-01-30
Close async initializers executors

fsamuel-bs:
**Goals (and why)**: Close async initializer executors. This had been attempted on #2520 and reverted on #2547. We avoid the issues discussed before by scheduling the shutdown of the executor — if the there's a close task to be ran, we shutdown the executor after it is ran.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2945)


atlasdb-commons/src/test/java/com/palantir/async/initializer/AsyncInitializerTest.java
```diff
@@ -58,11 +58,20 @@ protected String getInitializingClassName() {
...
         @Override
         ScheduledExecutorService getExecutorService() {
-            deterministicScheduler = new DeterministicScheduler();
+            deterministicScheduler = new DeterministicSchedulerShutdownAware();
             return deterministicScheduler;
         }
     }
 
+    private class DeterministicSchedulerShutdownAware extends DeterministicScheduler {
+        int numberOfTimesShutdownCalled = 0;
```
sandorw: Should probably be synchronized or volatile to avoid the test flaking.

PR #2947, opened 2018-01-31
Load Simulation

cbrockington:
**Goals (and why)**:
 - Provide a way of capturing and replaying transactions.  Useful for simulating increased load in atlas, without mutating underlying state, or directly requiring more humans to be clicking buttons.

A wrapper around the SerializableTransactionManager (WrappingSerializableTransactionManager) which allows for submitted tasks to be wrapped before they are passed along to the delegate manager.  The SuccessfulTaskInvocationCapture wraps these tasks to firstly (if a condition passes) wrap the Transaction in order to record following invocations, then after the original task is executed, submits the CapturedTransactions to a provided consumer.  In this first pass, the consumer is a TransactionReplayer which takes the CapturedTransactions and replays them.

The load-simulation-impl/src/main/java/com/palantir/atlasdb/simulated/LoadSimulator.java file is ultimately where the wiring of concrete classes is done so probably there.

**Do not merge until at least:**
- [ ] mavenLocal is removed
- [ ] git history is cleaned
- [ ] logging and documentation are sorted
- [x] load-simulation-* projects have their build.gradles cleaned up.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2947)


cbrockington:
Another thing I don't actually like at the moment is the endpoints for updating the rate of capture and replay.  I have largely put them in since ultimately I think it would be good to give Apollo and blue/green deploys some way of spiking load temporarily for testing.

cbrockington:
Review status: 0 of 35 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java, line 305 at r2](https://reviewable.io:443/reviews/palantir/atlasdb/2947#-L4BVuXGBFqJWw0iyppw-r2-305:-L59cG2cFq1HREUgPr2B:bwygxsx) ([raw file](https://github.com/palantir/atlasdb/blob/affea9be033524f8ecd97b9ca1e07cfac1462692/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L305)):*
<details><summary><i>Previously, jeremyk-91 (Jeremy Kong) wrote…</i></summary><blockquote>

Just note that this is typically only called once, so even though load simulation is in runtime config you can't live reload to enable (or disable) it
</blockquote></details>

Noted.

---

*[load-simulation-api/src/main/java/com/palantir/atlasdb/transaction/api/SamplingTransactionCondition.java, line 39 at r2](https://reviewable.io:443/reviews/palantir/atlasdb/2947#-L4BVuXH0EEJf1FRrrEh-r2-39:-L59cIqvTyYCqQAzNKau:bb6un1j) ([raw file](https://github.com/palantir/atlasdb/blob/affea9be033524f8ecd97b9ca1e07cfac1462692/load-simulation-api/src/main/java/com/palantir/atlasdb/transaction/api/SamplingTransactionCondition.java#L39)):*
<details><summary><i>Previously, jeremyk-91 (Jeremy Kong) wrote…</i></summary><blockquote>

nit: isn't 0.0 or 0d enough?
</blockquote></details>

Done

---

*[load-simulation-impl/src/main/java/com/palantir/atlasdb/transaction/impl/RefreshableReplayRepetition.java, line 24 at r2](https://reviewable.io:443/reviews/palantir/atlasdb/2947#-L4BVuXH0EEJf1FRrrEu-r2-24:-L59cK2XiX02gg2wRDMI:bb6un1j) ([raw file](https://github.com/palantir/atlasdb/blob/affea9be033524f8ecd97b9ca1e07cfac1462692/load-simulation-impl/src/main/java/com/palantir/atlasdb/transaction/impl/RefreshableReplayRepetition.java#L24)):*
<details><summary><i>Previously, jeremyk-91 (Jeremy Kong) wrote…</i></summary><blockquote>

see note on the other Refreshable class - seems like this should be volatile
</blockquote></details>

Done

---

*[load-simulation-impl/src/main/java/com/palantir/atlasdb/transaction/impl/RefreshableTransactionTaskCondition.java, line 29 at r2](https://reviewable.io:443/reviews/palantir/atlasdb/2947#-L4BVuXH0EEJf1FRrrEv-r2-29:-L59cLTb6B_e6xtVsslu:bb6un1j) ([raw file](https://github.com/palantir/atlasdb/blob/affea9be033524f8ecd97b9ca1e07cfac1462692/load-simulation-impl/src/main/java/com/palantir/atlasdb/transaction/impl/RefreshableTransactionTaskCondition.java#L29)):*
<details><summary><i>Previously, jeremyk-91 (Jeremy Kong) wrote…</i></summary><blockquote>

This _feels_ like it should be volatile (to ensure that writes from whichever thread puts in an update are visible to others)
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/atlasdb/2947)*

cbrockington:
It is, though I've separated all the functionality out into a separate project which can wrap things instead.

Closing.

PR #2951, opened 2018-01-31
Add Sweep Metric for table being swept

gmaretic:
**Goals (and why)**:
Expose a metric that tells us which table is currently being swept.

Reuse SweepMetrics, but needs additional generifying to allow for the update method to take a String argument.
 
Is this a horrible use of generics?

Wherever

Soon, this is useful.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2951)


gmaretic:
1) is done, 2) does not seem that useful because we can't get useful historical data from meters sicne we can only see exponentially-weighted moving averages :(

gmaretic:
@jeremyk-91 Can you take over reviewing this for @fsamuel-bs ? I addressed his comments, but would be good to get a fresh pair of eyes because the scope of this PR grew quite a bit

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AccumulatingValueMetric.java
```diff
@@ -0,0 +1,39 @@
...
+        return delegate.getValue();
+    }
+
+    @Override
+    public void setValue(Long newValue) {
+        delegate.setValue(newValue);
+    }
+
+    public void accumulateValue(Long newValue) {
+        setValue(newValue + getValue());
```
tboam: This is not thread safe.  Is it guaranteed that this will only be called in a single thread?  Does this guarantee still hold if Nimbus is sweeping many products at once?

We should either make this safe or make it very clear to our future selves that this is an issue we need to fix.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SpecificTableSweeper.java
```diff
@@ -128,11 +127,10 @@ void runOnceAndSaveResults(TableToSweep tableToSweep, SweepBatchConfig batchConf
 
     SweepResults runOneIteration(TableReference tableRef, byte[] startRow, SweepBatchConfig batchConfig) {
         try {
-            SweepResults results = sweepRunner.run(tableRef, batchConfig, startRow);
+            SweepResults results = sweepRunner.runWithMetricsUpdate(tableRef, batchConfig, startRow,
+                    this::updateMetricsDeleteBatch);
```
tboam: I dislike passing anything to do with metrics down here.  We have a nice separation of sweep's actual work and it updating metrics which this breaks.  My preferred solution would be to include more fields in the SweepResults object (or a new object containing a SweepResults and other information) that hold this information and the metrics update code continues to live in one place.

tboam: Failing that pass the `sweepMetricsManager` into the `sweepRunner` at construction time so the flow of code doesn't move back and forth between these 2 classes.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsManagerTest.java
```diff
@@ -110,204 +107,131 @@ public void tearDown() {
...
 
-        assertRecordedHistogramNonTaggedFullTable(CELLS_EXAMINED, TABLE_REF, EXAMINED);
-        assertRecordedHistogramNonTaggedFullTable(CELLS_SWEPT, TABLE_REF, DELETED);
-        assertRecordedHistogramNonTaggedFullTable(TIME_SPENT_SWEEPING, TABLE_REF, TIME_SWEEPING);
-        assertSweepTimeElapsedHistogramWithinMarginOfError(START_TIME);
+        assertRecordedCurrentValue(ImmutableMap.of(
+                CELLS_EXAMINED, OTHER_EXAMINED,
+                CELLS_SWEPT, OTHER_DELETED,
+                TIME_SPENT_SWEEPING, OTHER_TIME_SWEEPING,
+                TABLE_BEING_SWEPT, UNSAFE_FULLY_QUALIFIED));
```
jeremyk-91: nit: some of these blocks feel like they can be extracted to a common helper, though as this is test code I don't actually mind the verbosity too much.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepTaskRunner.java
```diff
@@ -179,6 +200,11 @@ private SweepResults doRun(TableReference tableRef,
                         candidates, batchConfig, sweepableCellFilter, limit);
             long totalCellTsPairsExamined = 0;
             long totalCellTsPairsDeleted = 0;
+
+            if (metricsManager.isPresent()) {
+                metricsManager.get().resetBeforeDeleteBatch();
+            }
```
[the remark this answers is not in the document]
gmaretic: 🐑

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepTaskRunner.java
```diff
@@ -189,10 +215,16 @@ private SweepResults doRun(TableReference tableRef,
...
-                        2 * batchConfig.deleteBatchSize());
+                long cellsDeleted = sweepBatch(tableRef, batch.cells(), runType, 2 * batchConfig.deleteBatchSize());
+                totalCellTsPairsDeleted += cellsDeleted;
+
+                long cellsExamined = batch.numCellTsPairsExamined();
+                totalCellTsPairsExamined += cellsExamined;
+
+                if (metricsManager.isPresent()) {
+                    metricsManager.get().updateAfterDeleteBatch(cellsExamined, cellsDeleted);
+                }
```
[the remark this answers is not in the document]
gmaretic: 🥇

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/SweeperTestSetup.java
```diff
@@ -113,11 +115,11 @@ protected void setProgress(SweepProgress progress) {
 
     protected void setNextTableToSweep(TableReference tableRef) {
         Mockito.doReturn(Optional.of(tableRef)).when(nextTableToSweepProvider)
-                .getNextTableToSweep(Mockito.any(), Mockito.anyLong());
+                .getNextTableToSweep(any(), Mockito.anyLong());
```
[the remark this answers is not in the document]
gmaretic: Might as well do all Mockito things then :)

gmaretic:
@jeremyk-91 Are we good?

PR #2952, opened 2018-01-31
Remove old comment from AtlasDbRuntimeConfig

bkrieger:
As of https://github.com/palantir/atlasdb/pull/2850, the runtime timelock block is _not_ ignored when the install config does not specify usage of TimeLock.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2952)


bkrieger:
@hsaraogi for review.

gsheasby:
Thanks for the correction!
I've replaced this by #2953 as we don't run builds on forked PRs.

PR #2953, opened 2018-01-31
Fix timelockRuntime javadoc

gsheasby:
As of #2850, the value is *not* ignored if the install config does not specify usage of Timelock.
Ported from #2952.

@bkrieger for SA.
@hsaraogi - one last PR to review? :-)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2953)

### DIFF
--- a/docs/source/configuration/external_timelock_service_configs/timelock_client_config.rst
+++ b/docs/source/configuration/external_timelock_service_configs/timelock_client_config.rst
@@ -77,6 +77,12 @@ Optional parameters:
 Runtime Configuration
 ---------------------
 
+.. danger::
+
+   Although we support live-reloading of the server configuration, AtlasDB needs to know at install time that it
+   should talk to TimeLock. Added the block to a running service using embedded timestamp and lock servers is unsafe,
+   as a rolling restart is likely to cause **SEVERE DATA CORRUPTION**.
+
 .. warning::
 
     Although we support starting up without knowledge of any TimeLock nodes, note that if you are using TimeLock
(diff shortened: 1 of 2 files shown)


bkrieger:
👍

pkoenig10:
@gsheasby Can we also reword the docs [here](https://github.com/palantir/atlasdb/blob/7ac1a624d204d6eeca29b516c97fb6eba1d7cbd5/docs/source/configuration/external_timelock_service_configs/timelock_client_config.rst#runtime-configuration) to make this behavior more clear?

gsheasby:
Added a scary warning - tossing this over to @jeremyk-91 for final review and merge (assuming everything's true and clear).

docs/source/configuration/external_timelock_service_configs/timelock_client_config.rst
```diff
@@ -77,6 +77,12 @@ Optional parameters:
 Runtime Configuration
 ---------------------
 
+.. danger::
+
+   Although we support live-reloading of the server configuration, AtlasDB needs to know at install time that it
+   should talk to TimeLock. Added the block to a running service using embedded timestamp and lock servers is unsafe,
+   as a rolling restart is likely to cause **SEVERE DATA CORRUPTION**.
```
pkoenig10: I think the important bit to mention here is that you must specify timelock config in both install and runtime. The error message you get otherwise is not overly helpful:
> The TimestampManagementService is currently unavailable, because configuration contains zero servers.

jeremyk-91: @pkoenig10 This is true up to and including 0.74.0, but was changed in 0.75.0 onwards to allow you to specify runtime only (at least, that was the intention of #2850).

@gsheasby Do you think it would be worth mentioning that this is true for older versions? (e.g. in an `info` block)

gsheasby: I think the info block would quickly become unnecessary as products move to 0.75.0+, and this part is already quite "blocky".

Issue #2944, opened 2018-01-30
Logging in between delete batches could be improved


jeremyk-91:
Currently if you try to delete say a million cells with a delete batch size of 1, it's possible that you'll see acquisition/release of the backup lock many times, without any other intervening log lines indicating that sweep is making progress. This is not ideal. Logging the number of cells read so far and possibly the number of deletes made at that time would be good. (Logging the cell would be great, but is unsafe.)

jeremyk-91:
Closing out as #2958 was merged, which addresses this.

PR #2958, opened 2018-02-02
CellsSweeper inter-delete-batch logging

jeremyk-91:
**Goals (and why)**:
- Improve visibility into Sweep when deleting wide / highly versioned rows. See #2944 

- Add logging into `CellsSweeper` so we log after deleting every individual batch of cells, as opposed to after the entire row for very wide rows

- Does this cause too much noise in the logs?
- Is this the right place?

**Where should we start reviewing?**: CellsSweeper

**Priority (whenever / two weeks / yesterday)**: two weeks; not that high, but useful for debugging

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2958)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
@@ -82,11 +82,12 @@ public class CellsSweeper {
             Multimap<Cell, Long> cellTsPairsToSweep,
             Collection<Cell> sentinelsToAdd) {
         if (cellTsPairsToSweep.isEmpty()) {
-            log.info("Attempted to sweep a batch of 0 cell+timestamp pairs.");
+            log.info("Attempted to delete 0 cell+timestamp pairs from table {}.",
+                    LoggingArgs.tableRef(tableRef));
             return;
         }
 
-        log.info("Attempting to delete {} stale cell+timestamp pairs from table {}, and add {} sentinels.",
+        log.info("Attempted to delete {} stale cell+timestamp pairs from table {}, and add {} sentinels.",
                 SafeArg.of("numCellTsPairsToDelete", cellTsPairsToSweep.size()),
                 LoggingArgs.tableRef(tableRef),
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -82,9 +82,16 @@ public void sweepCells(
             Multimap<Cell, Long> cellTsPairsToSweep,
             Collection<Cell> sentinelsToAdd) {
         if (cellTsPairsToSweep.isEmpty()) {
+            log.info("Attempted to delete 0 cell+timestamp pairs from table {}.",
+                    LoggingArgs.tableRef(tableRef));
             return;
         }
 
+        log.info("Attempted to delete {} stale cell+timestamp pairs from table {}, and add {} sentinels.",
```
jeremyk-91: oops, should this read `to add`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -82,9 +82,16 @@ public void sweepCells(
             Multimap<Cell, Long> cellTsPairsToSweep,
             Collection<Cell> sentinelsToAdd) {
         if (cellTsPairsToSweep.isEmpty()) {
```
tboam: Quick check on correctness here: this implies if `cellTsPairsToSweep` is empty but `sentinelsToAdd` is not then we don't do anything.  Is this the correct behaviour?  Can this case ever arise?

gsheasby: I don't think this case should ever happen.

jeremyk-91: Confirmed:
- this is only called from `sweepBatch` in `SweepTaskRunner`
- in that method, this is only called if the batch either has size `>= deleteBatchSize` (mid-loop) which is required to be positive, or if it's non-empty (at the end).

PR #2960, opened 2018-02-02
Better Cassandra retrying logic

gmaretic:
**Goals (and why)**:
Improve the retrying logic in Cassandra. This change gives us an alternative, more conservative retrying option that we can live-reload. This first version is meant more as an experimental feature for testing until we settle on a solution that we are happy with. This is also just the bare minimum, I expect to modify it before it gets merged in and am open to feedback.

- Add live-reloadable parameter to switch between impls of the exception handler.
- Make `TimedOutException` indicative of Cassandra load.

No real concerns, but am open to suggestions on how to handle retrying, backoff etc.

Probably `CassandraRequestExceptionHandler.java`

Not super urgent

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2960)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/AbstractRequestExceptionHandler.java
```diff
@@ -0,0 +1,203 @@
...
+                || ex instanceof CassandraClientFactory.ClientCreationFailedException
+                || isConnectionException(ex.getCause()));
+    }
+
+    private static boolean isTransientException(Throwable ex) {
+        return ex != null
+                // There's a problem with the connection to Cassandra.
+                && (ex instanceof TTransportException
+                // Not enough Cassandra nodes are up.
+                || ex instanceof InsufficientConsistencyException
```
fsamuel-bs: I think we want to have the same behavior for `InsufficientConsistencyException` and `UnavailableException`, since we try to wrap the latter in the former in several places.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ConservativeRequestExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    boolean shouldBackoff(Exception ex) {
+        return !isFastFailoverException(ex);
+    }
+
+    @Override
+    long getBackoffPeriod(int numberOfAttempts) {
+        return 500 * (long) Math.pow(2, numberOfAttempts);
```
fsamuel-bs: I don't know if we want longer backoffs, but I'm not sure of the behavior we want here. @tboam, what do you think?

tboam: Backing off by a longer amount each time seems reasonable but we should have a maximum as well to prevent threads sitting idle for 20 minutes.  I'd rather fail if we can't get C* to execute our query than wait indefinitely.

tboam:
Given we're experimenting with 2 implementations and we intend to throw the bad one away I'd be happy with 2 similar classes which share no code.  This makes it easier to see similarities/differences and saves work on refactoring.  Does the inheritance make sense if there is only one subclass?

I'll chat to @aggie05engineer about how we want to test this.  I'd rather we started experimenting/rolling out/refining our exception handling than produced something complicated in a vacuum so I'd bias towards to getting these 2 implementations into the field and then improving them.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ConservativeRequestExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    ConservativeRequestExceptionHandler(
+            Supplier<Integer> maxTriesSameHost,
+            Supplier<Integer> maxTriesTotal,
+            Blacklist blacklist) {
+        super(maxTriesSameHost, maxTriesTotal, blacklist);
+    }
+
+    @Override
+    boolean shouldBackoff(Exception ex) {
+        return !isFastFailoverException(ex);
```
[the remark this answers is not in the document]
gmaretic: This was done intentionally, to highlight that we backoff on all exceptions except fast failover.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/AbstractRequestExceptionHandler.java
gmaretic: Agree, also we want to backoff when we don't have enough nodes up

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ConservativeRequestExceptionHandler.java
gmaretic: Added upper bound of 30 seconds. We can consider making this configurable or changing to a different bound.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
```diff
@@ -29,100 +31,168 @@
...
-    private CassandraRequestExceptionHandler exceptionHandler = new CassandraRequestExceptionHandler(
-            () -> MAX_RETRIES_PER_HOST,
-            () -> MAX_RETRIES_TOTAL,
-            new Blacklist(config));
+    @Parameterized.Parameters
+    public static Iterable<? extends Object> data() {
+        return Arrays.asList(false, true);
+    }
+
+    @Parameterized.Parameter
```
tboam: I think this test would be easier to read and maintain if we just copied all the tests for the different handlers.  And when we delete one later it is easier to remove too.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -116,18 +126,19 @@
         }
     }
 
+    @VisibleForTesting
+    boolean shouldBlacklist(Exception ex, int numberOfAttempts) {
```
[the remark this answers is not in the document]
gmaretic: Will add a todo to track

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
gmaretic: done

PR #2961, opened 2018-02-02
Split ExecutorInheritableThreadLocal out of commons-executors

twilson-palantir:
The introduction of remoting3:tracing transitively added guava, jackson,
etc. to the dependency footprint of commons-executors. This is painful
for api projects that don't want these dependencies.

ExecutorInheritableThreadLocal is useful for api projects to declare
shared thread locals, so it's useful for it to be split out into a
separate project with no dependencies.

**Goals (and why)**:

Make it possible to use ExecutorInheritableThreadLocal in apis without pulling in extra dependencies.

Move ExecutorInheritableThreadLocal into another project and have commons-executors depend on it.

None

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2961)


twilson-palantir:
I'll add a release note. A comment in build.gradle should hopefully suffice for now to discourage adding dependencies.

PR #2962, opened 2018-02-05
[no release notes] Fixed broken test infra, added a test

gmaretic:
**Goals (and why)**:
Overrides for ConflictHandlers were not being propagated to transactions created by `runTaskWithRetry`. Tests were assuming they were being propagated.
 
Overriding `createTransaction` in `TestTransactionManagerImpl.java`

It's not the cleanest since it duplicates `SerializableTransactionManager` code, but didn't seem valuable to spend too much time on this. At least fixes the existing issue.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2962)

### DIFF
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
@@ -17,7 +17,9 @@ package com.palantir.atlasdb.transaction.impl;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
+import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cleaner.NoOpCleaner;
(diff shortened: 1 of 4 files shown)


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
```diff
@@ -108,10 +112,37 @@ public Transaction commitAndStartNewTransaction(Transaction tx) {
...
+                allowHiddenTableAccess,
+                timestampValidationReadCache,
+                lockAcquireTimeoutMs.get(),
+                getRangesExecutor,
+                defaultGetRangesConcurrency,
+                sweepQueueWriter);
+    }
+
+    @Override
+    public SnapshotTransaction createNewTransaction() {
```
jeremyk-91: Not sure I understand why we changed the return type of this one from `Transaction` to `SnapshotTransaction` - do we actually use the properties of a snapshot-transaction in the test?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
```diff
@@ -131,4 +162,11 @@ public Transaction createNewTransaction() {
     public void overrideConflictHandlerForTable(TableReference table, ConflictHandler conflictHandler) {
         conflictHandlerOverrides.put(table, conflictHandler);
     }
+
+    Map<TableReference, ConflictHandler> getConflictHandlerWithOverrides() {
```
jeremyk-91: nit: should be private

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1038,6 +1004,31 @@ private HeldLocksToken getExpiredHeldLocksToken() {
...
+                return null;
+            });
+        }
+        for (int i = 0; i < 1000; i++) {
+            Future<Void> future = executor.take();
+            future.get();
+        }
+        t1 = txManager.createNewTransaction();
+        return EncodingUtils.decodeVarLong(t1.get(TABLE, ImmutableSet.of(cell)).values().iterator().next());
+    }
```
jeremyk-91: **NON-ACTIONABLE**: The above could do with a significant cleanup, but given that it was just extracted from the bunch of duplicated methods and lives in test code I'm okay with living with this for now.

PR #2964, opened 2018-02-06
Wrap closeables and close them on failure

fsamuel-bs:
**Goals (and why)**: We've seen instances of we leaking resources when `TransactionManagers.serializable` fails. This attempts to stop this resource leakage by fixing the lifecycle of closeable resources.

- Make objects with `shutdown` AutoCloseable.
- Fix lifecycle of QoS objects by wiring `close` calls.
- Wrap closeables on `serializable`.

**Concerns (what feedback would you like?)**: Lacks tests. Will add a unit test with mocks and try to reproduce the issue on IL.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2964)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -205,15 +209,37 @@ public static SerializableTransactionManager createInMemory(Set<Schema> schemas)
     @JsonIgnore
     @Value.Derived
     public SerializableTransactionManager serializable() {
+        List<AutoCloseable> closeables = Lists.newArrayList();
```
sandorw: Could also do try-with-resources with a list that implements autocloseable and cut down on the code here, but probably isn't worth it.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -205,15 +209,37 @@ public static SerializableTransactionManager createInMemory(Set<Schema> schemas)
...
         AtlasDbRuntimeConfig defaultRuntime = AtlasDbRuntimeConfig.defaultRuntimeConfig();
-        java.util.function.Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier =
+        Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier =
                 () -> runtimeConfigSupplier().get().orElse(defaultRuntime);
 
-        QosClient qosClient = getQosClient(JavaSuppliers.compose(AtlasDbRuntimeConfig::qos, runtimeConfigSupplier));
+        QosClient qosClient = initializeCloseable(
+                () -> getQosClient(JavaSuppliers.compose(AtlasDbRuntimeConfig::qos, runtimeConfigSupplier)), closeables
+        );
+        closeables.add(qosClient);
```
sandorw: Didn't `initializeCloseable` already add the client to the list?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
sandorw: Nevermind, since we only want to clean them up in case of an exception.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -205,15 +209,36 @@ public static SerializableTransactionManager createInMemory(Set<Schema> schemas)
...
+        try {
+            return serializableInternal(closeables);
+        } catch (Throwable throwable) {
+            log.warn("Exception thrown when creating transaction manager. "
+                    + "Closing previously opened resources", throwable);
+            closeables.forEach(autoCloseable -> {
+                try {
+                    autoCloseable.close();
+                } catch (Exception ex) {
+                    log.info("Error closing {}", SafeArg.of("klass", autoCloseable.getClass()), ex);
```
jeremyk-91: nit: Probably can get away with `class` in the SafeArg, as it's a string

fsamuel-bs:
Smoke test passes:
1. EteServer correctly closes QoS resources when starts without Cassandra.
2. EteServer closes all closeable resources when a RuntimeException is thrown at the end of `TMs.serializable()`.

Merging on the next green build.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -237,81 +261,99 @@ public SerializableTransactionManager serializable() {
...
-
-        SchemaMetadataService schemaMetadataService = SchemaMetadataServiceImpl.create(rawKvs,
+        KeyValueService keyValueService = initializeCloseable(() -> {
+            KeyValueService kvs = atlasFactory.getKeyValueService();
+            kvs = ProfilingKeyValueService.create(kvs);
+            kvs = SweepStatsKeyValueService.create(kvs,
+                    new TimelockTimestampServiceAdapter(lockAndTimestampServices.timelock()));
+            kvs = TracingKeyValueService.create(kvs);
+            kvs = AtlasDbMetrics.instrument(KeyValueService.class, kvs, MetricRegistry.name(KeyValueService.class));
+            return ValidatingQueryRewritingKeyValueService.create(kvs);
```
fsamuel-bs: Note that if one of the KVS objects creation fail midway through, we won't clean the initially created objects. We could work around this, but currently none of the created KVSs can fail after the first one, so... it should be fine.

jeremyk-91: You can always fail with `OutOfMemoryError` or a messed-up classpath leading to linkage errors... 😛

I think this covers the majority of cases though, we can protect against these weird cases if this resurfaces in one of these directions in prod

qos-service-impl/src/main/java/com/palantir/atlasdb/qos/client/AtlasDbQosClient.java
```diff
@@ -98,4 +98,12 @@ public static AtlasDbQosClient create(QosRateLimiters rateLimiters) {
...
     }
 
+    @Override
+    public void close() {
+        try {
+            rateLimiters.read().close();
+        } finally {
+            rateLimiters.write().close();
+        }
+    }
```
jeremyk-91: This is clever, I didn't think of this 😄 

Might not scale if there were more resources, but since there were just two...

fsamuel-bs: It messes up the exceptions — if an exception is thrown in the first block and in the finally block, the finally one just shadows the first one. For a correct behavior we'd need to add the first one to the suppressed exceptions of the second one... but I think it should be fine.

PR #2966, opened 2018-02-06
Instrument TransactionManager

fsamuel-bs:
**Goals (and why)**: Instrument TransactionManager. Note that there are some AtlasDB consumers that are instrumenting by themselves. They should remove these metrics as of the version that's shipped with this change, to avoid having two similar metrics.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2966)


fsamuel-bs:
Should add tracing to the instrumentation as well.

sandorw:
Do we want to announce it as a dev break to call attention to it?

fsamuel-bs:
The base PR has release notes with the `devbreak` tag

PR #2967, opened 2018-02-06
Record number of cells written per table

fsamuel-bs:
**Goals (and why)**: Record number of cells written per table.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2967)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -0,0 +1,73 @@
...
+        this.taggedMetricRegistry = taggedMetricRegistry;
+    }
+
+    @Override
+    public CassandraClient delegate() {
+        return delegate;
+    }
+
+    @Override
+    public void batch_mutate(String kvsMethodName,
```
tboam: We don't appear to call the delegate in here which means the writes won't actually happen!

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -0,0 +1,73 @@
...
+            Map<ByteBuffer, Map<String, List<Mutation>>> mutation_map,
+            ConsistencyLevel consistency_level)
+            throws InvalidRequestException, UnavailableException, TimedOutException, TException {
+        mutation_map.forEach((key, tableToMutations) -> {
```
tboam: I think we can just loop over the values in `mutation_map`.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -0,0 +1,73 @@
...
+            tableToMutations.forEach((table, mutations) -> {
+                getCellsWrittenMeterForTable(table).mark(mutations.size());
+            });
+        });
+    }
+
+    private Meter getCellsWrittenMeterForTable(String table) {
+        return taggedMetricRegistry.meter(
```
tboam: As a Meter the intention is we sum across all hosts and accumulate over time to give us an estimate of accumulative size?  Which we can compare to accumulative cells swept, right?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
tboam: Should the calculation happen before or after the call to Cassandra?  Should it only happen if the write is successful?

fsamuel-bs: I think after the write is better, since QoS will act before the write.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
fsamuel-bs: Yep. We can `sum(host)` in the internal analytics tool.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -0,0 +1,81 @@
...
+            ConsistencyLevel consistency_level)
+            throws InvalidRequestException, UnavailableException, TimedOutException, TException {
+        delegate.batch_mutate(kvsMethodName, mutation_map, consistency_level);
+
+        Map<String, Long> tablesToCells = new HashMap<>(mutation_map.size());
+
+        mutation_map.values().forEach(tableToCellsMap -> {
+            tableToCellsMap.forEach((table, cells) -> {
+                Long numberOfCells = tablesToCells.getOrDefault(table, 0L);
+                tablesToCells.put(table, numberOfCells + cells.size());
```
fsamuel-bs: There's some boxing going on, but could not avoid it. :/

tboam: I'm not concerned about the performance of boxing here.

PR #2968, opened 2018-02-07
Delete cell at its timestamp plus one

tpetracca:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2968)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellDeleter.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellDeleter.java
@@ -88,7 +88,7 @@ class CellDeleter {
                             pred.setColumn_names(Arrays.asList(colName));
                             Deletion del = new Deletion();
                             del.setPredicate(pred);
-                            del.setTimestamp(Long.MAX_VALUE);
+                            del.setTimestamp(ts + 1);
                             Mutation mutation = new Mutation();
                             mutation.setDeletion(del);
 
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,6 +50,11 @@ develop
     *    - Type
          - Change
 
+    *    - |fixed|
+         - Fixed a bug which would make sweep deletes not be compacted by Cassandra.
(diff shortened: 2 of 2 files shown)


tpetracca:
Yea, sorry clocks. A lot of us were in the same room when discussing this. I'm going to ask @ashrayjain to do a brief write-up internally about why we want this; the BLUF is tombstones with a timestamp of Long.MAX_VALUE are bad.

To your concerns: Recently for some of the other work we've done I dug deep into the tombstone/TTL/gc_grace stuff, and this won't affect any of that. When you write a tomstone `USING TIMESTAMP X` the X is only considered when determining what cell wins when we have cells with conflicting primary keys across multiple sstables. The gc_grace determination is made off a different value which is just the wall clock time of the node at the time the delete was issued.

j-baker:
While we're here (and unrelated), clocks - what do we gain from writing at Atlas clock given that the same clock value is in the column so it will never overwrite any other?

tboam:
Assigning to @ashrayjain for a write up on why MAX_LONG is bad and why this is better.

ashrayjain:
Clocks already knows this but tombstone purging code in Cassandra works at a partition level and not at the cell level. Additionally, assuming the gc_grace constraint is already satisfied, the criteria that C* uses to decide when to drop tombstones during a compaction involves checking the timestamp of the tombstone (long.max for atlas) against the minimum write timestamp across all the data (this includes any data in the memtables for that partition key as well as all other sstables that contain that key, that are not involved in the compaction). A tombstone is dropped if the minimum timestamp found above is higher than the timestamp of the tombstone.

In case of Atlas, since we use Long.max, this essentially never ends up happening. Even more annoyingly, major compactions can also prove to be ineffective in clearing tombstones since the above logic includes stuff sitting in memtables as well.

By setting the timestamp to (the timestamp of the thing we are deleting + 1), we ensure that our delete is correctly applied on top of the value being deleting, without forcing C* to keep the tombstone around for long periods of time.

clockfort:
Understood, thanks for the summary.
I checked the localDeletionTime related code, and this lgtm.

tboam:
@ashrayjain @tpetracca is there any work needed for all of the values already written?  Are we ok to leave them as is?

PR #2969, opened 2018-02-07
Test NoOpPersistentLockService + PersistentLockManager interaction

fsamuel-bs:
**Goals (and why)**: The large internal product will use `NoOpPersistentLockService`, since they don't require backup locks to be taken for their backup procedure. Just wanted to future proof this interaction with this unit test.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2969)

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/PersistentLockManagerTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/PersistentLockManagerTest.java
@@ -20,6 +20,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/PersistentLockManagerTest.java
```diff
@@ -249,6 +251,14 @@ public void doesNotDeadlockOnShutdownIfLockCannotBeAcquired() throws Interrupted
         manager.shutdown();
     }
 
+    @Test
+    public void testNoOpPersistentLockDoesNotThrow() {
+        PersistentLockManager noOpManager = new PersistentLockManager(new NoOpPersistentLockService(), 0L);
+        assertTrue("NoOpPersistentLockService should return true when acquiring lock",
+                noOpManager.tryAcquirePersistentLock());
```
gsheasby: How about another test for `tryAcquire` twice without a `release`? This should work for `noOp`.

PR #2971, opened 2018-02-07
Be nice to Cassandra client's input buffers

gbonik:
Calling clear() on TMemoryInputTransport sets its 'buf' field to null
which can later result in a NullPointerException in places like
TBinaryProtocol.readI32().

Instead, we can call reset() with an empty byte array to achieve
a similar effect without using null pointers. For safety, we also do it
only if there are no bytes left in the buffer to avoid effectively
modifying its state. Not sure why there would be extra bytes left in the
buffer after a successful request, but the evidence suggests that it
might happen occasionally.

----

See QA-111187. Basically, there is a test flake in our internal product:

```
Caused by: java.lang.RuntimeException: java.lang.NullPointerException: null
--
at org.apache.thrift.protocol.TBinaryProtocol.readI32(TBinaryProtocol.java:320)
at org.apache.thrift.protocol.TBinaryProtocol.readMessageBegin(TBinaryProtocol.java:219)
at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:69)
at org.apache.cassandra.thrift.Cassandra$Client.recv_describe_keyspace(Cassandra.java:1413)
at org.apache.cassandra.thrift.Cassandra$Client.describe_keyspace(Cassandra.java:1400)
at com.palantir.atlasdb.keyvalue.cassandra.SchemaMutationLockTables.getAllLockTablesInternal(SchemaMutationLockTables.java:58)
 
```
Carefully reading the code points to `eagerlyCleanupReadBuffersFromIdleConnection()` being too eager

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2971)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -39,6 +39,7 @@ import com.codahale.metrics.Gauge;
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
 import com.palantir.atlasdb.cassandra.CassandraKeyValueServiceConfig;
+import com.palantir.atlasdb.encoding.PtBytes;
 import com.palantir.atlasdb.qos.QosClient;
 import com.palantir.atlasdb.util.MetricsManager;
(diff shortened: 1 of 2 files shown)


gsheasby:
@gbonik Thanks for tracking this down! 👍

fsamuel-bs:
I think we had seen this in the field (cc @hsaraogi), but were unable to track it down. Thanks!

PR #2972, opened 2018-02-07
[TM Part 1] Simplify TransactionManager inheritance tree

fsamuel-bs:
**Goals (and why)**:
1. Remove a potential bug in the future by unifying LockAware and ConditionAware — WrappingTM did not wrap methods from ConditionAwareTM.
2. Remove the cognitive overhead of keeping several interfaces, abstract classes and forwarding classes when we have just one impl that we care about.
3. Make it possible to instrument the TransactionManager we return in `TransactionManagers.serializable()`, since now we return an interface.
4. Get rid of the horrible shim in `SerializableTransactionManager`

1. Merge LockAwareTM and ConditionAwareTM in TM.
2. Remove ForwardingTM, replace it with AutoDelegate generated classes.

**Concerns (what feedback would you like?)**: This is a dev-break, but I think it's fine.

**Where should we start reviewing?**: TransactionManager

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2972)


fsamuel-bs:
@dxiao for SA. This will break the large internal product. I'll follow up with fixes after this merges.

fsamuel-bs:
Before:
screen shot 2018-02-08 at 11 42 50 am

After:
screen shot 2018-02-08 at 11 43 32 am

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
```diff
@@ -50,7 +50,8 @@
 import com.palantir.lock.v2.TimelockService;
 import com.palantir.timestamp.TimestampService;
 
-/* package */ class SnapshotTransactionManager extends AbstractLockAwareTransactionManager {
+class SnapshotTransactionManagerImpl extends AbstractTransactionManager implements
```
sandorw: We should remove this and just have `SerializableTransactionManagerImpl`, but keep the separation of `SerializableTransaction` vs. `SnapshotTransaction` for now.

It's pretty weird to have SerializableImpl extends SnapshotImpl implements Serializable

sandorw:
Closing this, implementing something similar in a new PR.

PR #2974, opened 2018-02-08
Make cells written metric a counter

fsamuel-bs:
**Goals (and why)**: Make the number of cells written metric a counter. We can do all the analysis we want in our metric tool with a counter, and this will reduce the number of tagged metrics we use.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2974)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
@@ -20,6 +20,7 @@ import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.cassandra.thrift.ConsistencyLevel;
 import org.apache.cassandra.thrift.InvalidRequestException;
@@ -28,7 +29,7 @@ import org.apache.cassandra.thrift.TimedOutException;
 import org.apache.cassandra.thrift.UnavailableException;
 import org.apache.thrift.TException;
 
-import com.codahale.metrics.Meter;
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.atlasdb.logging.LoggingArgs;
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -68,14 +67,14 @@ public void batch_mutate(String kvsMethodName,
...
 
-        tablesToCells.forEach((table, numberOfCells) -> getCellsWrittenMeterForTable(table).mark(numberOfCells));
+        tablesToCells.forEach((table, numberOfCells) -> updateCellsWrittenMeterForTable(table, numberOfCells));
     }
 
-    private Meter getCellsWrittenMeterForTable(String table) {
-        return taggedMetricRegistry.meter(
-                MetricName.builder()
+    private void updateCellsWrittenMeterForTable(String table, Long numberOfCells) {
+        taggedMetricRegistry.counter(MetricName.builder()
```
tboam: I do wonder if we should cache the table->counter mapping to save on constructing the MetricName every time.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
```diff
@@ -68,14 +67,14 @@ public void batch_mutate(String kvsMethodName,
...
         });
 
-        tablesToCells.forEach((table, numberOfCells) -> getCellsWrittenMeterForTable(table).mark(numberOfCells));
+        tablesToCells.forEach((table, numberOfCells) -> updateCellsWrittenMeterForTable(table, numberOfCells));
     }
 
-    private Meter getCellsWrittenMeterForTable(String table) {
-        return taggedMetricRegistry.meter(
-                MetricName.builder()
+    private void updateCellsWrittenMeterForTable(String table, Long numberOfCells) {
```
tboam: The method name mentions a Meter which is now slightly misleading.  `updateCellsWrittenForTable ` describes the intent without prescribing a certain implementation.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/InstrumentedCassandraClient.java
fsamuel-bs: That's what TaggedMetricRegistry does.

tboam: I meant avoiding having to do the `MetricName.builder()...build()` all the time and then having to has all the parts vs just a straight lookup.  But I agree the perf difference is probably negligible.

PR #2970, opened 2018-02-07
SSSS Part 2: Metadata Cleanup Task

jeremyk-91:
**Goals (and why)**:
- Implement a stream store's `MetadataCleanupTask` in a generic fashion, so that Nimbus can sweep a service with a stream store.

**Implementation Description (bullets)**:
- Implement an `OnCleanupTask` which is a generic form of the task used to clean a stream store's metadata table

**Concerns (what feedback would you like?)**:
- Is this a faithful translation of the generated version?

**Where should we start reviewing?**:`GenericStreamStoreMetadataCleanupTask`

**Priority (whenever / two weeks / yesterday)**: next week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2970)

PR #2976, opened 2018-02-08
SSSS Part 3: Index Cleanup Tasks and Deletion Filters

jeremyk-91:
**Goals (and why)**:
- Provide a generic implementation of a stream store's index cleanup task, so that Nimbus can sweep stream stores.

- Refactor the metadata delete task from #2970 to be a generic task taking a deletion filter, since the only part that differs between the index delete and this one is how we screen out stream IDs that we actually still want to keep
- Implement the index delete using a deletion filter (`UnindexedStreamDeletionFilter`)

- I take a defensive copy at the end of `UnindexedStreamDeletionFilter#getStreamIdentifiersToDelete` - is this needed?
- The unindexed stream deletion filter does a dynamic column range scan while grabbing all columns into memory at one shot. This is probably safe, right?

**Where should we start reviewing?**: Probably `GenericStreamStoreCleanupTask`.

**Priority (whenever / two weeks / yesterday)**: next week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2976)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/GenericStreamDeletionFilter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/external/GenericStreamDeletionFilter.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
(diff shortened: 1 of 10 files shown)


jeremyk-91:
closing for now - we can reopen later if needed

PR #2978, opened 2018-02-08
Check cause of exception when putting resource back in CCPC

### LINKED ISSUE #2973: Reusing Cassandra clients after an exception has been thrown can lead to errors
CassandraClientPoolingContainer.runWithGoodResource() returns a client to the pool, unless an exception of a specific type was caught (basically, TTransportException or TProtocolException). The problem is that those exceptions can sometimes be wrapped in a RuntimeException or similar, thus bypassing the "blacklist". This might result in broken connections being returned to the pool, and manifest as test flakes like QA-111187.

See also https://github.com/palantir/atlasdb/pull/2971 . In that pull request, I attempted to address the issue where a client was in a bad state right after checkout from the pool. While that PR should fix an immediate problem (NPE), further discussion with @clockfort revealed that the problem might be deeper. It seems like sometimes we get a connection from the pool that already has unread data in its buffer from a previous query. The only way this can happen in Thrift seems to be when it throws an exception in the middle of a call.

For example, see CassKVSImpl.createTablesInternal():
```
                } catch (TException thriftException) {
                    if (thriftException.getMessage() != null
                            && !thriftException.getMessage().contains("already existing table")) {
                        throw QosAwareThrowables.unwrapAndThrowRateLimitExceededOrAtlasDbDependencyException(
                                thriftException);
                    }
                }
```
This code re-throws a thrift exception as an unchecked exception. Therefore, CassandraClientPoolingContainer still returns it to the pool, and the next call that checks out the same connection from the pool will potentially get a broken connection.

fsamuel-bs:
**Goals (and why)**: Fix #2973.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2978)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -179,10 +179,11 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
         }
     }
 
-    private static boolean isInvalidClientConnection(Exception ex) {
+    private static boolean isInvalidClientConnection(Throwable ex) {
         return ex instanceof TTransportException
                 || ex instanceof TProtocolException
-                || ex instanceof NoSuchElementException;
+                || ex instanceof NoSuchElementException
+                || (ex.getCause() != null && isInvalidClientConnection(ex.getCause()));
     }
 
     private void invalidateQuietly(CassandraClient resource) {
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -60,6 +60,10 @@ develop
            (`Pull Request <https://github.com/palantir/atlasdb/pull/2948>`__)
 
     *    - |fixed|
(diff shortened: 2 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -179,10 +179,11 @@ private static void eagerlyCleanupReadBuffersFromIdleConnection(CassandraClient
         }
     }
 
-    private static boolean isInvalidClientConnection(Exception ex) {
+    private static boolean isInvalidClientConnection(Throwable ex) {
```
gbonik: This is still fragile. What if someone catches an exception and doesn't rethrow? Or throws without rewrapping? Or what if the exception ended up in suppressed exceptions because of a try-with-resources block? Please see my first comment on the issue for a proposal of a more robust fix. Essentially, we want to catch as close to the thrift call as possible.

tboam: Agree a more robust fix is better and should be implemented.  But is there value in shipping this now to fix most of the problems while we implement the more difficult fix?

fsamuel-bs:
Closing, as this was addressed by #3063.

PR #2981, opened 2018-02-09
Puncher no longer uses range scans

j-baker:
One of the things we'd like to do is to have some usages of AtlasDB use
the Murmur partitioner in Cassandra. A consequence of this is that
traditional AtlasDB range scans will be a lot harder, and that we should
be using dynamic columns for this.

This PR moves to a scheme for the puncher where the same row is used
for some period of time (14 days as currently configured), and we range
scan within this row.

If the request is not within the current row, we look back one, and then
30, giving up to 420 days worth of lookback (which seems like enough).

Additionally, I removed the 'get wall clock time for timestamp' method,
leaving it in the KVS implementation only. Right now it's only used in
a cli, so I consider it dead code.

**Goals (and why)**:

Does the code make sense? Does the idea make sense? I understand we can't do range scans in the same way when using Murmur3.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2981)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/PuncherStore.java
```diff
@@ -44,9 +44,4 @@ default boolean isInitialized() {
      * Find the latest timestamp created at or before timeMillis.
      */
     Long get(Long timeMillis);
-
-    /**
-     * Get the time in millis for the greatest timestamp punched less than or equal to the given timestamp.
-     */
-    long getMillisForTimestamp(long timestamp);
```
sandorw: We've got at least one internal use of this that we'll have to resolve if we want to remove it.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -105,59 +134,83 @@ public boolean isInitialized() {
...
-            if (result.hasNext()) {
-                byte[] encodedMillis = result.next().getRowName();
-                EncodingUtils.flipAllBitsInPlace(encodedMillis);
-                return EncodingUtils.decodeUnsignedVarLong(encodedMillis);
-            } else {
-                return 0L;
-            }
-        } finally {
-            result.close();
+        RangeRequest rangeRequest = RangeRequest.builder().batchHint(1).build();
```
sandorw: batch size 1 seems to be implied by the KVS API, but this shouldn't hurt

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStoreTest.java
```diff
@@ -113,9 +115,13 @@ private static long mean(long first, long second) {
...
+    private void initializePuncherStore(Map<Long, Long> timestampMap) {
+        keyValueService = new InMemoryKeyValueService(false);
+        puncherStore = KeyValueServicePuncherStore.create(keyValueService);
         timestampMap.forEach((key, value) -> puncherStore.put(key, value));
-        return puncherStore;
+    }
+
+    private long getMillisForTimestamp(long timestamp) {
+        return KeyValueServicePuncherStore.getMillisForTimestamp(keyValueService, timestamp);
     }
```
jeremyk-91: Probably would be good to have one test for the case where you have to go across a two week boundary to get the right answer (i.e. your timestamp is in an interval between a two week boundary and the first punch that occurred after said boundary - you should return the last one in the previous period).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -105,59 +134,83 @@ public boolean isInitialized() {
...
+    private static long timestamp(Cell cell) {
+        return MILLIS_IN_TWO_WEEKS * EncodingUtils.decodeLittleEndian(cell.getRowName(), 0)
+                + EncodingUtils.decodeUnsignedVarLong(EncodingUtils.flipAllBits(cell.getColumnName()));
+    }
+
+    private static byte[] row(long timeMillis) {
+        return EncodingUtils.encodeLittleEndian(Math.floorDiv(timeMillis, MILLIS_IN_TWO_WEEKS));
+    }
+
+    // floorDiv and floorMod are to ensure that this works properly if the time is somehow negative.
```
jeremyk-91: nit: this should go before `row()` probably

Nice though, I probably wouldn't have thought of this!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -43,7 +52,24 @@
...
+ * In order to look up a value for a give time, truncate in the same way and range scan within the row
+ * for the first cell beneath the provided time.
+ *
+ * Selecting the wall clock timestamp for a given Atlas timestamp still requires a full table scan,
+ * but should not be used in production code.
+ *
+ * It is expected that each row contains (with default setting) (1440 * 14 = 20k cells).
+ *
+ * Note one caveat of this implementation - as written, the get method only works if it is in the last
+ * 420 days, and defaults to Long.MIN_VALUE otherwise.
```
jeremyk-91: I'm a bit torn about this, but I can see why it could be beneficial, and the scenario should quickly correct itself (i.e. the puncher should punch something, which should mean that in an hour or so in the worst case future gets should work). 

The consequences of giving a pessimistic answer to get is that sweep will stop, but it's unlikely and as mentioned self-correcting.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -88,12 +114,15 @@ private void tryInitialize() {
         keyValueService.createTable(AtlasDbConstants.PUNCH_TABLE, new TableMetadata(
                 NameMetadataDescription.create(ImmutableList.of(
                         new NameComponentDescription.Builder()
-                                .componentName("time")
+                                .componentName("time_rounded_to_day")
```
jeremyk-91: This looks like an artifact of when the resolution was previously set to 1 day.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -105,59 +134,83 @@ public boolean isInitialized() {
...
+                .map(Map.Entry::getValue)
+                .map(KeyValueServicePuncherStore::getAtlasTimestamp)
+                .findFirst();
+    }
+
+    private static long getAtlasTimestamp(Value value) {
+        return EncodingUtils.decodeUnsignedVarLong(value.getContents());
     }
 
+    // Relies on a full table scan, but should only be used in a cli.
```
jeremyk-91: (in OSS Atlas at least, this is indeed only used by the CLIs.)

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
```diff
@@ -30,7 +30,7 @@ private AtlasDbConstants() {
         // Utility
     }
 
-    public static final TableReference PUNCH_TABLE = TableReference.createWithEmptyNamespace("_punch");
+    public static final TableReference PUNCH_TABLE = TableReference.createWithEmptyNamespace("_punch2");
```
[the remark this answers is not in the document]
jeremyk-91: Yep. The consequence of this would be that we would try to sweep `_punch`, which isn't very useful. (In previous versions this would cause sweep to explode - because we punch at commit timestamps, which don't have a transactions table entry - though we've made it more resilient since!)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -43,7 +52,24 @@
 /**
  * A PuncherStore implemented as a table in the KeyValueService.
  *
- * @author jweel
+ * This utilises dynamic columns to remove the necessity of range scans.
+ *
+ * The schema is:
+ *
+ * row: time rounded to 14 days
```
jeremyk-91: nit: `truncated`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -105,59 +134,83 @@ public boolean isInitialized() {
...
-            result.close();
+        RangeRequest rangeRequest = RangeRequest.builder().batchHint(1).build();
+        try (ClosableIterator<RowResult<Value>> result =
+                kvs.getRange(AtlasDbConstants.PUNCH_TABLE, rangeRequest, timestampExclusive)) {
+            // With Guava 21.0, can do Streams.stream(result), Streams.stream(row.getCells())
+            return StreamSupport.stream(Spliterators.spliteratorUnknownSize(result, 0), false)
+                    .flatMap(row -> StreamSupport.stream(row.getCells().spliterator(), false))
+                    .map(Map.Entry::getKey)
+                    .mapToLong(KeyValueServicePuncherStore::timestamp)
+                    .max()
```
fsamuel-bs: Can we do `max` and then `mapToLong(KeyValueServicePuncherStore::timestamp)`? This seems like is going to do 1 KVS call per atlas timestamp, and then take the max.

Not sure if it's needed, since this is also an approximation of an approximation...

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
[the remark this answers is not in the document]
j-baker: done

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
j-baker: exactly

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
j-baker: fixed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
j-baker: done

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/PuncherStore.java
j-baker: won't remove

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
j-baker: not quite truncated, since we always round down

sandorw:
Closing this out, since we're not tracking Murmur3 usage anymore.

PR #2982, opened 2018-02-12
PoC for Atlas-as-a-service write and read transactions

fsamuel-bs:
**Goals (and why)**: Proof of concept for Atlas-as-a-service getTransactions API. This service needs a `WriteTransaction` and a `ReadOnlyTransaction` to be able to perform transactions manually.

**Implementation Description (bullets)**: Hacked the internals of the transactions to get this done. This heavily relies on what is implemented currently, and the relationship between the transaction methods is not clear in the current impl. This could lead to a change in an area of the codebase creating bugs on the current implemented code.

The above could be said about the whole transactions code though... so maybe this is fine?

What needs to be done:

- [x] Extract the Transactions to separate classes
- [x] **Limit the public API** of the new transaction objects (throw `IllegalArgumentException` on methods we don't expect to be used?)
- [ ] Tests

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2982)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceReadOnlyTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceReadOnlyTransaction.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.impl;
+
+import com.palantir.atlasdb.transaction.api.AutoDelegate_Transaction;
+import com.palantir.atlasdb.transaction.api.Transaction;
+import com.palantir.processors.AutoDelegate;
+
+@AutoDelegate(typeToExtend = Transaction.class)
+class ServiceReadOnlyTransaction implements AutoDelegate_Transaction {
+    private final SnapshotTransaction delegate;
+
+    ServiceReadOnlyTransaction(SnapshotTransaction delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Transaction delegate() {
+        return delegate;
+    }
+}
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceWriteTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceWriteTransaction.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.impl;
+
+import com.palantir.atlasdb.transaction.api.AutoDelegate_Transaction;
+import com.palantir.atlasdb.transaction.api.Transaction;
+import com.palantir.atlasdb.transaction.api.TransactionFailedException;
+import com.palantir.atlasdb.transaction.service.TransactionService;
+import com.palantir.processors.AutoDelegate;
+
+@AutoDelegate(typeToExtend = Transaction.class)
+abstract class ServiceWriteTransaction implements AutoDelegate_Transaction {
+    private final SnapshotTransaction delegate;
+
+    ServiceWriteTransaction(SnapshotTransaction delegate) {
+        this.delegate = delegate;
+    }
+
+    public SnapshotTransaction delegate() {
+        return delegate;
+    }
+
+    public abstract void commit();
+}
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
@@ -37,6 +37,7 @@ import com.palantir.atlasdb.transaction.api.AtlasDbConstraintCheckingMode;
 import com.palantir.atlasdb.transaction.api.ConditionAwareTransactionTask;
 import com.palantir.atlasdb.transaction.api.KeyValueServiceStatus;
 import com.palantir.atlasdb.transaction.api.PreCommitCondition;
+import com.palantir.atlasdb.transaction.api.Transaction;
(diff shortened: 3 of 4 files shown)


fsamuel-bs:
Not sure how we test the new classes. Ideally we'd test that the new endpoints return working transactions... but since the new transactions are just a thin wrapper around the old impls, is this worth it?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
```diff
@@ -114,6 +115,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+        } finally {
+            condition.cleanup();
+        }
+    }
+
+    public <C extends PreCommitCondition, E extends Exception> ServiceReadOnlyTransaction getReadTransaction(
+            C condition) throws E {
+        checkOpen();
+        try {
+            SnapshotTransaction tx = createReadOnlySnapshotTransaction(condition);
```
fsamuel-bs: 1. We throw for THOROUGH tables for readOnly tasks on `ReadTransaction`. Since we create a `SnapshotTransaction`, we this check.

2. As requested we don't grab the immutableTs lock.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
```diff
@@ -114,6 +115,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+    public <C extends PreCommitCondition, E extends Exception> ServiceWriteTransaction getWriteTransaction(C condition)
+            throws E {
+        checkOpen();
+        try {
+            RawTransaction tx = setupRunTaskWithConditionThrowOnConflict(condition);
+            return new ServiceWriteTransaction(tx.delegate()) {
+                @Override
+                public void commit() {
+                    tx.commit();
+                    finishRunTaskWithLockThrowOnConflict(tx, transaction -> null);
```
fsamuel-bs: `finishRunTaskWithLocksThrowOnConflict` unlocks the immutableTs and adds stuff to the cleaner if the transaction is a `HARD_DELETE`, so figured we should probably leave them for correctness.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceWriteTransaction.java
```diff
@@ -0,0 +1,38 @@
...
+package com.palantir.atlasdb.transaction.impl;
+
+import com.palantir.atlasdb.transaction.api.AutoDelegate_Transaction;
+import com.palantir.atlasdb.transaction.api.Transaction;
+import com.palantir.atlasdb.transaction.api.TransactionFailedException;
+import com.palantir.atlasdb.transaction.service.TransactionService;
+import com.palantir.processors.AutoDelegate;
+
+@AutoDelegate(typeToExtend = Transaction.class)
+abstract class ServiceWriteTransaction implements AutoDelegate_Transaction {
```
fsamuel-bs: Do we want to support `HARD_DELETE` and `AGGRESSIVE_HARD_DELETE` for the service? If not, we could throw on some methods, to have a smaller public API.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceReadOnlyTransaction.java
```diff
@@ -0,0 +1,19 @@
+package com.palantir.atlasdb.transaction.impl;
+
+import com.palantir.atlasdb.transaction.api.AutoDelegate_Transaction;
+import com.palantir.atlasdb.transaction.api.Transaction;
+import com.palantir.processors.AutoDelegate;
+
+@AutoDelegate(typeToExtend = Transaction.class)
+class ServiceReadOnlyTransaction implements AutoDelegate_Transaction {
```
fsamuel-bs: Do we want to support `HARD_DELETE` and `AGGRESSIVE_HARD_DELETE` for the service? If not, we could throw on some methods, to have a smaller public API.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceReadOnlyTransaction.java
```diff
@@ -0,0 +1,35 @@
...
+    private final SnapshotTransaction delegate;
+
+    ServiceReadOnlyTransaction(SnapshotTransaction delegate) {
```
sandorw: I assume you chose `SnapshotTransaction` here to avoid buffering all reads from `SerializableTransaction`? 
👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ServiceWriteTransaction.java
```diff
@@ -0,0 +1,38 @@
...
+    private final SnapshotTransaction delegate;
+
+    ServiceWriteTransaction(SnapshotTransaction delegate) {
```
sandorw: This one should be `SerializableTransaction`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
```diff
@@ -114,6 +115,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
         return numTimesFailed > NUM_RETRIES;
     }
 
+    public <C extends PreCommitCondition, E extends Exception> ServiceWriteTransaction getWriteTransaction(C condition)
```
sandorw: I assume non-PoC version will add the methods to the API. This probably needs to be on the Serializable version.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
sandorw: Do we want to trust that this method can't throw? Could use try/finally instead.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
```diff
@@ -114,6 +115,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+            RawTransaction tx = setupRunTaskWithConditionThrowOnConflict(condition);
+            return new ServiceWriteTransaction(tx.delegate()) {
+                @Override
+                public void commit() {
+                    tx.commit();
+                    finishRunTaskWithLockThrowOnConflict(tx, transaction -> null);
+                }
+            };
+        } finally {
+            condition.cleanup();
```
sandorw: The condition needs to be cleaned up post-commit, just like `finishRunTaskWithLockThrowOnConflict`. This will clean it up immediately after creating the transaction.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerImpl.java
sandorw: I guess the contract here ends up being weird... We skip the THOROUGH checks that provide some safety by assuming that you're holding the immutableTs lock, but leave that up to the client. We'll need to be very explicit in docs that anyone using this with THOROUGH sweep must both acquire the lock themselves and have a precommit condition that verifies that the lock is still held.

sandorw: Actually, I think we should just consider removing the immutableTs lock checks on every read to THOROUGH sweep tables. It seems like a reasonable contract that no one should be making use of state from a failed transaction. To even do so, you have to try/catch around the transaction and store the read data in final vars.

sandorw: Will do that in a follow-up PR

fsamuel-bs:
Closing in favor of #2997.

PR #2985, opened 2018-02-12
Upgrade postgres jdbc driver to 42.2.1 (was 9.4.1209)

cakofony:
We're stuck on a postgres jdbc driver build from nearly two years ago. If we discover bugs we can file and fix upstream as opposed to sitting on deprecated builds.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2985)

### DIFF
--- a/atlasdb-cli-distribution/versions.lock
+++ b/atlasdb-cli-distribution/versions.lock
@@ -1107,7 +1107,7 @@
             ]
         },
         "org.postgresql:postgresql": {
-            "locked": "9.4.1209",
+            "locked": "42.2.1",
             "transitive": [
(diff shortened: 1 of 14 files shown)


gbonik:
DXiao and I went through the changelog of the JDBC driver and it seems that we should be fine. The tests in our internal product pass too - Carter tried that. The main concern is possible performance regressions, but we have benchmarks for that - let's merge and see what they show.

cakofony:
Updated the changelog

jeremyk-91:
Ah, we store the versions of dependencies in our version control. I'll fix this

PR #2989, opened 2018-02-14
accept sstable_size_in_mb defaults that we were already accidentally using

clockfort:
Found this as part of a bug that was causing our test stack to do schema mutations on startup and block on downed C* nodes despite not having a real upgrade to perform. (internal issue tracking number is 63705)

We were previously accidentally obliterating the sstable_size_in_mb setting by setting it once properly and another time clearing it out (I'll mark the place in the code we did this with a PR comment below)

This resulted in us having 160MB default sstable sizes instead of 80MB. 

We can either:
- Fix it so that what we thought we were setting as default is the default
  Pro: last time I did testing a billion years ago 80MB was a bit better for our workloads in terms of performance vs. extra IO work balance
  Con: would require anyone who upgraded atlas to have at least one startup where all cassandra nodes are up at once

- Accept the upstream default as our default (which I have done in this PR because I knew it was the safer thing to do, but I'm open to either option)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2989)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -150,7 +153,6 @@ static CfDef getStandardCfDef(String keyspace, String internalTableName) {
         cf.setKey_validation_class("org.apache.cassandra.db.marshal.BytesType");
         cf.setCompaction_strategy_options(new HashMap<String, String>());
```
clockfort: this line was the one obliterating what we already set above

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -160,11 +162,11 @@ static CfDef getStandardCfDef(String keyspace, String internalTableName) {
     @SuppressWarnings("CyclomaticComplexity")
     static boolean isMatchingCf(CfDef clientSide, CfDef clusterSide) {
         String tableName = clientSide.name;
-        if (!Objects.equal(clientSide.compaction_strategy_options, clusterSide.compaction_strategy_options)) {
+        if (clientSide.compaction_strategy != clusterSide.compaction_strategy) {
```
tpetracca: does `CfDef` not implement `equals` correctly? How does this work?

clockfort: compaction_strategy_options was a HashMap that could be null, so you can't just clientHashMap.equals(otherHashMap) without it possible NPEing.
now I'm only checking the string that either says "LCS" or "STCS",
since we weren't setting the options map correctly before / not setting it now

clockfort: oh I see what you're saying also.
Yes, this whole class is because CfDef doesn't implement equals sanely.

Basically the normal upstream CfDef.equals fails on defaults being specified vs unspecified, which is very annoying.

clockfort:
@ashrayjain because we no longer set it by default

clockfort:
added options validation back again, in a way such that the issue we were seeing on our test stack should still pass through correctly (null map vs. empty map)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -216,9 +224,45 @@ static boolean isMatchingCf(CfDef clientSide, CfDef clusterSide) {
     private static void logMismatch(String fieldName, String tableName,
             Object clientSideVersion, Object clusterSideVersion) {
         log.info("Found client/server disagreement on {} for table {}. (client = ({}), server = ({}))",
-                fieldName,
-                tableName,
-                clientSideVersion,
-                clusterSideVersion);
+                SafeArg.of("disagreementType", fieldName),
+                UnsafeArg.of("table", tableName),
```
hsaraogi: `LoggingArgs.tableRef()` but that works only for refs I think, maybe there is something for tablenames or we can log `TableRef.createUnsafe(name)` . This will result in greater visibility.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitionsTest.java
```diff
@@ -61,4 +61,21 @@ public void cfDefWithDifferingGcGraceSecondsValuesShouldNotMatch() {
...
+                TableReference.fromString("test_table"),
+                FOUR_DAYS_IN_SECONDS,
+                AtlasDbConstants.GENERIC_TABLE_METADATA);
+
+        CfDef cf2 = ColumnFamilyDefinitions.getCfDef(
+                "test_keyspace",
+                TableReference.fromString("test_table"),
+                FOUR_DAYS_IN_SECONDS,
+                AtlasDbConstants.GENERIC_TABLE_METADATA);
+
```
hsaraogi: nit: Maybe add a test where `isAppendHeavyAndReadLight` is set for the cfdef.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
hsaraogi: should be add the "sstable_size_in_mb" property to the map here? How will it be set now?

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitionsTest.java
clockfort: that's fair, I'll add one

clockfort:
(talked in person)
we're no longer setting sstable_size_in_mb , since we accidentally weren't before and it didn't kill us, and making the change to having our own custom setting again would be a somewhat difficult upgrade across everyone.

I'll address the remaining two comments, though. Thanks!

PR #2991, opened 2018-02-14
[Oracle Compaction][Part 2] Refactor `kvs.compactInternally`

fsamuel-bs:
**Goals (and why)**: Follow ups of `BackgroundCompactor` PR
1. Compact internally only on `BackgroundCompactor`
2. Remove all Cassandra compactInternally code

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2991)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1919,39 +1903,9 @@ private Column makeColumn(byte[] colName, byte[] contents, long timestamp) {
...
-            log.error("Compaction for {}.{} was interrupted.", UnsafeArg.of("keyspace", keyspace),
-                    LoggingArgs.tableRef(tableRef));
-        } finally {
-            alterGcAndTombstone(
-                    keyspace,
-                    tableRef,
-                    config.gcGraceSeconds(),
-                    CassandraConstants.TOMBSTONE_THRESHOLD_RATIO);
-        }
+        /* no-op */
```
tpetracca: INFO or DEBUG line saying it's a no-op?

tpetracca:
deleting code is the best

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlKeyValueService.java
```diff
@@ -1198,36 +1184,7 @@ String getFullTableName(TableReference tableRef) {
...
-        } catch (InterruptedException e) {
-            log.error("Compaction for {}.{} was interrupted.", config.getKeyspaceOrThrow(), tableRef);
-        } finally {
-            alterTableForCompaction(
-                    tableRef,
-                    config.gcGraceSeconds(),
-                    CassandraConstants.TOMBSTONE_THRESHOLD_RATIO);
-            CqlKeyValueServices.waitForSchemaVersionsToCoalesce("setting up tables post-compaction", this);
-        }
+        /* no-op */
```
gsheasby: should also override `shouldManuallyCompact`.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -700,4 +713,11 @@ default boolean isInitialized() {
     default boolean performanceIsSensitiveToTombstones() {
         return false;
     }
+
+    /**
+     * @return If {@link #compactInternally(TableReference)} should be called to free disk space.
+     */
+    default boolean shouldCompactManually() {
+        return true;
```
gsheasby: Let's default to false here and special case DbKvs instead of CKVS.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -76,9 +79,20 @@ public BackgroundCompactor(TransactionManager transactionManager,
         this.inSafeHours = inSafeHours;
     }
 
-    public synchronized void runInBackground() {
+    public void close() {
```
gsheasby: It might be worth unifying the logic of the BackgroundSweeper and BackgroundCompactor at some point.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -700,4 +712,11 @@ default boolean isInitialized() {
     default boolean performanceIsSensitiveToTombstones() {
         return false;
     }
+
+    /**
+     * @return If {@link #compactInternally(TableReference)} should be called to free disk space.
+     */
+    default boolean shouldManuallyCompact() {
```
jeremyk-91: The naming of this is a bit strange, though I'm not sure I can think of something much better...

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlKeyValueService.java
jeremyk-91: It is by default false, right? That's what we want for Cassandra based KVSes, reading the Javadoc (though the method name is a bit awkward I think)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -49,40 +50,57 @@
...
-    public synchronized void runInBackground() {
+    @Override
+    public void close() {
+        log.info("Closing BackgroundCompactor");
+        daemon.interrupt();
+        try {
+            daemon.join();
+            daemon = null;
+        } catch (InterruptedException e) {
+            throw Throwables.rewrapAndThrowUncheckedException(e);
```
jeremyk-91: Let's reset the interrupted flag before doing this

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -49,40 +50,57 @@
...
     }
 
-    public synchronized void runInBackground() {
+    @Override
+    public void close() {
+        log.info("Closing BackgroundCompactor");
+        daemon.interrupt();
+        try {
+            daemon.join();
+            daemon = null;
```
jeremyk-91: A bit uncertain why we need this assignment. Are we retaining references to the background compactor outside of a TM? Otherwise it looks like the TM after closing should itself be collected, and if that covers all references to this BackgroundCompactor then it should already be garbage-collected.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -49,40 +50,57 @@
...
+        } catch (InterruptedException e) {
+            throw Throwables.rewrapAndThrowUncheckedException(e);
+        }
+    }
+
+    private synchronized void runInBackground() {
```
jeremyk-91: This is only called in the static factory with a separate instance of the class each time, so I'm not sure why it needs to be `synchronized`.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -255,40 +258,46 @@ public void compactInternally() {
...
+                            SafeArg.of("elapsed time", shrinkTimer.elapsed(TimeUnit.MILLISECONDS)));
+                } else {
+                    Stopwatch shrinkAndCompactTimer = Stopwatch.createStarted();
+                    conns.get().executeUnregisteredQuery(
+                            "ALTER TABLE " + oracleTableNameGetter.getInternalShortTableName(conns, tableRef)
+                                    + " SHRINK SPACE COMPACT");
+                    log.info("Call to SHRINK SPACE COMPACT on table {} took {} ms.",
+                            LoggingArgs.tableRef(tableRef),
+                            SafeArg.of("elapsed time", shrinkAndCompactTimer.elapsed(TimeUnit.MILLISECONDS)));
+                }
```
jeremyk-91: (confirming these are the right way round - SHRINK SPACE COMPACT only takes row level locks, while SHRINK SPACE needs to reset the HWM)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -49,40 +50,57 @@
...
 
-    public static void createAndRun(TransactionManager transactionManager,
+    public static Optional<BackgroundCompactor> createAndRun(TransactionManager transactionManager,
             KeyValueService keyValueService,
             LockService lockService,
-            Supplier<CompactorConfig> compactorConfigSupplier) {
+            Supplier<Boolean> inSafeHours) {
+        if (!keyValueService.shouldManuallyCompact()) {
+            return Optional.empty();
+        }
```
jeremyk-91: To be honest for this kind of thing I usually prefer a null object pattern (basically you always create-and-run a background compactor, just that if your KVS shouldn't manually compact then the compactor does nothing) as it's easier to manage.

I realise it's quite a significant break from the existing API, though. Not actionable, to be clear.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
gsheasby: Is it a bit weird - we're saying that Atlas is manually compacting... `shouldCompactAfterSweep`?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlKeyValueService.java
gsheasby: Yep, this comment is from before we decided to switch the default from true to false, so now there's nothing to do :tada:

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
gsheasby: I think a lot of this code was copied from the background sweeper, so may have inherited some foibles.

gsheasby: Yep, this code was copied from `BackgroundSweeperImpl`, and previously was copied from $LARGE_INTERNAL_PROJECT back in Nov 2015. Potential tiny task in a few weeks to clean up some of the duplicated code/logic, but not worth it now.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
gsheasby: Ack and agreed, but will probably defer this refactor, especially as I don't think it would break anyone if we switched it around later.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
gsheasby: well, this method is still called indirectly by TMs, so we should probably be defensive and keep it synchronised.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
jeremyk-91: Discussed offline - I think we have `shouldTriggerCompactions` now which I like

PR #2995, opened 2018-02-15
SnasphotTransactionManager -> SerializableTransactionManager

fsamuel-bs:
**Goals (and why)**: We always want a SerializableTransaction, so no reason for keeping a SnapshotTransactionManager.


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerTest.java
```diff
@@ -45,7 +45,7 @@
     private final Cleaner cleaner = mock(Cleaner.class);
     private final KeyValueService keyValueService = mock(KeyValueService.class);
 
-    private final SnapshotTransactionManagerImpl snapshotTransactionManager = new SnapshotTransactionManagerImpl(
+    private final SerializableTransactionManager snapshotTransactionManager = new SerializableTransactionManagerImpl(
```
sandorw: We should change the class name.

PR #2996, opened 2018-02-15
Remove immutableTs lock checks on reads

sandorw:
**Goals (and why)**:
Remove checks that protect each read on a thorough table from an expired immutableTs lock. The lock is still checked at commit time (could be refined further if we checked whether or not we've read from any tables that have thorough sweep turned on). The extra calls to timelock only protect against clients making use of state from a failed transaction, which means try/catching around the transaction and putting state in final variables outside the closure. We should communicate that all reads/writes are only valid if the transaction successfully commits, but I'd argue that anyone getting invalid data out at the moment knows that they're doing something wrong.

Is there any other reason we're making these lock checks?

**Priority (whenever / two weeks / yesterday)**: no rush

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2996)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -100,7 +100,6 @@ import com.palantir.atlasdb.keyvalue.impl.LocalRowColumnRangeIterator;
 import com.palantir.atlasdb.keyvalue.impl.RowResults;
 import com.palantir.atlasdb.logging.LoggingArgs;
 import com.palantir.atlasdb.logging.LoggingArgs.SafeAndUnsafeTableReferences;
-import com.palantir.atlasdb.protos.generated.TableMetadataPersistence.SweepStrategy;
 import com.palantir.atlasdb.sweep.queue.SweepQueueWriter;
 import com.palantir.atlasdb.table.description.exceptions.AtlasDbConstraintException;
 import com.palantir.atlasdb.transaction.api.AtlasDbConstraintCheckingMode;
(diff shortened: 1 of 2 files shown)


fsamuel-bs:
@gbonik for SA.

sandorw:
Apparently this code is quite old, so I'd like to get more internal input before merging.

sandorw:
This does potentially affect clis or any users of the AtlasDbService, where one must call the commit endpoint explicitly.

dxiao:
Does the atlasdb repo contain any tests that would exercise the edge case we're worried about?

sandorw:
I don't believe so. The edge cases mostly revolve around the usage contract of atlas. I'm pretty comfortable making this change as it affects anyone who uses the runTask() methods on the transaction managers, as you have to deliberately try to extract data out of a failed transaction to make this a problem.

Now that I think of it, it could affect long running transactions on tables with thorough sweep, where you'd otherwise fail fast on reads if your lock expired. I don't believe that this was a reason for the design, though, since it *only* affects tables with thorough sweep.

Clis seem like the main case where this could be a problem. If they open a long running transaction and then a human interactively makes decisions as they see data, you don't get any warning that you lost your immutable timestamp lock and that some of the reads may have been invalid until the cli finally closes the transaction.

sandorw:
Good counterpoint from @carrino: clients that are not expecting this behavior may not handle non-existence of data very well. Given a consistent snapshot of the DB, one might assume existence of data that could otherwise disappear underneath you if you lost the lock with this change. This could open up the runTask lambda to throwing IllegalArgument/IllegalState exceptions that wouldn't be thrown and retried.

sandorw:
Closing this - we'll consider making this change in an opt-in way if we think the performance improvement is worthwhile.

PR #2997, opened 2018-02-15
[TM Part 3] Transactions factory + ServiceWriteTransaction / ServiceReadOnlyTransaction

fsamuel-bs:
**Goals (and why)**:
1. Extract the transaction creation to a factory called `TransactionsFactory`.
2. Implement the `ServiceWriteTransaction` and `ServiceReadTransaction` from #2982.

To-do:

- [ ] tests
- [ ] write API contract

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2997)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
@@ -23,18 +23,19 @@ import com.palantir.atlasdb.transaction.api.TransactionManager;
 import com.palantir.atlasdb.transaction.api.TransactionTask;
 
 public interface SerializableTransactionManager extends TransactionManager {
-    RawTransaction setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
+    <C extends PreCommitCondition, E extends Exception> ServiceWriteTransaction getWriteTransaction(
+            C condition) throws E;
 
-    <T, E extends Exception> T finishRunTaskWithLockThrowOnConflict(RawTransaction tx,
-            TransactionTask<T, E> task)
-            throws E, TransactionFailedRetriableException;
+    <C extends PreCommitCondition, E extends Exception> ServiceReadOnlyTransaction getReadTransaction(
+            C condition) throws E;
 
-    /**
-     * Registers a Runnable that will be run when the transaction manager is closed, provided no callback already
-     * submitted throws an exception.
-     *
-     * Concurrency: If this method races with close(), then closingCallback may not be called.
-     */
+
+        /**
+         * Registers a Runnable that will be run when the transaction manager is closed, provided no callback already
+         * submitted throws an exception.
+         *
+         * Concurrency: If this method races with close(), then closingCallback may not be called.
+         */
     void registerClosingCallback(Runnable closingCallback);
 
     Cleaner getCleaner();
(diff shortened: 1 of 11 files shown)


fsamuel-bs:
A follow up step could be to extract all those parameters from the `SerTM` creation and inject a factory instead.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -18,16 +18,26 @@
...
+            C condition) throws E;
 
-    <T, E extends Exception> T finishRunTaskWithLockThrowOnConflict(RawTransaction tx,
-            TransactionTask<T, E> task)
-            throws E, TransactionFailedRetriableException;
+    /**
+     * Returns a transaction that can just be used for reads.
+     * Note that this transaction does not lock the immutable timestamp. Therefore, it is possible that sweep has
+     * deleted the values this transaction would read, causing correctness issues.
+     * The burden for locking the immutable timestamp, and preventing correctness problems is on the client.
```
sandorw: We should expand docs a bit more - the client should lock the immutable timestamp and also provide a lock checking precommit condition. If we want to help avoid clients misusing this, we could require a lock checking precommit condition as an argument.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerImpl.java
```diff
@@ -249,27 +241,82 @@ public SerializableTransactionManagerImpl(KeyValueService keyValueService,
...
+            C condition) throws E {
+        checkOpen();
+        try {
+            RawTransaction tx = setupRunTaskWithConditionThrowOnConflict(condition);
+            return new ServiceWriteTransaction((SerializableTransaction) tx.delegate()) {
+                @Override
+                public void commit() {
+                    try {
+                        tx.commit();
+                        finishRunTaskWithLockThrowOnConflict(tx, transaction -> null);
```
sandorw: We probably want to avoid any failures in the finishRunTask method from making the commit unsuccessful. If it's all best effort clean up, let's try/catch.

fsamuel-bs:
If we want to trust the transactions themselves though — which we probably should, given we're just wrapping a Snapshot or Serializable transaction — we could add some testing around the immutable timestamp lock for the `ServiceWriteTransaction` and `ServiceReadTransaction`, and some smoke testing of these two transactions — that they can read and write correctly.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerImpl.java
```diff
@@ -249,27 +241,82 @@ public SerializableTransactionManagerImpl(KeyValueService keyValueService,
...
     protected boolean shouldStopRetrying(int numTimesFailed) {
         return numTimesFailed > NUM_RETRIES;
     }
 
+    @Override
+    public <C extends PreCommitCondition, E extends Exception> ServiceWriteTransaction getWriteTransaction(
+            C condition) throws E {
+        checkOpen();
+        try {
+            RawTransaction tx = setupRunTaskWithConditionThrowOnConflict(condition);
```
sandorw: If we can move all internal uses over to the transaction factory, we should see if we can drop setup/finishRunTask methods from the public API.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerImpl.java
```diff
@@ -249,27 +241,82 @@ public SerializableTransactionManagerImpl(KeyValueService keyValueService,
...
+            SnapshotTransaction tx = transactionsFactory.createReadOnlyTransaction(condition,
+                    startTimestampSupplier,
+                    getApproximateImmutableTimestamp());
+            return new ServiceReadOnlyTransaction(tx) {
+                @Override
+                public void commit() {
+                    try {
+                        tx.commit();
+                    } finally {
+                        condition.cleanup();
```
sandorw: We should probably clean up the condition on abort as well for both methods.

PR #2998, opened 2018-02-15
Make sweep stats configurable

CRogers:
**Goals (and why)**:

Make sweep stats configurable so that services that write a small amount of data can still have sweep run.

Made `WRITE_THRESHOLD` and `WRITE_SIZE_THRESHOLD` configurable.

Soon

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2998)


atlasdb-dagger/src/main/java/com/palantir/atlasdb/services/KeyValueServiceModule.java
```diff
@@ -56,7 +57,14 @@ public KeyValueService provideWrappedKeyValueService(@Named("rawKvs") KeyValueSe
...
         kvs = AtlasDbMetrics.instrument(KeyValueService.class, kvs);
         kvs = ValidatingQueryRewritingKeyValueService.create(kvs);
-        kvs = SweepStatsKeyValueService.create(kvs, tss);
+
+        SweepConfig sweepConfig = config.atlasDbRuntimeConfig().sweep();
+        kvs = SweepStatsKeyValueService.create(
+                kvs,
+                tss,
+                sweepConfig::writeThreshold,
+                sweepConfig::writeSizeThreshold);
```
CRogers: Didn't have a `Supplier<Config>` here, I guess the whole class gets regenerated when the config changes? Or is something else happening? Have to init one in `TransactionManagers` too that does have a `Supplier<Config>`.

fsamuel-bs: This is just used for the CLIs, console and ETE tests, so we're not worried with having live-reloadable configs here.

PR #3000, opened 2018-02-16
Fix/migration cli

gmaretic:
**Goals (and why)**:
Avoid dropping the temporary table as it makes LoggingArgs error out on startup due to hydrating empty metadata. This avoids people having to patch their version or do db surgery to be able to start AtlasDB.

We leave an empty table around

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3000)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schemas.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schemas.java
@@ -135,8 +135,8 @@ public final class Schemas {
         return schemaFullTableNames.stream().filter(allTables::contains).collect(Collectors.toSet());
     }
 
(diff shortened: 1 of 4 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schemas.java
```diff
@@ -135,8 +135,8 @@ public static void truncateTablesAndIndexes(Schema schema, KeyValueService kvs)
         return schemaFullTableNames.stream().filter(allTables::contains).collect(Collectors.toSet());
     }
 
-    public static void deleteTable(KeyValueService kvs, TableReference tableRef) {
-        kvs.dropTable(tableRef);
```
jeremyk-91: OK. I'm puzzled why this method even exists, but sure

PR #3003, opened 2018-02-16
[Oracle Compaction][Part 3] Final refactors and tests

fsamuel-bs:
**Goals (and why)**:
- Increase network timeout on compactions
- Refactor BackgroundCompactor to make it more testable

To-do:

- [x] tests of `CompactPriorityCalculator`
- [x] tests of `BackgroundCompactor`
- [x] metrics on `BackgroundCompactor`
- [x] logs on `CompactPriorityCalculator`
- [x] fix metric reregistration warnings

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3003)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
@@ -91,6 +91,26 @@ public class MetricsManager {
         registerMetric(clazz, metricName, gauge, tag);
     }
 
+    /**
+     * Add a new gauge to a metric (which may already exist)
+     *
+     * This will throw {@link IllegalArgumentException} if a non-Gauge metric has already been added.
+     */
+    public void registerOrAddToMetric(Class clazz, String metricName, Gauge gauge, Map<String, String> tag) {
+        MetricName metricToAdd = MetricName.builder()
+                .safeName(MetricRegistry.name(clazz, metricName))
+                .safeTags(tag)
+                .build();
+
+        taggedMetricRegistry.gauge(metricToAdd, gauge);
+    }
+
+    /**
+     * Add a new gauge metric of the given name.
+     *
+     * If the metric already exists, this will REPLACE it with a new metric.
+     * Consider using {@link MetricsManager#registerOrAddToMetric} instead.
+     */
     public void registerMetric(Class clazz, String metricName, Gauge gauge, Map<String, String> tag) {
         MetricName metricToAdd = MetricName.builder()
                 .safeName(MetricRegistry.name(clazz, metricName))
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
@@ -86,25 +86,6 @@ public class MetricsManager {
         registerMetric(clazz, metricName, (Metric) gauge);
     }
 
-    public void registerGaugeForTable(Class clazz, String metricName, TableReference tableRef, Gauge gauge) {
-        Map<String, String> tag = getTableNameTagFor(tableRef);
-        registerMetric(clazz, metricName, gauge, tag);
-    }
-
-    /**
-     * Add a new gauge to a metric (which may already exist)
-     *
-     * This will throw {@link IllegalArgumentException} if a non-Gauge metric has already been added.
-     */
-    public void registerOrAddToMetric(Class clazz, String metricName, Gauge gauge, Map<String, String> tag) {
-        MetricName metricToAdd = MetricName.builder()
-                .safeName(MetricRegistry.name(clazz, metricName))
-                .safeTags(tag)
-                .build();
-
-        taggedMetricRegistry.gauge(metricToAdd, gauge);
-    }
-
     /**
      * Add a new gauge metric of the given name.
      *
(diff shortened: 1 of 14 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -60,10 +61,12 @@
             return Optional.empty();
         }
 
+        CompactPriorityCalculator compactPriorityCalculator = new CompactPriorityCalculator(transactionManager);
```
gsheasby: Would rather use a factory method in BackgroundCompactor so we don't have to create this and then pass it into the constructor - it's an implementation detail.

gsheasby:
The history's a bit messed up right now, because part 1 was merged and new commits have been added to part 2. The new commits from part 3 look ok to me (except for the nit in background compactor's creation). Remaining tasks as on the PR description.

gsheasby:
Might still want some tests, but this is ready to review (we might want to deploy it quickly)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
```diff
@@ -0,0 +1,114 @@
...
+                });
+
+        List<String> uncompactedTables = tableToLastTimeSwept.keySet().stream()
+                .filter(table -> !tableToLastTimeCompacted.keySet().contains(table))
+                .collect(Collectors.toList());
+
+        if (uncompactedTables.size() > 0) {
+            log.info("There are some tables which have been swept, but not compacted. Choosing one at random.");
+            int randomTableIndex = ThreadLocalRandom.current().nextInt(uncompactedTables.size());
+            return Optional.of(uncompactedTables.get(randomTableIndex));
```
jeremyk-91: OK. Maybe would be nice to log which one was chosen

gsheasby: That's already done by the BackgroundCompactor, but I guess there's no harm in doing it here.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -301,4 +309,24 @@ public void compactInternally(boolean inSafeHours) {
...
+
+        try {
+            int originalNetworkTimeout = sqlConnection.getUnderlyingConnection().getNetworkTimeout();
+            int newNetworkMillis = config.compactionConnectionTimeout() > Integer.MAX_VALUE
+                    ? (int) config.compactionConnectionTimeout()
+                    : Integer.MAX_VALUE;
+
+            log.info("Increased sql socket read timeout from {} to {}",
+                    SafeArg.of("originalNetworkTimeout", originalNetworkTimeout),
+                    SafeArg.of("newNetworkTimeout", newNetworkMillis));
```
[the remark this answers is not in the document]
gsheasby: note for self: also in background sweeper

gsheasby:
The tests are done, but when running them, I noticed a lot of metric re-registration warnings (that also affect the SweepOutcome metrics). Planning to fix those up within this PR, but it would require a new method in MetricsManager.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -91,6 +91,26 @@ public void registerGaugeForTable(Class clazz, String metricName, TableReference
         registerMetric(clazz, metricName, gauge, tag);
     }
 
+    /**
+     * Add a new gauge to a metric (which may already exist)
+     *
+     * This will throw {@link IllegalArgumentException} if a non-Gauge metric has already been added.
```
gsheasby: Do we want to handle the possible IllegalArgumentException here? If so, how - catch and log metric registration failure?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
```diff
@@ -0,0 +1,110 @@
...
+            String table = entry.getKey();
+            long lastSweptTime = entry.getValue();
+            long lastCompactTime = tableToLastTimeCompacted.get(table);
+            long sweptAfterCompact = lastSweptTime - lastCompactTime;
+
+            if (sweptAfterCompact > maxSweptAfterCompact) {
+                tableToCompact = table;
+                maxSweptAfterCompact = sweptAfterCompact;
+            }
+        }
```
jeremyk-91: It feels like we can do something clever...

```
        tableToLastTimeSwept.entrySet().stream()
                .max(Comparator.comparingLong(entry -> entry.getValue() - tableToLastTimeCompacted.get(entry.getKey())))
                .map(Map.Entry::getKey);
```

and that returns an `Optional<String>`, but I'm probably getting ahead of myself here.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
jeremyk-91: I'm quite indifferent to the options - the `IllegalArgumentException` would indicate an Atlas/product bug right?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
gsheasby: Yep, saw that possibility but I kept the code I inherited unchanged for the sake of velocity.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
gsheasby: Most likely. It would be one that we would want to fix, but wouldn't want to be woken up for - so we should probably `log.warn` instead of throwing. I don't think we actually call this method any more...

gsheasby: Oh wait this is the method I added. Yep, we call it, and yep it's only ever used for gauges, so the error should not happen (unless a future bug is written!)

gsheasby:
Will merge this (into a branch that has part 2, not develop) and then cherry-pick to the backport branch where the actual testing will take place.

Issue #3004, opened 2018-02-17
LoggingArgs.hydrate should not explode for empty table metadata


gmaretic:
We should either catch and log, or filter it out at the KVS getMetadataForTables level.

PR #3006, opened 2018-02-19
Fix/logging args

gmaretic:
**Goals (and why)**:
Fix LoggingArgs preventing AtlasDB to start on hydrating issues

SImple

Is the log message good?

Relevant and small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3006)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/logging/SafeLoggableDataUtils.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/logging/SafeLoggableDataUtils.java
@@ -21,6 +21,9 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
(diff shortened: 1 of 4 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/logging/SafeLoggableDataUtils.java
```diff
@@ -41,8 +46,13 @@ public static SafeLoggableData fromTableMetadata(Map<TableReference, byte[]> tab
...
         tableRefToMetadata.forEach(
                 (ref, metadataBytes) -> {
-                    TableMetadata tableMetadata = TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(metadataBytes);
-                    addLoggableNamesToBuilder(builder, ref, tableMetadata);
+                    try {
+                        TableMetadata tableMetadata = TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(metadataBytes);
+                        addLoggableNamesToBuilder(builder, ref, tableMetadata);
+                    } catch (Exception e) {
+                        log.warn("Exception thrown hydrating table metadata for table {}. {}",
+                                UnsafeArg.of("tableName", ref), UnsafeArg.of("exception", e));
```
jeremyk-91: Usually you can just leave the exception on at the end without needing to wrap it in an Arg. This way you get the stack trace (which is considered safe), though the (possibly unsafe) message is redacted.

gmaretic:
Fixes https://github.com/palantir/atlasdb/issues/3004

Issue #3007, opened 2018-02-20
SnapshotTransaction ignores ColumnSelection when there are local writes


nartmal:
I was trying to run ``table.getRowColumns(tableRow, table.getColumnSelection(NamedColumn.my_column))`` where tableRow was written within the same atlasTransaction. 

Instead of getting back my single ``NamedColumn.my_column``, I got all the columns from my local write.

@hsaraogi says ``extractLocalWritesForRow`` and the lines below it doesn't take into account the specified columnSelection

https://github.com/palantir/atlasdb/blob/a7c40bd52b48e4904d4496c34d89302f7e107d11/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java#L379

hsaraogi:
We get all the inmemory writes to the row back instead of just for the column selected.

PR #3008, opened 2018-02-20
SnapshotTransaction#extractLocalWritesForRow takes columnSelection

nartmal:
**Goals (and why)**:
Addresses https://github.com/palantir/atlasdb/issues/3007
ColumnSelection should be applied regardless of whether or not it came from a local write.

Soon-ish? Will work-around around until push is released.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3008)

### DIFF
--- a/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
+++ b/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
@@ -18,7 +18,6 @@ package com.palantir.atlasdb.transaction.impl;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasItem;
-import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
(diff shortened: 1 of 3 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -480,6 +481,34 @@ public void testTransactionIsolation() throws Exception {
...
+    @Test
+    public void testGetRowsLocalWritesWithColumnSelection() {
+        // This test ensures getRows correctly applies columnSelection
+        byte[] row1 = PtBytes.toBytes("row1");
+        Cell row1Column1 = Cell.create(row1, PtBytes.toBytes("column1"));
+        Cell row1Column2 = Cell.create(row1, PtBytes.toBytes("column2"));
+        byte[] row1Column1Value = BigInteger.valueOf(1).toByteArray();
+        byte[] row1Column2Value = BigInteger.valueOf(2).toByteArray();
+
+        Transaction snapshotTx = serializableTxManager.createNewTransaction();
```
nartmal: The CachingTransaction seems to hide the bug. The value will be present in ``CachingTransaction#getCachedCellIfPresent``, which will pass an empty ``Set<byte[]> toLoad`` to the delegate ``SnapshotTransaction``.

For some reason the unwrap method at the bottom didn't work for me either :( so I'm pulling in the serializableTxManager.

PR #3011, opened 2018-02-21
Implement callback for SerializableTM

gmaretic:
**Goals (and why)**:
Allow clients to specify a custom Runnable to be ran after the TM that is being initialized asynchronously is ready, but before we expose its methods.

Went for the simple approach

By design, if the callback throws or gets stuck, we will permanently fail in initializing the TM. We can change this, but I think it should be handled by the client.
Also fixed a bug that prevented us from closing a SerializableTM before it becomes initialized -- this action also prevents the callback from ever happening.

I think both of the above make sense, but open to discussion.

Tests?

Soonish

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3011)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -169,6 +169,19 @@ boolean allowHiddenTableAccess() {
 
     abstract TaggedMetricRegistry globalTaggedMetricRegistry();
 
+    /**
+     * If config().initializeAsync() is set to true, the callback Runnable will be run when the TransactionManager is
+     * successfully initialized. The TransactionManager stay uninitialized and continue to throw for all other purposes
+     * until the callback returns at which point it will become initialized.
+     *
+     * Note that if the callback throws an exception or blocks forever, the TransactionManager will never become
```
tboam: This can be read as "if the callback throws an exception the TransactionManager will never become initialized." which isn't what we want.

Does the callback need to catch its own failures and rollback any changes it made?

gsheasby: I don't think we can safely catch and ignore exceptions thrown by the callback - I think it's down to the caller to (a) handle exceptions, and (b) decide whether to abort, retry (with or without rollback), or ignore.

sandorw: I think we need some way to get signal out. One use case for this is to check that the DB state agrees with what the client expects. If the schema doesn't match (example check in the callback), we'd like the service to be able to report as unhealthy and know that the initialization failed.

gmaretic: @tboam Is it not what we want? If we do not want to say we are initialized until the callback is done, then we should fail on an exception. We cannot guess the caller's intention.

@sandorw We could store the exception from the callback so we can then log it, wrap it in the NotInitialized exception on method calls, or something similar.

gmaretic:
In addition to be able to close the TM, any methods called on it now correctly throw IllegalStateExceptions. Callback failing now also closes the TM, and the Throwable is added to the cause of the above ISE.

atlasdb-commons/src/main/java/com/palantir/async/initializer/Callback.java
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.async.initializer;
+
+public abstract class Callback {
```
gsheasby: Let's have some javadoc on this class, since we expect other product teams to implement it.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -174,12 +175,13 @@ boolean allowHiddenTableAccess() {
      * successfully initialized. The TransactionManager stay uninitialized and continue to throw for all other purposes
      * until the callback returns at which point it will become initialized.
      *
-     * Note that if the callback throws an exception or blocks forever, the TransactionManager will never become
-     * initialized.
+     * Note that if the callback blocks forever, the TransactionManager will never become initialized. If the callback
+     * throws, the TransactionManager will not become initialized and it's close() method will be called. The callback
```
gsheasby: it's --> its

*glares at the inconsistent grammatical rules English has*

The "it is" later in the sentence can become "it's" :joy: (I wouldn't do this, but really don't mind if you do)

atlasdb-commons/src/main/java/com/palantir/async/initializer/Callback.java
gsheasby: Also, (nit) there are already about 15 classes called `Callback`. `InitializationCallback`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -123,6 +135,19 @@ private boolean isInitializedInternal() {
                     && txManager.getCleaner().isInitialized()
                     && initializationPrerequisite.get();
         }
+
+        private void runCallback(Callback callBack) {
```
gsheasby: You're passing in `callBack` instead of using the field `callback`. Ugly 👎

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
```diff
@@ -46,76 +48,100 @@
...
     private AsyncInitializer mockInitializer = mock(AsyncInitializer.class);
-    private Runnable mockCallback = mock(Runnable.class);
+    private Callback mockCallback = mock(Callback.class);
 
     private SerializableTransactionManager manager;
 
     @Before
     public void setUp() {
-        when(mockKvs.isInitialized()).thenReturn(false);
+        setInitializationStatus(false, false, false, false);
```
gsheasby: optional suggestion: since we're extracting helper methods... it's mostly "false * 4" or "true * 4", so consider adding e.g. `whenEverythingIsInitialized` and `whenNothingIsInitialized`.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
```diff
@@ -46,76 +48,100 @@
...
+    public void callbackRunsOnlyOnceAsInitializationStatusChanges() {
+        setInitializationStatus(true, true, true, true);
+        Awaitility.waitAtMost(2L, TimeUnit.SECONDS).until(() -> manager.isInitialized());
+
+        setInitializationStatus(false, false, false, false);
+        assertNotInitializedWithinTwoSeconds();
+
+        setInitializationStatus(true, true, true, true);
+        assertTrue(manager.isInitialized());
+        verify(mockCallback).runWithRetry();
```
gsheasby: does this verify that it does it `times(1)`?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
```diff
@@ -46,76 +48,100 @@
...
-        };
-        manager = getManagerWithCallback(true, throwingCallback);
-        assertNotInitializedWithinTwoSeconds();
-        assertTrue(invoked.get());
+    public void exceptionInCleanupClosesTransactionManager() {
+        RuntimeException cause = new RuntimeException("VALID REASON");
+        doThrow(cause).when(mockCallback).runWithRetry();
+        setInitializationStatus(true, true, true, true);
+
+        assertClosedWitoutInitializingWithinTwoSeconds();
```
gsheasby: spelling nit: Witout -> Without

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
```diff
@@ -176,34 +194,56 @@ private SerializableTransactionManager getManagerWithCallback(boolean initialize
...
 
-    private static class BlockingRunnable implements Runnable {
+    private void assertClosedWitoutInitializingWithinTwoSeconds() {
+        Assertions.assertThatThrownBy(() ->
+                Awaitility.waitAtMost(2L, TimeUnit.SECONDS).until(() -> manager.isInitialized()))
+                .isInstanceOf(IllegalStateException.class);
+        assertTrue(manager.isClosed.get());
+    }
+
+
```
gsheasby: nit: `\n\n\n\n`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
gmaretic: Oops, this is an artefact from previous version 👎

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
gmaretic: That's what I get for laughing at others

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
gmaretic: That's the default, yes -- but can make it explicit

atlasdb-commons/src/test/java/com/palantir/async/initializer/CallbackTest.java
```diff
@@ -40,6 +46,28 @@ public void cleanupExceptionGetsPropagatedAndStopsRetrying() {
         assertThat(countingCallback.initCounter).isEqualTo(5L);
     }
 
+    @Test
+    public void shutdownDoesNotBlockWhenTaskIsNotRunning() {
+        CountingCallback countingCallback = new CountingCallback(true);
+
+        long start = System.currentTimeMillis();
+        countingCallback.blockUntilSafeToShutdown();
+        assertThat(System.currentTimeMillis()).isLessThanOrEqualTo(start + 100L);
```
gsheasby: Flagging that this seems a little risky - is 0.1 seconds a big enough window for a slammed CircleCI node?

Not blocking on this flag, as it could be fine in practice, and I guess there's always FlakeRule.

gmaretic: Should be fine since it should really not block at all, but if it flakes we can increase it

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -42,45 +47,106 @@
...
         private final Supplier<Boolean> initializationPrerequisite;
+        private final Callback callback;
 
-        public InitializeCheckingWrapper(SerializableTransactionManager manager,
-                Supplier<Boolean> initializationPrerequisite) {
-            this.manager = manager;
+        private volatile boolean callbackDone = false;
+        private Optional<Exception> callbackException = Optional.empty();
+
+        private final ScheduledExecutorService executorService = PTExecutors.newSingleThreadScheduledExecutor(
```
tboam: This executor isn't shutdown when the TM is closed.  No new tasks will be added but we should still clean up after ourselves.

It is also possible Atlas should try and fewer ThreadPools which are shared by many classes and then we can manage them more easily.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -42,45 +47,106 @@
...
+                    throw new IllegalStateException("Operations cannot be performed on closed TransactionManager."
+                            + " Closed due to a callback throw.", callbackException.get());
+                }
+                throw new IllegalStateException("Operations cannot be performed on closed TransactionManager.");
+            }
+        }
+
+        private void scheduleInitializationCheckAndCallback() {
+            executorService.schedule(() -> {
+                if (isClosed.get()) {
```
tboam: I would prefer we coded defensively and caught exceptions at the top level to make sure the follow up task is always scheduled. Otherwise we run the risk of deadlock.

atlasdb-commons/src/test/java/com/palantir/async/initializer/CallbackTest.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.assertj.core.api.Assertions.fail;
+
+import java.util.concurrent.TimeUnit;
+
+import org.awaitility.Awaitility;
+import org.junit.Test;
+
+import com.palantir.common.concurrent.PTExecutors;
+
+public class CallbackTest {
```
tboam: We don't test the NoOp callback anywhere which is what most products will run.  Sounds dumb but I want to be sure that doing nothing and exiting immediately doesn't cause problems.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -168,6 +169,20 @@ boolean allowHiddenTableAccess() {
 
     abstract TaggedMetricRegistry globalTaggedMetricRegistry();
 
+    /**
+     * If config().initializeAsync() is set to true, the callback Runnable will be run when the TransactionManager is
+     * successfully initialized. The TransactionManager stay uninitialized and continue to throw for all other purposes
+     * until the callback returns at which point it will become initialized.
+     *
+     * Note that if the callback blocks forever, the TransactionManager will never become initialized. If the callback
```
tboam: If the callback blocks forever on a single iteration it will also be impossible to close the TM as the close method will block indefinitely.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -168,6 +170,21 @@ boolean allowHiddenTableAccess() {
 
     abstract TaggedMetricRegistry globalTaggedMetricRegistry();
 
+    /**
+     * The callback Runnable will be run when the TransactionManager is successfully initialized. The
+     * TransactionManager will stay uninitialized and continue to throw for all other purposes until the callback
+     * returns at which point it will become initialized. If asynchronous initialization is disabled, the callback will
+     * be run jut before the TM is returned.
```
sandorw: jut -> just

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -42,45 +50,132 @@
...
+            return status == State.CLOSED_BY_CALLBACK_FAILURE;
+        }
+
+        private void assertOpen() {
+            if (status == State.CLOSED) {
+                throw new IllegalStateException("Operations cannot be performed on closed TransactionManager.");
+            }
+            if (status == State.CLOSED_BY_CALLBACK_FAILURE) {
+                throw new IllegalStateException("Operations cannot be performed on closed TransactionManager."
+                            + " Closed due to a callback throw.", callbackException);
```
sandorw: throw -> failure or exception?

PR #3013, opened 2018-02-22
Shutdown Client Pool if Creating C*KVS Fails

jeremyk-91:
**Goals (and why)**:
- Fix another possible source of the thread leaks in `TransactionManagers.serializable()`

- If creating a client pool succeeds but creating the Cassandra KVS fails, close the client pool.

- This didn't feel like it addressed everything we saw in some internal support tickets.

**Where should we start reviewing?**: `CassandraKeyValueServiceImpl.java`

@CRogers for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3013)


fsamuel-bs:
Wouldn't it be better to call `close` if line https://github.com/palantir/atlasdb/blob/develop/atlasdb-commons/src/main/java/com/palantir/async/initializer/AsyncInitializer.java#L59 throws an exception?

This seems like it's going to affect every downstream dependency if an upstream dependency fails to be initialized. We patched this at the TM level, but if an intermediary object fails to be started, this would also affect it.

jeremyk-91:
@fsamuel-bs Separately that makes sense as a change.

With this specific code path though, it's possible to fail outside of the context of an `AsyncInitializer`. Roughly speaking this path does the following:

```
Initialize a Cassandra client pool (which creates a thread for refreshing the pool)
Create the Atlas Cassandra KVS thread pool
Create a TracingPrefsConfig and call run() on it
Create and schedule the tracing-prefs executor
Create a bunch of Cassandra helper classes
Initialize the Cassandra key-value-service
```

It's possible for some of the steps in the middle to fail.

docs/source/release_notes/release-notes.rst
```diff
@@ -61,6 +61,11 @@ develop
            This fixes an issue that prevented AtlasDB to start after performing a KVS migration.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3006>`__)
 
+    *    - |fixed|
```
sandorw: Follow-up: make sure these don't get magically added to 0.78

PR #3014, opened 2018-02-23
Docs/manual compaction

gsheasby:
Do not merge yet - this assumes that the Oracle Compaction chain has merged to develop.

**Goals (and why)**: Add documentation explaining how sweep frees up data. Include instructions on manual action, for those KVSs that may require it.

**Implementation Description (bullets)**: Wrote some docs

**Concerns (what feedback would you like?)**: Are the instructions sufficiently clear, and the warnings obvious enough?

**Where should we start reviewing?**: The docs

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3014)

### DIFF
--- a/docs/source/cluster_management/sweep/freeing-space.rst
+++ b/docs/source/cluster_management/sweep/freeing-space.rst
@@ -1,0 +1,62 @@
+.. _freeing-space:
+
+=========================
+Freeing space after sweep
+=========================
+
+The sweep job issues KVS-level deletes to remove stale data.
+However, this does not immediately free data, and for Oracle KVS, manual action may be required in order to gain space.
+
+DbKvs
+=====
+
+DbKvs uses a background compaction thread to free up disk space. Similar to the background sweeper, the background
+compactor periodically chooses a table [#tableChoice]_, and runs ``KeyValueService.compactInternally`` on that table.
+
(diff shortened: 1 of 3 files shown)


docs/source/cluster_management/sweep/freeing-space.rst
```diff
@@ -0,0 +1,62 @@
...
+Oracle - Standard Edition
+-------------------------
+
+The background compaction thread runs one of two commands, depending on whether the wall-clock time is within safe hours.
+Safe hours indicate when users are unlikely to be accessing the system, and so blocking operations can safely be run.
+They are configured by a live-reloaded boolean ``inSafeHours`` (config path ``runtime/compact/inSafeHours``)
+
+.. warning::
+
+   ``inSafeHours`` only determines when we *start* running ``SHRINK SPACE``.
```
jeremyk-91: I renamed the parameter to be `inMaintenanceMode`, but everything else here should be the same

gsheasby: Ack - I've actually reworded the surrounding text too (maintenance mode isn't necessarily determined by wall-clock), so you might want to give it the once-over to ensure that I didn't miss any references and that the text still scans well.

PR #3018, opened 2018-02-26
[PoC] Timestamp Corroboration Consistency Check

jeremyk-91:
- [ ] audit that this is correct with respect to async initialisation
- [ ] audit that this behaviour won't throw nasty errors when everything works fine (even in the presence of GCs/network overhead)
- [x] better javadoc
- [x] unit tests
- [x] ETE smoke-test (where I manually reset the timelock clock)
- [x] changelog

**Goals (and why)**:
Limit the damage fires from misconfigured TimeLock servers/AtlasDB clients can cause.

Separately: should Async Init have a terminal state? Or should the consistency checks be exposed elsewhere?

- Compare the unreadable timestamp against a fresh timestamp, *reading the unreadable timestamp first* (ordering is important). Unreadable > fresh is bad.

- Can this actually throw false positives?
- How does this respond to async initialization?
- How should these checks be exposed to users?

**Where should we start reviewing?**: `TimestampCorroborationConsistencyCheck`

**Priority (whenever / two weeks / yesterday)**: soon. This limits P0s - but itself isn't one.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3018)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -317,6 +319,14 @@ private SerializableTransactionManager serializableInternal(@Output List<AutoClo
                         .buildCleaner(),
                 closeables);
 
+        // Run consistency checks
+        ConsistencyCheckRunner checkRunner = new ConsistencyCheckRunner(
+                ImmutableList.of(
+                        new TimestampCorroborationConsistencyCheck(
+                                cleaner::getUnreadableTimestamp,
```
sandorw: 👍 to the timestamp sources being pluggable

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/startup/ConsistencyCheckRunner.java
```diff
@@ -0,0 +1,71 @@
...
+import com.palantir.async.initializer.AsyncInitializer;
+import com.palantir.atlasdb.factory.TransactionManagerConsistencyResult;
+import com.palantir.atlasdb.transaction.impl.consistency.TransactionManagerConsistencyCheck;
+import com.palantir.common.remoting.ServiceNotAvailableException;
+
+public final class ConsistencyCheckRunner extends AsyncInitializer {
+
+    private static final Logger log = LoggerFactory.getLogger(ConsistencyCheckRunner.class);
+
+    private static final RuntimeException UNKNOWN = new RuntimeException("unknown");;
```
sandorw: nit: ;;

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/startup/ConsistencyCheckRunner.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public ConsistencyCheckRunner(List<TransactionManagerConsistencyCheck> consistencyChecks) {
+        this.consistencyChecks = consistencyChecks;
+    }
+
+    @Override
+    protected void tryInitialize() {
+        TransactionManagerConsistencyResult consistencyResult = consistencyChecks.stream()
+                .map(Supplier::get)
+                .sorted(Comparator.comparing(
```
sandorw: Feels gross to depend on the ordering here. If we do this, leave a comment on the enum so that this doesn't change.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        if (freshTimestamp < lowerBound) {
+            log.error("Your AtlasDB client believes that the unreadable timestamp was {}, but that's newer than a"
+                    + " fresh timestamp of {}, which implies clocks went back. If using TimeLock, this could be because"
+                    + " timestamp bounds were not migrated properly - which can happen if you've moved TimeLock"
+                    + " Server without moving its persistent state. For safety, AtlasDB will refuse to start.",
+                    SafeArg.of("unreadableTimestamp", lowerBound),
+                    SafeArg.of("freshTimestamp", freshTimestamp));
+            return ImmutableTransactionManagerConsistencyResult.builder()
```
sandorw: We're not catching an exception here, but it would be nice if we had a reason other than unknown to put in the result to make the eventual exception more legible.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/startup/ConsistencyCheckRunner.java
tboam: You could maybe have a severity field on the enum which is a number and sort by that.

jeremyk-91: Yep, will add severity. (Separately, the comparator should have been `reversed()` too - the old code would accidentally have picked the healthiest status, which is really bad!)

jeremyk-91: Agreed, will add severity.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java
jeremyk-91: OK. I've created a 'dummy' `AssertionError` for this case that includes information about the lower bound/timestamp.

jeremyk-91:
Note: Rebasing on top of #3011 as fixing the issues with async initialization is likely to involve duplicating a lot of the work done in that PR

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/startup/ConsistencyCheckRunner.java
```diff
@@ -0,0 +1,68 @@
...
+                log.info("Cluster appears consistent.");
+                break;
+            default:
+                throw new IllegalStateException("Unexpected consistency state " + consistencyResult.consistencyState());
+        }
+    }
+
+    @Override
+    public void cleanup(TransactionManager resource, Exception initException) {
+        // No op; consistency checks shouldn't use their own resources / should clean up after themselves.
```
gmaretic: You have to rethrow here, otherwise the exception will be swallowed

gmaretic: Discussed offline, this is actually good

jeremyk-91: Note: It's good because `Error` breaks out of the callback framework, and we do want to retry the `NotInitializedException`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java
```diff
@@ -0,0 +1,100 @@
...
+    private final ToLongFunction<TransactionManager> freshTimestampSource;
+
+    public TimestampCorroborationConsistencyCheck(
+            ToLongFunction<TransactionManager> conservativeBound,
+            ToLongFunction<TransactionManager> freshTimestampSource) {
+        this.conservativeBound = conservativeBound;
+        this.freshTimestampSource = freshTimestampSource;
+    }
+
+
```
jeremyk-91: double line

gsheasby:
LGTM. There's one spot in the code which is left open to a careless programmer could cause bugs in the future, but it's very minor.
Some of the test coverage and documentation here (e.g. CallChain) is excellent.

---

Reviewed 1 of 5 files at r5, 4 of 11 files at r7, 1 of 5 files at r8, 6 of 6 files at r9.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java, line 69 at r1](https://reviewable.io:443/reviews/palantir/atlasdb/3018#-L6Iy9FtHBbDGq4TmbPY-r1-69:-L8aiGqnxWvfrPoT7Odv:by5qv8i) ([raw file](https://github.com/palantir/atlasdb/blob/8c7946dc7b9689aace31e1263f609df36036fc16/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java#L69)):*
<details><summary><i>Previously, jeremyk-91 (Jeremy Kong) wrote…</i></summary><blockquote>

OK. I've created a 'dummy' `AssertionError` for this case that includes information about the lower bound/timestamp.
</blockquote></details>

Looks informative enough to me

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java, line 42 at r9](https://reviewable.io:443/reviews/palantir/atlasdb/3018#-L8ahWfayTV5K-I05j1J:-L8ahWfayTV5K-I05j1K:b-8ky0yy) ([raw file](https://github.com/palantir/atlasdb/blob/bb40e206ddd7cbfb569a00af623cec4e7c2a2239/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/consistency/TimestampCorroborationConsistencyCheck.java#L42)):*
> ```Java
>     private final ToLongFunction<TransactionManager> freshTimestampSource;
> 
>     public TimestampCorroborationConsistencyCheck(
> ```

FYI, this is somewhat error-prone - the developer needs to remember that we first pass in the conservative bound, and then the source of fresh timestamps. A builder would make it clearer. Not blocking on this, though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/atlasdb/3018#-:-L8aoavJjWQWuCeuaK8p:bl8wvom)*

gsheasby:
Smoke testing underway. Steps:
1. Upgrade secure storage product to have this Atlas code
2. Start up, observe "Passed timestamp corroboration consistency check" message in logs
_3. Wait until the unreadable timestamp passes the timestamp at which the service migrated to timelock_
4. Shut the service down
5. Shut timelock down
6. Manually remove the service's subdirectory from `var/data/paxos` on all three timelock nodes
7. Start timelock up
8. Start the service up; verified that the "Failed consistency check" message was logged

This check will have a false _negative_ (i.e. time will go backwards and we won't know about it) if the data is lost within an hour of the product switching to timelock, but I don't know if we can do better than that.

gsheasby:
Smoke test successful - log from starting the service after removing paxos data:
```
{"type":"service.1","level":"ERROR","time":"2018-03-28T14:56:24.307Z","origin":"com.palantir.atlasdb.transaction.impl.consistency.TimestampCorroborationConsistencyCheck","thread":"main","message":"Your AtlasDB client believes that a strict lower bound for the timestamp was {} (typically by reading the unreadable timestamp), but that's newer than a fresh timestamp of {}, which implies clocks went back. If using TimeLock, this could be because timestamp bounds were not migrated properly - which can happen if you've moved TimeLock Server without moving its persistent state. For safety, AtlasDB will refuse to start.","params":{"timestampLowerBound":2048930,"freshTimestamp":1010005},"uid":null,"sid":null,"tokenId":null,"traceId":"23873b13b98f88e9","stacktrace":null,"unsafeParams":{},"tags":{}}
```

The service appeared as "failed to start" on the stack management service.

PR #3020, opened 2018-02-28
speed up certain types of read-only txn

clockfort:
remove a lock round trip for read only transactions that only involve non-thoroughly swept tables

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3020)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -213,6 +213,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
     protected final long lockAcquireTimeoutMs;
     protected final ExecutorService getRangesExecutor;
     protected final int defaultGetRangesConcurrency;
+    private final Set<TableReference> involvedTables = Sets.newConcurrentHashSet();
 
     protected volatile boolean hasReads;
 
@@ -352,6 +353,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
     }
 
     protected void checkGetPreconditions(TableReference tableRef) {
+        markTableAsInvolvedInThisTransaction(tableRef);
         if (transactionReadTimeoutMillis != null
                 && System.currentTimeMillis() - timeCreated > transactionReadTimeoutMillis) {
             throw new TransactionFailedRetriableException("Transaction timed out.");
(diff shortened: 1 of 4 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -213,6 +213,7 @@
     protected final long lockAcquireTimeoutMs;
     protected final ExecutorService getRangesExecutor;
     protected final int defaultGetRangesConcurrency;
+    private final Set<TableReference> involvedTables = Sets.newHashSet();
```
schlosna: Does this need to be a synchronized or concurrent set?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1976,6 +1979,18 @@ public void useTable(TableReference tableRef, ConstraintCheckable table) {
         constraintsByTableName.put(tableRef, table);
     }
 
+    /** the similarly-named-and-intentioned useTable method is only called on writes,
+     *  this one is more comprehensive and covers read paths as well
+     * (necessary because we wish to get the sweep strategies of tables in read-only transactions)
+     */
+    private void markTableAsInvolvedInThisTransaction(TableReference tableRef) {
```
sandorw: Confirmed that this gets called on all public API get/put methods.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -114,7 +114,7 @@ public void shouldNotMakeRemoteCallsInAReadonlyTransactionIfNoWorkIsDone() {
     }
 
     @Test
-    public void shouldConflictIfImmutableTimestampLockExpiresEvenIfNoWrites() {
+    public void shouldConflictIfImmutableTimestampLockExpiresEvenIfNoWritesOnThoroughSweptTable() {
```
sandorw: Can we add the opposite test - don't conflict if there were reads but not on a thorough sweep table?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1976,6 +1979,18 @@ public void useTable(TableReference tableRef, ConstraintCheckable table) {
...
+        involvedTables.add(tableRef);
+    }
+
+    private boolean validationNecessaryForInvolvedTables() {
+        return !Sets.filter(involvedTables, tableRef -> isValidationNecessary(tableRef)).isEmpty();
```
schlosna: nit: might be clearer as:

`return involvedTables.stream().anyMatch(this::isValidationNecessary)`

Issue #3021, opened 2018-03-02
BackgroundSweeper can fail to shut down correctly


chrisalice:
The BackgroundSweeper is a daemon thread but on shutdown, a non-daemon thread interrupts it, and then blocks waiting for it to stop. Unfortunately, if the sweeper is in PersistentLockManager.waitForRetry, the InterruptedException will be swallowed, meaning the thread never shuts down. This is causing an internal auth product to fail to stop gracefully, requiring a kill -9. (Logs demonstrating the exact sequence of events available on request.)

Recommended steps to solve:
 - [x] Fix PersistentLockManager.waitForRetry to correctly call Thread.currentThread().interrupt()
 - [ ] Audit Atlas for any other mishandling of InterruptedException
 - [x] Use a CountDownLatch as well as daemon.interrupt() to signal shutdown
 - [x] Replace all calls to Thread.sleep() with latch.await() in BackgroundSweeperImpl
 - [x] Add a timeout to the join() call in BackgroundSweeperImpl

This should provide a robust long-term solution that cannot be easily broken by any InterruptedException mishandling in future.

tboam:
Done and released on 0.78 (available but not announced yet).

chrisalice:
Note that I did not audit Atlas for any other mishandling of InterruptedException.

PR #3023, opened 2018-03-02
PersistentLockManager is swallowing thread interrupts

chrisalice:
**Goals (and why)**:

The PersistentLockManager is catching InterruptedException and rethrowing it wrapped in a RuntimeException, causing the interrupted status of the thread to be lost. This is causing the BackgroundSweeper to fail to shutdown cleanly.

See also #3021.

 * Call `currentThread().interrupt()` before wrapping the exception
 * `Throwables.propagate` is deprecated; in this case, we should just be creating our own RuntimeException.

Causing issues with an internal security product.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3023)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
@@ -25,7 +25,6 @@ import org.slf4j.LoggerFactory;
 import com.codahale.metrics.Meter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
 import com.palantir.atlasdb.keyvalue.api.CheckAndSetException;
 import com.palantir.atlasdb.persistentlock.LockEntry;
 import com.palantir.atlasdb.persistentlock.PersistentLockId;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
```diff
@@ -146,7 +145,8 @@ private void waitForRetry() {
         try {
             Thread.sleep(persistentLockRetryWaitMillis);
         } catch (InterruptedException e) {
-            throw Throwables.propagate(e);
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
         }
```
jeremyk-91: maybe there should be a baseline rule for 'catch InterruptedException and immediately rethrow'

chrisalice: Would be difficult to code — the internal version of Throwables in atlasdb-commons handles InterruptedException correctly, we wouldn't want to mess up that code.

schlosna: its probably worth an audit of the deprecated `com.google.common.base.Throwables#propagate` as there are definitely cases where we're not resetting interrupts, as well as other places catching `InterruptedException`  (e.g. where we just log or continue in Scrubber and LockAcquirer) that should also likely be resetting.

PR #3024, opened 2018-03-02
Robustify background sweeper

chrisalice:
**Goals (and why)**:

Harden the background sweeper against InterruptedException bugs. The current implementation relies on all code everywhere correctly handling thread interruption; if any wait call anywhere erroneously swallows the interrupt, the daemon will continue to run, preventing clean shutdown.

See issue #3021.

 * Trip a CountDownLatch after interrupting the thread
 * Replace all calls to `Thread.sleep` with a wait on this latch
 * Only wait for 10s for the background sweep thread

Causing issues with an internal security product.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3024)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
@@ -18,6 +18,7 @@ package com.palantir.atlasdb.sweep;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -297,9 +308,16 @@ public synchronized void shutdown() {
...
+        // Interrupt the daemon, whatever lock it may be waiting on.
         daemon.interrupt();
+        // Ensure we do not accidentally abort shutdown if any code incorrectly swallows InterruptedExceptions
+        // on the daemon thread.
+        shuttingDown.countDown();
         try {
-            daemon.join();
+            daemon.join(MAX_DAEMON_CLEAN_SHUTDOWN_TIME_MILLIS);
+            if (daemon.isAlive()) {
+                log.error("Background sweep thread failed to shut down");
```
jeremyk-91: Discussed offline - this might seem a bit harsh, though it definitively indicates a product bug (i.e. someone's swallowing an interrupt and slept for more than 10s)

PR #3025, opened 2018-03-02
[Oracle Compaction][Part 4] Bug Bash

jeremyk-91:
**Goals (and why)**:
- Make Oracle compaction work

Fixes the following issues with the original impl:

* We don't believe a compactor is needed
    * Delegating KVSs need to route shouldTriggerCompactions() to delegates
* Compaction never respects safe hours, always calling raw compactInternally() e.g. SHRINK SPACE COMPACT only
    * Delegating KVSs need to route compactInternally() to delegates
* Background Sweeper in 0.39 compacts at the end, disregarding safe hours
    * Sweeper shouldn't call compaction
* Compactor repeatedly throws exceptions in compactTable()
    * Cannot assume in CompactPriorityCalculator that table-names are fully qualified

* Are the delegating implementations correct? In particular, I'm uncertain of the Dual Write one.

**Where should we start reviewing?**:  wherever. I'd suggest looking at the issues above, identifying that the fixes proposed for each are reasonable, and then linking them up to the code changes made.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3025)

### DIFF
--- a/atlasdb-cli-distribution/versions.lock
+++ b/atlasdb-cli-distribution/versions.lock
@@ -322,7 +322,8 @@
             "project": true,
             "transitive": [
                 "com.palantir.atlasdb:atlasdb-config",
-                "com.palantir.atlasdb:atlasdb-dbkvs"
+                "com.palantir.atlasdb:atlasdb-dbkvs",
+                "com.palantir.atlasdb:atlasdb-service"
             ]
         },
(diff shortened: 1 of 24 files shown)


jeremyk-91:
Yes. I think compaction is expected to be supported regardless of whether we shouldTriggerCompactions or not. Longer term gating could be a good solution, though merging for now

PR #3026, opened 2018-03-04
Do not lose interrupt.

AmlingPalantir:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3026)


schlosna:
Is this a dupe of #3023 ?

gmaretic:
A fix for this was already merged to develop as pointed out above!

AmlingPalantir:
Sigh, that's what I get for only having looked at the version we're actually using.  See you next bump...

PR #3028, opened 2018-03-05
Oracle Compaction to 0.39.x

jeremyk-91:
**Goals (and why)**:
- Release a version of AtlasDB 0.39.x with Oracle compaction fixes

- see individual PRs

- none in particular; see individual PRs for questions around these

**Where should we start reviewing?**: probably easier if you look at the individual PRs

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3028)


atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -243,26 +250,80 @@ private void executeIgnoringError(String sql, String errorToIgnore) {
     }
 
     @Override
-    public void compactInternally() {
+    public void compactInternally(boolean inSafeHours) {
+        final String compactionFailureTemplate = "Tried to clean up {} bloat after a sweep operation,"
```
tboam: Minor: this message is a little out of date.  e.g. Assumes compact is running after a sweep, also not sure what "Since this can't be automated in your configuration," means and whether it is still correct.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -243,26 +250,80 @@ private void executeIgnoringError(String sql, String errorToIgnore) {
...
+                        "(Enterprise Edition that requires this user to be able to perform DDL operations)."
+                                + " Please change the `enableOracleEnterpriseFeatures` config to false.",
+                        e.getMessage());
+            } catch (TableMappingNotFoundException e) {
+                throw Throwables.propagate(e);
+            }
+        } else if (config.enableShrinkOnOracleStandardEdition()) {
+            Stopwatch timer = Stopwatch.createStarted();
+            try {
+                if (inSafeHours) {
```
tboam: Maybe too late but 'safe hours' could be misinterpreted.  Does it mean "safe to perform locking DB operations" or "keep these hours safe as users are using the system".  Maybe 'maintenance hours', or 'table move hours'.  I don't feel strongly though so if the messaging is clear where this is configured then ignore this.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresDdlTable.java
```diff
@@ -134,7 +134,7 @@ public void checkDatabaseVersion() {
     }
 
     @Override
-    public void compactInternally() {
+    public void compactInternally(boolean unused) {
         // VACUUM FULL is /really/ what we want here, but it takes out a table lock
```
tboam: Interesting comment, suggests we should maybe have a similar set up in postgres.  Don't need to worry about this just now though.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public void run() {
+        log.info("Attempting to start the background compactor for the very first time.");
+        try (SingleLockService compactorLock = createSimpleLocks()) {
+            log.info("Starting background compactor");
+            while (true) {
+                if (Thread.currentThread().isInterrupted()) {
+                    log.warn("Shutting down background compactor because someone interrupted its thread");
+                    Thread.currentThread().interrupt();
```
tboam: We had a bug in the background sweeper where we weren't spotting InterruptedException in all cases, does a similar bug exist in the background compactor?

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
jeremyk-91: Fixed the message on compact after sweep.

Regarding "since this can't be automated", I suspect the writer was referring to a world in which we didn't have safe hours (and so running SHRINK SPACE all the time may not have been safe)?

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
jeremyk-91: Right! I think `maintenanceHours` is substantially clearer - will rename to this.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresDdlTable.java
jeremyk-91: Postgres docs: https://www.postgresql.org/docs/9.5/static/sql-vacuum.html

> 
Plain VACUUM (without FULL) simply reclaims space and makes it available for re-use. This form of the command can operate in parallel with normal reading and writing of the table, as an exclusive lock is not obtained. However, extra space is not returned to the operating system (in most cases); it's just kept available for re-use within the same table. VACUUM FULL rewrites the entire contents of the table into a new disk file with no extra space, allowing unused space to be returned to the operating system. This form is much slower and requires an exclusive lock on each table while it is being processed.

So interestingly the (plain) VACUUM we run for postgres seems to be analogous to SHRINK SPACE COMPACT - and we always run the conservative command in postgres, never running VACUUM FULL (which is analogous to the blocking SHRINK SPACE). Could be something we could do in the future, yes.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
jeremyk-91: I've checked all of the try-catch blocks here and it doesn't look like we have a case where we catch an InterruptedException and then don't reset the flag or throw. (Note: in `grabLockAndRunOnce` the last two try-catch blocks don't throw InterruptedException).

PR #3029, opened 2018-03-05
Add API for releasing schema mutation lock

sandorw:
**Goals (and why)**:
Add an endpoint for releasing the schema mutation lock for environments where a CLI can't be supported.

Added a new persistent lock service endpoint that runs the equivalent of the clean lock state CLI command. This is only needed for Cassandra, so a ServiceLoader is used to avoid leaking Cassandra into general projects.

Interactions with async init? Will test this, but will the endpoints be mounted before we reach potential lock contention in creating tables?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3029)


sandorw:
Verified that this does work on a test stack. It does generally require using async initialization to ensure that the endpoint is mounted and available for use despite the transaction manager failing to initialize correctly (if the lock leaked, we expect the transaction manager not to initialize if there are schema changes).

sandorw:
Closing the PR and going with a different approach. It looks like under the circumstances where this endpoint is needed, the service will be in an unhealthy state and the endpoint won't be useable.

PR #3030, opened 2018-03-06
Go/timelock service status

gokcanozakdag:
**Goals (and why)**:
Expose a health check method in TM, available to be used by clients for their own health check reports.
Created a wrapper class for timelockService, that collects failure and success metrics each time a method is invoked. Health check compares failure and success rates to decide healthy or not.
Health check method only works if metrics are registered correctly, which requires TM to be created using `serializable()` method. If it is possible to create it other ways, health check won't be functional.
`InstrumentedTimelockService`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3030)


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -458,6 +459,19 @@ public void usesTimeLockIfInstallConfigIsTimeLockAndInitialRuntimeConfigDoesNotC
...
+        assertTrue(tm.getTimelockServiceStatus().isHealthy());
+    }
+
+    @Test
+    public void timelockServiceStatusReturnsHealthyAfterSuccessfulRequests() {
+        TransactionManager tm = TransactionManagers.createInMemory(GenericTestSchema.getSchema());
+        tm.getUnreadableTimestamp();
+        assertTrue(tm.getTimelockServiceStatus().isHealthy());
+    }
+
```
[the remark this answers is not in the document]
gsheasby: If the health check logic is extracted to its own class with a reference to an `InstrumentedTimelockService`, then we can set up a mocked `TimelockService` and add unit tests to verify that:
a) `InstrumentedTimelockService` correctly records successes/failures and rethrows exceptions (we may only want to write these for one or two methods, as it'll be tedious to write the same two or three tests for every method)
b) `TimelockServiceStatusHealthCheck` returns healthy when half or more of recent requests succeeded

We might need to be careful to ensure that either the health check cannot be invoked before the timelock is created, or that the health check returns something sensible in that case.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TimelockHealthCheckTest.java
```diff
@@ -0,0 +1,129 @@
...
+            e.printStackTrace();
+        }
+
+        Awaitility.await()
+                .ignoreExceptions()
+                .atMost(1000, TimeUnit.MILLISECONDS)
+                .until(() -> {
+                    try {
+                        serializableTransactionManager.getImmutableTimestamp();
+                        return true;
```
gsheasby: You want to return `false` here and `true` when there's an exception, since we expect an exception here.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TimelockHealthCheckTest.java
```diff
@@ -0,0 +1,129 @@
...
+        ensureTransactionManagerIsCreated();
+        try {
+            EteSetup.getContainer("timelock").stop();
+        } catch (InterruptedException | IOException e) {
+            e.printStackTrace();
+        }
+
+        Awaitility.await()
+                .ignoreExceptions()
+                .atMost(1000, TimeUnit.MILLISECONDS)
```
gsheasby: nit: `Duration.ONE_SECOND`

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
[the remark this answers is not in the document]
gokcanozakdag: Moved the health check logic to a separate class, and added unit tests for that. I have also removed the ete tests for healthcheck, as I thought it is redundant after adding unit tests; but I can add it back if you think it makes sense to keep both @gsheasby

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/health/MetricsBasedTimelockHealthCheckTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import com.palantir.atlasdb.transaction.impl.InstrumentedTimelockService;
+import com.palantir.atlasdb.util.AtlasDbMetrics;
+import com.palantir.lock.v2.TimelockService;
+
+public class MetricsBasedTimelockHealthCheckTest {
```
gsheasby: Please add test cases for "one success after multiple failures => unhealthy", and "multiple successes and one failure => healthy"

It might be easier to do these by mocking two `TimelockService` endpoints, and having one succeed and the other fail.

gokcanozakdag: Done!

PR #3034, opened 2018-03-07
Bump Cell creation error to ERROR logs

fsamuel-bs:
**Goals (and why)**: Debug seems very low this fails transactions in a non-retryable way.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3034)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Cell.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Cell.java
@@ -31,6 +31,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.primitives.Bytes;
 import com.google.common.primitives.UnsignedBytes;
 import com.palantir.atlasdb.encoding.PtBytes;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 
 /**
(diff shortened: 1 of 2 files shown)


fsamuel-bs:
I've also had the same doubt. I think this should be logged at ERROR though:

1. This should not happen very frequently, or with a certain cadence — this will only happen when someone tries to put something with a large key size, which should fail without any retries in Atlas, and probably will cause a 5xx to be thrown from the service.

2. Given our current ERROR/WARN distribution, I'd expect this to happen once every week, if most. In this case, it would not show up in our metrics / logs analysis unless its an error.

sandorw:
I think this is a case where we want fairly strong warnings, since it can suggest a bad schema design - if some keys exceed the length, it's not unreasonable that many might exceed it and the application is basically non-functional. Given that, error seems okay to me.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Cell.java
```diff
@@ -66,9 +67,11 @@ private void validateNameValid(byte[] name) {
                     "name must be no longer than {}.",
                     MAX_NAME_LENGTH);
         } catch (IllegalArgumentException e) {
-            log.debug("Cell creation that was attempted was: {}; since the vast majority of people encountering this "
+            log.error("Cell creation that was attempted was: {}; since the vast majority of people encountering this "
```
sandorw: This error message is pretty non-descriptive... can we include the exception message above? Name exceeded maximum name length

PR #3037, opened 2018-03-08
[Oracle Compaction][Part 6] Protect from Thread Death

jeremyk-91:
**Goals (and why)**:
- Be resilient to the case where the kvs / timelock isn't available when the background compactor starts

- Catch Exceptions and keep running the main loop, much like Sweep

- Are there still places which can throw and cause the background compactor to stop?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3037)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
@@ -86,7 +86,11 @@ class CompactPriorityCalculator {
             }
         }
 
-        logCompactionChoice(tableToCompact, maxSweptAfterCompact);
+        if (tableToCompact == null) {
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactPriorityCalculator.java
```diff
@@ -99,7 +103,7 @@ private void logCompactionChoice(String tableToCompact, long maxSweptAfterCompac
             log.info("All swept tables have been compacted after the last sweep. Choosing to compact {} anyway - "
                     + "this may be a no-op. It was last compacted {} milliseconds after it was last swept.",
                     tableToCompact,
-                    maxSweptAfterCompact);
+                    -maxSweptAfterCompact);
```
gmaretic: Math.abs()

PR #3038, opened 2018-03-08
Guava 21.0 -> 23.6-jre

jeremyk-91:
**Goals (and why)**:
- Upgrade Guava to a more modern version
- Unblock Atlas users from upgrading to more recent versions of Guava / avoid conflicts with using more recent versions of Guava.

- Bump Guava
- Fix the break in `BlockingTimeLimitedLockService` and its tests, owing to a Guava-side break in `SimpleTimeLimiter`; in particular, it throws a new type of exception.
- Release notes.

- Are there other ramifications of upgrading Guava that I should be aware of? Note that the large internal product already uses this version of Guava.

- Commit by commit probably makes sense. This is only large because of dependency locks

**Priority (whenever / two weeks / yesterday)**: 2 weeks?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3038)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/BlockingTimeLimitedLockService.java
```diff
@@ -65,7 +66,12 @@
 
     public static BlockingTimeLimitedLockService create(CloseableLockService lockService,
             long blockingTimeLimitMillis) {
-        return new BlockingTimeLimitedLockService(lockService, new SimpleTimeLimiter(), blockingTimeLimitMillis);
+        // TODO (jkong): Inject the executor to allow application lifecycle managed executors.
+        // Currently maintaining existing behaviour.
+        return new BlockingTimeLimitedLockService(
+                lockService,
+                SimpleTimeLimiter.create(Executors.newCachedThreadPool()),
```
sandorw: PtExecutors?

jeremyk-91: Sure. Didn't even cross my mind - will update

PR #3040, opened 2018-03-09
[WIP] Cassandra address translation

gmaretic:
**Goals (and why)**:
Add config for C* address transalation.

Static map that applies the translation whenever we get an endpoint from C* (which all goes through getAddressForHost()). Also fixed a bug there that was making testTokenMapping() a noop, the test now tests the translation.

Is this really all we need to do? There also seem to be quite a few bugs in the code around this, which will be addressed in follow-up PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3040)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -144,11 +144,14 @@ public class CassandraService implements AutoCloseable {
     }
 
     public InetSocketAddress getAddressForHost(String host) throws UnknownHostException {
-        InetAddress resolvedHost = InetAddress.getByName(host);
+        if (config.todoMap().containsKey(host)) {
+            return config.todoMap().get(host);
+        }
 
+        InetAddress resolvedHost = InetAddress.getByName(host);
         Set<InetSocketAddress> allKnownHosts = Sets.union(currentPools.keySet(), config.servers());
         for (InetSocketAddress address : allKnownHosts) {
-            if (address.getAddress().equals(resolvedHost)) {
+            if (resolvedHost.equals(address.getAddress())) {
                 return address;
             }
         }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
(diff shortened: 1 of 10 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -144,11 +144,14 @@ private InetSocketAddress getAddressForHostThrowUnchecked(String host) {
...
-        InetAddress resolvedHost = InetAddress.getByName(host);
+        if (config.addressTranslation().containsKey(host)) {
+            return config.addressTranslation().get(host);
+        }
 
+        InetAddress resolvedHost = InetAddress.getByName(host);
         Set<InetSocketAddress> allKnownHosts = Sets.union(currentPools.keySet(), config.servers());
         for (InetSocketAddress address : allKnownHosts) {
-            if (address.getAddress().equals(resolvedHost)) {
+            if (resolvedHost.equals(address.getAddress())) {
```
ashrayjain: Let's use `Objects.equals(address.getAddress(), resolvedHost)`  instead?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -144,11 +145,14 @@ private InetSocketAddress getAddressForHostThrowUnchecked(String host) {
     }
 
     public InetSocketAddress getAddressForHost(String host) throws UnknownHostException {
-        InetAddress resolvedHost = InetAddress.getByName(host);
+        if (config.addressTranslation().containsKey(host)) {
```
tboam: 1) Will describe_ring always give us back the same addresses for the other nodes?  Is there any variation in format?
We may have to put multiple entries for the same host into the map, not a problem just noting it.

2) What if a data centre rack is replaced and the C* node is moved, will the address remain the same?
I'm trying to think up scenarios where the config might have to be dynamic.  May not be needed but I want to be sure.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |new|
+         - Added a new parameter ``addressTranslation`` to ``CassandraKeyValueServiceConfig``.
+           This parameter is a static map specifying how internal Cassandra endpoints should be translated to InetSocketAddresses.
+           Previously, the client pool was not shut down, leading to a thread leak.
```
tboam: Copy and paste error.

clockfort:
@gmaretic @ashrayjain 
I sincerely apologize for not seeing this commit last week, but could we not automatically populate this from
- atlas has only the external hostnames in its configs that it can actually talk to the nodes on (aka solely the information it would have had before this commit)
- atlas resolves these external hostnames to external IPs
- atlas does for each external hostname: 
-- describe_ring() (returns all internal IPs)
-- 'select * from system.peers' (returns a node-specific view of all the internal IPs except for this nodes own internal IP)
and thus divines the internal IP of that node. And then knows the external hostname/IP to internal IP mapping.

If you want I can put my money where my mouth is and make a commit that does the above, as I'd prefer we have less configuration / more things to populate from $internalProjectName-container-land and $internalProjectName-configuration-management-service

ashrayjain:
Yea, that could work. We could get the external -> internal ip mapping by looking at the system.local table for each node or something like that.

PR #3042, opened 2018-03-09
Fix broken sanity check

gmaretic:
**Goals (and why)**:
The check can never fail because we were checking within the for loop instead fo creating the map first and checking afterwards.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3042)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -473,7 +473,7 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
 
         RuntimeException ex = new IllegalStateException("Hosts have differing ring descriptions."
                 + " This can lead to inconsistent reads and lost data. ");
-        log.error("Cassandra does not appear to have a consistent ring across all of it's nodes. This could cause us to"
+        log.error("Cassandra does not appear to have a consistent ring across all of its nodes. This could cause us to"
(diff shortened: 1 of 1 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -458,48 +458,49 @@ private void sanityCheckRingConsistency() {
...
-                        SafeArg.of("hosts2", CassandraLogHelper.collectionOfHosts(tokenRangesToHost.get(set2))));
-            }
+        if (tokenRangesToHost.keySet().size() == 1) { // all nodes agree on a consistent view of the cluster. Good.
+            return;
+        }
 
-            CassandraVerifier.logErrorOrThrow(ex.getMessage(), config.ignoreInconsistentRingChecks());
+        RuntimeException ex = new IllegalStateException("Hosts have differing ring descriptions."
+                + " This can lead to inconsistent reads and lost data. ");
+        log.error("QA-86204 {}: The token ranges to host are:\n{}",
```
sandorw: Let's put something appropriately scare in the log.error line, since the current message doesn't mean much. Including the message from the ISE above should do it.

gsheasby: +1 - this is already included as the message param, but putting it in the actual log message would make it more visible.

Note that ex.getMessage is used again later, so we might need to duplicate the string or extract a constant, assuming our intent is for the message to appear directly with other messages in log viewers, rather than as a separate parameter.

PR #3045, opened 2018-03-12
Fix old release note

sandorw:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3045)


sandorw:
Made it really hard to track down when this change happened when both the version and the PR number were wrong...

jeremyk-91:
(also, yeah, the new PR number is correct)

PR #3047, opened 2018-03-12
Adjust CellsSweeper to use dagger-based persistent lock service

dxiao:
internal ticket 111719

because the PersistentLockService was being created in two different places (BackgroundSweeperImpl and CellsSweeper) and their creation logic wasn't completely consistent.

pushing something for internal product.

sooner would be nice.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3047)


dxiao:
note that somebody else is welcome to revise this, I'm just here for when this merges and downstream products need to be fixed.

gsheasby:
Aiming to merge this today, since @dxiao should be around to fix the expected break internally.

PR #3048, opened 2018-03-13
Fix targeted sweep deletion timestamp

sandorw:
**Goals (and why)**:
Avoid deleting values at `Long.MAX_VALUE` to allow compaction of tombstones. Early targeted sweep work deletes at `Long.MAX_VALUE`.

Is this correct? The timestamp bound is exclusive, so all deleted cells should be at timestamps less than the exclusive bound. I assume that ranged tombstones function as normal ones with respect to compaction rules and the highest timestamp winning.

**Priority (whenever / two weeks / yesterday)**: Soon

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3048)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/Mutations.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/Mutations.java
@@ -25,7 +25,7 @@ public final class Mutations {
 
     public static Mutation rangeTombstoneForColumn(byte[] columnName, long maxTimestampExclusive) {
         Deletion deletion = new Deletion()
-                .setTimestamp(Long.MAX_VALUE)
+                .setTimestamp(maxTimestampExclusive)
                 .setPredicate(SlicePredicates.rangeTombstoneForColumn(columnName, maxTimestampExclusive));
 
         return new Mutation().setDeletion(deletion);


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/Mutations.java
```diff
@@ -25,7 +25,7 @@ private Mutations() { }
 
     public static Mutation rangeTombstoneForColumn(byte[] columnName, long maxTimestampExclusive) {
         Deletion deletion = new Deletion()
-                .setTimestamp(Long.MAX_VALUE)
+                .setTimestamp(maxTimestampExclusive)
```
ashrayjain: Should this be + 1?

sandorw: This is an exclusive bound, so it's the timestamp of the write that triggered the targeted sweep, so all swept values will be at a timestamp less than this one. Effectively, maxTimestampExclusive is 1 + the maximum timestamp of any cell that could be swept in this request.

gmaretic: Makes sense

PR #3049, opened 2018-03-13
log latest seq retrieval errors

gokcanozakdag:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3049)

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
@@ -19,9 +19,13 @@ package com.palantir.paxos;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.collect.ImmutableList;
 
 public class PaxosLatestRoundVerifierImpl implements PaxosLatestRoundVerifier {
+    private static final Logger log = LoggerFactory.getLogger(PaxosLatestRoundVerifierImpl.class);
 
     private final ImmutableList<PaxosAcceptor> acceptors;
     private final int quorumSize;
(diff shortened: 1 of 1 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
```diff
@@ -51,7 +55,12 @@ public PaxosQuorumStatus isLatestRound(long round) {
     }
 
     private boolean acceptorAgreesIsLatestRound(PaxosAcceptor acceptor, long round) {
-        return round >= acceptor.getLatestSequencePreparedOrAccepted();
+        try {
+            return round >= acceptor.getLatestSequencePreparedOrAccepted();
+        } catch (Exception e) {
+            log.debug("latest sequence retrieval failed for acceptor: {}", acceptor, e);
```
jeremyk-91: You probably want this to be a higher level - internally most things aren't logged at DEBUG. I realise that if a node is down this could cause a bit of spam, but INFO is probably okay.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
```diff
@@ -51,7 +55,12 @@ public PaxosQuorumStatus isLatestRound(long round) {
     }
 
     private boolean acceptorAgreesIsLatestRound(PaxosAcceptor acceptor, long round) {
-        return round >= acceptor.getLatestSequencePreparedOrAccepted();
+        try {
+            return round >= acceptor.getLatestSequencePreparedOrAccepted();
+        } catch (Exception e) {
+            log.info("latest sequence retrieval failed for acceptor: {}", acceptor, e);
```
tboam: I don't think the PaxosAcceptor has a useful toString so I don't think this is useful to log.  If it is useful it needs to be wrapped in `SafeArgs.of(...)` to conform to spec.

PR #3053, opened 2018-03-13
[Oracle Compaction][Part 7] Configurable Compactions

jeremyk-91:
**Goals (and why)**:
- Avoid a regression and maintain configurability of compaction parameters

- Expose a way to pass config parameters on sleeps to the background compactor

- There is some janky logic to handle the `NOTHING_TO_COMPACT` case. Does that make sense?

**Where should we start reviewing?**: BackgroundCompactor.java

**Priority (whenever / two weeks / yesterday)**: next 2 days please

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3053)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactorConfig.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactorConfig.java
@@ -26,6 +26,10 @@ public interface CompactorConfig {
     long DEFAULT_COMPACT_PAUSE_ON_FAILURE_MILLIS = TimeUnit.SECONDS.toMillis(1800);
     long DEFAULT_COMPACT_PAUSE_MILLIS = TimeUnit.SECONDS.toMillis(10);
 
+    /**
+     * Indicates whether 
+     * @return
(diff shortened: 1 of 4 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/CompactorConfig.java
```diff
@@ -16,13 +16,58 @@
...
+    }
 
-    static CompactorConfig defaultCompactorConfig() {
-        return ImmutableCompactorConfig.builder().inSafeHours(true).build();
+    @Value.Check
+    default void checkIntervalsNonnegative() {
+        Preconditions.checkState(compactPauseOnFailureMillis() >= 0,
+                "Compact pause-on-failure interval must be greater than 0 but found %s", compactPauseOnFailureMillis());
+        Preconditions.checkState(compactPauseMillis() >= 0,
+                "Compact pause interval must be greater than 0 but found %s", compactPauseMillis());
```
jeremyk-91: nit: slightly misleading (must be geq 0, not greater than 0)

PR #3056, opened 2018-03-14
Add logging to confirm address translation is working as intended

gmaretic:
**Goals (and why)**:
See title

Log when we do the translation while refreshing token ranges, log when we have to get a random host because we didn't have an entry in the tokenMap.

Is this enough, is it too spammy?

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3056)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -146,6 +146,8 @@ private InetSocketAddress getAddressForHostThrowUnchecked(String host) {
 
     public InetSocketAddress getAddressForHost(String host) throws UnknownHostException {
         if (config.addressTranslation().containsKey(host)) {
+            log.info("Found address translation mapping for host {}. The InetSocketAddress this host maps to is {}.",
```
sandorw: I'm not sure this is useful - if the translation is required, there will be very clear signal that the translation isn't working, since every cassandra operation will fail, plus this is very high volume.

PR #3057, opened 2018-03-14
Go/leader runtime config

gokcanozakdag:
**Goals (and why)**:
Making `logOnlyOnQuorumFailure` runtime configurable
Created `LeaderRuntimeConfig` which contains `logOnlyOnQuorumFailure`; passed to PaxosLeaderElectionService in a supplier.

`PaxosLeaderElectionService`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3057)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
@@ -22,7 +22,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 @JsonDeserialize(as = ImmutableLeaderRuntimeConfig.class)
-@JsonSerialize(as = ImmutableLeaderConfig.class)
+@JsonSerialize(as = ImmutableLeaderRuntimeConfig.class)
 @Value.Immutable
 public abstract class LeaderRuntimeConfig {
 
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
@@ -18,11 +18,6 @@ package com.palantir.atlasdb.config;
 
 import org.immutables.value.Value;
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-
-@JsonDeserialize(as = ImmutableLeaderRuntimeConfig.class)
-@JsonSerialize(as = ImmutableLeaderRuntimeConfig.class)
 @Value.Immutable
 public abstract class LeaderRuntimeConfig {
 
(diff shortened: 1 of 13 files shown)


timelock-agent/src/main/java/com/palantir/timelock/config/PaxosRuntimeConfiguration.java
```diff
@@ -33,6 +33,12 @@ default long leaderPingResponseWaitMs() {
         return 5000L;
     }
 
+    @JsonProperty("only-log-on-quorum-failure")
```
tboam: Why do we need this on both `PaxosRuntimeConfiguration` and `LeaderRuntimeConfig`?  We could just pass the `PaxosRuntimeConfiguration` around everywhere or even just an anonymous `Supplier<Boolean>`.  If we do stick with `LeaderRuntimeConfig` then I don't think it needs the JSON serializable stuff.

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@JsonDeserialize(as = ImmutableLeaderRuntimeConfig.class)
+@JsonSerialize(as = ImmutableLeaderRuntimeConfig.class)
+@Value.Immutable
+public abstract class LeaderRuntimeConfig {
+
+    @Value.Default
+    public boolean onlyLogOnQuorumFailure() { return true; }
```
jeremyk-91: I'm guessing we plan to make more things configurable here? Otherwise this seems a bit overkill

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.palantir.atlasdb.config;
+
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@JsonDeserialize(as = ImmutableLeaderRuntimeConfig.class)
+@JsonSerialize(as = ImmutableLeaderRuntimeConfig.class)
```
jeremyk-91: nit: if I'm not wrong, not needed (we don't serialize/deserialize these)

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderRuntimeConfig.java
gokcanozakdag: Yes, I also want to open a separate pr to `PaxosLeadershipCreator` for live reloading paxos ping rates, which will be passed in LeaderRuntimeConfig.

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosRuntimeConfiguration.java
gokcanozakdag: Removing the JSON serializable stuff. I am also planning to add paxos ping rates to this class. Thought would be tidier this way.

PR #3062, opened 2018-03-15
Exclude swept time from SweepResults.equals

### LINKED ISSUE #3059: AbstractSweepTaskRunnerTest is flaky
This seems to be failing
```
private void assertEmptyResultWithNoMoreToSweep(SweepResults results) {
        assertTimeSweepStartedWithinDeltaOfSystemTime(results);
        assertThat(results.getTimeInMillis()).isLessThanOrEqualTo(1000L);
        assertThat(results.getTimeInMillis()).isGreaterThanOrEqualTo(0L);
        assertEquals(results, SweepResults.builder()
                .from(SweepResults.createEmptySweepResultWithNoMoreToSweep())
                .timeSweepStarted(results.getTimeSweepStarted())
                .timeInMillis(results.getTimeInMillis())
                .build());
    }
```

Build 9817
```
[00:04:20] com.palantir.atlasdb.keyvalue.dbkvs.DbkvsPostgresTestSuite > com.palantir.atlasdb.keyvalue.dbkvs.DbkvsPostgresSweepTaskRunnerTest.testSweepStrategyNothing FAILED
[00:04:20]     java.lang.AssertionError: expected:<SweepResults{previousStartRow=Optional.empty, nextStartRow=Optional.empty, cellTsPairsExamined=0, staleValuesDeleted=0, minSweptTimestamp=9223372036854775807, timeInMillis=0, timeSweepStarted=1521052988386}> but was:<SweepResults{previousStartRow=Optional.empty, nextStartRow=Optional.empty, cellTsPairsExamined=0, staleValuesDeleted=0, minSweptTimestamp=9223372036854775807, timeInMillis=0, timeSweepStarted=1521052988385}>
```

gsheasby:
**Goals (and why)**: Fixes #3059 

A few tests break because we create a couple of empty results, which
have System.currentTimeMillis as the started time, and then expect them
to equals() each other. Adding Value.Auxiliary excludes this field from
equals().

As a side-effect, the field is also excluded from toString(), but since
this class isn't serialised, we don't mind.

[no release notes]

**Concerns (what feedback would you like?)**: Should we stop using currentTimeMillis in createEmptySweepResults instead?

**Priority (whenever / two weeks / yesterday)**: ASAP, causing tests to flake including external project tests

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3062)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
@@ -21,6 +21,7 @@ import org.immutables.value.Value;
 
 import com.palantir.atlasdb.encoding.PtBytes;
 
+// Note that this class has an Auxiliary field, and therefore serialization might not work as expected
 @Value.Immutable
 public abstract class SweepResults {
 
@@ -57,6 +58,7 @@ public abstract class SweepResults {
     /**
      * Time in milliseconds when we started sweeping this table.
      */
+    @Value.Auxiliary
     public abstract long getTimeSweepStarted();
 
     public long getTimeElapsedSinceStartedSweeping() {
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
@@ -21,7 +21,6 @@ import org.immutables.value.Value;
 
 import com.palantir.atlasdb.encoding.PtBytes;
 
-// Note that this class has an Auxiliary field, and therefore serialization might not work as expected
 @Value.Immutable
 public abstract class SweepResults {
 
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
@@ -65,7 +65,7 @@ public abstract class AbstractSweepTaskRunnerTest extends AbstractSweepTest {
     }
 
     @Test(timeout = 50000)
-    public void testSweepStrategyNothing() throws InterruptedException {
+    public void testSweepStrategyNothing() {
         createTable(SweepStrategy.NOTHING);
         putIntoDefaultColumn("foo", "bar", 50);
         putUncommitted("foo", "bad", 75);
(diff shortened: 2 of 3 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
```diff
@@ -21,6 +21,7 @@
 
 import com.palantir.atlasdb.encoding.PtBytes;
 
+// Note that this class has an Auxiliary field, and therefore serialization might not work as expected
```
jeremyk-91: I think it's fine. We don't specify the JSON serialize/deserialize annotations - and I did a quick test, and if we did the `timeSweepStarted` is still serialized.

Nice! I would have guessed such a feature must exist somewhere, but have never used it

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
```diff
@@ -243,6 +244,7 @@ public void testSweeperFailsHalfwayThroughOnDeleteTable() {
                         .maxCellTsPairsToExamine(DEFAULT_BATCH_SIZE)
                         .build(),
                 nextStartRow);
+        Thread.sleep(1);
```
jeremyk-91: Not sure I understand why we need the `Thread.sleep(1)` here - I get that it would be a failing test for the original impl with a non-Auxiliary start time for sweep, though I think we have the test in `SweepResultsTest` that already does that.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
```diff
@@ -57,6 +58,7 @@
     /**
      * Time in milliseconds when we started sweeping this table.
      */
+    @Value.Auxiliary
```
gmaretic: Awesome find :)

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
gsheasby: Oh cool. @gmaretic and I weren't sure if excluding the field from `toString()` would mess with `JsonSerialize` - thanks for doing the legwork here. I'll remove the comment, as it isn't needed.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/sweep/AbstractSweepTaskRunnerTest.java
gsheasby: Yep, that's why I put it there - I can revert that bit.

gmaretic: Looks to me like it was left from debugging. I'd remove it

PR #2637, opened 2017-11-06
Revert "allow Cassandra KVS task executor to grow..." 2534

gmaretic:
This reverts commit 74c3dae20ef83470b6f74aa9613ebff3dda85835.

**Goals (and why)**:
Fix perf regression.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2637)

PR #2534, opened 2017-10-20
allow Cassandra KVS task executor to grow past poolSize * servers size

jnelson15:
Fixes https://github.com/palantir/atlasdb/issues/2524.

Idea is basically that we should ideally delegate the blocking to the underlying client pools rather than in the KVS executor queue. For example, if we have 3 servers and pool size of 30 and get 90 requests to a single node, most of those will block and we won't be able to run queries against any of the other inactive nodes even though their pools are fully available.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2534)

PR #3064, opened 2018-03-17
Instrument kvs executor + Increase number of threads for CKVS

fsamuel-bs:
**Goals (and why)**: Following up on #2534 and #2637.

**Concerns (what feedback would you like?)**: Should we have `config.maxConnectionBurstSize() * config.servers().size()` all the times? By default, `config.maxConnectionBurstSize() = 100`, so on a 3 Cassandra node cluster we would have 300 threads allocated.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3064)


fsamuel-bs:
If we're afraid of increasing the number of threads, we could just instrument the executor and, using the new metrics, decide if the extra number of threads is needed. This PR increases the corePoolSize we use by, at least, 210 threads per AtlasDB user, so the more careful approach might be more advisable here.

tboam:
@gokcanozakdag this may be interesting to you both for the metrics it adds and the change it makes.  Can you take a look when you return?

gokcanozakdag:
yep, I will take a look @tboam

gmaretic:
@gokcanozakdag is this abandoned in favour of https://github.com/palantir/atlasdb/pull/3416 ?

gokcanozakdag:
We have addressed instrumenting the executor service on #3416; but that pr doesn't increase thread pool size. Looking at the first metrics we get from the field, increasing the thread pool size for cassandra executor service seems unnecessary; but I will be tracking it.

Closing this one out; will open a separate pr if we decide to increase thread pool size.

PR #3065, opened 2018-03-19
Stop using GCM in any circumstances

j-baker:
Also rename esoteric 'cypher' to 'cipher'.

Basically this says that a server or client 'may' reject other ciphers as insecure.
However, both internally and externally we do not do this on either
clients or servers.

We basically move to be the same as https://github.com/palantir/http-remoting/blob/bb2a852ab6be880f79f99d0b9005d5192ad90d05/http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3065)


SerialVelocity:
@j-baker Looks like you will have to fixup the integration tests too. Looks like they are failing because this cipher was removed.

j-baker:
Yeah, there's a bad dropwizard version which forces in the GCM ciphersuite.

sandorw:
ping @j-baker

PR #3069, opened 2018-03-22
Downgrade C* connection failure log to warn

### LINKED ISSUE #3054: If one Cassandra node is down, BackgroundSweeper spams ERROR logs
We require consistency ALL when deleting values. When we have a quorum of nodes up and BackgroundSweeper is enabled, BackgroundSweeper will constantly try to delete and spam ERROR logs while doing so, since Cassandra will throw an UnavailableException.

gsheasby:
**Goals (and why)**: Fixes #3054 (see ticket for discussion, BLUF: The exception is rethrown to the calling service, who can log at error if desired.)

- Error --> warn
- Also, added safe args to one of the log messages.

I wanted to add a note about _which_ exception is thrown, but from a partial investigation it seems we're not consistent in this. Sometimes we catch the C* exception and throw AtlasDbDependencyException, but not always. I deferred fully investigating this to a later time, as the issue on the ticket was fixed anyway.

**Where should we start reviewing?**: One java file

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3069)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
@@ -98,13 +98,17 @@ class CassandraRequestExceptionHandler {
         if (ex instanceof TTransportException
                 && ex.getCause() != null
                 && (ex.getCause().getClass() == SocketException.class)) {
-            log.error(CONNECTION_FAILURE_MSG, numberOfAttempts, ex);
-            String errorMsg =
-                    MessageFormatter.format(CONNECTION_FAILURE_MSG, numberOfAttempts).getMessage();
+            log.warn(CONNECTION_FAILURE_MSG_WITH_EXCEPTION_INFO,
+                    SafeArg.of("numTries", numberOfAttempts),
+                    SafeArg.of("exceptionClass", ex.getClass().getTypeName()),
+                    UnsafeArg.of("exceptionMessage", ex.getMessage()));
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -98,13 +98,17 @@ RequestExceptionHandlerStrategy getStrategy() {
...
+                    UnsafeArg.of("exceptionMessage", ex.getMessage()));
+            String errorMsg = MessageFormatter.format(CONNECTION_FAILURE_MSG, numberOfAttempts).getMessage();
             throw (K) new TTransportException(((TTransportException) ex).getType(), errorMsg, ex);
         } else {
-            log.error("Tried to connect to cassandra {} times.",
-                    SafeArg.of("numTries", numberOfAttempts), ex);
+            log.warn("Tried to connect to cassandra {} times. Exception message was: {} : {}",
+                    SafeArg.of("numTries", numberOfAttempts),
+                    SafeArg.of("exceptionClass", ex.getClass().getTypeName()),
+                    UnsafeArg.of("exceptionMessage", ex.getMessage()));
```
jeremyk-91: In general one advantage of logging the exception here is that you get the (safe) stack-trace, though I guess in these circumstances it isn't normally very useful?

gsheasby: We also log on the first failure (at INFO), so are not losing this information assuming we keep failing for the same reason.

Issue #3070, opened 2018-03-23
Stream store nontransactional write should support customizable backpressure


jeremyk-91:
From `AbstractPersistentStreamStore.java`:

```
    private void storeBlocksFromStream(@Nullable Transaction tx, long id, InputStream stream) throws IOException {
        long blockNumber = 0;

        while (true) {
            byte[] bytesToStore = new byte[BLOCK_SIZE_IN_BYTES];
            int length = ByteStreams.read(stream, bytesToStore, 0, BLOCK_SIZE_IN_BYTES);
            // Store only relevant data if it only filled a partial block
            if (length == 0) {
                break;
            }
            if (length < BLOCK_SIZE_IN_BYTES) {
                // This is the last block.
                storeBlockWithNonNullTransaction(tx, id, blockNumber, PtBytes.head(bytesToStore, length));
                break;
            } else {
                // Store a full block.
                storeBlockWithNonNullTransaction(tx, id, blockNumber, bytesToStore);
            }
            blockNumber++;
        }
    }
```

So if one tries to write a massive stream, this decomposes the stream into blocks of 1 MB, and then sequentially writes block 1 ... block N. In certain cases if you have a good connection, this can cause compaction tasks to build up on Cassandra, resulting in GC pressure and generally increased read and write latencies for other services using a given Cassandra cluster as well.

This issue might potentially only happen if you have a good connection such that serial transactions go through quickly enough, but we have witnessed it in production.

We probably want something like

```
    private void storeBlocksFromStream(@Nullable Transaction tx, long id, InputStream stream) throws IOException {
        long blockNumber = 0;

        while (true) {
            byte[] bytesToStore = new byte[BLOCK_SIZE_IN_BYTES];
            int length = ByteStreams.read(stream, bytesToStore, 0, BLOCK_SIZE_IN_BYTES);
            // Store only relevant data if it only filled a partial block
            if (length == 0) {
                break;
            }
            if (length < BLOCK_SIZE_IN_BYTES) {
                // This is the last block.
                storeBlockWithNonNullTransaction(tx, id, blockNumber, PtBytes.head(bytesToStore, length));
                break;
            } else {
                // Store a full block.
                storeBlockWithNonNullTransaction(tx, id, blockNumber, bytesToStore);
            }
            blockNumber++;
            if (blockNumber % NUM_BLOCKS_TO_WRITE_BEFORE_SLEEP == 0) {
               Thread.sleep(SLEEP_TIME);
            }
        }
    }
[...]

PR #3073, opened 2018-03-23
Stream Store Backoff Between Blocks

jeremyk-91:
**Goals (and why)**:
- Allow a configurable backoff between writing blocks to a stream store
- Allow configuration of the number of blocks to write before triggering the backoff

- Create a `StandardPeriodicBackoffStrategy` that given a number of blocks to write before backoff and a pause interval from live reloadable config, does any necessary backoff for a provided block number.
- Extend AtlasDbRuntimeConfig to include the live reloadable configs mentioned above
- Note that there is a drive by fix to the release notes for 0.79.0 (Oracle compaction should be on develop, not 0.79.0).

- There is a more complicated way of maintaining a tighter bound on the number of blocks written between sleeps in the event of live reloading (keep a counter), but I deemed this unimportant. Does that make sense?
- The intention here was to achieve the change I was looking for without forcing a dev break, though search product / products wanting to make use of this feature will need to regenerate their schemas (see release notes for discussion). Is this the right goal, and does this implementation achieve that?

**Where should we start reviewing?**: StandardPeriodicBackoffStrategy

**Priority (whenever / two weeks / yesterday)**: this week, ideally quickly - we're trying to test this with internal search product.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3073)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/stream/StandardPeriodicBackoffStrategy.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/stream/StandardPeriodicBackoffStrategy.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.stream;
+
+import java.util.function.Supplier;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class StandardPeriodicBackoffStrategy implements StreamStoreBackoffStrategy {
+    private final Supplier<StreamStorePersistenceConfiguration> persistenceConfiguration;
+    private final BackoffMechanism backoff;
+
+    @VisibleForTesting
+    StandardPeriodicBackoffStrategy(Supplier<StreamStorePersistenceConfiguration> persistenceConfiguration,
+            BackoffMechanism backoff) {
+        this.persistenceConfiguration = persistenceConfiguration;
+        this.backoff = backoff;
+    }
+
+    public static StandardPeriodicBackoffStrategy create(
+            Supplier<StreamStorePersistenceConfiguration> persistenceConfiguration) {
+        return new StandardPeriodicBackoffStrategy(persistenceConfiguration, Thread::sleep);
+    }
+
+    @Override
+    public void accept(long blockNumber) {
+        if (persistenceConfiguration.get().writePauseDurationMillis() == 0) {
+            return;
+        }
+
+        if (shouldBackoffBeforeWritingBlockNumber(blockNumber)) {
+            try {
+                backoff.backoff(persistenceConfiguration.get().writePauseDurationMillis());
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                // Preserving uninterruptibility, which is current behaviour.
+            }
+        }
+    }
+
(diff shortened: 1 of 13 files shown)


jeremyk-91:
See #3070

sandorw:
I think regenerating the schemas is fine and agree on not taking in more complications with live reloading the config mid-batch.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/StreamStorePersistenceConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+@JsonSerialize(as = ImmutableStreamStorePersistenceConfiguration.class)
+@JsonDeserialize(as = ImmutableStreamStorePersistenceConfiguration.class)
+@Value.Immutable
+public interface StreamStorePersistenceConfiguration {
+    StreamStorePersistenceConfiguration DEFAULT_CONFIG = ImmutableStreamStorePersistenceConfiguration.builder()
+            .build();
+
+    /**
+     * The number of blocks that a nontransactional storeStream() will store before pausing for
+     * writePauseDurationMillis. For example, if a value of 5 is specified, then storeStream() will
```
fsamuel-bs: I don't think "The number of blocks that a nontransactional storeStream() will store before pausing for writePauseDurationMillis" is accurate. This seems to affect both the transactional case and the non-transactional case. Check "storeBlockWithNonNullTransaction" for details.

docs/source/release_notes/release-notes.rst
```diff
@@ -55,6 +55,24 @@ develop
            An exception is thrown to the service who made the request; this service has the opportunity to log at a higher level if desired.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3069>`__)
 
+    *    - |new|
+         - AtlasDB now supports runtime configuration of throttling for stream stores when streams are written block by block in a non-transactional fashion.
+           Previously, such streams would be written using a separate transaction for each block, though in cases where data volume is high this may still cause load on the key-value-service Atlas is using.
+           Please note that if you wish to use this feature, you will need to regenerate your Atlas schemas and suitably inject the stream persistence configuration into your stream stores.
+           However, if you do not intend to use this feature, no action is required, and your stream stores' behaviour will not be changed.
```
fsamuel-bs: We should highlight here that clients should be aware that enabling this throttling will make their `storeStream` calls take longer, likely generating longer transactions, affecting sweep (immutable ts being stale) and increasing chances of rollback on read (not sure if we expose the new stream before it finished to be written).

I understand the reasons for this solution (it's quick to implement and gives us what we want), just think that we should have caution when recommending this for our consumers.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/StandardPeriodicBackoffStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public void accept(long blockNumber) {
+        if (persistenceConfiguration.get().writePauseDurationMillis() == 0) {
+            return;
+        }
+
+        if (shouldBackoffBeforeWritingBlockNumber(blockNumber)) {
+            try {
+                long sweepPauseMillis = persistenceConfiguration.get().writePauseDurationMillis();
```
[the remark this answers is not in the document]
jeremyk-91: Right. Yeah I copied this from where the sweep configs are.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/StreamStorePersistenceConfiguration.java
jeremyk-91: Ah, yes. Confirming this affects both cases.

jeremyk-91: I'll split the code paths - it doesn't make sense for a transactional case to bother with backing off since writes are buffered in memory.

docs/source/release_notes/release-notes.rst
jeremyk-91: I've changed the solution to only affect the non-transactional case. With that modification, while writing streams may take longer, I believe we avoid lengthening any transactions (sleep is never called as part of a transaction).

jeremyk-91:
Merging as internal smoke-tests passed.

PR #3075, opened 2018-03-26
Tagged Metrics for TimeLock: Clients and Leadership

jeremyk-91:
- [x] Needs smoketest/verification.

**Goals (and why)**:
- Current timelock server-side metrics are an aggregate across all clients, and don't expose leadership information. This makes reading and debugging timelock metrics cleaner and easier. See internal ticket 138.

- Implement a `TaggedMetricsEventInvocationHandler` based off Tritium's `MetricsEventInvocationHandler`, though with somewhat simplified functionality
- Instrument async timelock services (`TimeLockServices`, to be clear) on the server side with two tags: client -> the relevant timelock client, and isCurrentSuspectedLeader -> boolean

- This encodes an assumption that there is only one leader. Though I guess we can change PaxosLeadershipCreator.isCurrentSuspectedLeader() to be a function of client name.
- I tried to preserve the existing metric names, but was there an inadvertent break here?

**Where should we start reviewing?**: TaggedMetricsEventInvocationHandler

**Priority (whenever / two weeks / yesterday)**: not so high, unless TomP thinks otherwise / we encounter serious support issues that are difficult to troubleshoot because the metrics we have aren't clear.

@tpetracca for SA in general
@schlosna for SA / sanity check on event-invocation-handler.
Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3075)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.util;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.codahale.metrics.MetricRegistry;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.tritium.api.functions.BooleanSupplier;
+import com.palantir.tritium.event.AbstractInvocationEventHandler;
+import com.palantir.tritium.event.DefaultInvocationContext;
+import com.palantir.tritium.event.InstrumentationProperties;
+import com.palantir.tritium.event.InvocationContext;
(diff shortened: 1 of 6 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -0,0 +1,129 @@
...
+        markGlobalFailure();
+        if (context == null) {
+            logger.debug("Encountered null metric context likely due to exception in preInvocation: {}",
+                    UnsafeArg.of("exception", cause),
+                    cause);
+            return;
+        }
+
+        String failuresMetricName = MetricRegistry.name(getBaseMetricName(context), failuresMetricName());
+        taggedMetricRegistry.meter(MetricName.builder().safeName(failuresMetricName).build()).mark();
```
tboam: Should the failure metrics also have the tags applied?  (here, line below and in `markGlobalFailure`)

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
```diff
@@ -102,8 +103,10 @@ private static AsyncTimelockService createRawAsyncTimelockService(
...
     }
 
-    private static <T> T instrument(Class<T> serviceClass, T service, String client) {
-        // TODO(nziebart): tag with the client name, when tritium supports it
-        return AtlasDbMetrics.instrument(serviceClass, service, MetricRegistry.name(serviceClass));
+    private <T> T instrument(Class<T> serviceClass, T service, String client) {
+        return AtlasDbMetrics.instrumentWithTaggedMetrics(serviceClass, service, MetricRegistry.name(serviceClass),
+                context -> ImmutableMap.of(
+                        "client", client,
+                        "isCurrentSuspectedLeader", String.valueOf(leadershipCreator.isCurrentSuspectedLeader())));
```
tboam: If the current node is partitioned off from the other nodes there will be a period where 2 nodes report they think they are the leader.  Don't see an easy way around this though.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
jeremyk-91: Yep, good spot - that does make sense.

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
tboam: And potentially something interesting to look for in the metrics.

jeremyk-91:
I've patched an internal timelock implementation to use the new version - and observed that on leader nodes and on non-leader nodes, metrics were tagged appropriately. Furthermore, when a leader election occurred, the metric types being logged switched, which is what we want. Client names were also correctly associated with metrics.

PR #3082, opened 2018-04-02
Use correct logger class for cassandra request exceptions

pkoenig10:
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3082)


gsheasby:
@tboam was there a reason to keep the logs using `CassandraClientPoolImpl` when you extracted this class?

tboam:
I didn't want to break anyone's scripts or logging configuration so left messages using the appender they were using before the refactor.  I wasn't aware of any specific scripts or config but was just trying to minimise the breaks.  I'd be happy to break things if there's a note in the changelog.

pkoenig10:
I was motivated to make this change after trying to find the source of a log line. When the log class doesn't match the class where the log line was written, it makes tracing logs take that much longer.

I will add a changelog entry to this PR.

gsheasby:
Closing in favour of #3092 in order to run CI checks.

PR #3083, opened 2018-04-02
adding sampled logging

gokcanozakdag:
**Goals (and why)**:
To log errors when we call `getLatestSequencePreparedOrAccepted()`. It was depending on a runtime configuration before.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3083)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
```diff
@@ -63,7 +64,7 @@ private boolean acceptorAgreesIsLatestRound(PaxosAcceptor acceptor, long round)
         try {
             return round >= acceptor.getLatestSequencePreparedOrAccepted();
         } catch (Exception e) {
-            if (!onlyLogOnQuorumFailure.get()) {
+            if (isPicked()) {
```
gsheasby: I'd recommend a more descriptive name for the method, such as `shouldLogFailure`, but especially as this is temporary code, it's not a big deal.

PR #3087, opened 2018-04-04
Gozakdag/quorum checker logging

gokcanozakdag:
**Goals (and why)**:
Adding logging to quorum checker to debug a timelock issue.


leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -243,4 +260,8 @@ public static PaxosQuorumStatus getQuorumResult(List<PaxosResponse> responses, i
     private static boolean hasAnyDisagreements(List<PaxosResponse> responses) {
         return responses.stream().anyMatch(response -> !response.isSuccessful());
     }
+
+    private static boolean shouldLogCancellationStatus() {
+        return Math.random() < 0.01;
```
gsheasby: How often do we expect `cancelOutstandingRequestsAfterTimeout` to be called?

PR #3088, opened 2018-04-05
Create the _locks table with a deterministic cfid

### LINKED ISSUE #2583: Multiple schema mutation lock tables
Before timelock, we used to enforce that the lock leader creates the _locks schema mutation lock table if the config has a leader block. 

But now, every node in a HA service believes that they are the lock leader and create a schema mutration lock tables after checking if one already exists, thus presenting a race condition.

This race condition has been hit in the field quite a few times and the solution we have currently is to drop all _lockUUID tables manually and restart the service (hoping it doesnt hit the race again)

This presents a case where manual action is necessary and we can't automatically rollback or move forward.

One possible solution here is to always create _locks table and depend on Cassandra's `createIfNotExist` to ensure multiple tables are not created. Note: currently the lock table is called `_lockUUID`.

gsheasby:
Pushing this up as a proof of concept, to get feedback.
To do:
- [x] catch the error produced when multiple nodes create _locks at once
- [x] clean up the code (see TODO in the production code)
- [x] ensure that starting up with an old locks table or multiple locks tables works correctly

**Goals (and why)**: Fixes #2583

- Generate a cfid based on the keyspace name and the string "_locks"
- Use a cql query to set this cfid

- Do we need to set all of the parameters that were there? I copied them from a test stack, and I'm fairly sure they match the parameters from ColumnFamilyDefinitions.standardCfDef.

**Where should we start reviewing?**: SchemaMutationLockTables

**Priority (whenever / two weeks / yesterday)**: before EOW, I might do some work on this on Monday.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3088)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTables.java
```diff
@@ -63,31 +66,61 @@ public SchemaMutationLockTables(CassandraClientPool clientPool, CassandraKeyValu
...
-        // We replace '-' with '_' as hyphens are forbidden in Cassandra table names.
-        return UUID.randomUUID().toString().replace('-', '_');
-    }
+    private void createTableWithCustomId(TableReference tableRef) throws TException {
+        String internalTableName = CassandraKeyValueServiceImpl.internalTableName(tableRef);
+        String keyspace = config.getKeyspaceOrThrow();
+        String fullTableNameForUuid = keyspace + "." + internalTableName;
+        UUID uuid = UUID.nameUUIDFromBytes(fullTableNameForUuid.getBytes());
+
+        // TODO use parameters as in ColumnFamilyDefinitions.getStandardCfDef
```
tpetracca: I suggest making this simpler by having the method be very schema mutation lock table specific so you don't have to worry about arbitrary column family definitions

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTables.java
```diff
@@ -63,31 +66,61 @@ public SchemaMutationLockTables(CassandraClientPool clientPool, CassandraKeyValu
     }
 
     public TableReference createLockTable() throws TException {
-        return clientPool.runWithRetry(this::createLockTable);
-    }
-
-    private TableReference createLockTable(CassandraClient client) throws TException {
-        String lockTableName = LOCK_TABLE_PREFIX + "_" + getUniqueSuffix();
-        TableReference lockTable = TableReference.createWithEmptyNamespace(lockTableName);
+        TableReference lockTable = TableReference.createWithEmptyNamespace(LOCK_TABLE_PREFIX);
```
tpetracca: Do we still need to have a suffix in the table name for the other schema mutation lock code work?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTables.java
```diff
@@ -63,31 +66,61 @@ public SchemaMutationLockTables(CassandraClientPool clientPool, CassandraKeyValu
...
+                + "    AND gc_grace_seconds = 3600\n"
+                + "    AND max_index_interval = 2048\n"
+                + "    AND memtable_flush_period_in_ms = 0\n"
+                + "    AND min_index_interval = 128\n"
+                + "    AND read_repair_chance = 0.0\n"
+                + "    AND speculative_retry = 'NONE'"
+                + "    AND id = '%s'";
+        CqlQuery query = new CqlQuery(createTableStatement,
+                SafeArg.of("keyspace", keyspace),
+                SafeArg.of("internalTableName", internalTableName),
```
jeremyk-91: If this is general we should use LoggingArgs, though at this point I'm not sure if the schema metadata has been loaded yet. Of course if we follow @tpetracca's suggestion to be specific to the schema lock table (and I think we should), then saying that is safe is of course fine.

gsheasby:
"existing services with an old locks table of the _locks_uuid format will continue to use that table even after an upgrade to an Atlas with the new format, because we won't ever see a need to create a new one. Is that right/does that make sense?" - yep - the function for getting locks tables should return any table starting with `_locks`. I think that we already have integration tests to ensure that this works correctly.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTablesIntegrationTest.java
```diff
@@ -91,33 +104,59 @@ public void multipleSchemaMutationLockTablesObjectsShouldReturnSameLockTables()
...
     @Test
-    public void shouldCreateLockTablesStartingWithCorrectPrefix() throws TException {
+    public void shouldCreateLockTablesWithCorrectName() throws TException {
+        lockTables.createLockTable();
+
+        assertThat(Iterables.getOnlyElement(lockTables.getAllLockTables()).getTablename(), equalTo("_locks"));
+    }
+
+    @Test
+    public void shouldCreateLockTableWithFixedCfId() throws Exception {
```
tboam: Feels like this test is doing a lot of work just to check that we calculate the UUID using `UUID.nameUUIDFromBytes(fullTableName.getBytes())`.  This is testing the implementation rather than the behaviour so I'm not sure of the value.

gsheasby:
@tpetracca @jeremyk-91 this is ready for review again.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTables.java
```diff
@@ -63,31 +67,59 @@ public SchemaMutationLockTables(CassandraClientPool clientPool, CassandraKeyValu
...
-        // We replace '-' with '_' as hyphens are forbidden in Cassandra table names.
-        return UUID.randomUUID().toString().replace('-', '_');
-    }
+    private void createTableWithCustomId(TableReference tableRef) throws TException {
+        String internalTableName = CassandraKeyValueServiceImpl.internalTableName(tableRef);
+        String keyspace = config.getKeyspaceOrThrow();
+        String fullTableNameForUuid = keyspace + "." + internalTableName;
+        UUID uuid = UUID.nameUUIDFromBytes(fullTableNameForUuid.getBytes());
+
+        String createTableStatement = "CREATE TABLE \"%s\".\"%s\" (\n"
```
tpetracca: I'm not sure how much of this are just cassandra defaults (or even our own internal defaults) but the full schema representation I see in a production stack looks like:
```
CREATE TABLE gatekeeper."_locks_706c8369_7883_4add_974c_882172a005fb" (
    key blob,
    column1 blob,
    column2 bigint,
    value blob,
    PRIMARY KEY (key, column1, column2)
) WITH COMPACT STORAGE
    AND CLUSTERING ORDER BY (column1 ASC, column2 ASC)
    AND bloom_filter_fp_chance = 0.1
    AND caching = '{"keys":"ALL", "rows_per_partition":"NONE"}'
    AND comment = ''
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.LeveledCompactionStrategy'}
    AND compression = {}
    AND dclocal_read_repair_chance = 0.1
    AND default_time_to_live = 0
    AND gc_grace_seconds = 3600
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0.0
    AND speculative_retry = 'NONE';
```

gsheasby: So, that's precisely what I had before commit [acb6a36](https://github.com/palantir/atlasdb/pull/3088/commits/acb6a3662aec8d13f3cf05c755250b272b8c9c12).

You said "I suggest making this simpler by having the method be very schema mutation lock table specific so you don't have to worry about arbitrary column family definitions", which I took to mean "get rid of parameters we set for regular tables but don't need to tiny locks table" - maybe I was overzealous/mistaken?

Much of the values are set in [ColumnFamilyDefinitions](https://github.com/palantir/atlasdb/blob/0.80.0/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java#L133).

PR #3090, opened 2018-04-12
Sweep Priority/Blacklist Overrides

jeremyk-91:
- [ ] Smoke test on an actual stack/product

**Goals (and why)**:
- Allow sweep to be configured to target certain tables and/or ignore certain tables to ease debugging work

- Add 2 config parameters which are lists of tables to be prioritised/ignored.
- Modify the table selection algorithm to handle priority and blacklist accordingly.
- Re-run the table selection algorithm if the table one had been sweeping wasn't a priority table and priority tables existed, or if the table one had been sweeping was blacklisted.

- I'm only doing the check on a per-iteration basis, meaning that changes could take a long time if someone has a really wide row. Do I need to make this more invasive?

**Where should we start reviewing?**: I'd start with `sweep.rst` in the docs. For the code itself, probably NextTableToSweepProvider.

**Priority (whenever / two weeks / yesterday)**: whenever. Makes support easier, but obviously not critical.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3090)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProviderTest.java
```diff
@@ -96,6 +96,54 @@ public void calculatorReturnsManyTablesWithHighestPriority_thenProviderReturnsOn
         Assert.assertThat(tableToSweep.get(), anyOf(is(table("table2")), is(table("table3")), is(table("table4"))));
     }
 
+    @Test
+    public void calculatorReturnsNonMaximumForPriorityTable_thenProviderStillSelectsIt() {
+        givenPriority(table("table1"), 1.0);
+        givenPriority(table("table2"), 10_000.0);
+
+        whenGettingNextTableToSweepWithOverrides(ImmutableSweepPriorityOverrideConfig.builder()
```
tboam: Small style note: To keep with the BDD-style tests here I think this would have been better as:

```
givenPriority(table("table1"), 1.0);
givenPriority(table("table2"), 10_000.0);
givenPriorityOverride(table("table1"));

whenGettingNextTableToSweep();
...
```

This makes the test easier to read (the state is described in the 'given' phase) and if we ever change how overrides are specified we only have to change the `givenPriorityOverride` and `givenBlacklisted` methods.

Not essential to change though.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
```diff
@@ -100,10 +135,11 @@ private TableReference getRandomValueFromList(List<TableReference> tables) {
                 ? LoggingArgs.safeTableOrPlaceholder(chosenTable.get()).toString()
                 : "no table";
 
-        log.debug("Chose {} from scores: {}, unsafeScores: {}",
+        log.debug("Chose {} from scores: {}, unsafeScores: {}, overrides: {}",
                 SafeArg.of("chosenTable", chosenTableString),
                 SafeArg.of("scores", safeTableNamesToScore),
-                UnsafeArg.of("unsafeScores", scores));
+                UnsafeArg.of("unsafeScores", scores),
+                SafeArg.of("overrides", sweepPriorityOverrideConfig));
```
tboam: The overrides contains table names so we should treat it as unsafe, or sanitize it as we do above with `safeTableNamesToScore `.  The set of people able to view the safe logs may not be the same as the set of people able to view and edit the config (and hence know the table names).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -245,19 +251,34 @@ SweepOutcome runOnce() {
         return specificTableSweeper.getTxManager().runTaskWithRetry(
                 tx -> {
                     Optional<SweepProgress> progress = specificTableSweeper.getSweepProgressStore().loadProgress();
-                    if (progress.isPresent()) {
+                    SweepPriorityOverrideConfig overrideConfig = sweepPriorityOverrideConfig.get();
+                    if (progress.map(realProgress -> shouldContinueSweepingCurrentTable(realProgress, overrideConfig))
```
tboam: If we stop sweeping a table because of an override then do we need to do any clean up of the SweepProgressStore?  Is it ok for the SweepProgressStore to have the interrupted table and the priority table in it?

Also, since we don't complete the table we don't write the final results to the SweepPriorityStore.  Does this have an impact on the priority calculation in subsequent runs?

tboam: Based on answers to this there may be follow up work or the documentation may need to be updated with the implications of any interruption to a table being swept.

jeremyk-91:
Resuming work on this today.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProviderTest.java
jeremyk-91: OK, agreed that the above works better. First time working with this style of testing - must say the tests are very readable!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
jeremyk-91: Fair enough - I marked it as Safe based on assumptions on our current security model, but 'config view' and 'safe-log view' are indeed arguably different permissions. Will change to Unsafe for now - sanitising is not incredibly hard, but looks to get quite messy.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
jeremyk-91: Yes, agree that the implications of blacklisting/prioritizing tables could be better documented.

Sweep progress store:
We could clear the table, though I'm not sure there's a very significant benefit. Currently, the old entry for sweep progress will remain in the table; however, once the first batch of the new table is cleared it goes away. SweepProgressStore basically only stores 1 value - and supports PUT, GET and DELETE methods. This also means that after the blacklist/priority settings are removed, sweep won't return to where it left off. Will add some clarity in the docs.

Priority:
Currently we don't flush the stats to the sweep priority store, meaning that a non prioritized table will still retain its priority from before the sweep started. This likely means that the probability this table will be picked the next time for sweeping will increase, which seems fine. We could try to do something cleverer (e.g. by trying to scale down the number of writes and time since last sweep by some factor - possibly based on how far the sweep got assuming a uniform distribution across bytes), though I can't think of a good solution in general.

jeremyk-91:
@tboam Should be good for review again. I've added two paragraphs at the end of `Priority Overrides`, addressed the other comments and fixed some build conflicts.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
tboam: Thanks for the details.  I don't think we need to make any code changes, just want us to be clear on the tradeoffs when using this.

PR #3093, opened 2018-04-16
Remove LoggingArgs::isSafeForLogging(TableRef, Cell)

jeremyk-91:
**Goals (and why)**:
- Eliminate a method in `LoggingArgs` that doesn't behave as expected (and really isn't supported by our API), and also constitutes an awkward information leak

- Remove the method
- CellsSweeper now always logs the cell-ts map at Unsafe

- Is further action needed regarding the information leak? The main issue is that if you had a table with a row component `Foo`, you might discover a cell which has the row name `Foo` (which you don't have the right to know exists). It's small and very unlikely, but still...
- CellsSweeper now always logs at unsafe. Is this okay? I think it is, as in the past we would almost always have logged at unsafe anyway (apart from the row name matches row component thing)

**Where should we start reviewing?**: `LoggingArgs` probably, and then `CellsSweeper` which is the only place we use that method

@tboam for SA on the nature of the information hole as well

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3093)


atlasdb-client/src/main/java/com/palantir/atlasdb/logging/LoggingArgs.java
```diff
@@ -157,14 +155,6 @@ public static String safeInternalTableNameOrPlaceholder(String internalTableRefe
                 logArbitrator.isColumnNameSafe(tableReference, columnName));
     }
 
-    public static boolean isSafeForLogging(TableReference tableRef, Cell cell) {
-        String rowName = new String(cell.getRowName(), Charset.defaultCharset());
-        String columnName = new String(cell.getColumnName(), Charset.defaultCharset());
-
-        return logArbitrator.isRowComponentNameSafe(tableRef, rowName)
-                && logArbitrator.isColumnNameSafe(tableRef, columnName);
```
gmaretic: Hilarious 🥁

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CellsSweeper.java
```diff
@@ -93,24 +95,13 @@ public void sweepCells(
...
-        return cellTsPairsToSweep.keySet().stream().allMatch(cell -> LoggingArgs.isSafeForLogging(tableRef, cell));
+    private Arg<String> getLoggingArgForCells(Multimap<Cell, Long> cellTsPairsToSweep) {
+        return UnsafeArg.of("cellTsPairsToSweep", getMessage(cellTsPairsToSweep));
     }
 
-    private <T> String getMessage(Multimap<Cell, Long> cellTsPairsToSweep) {
+    private String getMessage(Multimap<Cell, Long> cellTsPairsToSweep) {
         return cellTsPairsToSweep.entries().stream()
-                .sorted((e1, e2) -> e1.getKey().compareTo(e2.getKey()))
+                .sorted(Comparator.comparing(Map.Entry::getKey))
```
gmaretic: 👍

PR #3097, opened 2018-04-17
[TS] Part 1: Targeted sweep data structures

gmaretic:
**Goals (and why)**:
Implement necessary data structures for persisting targeted sweep info.

TableReferenceAndCell self-explanatory
TargetedSweepMetadata:

- 1 bit for conservative vs. thorough sweep
- 1 bit for combined timestamp rows vs. dedicated rows for transactions with more than Y cells
- 8 bits for the shard number
- 6 bits for the number of the row if it's a dedicated row (max of 64 rows)
- 16 bits unused for future use

Open to naming suggestions, this is just a rough draft. Can the awful bitwise logic be made clearer, or is it enough to document it?

Anywhere

When you have time

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3097)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.api;
+
+import org.immutables.value.Value;
+import com.palantir.common.persist.Persistable;
+
+@Value.Immutable
+public abstract class TargetedSweepMetadata implements Persistable {
+    public abstract boolean conservative();
+    public abstract boolean dedicatedRow();
+    public abstract int shard();
+    public abstract int dedicatedRowNumber();
+
+    public static final Hydrator<TargetedSweepMetadata> BYTES_HYDRATOR = input ->
+            ImmutableTargetedSweepMetadata.builder()
+                    .conservative((input[0] & 0x80) != 0)
+                    .dedicatedRow((input[0] & 0x40) != 0)
+                    .shard((input[0] << 2 | input[1] >> 6) & 0xFF)
+                    .dedicatedRowNumber(input[1] & 0x3F)
+                    .build();
+
+    @Override
+    public byte[] persistToBytes() {
+        byte[] result = new byte[4];
+        if (conservative()) {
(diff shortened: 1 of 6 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+        if (conservative()) {
+            result[0] = (byte) 0x80;
+        }
+        if (dedicatedRow()) {
+            result[0] |= 0x40;
+        }
+        result[0] |= shard() >> 2;
+        result[1] = (byte) (shard() << 6);
+        result[1] |= dedicatedRowNumber();
+        return result;
```
jeremyk-91: Looks correct. I think you can just do `result[1] |= shard() << 6;` since bytes in an array by default are zero.

atlasdb-api/src/test/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadataTest.java
```diff
@@ -0,0 +1,74 @@
...
+            .shard(0)
+            .dedicatedRowNumber(0)
+            .build();
+
+    private static final TargetedSweepMetadata MAX_METADATA = ImmutableTargetedSweepMetadata.builder()
+            .conservative(true)
+            .dedicatedRow(true)
+            .shard(255)
+            .dedicatedRowNumber(63)
+            .build();
```
jeremyk-91: random thought: this makes it sound as though metadatas are comparable, which they aren't really (even if their byte array forms are)

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
```diff
@@ -52,7 +52,7 @@ private static Schema generateSchema() {
             dynamicColumns();
                 columnComponent("timestamp_modulus", ValueType.VAR_LONG);
                 columnComponent("write_index", ValueType.VAR_LONG);
-                value(ValueType.BLOB);
+                value(ValueType.STRING);
```
jeremyk-91: [non-blocking] Consider using `TableReferenceAndCell.class` and having a json-based persister for type safety. Might be pretty nontrivial so wouldn't block on this

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if (dedicatedRow()) {
+            result[0] |= 0x40;
+        }
+        result[0] |= shard() >> 2;
+        result[1] |= shard() << 6;
+        result[1] |= dedicatedRowNumber();
+        return result;
+    }
+}
```
jeremyk-91: General thought about this class: Do we want to add a `@Value.Check` limiting `shard` to 255 and `dedicatedRowNumber` to 63?

jeremyk-91:
also, @sandorw for SA - aware your bandwidth might be limited. I'm happy to take point on these reviews (also good to spread practical TS knowledge).

jeremyk-91:
Also, maybe this can wait till we become really certain of the sweep metadata schema, but it would be useful to document it somewhere in the code (javadoc?)

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
gmaretic: Makes sense

sandorw:
I won't have cycles until I get back next week, so I'll catch up then.

Issue #1412, opened 2017-01-06
Timelock Server repeatedly achieves unnecessary quorum under bounded clock drift assumptions


jeremyk-91:
Current implementation of InvalidatingLeaderProxy as well as the Leader Election Service require us to achieve consensus of a quorum of nodes in the cluster on every timestamp or lock request.

This is fine and strictly speaking required for safety, but in an environment where clock drift can be bounded by some constant `T` (e.g. systems synchronised with GPS, or if you have an atomic clock) it is very wasteful. I can't speak as strongly for Multi-Paxos, but for read-only operations in Raft (i.e. anything lock-related since lock service runs in memory, or individual timestamp operations working within a timestamp bound, which in normal use is probably >99% of operations), there is no reason to do this. 

Specifically, suppose the leader knows an AppendEntriesRPC dispatched at time `D` was successfully committed, and the minimum election timeout is `M`. Then the leader is also guaranteed that he remains the leader up to `D + M` without clock drift; we can set a cutoff at `D + M - T` to account for the bounded clock drift.

* A follower in the majority that acked the AppendEntriesRPC won't become a candidate until after the time it received the RPC plus `M`, which is at least `D + M`.
* A follower that did not ack the AppendEntriesRPC can't become the leader; even if he becomes a candidate at any point, the majority that acked the AppendEntriesRPC won't vote for him (by Raft's election restriction; the follower's logs aren't up to date enough), so he can't win an election.

And as a result, the leader can continue servicing lock requests and timestamp requests that don't touch shared state safely up to the time `D + M - T` without needing to contact a quorum of the other servers.

Preliminary benchmarking of a prototype JSimpleDB implementation of InvalidatingLeaderProxy with this optimisation included suggests a ~4x throughput improvement for the Timelock Server. I realise this may not be suitable for all use cases, so it may be worthwhile making this configurable if we go ahead with this idea.

jboreiko:
Posted this on a internal discussion but thought I would reiterate here.  I would imagine it would be rather difficult to set default values for M and T that would be considered safe in both Palantir controlled cloud environments and on prem hardware both of which may have poor clock synchronization.  In particular I can't imagine we could safely set T to something less than a reasonable M (given M is our minimum recovery time after a legitimate outage).

Some PDS tickets and drift times dredged up by @sarahfbeller:
PDS-47329 - 30 seconds!
PDS-43615 - 2 seconds
PDS-42827 - several hours
PDS-37549 - 39 hours
[...]

PR #3101, opened 2018-04-19
Reduce a number of error and warn messages to lower logging levels.

tboam:
**Goals (and why)**: Reduce the number of messages we log at ERROR and WARN levels.  These have caused unnecessary alarm for users and often obscure their own error messages.

**Implementation Description (bullets)**: I've reduced the logging level for a number of log lines.  In places by quite a lot.  I expect discussion and push back on some of the changes but I want to have the conversations rather than sitting on this.

**Concerns (what feedback would you like?)**:  Where have I reduced the logging level too much.

**Where should we start reviewing?**:  Each change is distinct so anywhere you want.

**Priority (whenever / two weeks / yesterday)**: would be good to merge this next week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3101)


diogoholanda:
Yeah, I would assume that debug level logs would not be available in general to address past problems. They are most likely manually enabled after an event.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -126,7 +126,7 @@ public TokenRangeWritesLogger getTokenRangeWritesLogger() {
             tokenRangeWritesLogger.updateTokenRanges(tokenMap.asMapOfRanges().keySet());
             return servers;
         } catch (Exception e) {
-            log.error("Couldn't grab new token ranges for token aware cassandra mapping!", e);
+            log.warn("Couldn't grab new token ranges for token aware cassandra mapping!", e);
```
jeremyk-91: The main implication of this is that we lose the perf optimisation we have where we talk directly to the correct nodes, right? If that is correct, `warn` is probably fine

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/SweepStatsKeyValueService.java
```diff
@@ -247,7 +247,7 @@ private Runnable createFlushTask() {
                 }
             } catch (Throwable t) {
                 if (!Thread.interrupted()) {
-                    log.error("Error occurred while flushing sweep stats: {}", t, t);
+                    log.warn("Error occurred while flushing sweep stats: {}", t, t);
```
jeremyk-91: 💯 this is bad but not service-haltingly-bad in most cases.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -462,7 +462,7 @@ private static void checkInstallConfig(AtlasDbConfig config) {
                 || config.getSweepReadLimit() != null
                 || config.getSweepCandidateBatchHint() != null
                 || config.getSweepDeleteBatchHint() != null) {
-            log.error("Your configuration specifies sweep parameters on the install config. They will be ignored."
+            log.warn("Your configuration specifies sweep parameters on the install config. They will be ignored."
```
jeremyk-91: This message (even at ERROR) didn't save one of our internal deployments from believing their install config worked. 😞 

I think for the most part I think of logs as something to guide diagnosis or investigation, not so much to be looked at in and of themselves...

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/clock/ClockSkewEvents.java
```diff
@@ -75,7 +75,7 @@ public void clockSkew(String server, long skew, long minRequestInterval, long du
     }
 
     public void clockWentBackwards(String server, long amount) {
-        log.error("The clock for server {} went backwards by {} nanoseconds",
+        log.debug("The clock for server {} went backwards by {} nanoseconds",
```
jeremyk-91: For these 3 messages, I don't think timelock where this is used logs at DEBUG at all. If we still wanted to monitor incidences of clock skew (e.g. as a precursor to our very contentious #1412) it's probably better to keep these at INFO, or change the default log level in timelock to be DEBUG for this logger specifically.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/Blacklist.java
```diff
@@ -70,7 +70,7 @@ private boolean isHostHealthy(CassandraClientPoolingContainer container) {
             container.runWithPooledResource(CassandraUtils.getValidatePartitioner(config));
             return true;
         } catch (Exception e) {
-            log.warn("We tried to add blacklisted host '{}' back into the pool, but got an exception"
+            log.debug("We tried to add blacklisted host '{}' back into the pool, but got an exception"
```
jeremyk-91: Very often this is `TimedOutException` or various Thrift-based things, though I don't know if this one is still useful for debugging?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/AdjustableSweepBatchConfigSource.java
```diff
@@ -87,7 +87,7 @@ public void decreaseMultiplier() {
         // Cut batch size in half, always sweep at least one row.
         reduceBatchSizeMultiplier();
 
-        log.warn("Sweep failed unexpectedly with candidate batch size {},"
+        log.info("Sweep failed unexpectedly with candidate batch size {},"
```
jeremyk-91: probably outside the scope of this PR, but there is a case for a higher log level (WARN I think) if all params are 1.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/Blacklist.java
gsheasby: I would put this to info, as otherwise the information will just be lost at a point when we're trying to diagnose any blacklisting issues.

However, I would like the message to be improved - what does it mean that we're distrusting it further? When will we next attempt to contact this host? This could potentially be done as a separate, small(?) PR.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -117,13 +117,13 @@ RequestExceptionHandlerStrategy getStrategy() {
...
         if (numberOfAttempts > 1) {
-            log.info("Error occurred talking to cassandra. Attempt {} of {}. Exception message was: {} : {}",
+            log.debug("Error occurred talking to cassandra. Attempt {} of {}. Exception message was: {} : {}",
                     SafeArg.of("numTries", numberOfAttempts),
                     SafeArg.of("maxTotalTries", maxTriesTotal),
                     SafeArg.of("exceptionClass", ex.getClass().getTypeName()),
                     UnsafeArg.of("exceptionMessage", ex.getMessage()));
         } else {
-            log.info("Error occurred talking to cassandra. Attempt {} of {}.",
+            log.debug("Error occurred talking to cassandra. Attempt {} of {}.",
```
gsheasby: :-1: this message has been useful for detecting ongoing issues, e.g. when sweep is misbehaving.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
```diff
@@ -85,13 +85,13 @@ private boolean shouldTraceQuery(Set<TableReference> tableRefs) {
...
     private void logFailedCall(Set<TableReference> tableRefs) {
-        log.error("A call to table(s) {} failed with an exception.",
+        log.warn("A call to table(s) {} failed with an exception.",
                 tableRefs.stream().map(TableReference::getQualifiedName).collect(Collectors.joining(", ")));
     }
 
     private void logTraceResults(long duration, Set<TableReference> tableRefs, ByteBuffer recvTrace, boolean failed) {
         if (failed || duration > tracingPrefs.getMinimumDurationToTraceMillis()) {
-            log.error("Traced a call to {} that {}took {} ms. It will appear in system_traces with UUID={}",
+            log.warn("Traced a call to {} that {}took {} ms. It will appear in system_traces with UUID={}",
```
gsheasby: This looks like it could even be an `info` message.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
gsheasby: Is this something that can be/is retried? If so, then the problem is likely to fix itself, possibly without us even realising that it has done so (as I don't think we log when the token ranges were successfully updated, unless it's logged somewhere else).

Again, if `refreshTokenRanges` is only called occasionally, we could consider a separate PR where we periodically retry the mapping update in the event of a failure. This can then be an `info`, as human action will probably not be required to remediate the issue.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -182,13 +182,13 @@ InetSocketAddress getAddressForHost(String host) throws UnknownHostException {
                 .filter(predicate)
                 .collect(Collectors.toSet());
         if (filteredHosts.isEmpty()) {
-            log.error("No hosts match the provided predicate.");
+            log.info("No hosts match the provided predicate.");
```
gsheasby: Suggestion for a quick improvement, while we're here: should also log the predicate, if plausible.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
gsheasby: Separate task: this should possibly be a healthcheck warning rather than a warning hidden in the logs. It's not severe enough to force a user to look at it (e.g. by refusing to start), but is still something that should be fixed, because ignoring it could lead people to believe their config is different to what it actually is. This would cause:
- Best case: less performant than intended sweep
- Average case: lost time in investigating an issue because it's not obvious enough what the problem is
- Worst case: full stack failure (in case the configuration options are set in order to avoid a Cassandra OOM or similar)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/clock/ClockSkewEvents.java
gsheasby: +1 to "change the default log level in timelock to be DEBUG for this logger specifically."

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
gsheasby: retracting this one - we'll log when we error at the last attempt.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
gsheasby: It's retried every 2 minutes, so I'm bumping it to info and adding that detail to the message.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
sandorw: Do we log more visibly on the last attempt if it fails?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -131,7 +131,9 @@ public TokenRangeWritesLogger getTokenRangeWritesLogger() {
             tokenRangeWritesLogger.updateTokenRanges(tokenMap.asMapOfRanges().keySet());
             return servers;
         } catch (Exception e) {
-            log.error("Couldn't grab new token ranges for token aware cassandra mapping!", e);
+            log.info("Couldn't grab new token ranges for token aware cassandra mapping. We will retry in {} seconds.",
```
sandorw: This one's unclear to me - whether it should be warn or info

gsheasby: The exception is thrown when:
* The C* call to get the token ring fails
* Something is borked and we invent a host that is not in the config (shouldn't happen)

Since we retry every 120 seconds by default, it seems very unlikely to me that any engineer will need to take action as a result of the log message, and therefore `INFO` seems appropriate to me.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
gsheasby: Yes, it's a `WARN` if the last attempt fails (see `handleExceptionFromRequest` and `logAndThrowException`.

PR #3104, opened 2018-04-20
[TS] Part2: SweepQueue writers impl

gmaretic:
**Goals (and why)**:
Implement `MultiTableSweepQueueWriter` persisting the sweep queue to the KVS.

Partition the writes by shard, sweep strategy, and timestamp; then persist the information in the two tables defined by `TargetedSweepSchema`.

General feedback. Keep in mind that this is still work in progress

KvsSweepQueueWriter.java

Fast and partial preferred to slow and thorough

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3104)


gsheasby:
Reviewed 26 of 26 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueuePersister.java, line 31 at r1](https://beta.reviewable.io/reviews/palantir/atlasdb/3104#-LAmw7Nc4N6u3L8upHTk:-LAmw7Nc4N6u3L8upHTl:b-vvpxc3) ([raw file](https://github.com/palantir/atlasdb/blob/05cc2ada841b23a5302236495a2d448fba631aa7/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueuePersister.java#L31)):*
> ```Java
>     private final SweepableTimestampsWriter sweepableTimestampsWriter;
> 
>     public KvsSweepQueuePersister(KeyValueService kvs, TargetedSweepTableFactory tableFactory) {
> ```

This is a lot of work for a constructor to do. Would prefer a factory method + private three-arg constructor.

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/MultiTableSweepQueueWriter.java, line 41 at r1](https://beta.reviewable.io/reviews/palantir/atlasdb/3104#-LAmwkgW4tTS0pEcnMvT:-LAmwkgX7N15yQaK5-lq:bxh663b) ([raw file](https://github.com/palantir/atlasdb/blob/05cc2ada841b23a5302236495a2d448fba631aa7/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/MultiTableSweepQueueWriter.java#L41)):*
> ```Java
>     default List<WriteInfo> toWriteInfos(Map<TableReference, ? extends Map<Cell, byte[]>> writes, long timestamp) {
>         return writes.entrySet().stream()
>                 .flatMap(writesEntry -> writesEntry.getValue().keySet().stream()
> ```

This is hard to read, due to the streams "write everything as a lambda" shorthand. If the steps to get from `writes` to `List<WriteInfo>` can be written in a few well-named functions, that would be better (and you might still be able to write it in a streamy way).

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/WriteInfoPartitioner.java, line 73 at r1](https://beta.reviewable.io/reviews/palantir/atlasdb/3104#-LAn4eaS8QhIofD_x6kN:-LAn4eaS8QhIofD_x6kO:b-flw5c0) ([raw file](https://github.com/palantir/atlasdb/blob/05cc2ada841b23a5302236495a2d448fba631aa7/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/WriteInfoPartitioner.java#L73)):*
> ```Java
> 
>     private TableMetadataPersistence.SweepStrategy getStrategyFromKvs(TableReference tableRef) {
>         // todo(gmaretic): fail gracefully if we cannot hydrate? How -- return NOTHING?
> ```

"conservative" might be a better fallback option.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/palantir/atlasdb/3104)*

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsWriter.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private long requiredDedicatedRows(List<WriteInfo> writes) {
+        return 1 + (writes.size() - 1) / MAX_CELLS_DEDICATED;
+    }
+
+    private long tsPartitionFine(long timestamp) {
+        return timestamp / TS_FINE_GRANULARITY;
+    }
+
+    public static long tsMod(long timestamp) {
```
jeremyk-91: confirming this was intentional (seems like package visibility here would do)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueuePersister.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public void initialize() {
+        Schemas.createTablesAndIndexes(TargetedSweepSchema.INSTANCE.getLatestSchema(), kvs);
+    }
+
+    @Override
+    public void enqueue(List<WriteInfo> writes) {
+        sweepableCellsWriter.enqueue(writes);
+        sweepableTimestampsWriter.enqueue(writes);
```
jeremyk-91: What happens if we can write the sweepable cells but then fail when trying to write the sweepable timestamps (e.g. cassandra goes down)?

sandorw: Should probably reverse the order

jeremyk-91:
Also not sure how much work this would be, but is some kind of queue-identifier (shard and boolean) worth wrapping in an object? Kind of like PartitionInfo but without the timestamp

i.e. queueId = shard + isConservative, partitionInfo = queueId + timestamp

Generally through Parts 3 and 4 it seems that those 2 are frequently passed around as a pair

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
```diff
@@ -0,0 +1,47 @@
...
+    private final TableReference tableRef;
+
+    public KvsSweepQueueWriter(KeyValueService kvs, TableReference tableRef) {
+        this.kvs = kvs;
+        this.tableRef = tableRef;
+    }
+
+    @Override
+    public void enqueue(List<WriteInfo> writes) {
+        kvs.put(tableRef, batchWrites(writes), 0L);
```
sandorw: @ashrayjain what are the compaction implications in Cassandra here? **All** writes would be at the same Cassandra timestamp.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDeleterImpl.java
```diff
@@ -52,7 +52,8 @@ public void sweep(Collection<WriteInfo> writes) {
         Map<Cell, Long> deleteTimestampByCell = Maps.newHashMap();
 
         for (WriteInfo write : writes) {
-            deleteTimestampByCell.merge(write.cell(), write.timestampToDeleteAtExclusive(sweeper), Long::max);
+            // todo(gmaretic): since we don't know about deletes, makes it awkward for thorough sweep?
```
sandorw: We should know about them if we plumb the isTombstone flag through on the `WriteInfo`s.

gmaretic: This was added before our discussion so that I don't forget :)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsWriter.java
gmaretic: This will be refactored in a follow-up PR into a utilities class

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TableReferenceAndCell.java
```diff
@@ -32,8 +32,12 @@
 @JsonSerialize(as = ImmutableTableReferenceAndCell.class)
 @Value.Immutable
 public abstract class TableReferenceAndCell implements Persistable {
-    abstract TableReference tableRef();
-    abstract Cell cell();
+    public abstract TableReference tableRef();
+    public abstract Cell cell();
```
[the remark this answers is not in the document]
gmaretic: Yes, but in a separate PR since it will require quite a few changes at this point

gmaretic:
The suggestion of bundling together shard and sweep strategy makes sense. But will do it in a separate PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/WriteInfoPartitioner.java
```diff
@@ -66,18 +80,22 @@ private PartitionInfo getPartitionInfo(WriteInfo write) {
...
     }
 
     private TableMetadataPersistence.SweepStrategy getStrategyFromKvs(TableReference tableRef) {
-        // todo(gmaretic): fail gracefully if we cannot hydrate? How -- return NOTHING?
-        return TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(kvs.getMetadataForTable(tableRef)).getSweepStrategy();
+        try {
+            return TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(kvs.getMetadataForTable(tableRef)).getSweepStrategy();
+        } catch (Throwable th) {
+            log.warn("Failed to obtain sweep strategy for table {}. Assuming sweep strategy is CONSERVATIVE.",
+                    UnsafeArg.of("tableRef", tableRef), th);
```
jeremyk-91: can we `LoggingArgs` the table ref? I guess there's not much point since the metadata is unloadable though

gmaretic:
A few remaining things that I will address in follow-up PRs to reduce merge conflicts

PR #3108, opened 2018-04-23
[TS] Part 3: Targeted Sweep Queue Progress Table + Class

gmaretic:
**Goals (and why)**:
Implement the progress table for sweep queue.

The table stores the last swept timestamp for each shard. It also stores the information about the number of shards we are using.

Just using a row of this table to store the number of shard information is hacky, but does not seem worth it to have a separate table just for that. Alternatively, we could create the new rows eagerly, but that is potentially a lot of CAS operations at once.

Also, the treatment of edge cases with CAS failing etc is open to discussion.

`KvsSweepProgressQueueTest`

Not urgent, but small (ignoring generated code)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3108)


atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
```diff
@@ -75,6 +75,19 @@ private static Schema generateSchema() {
...
+        schema.addTableDefinition("sweepProgressPerShard", new TableDefinition() {{
+            javaTableName("SweepShardProgress");
+            allSafeForLoggingByDefault();
+            rowName();
+                hashFirstRowComponent();
+                rowComponent("shard", ValueType.VAR_LONG);
+                rowComponent("sweep_conservative", ValueType.BLOB);
+            columns();
+                column("value", "v", ValueType.VAR_LONG);
+            sweepStrategy(TableMetadataPersistence.SweepStrategy.NOTHING);
```
gsheasby: It seems to me like we _would_ want to sweep this table.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.schema.generated.SweepShardProgressTable;
+import com.palantir.atlasdb.schema.generated.TargetedSweepTableFactory;
+import com.palantir.util.PersistableBoolean;
+
+public class KvsSweepQueueProgress {
+    private static final TableReference TABLE_REF = TargetedSweepTableFactory.of()
+            .getSweepShardProgressTable(null).getTableRef();
+    private static final int NUMBER_OF_SHARDS_INDEX = 1024;
```
gsheasby: name could be clearer. Do you mean "number of shards in index"? Or something else? `SHARDS_IN_INDEX`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final long CAS_TIMESTAMP = 1L;
+
+    private final KeyValueService kvs;
+
+    public KvsSweepQueueProgress(KeyValueService kvs) {
+        this.kvs = kvs;
+    }
+
+    public long numberOfShards() {
+        return getOrInitializeTo(NUMBER_OF_SHARDS_INDEX, true, 1L);
```
gsheasby: Strong suggestion: use an enum rather than bool, for readability and to reduce mental mapping.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public void updateNumberOfShards(int newNumber) {
+        Preconditions.checkArgument(newNumber <= 128);
+        increaseValue(NUMBER_OF_SHARDS_INDEX, true, newNumber);
+    }
+
+    public long lastSweptTimestamp(int shard, boolean conservative) {
```
gsheasby: I know it's out of fashion to use verbs, but `getLastSweptTimestamp` would be clearer here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
gsheasby: Oh, this is a special index for saying how many shards there are? -_-

If we must do that rather than have a new table, can we use `-1`? `SHARD_COUNT_INDEX`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private long initializeShard(int shard, boolean conservative, long initialVale) {
+        SweepShardProgressTable.Value colVal = SweepShardProgressTable.Value.of(initialVale);
+        CheckAndSetRequest initializeRequest = CheckAndSetRequest
+                .newCell(TABLE_REF, cellForShard(shard, conservative), colVal.persistValue());
+        try {
+            kvs.checkAndSet(initializeRequest);
+            return initialVale;
+        } catch (CheckAndSetException e) {
+            return getValue(getEntry(shard, conservative));
```
gsheasby: Do we care enough about this case to log it?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    private void increaseValue(int shard, boolean conservative, long newValue) {
+        long oldVal = getValue(getEntry(shard, conservative));
+        SweepShardProgressTable.Value colValNew = SweepShardProgressTable.Value.of(newValue);
+
+        while (oldVal < newValue) {
```
gsheasby: Again, do we care about the case where `oldVal` is already greater than `newValue`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+                .newCell(TABLE_REF, cellForShard(shard, conservative), colVal.persistValue());
+        try {
+            kvs.checkAndSet(initializeRequest);
+            return initialVale;
+        } catch (CheckAndSetException e) {
+            return getValue(getEntry(shard, conservative));
+        }
+    }
+
+    private void increaseValue(int shard, boolean conservative, long newValue) {
```
gsheasby: The value stored in the database after this call might be more than `newValue`. Do we want to return the last known value, so that any consequences of this can be handled by the caller? I can't tell this by looking at this class in isolation, but it's something to bear in mind as we continue building this.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
```diff
@@ -0,0 +1,122 @@
...
+        progress.numberOfShards();
+        progress.updateNumberOfShards(128);
+        assertThat(progress.numberOfShards()).isEqualTo(128);
+    }
+
+    @Test
+    public void attemptingToDecreaseNumberOfShardsIsNoop() {
+        progress.numberOfShards();
+        progress.updateNumberOfShards(64);
+        progress.updateNumberOfShards(32);
```
gsheasby: No exception in this case?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public void updatingTimestampForOneConsistencyDoesNotAffectOther() {
+        assertThat(progress.lastSweptTimestamp(10, true)).isEqualTo(0L);
+        assertThat(progress.lastSweptTimestamp(10, false)).isEqualTo(0L);
+        progress.updateLastSweptTimestamp(10, true, 128L);
+        assertThat(progress.lastSweptTimestamp(10, false)).isEqualTo(0L);
+
+        progress.updateLastSweptTimestamp(10, false, 32L);
```
gsheasby: Does it make sense to have different conservative and thorough timestamps per shard?

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
gsheasby: Upon re-reading the design doc: or maybe cleaning these tables up will come later?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
jeremyk-91: Depends on the implementation - but we're already treating them as separate queues (so we have 2 queues per shard) based on the timestamps writer in part 2, so I think it makes sense.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
```diff
@@ -0,0 +1,122 @@
...
+    public void attemptingToDecreaseNumberOfShardsIsNoop() {
+        progress.numberOfShards();
+        progress.updateNumberOfShards(64);
+        progress.updateNumberOfShards(32);
+        assertThat(progress.numberOfShards()).isEqualTo(64);
+    }
+
+    @Test
+    public void increasingNumberOfShardsAbove128Throws() {
+        progress.numberOfShards();
```
jeremyk-91: nit: this line seems a bit random; if it's only for setup then it could make sense being called in a `@Before`. Nothing major though

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
jeremyk-91: probably not. Maybe the method should be named `increaseToAtLeast` or something like that?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
jeremyk-91: I like `SHARD_COUNT_INDEX`.

Heads up: if you change the magic value to `-1`, then you'll need to declare the typing as `VAR_SIGNED_LONG`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -0,0 +1,116 @@
...
+                SweepShardProgressTable.SweepShardProgressNamedColumn.VALUE.getShortName());
+    }
+
+    private long getValue(Map<Cell, Value> entry) {
+        SweepShardProgressTable.Value value = SweepShardProgressTable.Value.BYTES_HYDRATOR.hydrateFromBytes(
+                Iterables.getOnlyElement(entry.values()).getContents());
+        return value.getValue();
+    }
+
+    private long initializeShard(int shard, boolean conservative, long initialVale) {
```
jeremyk-91: nit: `initialValue`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
gmaretic: Done in future PR

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
gmaretic: Subject to change, depending on how we implement the control over this parameter. I can see throwing if we set it as an install config, but probably want to WARN and continue using the old value still.

gmaretic:
Addressed the comments that have not already incidentally been fixed on other branches. If you're good with these changes and have no further comments, should be good to merge

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
gmaretic: Changed this on another branch, no longer necessary to get progress to initialize first

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueProgress.java
```diff
@@ -85,32 +88,33 @@ private long getValue(Map<Cell, Value> entry) {
...
             try {
                 kvs.checkAndSet(casRequest);
-                return;
+                return newValue;
             } catch (CheckAndSetException e) {
-                oldVal = getValue(getEntry(shard, conservative));
+                oldVal = getValue(getEntry(shardAndStrategy));
             }
         }
+        return  oldVal;
```
gsheasby: OCD nit: double space here

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepProgressQueueTest.java
```diff
@@ -63,60 +70,60 @@ public void increasingNumberOfShardsAbove128Throws() {
 
     @Test
     public void canReadInitialSweptTimestamp() {
-        assertThat(progress.lastSweptTimestamp(10, true)).isEqualTo(0L);
+        assertThat(progress.lastSweptTimestamp(ShardAndStrategy.conservative(10))).isEqualTo(INITIAL_TIMESTAMP);
```
gsheasby: nit: `CONSERVATIVE_TEN`

PR #3116, opened 2018-04-25
Include Cassandra semi-break in release notes

gsheasby:
Existing installations are not broken, and internally we already require later versions of Cassandra 2.2, so this isn't strong enough for a USER-BREAK.

**Where should we start reviewing?**: release-notes.rst

**Priority (whenever / two weeks / yesterday)**: today, tiny PR

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3116)

### DIFF
--- a/docs/source/services/key_value_services/index.rst
+++ b/docs/source/services/key_value_services/index.rst
@@ -19,9 +19,9 @@ This matrix details the current state of available Key Value Services supported 
 
     *    - :ref:`Cassandra KVS <cassandra-configuration>`
          - Supported
-         - 2.0+
+         - 2.1.15+, 2.2.5+, 3.0.3+, 3.2+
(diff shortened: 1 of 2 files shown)


docs/source/services/key_value_services/index.rst
```diff
@@ -19,9 +19,9 @@ This matrix details the current state of available Key Value Services supported
 
     *    - :ref:`Cassandra KVS <cassandra-configuration>`
          - Supported
-         - 2.0+
+         - 2.1.15+, 2.2.5+, 3.0.3+, 3.2+
```
tpetracca: has anyone ever actually tested atlas against 3.X? feels weird to declare support for it.
Practically speaking we only support 2.2.9+

PR #3117, opened 2018-04-25
adding start time to slow log messages

gokcanozakdag:
**Goals (and why)**:
Adding start time to kvs-slow-log messages, for easier debugging.

- Is there a cleaner way to do this?
- There may be a slight difference between start time we log, and actual start time in this implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3117)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ProfilingCassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ProfilingCassandraClient.java
@@ -65,15 +65,17 @@ public class ProfilingCassandraClient implements AutoDelegate_CassandraClient {
             throws InvalidRequestException, UnavailableException, TimedOutException, TException {
         int numberOfKeys = keys.size();
         int numberOfColumns = predicate.slice_range.count;
+        long startTime = System.currentTimeMillis();
 
         return KvsProfilingLogger.maybeLog(
                 (KvsProfilingLogger.CallableCheckedException<Map<ByteBuffer, List<ColumnOrSuperColumn>>, TException>)
                         () -> client.multiget_slice(kvsMethodName, tableRef, keys, predicate, consistency_level),
-                (logger, timer) -> logger.log("CassandraClient.multiget_slice({}, {}, {}, {}) on kvs.{} took {} ms",
+                (logger, timer) -> logger.log("CassandraClient.multiget_slice({}, {}, {}, {}) at time {}, on kvs.{} took {} ms",
                         LoggingArgs.tableRef(tableRef),
                         LoggingArgs.rowCount(numberOfKeys),
                         LoggingArgs.columnCount(numberOfColumns),
                         SafeArg.of("consistency", consistency_level.toString()),
+                        LoggingArgs.startTimeMillis(startTime),
                         SafeArg.of("kvsMethodName", kvsMethodName),
                         LoggingArgs.durationMillis(timer)),
                 (logger, rowsToColumns) -> logger.log("and returned {} cells in {} rows with {} bytes",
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ProfilingCassandraClient.java
```diff
@@ -156,24 +166,30 @@ public CASResult cas(TableReference tableReference,
...
             TException {
+        long startTime = System.currentTimeMillis();
+
         return KvsProfilingLogger.maybeLog(
                 (KvsProfilingLogger.CallableCheckedException<CqlResult, TException>)
                         () -> client.execute_cql3_query(cqlQuery, compression, consistency),
                 (logger, timer) -> cqlQuery.logSlowResult(logger, timer),
-                (logger, cqlResult) -> logger.log("and returned {} rows",
-                        SafeArg.of("numRows", cqlResult.getRows().size())));
+                (logger, cqlResult) -> logger.log("and returned {} rows, at time{}",
```
tboam: need a space between "time" and "{}"

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/ProfilingKeyValueService.java
```diff
@@ -367,12 +388,15 @@ public boolean isInitialized() {
...
                 // Log differently in maintenance mode - if a compactInternally is slow this might be bad in normal
                 // operational hours, but is probably okay in maintenance mode.
-                logger.log("Call to KVS.compactInternally (in maintenance mode) on table {} took {} ms.",
-                        LoggingArgs.tableRef(tableRef), LoggingArgs.durationMillis(stopwatch));
+                logger.log("Call to KVS.compactInternally (in maintenance mode) at time {}, on table {} took {} ms.",
+                        LoggingArgs.startTimeMillis(startTime),
+                        LoggingArgs.tableRef(tableRef),
+                        LoggingArgs.durationMillis(stopwatch));
             } else {
                 logTimeAndTable("compactInternally", tableRef);
```
tboam: need to pass the startTime into this method from `compactInternally` as `logTimeAndTable` will get a startTime only when the logger is called.

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/ProfilingKeyValueService.java
```diff
@@ -77,41 +77,54 @@ private ProfilingKeyValueService(KeyValueService delegate) {
...
             RangeRequest range) {
+        long startTime = System.currentTimeMillis();
         return (logger, stopwatch) ->
-                logger.log("Call to KVS.{} on table {} with range {} took {} ms.",
+                logger.log("Call to KVS.{} at time {}, on table {} with range {} took {} ms.",
                         LoggingArgs.method(method),
+                        LoggingArgs.startTimeMillis(startTime),
                         LoggingArgs.tableRef(tableRef),
                         LoggingArgs.range(tableRef, range),
                         LoggingArgs.durationMillis(stopwatch));
```
tboam: Does `logCellResultSize` need to include start time?  Could be helpful.

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/ProfilingKeyValueService.java
tboam: I think the BiConsumers used normally are fine (you call the method, it gets the start time and returns you the BiConsumer with a closure over the start time), it is just here where we call `logTimeAndTable` inside of a BiConsumer (to check maintenance mode) where `logTimeAndTable` is called too late.

Interestingly I think this method may be broken because nothing actually invokes the BiConsumer returned by logTimeAndTable so no logging happens.  This PR should also fix this.  Easiest method is probably inlining the contents of `logTimeAndTable` in this one invokation.

PR #3121, opened 2018-04-26
Refresh okHttp client after exceptions in a row

gokcanozakdag:
**Goals (and why)**:
- We are observing an issue causing timelock follower nodes not to receive any request from leader. Root cause seems to be corrupted okHttp clients as a result of frequent interrupts in `QuorumChecker`. Refreshing the client after detecting a row of exceptions should be a feasible workaround
- Wrapping the client with a `ExceptionCountingRefreshingClient`
- `ExceptionCountingRefreshingClient`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3121)

### DIFF
--- a/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
+++ b/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
@@ -74,7 +74,7 @@ public class ExceptionCountingRefreshingClient implements Client {
             try {
                 currentClient = refreshingSupplier.get();
             } catch (RuntimeException e) {
-                log.warn("An error occurred whilst trying to re-create an OkHttpClient, because {} exceptions have been thrown in a row by old client."
+                log.warn("An error occurred whilst trying to re-create an OkHttpClient."
                                 + " Continuing operation with the old client...",
                         SafeArg.of("count", currentCount),
                         e);
--- a/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
+++ b/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.http;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.palantir.logsafe.SafeArg;
+
+import feign.Client;
+import feign.Request;
+import feign.Response;
+
+public class ExceptionCountingRefreshingClient implements Client {
+    private static final Logger log = LoggerFactory.getLogger(ExceptionCountingRefreshingClient.class);
+
+    private static final long DEFAULT_EXCEPTION_COUNT_BEFORE_REFRESH = 10_000L;
+
+    private final Supplier<Client> refreshingSupplier;
+    private final long exceptionCountBeforeRefresh;
+    private final AtomicLong counter;
+
+    private volatile Client currentClient;
+
+    @VisibleForTesting
+    ExceptionCountingRefreshingClient(Supplier<Client> baseClientSupplier,
+            long exceptionCountBeforeRefresh) {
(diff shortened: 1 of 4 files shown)


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
```diff
@@ -0,0 +1,85 @@
...
+import com.palantir.logsafe.SafeArg;
+
+import feign.Client;
+import feign.Request;
+import feign.Response;
+
+public class ExceptionCountingRefreshingClient implements Client {
+    private static final Logger log = LoggerFactory.getLogger(ExceptionCountingRefreshingClient.class);
+
+    private static final long DEFAULT_EXCEPTION_COUNT_BEFORE_REFRESH = 10_000L;
```
tboam: How did you arrive at this number?  How long into the 'silent' issue would we be before we recreated the client?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private Client delegate() {
+        long currentCount = counter.get();
+        if (currentCount >= exceptionCountBeforeRefresh && counter.compareAndSet(currentCount, 0)) {
+            log.info("Creating a new Feign client, as {} exceptions have been thrown in a row by old client",
+                    SafeArg.of("count", currentCount));
+            try {
+                currentClient = refreshingSupplier.get();
+            } catch (RuntimeException e) {
+                log.warn("An error occurred whilst trying to re-create an OkHttpClient, because {} exceptions have been thrown in a row by old client."
```
tboam: The second part of the log message "because {} exceptions have been thrown in a row by old client" makes it sound like the Exceptions thrown by the old client are the reason we can't make a new one.  I'd just delete this part of the message and the counter as it is logged above anyway.

jeremyk-91: nit: The exceptions didn't necessarily cause the error in re-creating the client (I know you intend "an error occurred... while trying to (re-create because {} exceptions)", but to me at least it reads the other way). You already have the message above though so this probably isn't needed?

atlasdb-feign/src/test/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClientTest.java
```diff
@@ -0,0 +1,148 @@
...
+import feign.Request;
+import feign.Response;
+
+public class ExceptionCountingRefreshingClientTest {
+    private Supplier<Client> clientSupplier = (Supplier<Client>) mock(Supplier.class);
+    private Client client = mock(Client.class);
+
+    private Request request = Request.create("POST", "atlas.palantir.pt", ImmutableMap.of(), null, null);
+    private Response response = Response.create(204, "no content", ImmutableMap.of(), new byte[0]);
+    private Request.Options options = new Request.Options(1, 1);
```
jeremyk-91: nit: These 3 look like they're constants, so they could probably be `static final`

atlasdb-feign/src/test/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClientTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Test
+    public void requestsContinueWithOldClientIfDelegateSupplierThrows() throws IOException {
+        when(clientSupplier.get()).thenReturn(client);
+        when(client.execute(request, options)).thenThrow(CLIENT_EXCEPTION);
+
+        Client refreshingClient = new ExceptionCountingRefreshingClient(clientSupplier, EXCEPTION_COUNT_BEFORE_REFRESH);
+        callExecute(refreshingClient,1);
+
+        when(clientSupplier.get()).thenThrow(new IllegalStateException("bad"));
```
jeremyk-91: I _think_ this is equivalent to writing one line at the beginning `when(clientSupplier.get()).thenReturn(client).thenThrow(new ISE("bad"));`. This could help us avoid having to do `numberOfExecutions - 1` calls in line 98, which requires looking backwards to figure out why there's a `- 1` there.

atlasdb-feign/src/test/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClientTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private void callExecute(Client client, int numberOfExecutions) throws IOException {
+        for (int i = 0; i < numberOfExecutions; i++) {
+            try {
+                client.execute(request, options);
+            } catch (RuntimeException e) {}
+        }
+    }
+}
```
jeremyk-91: nit: newline please

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
```diff
@@ -0,0 +1,85 @@
...
+public class ExceptionCountingRefreshingClient implements Client {
+    private static final Logger log = LoggerFactory.getLogger(ExceptionCountingRefreshingClient.class);
+
+    private static final long DEFAULT_EXCEPTION_COUNT_BEFORE_REFRESH = 10_000L;
+
+    private final Supplier<Client> refreshingSupplier;
+    private final long exceptionCountBeforeRefresh;
+    private final AtomicLong counter;
+
+    private volatile Client currentClient;
```
jeremyk-91: Nice, I think you do need this (calls can go through multiple threads but this is only written in one, for a given refresh).

PR #3123, opened 2018-04-26
[TS] Part 5: Get WriteInfos To Tombstone For Timestamp Partition

gmaretic:
**Goals (and why)**:
Once we have a partition, we read the relevant entries from SweepableCellsTable, then construct a list of WriteInfos to use for creating the ranged tombstones

We use getRowsColumnRange to get all columns, then process each. In case of dedicated rows, we do another range scan to read them as well. Also fixed a horrible metadata bug (unfortunately BitSet solution is not great either :/)

Are the range scans correct? In particular, the ColumnRangeSelection that we use (it does not need to be a tight bound, but we must not miss anything). What about the batch hint -- what is reasonable?
Note that I plan to do a bit more refactoring and adding more tests -- feel free to suggest anything in particular you'd like to see :) 

`SweepableCellsReader.java`

Sooner better than later, but not blocked per se

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3123)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private boolean isReferenceToDedicatedRows(SweepableCellsTable.SweepableCellsColumn col) {
+        return col.getWriteIndex() < 0;
+    }
+
+    private void populateFromDedicatedRows(SweepableCellsTable.SweepableCellsRow row,
+            SweepableCellsTable.SweepableCellsColumn col, Map<TableReferenceAndCell, Long> results) {
+        TargetedSweepMetadata metadata = TargetedSweepMetadata.BYTES_HYDRATOR.hydrateFromBytes(row.getMetadata());
+        long partitionFine = row.getTimestampPartition();
+        int numberOfDedicatedRows = (int) -col.getWriteIndex();
```
gsheasby: It occurs to me that we'll have similar logic in reverse for when we're writing to dedicated rows. That logic would ideally live in the same class/layer. This class might then end up doing little more than delegating to one of two readers.

gsheasby: You wanted a test suggestion: we should have tests to verify that when you write something to the queue and then read it back, you get the same thing you wrote. That's more important than the above suggestion.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/WriteInfo.java
```diff
@@ -30,10 +30,14 @@
...
+    static WriteInfo of(TableReferenceAndCell tableRefCell, long timestamp) {
         return ImmutableWriteInfo.builder()
-                .tableRefCell(TableReferenceAndCell.of(tableRef, cell))
+                .tableRefCell(tableRefCell)
                 .timestamp(timestamp)
                 .build();
     }
+
+    static WriteInfo of(TableReference tableRef, Cell cell, long timestamp) {
+        return WriteInfo.of(TableReferenceAndCell.of(tableRef, cell), timestamp);
```
gsheasby: If `WriteInfo` = `TableReference` + `Cell` + `timestamp`, maybe `TableReferenceAndCell` is a `WriteLocation`? Or just a `Datapoint`?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReadWriteTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void canReadMultipleEntriesInSingleShardSameTransactionNotDedicated() {
+        List<WriteInfo> writes = writeToUniqueCellsInSameShard(writer, TS, 10, true);
+        ShardAndStrategy fixedShardAndStrategy = conservative(WriteInfoPartitioner.getShard(writes.get(0)));
+        assertThat(writes.size()).isEqualTo(10);
+        assertThat(reader.getLatestWrites(TS_REF, fixedShardAndStrategy)).hasSameElementsAs(writes);
+    }
+
+    @Test
+    public void canReadMultipleEntriesInSingleShardSameTransactionDedicated() {
+        List<WriteInfo> writes = writeToUniqueCellsInSameShard(writer, TS, 257, true);
```
gsheasby: Doesn't the premise of this test depend on `257` being above the limit of dedicated writes? Is there a way to assert that the thing we wrote was a dedicated write? If not, is it nice to use "threshold + 1" rather than 257 for the number of cells?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -36,30 +38,32 @@ void checkShardSize() {
 
     @Value.Check
     void checkRowNumber() {
-        Preconditions.checkArgument(dedicatedRowNumber() >= 0 && dedicatedRowNumber() <= 63,
-                "Dedicated row number must be between 0 and 63 inclusive, but it is %s.", dedicatedRowNumber());
+        Preconditions.checkArgument(dedicatedRowNumber() >= 0 && dedicatedRowNumber() < MAX_DEDICATED_ROWS,
+                "Dedicated row number must be between 0 and %s inclusive, but it is %s.",
```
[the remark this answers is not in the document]
gmaretic: I changed the lower or equal to lower

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
```diff
@@ -0,0 +1,139 @@
...
+    private void addIfMaxForCell(long ts, TableReferenceAndCell refAndCell, Map<TableReferenceAndCell, Long> result) {
+        result.merge(refAndCell, ts, Math::max);
+    }
+
+    private static ColumnRangeSelection allPossibleColumns() {
+        byte[] startCol = SweepableCellsTable.SweepableCellsColumn.of(0L, -TargetedSweepMetadata.MAX_DEDICATED_ROWS)
+                .persistToBytes();
+        byte[] endCol = SweepableCellsTable.SweepableCellsColumn.of(SweepQueueUtils.TS_FINE_GRANULARITY + 1, 0)
+                .persistToBytes();
+        return new ColumnRangeSelection(startCol, endCol);
```
jeremyk-91: I think this covers all the columns we need to. The first component ranges from `0` to `TS_FINE_GRANULARITY - 1` (both inclusive); the second component ranges from `-MAX_DEDICATED_ROWS` to `MAX_DEDICATED_ROWS` (both inclusive).

There are tighter bounds for endCol (I think you can actually specify `SweepQueueUtils.TS_FINE_GRANULARITY - 1, TargetedSweepMetadata.MAX_DEDICATED_ROWS + 1`) though they shouldn't affect anything. I'd actually suggest taking `TS_FINE_GRANULARITY` straight (i.e. without the `+ 1`) for readability.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
```diff
@@ -0,0 +1,139 @@
...
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.api.RowColumnRangeIterator;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.api.TableReferenceAndCell;
+import com.palantir.atlasdb.keyvalue.api.TargetedSweepMetadata;
+import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.schema.generated.SweepableCellsTable;
+import com.palantir.atlasdb.schema.generated.TargetedSweepTableFactory;
+
+public class SweepableCellsReader {
```
jeremyk-91: Fairly confident reading this and cross-checking with the `SweepableCellsWriter` that it is correct. There is (expectedly and justifiably) a fair bit of coupling with the `SweepableCellsWriter` and its implementation; I think we can try to reduce some of that. It might be worth explicitly defining some of the conventions being used (e.g. wrt using `false` and `0` in `getDefaultMetadata()`) by having one class that both the reader and writer use that translates objects into the byte[]s and rows we use to interface with the KVS...

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -36,30 +38,32 @@ void checkShardSize() {
...
 
     public static final Hydrator<TargetedSweepMetadata> BYTES_HYDRATOR = input ->
-            ImmutableTargetedSweepMetadata.builder()
-                    .conservative((input[0] & 0x80) != 0)
-                    .dedicatedRow((input[0] & 0x40) != 0)
-                    .shard((input[0] << 2 | input[1] >> 6) & 0xFF)
-                    .dedicatedRowNumber(input[1] & 0x3F)
-                    .build();
+        ImmutableTargetedSweepMetadata.builder()
+                .conservative((input[0] & 0x80) != 0)
```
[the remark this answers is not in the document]
gmaretic: Something like this?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -36,30 +38,32 @@ void checkShardSize() {
...
+                .dedicatedRow((input[0] & 0x40) != 0)
+                .shard((input[0] << 2 | (input[1] & 0xFF) >> 6) & 0xFF)
```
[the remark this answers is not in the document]
gmaretic: That's fair, probably makes more sense even though I'm confident now it's good

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -36,30 +38,32 @@ void checkShardSize() {
...
+                .build();
+
 
     @Override
     public byte[] persistToBytes() {
         byte[] result = new byte[] { 0, 0, 0, 0 };
-        if (conservative()) {
-            result[0] |= 0x80;
-        }
+        result[0] |= (shard() & 0xFF) >> 2;
```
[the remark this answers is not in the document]
gmaretic: Yes, this was just for completeness sake

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
gmaretic: I think this version now is the easiest to read

gsheasby:
Looks like you have some test failures due to mock misuse:

`SnapshotTransactionTest.noWritesAddedToSweepQueueOnException`
```
[00:05:41] com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest > noWritesAddedToSweepQueueOnException FAILED
[00:05:41]     org.mockito.exceptions.misusing.NotAMockException: 
[00:05:41]     Argument(s) passed is not a mock!
[00:05:41]     Examples of correct verifications:
[00:05:41]         verifyNoMoreInteractions(mockOne, mockTwo);
[00:05:41]         verifyZeroInteractions(mockOne, mockTwo);
[00:05:41]         at com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest.noWritesAddedToSweepQueueOnException(SnapshotTransactionTest.java:975)

[00:05:41] com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest > noWritesAddedToSweepQueueOnConflict FAILED
[00:05:41]     org.mockito.exceptions.misusing.NotAMockException: 
[00:05:41]     Argument passed to verify() is of type KvsSweepQueuePersister and is not a mock!
[00:05:41]     Make sure you place the parenthesis correctly!
[00:05:41]     See the examples of correct verifications:
[00:05:41]         verify(mock).someMethod();
[00:05:41]         verify(mock, times(10)).someMethod();
[00:05:41]         verify(mock, atLeastOnce()).someMethod();
[00:05:41]         at com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest.noWritesAddedToSweepQueueOnConflict(SnapshotTransactionTest.java:949)
```

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TargetedSweepMetadata.java
```diff
@@ -39,31 +43,30 @@ void checkShardSize() {
...
+            result[0] |= USE_DEDICATED_ROWS_MASK;
         }
         if (conservative()) {
-            result[0] |= 0x80;
+            result[0] |= SWEEP_STRATEGY_MASK;
         }
-        result[1] |= dedicatedRowNumber();
-        result[1] |= (shard() << 6) & 0xFF;
+        result[1] |= dedicatedRowNumber() & DEDICATED_ROW_NUMBER_MASK;
+        result[1] |= (shard() << 6) & BYTE_MASK;
```
jeremyk-91: Thanks! Reads much more clearly now

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
jeremyk-91: Yep, this is what I would prefer as well. Thanks

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private void addIfMaxForCell(long ts, WriteReference writeRef, Map<WriteReference, Long> result) {
+        result.merge(writeRef, ts, Math::max);
+    }
+
+    private static ColumnRangeSelection allPossibleColumns() {
+        byte[] startCol = SweepableCellsTable.SweepableCellsColumn.of(0L, -TargetedSweepMetadata.MAX_DEDICATED_ROWS)
+                .persistToBytes();
+        byte[] endCol = SweepableCellsTable.SweepableCellsColumn.of(SweepQueueUtils.TS_FINE_GRANULARITY, 0)
+                .persistToBytes();
```
jeremyk-91: yep, confirming this looks like it grabs everything (this is start inclusive end exclusive)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsReader.java
```diff
@@ -0,0 +1,139 @@
...
+                .dedicatedRowNumber(0)
+                .build();
+    }
+
+    private RowColumnRangeIterator getAllColumns(SweepableCellsTable.SweepableCellsRow row) {
+        return getAllColumns(ImmutableList.of(row.persistToBytes()));
+    }
+
+    private RowColumnRangeIterator getAllColumns(Iterable<byte[]> rows) {
+        return kvs.getRowsColumnRange(TABLE_REF, rows, ALL_COLUMNS, 1000, SweepQueueUtils.CAS_TS);
```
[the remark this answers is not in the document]
gsheasby: If we have metrics in place, then we can figure out what the numbers are like in practice, and then have tests for average and extreme cases.

PR #3124, opened 2018-04-27
[TS] Part 6: Reintroduce Tombstone Info

gmaretic:
**Goals (and why)**:
We need to know which writes were tombstones to properly set the ranged tombstone for thorough sweep

Pretty straightforward

Is my assumption that writing a tombstone == putting an empty byte array into the KVS correct?
Is the treatment of isTombstone as auxiliary confusing?

`WriteReference.java`

ASAP, so it can get merged into part 5 before too much changes.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3124)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 
 import org.immutables.value.Value;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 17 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
```diff
@@ -28,24 +29,29 @@
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
 import com.palantir.common.persist.Persistable;
 
-@JsonDeserialize(as = ImmutableTableReferenceAndCell.class)
-@JsonSerialize(as = ImmutableTableReferenceAndCell.class)
+@JsonDeserialize(as = ImmutableWriteReference.class)
+@JsonSerialize(as = ImmutableWriteReference.class)
 @Value.Immutable
-public abstract class TableReferenceAndCell implements Persistable {
+public abstract class WriteReference implements Persistable {
```
gsheasby: OK, happy with this name.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepQueueReadWriteTest.java
```diff
@@ -77,22 +76,32 @@ public void setup() {
...
 
-    public static List<WriteInfo> writeToUniqueCellsInSameShard(KvsSweepQueueWriter writer, long timestamp, int number, boolean conservative) {
+    public static List<WriteInfo> writeToUniqueCellsInSameShard(KvsSweepQueueWriter writer, long timestamp, int number,
+            boolean conservative) {
         TableReference tableRef = conservative ? TABLE_REF : TABLE_REF2;
         List<WriteInfo> result = new ArrayList<>();
         for (int i = 0; i < number; i++) {
             Cell cell = getCellWithFixedHash(i);
-            result.add(WriteInfo.of(tableRef, cell, timestamp));
+            result.add(WriteInfo.of(tableRef, cell, true, timestamp));
```
gsheasby: true is slightly ugly - maybe replace `of` with a couple of methods to hide the true/false?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/WriteInfoTest.java
```diff
@@ -0,0 +1,66 @@
...
+import com.palantir.atlasdb.sweep.Sweeper;
+
+public class WriteInfoTest {
+    private static final TableReference TABLE_REF = TableReference.createFromFullyQualifiedName("test.test");
+    private static final Cell CELL = Cell.create(new byte[]{1}, new byte[]{2});
+    private static final long ONE = 1L;
+    private static final long TWO = 2L;
+
+    @Test
+    public void tombstoneStatusIsIgnoredForEqualityAndSharding() {
```
gsheasby: Why is this desirable?

gmaretic: We need it to be the same for sharding. The equality is inherited from WriteReference, where the tombstone status is an auxiliary variable to simplify the logic since a tombstone and a normal write to the same cell are equivalent for the purposes of grouping writes and determining the one with the highest timestamp.

gmaretic:
Merging to reduce entropy. @jeremyk-91 will be happy to address any comments you have :)

PR #3129, opened 2018-04-30
[TS] Part 7.5: Fix Writes To Dedicated Rows

gmaretic:
**Goals (and why)**:
Make dedicated rows unique per transaction. Old impl was using the partition, which could clash in case we have multiple dedicated writes in transaction in the same partition. There was no correctness issue, as they were being written to separate columns, but the dedicated rows were able to grow out of proportion.

Just use the whole timestamp as the row component instead of the partition for dedicated rows.

None, but we could use 0 for the column component if we like -- it would slightly reduce the byte size and we could have a tighter range for the range scan. Probably not worth the added code obfuscation?

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3129)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsWriter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsWriter.java
@@ -84,13 +84,17 @@ public class SweepableCellsWriter extends KvsSweepQueueWriter {
                 .dedicatedRowNumber(dedicatedRow)
                 .build();
 
-        return SweepableCellsTable.SweepableCellsRow.of(
-                SweepQueueUtils.tsPartitionFine(info.timestamp()), metadata.persistToBytes());
+        return SweepableCellsTable.SweepableCellsRow.of(getTimestampOrPartition(info, dedicate),
+                metadata.persistToBytes());
     }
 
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCellsWriter.java
```diff
@@ -84,15 +84,19 @@ private void insert(PartitionInfo info, WriteReference writeRef, boolean dedicat
...
 
     private SweepableCellsTable.SweepableCellsColumnValue createColVal(long ts, long index, WriteReference writeRef) {
         SweepableCellsTable.SweepableCellsColumn col = SweepableCellsTable.SweepableCellsColumn.of(tsMod(ts), index);
         return SweepableCellsTable.SweepableCellsColumnValue.of(col, writeRef);
     }
 
+    private long getTimestampOrPartition(PartitionInfo info, boolean dedicate) {
+        return dedicate ? info.timestamp() : SweepQueueUtils.tsPartitionFine(info.timestamp());
+    }
+
```
jeremyk-91: 💯 yep, this version looks right.
We should probably have a test though given this passed multiple reviewers!

gmaretic: Yes, but writing a test is actually nontrivial. One thing the old version would fail and the new would pass is a check that the timestamp modulus part of the column name must be equal for all cells in the row

PR #3137, opened 2018-05-01
[TS] Part 8: Scrubber Impl & Refactor

gmaretic:
**Goals (and why)**:
This finishes all the parts for the first version of the sweep queue. What remains to be done before it is hooked up to TManagers is to set up the background threads (with proper cleanup), add support for increasing number of shards, and sweep at higher granularity than fine partition. Also tests.

I hope it is pretty self-explanatory. Might be time to add java docs soon as I imagine things will change much less after this refactor.

The tests are incoming, so it might not work correctly -- suggestions for tests welcome!

`KvsSweepQueueScrubber.java`

Test will take a while, so up to you

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3137)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
```diff
@@ -29,7 +29,8 @@
 public final class SweepQueueUtils {
     public static final long TS_COARSE_GRANULARITY = 10_000_000L;
     public static final long TS_FINE_GRANULARITY = 50_000L;
-    public static final long CAS_TS = 1L;
+    public static final long READ_TS = 1L;
```
sandorw: @ashrayjain for thoughts on timestamps here - we should make sure we're not interfering with compactions on the queue tables with our choice of timestamps

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueReader.java
```diff
@@ -18,39 +18,42 @@
...
 
 import com.google.common.collect.ImmutableList;
-import com.palantir.atlasdb.keyvalue.api.KeyValueService;
-import com.palantir.atlasdb.protos.generated.TableMetadataPersistence;
+import com.palantir.util.Pair;
 
-public class KvsSweepQueueReader implements SweepQueueReader {
-    private final SweepableCellsReader cellsReader;
-    private final SweepableTimestampsReader timestampsReader;
+public final class KvsSweepQueueReader implements SweepQueueReader {
```
sandorw: We could also have a single reader that is passed the relevant `ShardAndStrategy` along with the `consumeNexBatch()` call, since this class isn't stateful. Avoids some extra object allocations, but requires someone a layer up to iterate through the current shard count. I'm not sure which will be cleaner when we get to incrementing the shard count.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueReader.java
```diff
@@ -18,39 +18,42 @@
...
-            TableMetadataPersistence.SweepStrategy sweepStrategy) {
-        return new KvsSweepQueueReader(kvs, ShardAndStrategy.of(shard, sweepStrategy));
+    public static KvsSweepQueueReader create(KvsSweepQueueTables tables, ShardAndStrategy shardStrat) {
+        return new KvsSweepQueueReader(tables.sweepableCells, tables.sweepableTimestamps, tables.progress, shardStrat);
     }
 
     @Override
-    public void consumeNextBatch(Consumer<Collection<WriteInfo>> consumer, long maxTimestampExclusive) {
-        consumer.accept(getNextBatch(maxTimestampExclusive));
+    public void consumeNextBatch(Consumer<Collection<WriteInfo>> consumer, long maxTsExclusive) {
```
sandorw: This reads nicely 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
gmaretic: CAS is always done at timestamp 0 anyway, so just kept it 0 for everything else and then read at 1. Note that these are atlas timestamps, we don't specify cassandra timestamps anywhere directly.

ashrayjain: As long as we are not overriding c* timestamps, compactions should progress as usual.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+    void deleteDedicatedRows(ShardAndStrategy shardAndStrategy, long partitionFine) {
+        rangeRequestsDedicatedRows(shardAndStrategy, partitionFine).forEach(this::deleteRange);
+    }
+
+    void deleteNonDedicatedRow(ShardAndStrategy shardAndStrategy, long partitionFine) {
+        deleteRange(rangeRequestNonDedicatedRow(shardAndStrategy, partitionFine));
+    }
+
+    private void addReferenceToDedicatedRows(PartitionInfo info, List<WriteInfo> writes, Map<Cell, byte[]> cells) {
+        addCell(info, WriteReference.DUMMY, false, 0, -numberDedicatedRows(writes), cells);
```
sandorw: Definitely worth some comments here given fun things like the `-`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+    @VisibleForTesting
+    static final int MAX_CELLS_DEDICATED = 100_000;
+    private static final ColumnRangeSelection ALL_COLUMNS = allPossibleColumns();
+
+    public SweepableCells(KeyValueService kvs, WriteInfoPartitioner partitioner) {
+        super(kvs, TargetedSweepTableFactory.of().getSweepableCellsTable(null).getTableRef(), partitioner);
+    }
+
+    @Override
+    void populateCells(PartitionInfo partitionInfo, List<WriteInfo> writes, Map<Cell, byte[]> cells) {
```
sandorw: Clarify here that cells is effectively an output - why don't we just return the map instead?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+    void deleteNonDedicatedRow(ShardAndStrategy shardAndStrategy, long partitionFine) {
+        deleteRange(rangeRequestNonDedicatedRow(shardAndStrategy, partitionFine));
+    }
+
+    private void addReferenceToDedicatedRows(PartitionInfo info, List<WriteInfo> writes, Map<Cell, byte[]> cells) {
+        addCell(info, WriteReference.DUMMY, false, 0, -numberDedicatedRows(writes), cells);
+    }
+
+    private void addWrite(PartitionInfo info, WriteInfo write, boolean dedicate, long index, Map<Cell, byte[]> cells) {
+        addCell(info, write.writeRef(), dedicate, index / MAX_CELLS_DEDICATED, index % MAX_CELLS_DEDICATED, cells);
```
sandorw: Clever, but was slightly puzzling why `MAX_CELLS_DEDICATED` was being used for non-dedicated rows. Worth some comments.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.keyvalue.api.WriteReference;
+import com.palantir.atlasdb.schema.generated.SweepableCellsTable;
+import com.palantir.atlasdb.schema.generated.TargetedSweepTableFactory;
+
+public class SweepableCells extends KvsSweepQueueWriter {
+    @VisibleForTesting
+    static final int MAX_CELLS_GENERIC = 50;
+    @VisibleForTesting
+    static final int MAX_CELLS_DEDICATED = 100_000;
```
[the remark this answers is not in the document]
sandorw: @j-baker for thoughts or follow-up discussion

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+import com.palantir.atlasdb.keyvalue.api.RangeRequest;
+import com.palantir.atlasdb.keyvalue.api.RangeRequests;
+import com.palantir.atlasdb.keyvalue.api.RowColumnRangeIterator;
+import com.palantir.atlasdb.keyvalue.api.TargetedSweepMetadata;
+import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.keyvalue.api.WriteReference;
+import com.palantir.atlasdb.schema.generated.SweepableCellsTable;
+import com.palantir.atlasdb.schema.generated.TargetedSweepTableFactory;
+
+public class SweepableCells extends KvsSweepQueueWriter {
```
sandorw: Let's follow up later in the granularity/batching discussion and discuss how partial sweep batches of dedicated rows will work.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueuePersister.java
```diff
@@ -18,38 +18,22 @@
...
+    private KvsSweepQueuePersister(SweepableCells cells, SweepableTimestamps ts) {
+        this.sweepableCells = cells;
+        this.sweepableTimestamps = ts;
     }
 
-    // todo(gmaretic): remove this
-    public void initialize() {
-        Schemas.createTablesAndIndexes(TargetedSweepSchema.INSTANCE.getLatestSchema(), kvs);
+    public static KvsSweepQueuePersister create(KvsSweepQueueTables tables) {
+        return new KvsSweepQueuePersister(tables.sweepableCells, tables.sweepableTimestamps);
```
gsheasby: Here, two internals of `tables` are exposed, just so that we can call the same method (`enqueue`) on them at the same time, with the same argument. `KvsSweepQueueTables` itself doesn't do anything.

Let's keep `tables` as a field here and call `tables.enqueue(writes)` below, which will then enqueue the writes for both `sweepableCells` and `sweepableTimestamps`. Then, `KvsSweepQueueTables` has more purpose, and we're closer to making its fields private.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
```diff
@@ -16,29 +16,45 @@
...
 
     @Override
-    public void enqueue(List<WriteInfo> writes) {
-        Map<Cell, byte[]> batchedWrites = batchWrites(writes);
-        if (!batchedWrites.isEmpty()) {
-            kvs.put(tableRef, batchWrites(writes), 0L);
+    public void enqueue(List<WriteInfo> allWrites) {
+        Map<Cell, byte[]> cellsToWrite = new HashMap<>();
+        Map<PartitionInfo, List<WriteInfo>> partitionedWrites = partitioner.filterAndPartition(allWrites);
+        partitionedWrites.forEach((partitionInfo, writes) -> populateCells(partitionInfo, writes, cellsToWrite));
```
gsheasby: Note: It's kind of janky that `cellsToWrite` gets passed down through many levels of the call stack before someone calls `put`, rather than having `populateCells` fetch the cells which are then `putAll` here. I guess your way works out more efficient, as we don't need to create tons of mini-maps but just have this one map to rule them all?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
```diff
@@ -21,10 +21,15 @@
...
-    // On the other hand, if the consumer fails by throwing an exception, the same batch will be consumed
-    // on the next invocation.
+    /**
+     * Passes the next batch of writes to the given consumer. If the consumer returns without throwing,
+     * the batch is considered consumed, and future invocations of this method will consume only newer writes.
+     * On the other hand, if the consumer fails by throwing an exception, the same batch will be consumed
+     * on the next invocation.
+     * @param consumer
+     * @param maxTimestampExclusive
+     * @return
```
gsheasby: nit: I think `@return` can be deleted here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gsheasby: +1, there is some sorcery here. Also `false` sounds like it means non-dedicated, but the method talks about dedicated rows.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+        addCell(info, WriteReference.DUMMY, false, 0, -numberDedicatedRows(writes), cells);
+    }
+
+    private void addWrite(PartitionInfo info, WriteInfo write, boolean dedicate, long index, Map<Cell, byte[]> cells) {
+        addCell(info, write.writeRef(), dedicate, index / MAX_CELLS_DEDICATED, index % MAX_CELLS_DEDICATED, cells);
+    }
+
+    private void addCell(PartitionInfo info, WriteReference writeRef, boolean dedicate, long dedicatedRow,
+            long index, Map<Cell, byte[]> cells) {
+        SweepableCellsTable.SweepableCellsRow row = computeRow(info, dedicate, dedicatedRow);
```
gsheasby: nit: method ordering is all over the place in this file, which makes it hard to follow.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private boolean isReferenceToDedicatedRows(SweepableCellsTable.SweepableCellsColumn col) {
+        return col.getWriteIndex() < 0;
+    }
+
+    private void addWritesFromDedicated(SweepableCellsTable.SweepableCellsRow row,
+            SweepableCellsTable.SweepableCellsColumn col, Map<CellReference, WriteInfo> writes) {
+        List<byte[]> dedicatedRows = computeDedicatedRows(row, col);
+        RowColumnRangeIterator iterator = getColumnRangeAll(dedicatedRows);
```
gsheasby: `getColumnRangeAll` name could be clearer. Are we getting all columns from the given dedicated rows?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    private void addWriteFromValue(long timestamp, Value value, Map<CellReference, WriteInfo> writes) {
+        WriteReference writeRef = SweepableCellsTable.SweepableCellsColumnValue.hydrateValue(value.getContents());
+        updateLatestForCell(timestamp, writeRef, writes);
+    }
+
+    private void updateLatestForCell(long ts, WriteReference writeRef, Map<CellReference, WriteInfo> writes) {
+        WriteInfo newWrite = WriteInfo.of(writeRef, ts);
+        writes.merge(writeRef.cellReference(), newWrite,
```
gsheasby: ok, so we're doing some kind of merge here which means we need to know what the old value was. Is it possible/efficient to do that at a higher level to avoid all the write passing?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+                (writeOne, writeTwo) -> writeOne.timestamp() > writeTwo.timestamp() ? writeOne : writeTwo);
+    }
+
+    private List<RangeRequest> rangeRequestsDedicatedRows(ShardAndStrategy shardAndStrategy, long partitionFine) {
+        SweepableCellsTable.SweepableCellsRow startingRow = computeRow(partitionFine, shardAndStrategy);
+        RowColumnRangeIterator rowIterator = getColumnRangeAllForRow(startingRow);
+        List<RangeRequest> requests = new ArrayList<>();
+        rowIterator.forEachRemaining(entry -> addRangeRequestIfDedicated(startingRow, computeColumn(entry), requests));
```
gsheasby: Can we rework this to filter the rows and then add? This would avoid having to pass around another output parameter.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -0,0 +1,251 @@
...
+
+    private SweepableCellsTable.SweepableCellsRow computeRow(long timestamp, TargetedSweepMetadata metadata) {
+        return SweepableCellsTable.SweepableCellsRow.of(timestamp, metadata.persistToBytes());
+    }
+
+    private List<byte[]> computeDedicatedRows(SweepableCellsTable.SweepableCellsRow row,
+            SweepableCellsTable.SweepableCellsColumn col) {
+        TargetedSweepMetadata metadata = TargetedSweepMetadata.BYTES_HYDRATOR.hydrateFromBytes(row.getMetadata());
+        long timestamp = getTimestamp(row, col);
+        int numberOfDedicatedRows = (int) -col.getWriteIndex();
```
gsheasby: here's the minus again. Would it be better to put that logic in one place?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableTimestamps.java
```diff
@@ -0,0 +1,137 @@
...
+    private Optional<ColumnRangeSelection> getColRangeSelection(long minFine, long maxFine) {
+        if (minFine >= maxFine) {
+            return Optional.empty();
+        }
+        byte[] start = SweepableTimestampsTable.SweepableTimestampsColumn.of(minFine).persistToBytes();
+        byte[] end = SweepableTimestampsTable.SweepableTimestampsColumn.of(maxFine).persistToBytes();
+        return Optional.of(new ColumnRangeSelection(start, end));
+    }
+
+    private Optional<Long> getCandidatesInCoarsePartition(ShardAndStrategy shardStrategy, long partitionCoarse,
```
gsheasby: would this be more accurately called `getNextCandidateInCoarsePartition`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1377,6 +1377,9 @@ private void commitWrites(TransactionService transactionService) {
             Timer.Context conflictsTimer = getTimer("commitCheckingForConflicts").time();
             throwIfConflictOnCommit(commitLocksToken, transactionService);
             long millisCheckingForConflicts = TimeUnit.NANOSECONDS.toMillis(conflictsTimer.stop());
+
+            sweepQueue.enqueue(writesByTable, getStartTimestamp());
```
gsheasby: This may succeed, and then the `multiPut` might fail. Then when we read from the queue, we'll be trying to process writes that don't exist. Is this an issue?

Sanity check: `getStartTimestamp()` is a `Supplier<Long>`. How sure are we that calling it twice will return the same value, both now and in the future?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -18,19 +18,20 @@
...
-import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 import static com.palantir.atlasdb.protos.generated.TableMetadataPersistence.SweepStrategy.CONSERVATIVE;
 import static com.palantir.atlasdb.protos.generated.TableMetadataPersistence.SweepStrategy.NOTHING;
 import static com.palantir.atlasdb.protos.generated.TableMetadataPersistence.SweepStrategy.THOROUGH;
+import static com.palantir.atlasdb.sweep.queue.SweepQueueTablesTest.metadataBytes;
```
gsheasby: it's slightly weird to import functions statically like this.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -195,6 +188,169 @@ public void sweepDeletesTombstonesWhenWriteHasHigherTimestamp() {
         assertReadAtTimestampReturnsValue(TABLE_CONSERVATIVE, TS2 + 1, TS2);
     }
 
+    @Test
+    public void sweepProgressesAndSkipsEmptyFinePartitions() {
```
gsheasby: This test could use some comments to explain the state of the world at each stage. (same for various other tests in this class)

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -195,6 +188,169 @@ public void sweepDeletesTombstonesWhenWriteHasHigherTimestamp() {
...
+        long tsSecondPartitionFine = TS + TS_FINE_GRANULARITY;
+        long largestFirstPartitionCoarse = TS_COARSE_GRANULARITY - 1L;
+        enqueueWrite(TABLE_CONSERVATIVE, TS);
+        enqueueWrite(TABLE_CONSERVATIVE, tsSecondPartitionFine);
+        enqueueWrite(TABLE_CONSERVATIVE, largestFirstPartitionCoarse);
+
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertLowestFinePartitionInSweepableTimestampsEquals(tsPartitionFine(TS));
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertLowestFinePartitionInSweepableTimestampsEquals(tsPartitionFine(TS));
```
gsheasby: why doesn't this advance? (ditto below)

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -195,6 +188,169 @@ public void sweepDeletesTombstonesWhenWriteHasHigherTimestamp() {
...
+
+        // next sweep goes into a new coarse partition
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertNoEntriesInSweepableTimestampsBeforeImmutableTimestamp();
+    }
+
+    @Test
+    public void sweepableTimestampsGetsScrubbedWhenLastSweptProgressesInNewCoarsePartition2() {
```
gsheasby: nit: `2`?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -195,6 +188,169 @@ public void sweepDeletesTombstonesWhenWriteHasHigherTimestamp() {
...
+        assertLowestFinePartitionInSweepableTimestampsEquals(tsPartitionFine(TS));
+
+        // next sweep goes into a new coarse partition
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertLowestFinePartitionInSweepableTimestampsEquals(tsPartitionFine(thirdPartitionCoarse));
+    }
+
+    @Test
+    public void sweepableCellsGetsScrubbedAfterEachBatch() {
+        long tsSecondPartitionFine = TS + TS_FINE_GRANULARITY;
```
gsheasby: This is defined thrice. Consider extracting a constant.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -241,4 +397,21 @@ private void assertReadAtTimestampReturnsNothing(TableReference tableRef, long r
         Map<Cell, Long> singleRead = ImmutableMap.of(DEFAULT_CELL, ts);
         return kvs.get(tableRef, singleRead);
     }
+
+    private void assertProgressUpdatedToFineTimestampPartition(long partitionFine) {
+        assertThat(sweepQueue.tables.progress.getLastSweptTimestampPartition(ShardAndStrategy.conservative(CONS_SHARD)))
```
gsheasby: Demeter is displeased with you for doubly breaking her Law :zap:

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepableCellsTest.java
```diff
@@ -31,96 +31,96 @@
...
     @Test
     public void canReadMultipleEntriesInSingleShardMultipleTransactionsCombined() {
-        List<WriteInfo> writesFirst = writeToUniqueCellsInFixedShard(writer, TS, MAX_CELLS_GENERIC * 2 + 1, TABLE_CONS);
-        List<WriteInfo> writesLast = writeToUniqueCellsInFixedShard(writer, TS + 2, 1, TABLE_CONS);
-        List<WriteInfo> writesMiddle = writeToUniqueCellsInFixedShard(writer, TS + 1, MAX_CELLS_GENERIC + 1, TABLE_CONS);
+        List<WriteInfo> first = writeToCellsInFixedShard(sweepableCells, TS, MAX_CELLS_GENERIC * 2 + 1, TABLE_CONS);
+        List<WriteInfo> last = writeToCellsInFixedShard(sweepableCells, TS + 2, 1, TABLE_CONS);
+        List<WriteInfo> middle = writeToCellsInFixedShard(sweepableCells, TS + 1, MAX_CELLS_GENERIC + 1, TABLE_CONS);
         // The expected list of latest writes contains:
         // ((0, 0), TS + 2
```
gsheasby: OCD nit: missing trailing `)`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/WriteInfoPartitionerTest.java
```diff
@@ -23,6 +23,7 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import static com.palantir.atlasdb.sweep.queue.SweepQueueTablesTest.metadataBytes;
```
gsheasby: Static method import. Maybe this is personal preference? Not sure if we have specific guidelines about it.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
sandorw: That's by design, to avoid any case where we've written data but not updated the queue. If the `multiPut` fails (or only partially succeeds), we'll just issue some possibly empty range deletes. 

As an extra detail, we must have targeted sweep check if the transactions actually committed or not and clean up *only* the data they wrote if they did not commit. Otherwise, moving the location of the write would be a correctness issue - we may range delete valid data behind an uncommitted transaction.

The `getStartTimestamp()` issue is present all over this class. We could memoize locally to be defensive, but all normal suppliers that I know of are already memoized.

sandorw: Also, make sure to add tests for failed transactions :)

gmaretic: Very good point. This will be done in a separate PR

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -236,7 +236,7 @@ static boolean isFastFailoverException(Throwable ex) {
     }
 
     private static class Default implements RequestExceptionHandlerStrategy {
-        private static final RequestExceptionHandlerStrategy INSTANCE = new Default();
+        private static final RequestExceptionHandlerStrategy  INSTANCE = new Default();
```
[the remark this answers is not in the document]
gmaretic: That's what happens when I use my current branch to answer support questions 🤓

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueuePersister.java
gmaretic: Will merge these two into one class

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
gmaretic: This is to keep only the write with the maximum timestamp for each cell in memory at a time. Given that we will also need to do lookups to determine which writes were actually committed, the approach you mention makes more sense. Will implement that all in one PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gmaretic: This was an attempt to make things less chaotic by grouping methods that did similar things together. Would it be easier if I just ordered them top to bottom as they were invoked in the non-private methods?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gmaretic: getWithColumnRangeAll?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableTimestamps.java
gmaretic: Renamed in next PR

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
gmaretic: It makes the tests much more readable, but don't feel too strongly about it

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
gmaretic: Planned to rename this later and forgot 👎

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
gmaretic: Doubly visible for testing? I guess I can have a new KvsSweepProgress to check it out 🦌

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/WriteInfoPartitionerTest.java
gmaretic: Checkstyle doesn't complain, and we routinely do it for `Mockito`. I'm fine either way, so if it is too jarring, will remove :)

gsheasby: OK, ignore me then. Your point about readability in the other comment is valid, too.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
gsheasby: I meant that you are reaching into the internals of sweepQueue, and then into the internals of tables, to access functionality exposed by progress. Merging a couple of the objects will help, as will allowing accessors. Doing it only in test methods is less bad then also doing it in production code.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gsheasby: `getAllColumns`? At this level, we don't care that we're doing it with `kvs.getRowsColumnRange`, and in any case that is implied by the return type.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gsheasby: Not sure. It might be that this class simply does too much, but at this point I wouldn't pour too much time into redesigning or reordering.

gmaretic:
Again, merging to reduce loose threads; addressed some things in part 9, and tracking others

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableTimestamps.java
gsheasby: Noting that this isn't in part 9. I don't care that much, though.

PR #3145, opened 2018-05-03
Resiliency to Cassandra Rolling Restarts + Respect for maxTriesOnHost()

### LINKED ISSUE #3139: Cassandra rolling restarts cause Atlas to be unavailable for a refresh interval
At several internal deployments, it is possible for Cassandra hosts to remain blacklisted for considerably longer periods of time than the nodes are actually down for. This pattern tends to be brought about by a rolling restart of Cassandra nodes.

There is a background thread that is scheduled with a fixed configurable delay that unblacklists nodes. However, setting this to say <10 seconds will cause a lot of network requests. Having say 200 nodes (across all Atlas-using services) all fire their refreshes to a 6 node cluster every 5 seconds, say, could be troublesome.

Quoting from the ticket:

>Ideally:
>1. If things are blacklisted then generally skip them
>2. If you can't contact enough hosts, try the blacklisted ones (maybe just once)
>3. If you try a blacklisted host, and it works, take it off the blacklist

Internal reference: PDS-66807

jeremyk-91:
**Goals (and why)**: Fix #3139.

- Respect the `maxTriesOnHost()` configuration option even when dealing with 'transient' exceptions - it is apparently possible for six `TTransportException`s to crush a request.
- In addition to having the background thread refresh the blacklist, after a successful operation has completed we also un-blacklist a node.

- Is a successful request completion sufficient in place of running the health-check / are there reasons we might want a node that can answer a client request but cannot answer our health-check to stay on the blacklist?
- What was the intention of `maxTriesOfHost()` not actually being the maximum number of tries on a given host? It's clearly simpler to not have a check on exception type, meaning that initially it was desired that we re-try on the same node in the event of a `TTransportException`.

**Where should we start reviewing?**: CassandraRequestExceptionHandler / CassandraClientPoolImpl.

**Priority (whenever / two weeks / yesterday)**: tomorrow would be great; if not early next week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3145)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -381,7 +381,9 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
             CassandraClientPoolingContainer hostPool = getPreferredHostOrFallBack(req);
 
             try {
-                return runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                V response = runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                blacklist.remove(hostPool.getHost()); // successful request -> can un-blacklist
+                return response;
             } catch (Exception ex) {
                 exceptionHandler.handleExceptionFromRequest(req, hostPool.getHost(), ex);
             }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -415,7 +415,9 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
     public <V, K extends Exception> V runOnHost(InetSocketAddress specifiedHost,
             FunctionCheckedException<CassandraClient, V, K> fn) throws K {
         CassandraClientPoolingContainer hostPool = cassandra.getPools().get(specifiedHost);
-        return runWithPooledResourceRecordingMetrics(hostPool, fn);
+        V response = runWithPooledResourceRecordingMetrics(hostPool, fn);
+        blacklist.remove(specifiedHost);
+        return response;
     }
 
     private <V, K extends Exception> V runWithPooledResourceRecordingMetrics(
--- a/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
+++ b/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
@@ -187,7 +187,6 @@ public class CassandraClientPoolTest {
         verifyBlacklistMetric(1);
     }
 
-    @SuppressWarnings("unchecked") // We know the types are correct within this test.
     @Test
     public void successfulRequestCausesHostToBeRemovedFromBlacklist() {
         CassandraClientPool cassandraClientPool = clientPoolWithServersInCurrentPool(ImmutableSet.of(HOST_1));
--- a/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
+++ b/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
@@ -15,10 +15,13 @@
  */
 package com.palantir.atlasdb.keyvalue.cassandra;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
+import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -27,6 +30,7 @@ import java.net.SocketTimeoutException;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -46,7 +50,6 @@ public class CassandraClientPoolTest {
     private static final int TIME_BETWEEN_EVICTION_RUNS_SECONDS = 20;
     private static final int UNRESPONSIVE_HOST_BACKOFF_SECONDS = 5 * 60;
     private static final int DEFAULT_PORT = 5000;
-    private static final int OTHER_PORT = 6000;
     private static final String HOSTNAME_1 = "1.0.0.0";
     private static final String HOSTNAME_2 = "2.0.0.0";
     private static final String HOSTNAME_3 = "3.0.0.0";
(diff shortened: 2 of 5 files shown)


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
```diff
@@ -172,10 +181,10 @@ public void fastFailoverExceptionAlwaysRetriesOnDifferentHostDefault() {
...
     public void retryableExceptionsAreRetryableConservative() {
         Set<Exception> allExceptions = new HashSet<>();
-        allExceptions = Sets.union(allExceptions, connectionExceptions);
-        allExceptions = Sets.union(allExceptions, transientExceptions);
-        allExceptions = Sets.union(allExceptions, indicativeOfCassandraLoadException);
-        allExceptions = Sets.union(allExceptions, fastFailoverExceptions);
+        allExceptions = Sets.union(allExceptions, CONNECTION_EXCEPTIONS);
+        allExceptions = Sets.union(allExceptions, TRANSIENT_EXCEPTIONS);
+        allExceptions = Sets.union(allExceptions, INDICATIVE_OF_CASSANDRA_LOAD_EXCEPTIONS);
+        allExceptions = Sets.union(allExceptions, FAST_FAILOVER_EXCEPTIONS);
```
jeremyk-91: didn't see this - this use should be removed too

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
```diff
@@ -44,12 +48,23 @@
...
-    private static Set<Exception> fastFailoverExceptions = Sets.newHashSet(new InvalidRequestException());
+    private static final Set<Exception> TRANSIENT_EXCEPTIONS = ImmutableSet.of(new TTransportException());
+    private static final Set<Exception> INDICATIVE_OF_CASSANDRA_LOAD_EXCEPTIONS =
+            ImmutableSet.of(
+                    new NoSuchElementException(),
+                    new TimedOutException(),
+                    new UnavailableException(),
+                    new InsufficientConsistencyException(MESSAGE));
+    private static final Set<Exception> FAST_FAILOVER_EXCEPTIONS = ImmutableSet.of(new InvalidRequestException());
+    private static final Set<Exception> ALL_EXCEPTIONS = Stream.of(
```
gmaretic: 🥇

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
```diff
@@ -183,6 +187,40 @@ public void testBlacklistMetrics() {
...
+        AtomicReference<InetSocketAddress> downHost = new AtomicReference<>(HOST_1);
+        cassandraClientPool.getCurrentPools().values().forEach(pool -> setConditionalTimeoutFailureForHost(
+                pool, container -> container.getHost().equals(downHost.get())));
+
+        runNoopWithRetryOnHost(HOST_1, cassandraClientPool);
+        assertThat(blacklist.contains(HOST_1), is(true));
+
+        downHost.set(HOST_2);
+
+        runNoopWithRetryOnHost(HOST_1, cassandraClientPool);
```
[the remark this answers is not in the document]
jeremyk-91: Good spot - I wrote this first and then wrote the unblacklist test, but forgot to change this. Thanks!

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
jeremyk-91: 🎉

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -381,7 +381,9 @@ void runOneTimeStartupChecks() {
             CassandraClientPoolingContainer hostPool = getPreferredHostOrFallBack(req);
 
             try {
-                return runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                V response = runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                blacklist.remove(hostPool.getHost()); // successful request -> can un-blacklist
```
tpetracca: will this only do something if we randomly queried this host because all hosts are currently blacklisted?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -413,7 +415,9 @@ void runOneTimeStartupChecks() {
     public <V, K extends Exception> V runOnHost(InetSocketAddress specifiedHost,
             FunctionCheckedException<CassandraClient, V, K> fn) throws K {
         CassandraClientPoolingContainer hostPool = cassandra.getPools().get(specifiedHost);
-        return runWithPooledResourceRecordingMetrics(hostPool, fn);
+        V response = runWithPooledResourceRecordingMetrics(hostPool, fn);
+        blacklist.remove(specifiedHost);
```
tpetracca: again

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
```diff
@@ -44,12 +46,23 @@
...
 
-    private static Set<Exception> connectionExceptions = Sets.newHashSet(new SocketTimeoutException(MESSAGE),
+    private static final Set<Exception> CONNECTION_EXCEPTIONS = ImmutableSet.of(new SocketTimeoutException(MESSAGE),
             new CassandraClientFactory.ClientCreationFailedException(MESSAGE, CAUSE));
-    private static Set<Exception> transientExceptions = Sets.newHashSet(new TTransportException());
-    private static Set<Exception> indicativeOfCassandraLoadException = Sets.newHashSet(new NoSuchElementException(),
-            new TimedOutException(), new UnavailableException(), new InsufficientConsistencyException(MESSAGE));
-    private static Set<Exception> fastFailoverExceptions = Sets.newHashSet(new InvalidRequestException());
+    private static final Set<Exception> TRANSIENT_EXCEPTIONS = ImmutableSet.of(new TTransportException());
+    private static final Set<Exception> INDICATIVE_OF_CASSANDRA_LOAD_EXCEPTIONS =
```
tpetracca: What does CASSANDRA_LOAD mean here?
- entire cluster
- specific node
- specific set of nodes with the replica we're querying

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandlerTest.java
```diff
@@ -44,12 +46,23 @@
...
+            ImmutableSet.of(
+                    new NoSuchElementException(),
```
tpetracca: This is thrown by the client pool, not cassandra, right? Could also be indicative of too high query volume or over parallelization, rather than c* load.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
```diff
@@ -183,6 +187,40 @@ public void testBlacklistMetrics() {
...
+        assertThat(blacklist.contains(HOST_1), is(true));
+
+        fail.set(false);
+
+        runNoopWithRetryOnHost(HOST_1, cassandraClientPool);
+        assertThat(blacklist.contains(HOST_1), is(false));
+    }
+
+    @Test
+    public void resilientToRollingRestarts() {
```
sandorw: 🥇

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
sandorw: I believe tests show that you un-blacklist hosts even before all hosts are blacklisted.

jeremyk-91: In practice, yes. The `getPreferredHostOrFallBack` will screen out blacklisted hosts unless everyone was blacklisted.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
jeremyk-91: For this one, if you call `runOnHost` you're already committing to running your request on the host you provide (even if it is blacklisted, but there are already other good hosts).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
jeremyk-91: @sandorw there's actually an intermediate state in the rolling restart test where both 1 and 2 are on the blacklist. We then pick a random host we haven't already tried, which is only host 1, and then 1 is removed from the blacklist

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolTest.java
jeremyk-91: 🎉🎉

PR #3147, opened 2018-05-04
Initialize column range map with mutable list

ellisjoe:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3147)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -627,7 +627,7 @@ public class SerializableTransaction extends SnapshotTransaction {
                     if (rangesToRows.get(range) != null) {
                         rangesToRows.get(range).add(row);
                     } else {
-                        rangesToRows.put(range, ImmutableList.of(row));
+                        rangesToRows.put(range, Lists.newArrayList(row));
                     }
                 }
             }


sandorw:
Filed https://github.com/palantir/atlasdb/issues/3148 as a follow-up to add test coverage

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -627,7 +627,7 @@ private void verifyColumnRanges(Transaction readOnlyTransaction) {
                     if (rangesToRows.get(range) != null) {
                         rangesToRows.get(range).add(row);
                     } else {
-                        rangesToRows.put(range, ImmutableList.of(row));
+                        rangesToRows.put(range, Lists.newArrayList(row));
```
schlosna: could simplify the whole if/else to:

```
                    rangesToRows.computeIfAbsent(range, ignored -> Lists.newArrayList()).add(row);
```

ellisjoe: I'm happy either way. @sandorw?

sandorw: Going to keep this for now for the sake of merging and getting the release out. I'll bundle it with another PR which is fixing some other code style in the same class (https://github.com/palantir/atlasdb/pull/3149).

PR #3151, opened 2018-05-07
don't overwrite cass ts in all cases

tpetracca:
Specifically want to stop overwriting it for tombstone reclamation affected by these codepaths:
- KVS.multiPut
- KVS.addGarbageCollectionSentinelValues
- KVS.delete

In practice this PR stops overwriting it (which is generally bad practice) in two other places as well:
- CassandraTableDropper.putMetadataWithoutChangingSettings - which is a no-op given it was already using wall clock time
- KVS.putWithTimestamps - I believe this codepath is dead for Cassandra KVS

See internal document for a very in-detail look at why overwriting these timestamps is bad for tombstone reclamation in cassandra.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3151)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellValuePutter.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellValuePutter.java
@@ -20,6 +20,7 @@ import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.function.Supplier;
 
 import org.apache.cassandra.thrift.Column;
 import org.apache.cassandra.thrift.ColumnOrSuperColumn;
@@ -45,6 +46,8 @@ public class CellValuePutter {
     private static final Function<Map.Entry<Cell, Value>, Long> ENTRY_SIZING_FUNCTION = input ->
             input.getValue().getContents().length + 4L + Cells.getApproxSizeOfCell(input.getKey());
 
+    private final Supplier<Long> freshTimestampSupplier;
+
     private CassandraKeyValueServiceConfig config;
     private CassandraClientPool clientPool;
     private TaskRunner taskRunner;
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellValuePutter.java
```diff
@@ -101,6 +115,9 @@ public Void apply(CassandraClient client) throws Exception {
                             for (Map.Entry<Cell, Value> e : partition) {
                                 Cell cell = e.getKey();
                                 Column col = CassandraKeyValueServices.createColumn(cell, e.getValue());
+                                if (overwriteTimestamps) {
```
gsheasby: This is wasteful as you end up calling `createColumn` twice - this matters more if the value is large, as you'd extract it twice. It's easy to avoid:

```
Column col = overwriteTimestamps
        ? CassandraKeyValueServices.createColumnAndOverwriteTimestamp(cell, e.getValue())
        : CassandraKeyValueServices.createColumn(cell, e.getValue());
```

tpetracca:
So as it turns out the thrift driver actually requires you override the timestamp. I'll push another version of this that uses client side wall clock time (as opposed to cassandra having used server side wall clock time).

ashrayjain:
Yes, we should use a fresh atlas timestamp instead. The only thing that we really care about at the C* level is that we never write in the "past". As long as the timestamps are always monotonically increasing, we should be fine.

tpetracca:
As discussed offline, agree with your sentiments about using atlas logical time instead of wall clock time here. Unfortunately this inherently means taking a timestamp service dependency in `atlasdb-cassandra`. I attempted to show what this would look like without introducing a compile time dep on it by using a `Supplier<Long>` and passing it around.

Unfortunately the abstract-ness of the AtlasDbFactory makes this hard to do. Actually thinking about it, breaking the `AtlasDbFactory.createRawKeyValueService()` API probably isn't a big deal; other implementers can jsut ignore the extra arg.

sandorw:
Yep, that was my thinking - it's a pretty small change to the KVSs.

tboam:
I think this is going to be pretty gnarly whatever you do.  Async initialisation will keep trying to create an object until it is successful.  If it gets an Exception then it cleans up and will then try again.

If KVS construction needs to query Cassandra with a timestamp from timelock then this needs to happen inside the tryInitialise function so that Exceptions are caught, tidy up is performed and creation is tried again.  It may look weird that the KVS reports unavailable when C* is up just because Timelock is down.

tboam:
@tpetracca @jeremyk-91 is this still required after #3224 ?

Issue #3122, opened 2018-04-26
Release schema-mutation lock on shutdown


sandorw:
We release the backup/sweep lock on shutdown, but not the schema-mutation lock. There's been a number of issues recently where services were double restarted after an upgrade, leading to a large number of lost locks. Since the restarts are graceful shutdowns, we could in theory release the lock during the process.

If we consider this, we should confirm that there's no correctness issue if we release the lock, but C* is still processing our createTable command.

sandorw:
Discussed offline, we just need to make sure we wait for a createTable command to complete before releasing the lock and shutting down. Given that, we may want to decrease the timeout for waiting to grab the lock (currently 60s).

sandorw:
We'll certainly do it this week - will update the ticket when it's complete and a release is out.

tboam:
Released in 0.83

PR #3152, opened 2018-05-08
Release schema mutation lock upon txn manager close

sandorw:
**Goals (and why)**:
Best-effort clean-up of the C* schema mutation lock when a transaction manager is shut down.

Run `createTables` and `dropTables` calls on an executor service. When the txn manager is closed, any running task is given 2 minutes to complete before continuing to close the txn manager. 

Ideas for testing? Any concerns with interrupt handling?

This week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3152)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -29,6 +29,9 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -64,6 +67,7 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.Futures;
 import com.palantir.async.initializer.AsyncInitializer;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cassandra.CassandraKeyValueServiceConfig;
@@ -181,15 +185,18 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
         }
     }
 
+    private static final long SCHEMA_MUTATION_TASK_TIMEOUT_SECONDS = 120;
+
     private final Logger log;
 
     private final CassandraKeyValueServiceConfig config;
     private final CassandraClientPool clientPool;
 
     private SchemaMutationLock schemaMutationLock;
-    private final Optional<LeaderConfig> leaderConfig;
-
+    private final ExecutorService schemaMutationExecutor;
     private final UniqueSchemaMutationLockTable schemaMutationLockTable;
+
+    private final Optional<LeaderConfig> leaderConfig;
 
     private ConsistencyLevel readConsistency = ConsistencyLevel.LOCAL_QUORUM;
     private final ConsistencyLevel writeConsistency = ConsistencyLevel.EACH_QUORUM;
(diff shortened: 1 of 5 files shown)


sandorw:
Fixes https://github.com/palantir/atlasdb/issues/3122

sandorw:
Remaining test failure is due to the exception wrapping caused by using the executor service, in particular:

`
    @Test
    public void dropTableThrows() {
        assertThat(OneNodeDownTestSuite.kvs.getAllTableNames()).contains(OneNodeDownTestSuite.TEST_TABLE_TO_DROP);
        assertThatThrownBy(() -> OneNodeDownTestSuite.kvs.dropTable(OneNodeDownTestSuite.TEST_TABLE_TO_DROP))
                .isExactlyInstanceOf(AtlasDbDependencyException.class)
                .hasCauseInstanceOf(IllegalStateException.class);
        // This documents and verifies the current behaviour, dropping the table in spite of the exception
        // Seems to be inconsistent with the API
        assertThat(OneNodeDownTestSuite.kvs.getAllTableNames()).doesNotContain(OneNodeDownTestSuite.TEST_TABLE_TO_DROP);
    }
`

We could unwrap all exceptions from the `UncheckedExecutionException` if we want to keep current exception behavior, though it's not clear to me how important those exception types are for any clients.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1438,6 +1448,10 @@ public void createTables(final Map<TableReference, byte[]> tableNamesToTableMeta
         return filteredTables;
     }
 
+    private void runSchemaMutationTask(Runnable task) {
+        Futures.getUnchecked(schemaMutationExecutor.submit(task));
+    }
```
jeremyk-91: This will change the exception type of what is thrown (from `IllegalStateException` to `UncheckedExecutionException`) - so technically is a break though is probably okay. 

Either way we should document this in release notes / javadoc for `CassandraKeyValueServiceImpl` which has changed.

jeremyk-91: Yep, makes sense (this happens when Cassandra, an AtlasDB dependency, is down).

One last thing - JavaDoc for `createTable` / `createTables` / `dropTable` / `dropTables` in this class should be `@throws AtlasDbDependencyException` rather than `@throws IllegalStateException` after this change right?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1438,6 +1450,18 @@ public void createTables(final Map<TableReference, byte[]> tableNamesToTableMeta
...
+        try {
+            Futures.getUnchecked(schemaMutationExecutor.submit(task));
+        } catch (UncheckedExecutionException e) {
+            // If this wraps an AtlasDbDependencyException, rewrap and throw
+            if (e.getCause() instanceof AtlasDbDependencyException) {
+                throw Throwables.unwrapAndThrowAtlasDbDependencyException(e);
+            }
+            throw e;
+        }
+    }
```
jeremyk-91: Yep, makes sense (this happens when Cassandra, an AtlasDB dependency, is down).

One last thing - JavaDoc for `createTable` / `createTables` / `dropTable` / `dropTables` in this class should be `@throws AtlasDbDependencyException` rather than `@throws IllegalStateException` after this change right?

sandorw: Yep, thanks for catching

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1759,6 +1773,13 @@ private void insertRangeTombstones(CassandraClient client, Map<Cell, Long> maxTi
      */
     @Override
     public void close() {
+        schemaMutationExecutor.shutdown();
+        try {
+            schemaMutationExecutor.awaitTermination(SCHEMA_MUTATION_TASK_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            // Continue with further transaction manager clean-up
+            Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
gsheasby: Do we want to shut down the client pool in this case, too?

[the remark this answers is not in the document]
sandorw: This just sets the interrupt flag, so we do continue on to run the remaining clean-up methods.

PR #3156, opened 2018-05-10
get more info from logging on directory creation failure.

hsaraogi:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3156)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
@@ -16,17 +16,22 @@
 package com.palantir.atlasdb.config;
 
 import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.Set;
 
 import javax.validation.constraints.Size;
 
 import org.immutables.value.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
```diff
@@ -35,6 +40,8 @@
 @SuppressWarnings("DesignForExtension")
 public abstract class LeaderConfig {
 
+    private final Logger log = LoggerFactory.getLogger(LeaderConfig.class);
```
jeremyk-91: nit: Loggers are usually static

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
```diff
@@ -87,12 +94,25 @@ protected final void check() {
...
+    private boolean ensureDirectoryExists(File directory) {
+        if (learnerLogDir().exists()) {
+            return true;
+        }
+        try {
+            Files.createDirectories(Paths.get(directory.getPath()));
+            return true;
+        } catch (Throwable t) {
+            log.error("Could not create the directory {}", SafeArg.of("dirName", directory.getPath()), t);
+            return false;
```
jeremyk-91: this will make timelock blow up so error is legit 👍

PR #3170, opened 2018-05-17
Compaction fixes for internal LTS

gsheasby:
**Goals (and why)**: Make compaction thread more robust, to ensure that we don't leave compaction threads hanging around

- Sleep before first compaction, to avoid short-lived CLIs etc trying to compact
- Ensure we shut down the compaction thread when the system shuts down, and that we handle interrupts in this process correctly

**Concerns (what feedback would you like?)**: No real concerns, a lot of this was copied from sweep code on develop

**Where should we start reviewing?**: The code =)

**Priority (whenever / two weeks / yesterday)**: ASAP, blocking multiple internal product LTSes

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3170)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
@@ -17,6 +17,7 @@
 package com.palantir.atlasdb.compact;
 
 import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
```diff
@@ -88,8 +91,14 @@
     public void close() {
         log.info("Closing BackgroundCompactor");
         daemon.interrupt();
+        // Ensure we do not accidentally abort shutdown if any code incorrectly swallows InterruptedExceptions
+        // on the daemon thread.
+        shuttingDown.countDown();
         try {
             daemon.join();
+            if (daemon.isAlive()) {
```
sandorw: From join docs, it seems like this shouldn't occur. I do see that we have similar code in place for sweep though.

PR #3171, opened 2018-05-17
[TS] Part 12: Transaction Table Lookups And Abort/Cleanup

gmaretic:
**Goals (and why)**:
Implement the logic for aborting in-progress transactions and cleanup of writes corresponding to aborted transactions.
More tests incoming.

Use a similar method for checking as old sweep uses, except that we batch the lookup for all non cached transactions to hit the kvs only once. Then do the usual aborting logic and delete all writes from aborted transactions.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3171)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -131,20 +144,33 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
         SweepableCellsTable.SweepableCellsRow row = computeRow(partitionFine, shardStrategy);
         RowColumnRangeIterator resultIterator = getRowColumnRange(row, partitionFine, minTsExclusive, maxTsExclusive);
 
-        Map<CellReference, WriteInfo> writes = new HashMap<>();
+        Multimap<Long, WriteInfo> startTsWrites = HashMultimap.create();
```
gsheasby: nit: I would prefer `writesByStartTs`. Would you?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -131,20 +144,33 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
         SweepableCellsTable.SweepableCellsRow row = computeRow(partitionFine, shardStrategy);
```
gsheasby: `getBatchForPartition` now does the following:
1. Fetches the values from the `SweepableCellsTable`
2. Groups the writes by start timestamp, filtering out those that were not committed
3. Bundles the writes that were committed into a `SweepBatch`

This seems like too many things for one method, and I just took ten minutes trying to understand it all. However, extracting methods might be tricky, because we compute the last swept ts along the way. Might be worth doing though, if it's not too gnarly (don't introduce `Pair<lastSweepTs, writes>` for example).

If it would not be straightforward or nice to cut this method down to size, then adding a few comments would make it more readable.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/AbstractSweepQueueTablesTest.java
```diff
@@ -86,37 +94,70 @@ public int writeToDefault(KvsSweepQueueWriter queueWriter, long timestamp, Table
...
         queueWriter.enqueue(ImmutableList.of(write));
         return write.toShard(numShards);
     }
 
-    public static List<WriteInfo> writeToCellsInFixedShard(KvsSweepQueueWriter writer, long timestamp, int number,
+    protected void putTimestampIntoTransactionTable(long ts, long commitTs) {
+        try {
+            txnService.putUnlessExists(ts, commitTs);
+        } catch (KeyAlreadyExistsException e) {
+            // this is fine
```
gsheasby: not if you're trying to put `commitTs = FAILED_COMMIT_TS` and the key there is a legit commit timestamp. Probably want to check and fail if that's the case.

We don't care, in this case, if we're putting some different commit ts where they're both non-failed (but by construction I don't think we'd ever do this).

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueTest.java
```diff
@@ -455,4 +466,12 @@ private void assertNoEntriesInSweepableTimestampsBeforeSweepTimestamp() {
                 .nextSweepableTimestampPartition(ShardAndStrategy.conservative(CONS_SHARD), -1L, sweepTs))
                 .isEmpty();
     }
+
+    protected void putTimestampIntoTransactionTable(long ts, long commitTs) {
+        try {
+            txnService.putUnlessExists(ts, commitTs);
+        } catch (KeyAlreadyExistsException e) {
+            // this is fine
```
gsheasby: No need to change this.

The duplication makes me twitch a little bit, but the comment I wrote above doesn't apply here - in this case we don't care because the commit ts is always the same.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepableCellsTest.java
```diff
@@ -211,8 +254,16 @@ private SweepBatch readThorough(long partition, long minExclusive, long maxExclu
         return sweepableCells.getBatchForPartition(thorough(shardThor), partition, minExclusive, maxExclusive);
     }
 
+    private void assertDeleted(TableReference tableRef, Multimap<Cell, Long> expectedDeletes) {
```
gsheasby: 👍 smart

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gmaretic: Refactored

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/AbstractSweepQueueTablesTest.java
gmaretic: Fixed in follow-up PR

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepableCellsTest.java
gmaretic: 🥇

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gsheasby: Nice! 🎉

PR #3173, opened 2018-05-18
More logging for the serializabletxn verifyColumnRanges method

j-baker:
This method is toying with us. The implementation is now less efficient,
but until 2 weeks ago it never worked and now it only somewhat works.

**Goals (and why)**:

Deal with correctness issue.

Don't iterate, just copy so we can log better.

No real concerns

The lines that got changed

Yesterday

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3173)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -79,6 +79,7 @@ import com.palantir.common.collect.IterableUtils;
 import com.palantir.common.collect.Maps2;
 import com.palantir.lock.v2.LockToken;
 import com.palantir.lock.v2.TimelockService;
+import com.palantir.logsafe.UnsafeArg;
 import com.palantir.util.Pair;
 
 /**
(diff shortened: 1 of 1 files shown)


gmaretic:
LGTM

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -638,9 +639,20 @@ private void verifyColumnRanges(Transaction readOnlyTransaction) {
...
                             input -> ByteBuffer.wrap(input));
-                    boolean isEqual = bv.transformBatch(input -> filterWritesFromCells(input, writes))
-                            .isEqual(readsInRange.entrySet());
+                    List<Entry<Cell, ByteBuffer>> realValuesAsList =
+                            bv.transformBatch(input -> filterWritesFromCells(input, writes)).immutableCopy();
+                    List<Entry<Cell, ByteBuffer>> readValuesAsList = ImmutableList.copyOf(readsInRange.entrySet());
+                    boolean isEqual = realValuesAsList.equals(readValuesAsList);
                     if (!isEqual) {
+                        log.info("Failing a serializable transaction because column ranges not equal",
+                                UnsafeArg.of("table", table),
```
tboam: You can use `LoggingArgs...` here to make this safe if you want, not essential though

j-baker: oh interesting, I knew there was a class like that. Meh at this point I want it merged fast enough that a rebuild is annoying.

PR #3174, opened 2018-05-18
Fix the ordering bug

j-baker:
Oh god oh god oh god oh god oh god

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -27,6 +27,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
@@ -493,7 +494,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 if (raw.isEmpty()) {
                     return endOfData();
                 }
-                Map<Cell, byte[]> post = new LinkedHashMap<>();
+                SortedMap<Cell, byte[]> post = new TreeMap<>();
                 getWithPostFiltering(tableRef, raw, post, Value.GET_VALUE);
                 batchIterator.markNumResultsNotDeleted(post.keySet().size());
                 return post.entrySet().iterator();
--- a/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
+++ b/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.atlasdb.transaction.impl;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasItem;
@@ -112,6 +113,7 @@ import com.palantir.atlasdb.transaction.api.TransactionReadSentinelBehavior;
 import com.palantir.common.base.AbortingVisitor;
 import com.palantir.common.base.AbortingVisitors;
 import com.palantir.common.base.BatchingVisitable;
+import com.palantir.common.base.BatchingVisitableView;
 import com.palantir.common.concurrent.PTExecutors;
 import com.palantir.common.proxy.MultiDelegateProxy;
 import com.palantir.lock.AtlasRowLockDescriptor;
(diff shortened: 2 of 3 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -975,6 +977,43 @@ public void noWritesAddedToSweepQueueOnException() {
...
+        byte[] value = new byte[1];
+
+        serializableTxManager.runTaskWithRetry(tx -> {
+            tx.put(TABLE, ImmutableMap.of(firstCell, value, secondCell, value));
+            return null;
+        });
+
+        // this will write into the DB, because the protocol demands we write before we get a commit timestamp
+        RuntimeException conditionFailure = new RuntimeException();
+        assertThatThrownBy(() ->  serializableTxManager.runTaskWithConditionWithRetry(() -> new PreCommitCondition() {
```
sandorw: note that doing this with retry is likely overkill, but it doesn't matter

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -975,6 +977,43 @@ public void noWritesAddedToSweepQueueOnException() {
...
+        })).isSameAs(conditionFailure);
+
+        List<Cell> cells = serializableTxManager.runTaskReadOnly(tx ->
+                BatchingVisitableView.of(tx.getRowsColumnRange(
+                        TABLE,
+                        ImmutableList.of(row),
+                        BatchColumnRangeSelection.create(null, null, 10)).get(row))
+                        .transform(Map.Entry::getKey)
+                        .immutableCopy());
+        assertEquals(ImmutableList.of(firstCell, secondCell), cells);
```
sandorw: I assume this failed pre-fix?

gsheasby: I checked out the branch verified that it did. ✅

docs/source/release_notes/release-notes.rst
```diff
@@ -42,6 +42,22 @@ develop
...
+.. list-table::
+    :widths: 5 40
+    :header-rows: 1
+
+    *    - Type
+         - Change
+
+    *    - |fixed|
+         - Snapshot transaction is now guaranteed to return getRowsColumnRange results in the correct order.
+           Previously while paging over row dynamic columns, if uncommitted or aborted transaction data was
```
uschi2000: why would aborted data ever show up at all?

uschi2000: @j-baker

sandorw: When we read data from the KVS, we scan backwards from the maximum possible Atlas timestamp until we find a value - this is how reads and conflict detection work. Transactions necessarily write their data before they commit, so KVS reads can return data that belongs to another transaction, which could be committed or uncommitted. Aborting a transaction means marking in the _transactions table that it did not (and cannot) commit, but it's not possible to delete all the data written by that transaction at the time of the abort for a variety of reasons (don't know all the places that transaction wrote, all necessary KVS nodes may not be up), so transactions in general may happen across invalid data. 

As implemented, whenever an invalid value is found, we do another batch of reads to continue scanning for a valid value at a lower timestamp. Unfortunately, any value where we had to do additional iterative lookups ended up behind the other results.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -493,7 +493,7 @@ protected void checkGetPreconditions(TableReference tableRef) {
                 if (raw.isEmpty()) {
                     return endOfData();
                 }
-                Map<Cell, byte[]> post = new LinkedHashMap<>();
+                SortedMap<Cell, byte[]> post = new TreeMap<>();
                 getWithPostFiltering(tableRef, raw, post, Value.GET_VALUE);
```
mcintyret: I've come across cases in hot code where building a TreeMap from almost-sorted data can cause significant perf issues (since the red-black tree is constantly rebalancing). It _may_ be worth collecting these in a list and then sorting at the end - java default to TimSort which is particularly good on almost-sorted data. Of course whether this is worthwhile depends on the scale of the data, how hot the code is, and how out-of-order the normal case is.

gsheasby: How large would the result set need to be for these perf issues to become apparent? We have a pretty decent suite of performance tests, and could add a test case to expose the large-but-almost-sorted case.

mcintyret: Quick and dirty benchmark:
```
    public static void main(String[] args) {
        for (int i = 10; i < 1000000000; i *= 10) {
            runTest(i);
            System.out.println();
        }
    }

    static class CountingComparator implements Comparator<String> {
        int comparisons = 0;

        @Override
        public int compare(String o1, String o2) {
            comparisons++;
            return o1.compareTo(o2);
        }
    }

    private static void runTest(int howMany) {
        List<String> list = new ArrayList<>(howMany);
        for (int i = 0; i < howMany; i++) {
            list.add(RandomStringUtils.random(20));
        }

        list.sort(null);
        // now make 'almost sorted'
        int numSwaps = (int) Math.ceil(Math.log(howMany));
        for (int i = 0; i < numSwaps; i++) {
            swapRandom(list);
        }
        System.out.println("Testing with " + howMany + "; num swaps: " + numSwaps);

        List<String> listToSort = new ArrayList<>(list);
        CountingComparator listComp = new CountingComparator();
        long listSortTimeStart = System.nanoTime();
        listToSort.sort(listComp);
        long timeMicros = (System.nanoTime() - listSortTimeStart) / 1000;
        System.out.println("List: " + timeMicros + "µs; " + listComp.comparisons + " comparisons");

        CountingComparator treeComp = new CountingComparator();
        long treeSortTimeStart = System.nanoTime();
        TreeSet<String> tree = new TreeSet<>(treeComp);
        tree.addAll(list);
        long timeMicrosTree = (System.nanoTime() - treeSortTimeStart) / 1000;
        System.out.println("Tree: " + timeMicrosTree + "µs; " + treeComp.comparisons + " comparisons");

    }

    private static void swapRandom(List<String> list) {
        int first = ThreadLocalRandom.current().nextInt(list.size());
        int second = ThreadLocalRandom.current().nextInt(list.size());
        String atFirst = list.set(first, list.get(second));
        list.set(second, atFirst);
    }
```
gave these results:
```
Testing with 10; num swaps: 3
List: 222µs; 20 comparisons
Tree: 1096µs; 27 comparisons

Testing with 100; num swaps: 5
List: 260µs; 403 comparisons
Tree: 563µs; 785 comparisons

Testing with 1000; num swaps: 7
List: 6362µs; 1407 comparisons
Tree: 3387µs; 14411 comparisons

Testing with 10000; num swaps: 10
List: 658µs; 10886 comparisons
Tree: 7446µs; 211178 comparisons

Testing with 100000; num swaps: 12
List: 10479µs; 101214 comparisons
Tree: 42841µs; 2772129 comparisons

Testing with 1000000; num swaps: 14
List: 61146µs; 1002123 comparisons
Tree: 422747µs; 34378259 comparisons

Testing with 10000000; num swaps: 17
List: 866504µs; 10002954 comparisons
Tree: 9250023µs; 410639440 comparisons
```
So the difference becomes pretty significant >10k. It looks like the tree does far more comparisons that the list, so I imagine it would be fairly sensitive to the complexity of the comparison.

gsheasby: Interesting. I'd guess that we typically have 1000 or fewer results. It's a 1ms improvement at that level, which in the context of the whole transaction is not very much at all. If we get beyond 10k results then we start to have other problems...

sandorw: If there isn't a downside, we should make the change. 1 ms not being an improvement can and should change as we work on transaction overheads to improve performance.

PR #3176, opened 2018-05-18
Fix LoggingArgs NPE

gmaretic:
**Goals (and why)**:
Fix the issue arising from `SnapshotTransaction.getCommitTimestamps` hitting this method with a null TableReference causing an NPE.

We deal explicitly with the case where the argument is null

LoggingArgs should never throw exceptions, being an logging utility class, and this is the second such issue. Are there any other methods that can end up biting us?
Also, we need better test coverage!

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3176)


gmaretic:
Gokcan fixes this in another PR

PR #3177, opened 2018-05-18
Revert extra logging

j-baker:
Reverts palantir/atlasdb#3173

Assuming that the fix fixed the issue, no need to log all the time again.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3177)


sandorw:
This is just reverting the diagnostic PR that added extra logging.

PR #3178, opened 2018-05-21
Jbaker/delete some code

j-baker:
**Goals (and why)**:

The more code that is in Atlas, the harder it is to maintain. At the moment, we can see that AtlasDB is hard to maintain.

There are two getRowsColumnRange overloads. One of them is fine. This one however, is

1. **Broken**. No one bothered to implement it on serializable transactions, so it breaks AtlasDB's consistency guarantees. Like, you can literally corrupt your DB by using this method.
2. **Fairly complex**. It takes a while to understand this method. It's hard to implement this method correctly for serializable transactions, which is presumably why nobody bothered to not inject that bug.
3. **Largely untested**. This makes it very hard to maintain. We found a correctness bug in a very related method _just last week_.
4. **Redundant**. It only exists for performance reasons. I'm currently running the performance benchmarks in order to prove whether it's useful or not; I would imagine not but will circle back.
5. **Low value**. So far as I can see, there is exactly one use of this thing internally. If hundreds of lines of code has only one internal usage, it's a maintenance sink.

Alternative approaches:

We could just UnsupportedOperationException this when you call it on serializable txn. However, I would suggest that the value of reducing our API surface is high, and that we should take this opportunity to do it, assuming that the perf costs are not obscene.

Delete method, + the codegen that generates it.

- Might aggressively affect perf. I'll circle back with number comparisons.

Probably the bits I didn't delete? There are some benchmarks I changed to use the other method of dynamic columns.

Few days? It's not a hard review.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3178)


j-baker:
Well, on the plus side the benchmark is 1000 times faster in Cassandra. On the minus side it crashes on Oracle with this error, which I maintain is hilarious

image

sandorw:
See also https://github.com/palantir/atlasdb/pull/3200

sandorw:
Closing since there was a sizable perf difference

PR #3179, opened 2018-05-21
Chain where and how a client is invalidated.

AmlingPalantir:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3179)


gsheasby:
Unforked in #3189.

PR #3182, opened 2018-05-22
lower log level to debug for silent timelock debugging messages

gokcanozakdag:
**Goals (and why)**:
Reduce the logging noise by lowering log level to debug for the messages we used to debug silent timelock issue.

This week would be nice.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3182)

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -235,7 +235,7 @@ public final class PaxosQuorumChecker {
             }
         }, OUTSTANDING_REQUEST_CANCELLATION_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
 
-        if (shouldLogCancellationStatus()) {
+        if (log.isDebugEnabled() && shouldLogCancellationStatus()) {
             log.debug("Quorum checker canceled pending requests"
                     + ". Rate of successful cancellations: {}, rate of no-op cancellations: {}",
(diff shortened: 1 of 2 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -236,7 +236,7 @@ private PaxosQuorumChecker() {
         }, OUTSTANDING_REQUEST_CANCELLATION_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
 
         if (shouldLogCancellationStatus()) {
-            log.warn("Quorum checker canceled pending requests"
+            log.debug("Quorum checker canceled pending requests"
```
jeremyk-91: Does it make sense to add a `log.isDebugEnabled()` above? I don't know how `getOneMinuteRate()` is implemented, but if that is actually costly then we should avoid calculating it if we don't have to

snegrus-adb: Was there an offline discussion related to this question?

gokcanozakdag: I have changed it to check `log.isDebugEnabled()`. Metrics shouldn't be costly, still checking if debug enabled is easy enough. Also, it prevents generating a random number every time (by checking `isDebugEnabled` before `shouldLog` which generates a random value to sample logs)

snegrus-adb: cool, thanks!

PR #3185, opened 2018-05-23
Row paging hands back in hashmap order from the second page and beyond

j-baker:
**Goals (and why)**: Contract of KeyValueService (and real life code) dictates that cells be handed back ordered lexicographically. In practice, whilst this is true for the first page, subsequent are handed back in HashMap order.

Change the method that returns the second-and-beyond-page to use the RowColumnRangeExtractor not the RowExtractor.

I'd like this merged and released ASAP, since internally we have stuff blocking on this.

If you want background context, the single host version of the getRowsColumnRange method is sensible place to start.

Gist is that the algorithm is:

Do a batch call to get all the columns, limit foo. See if the last column returned indicates to us that there are no more, and if so, that's it. Otherwise, create a lazy batching iterator that will read the second page and beyond.

The codepaths for getting the first page and getting the second page are sadly different, but I've unified them a little bit here.

Yesterday

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3185)


j-baker:
Oh and BTW, I already validated that this fails, so no need to check out the code.

j-baker:
ValueExtractor is great - it's just not good for this

j-baker:
it's designed for the getRows case wher eordering doesn't matter

PR #3188, opened 2018-05-23
TxnManager refactors

sandorw:
**Goals (and why)**:
Make progress on cleaning up the `TransactionManager` inheritance tree. Currently, there's no central interface that includes the relevant methods supported by `SerializableTransactionManager`, so it's very difficult to wrap it to add functionality. For example, we have several workflows where we'd like to transparently wrap all tasks on a `TransactionManager` with additional locking or pre-commit conditions. Extending `SerializableTransactionManager` doesn't work due to the number of arguments in the constructor and the need to duplicate the code in `TransactionManagers`.

- Condense all relevant methods to the `TransactionManager` interface, including the removal of `ConditionAwareTransactionManager` and `LockAwareTransactionManager`
- Use auto delegate to replace `ForwardingTransactionManager`
- `TransactionManagers` returns a `TransactionManager` instead of a `SerializableTransactionManager` and automatically instruments it. Most internal references now pass the interface class rather than the implementation class.
- Consistent naming on the runTask methods (`runTaskReadOnlyWithCondition` -> `runTaskWithConditionReadOnly`).

- Do we anticipate dev breaks that we won't be able to resolve? 
- Any further changes that we should bundle with this dev break?
- For `WrappingTransactionManager`, did I appropriately wrap all of the relevant methods?

`TransactionManager.java`

soon

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3188)


sandorw:
Immediate follow-ups:
 - Should likely instrument in TransactionManagers

Potential future follow-up PRs:
 - Combine AbstractTM, AbstractConditionAwareTM, and AbstractLockAwareTM
 - Combine SnapshotTM and SerializableTM together
 - Possibly combine all of the above into one class
 - Consider a transaction factory that can handle all WrappingTM behavior

gsheasby:
:lgtm:

---

Reviewed 108 of 112 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java, line 266 at r2](https://reviewable.io/reviews/palantir/atlasdb/3188#-LDGgG7J19XxxY4fLX8N:-LDGgG7J19XxxY4fLX8O:b-90pc0z) ([raw file](https://github.com/palantir/atlasdb/blob/6db36241dae646734f96dd853c19de5b637e24a4/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java#L266)):*
> ```Java
> 
>     /**
>      * Returns the KVS used by this transaction manager. In general, this should not be used by clients, as
> ```

I wondered: Why is it exposed here? If it's for testing, could it be moved to `TestTransactionManager`?

Then I saw it used in SnapshotTransactionManager. That's a bit sad, but beyond the scope of this already-hefty PR.

---

*[atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ReadOnlyTransactionManager.java, line 215 at r2](https://reviewable.io/reviews/palantir/atlasdb/3188#-LDGiQA_-ubZbOBBgX5q:-LDGiQA_-ubZbOBBgX5r:b-uo4kvs) ([raw file](https://github.com/palantir/atlasdb/blob/6db36241dae646734f96dd853c19de5b637e24a4/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ReadOnlyTransactionManager.java#L215)):*
> ```Java
> 
>     @Override
>     public TimestampService getTimestampService() {
> ```

These seem somewhat dangerous.

---


*Comments from [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3188#-:-LDH-d7S3HdtjPcmuEbi:bnfp4nl)*

gsheasby:
Would be good to compile against the auth product, too.

jeremyk-91:
@gsheasby getTimestampService and friends are already used pretty widely (for generation of unique IDs, event queue, etc.) so might be a bit tricky to remove access to that.

At least we don't expose `TimestampManagementService` which would be really fun

sandorw:
I would prefer not to expose most of the stuff that's bolted on to `TransactionManager`, but it's an artifact of minimizing breaks for existing codepaths (`SerializableTransactionManager`s have been providing them all for ages) and the fact that `TransactionManagers` doesn't provide access to these intermediate building blocks (KVS, timelock, etc.) for the cases where they are needed. If we want to stomach further changes that restrict the `TransactionManager` interface down but provide access to those classes, I'd be game, assuming the further dev breaks are manageable.

As for `registerClosingCallback`, I'd prefer to expose that only at creation time in some way on `TransactionManagers`, but we have at least one example internally where the callback requires the transaction manager itself (similar chicken and egg problem to async initialization callbacks). 

Probably worth discussing offline what the options are for cutting down the interface much more and seeing if they're doable.

sandorw:
Once we've solidified, I'll check building against the internal projects specified above. I won't merge until we've resolved the dev breaks and are ready to help any others start rolling it out - I don't want this to sit on develop and prevent us from releasing with an important bugfix.

sandorw:
I'll add the concurrency note from the `registerClosingCallback` docs on the implementation. Is that sufficient, @jeremyk-91, or do you think we should do more work on it (analogous to txn manager async init callbacks)?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -212,16 +212,16 @@ boolean allowHiddenTableAccess() {
      *
      * @see TransactionManagers#createInMemory(Set)
      */
-    public static SerializableTransactionManager createInMemory(Schema schema) {
+    public static TransactionManager createInMemory(Schema schema) {
```
j-baker: you fancy deleting the no-args constructor of SerializableTransactionManager?

sandorw: This is currently used for `AutoDelegate_SerializableTM`, which is used once internally for the async init wrapper and once externally in a case where they want to change the `createTransaction` behavior but need access to the ~75 member variables. The second case can be switched to `AutoDelegate_TM` instead. I have to dig in some to see if we can fix the init wrapper usage.

sandorw:
Should be good to review the last additions @gsheasby @jeremyk-91 @j-baker, though not much substantive has changed. I may have to reopen the PR, since the Palantir membership check seems to be broken.

sandorw:
Will try closing and reopening to see if that fixes the membership issue, otherwise I'll just open a new PR and merge there.

PR #3189, opened 2018-05-23
Chain where and how a client is invalidated.

gokcanozakdag:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3189)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientImpl.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


gokcanozakdag:
Original pr: https://github.com/palantir/atlasdb/pull/3179
Opened it as a branch in atlas repo, to be able to run tests. Original pr looks good to me, but I cannot add myself as a reviewer.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientImpl.java
```diff
@@ -282,16 +282,21 @@ private void executeHandlingExceptions(ThrowingVoidSupplier supplier)
     }
 
     private void updateIsValid(Exception e) {
```
gsheasby: Was wondering if we need to `synchronise` this method.

Turns out we don't, since `compareAndSet` simply returns false if we've lost a race with another thread, and attempting to invalidate something twice has the same result as attempting it once.

PR #3190, opened 2018-05-23
Fix/remove jacoco

gsheasby:
**Goals (and why)**: Remove jacoco, because we don't use it and it added an extra 5 mins to our biuld

**Implementation Description (bullets)**: Removed jacoco from gradle/yml

**Concerns (what feedback would you like?)**: Does anyone actually use this?

**Where should we start reviewing?**: Anywhere

**Priority (whenever / two weeks / yesterday)**: whenever you want your builds to be faster :-)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3190)


sandorw:
Looks like that's all the references to jacoco, and the builds took 13 minutes!

PR #3193, opened 2018-05-23
Bump remoting3

hsaraogi:
**Goals (and why)**: Causes conflicts with the latest WC version in upstream products. Alta is a more longer term solution to avoid this.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3193)


sandorw:
Closing out since another PR merged with this bump

PR #3198, opened 2018-05-25
Findbugs annotations instead of findbugs:annotations

j-baker:
Findbugs:annotations is this fairly lame megajar which contains all
the classes in findbugs-annotations and jsr305. Internally we've
roughly standardized on the one and not the other (and the migration
is easy) so just use that one.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3198)


j-baker:
Also, @sandorw would you consider a PR that removes all the lock files?

This PR's addition of qos-api to atlasdb api merely indicates that no one regenerated the lock files for 6 months, no?

gsheasby:
// assigning to @jeremyk-91 for load-balancing. The diff looks intimidating, but it's mostly lock files. Given the relative priority, I wouldn't expect this to be done this week.

atlasdb-perf/build.gradle
```diff
@@ -7,6 +7,14 @@ apply plugin: 'application'
 mainClassName = 'com.palantir.atlasdb.performance.cli.AtlasDbPerfCli'
 applicationName = 'atlasdb-perf'
 
+configurations.all {
+    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        if (details.requested.group == 'com.google.code.findbugs' && details.requested.name == 'annotations') {
+            details.useTarget group: 'com.google.code.findbugs', name: 'findbugs-annotations', version: '3.0.1'
+        }
+    }
+}
```
jeremyk-91: OK, a bit nasty but it looks like these annotations are a transitive of some of the dependencies.

PR #3199, opened 2018-05-25
Don't greedily delete data from aborted txns

gsheasby:
This caused reads and writes to become extremely slow (but still
eventually pass) when a database node is offline.

Potential fixes:
- Add the rolled-back data to the sweep queue
- Delete the rolled-back data asynchronously

**Goals (and why)**: Make AtlasDB (performantly) HA (P-HA) Again.
Reads and writes were previously not P-HA when aborted transaction data existed. This caused a P0 with the internal Megatron product.

- Remove the try block where we were attempting to delete the data
- Note that we already (correctly) abort the transaction earlier in that same method, so no other change was necessary

**Concerns (what feedback would you like?)**: Why didn't this change break any tests? We should also track adding this cleanup step back in future, but possibly not until Targeted Sweep is a thing

**Where should we start reviewing?**: One java file

**Priority (whenever / two weeks / yesterday)**: Yesterday. In fact, like last year, because that's when we started promising to be HA.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3199)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1696,26 +1696,10 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             }
         }
 
-        try {
-            log.debug("For table: {} we are deleting values of an uncommitted transaction: {}",
-                    LoggingArgs.tableRef(tableRef),
-                    UnsafeArg.of("keysToDelete", keysToDelete));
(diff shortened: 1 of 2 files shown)


sandorw:
Let's make sure to file a ticket tracking possible re-addition with an implementation that's async with respect to the thread committing the transaction (after the synchronous abort).

We should also check with certain internal projects to be sure that this behavior change isn't a problem for them.

gsheasby:
If we're concerned about internal products, we could always hide the behaviour behind a config flag (defaulted to don't-delete, but set to greedy-delete for the internal product). I've pinged those responsible to find out if it's necessary.

@dxiao for SA.

Issue #3172, opened 2018-05-17
Serializable isolation seems to be broken by getRowsColumnRange


j-baker:
There are two methods in SnapshotTransaction called getRowsColumnRange. One is overridden by SerializableTransaction. The other is not. I would tentatively say that the other one does not properly enforce serializable isolation, because so far as I can see there's no mechanism by which it would be.

PR #3200, opened 2018-05-25
Throw for unsupported SerializableTM method

sandorw:
**Goals (and why)**:
Address https://github.com/palantir/atlasdb/issues/3172

throw UOE if someone attempts to use this method on a table with a serializable conflict handler

next week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3200)


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -532,6 +534,27 @@ public void testColumnSelection2() {
...
+        } catch (UnsupportedOperationException e) {
+            // expected
+        }
+    }
+
+    @Test
+    public void testColumnRangeReadSupported() {
+        Transaction t1 = startTransaction();
+        // The transactions table is registered as IGNORE_ALL, so the request is supported
+        // Reading at a negative timestamp to avoid any repercussions for in-flight transactions
```
jeremyk-91: 🥇I like this!

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -532,6 +534,27 @@ public void testColumnSelection2() {
                 RangeRequest.builder().retainColumns(ImmutableList.of(PtBytes.toBytes(col))).build());
     }
 
+    @Test
+    public void testColumnRangeReadUnsupported() {
+        Transaction t1 = startTransaction();
+        try {
+            t1.getRowsColumnRange(TEST_TABLE, ImmutableList.of(PtBytes.toBytes("row1")),
+                    new ColumnRangeSelection(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY), 1);
```
jeremyk-91: There needs to be a `fail()` after this line, otherwise not throwing also passes

PR #3204, opened 2018-05-29
[TS] Part 19: Hash Metadata to avoid hotspotting

jeremyk-91:
**Goals (and why)**:
- Avoid hotspotting when writing to dedicated rows (currently everything on the same shard goes to the same node during a `FINE_PARTITION_TS` time-block)

- Change `SweepableCells` to hash the first two row components
- Change `deleteDedicatedRows` to perform deletions on a per-row basis as opposed to across a row range (see first point under Concerns for details)

- The main tradeoff here is that multiple dedicated rows in a single transaction were previously deleted as a single `RangeRequest` in a single call to `deleteRange`, but now we have to delete them one row at a time (because hashing metadata means they'll no longer be contiguous). Given these deletes take place on a background thread and the current impl of `deleteRange` basically will make multiple delete calls to Cassandra anyway (we don't do something very clever) I think this is fine.
- Is the performance cost of hashing the metadata significant? I'd guess not.
- Are the tests too specific?

**Where should we start reviewing?**: `TargetedSweepSchema`, but the real meat is in `SweepableCells` which has the fiddly change in range deletions.

**Priority (whenever / two weeks / yesterday)**: ideally by EOD tomorrow

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3204)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepableCellsTable.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepableCellsTable.java
@@ -55,7 +55,6 @@ import com.palantir.atlasdb.keyvalue.api.Prefix;
 import com.palantir.atlasdb.keyvalue.api.RangeRequest;
 import com.palantir.atlasdb.keyvalue.api.RowResult;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
-import com.palantir.atlasdb.keyvalue.api.WriteReference;
 import com.palantir.atlasdb.keyvalue.impl.Cells;
 import com.palantir.atlasdb.ptobject.EncodingUtils;
 import com.palantir.atlasdb.table.api.AtlasDbDynamicMutablePersistentTable;
@@ -147,7 +146,7 @@ public final class SweepableCellsTable implements
         private final byte[] metadata;
 
         public static SweepableCellsRow of(long timestampPartition, byte[] metadata) {
-            long hashOfRowComponents = computeHashFirstComponents(timestampPartition);
+            long hashOfRowComponents = computeHashFirstComponents(timestampPartition, metadata);
             return new SweepableCellsRow(hashOfRowComponents, timestampPartition, metadata);
         }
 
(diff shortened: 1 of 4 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepableCellsTable.java
```diff
@@ -205,9 +204,10 @@ public SweepableCellsRow hydrateFromBytes(byte[] __input) {
             }
         };
 
-        public static long computeHashFirstComponents(long timestampPartition) {
+        public static long computeHashFirstComponents(long timestampPartition, byte[] metadata) {
             byte[] timestampPartitionBytes = EncodingUtils.encodeUnsignedVarLong(timestampPartition);
-            return Hashing.murmur3_128().hashBytes(EncodingUtils.add(timestampPartitionBytes)).asLong();
+            byte[] metadataBytes = metadata;
```
sandorw: nit: just use metadata directly below

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepableCellsTable.java
```diff
@@ -375,19 +375,19 @@ public int compareTo(SweepableCellsColumn o) {
      *   {@literal Long writeIndex};
      * }
      * Column value description {
-     *   type: com.palantir.atlasdb.keyvalue.api.TableReferenceAndCell;
+     *   type: com.palantir.atlasdb.keyvalue.api.WriteReference;
      * }
      * </pre>
      */
-    public static final class SweepableCellsColumnValue implements ColumnValue<WriteReference> {
+    public static final class SweepableCellsColumnValue implements ColumnValue<com.palantir.atlasdb.keyvalue.api.WriteReference> {
```
sandorw: Was this import drop intentional?

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepableCellsTable.java
sandorw: nvm, generated

jeremyk-91:
Tests are failing because the underlying ETE test branch is failing. I'll investigate that now, having reviewed 17

PR #3205, opened 2018-05-29
Adding coverage for putUnlessExists, with fix for NPE

gokcanozakdag:
**Goals (and why)**:
- Adding test coverage for the case when we try to commit a transaction that is already committed; possibly as a result of putUnlessExists retry logic. Also fixing NPE while logging the `tableRef`.

- Forcing timestampService to return same timestamp (`commitTs`) when fresh timestamp is requested. Putting a transaction with `startTs` and `commitTs` to `transactionService` before calling `commit` on transaction, which will cause `putUnlessExists` to throw `KeyAlreadyExists`.

- Is there a better way to test this case?
- Checking `tableRef != null` and logging accordingly seems a bit ugly to me, not sure if logging without tableRef is useful at all. If not, we can just remove that part, and log only if tableRef is not null.

- SnapshotTransaction.java

- Not high priority, P2.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3205)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1014,6 +1015,35 @@ public void cleanup() {}
         assertEquals(ImmutableList.of(firstCell, secondCell), cells);
     }
 
+    @Test
+    public void commitDoesNotThrowIfAlreadySuccessfullyCommitted() {
```
gsheasby: what should happen if we try to commit, but somebody has rolled us back - is that case covered?

gsheasby:
Next action: @gokcanozakdag to add the above mentioned test case

gokcanozakdag:
@gsheasby @sandorw I have revised the pr, can you take a final look?

docs/source/release_notes/release-notes.rst
```diff
@@ -66,6 +66,11 @@ develop
            transactions had significantly impaired performance if a database node was down.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3199>`__)
 
+    *    - |fixed|
+         - Fixed an issue occurring after a failure to putUnlessExists a commit timestamp that was causing an NPE leading to a confusing error message.
```
gsheasby: The release note should capture in which context we would run into this issue in the field.

gsheasby:
The new test failed :-(
```
[00:04:30] com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest > commitThrowsIfRolledBackAtCommitTime STARTED
[00:04:30] 
[00:04:30] com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest > commitThrowsIfRolledBackAtCommitTime FAILED
[00:04:30]     java.lang.AssertionError: 
[00:04:30]     Expecting:
[00:04:30]      <java.lang.AssertionError: BUG: Someone tried to roll back our transaction but our locks were still valid; this is not allowed. Held immutable timestamp lock: Optional.empty; held commit locks: LockToken{requestId=38450ec9-0151-bdc8-3e52-74c490b057f6}>
[00:04:30]     to be an instance of:
[00:04:30]      <com.palantir.atlasdb.transaction.api.TransactionCommitFailedException>
[00:04:30]     but was instance of:
[00:04:30]      <java.lang.AssertionError>
[00:04:30]         at com.palantir.atlasdb.transaction.impl.SnapshotTransactionTest.commitThrowsIfRolledBackAtCommitTime(SnapshotTransactionTest.java:1057)
[00:04:30] 
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1048,15 +1050,50 @@ public void commitDoesNotThrowIfAlreadySuccessfullyCommitted() {
...
+        transactionService.putUnlessExists(snapshot.getTimestamp(), TransactionConstants.FAILED_COMMIT_TS);
 
-        tx.put(TABLE, ImmutableMap.of(cell, PtBytes.toBytes("value")));
-        assertThatThrownBy(() -> tx.commit()).isInstanceOf(TransactionCommitFailedException.class);
+        snapshot.put(TABLE, ImmutableMap.of(cell, PtBytes.toBytes("value")));
+        assertThatThrownBy(() -> snapshot.commit()).isInstanceOf(TransactionLockTimeoutException.class);
     }
 
+
+
```
gsheasby: nit: triple-blank line

PR #3206, opened 2018-05-29
InterruptibleProxy tests and configurable ExecutorService

cakofony:
**Goals (and why)**:
commit 1: Tests should live in the same codebase as the implementation
commit 2: Reduce unnecessary thread creation for operations that take just over 100ms by providing a custom ExecutorService.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3206)

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/proxy/InterruptibleProxy.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/proxy/InterruptibleProxy.java
@@ -21,7 +21,9 @@ import java.lang.reflect.Proxy;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.function.Supplier;
 
+import com.google.common.base.Suppliers;
 import com.palantir.common.base.Throwables;
 import com.palantir.common.concurrent.NamedThreadFactory;
 import com.palantir.common.concurrent.PTExecutors;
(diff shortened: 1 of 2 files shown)


atlasdb-commons/src/main/java/com/palantir/common/proxy/InterruptibleProxy.java
```diff
@@ -47,7 +47,9 @@
 
     private final Object delegate;
     private static final ExecutorService executor = PTExecutors.newCachedThreadPool(
-            new NamedThreadFactory("Interruptible Proxy", true /* isDaemon */ ));
+            new NamedThreadFactory("Interruptible Proxy", true /* isDaemon */ ),
+            // Timeout idle threads after 5 seconds, up from the default of 100ms
+            5000);
```
schlosna: Chatted with @cakeofny this morning, might want to just inject executor in constructor as it simplifies the proxy and lets user control lifecycle of executor (right now nothing shuts this down)

schlosna: Separate from this PR we might want to look at executors across AtlasDB as 100ms default TTL for cached pools is quite short in general

atlasdb-commons/src/main/java/com/palantir/common/proxy/InterruptibleProxy.java
```diff
@@ -35,23 +35,28 @@
  *
  */
 public final class InterruptibleProxy implements DelegatingInvocationHandler {
+    private static final ExecutorService defaultExecutor = PTExecutors.newCachedThreadPool(
+            new NamedThreadFactory("Interruptible Proxy", true /* isDaemon */ ));
```
schlosna: Wondering if we should lazily create this as it really shouldn't be used directly, was thinking along lines:

```
    private static final Supplier<ExecutorService> defaultExecutor = Suppliers.memoize(
            () -> PTExecutors.newCachedThreadPool(new ThreadFactoryBuilder()
                    .setNameFormat("Interruptible-Proxy-%d")
                    .setDaemon(true)
                    .build()));
```

atlasdb-commons/src/main/java/com/palantir/common/proxy/InterruptibleProxy.java
```diff
@@ -35,23 +35,28 @@
...
 
-    private final CancelDelegate cancel;
+    public static <T> T newProxyInstance(Class<T> interfaceClass, T delegate, CancelDelegate cancel) {
```
schlosna: should probably `@Deprecated` the old constructor

atlasdb-commons/src/test/java/com/palantir/common/proxy/InterruptibleProxyTest.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        try {
+            proxy.remove(0);
+            fail("Should be interrupted");
+        } catch (PalantirInterruptedException e) {
+            // Should get here
+        }
+
+        barrier.await();
```
sandorw: It's unclear to me what the barrier is doing for us, since `gotInterrupted` will get set regardless of whether the proxy thread gets stuck waiting on the barrier.

schlosna:
👍 LGTM

PR #3207, opened 2018-05-29
[TS] Part 18+: Fix ETE Shenanigans?

jeremyk-91:
**Goals (and why)**:
- Fix the ETEs in #3203 

- Rename the test to end with `EteTest` (for why this is important, consider `atlasdb-ete-tests/build.gradle:67`)

- Did this actually fix the problem?

**Where should we start reviewing?**: +2 -2

**Priority (whenever / two weeks / yesterday)**: asap

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3207)


gmaretic:
Thanks!

PR #3209, opened 2018-05-30
[TS] Part 20: Thorough Sweep Deletes Sentinels

gmaretic:
**Goals (and why)**:
Have thorough sweep include sentinels in the ranged tombstone.

Added a kvs method that adds the appropriate ranged tombstones.

Is it worth bloating the API?

Anywhere, really

Now :)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3209)


sandorw:
Discussed offline, but the alternative to at least try briefly is for the CassandraKVS to look up/cache the strategy based on the passed in `TableReference`. It might not be worth it, but it wouldn't require the extra method if we can use that.

gmaretic:
Another alternative is to just break the current API and have an additional argument specifying whether to delete the sentinel or not. This may actually be the cleanest solution as I don't expect anyone was actually using `deleteAllTimestamps`

PR #3215, opened 2018-05-31
no longer blacklist connections due to a CAS error

j-baker:
**Goals (and why)**: We currently throw exceptions in the client pool on a check and set failure. This causes retries and blacklisting to be invoked, making those methods detrimental to overall performance and more importantly causing a lot of logspam.

**Implementation Description (bullets)**: Throw _outside_ the client pool.

**Concerns (what feedback would you like?)**: Not sure how to test this, given it's not observable behaviour at this level.

**Priority (whenever / two weeks / yesterday)**: Now?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3215)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -1866,18 +1866,21 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     public void putUnlessExists(final TableReference tableRef, final Map<Cell, byte[]> values)
             throws KeyAlreadyExistsException {
         try {
-            clientPool.runWithRetry(client -> {
+            Optional<KeyAlreadyExistsException> failure = clientPool.runWithRetry(client -> {
                 for (Entry<Cell, byte[]> e : values.entrySet()) {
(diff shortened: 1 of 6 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1866,19 +1866,24 @@ public void addGarbageCollectionSentinelValues(TableReference tableRef, Iterable
...
                     if (!casResult.isSuccess()) {
-                        throw new KeyAlreadyExistsException(
+                        return Optional.of(new KeyAlreadyExistsException(
                                 String.format("The row in table %s already exists.", tableRef.getQualifiedName()),
-                                ImmutableList.of(e.getKey()));
+                                ImmutableList.of(e.getKey())));
                     }
                 }
                 clientPool.markWritesForTable(values, tableRef);
                 return null;
```
jeremyk-91: should this be `Optional.empty()`?

j-baker: yeah, i switched to Optional accidentally...

circle.yml
```diff
@@ -1,7 +1,7 @@
 machine:
   java:
     version:
-      oraclejdk8
+      openjdk8
```
jeremyk-91: Saw your commit message - ok. Thanks

PR #3220, opened 2018-05-31
[TS] Targeted Sweep Fixes

gmaretic:
**Goals (and why)**:
Further reduce noisy logging, do not give when sweep throws, add test that verifies deleteAllTimestamps runs at consistency all.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3220)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
@@ -171,7 +171,7 @@ public class TargetedSweeper implements MultiTableSweepQueueWriter {
                 log.warn("Targeted sweep for {} failed and will be retried later.",
                         SafeArg.of("shardStrategy", shardStrategy.toText()), th);
             } finally {
-                shardsBeingSwept.remove(shardStrategy.shard());
+                unlockShard(shardStrategy.shard());
             }
         }
 
@@ -185,6 +185,10 @@ public class TargetedSweeper implements MultiTableSweepQueueWriter {
 
         private int getShardAndIncrement() {
             return queue.modShards(counter.getAndIncrement());
+        }
+
+        private void unlockShard(int shard) {
+            shardsBeingSwept.remove(shard);
         }
 
         @Override
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
@@ -17,6 +17,8 @@
 package com.palantir.atlasdb.sweep.queue;
 
 import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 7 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -149,13 +154,21 @@ private void scheduleBackgroundThreads() {
...
                 }
             }
         }
 
+        private void runOneIteration() {
+            ShardAndStrategy shardStrategy = ShardAndStrategy.of(getShardAndIncrement(), sweepStrategy);
+            try {
+                sweepNextBatch(shardStrategy);
+            } catch (Throwable th) {
+                log.error("Targeted sweep for {} failed and will be retried later.",
```
gsheasby: Are we guaranteed to try the next `shardStrategy` next time?

We could test it: something like:
`when(sweepNextBatch(any()).thenFail().thenReturn()`
followed by verifying that the arguments to both `sweepNextBatch()` calls are the same.

jeremyk-91: Actually this won't do that if I'm not wrong - we will try if we loop around.

I don't have a nice way of doing this offhand apart from having a queue of failed nodes to try again first, though that might be too much complexity for now?

---

Separately: should this be `warn`, since Background Sweep's analogous message (`The background sweep job failed unexpectedly`) is at warn?

gmaretic: Discussed offline -- we are retrying the shard when we come back around to it

gmaretic:
Does this fix for preventing threads from sweeping the same shard make sense?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -149,17 +158,39 @@ private void scheduleBackgroundThreads() {
...
+                unlockShard(shardStrategy.shard());
+            }
+        }
+
+        private int lockNextShardToSweep() {
+            int nextShardCandidate = getShardAndIncrement();
+            while (!shardsBeingSwept.add(nextShardCandidate)) {
+                nextShardCandidate = getShardAndIncrement();
+            }
+            return nextShardCandidate;
```
jeremyk-91: Discussed offline - might be a little awkward if we have more threads than shards, but we could at initialisation transform the numbers so things work

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -149,17 +158,39 @@ private void scheduleBackgroundThreads() {
...
 
+        private void runOneIteration() {
+            ShardAndStrategy shardStrategy = ShardAndStrategy.of(lockNextShardToSweep(), sweepStrategy);
+            try {
+                sweepNextBatch(shardStrategy);
+            } catch (Throwable th) {
+                log.warn("Targeted sweep for {} failed and will be retried later.",
+                        SafeArg.of("shardStrategy", shardStrategy.toText()), th);
+            } finally {
+                unlockShard(shardStrategy.shard());
```
jeremyk-91: 👍

PR #3224, opened 2018-06-01
Use TimestampService Timestamps For Cassandra multiPut() and delete()

jeremyk-91:
- [x] Add test coverage for new code paths
- [x] Add timestamp caching test

**Goals (and why)**:
- Use timestamp service timestamps for Cassandra `multiPut` and `delete`, to increase efficiency of compactions for tables which churn a lot
- Cut the knot mentioned in #3151 regarding initializing suitable `LongSupplier`s while minimising breaks elsewhere

- Cassandra mutation timestamps have changed.
  - Cassandra `multiPut` and `delete` now take place at a fresh timestamp from the timestamp service.
  - `put` and `putWithTimestamps` are unaffected
  - Atomic tables and `putUnlessExists` are unaffected
  - **Metadata table accesses use wall clock time.** Contentious, but I struggled with providing a timestamp service in some of the places that access the metadata table. It's usually relatively small so I don't think it's a major concern.
  - For some tests, I've added a fast forward of the current clock time up to 1 million or 1 billion, so that put and putWithTimestamps happen at legitimately possible points in time.
- Cassandra KVS now takes a `CassandraMutationTimestampSupplier` at startup, which is a pair of `Function<Long, Long>` (atlas write ts -> cass put ts, atlas delete ts -> cass tombstone ts).
  - Existing codepaths apart from `TMs.create()` can create a `legacy` supplier, which uses the existing mappings (put: `ts -> ts`, delete: `ts -> ts + 1`).
  - TM.create uses one that takes a `LongSupplier` and is basically `unused -> timestampService.getFreshTimestamp()`.
  - `CellValuePutter` and `CellDeleter` take `Function<Long, Long>` which are supplied from the `CassandraMutationTimestampSupplier` a given Cassandra KVS object might have.
- Atlas factories can now make use of a `LongSupplier`, which is unused in all implementations apart from `CassandraAtlasDbFactory`. This `LongSupplier` is initialized in `TransactionManagers.create()`. Other users, like the large internal product, won't have the `LongSupplier` (so if using Cassandra will continue to use the legacy scheme).
  - The process looks like this: (kvs initialized but can't put or delete) -> (timestamp and lock initialized) -> (kvs initialized and can put or delete) -> (kvs initialized), with a long gap in between steps 3 and 4.
  - Calling a Cassandra `multiPut` or `delete` before/while the timestamp service has been/is being initialized will result in Atlas throwing an exception. I think this is reasonable as we only mark the KVS as initialized long after the raw KVS (with put/delete capability) has been initialized, and setting up the timestamp and lock services does not need multiput or delete (at least for now). Manifestations of this issue would likely only occur if someone reflected into Atlas internals and/or there is a bug in Atlas code.

- `multiPut` and `delete` now add a round trip to TimeLock at TimeLock deployments. Is the perf impact reasonable?
- I still find the two-phase initialization of the KVS gross, though I don't have a better solution offhand.
- Is there a scenario where the KVS can fail to become properly initialized even when TimeLock is happy? I opted to initialize the TimeLock before finishing initialization of the KVS, since (TimeLock + uninitialized KVS) is possibly less awful than (KVS + uninitialized TimeLock) as the later case is prone to data corruption.
- Will this affect KVS healthchecks if timelock is sad? I don't think so as timestamps are only required for `multiPut` and `delete`, and healthchecks seem unlikely to do either of these.

**Where should we start reviewing?**: `AtlasDbFactory`

**Priority (whenever / two weeks / yesterday)**: early next week please!

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3224)


j-baker:
Right now I think there's a round trip per cell written in multiput, so we're likely to wreck our perf in the current state.

We shouldn't really need to change behaviour for puts, because the timestamps we use for puts at the moment are already sufficiently recent to not cause issues here - I wouldn't want to shove yet another RPC in front of transactions.

That has the additional benefit of getting rid of the Cell Value Putter unused -> System.currentTimeMillis().

Additional notes:

It's weird that we'd add an extra method to the KeyValueServiceFactory; we should probably feel comfortable breaking it and just ignoring it in the current places - it's not used elsewhere and it'll just be tech debt (like you should always use the one, and never the other).

j-baker:
and even for deletes, we should also remove the thing where it gets a timestamp every cell

j-baker:
I'd be minded to shove in a Suppliers.memoizeWithExpiration in somewhere btw - just make it as hard as you can for someone to accidentally turn their O(1) rpcs algorithm into O(n).

j-baker:
Yes, exactly that. The point is that if I'm doing deletes A, B, C, the fact that I'm sending them to the KVS at the same time means that there's little benefit from doing deletes (A, 5), (B, 6), (C, 7) as opposed to (A, 5), (B, 5), (C, 5), because A will still screw things up if there's screwing up to be had.

jeremyk-91:
> Right now I think there's a round trip per cell written in multiput, so we're likely to wreck our perf in the current state.

Ah, you're right. Didn't notice this initially - I intended it to be 1/api call, but as written it is 1/cell. As discussed I'll bump delete to be 1/api call (and remove this for multiputs). 

> it's not used elsewhere and it'll just be tech debt (like you should always use the one, and never the other).

This is rather Cassandra-specific (while the KVS factory is not) though - non-Cassandra users could of course use the legacy provider (or even a throwing one), but still seems a bit awkward.

> I'd be minded to shove in a Suppliers.memoizeWithExpiration (10 seconds?) in somewhere btw - just make it as hard as you can for someone to accidentally turn their O(1) rpcs algorithm into O(n).

Yep, makes sense, as we don't need strict guarantees on increasing-ness.

Will implement the changes mentioned.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -293,13 +293,22 @@ static String getFilteredStackTrace(String filter) {
...
 
     static Column createColumn(Cell cell, Value value) {
+        return createColumn(cell, value, value.getTimestamp());
+    }
+
+    static Column createColumn(Cell cell, Value value, long writeTimestamp) {
         byte[] contents = value.getContents();
-        long timestamp = value.getTimestamp();
-        ByteBuffer colName = makeCompositeBuffer(cell.getColumnName(), timestamp);
+        ByteBuffer colName = makeCompositeBuffer(cell.getColumnName(), value.getTimestamp());
```
gsheasby: This isn't a bug, but at first glance it looked like one. Might need more specific variable names and/or a comment to make explicit that `writeTimestamp` means `cassandraTimestamp` and `value.getTimestamp()` is `atlasTimestamp`.

If we only intend to supply a `writeTimestamp` when deleting, we could call the method `createColumnForDelete`, and add a javadoc explaining why we use an unusual (for Atlas) cass timestamp.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -293,13 +293,22 @@ static String getFilteredStackTrace(String filter) {
...
         Column col = new Column();
         col.setName(colName);
         col.setValue(contents);
-        col.setTimestamp(timestamp);
+        col.setTimestamp(writeTimestamp);
+        return col;
+    }
+
+    static Column createColumnAndOverwriteTimestamp(Cell cell, Value value) {
```
gsheasby: This provides identical functionality to `createColumn(Cell, Value)`. Can we `@Deprecate` the original method?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
gsheasby: Or do the code changes to this file just disappear once we roll back the changes related to `multiPut`?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellDeleter.java
```diff
@@ -88,7 +93,7 @@ public Void apply(CassandraClient client) throws Exception {
                             pred.setColumn_names(Arrays.asList(colName));
                             Deletion del = new Deletion();
                             del.setPredicate(pred);
-                            del.setTimestamp(ts + 1);
+                            del.setTimestamp(deleteTimestampGetter.apply(ts));
```
gsheasby: As mentioned by others, move the `deleteTimestampGetter.apply` outside the for loops for efficiency.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellValuePutter.java
```diff
@@ -100,7 +123,12 @@ public Void apply(CassandraClient client) throws Exception {
                             MutationMap map = new MutationMap();
                             for (Map.Entry<Cell, Value> e : partition) {
                                 Cell cell = e.getKey();
-                                Column col = CassandraKeyValueServices.createColumn(cell, e.getValue());
+                                Column col = overwriteTimestamps
+                                        ? CassandraKeyValueServices.createColumn(cell, e.getValue())
+                                        : CassandraKeyValueServices.createColumn(
+                                                cell,
+                                                e.getValue(),
+                                                mutationTimestampGetter.apply(e.getValue().getTimestamp()));
```
gsheasby: expecting this code to go away, but again, yeah this gets called once per cell.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
```diff
@@ -122,6 +117,34 @@ public ServiceDiscoveringAtlasSupplier(
         timestampStoreInvalidator = () -> atlasFactory.createTimestampStoreInvalidator(getKeyValueService());
     }
 
+    public ServiceDiscoveringAtlasSupplier(
+            KeyValueServiceConfig config,
+            java.util.function.Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig,
+            Optional<LeaderConfig> leaderConfig,
+            Optional<String> namespace,
+            Optional<TableReference> timestampTable,
```
gsheasby: This timestampTable arg is always empty

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timestamp/FreshTimestampSupplierAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+public class FreshTimestampSupplierAdapter implements LongSupplier {
+    private volatile TimestampService timestampService;
+
+    public void setTimestampService(TimestampService timestampService) {
+        this.timestampService = timestampService;
+    }
+
+    @Override
+    public long getAsLong() {
+        TimestampService timestampServiceForThisCall = timestampService;
```
gsheasby: Why is this variable necessary?

Is it to avoid the edge case where after you do the below check but before you get fresh timestamp, someone calls `setTimestampService(null)`? If so (maybe even if not so), is `setTimestampService(null)` ever a valid thing to do?

jeremyk-91:
Got through some of this. I still need to
- [ ] go through @gsheasby's line by line review in detail
- [ ] add test coverage

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.cassandra;
+
+public interface CassandraMutationTimestampProvider {
```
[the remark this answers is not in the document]
jeremyk-91: Exactly - there's actually still a requirement for 'what should I set the timestamp on a sweep sentinel' too, as we'd previously set it to 0 but that's not what we want

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellValuePutter.java
jeremyk-91: (Now only called once per API call - interface mentions that this should be cheap as well)

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
```diff
@@ -45,6 +48,14 @@ default KeyValueService createRawKeyValueService(
                 FakeQosClient.INSTANCE);
     }
 
+    KeyValueService createRawKeyValueService(
```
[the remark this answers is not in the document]
jeremyk-91: I've decided not to copy this and force a break on `createRawKeyValueService` instead

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraBackgroundSweeperIntegrationTest.java
```diff
@@ -48,6 +48,7 @@ protected KeyValueService getKeyValueService() {
                 : CassandraContainer.KVS_CONFIG;
         return CassandraKeyValueServiceImpl.create(
                 config,
-                CassandraContainer.LEADER_CONFIG);
+                CassandraContainer.LEADER_CONFIG,
+                CassandraTestTools.getMutationProviderWithStartingTimestamp(1_000_000));
```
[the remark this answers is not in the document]
jeremyk-91: Added. Thanks

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
```diff
@@ -64,6 +102,7 @@ public KeyValueService createRawKeyValueService(
                 preprocessedConfig,
                 cassandraRuntimeConfig,
                 leaderConfig,
+                mutationTimestampProvider,
```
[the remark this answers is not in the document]
jeremyk-91: I've made the order consistent. I think @tpetracca and I added it at different points and I cherry-picked and plugged into his API

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -214,30 +216,43 @@ public void close() {
...
     }
 
     public static CassandraKeyValueService create(
             CassandraKeyValueServiceConfig config,
             Optional<LeaderConfig> leaderConfig,
+            CassandraMutationTimestampProvider mutationTimestampProvider) {
+        return create(config, leaderConfig, mutationTimestampProvider, AtlasDbConstants.DEFAULT_INITIALIZE_ASYNC);
+    }
+
+    public static CassandraKeyValueService create(
```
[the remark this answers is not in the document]
jeremyk-91: I've removed one where I could, though a lot of these are used by multiple tests that want to inject certain specific parameters.

jeremyk-91:
#3241 should be handled separately and is something I want to do, but not here - this has ballooned into a 900-line monster 😞

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -279,13 +279,35 @@ static String getFilteredStackTrace(String filter) {
     }
 
     static Column createColumn(Cell cell, Value value) {
+        return createColumnAtSpecificCassandraTimestamp(cell, value, value.getTimestamp());
```
j-baker: I think this PR would be much, much simpler if we just wrote _everything_ in with the timestamp the supplier is currently giving us

j-baker: sorry, I know this is kinda the opposite of what we suggested before, but given the sweep sentinels codepath and the caching we're doing the cost should be minimal.

Basically always force the timestamp to the one the supplier's given us (assuming that the caching is cheap), and never have it be the one that is passed in by the user.

j-baker:
At the moment this is a little bit broken because it means that deletes can end up being behind the puts that they're supposed to dominate.

Think removing the Suppliers.memoizeWithExpiration is sufficient to fix, although with an extra RPC on deletes. Think this is fine though, since it'll be batched together with all the other ones.

Beyond this this PR seems fine. I'd probably still coalesce all the 3 methods into 1 (basically a fresh timestamp supplier).

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
```diff
@@ -63,6 +68,7 @@ KeyValueService createRawKeyValueService(
             Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig,
             Optional<LeaderConfig> leaderConfig,
             Optional<String> namespace,
+            Optional<Supplier<Long>> freshTimestampSource,
```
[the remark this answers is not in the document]
gsheasby: +1

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -328,6 +338,86 @@ public void testCleanCassLocksStateCli() throws Exception {
...
+                tableReference,
+                ImmutableMap.of(CELL, 1_234_567L),
+                true);
+
+        putDummyValueAtCellAndTimestamp(tableReference, CELL, 1337L, STARTING_ATLAS_TIMESTAMP - 5_000_000);
+        Map<Cell, Value> results = keyValueService.get(tableReference, ImmutableMap.of(CELL, 1337L + 1));
+        assertThat(results.containsKey(CELL), is(false));
+
+        putDummyValueAtCellAndTimestamp(tableReference, CELL, 1_333_337L, STARTING_ATLAS_TIMESTAMP - 5_000_000);
+        Map<Cell, Value> results2 = keyValueService.get(tableReference, ImmutableMap.of(CELL, Long.MAX_VALUE));
```
gsheasby: nits: `results2` is a bit of a lazy name, and I'm not sure what you're testing with this part.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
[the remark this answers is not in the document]
gsheasby: Thanks for not adding to the pain. Cutting down on it is way outside the scope of the PR.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timestamp/FreshTimestampSupplierAdapterTest.java
```diff
@@ -0,0 +1,88 @@
...
+public class FreshTimestampSupplierAdapterTest {
+    private final FreshTimestampSupplierAdapter adapter = new FreshTimestampSupplierAdapter();
+
+    @Test
+    public void throwsNotInitializedIfTimestampServiceNotSet() {
+        assertThatThrownBy(adapter::get).isInstanceOf(NotInitializedException.class);
+    }
+
+    @Test
+    public void throwsNullPointerExceptionIfSettingTimestampServiceToNull() {
```
gsheasby: 💯

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timestamp/FreshTimestampSupplierAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+        Preconditions.checkNotNull(timestampService, "Should not re-set timestamp service in a"
+                + " FreshTimestampSupplierAdapter to null");
+        this.timestampService = timestampService;
+    }
+
+    @Override
+    public Long get() {
+        // There are 2 reads which are not necessarily consistent. However, if the timestamp service changes in between
+        // read one and two, it will not be null and we will get a fresh timestamp from the second service read.
+        // In practice we don't expect this method to be called more than once per TransactionManager.
```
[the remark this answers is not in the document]
jeremyk-91: Fair enough. I think the comment was there cause I initially grabbed a `tsservice` from the beginning which might seem a bit awkward to readers who aren't as familiar with `volatile`s. Now it's pretty obvious

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timestamp/FreshTimestampSupplierAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.palantir.atlasdb.factory.timestamp;
+
+import java.util.function.Supplier;
+
+import com.google.common.base.Preconditions;
+import com.palantir.exception.NotInitializedException;
+import com.palantir.timestamp.TimestampService;
+
+public class FreshTimestampSupplierAdapter implements Supplier<Long> {
```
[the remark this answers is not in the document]
jeremyk-91: Now we aren't planning on using memoizeWithExpiration I'd switch back.
I actually initially used primitive LongSupplier, but switched to this because we were using MWE. I could write an adapter but that would box/unbox the longs anyway, and I didn't really fancy copying the ExpiringMemoizingSupplier from Guava and writing a primitive version

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -34,13 +34,20 @@
 import static org.mockito.Mockito.verify;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import javax.xml.bind.DatatypeConverter;
```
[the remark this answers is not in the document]
jeremyk-91: This is a hack to get the format correct for those Cassandra insert-using-timestamp fixtures. I'll see if there's a way within Atlas to get that though.

jeremyk-91:
(I have some bandwidth to work on this today, in the absence of support explosions)

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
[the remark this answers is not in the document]
jeremyk-91: Agreed, will make it mandatory

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
jeremyk-91: I've renamed the variable, and split the test into 2 parts.

Part 1 is to check that the range tombstones have reasonably current Cassandra timestamps, as this test name suggests.

Part 2 (which includes this check) was intended to be a check against confusing the Cassandra and Atlas timestamps - in particular, if the Cassandra timestamp was used as the Atlas timestamp to apply the range tombstone for, then this test will fail.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolIntegrationTest.java
```diff
@@ -57,6 +58,7 @@
     private CassandraKeyValueService kv = CassandraKeyValueServiceImpl.create(
             CassandraContainer.KVS_CONFIG,
             CassandraContainer.LEADER_CONFIG,
+            CassandraMutationTimestampProviders.legacy(),
```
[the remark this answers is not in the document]
jeremyk-91: Yep, will rename.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -188,7 +188,7 @@ public void lockRequestCanTimeOut() {
     @Test
     public void waitForLocksRequestCanTimeOut() {
         if (isUsingSyncAdapter(cluster)) {
-            // legacy API does not support timeouts on this endpoint
+            // legacyModeForTestsOnly API does not support timeouts on this endpoint
```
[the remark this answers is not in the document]
jeremyk-91: Yep - the refactor tools can be _too_ powerful when renaming methods that are very common words (if you have something like `SomeClass.of()` that would be even worse)

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -324,6 +337,107 @@ public void testCleanCassLocksStateCli() throws Exception {
...
+                TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
+        keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
+
+        keyValueService.addGarbageCollectionSentinelValues(tableReference, ImmutableList.of(CELL));
+
+        putDummyValueAtCellAndTimestamp(
+                tableReference,
+                CELL,
+                Value.INVALID_VALUE_TIMESTAMP,
+                STARTING_ATLAS_TIMESTAMP - 5_000_000);
```
sandorw: -1 would be better to be as precise as possible

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -324,6 +337,107 @@ public void testCleanCassLocksStateCli() throws Exception {
...
+        keyValueService.delete(tableReference, ImmutableListMultimap.of(CELL, 8L));
+
+        // We want to verify that the delete wrote a Cassandra tombstone with timestamp > 5M
+        putDummyValueAtCellAndTimestamp(tableReference, CELL, 8L, STARTING_ATLAS_TIMESTAMP - 5_000_000);
+        Map<Cell, Value> results = keyValueService.get(tableReference, ImmutableMap.of(CELL, 8L + 1));
+        assertThat(results.containsKey(CELL), is(false));
+    }
+
+    @Test
+    public void rangeTombstonesWrittenAtRelativelyCurrentTimestamp() throws Exception {
```
sandorw: also worth renaming these to fresh timestamp to make the intent clearer

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.palantir.atlasdb.cassandra;
+
+import com.google.common.collect.Multimap;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+
+/**
+ * This interface produces Cassandra timestamps for mutations, where it makes sense for these to be overridden.
+ * Generally, to ensure that Cassandra is able to compact tombstones smoothly, these values should increase over
```
sandorw: Probably worth noting the stronger constraints we have around the ordering of actions and the ordering of the C* timestamps that they appear at, since that constraint implies much more than using a relatively recent timestamp.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
```diff
@@ -0,0 +1,49 @@
...
+ * time and should largely be in line with AtlasDB's notion of logical time.
+ *
+ * Calling these methods must be cheap, at least on an amortised basis; they may be called multiple times
```
sandorw: With this PR, can you let me know how often the three methods below will be called, including any higher level batching or extra uses for fetched timestamps?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
jeremyk-91: I've added a more detailed discussion of the constraints (discussed offline).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
jeremyk-91: So beneath this level, I've added it to the docs.

Higher level batching takes place in `RequestBatchingTimestampService`, where a given Atlas client will always maintain only one open timestamp request to timelock.

sandorw: @gmaretic we may need to do follow up work here or at least make sure we understand how much overhead the timestamp call is adding to each ranged tombstone. If this slows targeted sweep down by a factor of 2-3x, we should definitely revisit.

jeremyk-91:
@tpetracca There are integration tests that cover both of these cases, so I'm reasonably confident startup at least behaves properly. 

As far as I understand, the affected code-paths in terms of getting these timestamps are in `KVS.delete()`, `KVS.deleteAllTimestamps()` and `KVS.addGarbageCollectionSentinelValues()`.

These only take place in Background and Targeted Sweep. These sweepers are only instantiated and scheduled long after the adapter is initialized (see `TransactionManager#buildSerializable`).

Timelock Migration takes place before the adapter is initialized - by the time you get the TimestampService that talks to timelock returned to you, timelock migration is already complete.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProviders.java
```diff
@@ -56,8 +57,9 @@ public long getSweepSentinelWriteTimestamp() {
             }
 
             @Override
-            public long getDeletionTimestamp(long atlasDeletionTimestamp) {
-                return longSupplier.getAsLong();
+            public LongUnaryOperator getDeletionTimestampOperatorForBatchDelete() {
+                long deletionTimestamp = longSupplier.getAsLong();
+                return unused -> deletionTimestamp;
```
gsheasby: Is this edit needed? Looks like it was used for debugging?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
```diff
@@ -16,34 +16,47 @@
...
     /**
-     * Cassandra timestamp at which KVS-level deletes should be performed.
-     * @param atlasDeletionTimestamp Atlas timestamp of the relevant delete.
+     * Returns an operator, which determines the Cassandra timestamp to write a delete at, given the Atlas timestamp
+     * at which the deletion occurred.
+     * The operator is intended to be use for a single batch deletion, and must not be re-used outside of the
+     * context of a single batch delete.
+     *
+     * The operator is obtained once per API call to CassandraKeyValueServiceImpl.delete().
+     * The operator is invoked once per cell involved in such an API call, and thus must be cheap to invoke.
```
gsheasby: Is this still correct?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProviders.java
jeremyk-91: It's needed actually - we don't want the timestamp operator to make a remote call, because it is called on every cell. Preparing the timestamp operator for a single batch delete can make a remote call (which is OK)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraMutationTimestampProvider.java
jeremyk-91: Yep - see `CellDeleter:96`. The problem here is that in general you can't call this outside the loop, because in legacy/test mode you want to return `ts + 1` where ts is the delete timestamp

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -324,6 +337,107 @@ public void testCleanCassLocksStateCli() throws Exception {
...
+        keyValueService.addGarbageCollectionSentinelValues(tableReference, ImmutableList.of(CELL));
+
+        putDummyValueAtCellAndTimestamp(
+                tableReference,
+                CELL,
+                Value.INVALID_VALUE_TIMESTAMP,
+                STARTING_ATLAS_TIMESTAMP - 1);
+
+        // The following implies that the sentinel was written with a timestamp greater than 5M, which is reasonable
+        // as the current time in Atlas is 10M.
```
jeremyk-91: remove this comment - not true any more

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -324,6 +337,107 @@ public void testCleanCassLocksStateCli() throws Exception {
...
+                TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
+        keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
+        byte[] data = PtBytes.toBytes("data");
+        byte[] moreData = PtBytes.toBytes("data2");
+
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL, Value.create(data, 8L)));
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL, Value.create(moreData, 88L)));
+        keyValueService.delete(tableReference, ImmutableListMultimap.of(CELL, 8L));
+
+        // We want to verify that the delete wrote a Cassandra tombstone with timestamp > 5M
```
jeremyk-91: as above

jeremyk-91:
awesome - thanks @j-baker for doing the comment cleanups for me!

Internal smoke-tests for the embedded as well as migration cases passed (with internal blob storage product) - I started the service, wrote a few pieces of data to Atlas and checked they could still be read. Also kicked off sweep manually and checked that the writetimes for sentinels were fresh timestamps.

PR #3227, opened 2018-06-01
There is now only one way to put data into the KVS

j-baker:
**Goals (and why)**: At the moment there are three ways of putting data into AtlasDB. One takes a table, a map of cell -> byte[], and a timestamp. One takes a `map<table<cell, byte[]>>`, and a timestamp. One takes a table, a map of cell to byte[] and timestamp (allowing you to specify multiple timestamps.

Hilariously, none of these three are the one I want, which is one that lets me hand a bunch of (table, cell, timestamp, value) tuples and get them written (I wanna do some [smart batching](https://mechanical-sympathy.blogspot.com/2011/10/smart-batching.html) to merge writes from different transactions into the same KVS call).

Even better, the behaviours are slightly different (put and putWithTimestamps typically are slower to run that multiPut).

This PR deletes all of the implementations of the put methods and deletes the interface methods. It also removes all internal usages of the put methods.

Why am I deleting the interface methods, obviously providing a dev break? because by adding a new interface method, the same people (large internal product) will get broken anyway (they implement, they don't use the methods themselves). So i'll have to make that PR whatever happens.

What are the consequences?

- The most naive consequence is an extra object allocation for every value stored (the 'Write' object). That said, in most cases this is offset by removing an object allocation for those same puts elsewhere (the 'Value' allocation).
- The second consequence is some shenanigans around the TableSplittingKeyValueService - at the moment it buffers all the writes one extra time. In practice this isn't as bad as one might imagine (after all, creating a list of a few thousand elements is cheap), and in all other cases we avoid the extra buffer (usually the writes need to be transformed anyway - in DbKvs they are sorted, in Cassandra they are partitioned etc).

I'm going to run some performance tests, and will get back with results. Obviously as usual if the perf degradation is huge, this likely won't happen :).

What decisions did I specifically make?

Well, the decision to take the stream is one. I could have instead taken a `Map<TableReference, Map<Cell, Value>>` which would have expressed the same semantics - the downside is that it's dramatically harder to autobatch (which is the point) and a huge pain to construct if you just want to store something (it's why put in its current form existed until now, for example).

I also elected to remove some of the logic from the tableremappingkeyvalueservice.

The logic is basically unused given it's largely just on the putWithTimestamps one, and I don't think keeping it has much value. The short circuit methods are just delaying the inevitable shortcircuit later.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3227)


tboam:
@j-baker  Sorry for the length of time taken to review this.  If you still want this can you fix the conflicts and then someone can take a look.

gmaretic:
Do we still want to do this?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Write.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
```
felixdesouza: - [ ] License

PR #3228, opened 2018-06-02
[no release notes] Delete the CQL KVS and some other unused stuff

j-baker:
This is likely to be contentious, but the gist is:

1. The CQL stuff has never been used in production, so likely doesn't
work well enough.
2. There's no migration path; the CQL stuff likely can _never_ be used
in production. In particular, we're unlikely to move over in this way
(big-bang switch) - the only way to migrate is the way we're currently
doing it (migrating one call at a time, tactically).
3. There's almost no testing, as you can see there are no test changes
here. In AtlasDB, what this generally means is that there are loads of
bugs.
4. This code actually ends up getting modified a lot (100 times in the
last 3 years) so is clearly a burden on the AtlasDB team.

Given that, would like to delete, and we can always resurrect later?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3228)


tpetracca:
> the only way to migrate is the way we're currently doing it (migrating one call at a time, tactically).

This is the biggest reason why I think this is fine. Thankfully version control is great and we can always resurrect this stuff if we'd like. I have feeling we'd likely rethink a bunch of the implementations here anyway if we were to swap over. The datastax driver and CQL generally have come a long way since this was written.

gsheasby:
@clockfort for SA as this was his pet project at some point

sandorw:
also @fsamuel-bs

tpetracca:
Yea, there are even some reasons why we can't switch to it in it's current form anyway (like the cass timestamp overwriting stuff; all the CQL stuff uses wallclock)

fsamuel-bs:
The biggest value that I saw for leaving it around was to have a reference code for moving to CQL down the road. We can always use `git history` for checking out how this implementation. 👍 for removing it.

PR #3229, opened 2018-06-02
[no release notes] delete lockfiles

j-baker:
The lock files make trivial PRs have massive diffs, and provide
relatively little value. Wanted to make this PR to suggest
their deletion, though this is largely a function of me not knowing that
no one gets value rather than asserting that no one does.

Anyone prefer that these stay?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3229)


tboam:
Checking with the team on whether we should proceed with this.

PR #3245, opened 2018-06-06
PDS-69180 - CommitTsLoader is too aggressive when warming its cache

### LINKED ISSUE #2979: Sweep should batch loads from _transactions
Saw this message when sweeping a wide row on a production stack:
```
Re-batching in getLoadWithTsTasksForSingleHost a call to {} for table {} that attempted to multiget {} rows; this may indicate overly-large batching on a higher level. {}
tableRef: "_transactions"
rows: "117147.00"
```
And again with rows = 116123, 7300, and 109687

Sweep attempted to fetch over 100,000 rows from the `_transactions` table. This gets rebatched to a maximum of 5000 rows; we should do some batching ourselves, perhaps `min(16*maxCellTsPairsToExamine, 5000)`.

jeremyk-91:
**Goals (and why)**: 
- fix suspected root cause of internal ticket PDS-69180
- fix #2979 - see more details there

- If Sweep is checking the status of a ton of candidates, split its reads into smaller batches of 5000, and warm the cache one batch at a time instead of all in parallel.

- should 5000 be configurable? Might be worth doing this, though I refrained owing to high priority and also because this is not a regression (existing is already a flat 5000)
- this obviously slows this phase of sweep down, but I think that's what we want to do!

**Where should we start reviewing?**: `CommitTsLoader.java`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3245)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
@@ -38,6 +38,7 @@ import com.google.common.collect.Multimaps;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.TreeMultimap;
 import com.google.common.primitives.UnsignedBytes;
+import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cassandra.CassandraKeyValueServiceConfig;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
@@ -135,7 +136,7 @@ class CellLoader {
             cellsByCol.put(cell.getColumnName(), cell);
         }
         List<Callable<Void>> tasks = Lists.newArrayList();
-        int fetchBatchCount = config.fetchBatchCount();
+        int fetchBatchCount = AtlasDbConstants.TRANSACTION_TIMESTAMP_LOAD_BATCH_LIMIT;
         for (Map.Entry<byte[], Collection<Cell>> entry : Multimaps.asMap(cellsByCol).entrySet()) {
             final byte[] col = entry.getKey();
             Collection<Cell> columnCells = entry.getValue();
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
@@ -140,8 +140,11 @@ class CellLoader {
             final byte[] col = entry.getKey();
             Collection<Cell> columnCells = entry.getValue();
             if (columnCells.size() > fetchBatchCount) {
-                log.warn("Re-batching in getLoadWithTsTasksForSingleHost a call to {} for table {} that attempted to "
-                                + "multiget {} rows; this may indicate overly-large batching on a higher level.\n{}",
+                log.warn("Re-batching in getLoadWithTsTasksForSingleHost a call to {} for table {} that attempted to"
+                                + " multiget {} rows; this may indicate overly-large batching on a higher level."
+                                + " Note that batches are executed in parallel, which may cause load on both"
+                                + " your Atlas client as well as on Cassandra if the number of rows is exceptionally"
+                                + " high.\n{}",
                         SafeArg.of("host", CassandraLogHelper.host(host)),
                         LoggingArgs.tableRef(tableRef),
                         SafeArg.of("rows", columnCells.size()),
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
 import com.palantir.atlasdb.keyvalue.api.KeyAlreadyExistsException;
 import com.palantir.atlasdb.transaction.api.TransactionFailedRetriableException;
@@ -39,7 +40,8 @@ import gnu.trove.set.TLongSet;
 public final class CommitTsLoader {
     private static final Logger log = LoggerFactory.getLogger(CommitTsLoader.class);
 
-    private static final int TS_PER_BATCH = 5000;
+    @VisibleForTesting
+    static final int TS_PER_BATCH = 5000;
 
     private final TLongLongMap commitTsByStartTs;
     private final TransactionService transactionService;
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
@@ -62,7 +62,7 @@ public final class CommitTsLoader {
             TransactionService transactionService,
             TLongSet startTssToWarmingCache,
             TLongLongMap cache) {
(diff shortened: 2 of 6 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
```diff
@@ -49,12 +52,22 @@ private CommitTsLoader(TLongLongMap commitTsByStartTs, TransactionService transa
     public static CommitTsLoader create(TransactionService transactionService, TLongSet startTssToWarmingCache) {
         TLongLongMap cache = new TLongLongHashMap();
         if (!startTssToWarmingCache.isEmpty()) {
-            // Ideally TransactionService should work with primitive collections to avoid GC overhead..
-            cache.putAll(transactionService.get(TDecorators.wrap(startTssToWarmingCache)));
+            warmCache(transactionService, startTssToWarmingCache, cache);
```
sandorw: Curious why we don't consider warming the cache async, especially if we may be loading a million pairs during the warming phase. Especially since the default cache size is one million, so loading any more than that is going to end up being useless.

sandorw: (Not necessarily in the scope of this PR - we should resolve the P0 as a higher priority)

sandorw: As @j-baker noted, sweep is also be a bad user of the cache, since it pollutes it with timestamps that are likely to be uncorrelated with the timestamps read by live requests.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CommitTsLoaderTest.java
```diff
@@ -88,4 +95,26 @@ public void loadShouldThrowIfANullIsToBeReturned() throws Exception {
...
+            return timestamps.stream().collect(Collectors.toMap(n -> n, n -> n));
+        }).when(mockTransactionService).get(any());
+
+        TLongSet initialTimestamps = new TLongHashSet();
+        initialTimestamps.addAll(LongStream.range(0, valuesToInsert).boxed().collect(Collectors.toList()));
+
+        CommitTsLoader loader = CommitTsLoader.create(mockTransactionService, initialTimestamps);
+        assertThat(loader.load(valuesToInsert - 1)).isEqualTo(valuesToInsert - 1);
+
+        verify(mockTransactionService, atLeast(2)).get(any());
```
sandorw: Is this trying to avoid assuming the 5000 batch size? It's already checked explicitly above. I'd prefer to just make the batch size protected and refer to it directly.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
sandorw: Checked the above and we _only_ use the cache for sweep, not for actual transactions, where I'd guess it may be more useful...

sandorw: Fun. 

Transactions use a `TimestampCache` which is by default limited to 1 million entries. It's a dumb cache class with no logic for loading timestamps or rolling back transactions.

Sweep uses a different class `CommitTsLoader` instead, which doesn't really act as a cache. Each `CommitTsLoader` only lives as long as it takes to complete a call to `SweepableCellFilter.getCellsToSweep`, so we only benefit from the cache if, if I understand correctly, multiple cells in a single row were written with the same timestamp, which certainly should happen. That said, we are throwing away the cache completely as soon as we move on to the next batch. Background sweep should use a different cache from the transactions, since the timestamps they load are very likely to be different, but we might benefit from combining the classes used and not blowing away the cache so frequently.

Also, `CommitTsLoader` has no eviction and targeted sweep keeps a long lived one around... filed https://github.com/palantir/atlasdb/issues/3246.

Sorry for all the confusion above, it was pretty surprising to me that we have two separate commit timestamp cache implementations :(

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
```diff
@@ -140,8 +140,11 @@ void loadWithTs(String kvsMethodName,
             final byte[] col = entry.getKey();
             Collection<Cell> columnCells = entry.getValue();
             if (columnCells.size() > fetchBatchCount) {
```
gmaretic: Should this even be configured by this config? I'd err on the side of making it constant. Also make sure it is consistent with the `TS_PER_BATCH` below

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
```diff
@@ -38,6 +39,8 @@
 public final class CommitTsLoader {
     private static final Logger log = LoggerFactory.getLogger(CommitTsLoader.class);
 
+    private static final int TS_PER_BATCH = 5000;
```
gmaretic: See above

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
jeremyk-91: Correct, yeah these are two distinct caches 😞

Yep, rather confusingly we also throw away the `CommitTsLoader`. (Technically the cache gives you benefits if your rows are small and rows that are nearby share timestamps too)

I wrote this PR as a fairly quick hack to defuse the P0, but I'd want to unify the caches at some point

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CommitTsLoaderTest.java
jeremyk-91: :+1: will change this

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
```diff
@@ -38,6 +39,8 @@
 public final class CommitTsLoader {
```
gmaretic: This class should really just use a LoadingCache internally so that we have eviction, I'm happy to do that in a separate PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
```diff
@@ -49,12 +52,22 @@ private CommitTsLoader(TLongLongMap commitTsByStartTs, TransactionService transa
...
         if (!startTssToWarmingCache.isEmpty()) {
-            // Ideally TransactionService should work with primitive collections to avoid GC overhead..
-            cache.putAll(transactionService.get(TDecorators.wrap(startTssToWarmingCache)));
+            warmCache(transactionService, startTssToWarmingCache, cache);
+
         }
         return new CommitTsLoader(cache, transactionService);
     }
 
+    private static void warmCache(
```
gmaretic: I'd like to get rid of this and try to reuse the cache instead. But outside the scope of this PR.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/CommitTsLoader.java
jeremyk-91: Yes. Will be coming in part 2

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
jeremyk-91: We don't have visibility of `CassandraConstants` here. I can shove this into `AtlasDbConstants` maybe - will have a look at doing that

PR #3253, opened 2018-06-07
[TS] Targeted Sweep: do not sweep cells with commitTs >= sweepTs

gmaretic:
**Goals (and why)**:
Fix the following corner case:
Txn 1 starts at 100, deletes a cell, commits at 200.
Txn 2 starts at 150, writes to the cell.

Run thorough sweep at sweepTs 150
Sweep puts a ranged tombstone including the delete of Txn 1.
Txn 2 successfully commits at 300

Stop getting cells to sweep when we hit a startTs that was committed after sweepTs.

Are the tests enough? Probably can do with a refactor, this was done rather hastily

small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3253)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -156,54 +157,65 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
...
-                committedTimestamps.add(entry.getKey());
+                processedAll = false;
+                break;
             }
         }
+
         cellsToDelete.forEach((tableRef, multimap) -> {
             kvs.delete(tableRef, multimap);
             maybeMetrics.ifPresent(metrics -> metrics.updateAbortedWritesDeleted(shardStrategy, multimap.size()));
             log.info("Deleted {} aborted writes from the KVS.", SafeArg.of("number", multimap.size()));
```
sandorw: Either log the tableRef here as well, or move the log line outside the lambda and just log the total number of writes aborted.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -156,54 +157,65 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
...
+            long commitTs = startToCommitTs.get(startTs);
+            if (commitTs == TransactionConstants.FAILED_COMMIT_TS) {
+                lastSweptTs = startTs;
+                writesByStartTs.get(startTs)
                         .forEach(write -> cellsToDelete
                                 .computeIfAbsent(write.tableRef(), ignore -> HashMultimap.create())
                                 .put(write.cell(), write.timestamp()));
+            } else if (commitTs < sweepTs) {
+                lastSweptTs = startTs;
+                committedTimestamps.add(startTs);
```
sandorw: Likely doesn't matter, but you could use a reverse ordered list or stack to avoid the extra sort at the end.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -156,54 +157,65 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
...
             } else {
-                committedTimestamps.add(entry.getKey());
+                processedAll = false;
+                break;
             }
         }
+
         cellsToDelete.forEach((tableRef, multimap) -> {
```
sandorw: For future note, I'm curious how often we need to delete an aborted txn's writes directly vs. the batch containing a valid write already that would cover it with a ranged tombstone. If batch sizes are sufficiently high, we might be able to avoid most of the kvs.delete calls here if each sweep iteration needs to be sped up.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -156,54 +157,65 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
...
+                .map(writesByStartTs::get)
+                .flatMap(Collection::stream)
+                .forEach(write -> writesToSweepFor
+                        .putIfAbsent(write.writeRef().cellReference(), write));
+        return writesToSweepFor.values();
+    }
+
+    private long getLastSweptTs(TimestampsToSweep startTsCommitted, RowColumnRangeIterator resultIterator,
+            long partitionFine, long maxTsExclusive) {
+        if (startTsCommitted.processedAll() && exhaustedAllColumns(resultIterator)) {
```
sandorw: Remind me when we wouldn't have exhausted all columns?

gmaretic: We get the whole row, but we only scan the columns until we go over the threshold of ``SweepQueueUtils.SWEEP_BATCH_SIZE`` cells.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -591,6 +593,67 @@ public void testSweepTimestampMetric() {
...
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(920L);
+        ArgumentCaptor<Map> argument = ArgumentCaptor.forClass(Map.class);
+        verify(spiedKvs, times(1)).deleteAllTimestamps(eq(TABLE_CONS), argument.capture(), eq(false));
+        assertThat(argument.getValue()).containsValue(920L);
+
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(920L);
+        verify(spiedKvs, times(1)).deleteAllTimestamps(any(TableReference.class), anyMap(), eq(false));
+
+        immutableTs = 1101L;
```
sandorw: Would 1001L make more sense to catch off by one errors and more precisely check the logic?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -591,6 +593,67 @@ public void testSweepTimestampMetric() {
...
     }
 
+    @Test
+    public void doNotSweepAnythingAfterEntryWithCommitTsAfterSweepTs() {
+        immutableTs = 1000L;
+        enqueueWrite(TABLE_CONS, 900);
+        enqueueWrite(TABLE_CONS, 910);
+        enqueueWriteCommitedAt(TABLE_CONS, 950, 1000);
+        enqueueWrite(TABLE_CONS, 970);
+        enqueueWrite(TABLE_CONS, 920);
```
sandorw: nit: The order here isn't meaningful, right? Might as well order them based on `startTs` to make it clearer.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -591,6 +593,67 @@ public void testSweepTimestampMetric() {
...
+        assertThat(argument.getValue()).containsValue(920L);
+
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(920L);
+        verify(spiedKvs, times(1)).deleteAllTimestamps(any(TableReference.class), anyMap(), eq(false));
+
+        immutableTs = 1101L;
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(1101L - 1L);
+        verify(spiedKvs, times(2)).deleteAllTimestamps(eq(TABLE_CONS), argument.capture(), eq(false));
```
sandorw: This is one additional call to `deleteAllTimestamps`, not two more, correct?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -591,6 +593,67 @@ public void testSweepTimestampMetric() {
...
+
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(920L);
+        verify(spiedKvs, never()).deleteAllTimestamps(any(TableReference.class), anyMap(), eq(false));
+
+        ArgumentCaptor<Multimap> multimap = ArgumentCaptor.forClass(Multimap.class);
+        verify(spiedKvs, times(1)).delete(eq(TABLE_CONS), multimap.capture());
+        assertThat(multimap.getValue().keySet()).containsExactly(DEFAULT_CELL);
+        assertThat(multimap.getValue().values()).containsExactly(900L, 920L);
+        verify(spiedKvs, times(1)).delete(eq(TABLE_CONS), multimap.capture());
```
sandorw: duplicate of a few lines above

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -591,6 +593,67 @@ public void testSweepTimestampMetric() {
...
+        enqueueWriteCommitedAt(TABLE_CONS, 950, 2000);
+        sweepQueue.enqueue(writeToDefaultCell(TABLE_CONS, 970), 970);
+        sweepQueue.enqueue(writeToDefaultCell(TABLE_CONS, 1110), 1110);
+
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
+        assertThat(progress.getLastSweptTimestamp(ShardAndStrategy.conservative(CONS_SHARD))).isEqualTo(920L);
+        verify(spiedKvs, never()).deleteAllTimestamps(any(TableReference.class), anyMap(), eq(false));
+
+        ArgumentCaptor<Multimap> multimap = ArgumentCaptor.forClass(Multimap.class);
+        verify(spiedKvs, times(1)).delete(eq(TABLE_CONS), multimap.capture());
```
sandorw: Why were 900 and 920 deleted directly rather than via `deleteAllTimestamps`? From the code on this PR, the direct delete should only occur if the transactions weren't committed, and it looks like they should be.

nit: The two new tests use  `enqueueWrite`, `sweepQueue.enqueue`, and `enqueueWriteCommittedAt`, which are hard to reason about given that I assume they must behave differently, but seem very similar.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gmaretic: I replaced the sort with Lists.reverse()

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
gmaretic: Good point

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
gmaretic: Sure

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
gmaretic: Fair enough

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
gmaretic: Yes, this counts the total number of calls, will add a comment

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
gmaretic: Definitely wasn't clear enough. Should be good now

sandorw: Looks good now, thanks.

PR #3257, opened 2018-06-08
Feature/multi sweep

### LINKED ISSUE #3225: MultiSweep: allow multiple simultaneous sweep processes
At some $large_internal_product deployments, boxes are beefy and one sweep process isn't enough.

Multiple sweep threads doing different tables would allow them to keep up better.

This would need:
* Launching multiple threads, e.g. in BackgroundSweeperImpl
* Changes to `sweep_progress` to allow each thread to process a different table
* Configuration for number of threads
* Handling changes to number of threads (spinning up a new thread or shutting down other threads)

Note that the backup lock will still slow things down as only one thread would be able to delete at once (unless the backup lock is disabled, which is the case for the internal product)

gsheasby:
**Goals (and why)**: Fixes #3225.
Allow sweep to be run in multiple threads. This will be most useful for $large_internal_product, where the database is run on beefy boxes that can handle multiple concurrent sweeps

- [x] Split BackgroundSweepThread out of BackgroundSweepImpl, allowing the Impl to purely do thread management
- [x] Apply config to start multiple threads
- [x] Make each thread use separate locks and separate columns in the SweepProgress table
- [x] Grab locks per-table to ensure that different threads sweep different tables
- [x] Tests
- [x] Docs
- [x] Release notes 
- [x] (stretch) Allow a sweep thread to resume the previously running sweep (stored in column "s")

Not making numThreads live-reloadable in this PR.

- The tricky part was creating the locks and passing them around. Would it be better to have each thread grab locks when choosing a table, and then grab a new lock when it moves on to a new table? The caveat here is that multi-node services will start running multiple sweeps, which I don't think is desirable in the general case.

**Where should we start reviewing?**: BackgroundSweeperImpl, but a lot of the interesting stuff is also in NextTableToSweepCalculator

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3257)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepOutcomeMetrics.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
+
+import com.codahale.metrics.SlidingTimeWindowReservoir;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.atlasdb.util.MetricsManager;
+
+class SweepOutcomeMetrics {
+    private final MetricsManager metricsManager = new MetricsManager();
```
sandorw: This should be updated to reflect https://github.com/palantir/atlasdb/pull/3237

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/SweeperTestSetup.java
```diff
@@ -42,16 +43,19 @@
...
 import com.palantir.lock.LockService;
+import com.palantir.lock.SingleLockService;
 
 public class SweeperTestSetup {
-
-    protected static final TableReference TABLE_REF = TableReference.createFromFullyQualifiedName(
+    protected static final TableReference OTHER_TABLE = TableReference.createFromFullyQualifiedName(
             "backgroundsweeper.fasttest");
+    protected static final TableReference TABLE_REF = TableReference.createFromFullyQualifiedName(
+            "backgroundsweeper.fasttest_other");
```
sandorw: 😆 OTHER_TABLE = "fasttest", TABLE = "fasttest_other"

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/BackgroundSweeperFastTest.java
```diff
@@ -161,6 +161,46 @@ public void testWriteProgressAfterIncompleteRunWithPreviousProgress() {
...
+        setNoProgress(otherThreadIndex);
+        setupTaskRunner(OTHER_TABLE, ImmutableSweepResults.builder()
+                .staleValuesDeleted(2)
+                .cellTsPairsExamined(10)
+                .minSweptTimestamp(12345L)
+                .nextStartRow(Optional.of(new byte[] {4, 5, 6}))
+                .timeInMillis(20L)
+                .timeSweepStarted(50L)
+                .build());
+        setNextTableToSweep(OTHER_TABLE);
```
sandorw: A little unclear what code this test is covering, since the cross thread lock contention seems to be mocked and skipped over here.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProviderTest.java
```diff
@@ -89,6 +102,32 @@ public void calculatorReturnsSingleHighestPriority_thenProviderReturnsThatTable(
...
+        givenPriority(table("table2"), 20.0);
+        givenTableIsLocked("table1");
+
+        whenGettingNextTableToSweep();
+
+        thenTableChosenIs(table("table2"));
+    }
+
+    @Test
+    public void calculatorReturnsMultiplePriorities_allLocked_thenProviderReturnsNothing()
```
sandorw: rename? 2/3 are locked and the last one has sweep priority (which apparently means not to sweep it)

docs/source/cluster_management/sweep/standard-sweep.rst
```diff
@@ -36,6 +36,7 @@ Note that some of these parameters are just used as a hint. Sweep dynamically mo
    :widths: 20, 20, 40, 200
 
    ``enabled``, "Only specified in config", "true", "Whether the background sweeper should run."
+   ``sweepThreads``, "Only specified in config", "1", "The number of threads to run sweep with. Changes require a restart of any service node to take effect. Not recommended for Cassandra KVS."
```
sandorw: Might be worth specifying here the contention for the backup lock. Outside of that, is there a reason to avoid this?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+        return overrideConfig.priorityTables().contains(currentTableName);
+    }
+
+    private Optional<TableToSweep> getNextTableToSweep(Transaction tx, SweepPriorityOverrideConfig overrideConfig) {
+        Optional<TableToSweep> nextTableToSweep = nextTableToSweepProvider.getNextTableToSweep(
+                tx,
+                specificTableSweeper.getSweepRunner().getConservativeSweepTimestamp(),
+                overrideConfig);
+        nextTableToSweep.ifPresent(tableToSweep -> lockForCurrentTable = tableToSweep.getSweepLock());
```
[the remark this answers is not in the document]
gsheasby: After the latest change, I think we actually should just keep the `TableToSweep`, rather than tracking a few separate things from it. We'd need to remember to update it each time with the latest progress, though (probably with a method on `TableToSweep`).

Will refactor after the per-table progress stuff has been checked.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/BackgroundSweeperFastTest.java
gsheasby: My intention was to test that if one thread is sweeping `TABLE`, then the other thread will sweep some `OTHER_TABLE`.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProviderTest.java
gsheasby: yep, we don't sweep a table if the sweep priority is zero.

docs/source/cluster_management/sweep/standard-sweep.rst
gsheasby: Sweep is still quite a heavy operation for Cassandra, and extra sweep threads sound like a recipe for trouble, and we don't need to take that risk given the proximity of targeted sweep. I'll add something about the backup lock, though.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweeperImpl.java
```diff
@@ -325,73 +140,30 @@ public void close() {
...
-        log.info("Signalling background sweeper to shut down.");
+        log.info("Signalling background sweepers to shut down.");
         // Interrupt the daemon, whatever lock it may be waiting on.
-        daemon.interrupt();
+        daemons.forEach(Thread::interrupt);
         // Ensure we do not accidentally abort shutdown if any code incorrectly swallows InterruptedExceptions
         // on the daemon thread.
         shuttingDown.countDown();
+
+        // TODO what happens if one of the daemons (but not the last one) causes this to throw?
```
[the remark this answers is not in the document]
gsheasby: Sounds sub-optimal. Will rewrite it to collect the exceptions in a list and rethrow at the end.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -0,0 +1,326 @@
...
+    private void createLockForTableIfNecessary(SweepProgress progress) {
+        if (currentTable.isPresent() && !currentTable.get().getTableRef().equals(progress.tableRef())) {
+            // If currentTableToSweep is present, we're stopping the current table due to a new override. Release the lock
+            currentTable.ifPresent(tableToSweep -> tableToSweep.getSweepLock().close());
+        }
+
+        TableReference tableRef = progress.tableRef();
+        SingleLockService singleLockService = SingleLockService.createNamedLockServiceForTable(
+                lockService, TABLE_LOCK_PREFIX, tableRef);
+        currentTable = Optional.of(TableToSweep.continueSweeping(tableRef, singleLockService, progress));
```
gsheasby: There was a bug in the "keeping track of current table logic" (see the last commit), partly because the logic is split over a few classes.
* `TableToSweep.continueSweeping` sets the `hasPreviousProgress` flag to true. Even though we're sweeping the same table, we need to update the progress, which at present means creating a whole new TableToSweep object (and creating a new `SingleLockService` too). If you don't do this, then the issue shows up in [a completely different part of the code](https://github.com/palantir/atlasdb/pull/3257/files#diff-94bc098d7947536a70c47437788b4849R219) (`SpecificTableSweeper.processFinishedSweep`), causing the sweep.priority table to be continually updated. Either this or another side-effect caused a sweep of the sweep.priority table to loop infinitely.

I'm considering a refactor to pass some of these responsibility into `TableToSweep`, or perhaps a new class `SweepState` which manages the sweep state machine and holds the `Optional<TableToSweep>` object. This class would be responsible for opening, refreshing and closing the per-table locks, as well as keeping the sweep progress state updated. However, this PR is large already, so perhaps this is best deferred to a later date.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -203,40 +203,46 @@ SweepOutcome runOnce() {
...
+                                    tableToSweep.getTableRef()));
+                    SweepPriorityOverrideConfig overrideConfig = sweepPriorityOverrideConfig.get();
+                    if (progress.map(
+                            realProgress -> shouldContinueSweepingCurrentTable(realProgress, overrideConfig))
+                            .orElse(false)) {
+                        try {
+                            createLockForTableIfNecessary(progress.get());
+                            if (currentTable.isPresent()) {
+                                //noinspection ConstantConditions // class runs in a single thread, so this is fine
+                                currentTable.get().refreshLock();
```
sandorw: Expected behavior here? If we fail to refresh, the lock just gets set to null.

docs/source/release_notes/release-notes.rst
```diff
@@ -118,6 +118,30 @@ develop
          - The ``BackgroundSweeper`` is no longer a ``Runnable``. Its job is now to manage ``BackgroundSweeperThread`` instances, which are ``Runnable``.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3257>`__)
 
+    *    - |fixed|
```
sandorw: Missing from other PRs?

sandorw:
I'm still unsure whether we're handling the table locks properly here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -203,40 +203,46 @@ SweepOutcome runOnce() {
...
+                    return getNextTableToSweep(tx, overrideConfig);
+                });
     }
 
     // This is needed if we've just grabbed the sweep thread lock and a previous thread was in the middle of a table
-    private void createLockForTableIfNecessary(TableReference tableRef) {
-        if (lockForCurrentTable != null) {
-            return;
+    private void createLockForTableIfNecessary(SweepProgress progress) {
+        if (currentTable.isPresent() && !currentTable.get().getTableRef().equals(progress.tableRef())) {
```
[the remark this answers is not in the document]
gsheasby: This occurs if the sweep priority overrides changed since the last iteration.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -203,40 +203,46 @@ SweepOutcome runOnce() {
...
+                    Optional<SweepProgress> progress = currentTable.flatMap(
+                            tableToSweep -> specificTableSweeper.getSweepProgressStore().loadProgress(
+                                    tableToSweep.getTableRef()));
+                    SweepPriorityOverrideConfig overrideConfig = sweepPriorityOverrideConfig.get();
+                    if (progress.map(
+                            realProgress -> shouldContinueSweepingCurrentTable(realProgress, overrideConfig))
+                            .orElse(false)) {
+                        try {
+                            createLockForTableIfNecessary(progress.get());
+                            if (currentTable.isPresent()) {
```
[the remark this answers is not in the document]
gsheasby: The comment suppresses an intellij warning, which I guess occurs because potentially another thread can make currentTable empty in between isPresent and get, although in practice this won't happen here, because the class runs in a single thread.

Re the nit: I wish it could be, but that gives me a compile error (unhandled InterruptException) - yes, even though we catch it later.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -203,40 +203,46 @@ SweepOutcome runOnce() {
...
-            return;
+    private void createLockForTableIfNecessary(SweepProgress progress) {
+        if (currentTable.isPresent() && !currentTable.get().getTableRef().equals(progress.tableRef())) {
+            // If currentTableToSweep is present, we're stopping the current table due to a new override. Release the lock
+            currentTable.ifPresent(tableToSweep -> tableToSweep.getSweepLock().close());
         }
 
-        lockForCurrentTable = SingleLockService.createNamedLockServiceForTable(
+        TableReference tableRef = progress.tableRef();
+        SingleLockService singleLockService = SingleLockService.createNamedLockServiceForTable(
```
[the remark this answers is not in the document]
gsheasby: Yeah, we should keep the same lock. The logic here is a bit funky, which is why I suggested a state machine above.

gsheasby:
✅ Verified on a test stack that the different threads choose different tables, and operate their sweeps independently.

atlasdb-config/src/main/java/com/palantir/atlasdb/config/SweepConfig.java
```diff
@@ -90,6 +100,12 @@ public SweepPriorityOverrideConfig sweepPriorityOverrides() {
         return SweepPriorityOverrideConfig.defaultConfig();
     }
 
+    @Value.Check
+    public void check() {
+        Preconditions.checkState(sweepThreads() > 0, "Must have a positive number of threads! "
+                        + "If your intention was to disable sweep, please set enabled to false.");
```
jeremyk-91: Nice!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
```diff
@@ -81,15 +84,43 @@ public static NextTableToSweepProvider create(KeyValueService kvs, SweepPriority
...
+    private Optional<TableToSweep> attemptToChooseTableFromList(List<TableReference> priorityTables, String reason) {
+        for (TableReference tableRefToSweep : priorityTables) {
+            SingleLockService sweepLockForTable = SingleLockService.createNamedLockServiceForTable(
+                    lockService, BackgroundSweepThread.TABLE_LOCK_PREFIX, tableRefToSweep);
+            try {
+                sweepLockForTable.lockOrRefresh();
+                if (sweepLockForTable.haveLocks()) {
+                    log.info("Decided to start sweeping {} because {}.",
+                            LoggingArgs.safeTableOrPlaceholder(tableRefToSweep), reason);
+                    return Optional.of(TableToSweep.newTable(tableRefToSweep, sweepLockForTable));
```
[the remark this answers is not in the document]
jeremyk-91: Yep, we do refresh the lock. Right now it doesn't look like the thread flags something if it loses the lock though (unless I'm missing something!)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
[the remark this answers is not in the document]
sandorw: I might be missing something, but it looks like the progress comes from querying the `ProgressStore` with currentTable as a table name, which I assume can only return the progress for that same table.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/BackgroundSweeperFastTest.java
```diff
@@ -92,6 +93,117 @@ public void testWritePriorityAfterSecondRunCompletesSweep() {
...
+                .minimumSweptTimestamp(4567L)
+                .startRow(new byte[] {4, 5, 6})
+                .startColumn(PtBytes.toBytes("unused"))
+                .timeInMillis(20L)
+                .startTimeInMillis(50L)
+                .build();
+        Mockito.verify(progressStore).saveProgress(
+                eq(progressAfterSecondIteration));
+    }
+
```
jeremyk-91: Tests look good. Is there a way we can verify the states of timelock / the lock service at that time?

jeremyk-91:
reminder from offline discussion - we should make the backup lock be re-entrant and counting before we merge. There is a risk of starving another user of the backup lock, but we don't anticipate deploying this in configurations where the other user is present

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
```diff
@@ -49,6 +49,19 @@
...
+     * 2. Sweep thread 2 grabs the lock for deleting stuff
+     * 3. Sweep thread 1 releases the lock
+     * 4. A backup starts
+     * 5. Sweep thread 2 releases the lock
+     *
+     * A backup taken between steps 4 and 5 is not guaranteed to be consistent, because sweep may be deleting data
+     * that it relies on.
+     */
+    @GuardedBy("this")
+    int referenceCount = 0;
```
gsheasby: I *think* this is safe, and doesn't need to be Atomic because it's only used in synchronised methods. Does that make sense? (@jeremyk-91 @sandorw  for SA in case this commit comment doesn't show up in the set of commits they view)

sandorw: For the sake of the visibility of changes, I believe this needs to be either `volatile` or an `AtomicInt`

sandorw: Looks like I was actually wrong here - `synchronized` access to `referenceCount` offers similar visibility guarantees to `volatile`.

sandorw: For reference, see comments on https://stackoverflow.com/questions/1850270/memory-effects-of-synchronization-in-java

PR #3264, opened 2018-06-14
Compaction fixes to develop

gsheasby:
**Goals (and why)**: Clean cherry-pick of #3170 to develop

* Sleep before the first compaction

This is so that short-lived CLIs/tests do not try to compact.
Compaction threads were interfering with tests of an internal product.

* Interrupt compaction thread if it's shut down while we're sleeping

**Concerns (what feedback would you like?)**: [no release notes]?

**Where should we start reviewing?**: one file

**Priority (whenever / two weeks / yesterday)**: today - required for port of compaction changes to internal develop branch

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3264)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/BackgroundCompactor.java
@@ -17,6 +17,7 @@
 package com.palantir.atlasdb.compact;
 
 import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


gsheasby:
Tagging as `do not merge` because a test that flaked, `DbKvs.testSweepHighlyVersionedCell` used to take 8-15 seconds on circle and now on two runs: 1) timed out after 50 seconds; 2) took 49 seconds and passed. Investigating.

gsheasby:
Never mind - the third time, it passed in 14 seconds, which is a duration also seen on other branches. We should consider flake-ruling the test (and in general I'm wondering if there's a way to pull test runtimes out of circle and track them).

PR #3266, opened 2018-06-14
C* Timestamps, Part 3: ETE Test

jeremyk-91:
**Goals (and why)**:
- Have an ete test that checks that SSTables are written in a sane way

- The test writes a value, overwrites it, and then forces the old version of the value to be swept.
It then checks the relevant SSTable, ensuring that it has a minimum timestamp that is recent (greater than the write timestamp of the first value) and that it has droppable tombstones.
- The test is run as part of a Cassandra+TimeLock test, and also on Cassandra+Embedded.

- Are the expectations of this test too strong?
- Tests that Atlas starts properly and behaves in a sane way already happen on the other paths, so it should be fine to just test the two paths I did.
- I kind of abused awaitility - is this _too_ gross?

**Where should we start reviewing?**: CassandraTimestampsEteTest

**Priority (whenever / two weeks / yesterday)**: this week or so

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3266)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -332,6 +332,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             CassandraMutationTimestampProvider mutationTimestampProvider) {
         super(AbstractKeyValueService.createFixedThreadPool("Atlas Cassandra KVS",
                 config.poolSize() * config.servers().size()));
+//        mutationTimestampProvider = CassandraMutationTimestampProviders.legacyModeForTestsOnly();
         this.log = log;
         this.config = config;
         this.clientPool = clientPool;
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -332,7 +332,6 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             CassandraMutationTimestampProvider mutationTimestampProvider) {
         super(AbstractKeyValueService.createFixedThreadPool("Atlas Cassandra KVS",
                 config.poolSize() * config.servers().size()));
-//        mutationTimestampProvider = CassandraMutationTimestampProviders.legacyModeForTestsOnly();
         this.log = log;
         this.config = config;
         this.clientPool = clientPool;
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.ete;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.awaitility.Awaitility;
+import org.junit.After;
+import org.junit.Before;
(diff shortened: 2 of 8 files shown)


atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .until(() -> {
+                        todoClient.runIterationOfTargetedSweep();
+                        return todoClient.doesNotExistBeforeTimestamp(id, timestamp);
+                });
+    }
+
+    private void assertMinimumTimestampIsAtLeast(String sstableMetadata, long expectedMinimum) {
+        Pattern pattern = Pattern.compile("Minimum timestamp: ([0-9]*)");
+        Matcher timestampMatcher = pattern.matcher(sstableMetadata);
+        assertThat(timestampMatcher.find()).isTrue();
```
gsheasby: Does this give a friendly error message when it fails?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
```diff
@@ -0,0 +1,103 @@
...
+        todoClient.truncate();
+    }
+
+    @After
+    public void cleanup() {
+        todoClient.truncate();
+    }
+
+    @Test
+    public void timestampsInOneSSTableAreCurrent() throws IOException, InterruptedException {
```
gsheasby: Consider adding a comment about the implications if this test fails. Out of curiosity, have you attempted to run it on a pre-3224 branch?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/cassandra/util/CassandraCommands.java
```diff
@@ -0,0 +1,53 @@
...
+public class CassandraCommands {
+    public static void nodetoolFlush(String containerName) throws IOException, InterruptedException {
+        EteSetup.execCliCommand(containerName, "nodetool flush");
+    }
+
+    public static List<String> nodetoolGetSSTables(String containerName,
+            String keyspace,
+            TableReference tableRef,
+            byte[] rowKey) throws IOException, InterruptedException {
+        String query = String.format("nodetool getsstables %s %s__%s %s",
```
gsheasby: is there a utility that converts a tablespace to %s__%s? We must have done this conversion when creating the table.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
jeremyk-91: No (it gives `expected: <true> but was: <false>`) but I've used assertj's `as` to give a more friendly error message (`[SSTableMetadata output contains the minimum timestamp] expected: <true> but was: <false>`). I did the same for the other regex match as well.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
jeremyk-91: Added several comments including an additional assertion that's important. Actually the todo table is thorough (with regards to the general comment I wanted to do one for a conservative table that I thought maybe could be a part 4).

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/cassandra/util/CassandraCommands.java
jeremyk-91: Yep - `CassandraKeyValueServiceImpl.internalTableName(tableRef)` should do it, changed to use that instead.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -332,6 +332,7 @@ private CassandraKeyValueServiceImpl(Logger log,
             CassandraMutationTimestampProvider mutationTimestampProvider) {
         super(AbstractKeyValueService.createFixedThreadPool("Atlas Cassandra KVS",
                 config.poolSize() * config.servers().size()));
+//        mutationTimestampProvider = CassandraMutationTimestampProviders.legacyModeForTestsOnly();
```
gsheasby: oops

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
```diff
@@ -58,13 +58,14 @@ public void cleanup() {
     }
 
     @Test
-    public void timestampsInOneSSTableAreCurrent() throws IOException, InterruptedException {
+    public void timestampsForRangeTombstonesAreCurrentInThoroughTables() throws IOException, InterruptedException {
```
gsheasby: better name 👍

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
jeremyk-91: oops!

jeremyk-91: and that's how I tested that things wouldn't work in the old world 😅

jeremyk-91:
Something has made the startup independence ETE tests unhappy - investigating.

PR #3267, opened 2018-06-15
serializable_cell conflict handler

gokcanozakdag:
**Goals (and why)**:
We have retry_on_write_write_cell; but we don't have an serializable analogue of this conflict handler. This PR is adding `SERIALIZABLE_CELL` aiming to give clients the ability to lock at cell level in their serializable transactions.

Conflict handler enums are keeping the information of whether it is locking at cell level or row level, and if checking w/r and w/w conflicts. SERIALIZABLE_CELL locks at cell level and checks both w/w and w/r conflicts.

Is there any other test that can be useful to add?

`ConflictHandler.java`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3267)


j-baker:
one thing that would be cool is if we could avoid checking for write write conflicts (based on our internal usage patterns). This avoids a bunch of Atlas traffic.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsSerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,63 @@
...
+import com.palantir.atlasdb.sweep.queue.SpecialTimestampsSupplier;
+import com.palantir.atlasdb.sweep.queue.TargetedSweepFollower;
+import com.palantir.atlasdb.sweep.queue.TargetedSweeper;
+import com.palantir.atlasdb.transaction.impl.AbstractSerializableTransactionWithCellLockTest;
+
+public class CassandraKvsSerializableTransactionWithCellLockTest
+        extends AbstractSerializableTransactionWithCellLockTest {
+    @ClassRule
+    public static final Containers CONTAINERS =
+            new Containers(CassandraKvsSerializableTransactionWithRowLockTest.class)
```
jeremyk-91: Should be `CellLockTest` - this might mess with the logs as we'll believe that we should put the logs in the same place as the row test

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsSerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,63 @@
...
+    protected MultiTableSweepQueueWriter getSweepQueueWriterUninitialized() {
+        return TargetedSweeper.createUninitializedForTest(() -> 128);
+    }
+
+    @Override
+    protected MultiTableSweepQueueWriter getSweepQueueWriterInitialized() {
+        TargetedSweeper queue = TargetedSweeper.createUninitializedForTest(() -> 128);
+        queue.initialize(new SpecialTimestampsSupplier(() -> 0, () -> 0), keyValueService,
+                mock(TargetedSweepFollower.class));
+        return queue;
```
jeremyk-91: These defaults scared me for a bit at first, though they are indeed used by the existing tests, so should be correct.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1785,7 +1786,8 @@ protected LockToken acquireLocksForCommit() {
                 continue;
             }
             ConflictHandler conflictHandler = getConflictHandlerForTable(tableRef);
-            if (conflictHandler == ConflictHandler.RETRY_ON_WRITE_WRITE_CELL) {
+            if (conflictHandler == ConflictHandler.RETRY_ON_WRITE_WRITE_CELL
+                    || conflictHandler == ConflictHandler.SERIALIZABLE_CELL) {
```
jeremyk-91: This is the heart of the change, and is correct.

nit: Maybe have helpers `mustCheckWriteWriteConflicts()` and `useCellLevelLocking()`?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -696,6 +703,27 @@ public void testMultipleReadsToSameColumnRange() {
...
+        return lock(rowLockDescriptor);
+    }
+
+    protected LockResponse getCellLock(String rowName, String colName) {
+        LockDescriptor cellLockDescriptor = AtlasCellLockDescriptor.of(
+                TEST_TABLE.getQualifiedName(),
+                PtBytes.toBytes(rowName),
+                PtBytes.toBytes(colName));
+        return lock(cellLockDescriptor);
+    }
```
jeremyk-91: nit: I'd prefer calling these `acquireRowLock` and `acquireCellLock` - given the return type it's clear they're trying to do work, but I don't normally expect a getter to have side effects

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,130 @@
...
+        Transaction t0 = startTransaction();
+        put(t0, "row1", "col1", "100");
+        put(t0, "row1", "col2", "101");
+        put(t0, "row2", "col1", "102");
+        t0.commit();
+
+        Transaction t1 = startTransactionWithPreCommitCondition((ignored) -> {
+            LockResponse response = getRowLock("row1");
+            assertTrue(response.wasSuccessful());
+        });
```
jeremyk-91: This is pretty clever ⭐️

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/keyvalue/MemorySerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,30 @@
...
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.impl.InMemoryKeyValueService;
+import com.palantir.atlasdb.transaction.impl.AbstractSerializableTransactionWithCellLockTest;
+import com.palantir.common.concurrent.PTExecutors;
+
+public class MemorySerializableTransactionWithCellLockTest extends AbstractSerializableTransactionWithCellLockTest {
+    @Override
+    protected KeyValueService getKeyValueService() {
+        return new InMemoryKeyValueService(false,
+                PTExecutors.newSingleThreadExecutor(PTExecutors.newNamedThreadFactory(false )));
```
jeremyk-91: nit: space

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
j-baker: or even better, have ConflictHandler have that method on it!

jeremyk-91: Yep, that would work - though there is an awkward `RETRY_ON_VALUE_CHANGED` that I think we'll still need to reference explicitly. It makes sense to add check w/w conflicts and lock levels to the enum, though I don't know what people feel in general about enum members that are equivalent in all aspects but unequal 🤷‍♂️

gokcanozakdag:
@j-baker will be adding SERIALIZABLE_INDEX in a separate pr, just after merging this one. Will be a very small change thanks to last enum changes.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
gokcanozakdag: thanks! Done 👍

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/ConflictHandlersTest.java
```diff
@@ -0,0 +1,35 @@
...
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import com.palantir.atlasdb.protos.generated.TableMetadataPersistence;
+import com.palantir.atlasdb.transaction.api.ConflictHandler;
+
+public class ConflictHandlersTest {
+    @Test
+    public void shouldConvertToProtosAndBack() {
```
j-baker: +1

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1504,9 +1504,7 @@ protected void throwIfWriteAlreadyCommitted(TableReference tableRef,
 
         if (conflictHandler == ConflictHandler.RETRY_ON_VALUE_CHANGED) {
             throwIfValueChangedConflict(tableRef, writes, spanningWrites, dominatingWrites, commitLocksToken);
-        } else if (conflictHandler == ConflictHandler.RETRY_ON_WRITE_WRITE
-                || conflictHandler == ConflictHandler.RETRY_ON_WRITE_WRITE_CELL
-                || conflictHandler == ConflictHandler.SERIALIZABLE) {
+        } else if (conflictHandler.checkWriteWriteConflicts()) {
```
j-baker: there's some simplification you can do here - notably you know this is true.

j-baker: as in you can get rid of the else block and just make this be else

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsSerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.palantir.atlasdb.containers.Containers;
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.sweep.queue.MultiTableSweepQueueWriter;
+import com.palantir.atlasdb.sweep.queue.SpecialTimestampsSupplier;
+import com.palantir.atlasdb.sweep.queue.TargetedSweepFollower;
+import com.palantir.atlasdb.sweep.queue.TargetedSweeper;
+import com.palantir.atlasdb.transaction.impl.AbstractSerializableTransactionWithCellLockTest;
+import com.palantir.lock.v2.TimelockService;
+
+public class CassandraKvsSerializableTransactionWithCellLockTest
```
j-baker: Note: you're definitely only doing this because of precedence, so i'm cool with you leaving it - but I'll ask anyway as food for thought - is there really a behaviour that having a Cassandra around will catch issues with that your unit tests won't?

j-baker: right now you're just running all the tests 3 times, but it's not like they even use the KVS particularly

j-baker:
I think your tests flaked

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionWithCellLockTest.java
```diff
@@ -0,0 +1,117 @@
...
+import com.palantir.atlasdb.transaction.api.ConflictHandler;
+import com.palantir.atlasdb.transaction.api.PreCommitCondition;
+import com.palantir.atlasdb.transaction.api.Transaction;
+import com.palantir.atlasdb.transaction.api.TransactionReadSentinelBehavior;
+import com.palantir.atlasdb.transaction.impl.logging.CommitProfileProcessor;
+import com.palantir.atlasdb.util.MetricsManagers;
+import com.palantir.lock.impl.LegacyTimelockService;
+import com.palantir.lock.v2.LockResponse;
+
+public abstract class AbstractSerializableTransactionWithCellLockTest extends AbstractSerializableTransactionTest {
```
j-baker: wait, i think it's overkill to just run all the tests again with row locks

j-baker: and cell locks

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -20,13 +20,13 @@
      * If two transactions write to the same cell the one that was committed later will win.  We
      * are ignoring write conflicts in this case.
      */
-    IGNORE_ALL,
+    IGNORE_ALL(false, false, false),
```
j-baker: think for online upgrades you specifically need a 'lock on cells', 'lock on rows' thing for migration

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1743,15 +1741,15 @@ protected LockToken acquireLocksForCommit() {
...
+            if (conflictHandler.lockCellsForConflicts()) {
                 for (Cell cell : getLocalWrites(tableRef).keySet()) {
                     result.add(
                             AtlasCellLockDescriptor.of(
                                     tableRef.getQualifiedName(),
                                     cell.getRowName(),
                                     cell.getColumnName()));
                 }
-            } else if (conflictHandler != ConflictHandler.IGNORE_ALL) {
+            } else if (conflictHandler.lockRowsForConflicts()) {
```
j-baker: make this be an 'if' not 'else if' to make live migrations work

gokcanozakdag:
@j-baker @jeremyk-91 I have added the migration conflict handler, and changed the testing workflow to avoid running unnecessary tests.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -47,7 +47,7 @@
      * Note: This ConflictHandler has the ABA problem (https://en.wikipedia.org/wiki/ABA_problem).  The easiest
      * way to fix this is to have the value always increase.
      */
-    RETRY_ON_VALUE_CHANGED,
+    RETRY_ON_VALUE_CHANGED(false, true, true, false),
```
gsheasby: Is this handler truly the same as `RETRY_ON_WRITE_WRITE`?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -60,13 +60,76 @@
...
+    /**
+     * This conflict handler is designed to be used for migrating a table from SERIALIZABLE to SERIALIZABLE_CELL
+     * conflict handler, or vice versa without requiring a shutdown.
+     * <p>
+     * Current implementation of lock service permits a client to lock a cell after locking the row that cell
+     * belongs to. In a scenario where we migrate table A to SERIALIZABLE_CELL from SERIALIZABLE, and conduct a rolling
+     * upgrade, if client_1 is on the newer version, it will try to acquire cell locks for its transaction t1; while
+     * client_2 with an older version, acquiring row locks for its transaction t2. If t1 and t2 are modifying the same
+     * cell then correctness issues may occur as there is no way for t1 to be aware of t2. Locking on both cell and row
+     * level prevents this issue.
```
gsheasby: When can clients safely migrate from `SERIALIZABLE_LOCK_LEVEL_MIGRATION` to `SERIALIZABLE_CELL`?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -60,13 +60,76 @@
...
+     * table with the main table.
+     * <p>
+     * This conflict handler locks at cell level. As index tables use dynamic columns, locking at row level would
+     * result in higher contention.
+     */
+    SERIALIZABLE_INDEX(true, false, false, true),
+
+    /**
+     * This conflict handler is designed to be used for migrating a table from SERIALIZABLE to SERIALIZABLE_CELL
+     * conflict handler, or vice versa without requiring a shutdown.
```
sandorw: simultaneous shutdown of all clients

sandorw: Maybe also clarify that once all nodes are on `SERIALIZABLE_LOCK_LEVEL_MIGRATION`, then can move to the desired end state

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommitLockTest.java
```diff
@@ -0,0 +1,216 @@
...
+import com.palantir.common.concurrent.PTExecutors;
+import com.palantir.lock.AtlasCellLockDescriptor;
+import com.palantir.lock.AtlasRowLockDescriptor;
+import com.palantir.lock.LockDescriptor;
+import com.palantir.lock.impl.LegacyTimelockService;
+import com.palantir.lock.v2.LockRequest;
+import com.palantir.lock.v2.LockResponse;
+import com.palantir.lock.v2.TimelockService;
+
+@RunWith(Theories.class)
```
gsheasby: Didn't know about this runner. Nice to learn about it! 👍

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommitLockTest.java
```diff
@@ -0,0 +1,216 @@
...
+    private LockResponse acquireCellLock(String rowName, String colName) {
+        LockDescriptor cellLockDescriptor = AtlasCellLockDescriptor.of(
+                TEST_TABLE.getQualifiedName(),
+                PtBytes.toBytes(rowName),
+                PtBytes.toBytes(colName));
+        return lock(cellLockDescriptor);
+    }
+
+    private LockResponse lock(LockDescriptor lockDescriptor) {
+        TimelockService timelockService = new LegacyTimelockService(timestampService, lockService, lockClient);
```
gsheasby: Creating a new timelock service each time we want a lock seems suspicious (but apparently it works!)

docs/source/release_notes/release-notes.rst
```diff
@@ -56,6 +56,13 @@ develop
            Note that the sweep CLI itself has been deprecated in favour of using the sweep priority override configuration, possibly in conjunction with the thread count (:ref:`Docs<sweep_tunable_parameters>`).
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3338>`__)
 
+    *    - |new|
+         - Three new conflict handlers ``SERIALIZABLE_CELL``, ``SERIALIZABLE_INDEX`` and ``SERIALIZABLE_LOCK_LEVEL_MIGRATION`` are added. 
+           ``SERIALIZABLE_CELL`` conflict handler is same as ``SERIALIZABLE``, but checks for conflicts by locking cells during commit instead of locking rows. Cell locks are more fine-grained, so this will produce less contention at the expense of requiring more locks to be acquired.
```
gsheasby: Are there docs about conflict handlers? Some of this information should be moved to or duplicated in docs, for long-term discoverability.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommitLockTest.java
```diff
@@ -0,0 +1,216 @@
...
+            LockResponse response = acquireCellLock("row1", "col2");
+            assertTrue(response.wasSuccessful());
+        };
+
+        Transaction transaction = startTransaction(canAcquireLockOnDifferentCell, conflictHandler);
+        put(transaction, "row1", "col1", "100");
+        transaction.commit();
+    }
+
+    private Transaction startTransaction(PreCommitCondition preCommitCondition, ConflictHandler conflictHandler) {
```
sandorw: Sad times, but I don't think we can easily work around this.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommitLockTest.java
```diff
@@ -0,0 +1,216 @@
...
+                true,
+                timestampCache,
+                AtlasDbConstants.DEFAULT_TRANSACTION_LOCK_ACQUIRE_TIMEOUT_MS,
+                AbstractTransactionTest.GET_RANGES_EXECUTOR,
+                AbstractTransactionTest.DEFAULT_GET_RANGES_CONCURRENCY,
+                MultiTableSweepQueueWriter.NO_OP,
+                MoreExecutors.newDirectExecutorService(),
+                CommitProfileProcessor.createNonLogging(metricsManager)) {
+            @Override
+            protected Map<Cell, byte[]> transformGetsForTesting(Map<Cell, byte[]> map) {
```
sandorw: The method that necessitates this on `SerializableTransactionManager` is scary...

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
gsheasby: Ah, I see the difference is in the exception handling [here](https://github.com/palantir/atlasdb/pull/3267/files#diff-bf062bd4c13b8c0022080ca8ef49c4d2R1515).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -60,13 +60,76 @@
...
+    SERIALIZABLE_CELL(true, false, true, true),
+
+    /**
+     * This conflict handler is designed to be used by index tables. As any write/write conflict on an index table
+     * will necessarily also be a write/write conflict on base table, we do not need to check write/write conflicts
+     * on index tables. Read/write conflicts should still need to be checked, since we do not need to read the index
+     * table with the main table.
+     * <p>
+     * This conflict handler locks at cell level. As index tables use dynamic columns, locking at row level would
+     * result in higher contention.
```
jeremyk-91: Nice, thanks for the explanation

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1753,15 +1749,17 @@ protected LockToken acquireLocksForCommit() {
...
                     result.add(
                             AtlasCellLockDescriptor.of(
                                     tableRef.getQualifiedName(),
                                     cell.getRowName(),
                                     cell.getColumnName()));
                 }
-            } else if (conflictHandler != ConflictHandler.IGNORE_ALL) {
+            }
+
+            if (conflictHandler.lockRowsForConflicts()) {
```
jeremyk-91: Good - we still only make one call to timelock even if we're taking both row and cell locks

docs/source/release_notes/release-notes.rst
```diff
@@ -56,6 +56,13 @@ develop
...
+           ``SERIALIZABLE_CELL`` conflict handler is same as ``SERIALIZABLE``, but checks for conflicts by locking cells during commit instead of locking rows. Cell locks are more fine-grained, so this will produce less contention at the expense of requiring more locks to be acquired. 
+           ``SERIALIZABLE_INDEX`` conflict handler is designed to be used by index tables. As any write/write conflict on an index table will necessarily also be a write/write conflict on base table, this conflict handler does not check write/write conflicts. Read/write conflicts should still need to be checked, since we do not need to read the index table with the main table. This conflict handler also locks at cell level.
+           If your schema already has a table with ``SERIALIZABLE`` conflict handler, and you would like to migrate it to ``SERIALIZABLE_CELL`` or ``SERIALIZABLE_INDEX`` with a rolling upgrade (without a shutdown); then you should first migrate it to ``SERIALIZABLE_LOCK_LEVEL_MIGRATION`` conflict handler to avoid data corruption.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3267>`__)
```
jeremyk-91: Good, thanks. There's a section on Conflict Detection in `performance_ideas.rst` that could be worth a look/update

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommitLockTest.java
jeremyk-91: This works because the underlying lock service is consistent, but we really should just store the single timelock service somewhere

jeremyk-91: It delegates to the same underlying `lockService` :) though maybe best not to rely on that

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
j-baker: when all nodes are running with one, or the other

j-baker: it's safe to run S_C with S_L_L_M and it's safe to run S_L_L_M with S, but not S_C with S. So you have a quick online migration for the change.

PR #3268, opened 2018-06-15
C* Timestamps, Part 4: Conservative ETE Test

jeremyk-91:
**Goals (and why)**: Write an ETE test that checks the new behaviour for sweep sentinels (this arguably is the more significant change in #3224)

- Make a new NamespacedTodoTable that uses a conservative sweep strategy (and happens to use dynamic columns, in case we want future tests that use those)

**Where should we start reviewing?**: CassandraTimestampsEteTest

**Priority (whenever / two weeks / yesterday)**: early next week - good for noting if we accidentally touch the timestamp logic again

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3268)

### DIFF
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
@@ -20,6 +20,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -31,19 +32,18 @@ import org.junit.Test;
 
 import com.google.common.collect.Iterables;
 import com.palantir.atlasdb.ete.cassandra.util.CassandraCommands;
-import com.palantir.atlasdb.keyvalue.api.Namespace;
-import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.table.description.ValueType;
 import com.palantir.atlasdb.todo.ImmutableTodo;
 import com.palantir.atlasdb.todo.Todo;
 import com.palantir.atlasdb.todo.TodoResource;
-import com.palantir.atlasdb.todo.generated.TodoTable;
+import com.palantir.atlasdb.todo.TodoSchema;
 
 public class CassandraTimestampsEteTest {
     private static final Todo TODO = ImmutableTodo.of("todo");
     private static final Todo TODO_2 = ImmutableTodo.of("todo_two");
     private static final String CASSANDRA_CONTAINER_NAME = "cassandra";
     private static final long ID = 1L;
+    private static final String NAMESPACE = "tom";
 
     private TodoResource todoClient = EteSetup.createClientToSingleNode(TodoResource.class);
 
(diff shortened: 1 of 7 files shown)


atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
```diff
@@ -87,12 +87,42 @@ public void timestampsForRangeTombstonesAreCurrentInThoroughTables() throws IOEx
...
+        // original value (unlikely since we confirmed no value existed at the first write timestamp), or
+        // that the sweep sentinel was written at a lower Cassandra timestamp than the original value (and hence not
+        // a fresh timestamp).
+        assertMinimumTimestampIsAtLeast(sstableMetadata, firstWriteTimestamp);
+
+        // Failure here means that the range tombstone isn't droppable, which would be weird
+        assertDroppableTombstoneRatioPositive(sstableMetadata);
+    }
+
+    private void sweepUntilNoValueExistsAtTimestamp(long id, long timestamp, Optional<String> namespace) {
```
gsheasby: The optional parameter is slightly hacky here. I guess you're trying to avoid duplicating the Awaitility logic, but this could be done without an optional parameter by having a helper method that takes a boolean, e.g.
```
private void sweepUntilNoValueExistsForNamespaceAtTimestamp(long id, String namespace, long timestamp) {
    sweepUntilConditionSatisfied(
        () -> todoClient.namespacedTodoDoesNotExistBeforeTimestamp(id, timestamp, namespace));
}
```

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
```diff
@@ -87,12 +87,42 @@ public void timestampsForRangeTombstonesAreCurrentInThoroughTables() throws IOEx
...
+                CASSANDRA_CONTAINER_NAME, Iterables.getOnlyElement(ssTables));
+
+        // The table should contain the first value, and timestamps thereafter should be higher.
+        // Failure here means either that the range tombstone was written at a lower Cassandra timestamp than the
+        // original value (unlikely since we confirmed no value existed at the first write timestamp), or
+        // that the sweep sentinel was written at a lower Cassandra timestamp than the original value (and hence not
+        // a fresh timestamp).
+        assertMinimumTimestampIsAtLeast(sstableMetadata, firstWriteTimestamp);
+
+        // Failure here means that the range tombstone isn't droppable, which would be weird
```
gsheasby: This would happen if the range tombstone was written at the wrong C* timestamp, right?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
jeremyk-91: Yep, having a BooleanSupplier condition works well! Changed to use one of these

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimestampsEteTest.java
jeremyk-91: If I'm not wrong right now these are just calculated as a ratio of tombstones to values - see https://issues.apache.org/jira/browse/CASSANDRA-13024. So this would be 0 if the tombstone was somehow already dropped which shouldn't happen. I've clarified the comment - `weird` isn't a good description.

jeremyk-91:
Awesome, thanks @gsheasby - merging 🎉

PR #3271, opened 2018-06-17
Use a daemon executor

j-baker:
Internally we've had a few cases recently where we've failed to shut
down properly. All the executors _we_ create are properly shut down
when a termination signal is seen, and this is the case almost all
of the time in AtlasDB, except in this place, which is brand new (and does not use daemon threads).

I've diffed our dependencies, and this seems like the most likely
place for our issues to show up.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3271)


j-baker:
related: https://github.com/palantir/atlasdb/pull/3237/commits/1908f74b6bcb060d2b605bbfa55c7caa8318306b

j-baker:
which flaked for me because someone was shutting down the executor before it should have been (kinda the opposite of this issue!)

j-baker:
Please don't merge this until https://github.com/palantir/atlasdb/pull/3237 is merged!

sandorw:
We should make the same change as we propagate breaks to the big internal product.

PR #3272, opened 2018-06-18
The static cached executors in BasicSQL are configurable

cakofony:
A configured executor may be supplied rater than the default
implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3272)

### DIFF
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorServiceFactory.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorServiceFactory.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.common.concurrent;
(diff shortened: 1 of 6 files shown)


commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorServiceFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    ExecutorService getCachedExecutorService(String threadNamePrefix, int threadIdleTimeoutMillis);
+
+    /**
+     * Gets an unbounded {@link ExecutorService} with daemon threads using the provided thread name prefix.
+     * Creates an executor using the default idle timeout.
+     */
+    ExecutorService getCachedExecutorService(String threadNamePrefix);
+
+    /**
+     * Get the priority of this factory. The factory service with the highest priority will be used.
```
gsheasby: This logic seems hacky to me.
Notes from syncing up with @cakofony :
* The hackery is because the services in `BasicSQL` are static
* It appears that one of them, `service`, need not be static
* The other, `executeService`, is used in a public static method
* it's an open question how often that public static method is used outside of AtlasDB code

Outcome: @cakofony will look in to adding a parameter to the static method that uses `executeService`, and to making `service` a field that can be initialised in the constructor on the internal product side.

jeremyk-91: @cakofony did you have any bandwidth to look at the above recently?

tboam:
@cakofony Have you had a chance to look through the comments on this PR yet?  We're happy to merge once you've looked into the suggested refactor.

jeremyk-91:
@cakofony for SA - I'm going to open a separate branch starting from develop as it'll be easier for me to verify that this merges well (I think it does, but the merge is pretty large), so this might get closed today or tomorrow but there should be a replacement coming up

jeremyk-91:
@cakofony I'm closing this out as #3487 which makes the executors configurable has merged - please let us know if you need anything from us on this.

cakofony:
Thanks Jeremy!

PR #3273, opened 2018-06-18
Faster Firing Transactions (FFT) Part 1: Snapshot Transaction Profiling

jeremyk-91:
TODOs:
- [x] Add unit tests
- [ ] Make the rate limit interval configurable?

**Goals (and why)**:
- Expose better logging and metrics for "what takes up time when committing writes"

- Log more individual stages in commitWrites()
- Have a rate limiter to allow normal services to log this information once every time interval (currently set to 5 seconds).
- Build some infrastructure to manage conditional logging while not evaluating a not-so-cheap operation more than once (and zero times if we're not logging it)

- Should the interval be configurable?
- Is the infrastructure too much overhead?

**Where should we start reviewing?**: `LogTemplate`

**Priority (whenever / two weeks / yesterday)**: this week, I'd like to get metrics soon

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3273)


jeremyk-91:
@tboam The main issue is that from separate metrics I don't have a good way of estimating the overhead. The metrics give us various order stats (p95, p99 etc), so I can compare the p99s of transaction parts against the overall p99; but from a statistical point of view it's not quite right. 

If I have `A + B = C`, where `A` is KVS operations and `B` is other transaction overhead (so `C` is total transaction time), I'm interested in the distributions of `B` and `B/C`. However right now we don't have `B` (we only have some components that sum to `B`). The best I can do right now to estimate the ratio would be something like `p99(B)/p99(C)`, but I can't say much about `B/C` from that!

I probably want to add new metrics that are `B` and `B/C` actually, in addition to the other pieces already here.

jeremyk-91:
@tboam should be good for review again - changes were non-trivial so probably worth a second look before merging

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1305,70 +1312,113 @@ private void commitWrites(TransactionService transactionService) {
...
+                    });
         } finally {
             timelockService.unlock(ImmutableSet.of(commitLocksToken));
         }
     }
 
+    private void updateNonPutOverheadMetrics(long microsWritingToTargetedSweepQueue, long microsForWrites,
+            long microsForCommitStage) {
+        long nonPutOverhead = microsForCommitStage - microsForWrites - microsWritingToTargetedSweepQueue;
+        getTimer("nonPutOverhead").update(nonPutOverhead, TimeUnit.MILLISECONDS);
```
tboam: I think this is in micros as all of the inputs to the calculation are in micros

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/logging/ChainingLogConsumerProcessorTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final LogConsumerProcessor SWALLOWING_PROCESSOR = unused -> {};
+
+    @SuppressWarnings("unchecked") // Mocking
+    private final Supplier<LogTemplate> templateSupplier = mock(Supplier.class);
+
+    private LogConsumerProcessor chainingProcessor;
+
+    @Test
+    public void doesNotInvokeSupplierIfNoDelegates() {
+        givenProcessors();
```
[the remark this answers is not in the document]
jeremyk-91: It seemed like a good fit for a bunch of tests that would otherwise have a lot of repetition :)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: Ooh, good catch! Will fix. Yep, a better fix is to switch everything out to Duration but agree that in this instance this fix will do

PR #3275, opened 2018-06-19
Add logging to leadership election

datval:
**Goals (and why)**:
Lock leader election fails for internal product, adding more logging to debug the issue.

Just adding some logging lines.

This week maybe?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3275)


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -165,6 +168,7 @@ private void proposeLeadershipOrWaitForBackoff(LeadershipState currentState)
         }
 
         long backoffTime = (long) (randomWaitBeforeProposingLeadership * Math.random());
+        log.debug("Waited for [{}] ms before proposing leadership", SafeArg.of("waitTimeMs", backoffTime));
```
schlosna: The log message indicates that the wait has already occurred, but the sleep is below. Seems like we either want to move the log line after sleep or adjust message wording indicating that it will wait for given backoff

datval:
Fixed the message. I don't have permissions to merge

jeremyk-91:
Checkstyle failed. I'll fix this

PR #3278, opened 2018-06-19
less tablescans for grabbing CKVS tablemetadata

clockfort:
the diff view for internalPutMetadataForTables() sucks since its mostly a rewrite, you might want to look at the versions separately.

The drop table metadata cleanup case is also addressed by the new putMetadataForTables change, because a 'delete' of the table metadata is currently accomplished via  putMetadata(myTable, byte[0])

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3278)

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
@@ -79,7 +79,10 @@ import com.palantir.atlasdb.keyvalue.impl.AbstractKeyValueServiceTest;
 import com.palantir.atlasdb.keyvalue.impl.TableSplittingKeyValueService;
 import com.palantir.atlasdb.keyvalue.impl.TracingPrefsConfig;
 import com.palantir.atlasdb.protos.generated.TableMetadataPersistence;
+import com.palantir.atlasdb.table.description.ColumnMetadataDescription;
+import com.palantir.atlasdb.table.description.NameMetadataDescription;
 import com.palantir.atlasdb.table.description.TableDefinition;
+import com.palantir.atlasdb.table.description.TableMetadata;
 import com.palantir.atlasdb.table.description.ValueType;
 import com.palantir.atlasdb.transaction.api.ConflictHandler;
 import com.palantir.atlasdb.util.MetricsManager;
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1566,6 +1576,12 @@ private void createTablesInternal(final Map<TableReference, byte[]> tableNamesTo
         } else {
             log.debug("Found table metadata for {} at matching name {}", LoggingArgs.tableRef(tableRef),
                     LoggingArgs.tableRef("matchingTable", match.get().getKey()));
+
+            // rewrite this row with the key we initially expected, so we don't have to do this again
+            putMetadataAndMaybeAlterTables(
```
sandorw: Let's delete the old versions, since a future delete would clean up the lower case version and the stale versions would become visible again.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -416,6 +419,71 @@ public void cassandraTimestampsAreNotUsedAsAtlasTimestampsForRangeTombstone() th
...
+
+        assertThat(
+                Arrays.equals(keyValueService.getMetadataForTable(userTable), totallyLegitTableMetadata),
+                is(true));
+    }
+
+    @Test
+    public void metadataForNewTableIsLowerCased() {
+        TableReference userTable = TableReference.createFromFullyQualifiedName("test.xXcOoLtAbLeNaMeXx");
+        byte[] originalMetadata = new TableMetadata(
```
sandorw: nit: static var to avoid redundant declarations

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1666,48 +1676,59 @@ public void putMetadataForTables(final Map<TableReference, byte[]> tableRefToMet
...
+                    // found existing metadata at new name, but we're performing an update
+                    updatedMetadata.put(newCell, tableRefToMetadata.get(tableRef));
+                    updatedCfs.add(getCfForTable(tableRef, tableRefToMetadata.get(tableRef), config.gcGraceSeconds()));
+                }
+            } else if (existingMetadataAtOldName.containsKey(tableRefToOldCell.get(tableRef))) {
+                if (metadataIsDifferent(
+                        existingMetadataAtOldName.get(tableRefToOldCell.get(tableRef)).getContents(),
+                        tableRefToMetadata.get(tableRef))) {
+                    // found existing metadata at old name, but we're performing an update
+                    updatedMetadata.put(tableRefToOldCell.get(tableRef), tableRefToMetadata.get(tableRef));
```
sandorw: So we keep the old name in cases where it's used?

clockfort: Yep, we continue to use the old name, in read, update, and delete cases.
The main improvements are thus for {existing deployments with old tables that were always lowercased} and {old and new deployments who will just write out all future tables more nicely}.

The negative is a light amount of legacy tech debt and bonus complexity baked into CKVS.

sandorw: I will always 👍 bonus complexity

PR #3281, opened 2018-06-19
FFT Part 2: Timers For Setup and Finalization in SnapshotTransactionManager

jeremyk-91:
**Goals (and why)**:
- Track timing for the setup and tear-down of a TransactionTask in SnapshotTransactionManager.
- There is an RPC involved in each part, plus some additional work.
- Useful for tracking improvements since the RPC in the teardown step is likely to be my first 'real' target.

- Add two timers in SnapshotTransactionManager, one for the setup phase of a task, and one from after the task finishes to the actual end of runTaskWithLockThrowOnConflict.

**Where should we start reviewing?**: SnapshotTransactionManagers.java

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3281)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerTest.java
```diff
@@ -16,34 +16,46 @@
...
 import com.palantir.lock.CloseableLockService;
 import com.palantir.lock.LockClient;
+import com.palantir.lock.LockRefreshToken;
 import com.palantir.lock.LockService;
 import com.palantir.lock.impl.LegacyTimelockService;
 import com.palantir.timestamp.InMemoryTimestampService;
 
 public class SnapshotTransactionManagerTest {
+    private static final String SETUP_TASK_METRIC_NAME =
+            SnapshotTransactionManager.class.getCanonicalName() + ".setupTask";
```
gsheasby: In future, if we have a bunch of these task metric names, it might be good to have a constant that we can use both in production and test code.

I'd rather you didn't make the change now, though, because it's not that bad at the moment, and I'd like to see this merged in quickly so that it can be included in the next release.

jeremyk-91: Yep. Makes sense to centralise the metrics names if there are many of them, but won't do this yet.

PR #3282, opened 2018-06-19
[wip] FFT Part 3: The First Fusion - unlockCommitLock() and unlockImmutableTimestamp()

jeremyk-91:
**Note to self/reviewers: This needs to be rebased on develop - currently it's based on top of a branch that is develop + #3273 + #3281.**

- [ ] unit tests
- [ ] changelog
- [ ] smoke test

**Goals (and why)**:
- Batch unnecessary separate RPCs to timelock. There are 2 that are safely batch-able, and a 3rd that is more contentious.
- This PR includes an implementation of both batch-2 and batch-3, but all code paths go through batch-2 for now.

- Add a lock manager to a snapshot transaction, that keeps track of tokens for both the immutable timestamp and the row lock
- Implement algorithms Batch-2 and Batch-3:
  - Batch-2 combines unlocking the commit lock and unlocking the immutable timestamp lock
  - Batch-3 rolls the expiry check in as well. The danger here is that we may end up with more conflicts and/or readers rolling back our write transactions. However, Batch-3 is good for tables that are immutable or otherwise have low contention.
- Wire things up so that, for now, Batch-2 is always used. Batch-3 needs to be rolled out more carefully if we decide to go ahead with it, and possibly controlled at a finer level.

- Is it indeed the case that after you call unlock() once on a SimpleTransactionLockManager, subsequent calls don't make an RPC?
- Does the TransactionLockManager need to be synchronized in any way?
- Does this actually improve performance?

**How should we measure success as a perf improvement?**:
- This PR should cause the `SnapshotTransactionManager.finishTask` timer to take a considerable drop. Assuming that most transactions are successful this should eliminate the RPC where we unlock the immutable timestamp lock at the very end of the transaction.

**Where should we start reviewing?**: SimpleTransactionLockManager


jeremyk-91:
On second thoughts, there is an easier way to do this. Closing out - will reopen if we decide to push this approach.

PR #3283, opened 2018-06-19
smaller batch size

j-baker:
Right now it'll read up to 100k cells in one go, which is quite a lot (internally we got a ticket from the ops guys).

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3283)


gmaretic:
We can reduce the batch size, but this is not going to work for dedicated rows, as the current implementation puts all the writes from a single reference cell to the same batch. It is nontrivial to avoid doing so, because we then need to maintain more state than just the last swept timestamp between iterations or we need to delete entries from the queue appropriately after each iteration, which we are not currently doing.

gmaretic:
@j-baker is this what you wanted? Kind of took over the PR, so would like your 👍 to approve

gmaretic:
Merging since @j-baker OK'd it

PR #3286, opened 2018-06-20
log before removing from blacklist

gokcanozakdag:
**Goals (and why)**:
Currently we have two ways of un-blacklisting a host. One of them is running a background thread that healthcheck blacklisted hosts periodically and adds them back into pool if healthy (we are logging messages for this one). 
Second one is adding the host back into pool if we receive a successful response. It would be easier to debug some issues if we had logging for this one also.

This week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3286)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -392,6 +392,8 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
 
             try {
                 V response = runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                log.info("Added host {} back into the pool after receiving a successful response",
+                        SafeArg.of("host", CassandraLogHelper.host(hostPool.getHost())));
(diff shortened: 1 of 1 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -392,6 +392,8 @@ void runOneTimeStartupChecks() {
 
             try {
                 V response = runWithPooledResourceRecordingMetrics(hostPool, req.getFunction());
+                log.info("Added host {} back into the pool after receiving a successful response",
```
gsheasby: nit: it's a tiny duplication, but did you consider factoring out a method to hold both the log statement and the `blacklist.remove` call?

gokcanozakdag: yeah, I was in between. The reason why I choose this was, if I extract it to a function it can only be used for this specific case (removing the host from blacklist after receiving a successful response, couldn't be used if you are adding the host back to pool for some other reason because of the specific log message); so I should have named it to something like removeFromBlacklistAfterResponse(). But yeah, I was in between.

PR #3287, opened 2018-06-20
Add SchemaMutationLockReleasingRule to OneNodeDownTestSuite

gsheasby:
**Goals (and why)**: Have seen this suite flake a number of times in the class method - followed [this advice](https://automationrhapsody.com/retry-junit-failed-tests-immediatelly/).

**Implementation Description (bullets)**: see PR title

**Concerns (what feedback would you like?)**: Will this actually work? The arguments seemed a bit sketchy to me.

**Where should we start reviewing?**: OneNodeDownTestSuite

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3287)


atlasdb-cassandra-multinode-tests/src/test/java/com/palantir/cassandra/multinode/OneNodeDownTestSuite.java
```diff
@@ -35,6 +37,10 @@
         })
 public final class OneNodeDownTestSuite extends NodesDownTestSetup {
 
+    @ClassRule
+    public final RuleChain ruleChain = SchemaMutationLockReleasingRule.createChainedReleaseAndRetry(
+            createCassandraKvs(), CONFIG);
```
jeremyk-91: This one should probably be a standard Rule, not a ClassRule - as written I think this releases the lock at the end of the tests rather than after each test.

Not too sure what `FlakeRetryingRule` does as a ClassRule as well - I don't think it actually retries stuff, but could be mistaken!

gsheasby: I think we actually want a class rule, becasue this test is flaking in the BeforeClass part. My aim was to have this part be retried. Not sure if that's plausible though?

jeremyk-91:
Right, I see (and it's plausible that this should work in the event of a beforeclass failing, though a test would be good).

I think we need to wrap this and the other class rule in a rule-chain with the container setup being outside, because the order is [undefined](https://junit.org/junit4/javadoc/4.12/org/junit/ClassRule.html), so if I'm understanding things correctly then you can either have:

- Container set up is outside: if BeforeClass fails, then it's caught by the flake rule and tried again, after releasing the SML.
- Container set up is inside: if BeforeClass fails, _the containers are torn down, then the flake rule kicks in, and brings up the containers again_.

jeremyk-91:
One more thing: you'll need the ShouldRetry annotation on the class! Unfortunately it isn't a described item as far as the tests are concerned, so you'll have to apply it to the class level. I'll be opening a PR shortly that has an explanation of what's going on

gsheasby:
Thanks for doing the legwork here - fixing this up now.

atlasdb-cassandra-multinode-tests/src/test/java/com/palantir/cassandra/multinode/OneNodeDownTestSuite.java
```diff
@@ -33,11 +36,16 @@
...
 
+    // Reusing containers when the FlakeRule kicks in.
     @ClassRule
-    public static final Containers CONTAINERS = new Containers(NodesDownTestSetup.class)
-            .with(new ThreeNodeCassandraCluster());
+    public static final RuleChain ruleChain = RuleChain
+            .outerRule(new Containers(NodesDownTestSetup.class)
+                    .with(new ThreeNodeCassandraCluster()))
+            .around(SchemaMutationLockReleasingRule.createChainedReleaseAndRetry(
+                    createCassandraKvs(), CONFIG));
```
jeremyk-91: ooh, I didn't know you can wrap a RuleChain in a RuleChain (though it kind of makes sense!)

gsheasby: :yodawg:

PR #3288, opened 2018-06-20
retry on different host if cassandra load is high in default strategy

gokcanozakdag:
**Goals (and why)**:
Internal ticket pds-69886 indicates that default retry strategy gets stuck in retrying on the same node while that C* node is coming back to life. I don't see a reason not to try on a different host if we have high load (either on cassandra cluster as a whole or on a single node).

Conservative strategy retries on a different host if exception is indicative of C* load already. Is there a reason why we used a different logic for default strategy?

This week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3288)


gokcanozakdag:
@jeremyk-91 I think logging why we are retrying on a different host is useful. Just added that.

gokcanozakdag:
Closing this one as observing how conservative strategy behaves on the field, then moving to that all together would make more sense then modifying the default one.

PR #3291, opened 2018-06-20
C* Request Exception Handler: Log number of tries, not the Supplier

### LINKED ISSUE #3294: CassandraRequestExceptionHandler is not logging maxTotalTries correctly
CassandraRequestExceptionHandler is not logging maxTotalTries correctly. The SafeArg is using the Supplier instead of the Supplier's value.

jeremyk-91:
**Goals (and why)**:
- Fixes #3294.
- Fix bug where we log a supplier rather than the value it returns.

- Call get on the supplier

- Technically this retrieves the value AFTER the decision to make an attempt is made, meaning you can get weirdness like `attempt 10 of 2`.

**Where should we start reviewing?**: CassandraRequestExceptionHandler.java

@SerialVelocity for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3291)


gsheasby:
Presciently, fixes #3294.

PR #3292, opened 2018-06-20
Revert "Excavator: Upgrades Baseline to the latest version (#3277)"

gsheasby:
This reverts commit 4fe83411946621799efd9794ad05b7cc2f22a99e.

**Goals (and why)**: After this, we have 2,362 baseline-checkstyle errors instead of about 11,500. That sounds like a (marginally) more manageable number...

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3292)


gsheasby:
I think these errors are mostly (entirely?) down to Baseline-Checkstyle (on intellij) not taking into consideration the list of blacklisted projects. So hopefully the next step would be to exclude those projects from the Intellij tests, too.

snegrus-adb:
What's the plan in the end? Stay with the lower number of errors, fix them, add back that commit and fix the rest of the errors?

gsheasby:
So I counted the errors. There are 2362 baseline errors according to intellij. 2191 (over 90%) come from projects that are blacklisted. Of those, 1537 (65%) come from three projects; the rest are from about ten other projects.

If we switch to actually use baseline-checkstyle in gradle, we'll find about 171 errors.

gsheasby:
The plan would roughly be:
First PR (this one): Switch back to baseline 18 (so that the number of errors outside blacklisted projects is small enough to fix in one go)
Second PR: Suppress the 12 projects that are suppressed *and* have errors, switch gradle to actually use baseline, and fix or suppress the ~171 other errors
two-and-a-halfth PR (maybe a few PRs): Unsuppress the projects that have between 1 and 159 errors 
Third PR: Upgrade to baseline 20, suppressing the more annoying rules to make this tractable

Suppressed projects and their errors:
atlasdb-client | 379 | 
atlasdb-client-protobufs | 0 | Generated code
atlasdb-commons | 540 |  
atlasdb-console | 90 |  
atlasdb-dbkvs-hikari | 41 |  
atlasdb-jdbc | 124 | Deprecated
atlasdb-jdbc-tests | 1 | Deprecated
atlasdb-service-server | 0 | Deleted!
cassandra-partitioner | 62 | single file
commons-db | 618 |  
commons-executors | 17 |  
leader-election-impl | 138 |  
lock-impl | 159 |  
profile-client-protobufs | 0 |  Generated code
timestamp-impl | 22 |

gmaretic:
That sounds good! Thanks :)

PR #3293, opened 2018-06-20
[FlakeRule] BeforeClass example

jeremyk-91:
**Goals (and why)**:
- Give an example of how to use the flake rule to protect against a flaky BeforeClass scenario

- Give an example of how to use the flake rule to protect against a flaky BeforeClass scenario

None in particular

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3293)

### DIFF
--- a/flake-rule/src/test/java/com/palantir/flake/BeforeClassFlakeTest.java
+++ b/flake-rule/src/test/java/com/palantir/flake/BeforeClassFlakeTest.java
@@ -1,0 +1,45 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


gsheasby:
Thanks for adding this! :tada:

flake-rule/src/test/java/com/palantir/flake/BeforeClassFlakeTest.java
```diff
@@ -0,0 +1,45 @@
...
+package com.palantir.flake;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+@ShouldRetry(numAttempts = 3)
```
gsheasby: ...did this fail twice?

jeremyk-91: Nope, I just picked a random number :) changed to 2.

PR #3303, opened 2018-06-21
FFT - 1st ed., Part 3: Async TimeLock Unlocker

jeremyk-91:
To do in this PR
- [x] 1 or 2 more smoky/fuzzy tests

To do in Part 5
- [ ] Short ADR for the async timelock unlocker

**Goals (and why)**: Enable faster-firing transactions, by removing synchronous unlocking of row locks and the immutable timestamp at the end of one (there's no need for these to be synchronous)

- Expose an 'Unlock but I don't care about the result" API on TimelockService
- Unlock the locks at the end of a transaction asynchronously.
- Batching logic so that only 1 unlock request is outstanding to timelock at a time.

- Is there an error in the concurrency?
- There was some complexity in avoiding edge cases especially with the background kicking job - is this worthwhile?
- The original version of Part 3 merged the two unlock calls into one - this doesn't because accumulating seems cheap enough to me. Is this still worth doing?

**How do we tell if we improved perf?**:
Time for the post-commit work in SnapshotTransactionManager should go down, as would the extra stage added in #3307 

**Where should we start reviewing?**: Async TimeLock Unlocker

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3303)


lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Adds all provided lock tokens to a queue to eventually be scheduled for unlocking.
+     * Locks in the queue are unlocked asynchronously, and users must not depend on these locks being unlocked /
+     * available for other users immediately.
+     *
+     * @param tokens Lock tokens to schedule an unlock for.
+     */
+    public void enqueue(Set<LockToken> tokens) {
+        outstandingLockTokens.getAndAccumulate(tokens, Sets::union);
```
sandorw: What's the rationale for this vs. just using a concurrent set implementation and adding the elements directly? 

I'm worried about the number of set iterators we might have sitting with this implementation - could be dozens of unioned and diffed Sets in here. Of course, we'd have to be more careful about read/write contention in this class with the other approach.

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -0,0 +1,114 @@
...
+        scheduleIfNoTaskRunning();
+    }
+
+    private void scheduleIfNoTaskRunning() {
+        if (available.compareAndSet(true, false)) {
```
sandorw: Naming seems off here - this is really scheduling a new task unless another task is concurrently being scheduled, since `submit` is async. Are you just concerned with excessive numbers of tasks on the executor?

I don't have a good sense of how long the `submit` call takes vs. `compareAndSet`, so I don't know how many extra tasks we're saving vs. the cost of each `unlockOutstanding`.

I have a rough idea for an alternative design, will discuss it offline.

lock-api/src/test/java/com/palantir/lock/client/AsyncTimeLockUnlockerTest.java
```diff
@@ -0,0 +1,154 @@
...
+import com.palantir.lock.v2.TimelockService;
+
+public class AsyncTimeLockUnlockerTest {
+    private static final int ONE_THOUSAND = 1000;
+
+    private List<LockToken> tokenList;
+    private List<LockToken> unlockedTokens;
+
+    private TimelockService timelockService = mock(TimelockService.class);
+    private AtomicReference<Set<LockToken>> outstandingTokens = new AtomicReference<>(Sets.newConcurrentHashSet());
```
sandorw: The real impl starts out with an `ImmutableSet` here - why change the behavior for the test?

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
gokcanozakdag: If I correctly understand the behavior here, if tryUnlock is taking longer then the time between two enqueues, we won't be batching the tokens, instead we will continue to call unlock in separate requests.

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -71,10 +71,31 @@ default boolean isInitialized() {
     @Path("refresh-locks")
     Set<LockToken> refreshLockLeases(Set<LockToken> tokens);
 
+    /**
+     * Releases locks associated with the set of {@link LockToken}s provided.
+     * The set of tokens returned are the tokens for which the associated locks were unlocked in this call.
+     * It is possible that a token that was provided is NOT in the returned set (e.g. if it expired).
+     * However, in this case it is guaranteed that that token is no longer valid.
```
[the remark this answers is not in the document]
jeremyk-91: We don't actually use the output in Atlas. It can be useful to distinguish this if you wanted to do a 'check I still have the locks, and then release them' type of operation - this was considered as a possible tx perf improvement (and was going to be used as the next step on the #3282 route)

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -71,10 +71,31 @@ default boolean isInitialized() {
...
+     * @param tokens Tokens for which associated locks should be unlocked.
+     * @return Tokens for which associated locks were unlocked.
+     */
     @POST
     @Path("unlock")
     Set<LockToken> unlock(Set<LockToken> tokens);
 
+    /**
+     * A version of {@link TimelockService#unlock(Set)} where one does not need to know whether the locks associated
+     * with the provided tokens were successfully unlocked or not.
```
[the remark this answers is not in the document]
jeremyk-91: Discussed offline: this is true if you're making a request directly to a timelock server, but not necessarily true if you're using some wrappers on a proxy that talks to a timelock server. (Conceptually in `TimeLockClient`, we could schedule something async and have the JVM crash before the locking call gets scheduled)

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
jeremyk-91: Discussed offline: alternative design makes sense, and also eliminates the need for the background scheduled task.

Originally the set to false was in the wrong place (should have been at end of unlockOutstanding).

sandorw: I believe it should be at the beginning of `unlockOutstanding`.

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
jeremyk-91: Fair enough - main thing was that this gives a very easy way of draining the set in `unlockOutstanding`. Discussed offline - we now have a solution that uses read-write locks to maintain high concurrency in adding to a straight concurrentset, and keeps the synchronized bits small.

lock-api/src/test/java/com/palantir/lock/client/AsyncTimeLockUnlockerTest.java
jeremyk-91: No reason to - and yeah this should have been an `ImmutableSet` in the previous version. It's now a `ConcurrentHashSet` as that's what the real impl uses

jeremyk-91: Actually removed as it's no longer meaningful to test that adding directly to the set won't leave stragglers (it will, but that's not a valid state any longer, at least in theory)

jeremyk-91:
There is some bug with the read-write locks right now (read unlock does not happen-before write lock it seems). The PR should still be reviewable for the most part - what's left is to fix the concurrency semantics

lock-api/src/test/java/com/palantir/lock/client/AsyncTimeLockUnlockerTest.java
```diff
@@ -1,11 +1,11 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
sandorw: Checkstyle sent this file back in time

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -39,59 +36,59 @@
  * There is another layer of retrying below us (at the HTTP client level) for external timelock users.
  * Also, in the event we fail to unlock (e.g. because of a connection issue), locks will eventually time-out.
  * Thus not retrying is reasonably safe (as long as we can guarantee that the lock won't otherwise be refreshed).
+ *
+ * Concurrency: We want to guarantee that a token T that is enqueued is included in some call to unlockOutstanding.
+ * If T can pass the compareAndSet, then T itself is scheduled. If T does not, that means there is some other
+ * thread that has scheduled the task, but the task has not retrieved the reference to the set of tokens to be
+ * unlocked (because it re-sets unlockIsScheduled to true first, before extracting the reference to the set).
```
sandorw: re-sets unlockIsScheduled to false

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -39,59 +36,59 @@
...
         this.scheduledExecutorService = scheduledExecutorService;
-        this.outstandingLockTokens = outstandingLockTokens;
-        schedulePeriodicKickJob();
     }
 
     /**
      * Adds all provided lock tokens to a queue to eventually be scheduled for unlocking.
      * Locks in the queue are unlocked asynchronously, and users must not depend on these locks being unlocked /
-     * available for other users immediately.
+     * unlockIsScheduled for other users immediately.
```
sandorw: I think this was an over-zealous find-replace :)

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -102,14 +99,20 @@ private void unlockOutstanding() {
...
-                this::scheduleIfNoTaskRunning, 0, KICK_JOB_INTERVAL.getSeconds(), TimeUnit.SECONDS);
+    private Set<LockToken> getOutstandingLockTokenSnapshot() {
+        // Ensure that we acquire the lock for as short as possible (i.e. only 2 writes)
+        Set<LockToken> toUnlock;
+        Set<LockToken> newSet = Sets.newConcurrentHashSet();
+
+        readWriteLock.writeLock().lock();
+        try {
+            toUnlock = outstandingLockTokens;
+            outstandingLockTokens = newSet;
```
sandorw: Are we back to needing a `volatile` modifier on `outstandingLockTokens` since we're not using explicit `synchronized` blocks anymore? I'll try to dig in to understand this a bit better.

lock-api/src/test/java/com/palantir/lock/client/AsyncTimeLockUnlockerTest.java
jeremyk-91: Yep. I think when I was fixing checkstyle I decided to copy it from the first file I could find that had the right version 🕙 changed

jeremyk-91:
update: The tests are failing because I wrote bad tests, not because the synchronization is wrong!

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
jeremyk-91: Yep, I'm surprised this hit for `available` which isn't such a common word! Fixed.

lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
sandorw: Looks like this isn't necessary according to the `Lock` interface: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html. These locks must provide the same guarantees as the built-in locks.

jeremyk-91: Yep, I don't think we need the `volatile` modifier here. I'm pretty confident the lock gives us the required memory visibility semantics.

Gory details (there might be some errors here):
- A `ReentrantReadWriteLock` uses an inner-class implementation of [AbstractQueuedSynchronizer](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/AbstractQueuedSynchronizer.html) for its synchronization.
- Internally, this uses volatiles - the release and acquire methods (both exclusive and shared) call AQS methods that perform volatile reads and writes.
- Read-lock lock/unlock and write-lock lock/unlock methods call the AQS methods you'd expect them to call; crucially they are on the *same* synchronizer.
- Here, the volatile variable we're synchronizing on is inside the AbstractQueuedSynchronizer. When we lock the write lock, we are doing a volatile read of some variable that was written to when the last reader unlocked the read lock. So lock/unlock operations have the same semantics as a volatile read _and_ write.

Happy to talk about this in more detail!

jeremyk-91:
closing/reopening for CLA

jeremyk-91:
Merging - thanks @sandorw!

PR #3305, opened 2018-06-22
Better batching with timelock

j-baker:
Internally we were seeing that atm it takes about 2.5 seconds to delete
1k things. This is almost surely exclusively due to the overhead of
calling timelock once per cell.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3305)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -1783,7 +1783,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             InetSocketAddress host,
             Map<Cell, Long> maxTimestampExclusiveByCell,
             boolean deleteSentinels,
-            long rangeTombstoneCassandraTimestamp) {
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1757,18 +1757,33 @@ public void deleteRange(final TableReference tableRef, final RangeRequest range)
...
             boolean deleteSentinels) {
+        if (maxTimestampExclusiveByCell.isEmpty()) {
+            return;
+        }
+
         Map<InetSocketAddress, Map<Cell, Long>> keysByHost = HostPartitioner.partitionMapByHost(
                 clientPool, maxTimestampExclusiveByCell.entrySet());
+
+        // this is required by the interface of the CassandraMutationTimestampProvider, although it exists for tests
+        long maxTimestampForAllCells = maxTimestampExclusiveByCell.values().stream().mapToLong(x -> x).max().getAsLong();
```
jeremyk-91: This is definitely fine for now. Longer term we should either find a way to fix the tests to kill off the legacy usage, or at least make the provider take a LongSupplier that it can call or not, since it's a bit sad that the result of this computation isn't actually used

PR #3307, opened 2018-06-22
FFT Part 4: Track post-commit unlocking and factor this into nonPutOverhead metrics

jeremyk-91:
**Goals (and why)**: Instrument a call that is part of tx overhead that was missed

- Instrument post-commit unlocks

- None in particular, though more broadly: is there anything else I missed?

**Priority (whenever / two weeks / yesterday)**: today, should be really small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3307)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1399,7 +1399,8 @@ private void commitWrites(TransactionService transactionService) {
                         .build();
                     });
         } finally {
-            timelockService.unlock(ImmutableSet.of(commitLocksToken));
+            runAndGetDurationMicros(() -> timelockService.unlock(ImmutableSet.of(commitLocksToken)),
```
gsheasby: From what I can see, this returns a long... which you're never using, reporting, logging or otherwise storing. I think you need to move the massive log message into the finally block, too.

jeremyk-91: I think we only want to do the big logging in the event of a successful tx though. Might need some nontrivial refactoring to get this right!

Actually `runAndGetDurationMicros` does have side effects that report the time to a Dropwizard metric timer, which is what I was going for. I'll... rename that as well.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,11 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
```
gsheasby: nit: do we still add `|metrics|`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1341,80 +1343,56 @@ private void commitWrites(TransactionService transactionService) {
...
-        long nonPutOverhead = microsForCommitStage - microsForWrites - microsWritingToTargetedSweepQueue;
-        getTimer("nonPutOverhead").update(nonPutOverhead, TimeUnit.MICROSECONDS);
+            long microsForPostCommitUnlock = runAndReportTimeAndGetDurationMicros(
+                    () -> timelockService.unlock(ImmutableSet.of(commitLocksToken)), "postCommitUnlock");
 
-        // Dropwizard Metrics doesn't support histograms of double yet, so using longs as a workaround
-        getHistogram("nonPutOverheadMillionths").update(
-                Math.round((1_000_000. * nonPutOverhead) / microsForCommitStage));
+            // We only care about detailed profiling for successful transactions
+            optionalProfile.ifPresent(profile -> profileProcessor.consumeProfilingData(
```
gsheasby: This part appeared slightly sketchy (you get the profile and a few random extra bits), but it's ok, especially as the name (TransactionCommitProfile) is accurate and descriptive.

gsheasby:
A bunch of tests fail because of:
```
java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Cannot build TransactionCommitProfile, some of required attributes are not set [totalCommitStageMicros, totalTimeSinceTransactionCreationMicros]
```

docs/source/release_notes/release-notes.rst
jeremyk-91: Yeah, we do. Forgot about that - added

Issue #3297, opened 2018-06-21
AtlasDB does not actually use baseline checkstyle


gsheasby:
Filing for tracking.

Intellij's configured checkstyle (Baseline checkstyle) does not match `./gradlew check`. Instead, the gradle function runs a very restricted set of rules that pre-dates baseline.

There are ~2300 errors according to baseline 0.18, and ~11,500 according to baseline 0.20. For 0.18, most of the errors are confined to a few projects with a lot of legacy code that we didn't get to in the baseline bash last year (see #334).

- [x] Step 1 (PR #3292) - Revert back to baseline 0.18
- [x] Step 2 (PR #3301) - Switch gradle to actually use baseline, and fix or suppress the ~170 errors that will appear
- [x] Step 2.25 (PR #3308) - Fix the 80 or so errors in impl-shared
- [ ] Step 2.5 (optional) - Reduce the number of suppressed projects
- [x] Step 3 (forked in PR #3300) - Upgrade to baseline 0.20 (potentially involves changing copyright and import order of every file)

Suppressed projects and their errors (for baseline 0.18)
- [ ] atlasdb-client | 379 |
- [x] atlasdb-client-protobufs | 0 | Generated code
- [ ] atlasdb-commons | 540 |  
- [ ] atlasdb-console | 90 |  
- [x] atlasdb-dbkvs-hikari | 41 |  #3380
- [x] atlasdb-jdbc | 124 | Deprecated
- [x] atlasdb-jdbc-tests | 1 | Deprecated
- [x] atlasdb-service-server | 0 | Deleted!
- [ ] cassandra-partitioner | 62 | 
- [ ] commons-db | 618 |  
- [x] commons-executors | 17 |  #3341
- [ ] leader-election-impl | 138 |  
- [ ] lock-impl | 159 |  
- [x] profile-client-protobufs | 0 |  Generated code
- [x] timestamp-impl | 22 |  #3341

gmaretic:
I think we did as much as is reasonable here, also we are on 0.30.0 now: https://github.com/palantir/atlasdb/blob/c4845828b3f9c28f79a7cafacdf784069e5efac3/build.gradle#L18

PR #3308, opened 2018-06-22
Fix/baseline impl shared

gsheasby:
**Goals (and why)**: Continue to address #3297. Resume baseline checks on atlasdb-impl-shared, since we had them enabled before.

**Implementation Description (bullets)**: Checkstyle!

**Concerns (what feedback would you like?)**: I suppressed some things, particularly AvoidStaticImport - allowing these for all tests, because Mockito seems to have sprouted a bunch of helper classes that we sometimes use, and we have our own constants (e.g. in SweepQueueUtils) that are used in a lot of places in a handful of files.

**Where should we start reviewing?**: Commit by commit?

**Priority (whenever / two weeks / yesterday)**: Today

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3308)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
@@ -118,8 +118,9 @@ public class ShardProgress {
     private long increaseValueFromToAtLeast(ShardAndStrategy shardAndStrategy, long oldVal, long newVal) {
         byte[] colValNew = SweepShardProgressTable.Value.of(newVal).persistValue();
 
-        while (oldVal < newVal) {
-            CheckAndSetRequest casRequest = createRequest(shardAndStrategy, oldVal, colValNew);
+        long currentValue = oldVal;
+        while (currentValue < newVal) {
+            CheckAndSetRequest casRequest = createRequest(shardAndStrategy, currentValue, colValNew);
             try {
                 kvs.checkAndSet(casRequest);
                 return newVal;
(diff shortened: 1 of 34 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
```diff
@@ -128,10 +129,10 @@ private long increaseValueFromToAtLeast(ShardAndStrategy shardAndStrategy, long
                         + "value changed under us.",
                         SafeArg.of("old value", oldVal),
                         SafeArg.of("new value", newVal));
-                oldVal = updateOrRethrowIfNoChange(shardAndStrategy, oldVal, e);
```
gmaretic: awkward

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -85,7 +85,7 @@ private TargetedSweeper(MetricsManager metricsManager, Supplier<Boolean> runSwee
      * @param conservativeThreads number of conservative threads to use for background targeted sweep.
      * @param thoroughThreads number of thorough threads to use for background targeted sweep.
      * @param followers follower used for sweeps, as defined by your schema.
-     * @return returns an uninitialized targeted sweeper.
+     * @return returns an uninitialized targeted sweeper
```
gmaretic: Never understood why sentences should not end with a full stop. 😕

PR #3310, opened 2018-06-22
[TS]Have Targeted Sweep Use Timelock Locks for Synchronization

gmaretic:
**Goals (and why)**:
Currently all nodes of a single service run targeted sweep independently, which can manifest itself in doing all the work twice. This avoids that by having background threads take out timelock locks on the shard and strategy they are about to sweep.

Relatively simple -- when a thread wants to sweep, it tries to find a shard that is free, internally cycling using an AtomicLong to avoid unnecessary clashes and ensure a uniform distribution. If it manages to get a lock, it sweeps and then releases; otherwise it gives up after trying n times, where n is the number of shards, and just waits until next iteration.

Do we need more testing?

Small

Would be good to get it in ASAP as it should speed things up by a factor of the number of nodes we use.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3310)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperLock.java
```diff
@@ -0,0 +1,62 @@
...
+    public boolean isHeld() {
+        return lockRefreshToken.isPresent();
+    }
+
+    public ShardAndStrategy getShardAndStrategy() {
+        return shardStrategy;
+    }
+
+    public void unlock() {
+        lockRefreshToken.ifPresent(lockService::unlock);
```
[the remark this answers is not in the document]
gmaretic: We are discarding the object after this call, but makes sense to do it this way to be consistent with the expected behaviour.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -103,23 +115,17 @@ public void initializingWithUninitializedKvsThrows() {
...
-                .createUninitialized(MetricsManagers.createForTests(), null, null, DEFAULT_SHARDS + 5, 0, ImmutableList.of());
-        sweeperConservative.initialize(timestampsSupplier, spiedKvs, mock(TargetedSweepFollower.class));
-        assertThat(progress.getNumberOfShards()).isEqualTo(DEFAULT_SHARDS + 5);
-
-        TargetedSweeper sweeperThorough = TargetedSweeper
-                .createUninitialized(MetricsManagers.createForTests(), null, null, 0, DEFAULT_SHARDS + 10, ImmutableList.of());
-        sweeperThorough.initialize(timestampsSupplier, spiedKvs, mock(TargetedSweepFollower.class));
-        assertThat(progress.getNumberOfShards()).isEqualTo(DEFAULT_SHARDS + 10);
+    public void enqueueUpdatesNumberOfShards() {
+        assertThat(DEFAULT_SWEEP_QUEUE_SHARDS).isLessThan(DEFAULT_SHARDS);
```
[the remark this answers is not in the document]
gmaretic: I fully qualified the one that's the default config, so should be clearer.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -42,17 +41,17 @@
...
 @SuppressWarnings({"FinalClass", "Not final for mocking in tests"})
 public class TargetedSweeper implements MultiTableSweepQueueWriter {
     private static final Logger log = LoggerFactory.getLogger(TargetedSweeper.class);
     private final Supplier<Boolean> runSweep;
     private final Supplier<Integer> shardsConfig;
-    private final int minShards;
     private final List<Follower> followers;
-
     private final MetricsManager metricsManager;
+    private LockService lockService;
```
j-baker: let's put this with the other private fields, not the final ones

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperLock.java
```diff
@@ -0,0 +1,65 @@
...
+    private final LockService lockService;
+    private Optional<LockRefreshToken> lockRefreshToken;
+
+    private TargetedSweeperLock(ShardAndStrategy shardStrategy, LockService lockService, LockRefreshToken token) {
+        this.shardStrategy = shardStrategy;
+        this.lockService = lockService;
+        this.lockRefreshToken = Optional.ofNullable(token);
+    }
+
+    public static TargetedSweeperLock acquire(int shard, TableMetadataPersistence.SweepStrategy strategy,
```
j-baker: I'd recommend you call this tryAcquire and have it return optional if you could not lock. At the moment it's weird to get back a targeted sweeper lock that may or may not be valid.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -193,31 +189,39 @@ private void scheduleBackgroundThreads() {
...
+        private Optional<TargetedSweeperLock> tryToAcquireLockForNextShardAndStrategy() throws InterruptedException {
+            for (int i = 0; i < queue.getNumShards(); i++) {
+                int nextShardCandidate = getShardAndIncrement();
+                TargetedSweeperLock lock = TargetedSweeperLock.acquire(nextShardCandidate, sweepStrategy, lockService);
+                if (lock.isHeld()) {
+                    return Optional.of(lock);
+                }
             }
-            return nextShardCandidate;
+            log.info("There are no candidates for this iteration of targeted sweep as all shards for sweep strategy {} "
```
j-baker: this isn't info, it'll get logged like all the time

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -193,31 +189,39 @@ private void scheduleBackgroundThreads() {
...
+                maybeLock.ifPresent(TargetedSweeperLock::unlock);
             }
         }
 
-        private int lockNextShardToSweep() {
-            int nextShardCandidate = getShardAndIncrement();
-            while (!shardsBeingSwept.add(nextShardCandidate)) {
-                nextShardCandidate = getShardAndIncrement();
+        private Optional<TargetedSweeperLock> tryToAcquireLockForNextShardAndStrategy() throws InterruptedException {
+            for (int i = 0; i < queue.getNumShards(); i++) {
```
j-baker: if you wanna go fluent

```java
return IntStream.range(0, queue.getNumShards)
    .map(this::getShardAndIncrement)
    .mapToObj(shard -> TargetedSweeperLock.tryAcquire(nextSharedCandidate, sweepStrategy, lockService))
    .flatMap(Streams::stream)
    .findFirst();
```

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -775,6 +779,51 @@ public void deletesGetBatched() {
...
+    public void extraSweepersGiveUpAfterFailingToAcquireEnoughTimes() throws InterruptedException {
+        int shards = 8;
+        int sweepers = 2;
+        int threads = shards / sweepers + 1;
+        LockService stickyLockService = createStickyLockService();
+        createAndInitializeSweepersAndWaitForOneBackgroundIteration(sweepers, shards, threads, stickyLockService);
+
+        // We issue between 2 * shards - shards/sweepers and 2 * shards - 1 lock requests to acquire all locks, and then
+        // an additional 2 * shards requests by the unlucky threads that have been unable to get a lock
+        ArgumentCaptor<LockRequest> captor = ArgumentCaptor.forClass(LockRequest.class);
```
jeremyk-91: I'm not sure these bounds are correct generally, assuming you take `sweepers` as a variable (they are correct in this instance where `sweepers = 2`, but in that case I'd expect to see `2 * shards - shards / 2`).

More generally I get, where `shards / sweepers` is an integer, a minimum of `shards * (sweepers + 1) / 2` and a maximum of `shards * sweepers - (sweepers * (sweepers - 1)) / 2`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -775,6 +779,51 @@ public void deletesGetBatched() {
...
+        verify(stickyLockService, atLeast(2 * shards - shards / sweepers + 2 * shards))
```
jeremyk-91: May be worth extracting something or a comment for why there's the hanging `2 * shards` here

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
gsheasby: @gmaretic I saw this test flake on circle with message "Wanted at most 314 times but was 322"

gsheasby: Internal build #11157.

PR #3311, opened 2018-06-22
Moved CancelableServerCall annotation to commons-annotations

cakofony:
from atlasdb-commons

**Goals (and why)**:

It's an annotation. Annotations go in commons-annotations.
Not a break because atlasdb-commons depends on commons-annotations

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3311)

### DIFF
--- a/commons-annotations/src/main/java/com/palantir/annotations/remoting/CancelableServerCall.java
+++ b/commons-annotations/src/main/java/com/palantir/annotations/remoting/CancelableServerCall.java
@@ -21,7 +21,7 @@ import java.lang.annotation.RetentionPolicy;
 
 /**
  * Marks a server call cancelable - the server call can now throw interrupted exceptions on the client,
- * and interrupting on the client will propagate to the server
+ * and interrupting on the client will propagate to the server.
  *
  * @author kbrainard
  */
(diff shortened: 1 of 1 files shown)


docs/source/release_notes/release-notes.rst
```diff
@@ -310,6 +310,9 @@ v0.89.0
          - Increased PTExecutors default thread timeout from 100 milliseconds to 5 seconds to avoid recreating threads unnecessarily.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3208>`__)
 
+    *    - |improved|
```
gsheasby: We don't need a release note here - getting around the check with `[no release notes]` is better - especially since this will go out in 0.93.0, not 0.89.0.

cakofony: updated

gsheasby: Thanks - I was also updating it (and fixing checkstyle, since checkstyle rules apply in commons-annotations and not in atlasdb-commons).

By the way, please apply fixes in new commits rather than amending and force-pushing.  We squash commits on merge anyway. Here the change is trivial, but for substantial changes it gets a real pain to CR in Github if you amend (it doesn't have patch sets like some internally used CR tools).

PR #3314, opened 2018-06-25
log only if host is really removed from blacklist

gokcanozakdag:
**Goals (and why)**:
Fix the bug where we are logging bunch of "added back host.." logs, as we didn't check if we really removed the host from blacklist.

Check if host is blacklisted first.

This week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3314)


gokcanozakdag:
Does it worth adding a changelog?

PR #3318, opened 2018-06-26
Fix flaky test

gmaretic:
**Goals (and why)**:
There is a tricky corner case that results in more requests being made than anticipated.

Increase the required upper bound so that if the test fails, we falsify TargetedSweeperLock's guarantees.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3318)

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
@@ -816,7 +816,7 @@ public class TargetedSweeperTest extends AbstractSweepQueueTest {
         // threads + ... + threads * (shards / threads) + shards * (threads * sweepers - shards)
         verify(stickyLockService, atLeast(shards * (shards / threads + 1) / 2 + shards * (threads * sweepers - shards)))
                 .lock(any(LockRequest.class));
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -816,10 +816,11 @@ public void extraSweepersGiveUpAfterFailingToAcquireEnoughTimes() throws Interru
...
         verify(stickyLockService, atLeast(shards * (shards / threads + 1) / 2 + shards * (threads * sweepers - shards)))
                 .lock(any(LockRequest.class));
-        // maximum: as in the example above, but we have extra threads
+        // maximum: one would think that it is
         // shards + shards - 1 + ... + shards - (sweepers - 1) + shards * (threads * sweepers - shards)
-        verify(stickyLockService, atMost(shards * ((threads + 1) * sweepers - shards) - sweepers * (sweepers - 1) / 2))
-                .lock(captor.capture());
+        // but actually the logic is much more complicated since threads from the same sweeper can loop back and hit a
+        // race condition with each other, so we go with the more conservative upper bound
+        verify(stickyLockService, atMost(threads * sweepers * shards)).lock(captor.capture());
```
jeremyk-91: very conservative, but this is fine!

PR #3319, opened 2018-06-26
FFT - getRowColumnRange Merge, Part 2: SortedMap -> List + Sort

jeremyk-91:
**Goals (and why)**:
- Improve the performance of the post-filtering stage of `getRowsColumnRange`. During this process, we accumulate multiple lists of entries that are known to be in sorted order as we try to read the newest version, then the next newest if that hasn't committed, etc. We need to merge them to produce a final, fully sorted list.

**UPDATE Opted to go for @j-baker's `ImmutableSortedMap` solution as it's cleaner and gets us 75-80% of the performance gain without hurting readability of SnapshotTransaction.**
- ~Replace the use of a `TreeSet` with a plain `ArrayList` and sorting at the end. It makes sense that `TreeSet` is slow as it reads a mostly sorted list of elements and has to rebalance its internal structure a lot.~
- ~Create a `PostFilteringPopulator` which processes elements. Crucially, the method calls are only invoked *once* on. There is a much easier implementation with a `BiConsumer<Cell, Value>`, but benchmarks found that function calls when done on a per-cell basis can actually add a nontrivial overhead - in JMH benchmarks this was deemed to be about 2 and 4 percent respectively when dealing with reads with 100 and 50,000 columns.~
- Alternative solutions benchmarked were performing an N-way merge (linear time, but more complex and not clearly better in tests, using a BiConsumer to process output using a sorted list, and using an ImmutableSortedMap's builder which waits to the end to sort. All units were in microseconds. Tables read had one row with Size columns; a randomly selected %Aborted of these were aborted. I used the in-memory KVS to remove uncertainty from Docker/Cassandra performance overheads (this is also why the perf improvements in the release notes are only ~10 percent, not the ~40 percent this could suggest)

| Size  | %Aborted |   TreeSet   |  List+Sort  | N-way Merge | BiConsumer | Imm.SortMap
|-------|----------|-------------|-------------|-------------|-----------------|----|
|   100 |        0 | 94 ± 8      | 79 ± 6      | 78 ± 6      | 81 ± 5          | 80 ± 5 |
|   100 |        2 | 96 ± 6      | 86 ± 6      | 87 ± 6      | 87 ± 5          | 86 ± 5 |
|   100 |       20 | 95 ± 6      | 88 ± 6      | 87 ± 7      | 87 ± 6          | 88 ± 6 |
| 50000 |        0 | 34325 ± 115 | 18936 ± 68  | 19050 ± 75  | 19501 ± 84      | 20666 ± 82      |
| 50000 |        2 | 37316 ± 143 | 24858 ± 100 | 24837 ± 108 | 25477 ± 101     | 26855 ± 127      |
| 50000 |       20 | 37751 ± 171 | 24988 ± 100 | 24581 ± 101 | 25473 ± 102     | 27001 ± 286      |

- ~The inner classes in SnapshotTransaction feature egregious and reprehensible duplication, though I can't really think of a better way to do it offhand. Note earlier comments on benchmarking as this is a very hot code-path.~
- This isn't the fastest solution, is this the right tradeoff?
- I didn't add tests as this PR should not change behaviour. The critical test is `SnapshotTransactionTest#getRowsColumnRangesReturnsInOrderInCaseOfAbortedTxns` which I've confirmed fails if you remove the list-sort line.

**How can we see a perf improvement?**:
By JMH benchmarks (to come in Part 2), and general improvements in `runTaskWithRetry` for heavy users of dynamic column queries.
Unfortunately we don't quite have metrics for the performance of `getRowColumnRange` yet.

**Where should we start reviewing?**: `SnapshotTransaction.java`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3319)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -31,6 +31,7 @@ import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
@@ -1048,10 +1049,13 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
         }
 
         Map<Cell, Value> remainingResultsToPostfilter = rawResults;
+        AtomicInteger resultCount = new AtomicInteger();
         while (!remainingResultsToPostfilter.isEmpty()) {
             remainingResultsToPostfilter = getWithPostFilteringInternal(
-                    tableRef, remainingResultsToPostfilter, results, transformer);
+                    tableRef, remainingResultsToPostfilter, results, resultCount, transformer);
         }
+
+        getMeter(AtlasDbMetricNames.SNAPSHOT_TRANSACTION_CELLS_RETURNED).mark(resultCount.get());
     }
 
     /**
@@ -1061,6 +1065,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
     private <T> Map<Cell, Value> getWithPostFilteringInternal(TableReference tableRef,
             Map<Cell, Value> rawResults,
             @Output ImmutableMap.Builder<Cell, T> results,
+            @Output AtomicInteger count,
             Function<Value, T> transformer) {
         Set<Long> startTimestampsForValues = getStartTimestampsForValues(rawResults.values());
         Map<Long, Long> commitTimestamps = getCommitTimestamps(tableRef, startTimestampsForValues, true);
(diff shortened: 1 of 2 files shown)


j-baker:
hey Jeremy,
It'd be cool if you could check out https://github.com/palantir/atlasdb/compare/jbaker/no_txns_table_hotspot?expand=1 - I use ImmutableMap/ImmutableSortedMap to achieve the same effect, achieving most of the change with basically zero code additions overall.

Difference in perf is something like:

Baseline: 850 ns/cell
James: 703 ns/cell
Jeremy: 655ns/cell

so you basically get rid of almost all of the difference without the code change.

j-baker:
basically.... I am not convinced that this is a sufficiently hot code path in Atlas that we should compromise on code readability :)

j-baker:
because e.g. the benefit at this level is roughly 50 nanoseconds per cell, which means that a single CPU core can do 20 million reads a second, which seems good enough :)

jeremyk-91:
Yep, @sandorw let me know about that. Agree that most of the win is already obtained by killing the treeset rebalancing. Might have gone down a bit of a rabbit hole trying to optimize this too aggressively (I even had some further perf improvements in mind)

I'll discuss with the team tomorrow. If we're not going for perf-at-all-costs then there certainly are much more readable approaches!

jeremyk-91:
Offline discussion - let's go ahead with the smaller change for now then.

PR #3320, opened 2018-06-26
FFT - getRowColumnRange Merge, Part 1: JMH Benchmarks

jeremyk-91:
**Goals (and why)**:
- Have benchmarks that cover the post-filtering stage of `getRowColumnRange`

- Add a state table that is set up with one row, where some columns are uncommitted, and other columns are committed but have uncommitted values written on top of them. The table is parameterised in number of columns that are purely committed, committed with uncommitted values, and uncommitted.
- Add benchmarks that read all columns from tables with various mixtures of these parameters and check that the number of visible columns (purely committed + committed with uncommitted values) is correct.

- Is the scrambling of the rows that were uncommitted something reasonable to do?
- This adds a good eight minutes to the benchmarks. Is that okay?

**Where should we start reviewing?**: `WideRowTableWithAbortedValues.java`

**Priority (whenever / two weeks / yesterday)**: this week. Ideally before #3319

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3320)

### DIFF
--- a/atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/TransactionGetRowsColumnRangeBenchmarks.java
+++ b/atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/TransactionGetRowsColumnRangeBenchmarks.java
@@ -46,50 +46,50 @@ public class TransactionGetRowsColumnRangeBenchmarks {
     @Threads(1)
     @Warmup(time = 16, timeUnit = TimeUnit.SECONDS)
     @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
-    public Object getAllColumnsSingleBigRow(VeryWideRowTable table, Blackhole blackhole) {
-        return getAllRowsAndAssert(table, blackhole,
+    public void getAllColumnsSingleBigRow(VeryWideRowTable table, Blackhole blackhole) {
+        getAllRowsAndAssert(table, blackhole,
                 count -> Preconditions.checkState(count == table.getNumCols(),
                         "Should be %s columns, but was: %s", table.getNumCols(), count));
     }
 
     @Benchmark
     @Threads(1)
-    @Warmup(time = 16, timeUnit = TimeUnit.SECONDS)
-    @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
-    public Object getAllColumnsModeratelyWideRow(
+    @Warmup(time = 5, timeUnit = TimeUnit.SECONDS)
+    @Measurement(time = 20, timeUnit = TimeUnit.SECONDS)
+    public void getAllColumnsModeratelyWideRow(
             ModeratelyWideRowTable table,
             Blackhole blackhole) {
-        return getAllRowsAndAssert(table, blackhole,
+        getAllRowsAndAssert(table, blackhole,
                 count -> Preconditions.checkState(count == table.getNumCols(),
                         "Should be %s columns, but was: %s", table.getNumCols(), count));
     }
 
     @Benchmark
     @Threads(1)
-    @Warmup(time = 16, timeUnit = TimeUnit.SECONDS)
-    @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
-    public Object getAllColumnsModeratelyWideRowWithSomeUncommitted(
+    @Warmup(time = 5, timeUnit = TimeUnit.SECONDS)
+    @Measurement(time = 20, timeUnit = TimeUnit.SECONDS)
+    public void getAllColumnsModeratelyWideRowWithSomeUncommitted(
             CleanModeratelyWideRowTable table,
             Blackhole blackhole) {
-        return getAllRowsAndAssert(table, blackhole,
+        getAllRowsAndAssert(table, blackhole,
                 count -> Preconditions.checkState(count == table.getNumReadableCols(),
                         "Should be %s columns, but was: %s", table.getNumReadableCols(), count));
     }
 
     @Benchmark
     @Threads(1)
-    @Warmup(time = 16, timeUnit = TimeUnit.SECONDS)
-    @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
-    public Object getAllColumnsModeratelyWideRowWithManyUncommitted(
+    @Warmup(time = 5, timeUnit = TimeUnit.SECONDS)
+    @Measurement(time = 20, timeUnit = TimeUnit.SECONDS)
+    public void getAllColumnsModeratelyWideRowWithManyUncommitted(
             DirtyModeratelyWideRowTable table,
             Blackhole blackhole) {
-        return getAllRowsAndAssert(table, blackhole,
+        getAllRowsAndAssert(table, blackhole,
                 count -> Preconditions.checkState(count == table.getNumReadableCols(),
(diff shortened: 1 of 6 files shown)


j-baker:
I don't think it's reasonable to merge these benchmarks as is. Basically the amount of time you're benchmarking for is unreasonably high; if you use e.g. the inmemory kvs, you can run 30 iterations of the benchmark a second, which means that you're doing overall (for your 160 second per iteration of measurement (JMH will run each test like 20 times)) you're doing 250 million reads to build this number.

I'd expect we can get something reasonable out of a much smaller sample size.

jeremyk-91:
Fair enough to cut down the benchmarking length. In Atlas at least though I'm pretty sure the benchmarks are only run for one iteration (@gmaretic would know better regarding the actual setup, but when I run the CLI).

What do you think would be a reasonable number of reads?

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/TransactionGetRowsColumnRangeBenchmarks.java
```diff
@@ -42,6 +47,48 @@
...
+    @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
+    public Object getAllColumnsModeratelyWideRowWithManyUncommitted(
+            DirtyModeratelyWideRowTable table,
+            Blackhole blackhole) {
+        return getAllRowsAndAssert(table, blackhole,
+                count -> Preconditions.checkState(count == table.getNumReadableCols(),
+                        "Should be %s columns, but was: %s", table.getNumReadableCols(), count));
+    }
+
+    private Object getAllRowsAndAssert(WideRowTable table, Blackhole blackhole, Consumer<Integer> assertion) {
```
gmaretic: You don't actually accept the count anywhere

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/TransactionGetRowsColumnRangeBenchmarks.java
```diff
@@ -42,6 +47,48 @@
...
+    }
+
+    @Benchmark
+    @Threads(1)
+    @Warmup(time = 16, timeUnit = TimeUnit.SECONDS)
+    @Measurement(time = 160, timeUnit = TimeUnit.SECONDS)
+    public Object getAllColumnsModeratelyWideRowWithManyUncommitted(
+            DirtyModeratelyWideRowTable table,
+            Blackhole blackhole) {
+        return getAllRowsAndAssert(table, blackhole,
```
gmaretic: You don't need to return anything here as you're already consuming all the columns with the blackhole, no?

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/WideRowTableWithAbortedValues.java
```diff
@@ -0,0 +1,135 @@
...
+ * columns which have committed versions, but have newer uncommitted versions on top.
+ */
+public abstract class WideRowTableWithAbortedValues extends WideRowTable {
+    public static final byte[] DUMMY_VALUE = PtBytes.toBytes("dummy");
+
+    public abstract int getNumColsCommitted();
+    public abstract int getNumColsCommittedAndNewerUncommitted();
+    public abstract int getNumColsUncommitted();
+
+    public abstract int getNumUncommittedValues();
```
gmaretic: I'd add a bit more explanation that we write into the cells that have uncommitted writes this many times

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/WideRowTableWithAbortedValues.java
```diff
@@ -0,0 +1,135 @@
...
+        for (Cell cell : Iterables.concat(committedWithNewerUncommitted, uncommitted)) {
+            values.put(cell, DUMMY_VALUE);
+        }
+
+        // Simulate getting a timestamp, writing the values, but not putting into the tx table
+        long freshTimestamp = services.getTransactionManager().getTimestampService().getFreshTimestamp();
+        services.getKeyValueService().multiPut(ImmutableMap.of(tableRef, values), freshTimestamp);
+    }
+
+    private List<Cell> getCells(int numCells, CellType cellType) {
```
gmaretic: nice!

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/WideRowTableWithAbortedValues.java
```diff
@@ -0,0 +1,135 @@
...
+                .map(index -> cell(index, cellType))
+                .collect(Collectors.toList());
+    }
+
+    private Cell cell(int index, CellType cellType) {
+        return Cell.create(Tables.ROW_BYTES.array(), getColumnName(index, cellType));
+    }
+
+    private byte[] getColumnName(int index, CellType cellType) {
+        // Prepend a hash of the column name, to ensure an even distribution of the various cell types.
```
gmaretic: 👍

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/TransactionGetRowsColumnRangeBenchmarks.java
jeremyk-91: Oops! Good catch. Fixed!

jeremyk-91:
Reducing the benchmarks to 20 seconds each (which gives counts of about 100x). Using Cassandra KVS I get something like

```
Benchmark                                                                                                                                                                                 (uri)    Mode  Cnt        Score       Error  Units
TransactionGetRowsColumnRangeBenchmarks.getAllColumnsModeratelyWideRow                     com.palantir.atlasdb.performance.backend.CassandraKeyValueServiceInstrumentation@192.168.99.100:9160  sample  118   169511.641 ±  3494.785  us/op
TransactionGetRowsColumnRangeBenchmarks.getAllColumnsModeratelyWideRowWithManyUncommitted  com.palantir.atlasdb.performance.backend.CassandraKeyValueServiceInstrumentation@192.168.99.100:9160  sample  102   196913.835 ±  6629.162  us/op
TransactionGetRowsColumnRangeBenchmarks.getAllColumnsModeratelyWideRowWithSomeUncommitted  com.palantir.atlasdb.performance.backend.CassandraKeyValueServiceInstrumentation@192.168.99.100:9160  sample  104   192872.448 ±  4598.556  us/op
```

gmaretic:
You could probably go even lower, as it should be faster on a dedicated machine and that's what we care about, but 20 seconds is reasonable

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/WideRowTableWithAbortedValues.java
jeremyk-91: I've added some javadoc that should hopefully make things a lot clearer, and renamed this to `getNumUncommittedValuesPerCell` as well.

Issue #3322, opened 2018-06-26
Investigate why targeted sweep reads so many entries


gmaretic:
Observed that we have written 6M writes, when we sweep them we put 6M tombstones, but we read 40M entries from the sweep queue.

Hypothesis is that this is because we repeatedly reread a batch of entries, but cannot sweep any of them away because the first one is committed after the sweep timestamp. This can persist for multiple iterations until the sweep timestamp progresses far enough to unblock the sweeps.

If this is the case, we could give up early since we should have the commit timestamp cached from the previous iteration and we can just skip reading further until we get unblocked

PR #3325, opened 2018-06-27
[TS] Immediately stop reading from sweep queue if we know we should pause

gmaretic:
**Goals (and why)**:
This should address https://github.com/palantir/atlasdb/issues/3322. We do not want to read an entire batch of entries if we know early on we will have to pause anyway.

For each start timestamp check if we happen to have the commit timestamp cached (which we will if we have recently swept from the same node and had to pause here). If yes, and it is greater than the sweep timestamp, we know everything after will be filtered out anyway, so immediately stop reading.

Should there be a more detailed explanation in the code?

Small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3325)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
@@ -140,7 +141,7 @@ public class SweepableCells extends KvsSweepQueueWriter {
         SweepableCellsTable.SweepableCellsRow row = computeRow(partitionFine, shardStrategy);
         RowColumnRangeIterator resultIterator = getRowColumnRange(row, partitionFine, minTsExclusive, sweepTs);
         PeekingIterator<Map.Entry<Cell, Value>> peekingResultIterator = Iterators.peekingIterator(resultIterator);
-        Multimap<Long, WriteInfo> writesByStartTs = getBatchOfWrites(row, peekingResultIterator);
+        Multimap<Long, WriteInfo> writesByStartTs = getBatchOfWrites(row, peekingResultIterator, sweepTs);
         maybeMetrics.ifPresent(metrics -> metrics.updateEntriesRead(shardStrategy, writesByStartTs.size()));
         log.info("Read {} entries from the sweep queue.", SafeArg.of("number", writesByStartTs.size()));
         TimestampsToSweep tsToSweep = getTimestampsToSweepDescendingAndCleanupAborted(shardStrategy,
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -151,12 +152,17 @@ SweepBatch getBatchForPartition(ShardAndStrategy shardStrategy, long partitionFi
...
-            PeekingIterator<Map.Entry<Cell, Value>> resultIterator) {
+            PeekingIterator<Map.Entry<Cell, Value>> resultIterator, long sweepTs) {
         Multimap<Long, WriteInfo> writesByStartTs = HashMultimap.create();
         while (resultIterator.hasNext() && writesByStartTs.size() < SweepQueueUtils.SWEEP_BATCH_SIZE) {
             Map.Entry<Cell, Value> entry = resultIterator.next();
             SweepableCellsTable.SweepableCellsColumn col = computeColumn(entry);
-            writesByStartTs.putAll(getTimestamp(row, col), getWrites(row, col, entry.getValue()));
+            long startTs = getTimestamp(row, col);
+            if (knownToBeCommittedAfterSweepTs(startTs, sweepTs)) {
+                writesByStartTs.put(startTs, getWriteInfo(startTs, entry.getValue()));
```
jeremyk-91: You've mentioned this in the PR description - but could be worth explaining this. Something like maybe `In this case we will stop here in getTimestampsToSweepDescendingAndCleanupAborted() anyway, so there's no point getting more values`.

An uninitiated but astute reader may wonder why you're doing this if you have something like (write at 80, commit at 105) and (write at 83, commit at 96) on different cells in the same shard, with an immutable timestamp of 100.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableCells.java
```diff
@@ -281,6 +287,11 @@ private long getTimestamp(SweepableCellsTable.SweepableCellsRow row, SweepableCe
         return row.getTimestampPartition() * SweepQueueUtils.TS_FINE_GRANULARITY + col.getTimestampModulus();
     }
 
+    private boolean knownToBeCommittedAfterSweepTs(long startTs, long sweepTs) {
+        Optional<Long> commitTsIfCached = commitTsCache.loadIfCached(startTs);
+        return commitTsIfCached.isPresent() && commitTsIfCached.get() >= sweepTs;
```
jeremyk-91: nit: this is equivalent to `commitTsIfCached.map(x -> x >= sweepTs).orElse(false)` I think (and you can make the whole thing fluent if you want from there!)

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CommitTsCacheTest.java
```diff
@@ -148,6 +150,34 @@ public void onlyRequestNonCachedTimestamps() throws Exception {
...
+        verifyNoMoreInteractions(mockTransactionService);
+    }
+
+    @Test
+    public void loadIfCachedDoesNotAbortTransactionsAndCorrectlyGetsAbortedTransactions() {
+        when(mockTransactionService.get(anyLong())).thenReturn(null);
+        assertThat(loader.loadIfCached(VALID_START_TIMESTAMP)).isEmpty();
+        assertThat(loader.load(VALID_START_TIMESTAMP)).isEqualTo(ROLLBACK_TIMESTAMP);
+        assertThat(loader.loadIfCached(VALID_START_TIMESTAMP)).contains(ROLLBACK_TIMESTAMP);
+    }
```
jeremyk-91: Good set 👍

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -704,6 +705,68 @@ public void doNotDeleteAnythingAfterEntryWithCommitTsAfterSweepTs() {
...
+        // we read a reference to bad entries and give up
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(otherShard));
+        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(writesInDedicated + 1);
+
+        immutableTs = 250;
+
+        // we now read all to the end
+        sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(otherShard));
+        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(writesInDedicated + 1 + writesInOther);
+    }
```
jeremyk-91: Nice tests!

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -1011,4 +1075,24 @@ private void createAndInitializeSweepersAndWaitForOneBackgroundIteration(int swe
...
+    private Map<Integer, Integer> enqueueAtLeastThresholdWritesInDefaultShardWithStartTs(long threshold, long startTs) {
+        List<WriteInfo> writeInfos = new ArrayList<>();
+        int counter = 0;
+        while (writeInfos.stream().filter(write -> write.toShard(DEFAULT_SHARDS) == CONS_SHARD).count() < threshold) {
+            writeInfos.addAll(generateHundredWrites(counter++, startTs));
+        }
+        sweepQueue.enqueue(writeInfos);
+        return writeInfos.stream()
+                .collect(Collectors.toMap(write -> write.toShard(DEFAULT_SHARDS), write -> 1, (fst, snd) -> fst + snd));
+    }
```
jeremyk-91: A little hacky, but the point of hashes is to have preimage resistance I guess 😄

PR #3326, opened 2018-06-27
Actually rate limit detailed profile logging

jeremyk-91:
**Goals (and why)**:
- The detailed profile logging was supposed to be written once every 5 seconds, but the rate-limiter was applied on a per-`SnapshotTransaction` level so it wasn't. This PR fixes that.

- Have a single `RateLimitedBooleanSupplier` in the `SnapshotTransactionManager` that is passed to all transactions created by that TM.

- Does this actually work?
- This does not rate limit across TransactionManagers - is that a problem?

**Where should we start reviewing?**: `SnapshotTransactionManager` and the test associated with it, probably

**Priority (whenever / two weeks / yesterday)**: yesterday-ish. Probably doesn't break stacks, but generates excessive log volume and makes reading logs in internal tool without the relevant filter difficult.

@ashrayjain for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3326)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -35,6 +35,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
+import java.util.function.BooleanSupplier;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -117,7 +118,6 @@ import com.palantir.atlasdb.transaction.impl.logging.ImmutableChainingLogConsume
 import com.palantir.atlasdb.transaction.impl.logging.ImmutableTransactionCommitProfile;
 import com.palantir.atlasdb.transaction.impl.logging.LogConsumerProcessor;
 import com.palantir.atlasdb.transaction.impl.logging.PredicateBackedLogConsumerProcessor;
-import com.palantir.atlasdb.transaction.impl.logging.RateLimitedBooleanSupplier;
 import com.palantir.atlasdb.transaction.impl.logging.TransactionCommitProfile;
 import com.palantir.atlasdb.transaction.service.TransactionService;
 import com.palantir.atlasdb.util.MetricsManager;
@@ -222,8 +222,8 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
     protected volatile boolean hasReads;
 
     private final Timer.Context transactionTimerContext;
-
-    private final CommitProfileProcessor profileProcessor = createDefaultCommitProfileProcessor();
+    protected final BooleanSupplier shouldProfile;
+    protected final CommitProfileProcessor profileProcessor;
 
     /**
      * @param immutableTimestamp If we find a row written before the immutableTimestamp we don't need to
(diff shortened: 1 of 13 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -275,6 +276,8 @@
         this.sweepQueue = sweepQueue;
         this.deleteExecutor = deleteExecutor;
         this.hasReads = false;
+        this.shouldProfile = shouldProfile;
+        this.profileProcessor = createDefaultCommitProfileProcessor();
```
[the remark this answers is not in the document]
gsheasby: Or a spare 5 minutes?
1. pass `shouldProfile` into `createDefaultCommitProfileProcessor`, and remove the field
2. move the `createDefaultCommitProfileProcessor` logic to `CommitProfileProcessor`, taking `getTimer` and `getHistogram` with it. All the calls to those methods would become `profileProcessor.getTimer`, which doesn't sound too bad. I would keep `SnapshotTransaction.class` as the metric class in `getTimer` and `getHistogram` for legacy reasons, and because the `ProfileProcessor` is responsible for recording metrics for SnapshotTransactions.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerTest.java
```diff
@@ -168,4 +170,15 @@ public void registersMetrics() throws InterruptedException {
         assertThat(registry.getTimers().get(SETUP_TASK_METRIC_NAME).getCount()).isGreaterThanOrEqualTo(1);
         assertThat(registry.getTimers().get(FINISH_TASK_METRIC_NAME).getCount()).isGreaterThanOrEqualTo(1);
     }
+
+    @Test
+    public void profilingIsSharedAcrossTransactions() {
```
gsheasby: :100:

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -117,7 +117,8 @@ protected Transaction startTransaction() {
                 AbstractTransactionTest.GET_RANGES_EXECUTOR,
                 AbstractTransactionTest.DEFAULT_GET_RANGES_CONCURRENCY,
                 getSweepQueueWriterInitialized(),
-                MoreExecutors.newDirectExecutorService()) {
+                MoreExecutors.newDirectExecutorService(),
+                () -> false) {
```
gsheasby: When we're passing `() -> false` in, we could pass in something like `CommitLogProcessor.neverProfile()`

docs/source/release_notes/release-notes.rst
```diff
@@ -60,6 +60,11 @@ develop
            This should reduce request volumes on TimeLock Server.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3303>`__)
 
+    *    - |fixed|
+         - Snapshot transactions now write detailed profiling logs of the form ``Committed {} bytes with locks...`` only once every 5 seconds per ``TransactionManager`` used.
+           Previously, they were written on every transaction.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/33>`__)
```
gsheasby: polite reminder: number :)

jeremyk-91: yep, thanks! Added.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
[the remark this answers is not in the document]
jeremyk-91: There are some non-commit things that also use `getTimer`/`getHistogram` though that aren't part of commit-stage profiling (e.g. `getRows`), so I wouldn't want to exclusively move these into the profiler.

We can discuss ideas for cleaning this up offline. The tricky bit here is that serializable transactions create read transactions of their own (we could of course have 2 constructors where one takes `shouldProfile` and one takes the processor, but that's pretty nasty)

PR #3332, opened 2018-06-29
Release Notes 94

jeremyk-91:
**Goals (and why)**: Make sure people know what is in what version

- Add a header. That's all

- Not having any edits is rare, but what follows seems fine. Is this the case?

**Where should we start reviewing?**: `release-notes.rst`

**Priority (whenever / two weeks / yesterday)**: today, it's small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3332)

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,6 +50,22 @@ develop
     *    - Type
          - Change
 
+    *    -
+         -
+
+=======
+v0.94.0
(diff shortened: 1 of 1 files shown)


gsheasby:
Um. actually I can't see the multi sweep note at all...

jeremyk-91:
... discussed offline - they're in 0.90 🔥but we'll fix it in a separate PR

PR #3337, opened 2018-07-03
FFT - 1st ed., Part 6: Dependency Chain

jeremyk-91:
**Goals (and why)**:
- Outline why certain steps must run before others, so no one accidentally tries to parallelize them 

- Add comments explaining why the sequencing matters

- Is this the right place, or should this live in the docs? I thought it reasonable to put this here as it's already present to some extent and is more visible.
- Are any of the error cases proposed incorrect?
- There are two parts of the proof (serializable tx, and lock check) where I'm not claiming a dependency on the immediate next step. Are there stronger proofs, or are those actually safe to parallelize?

**Where should we start reviewing?**: `SnapshotTransaction`

**Priority (whenever / two weeks / yesterday)**: this week, it's small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3337)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1323,47 +1323,57 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
         Optional<TransactionCommitProfile> optionalProfile = Optional.empty();
         Timer.Context commitStageTimer = getTimer("commitStage").time();
 
+        // Acquire row locks and a lock on the start timestamp row in the transactions table.
+        // This must happen before conflict checking, otherwise we could complete the checks and then have someone
+        // else write underneath us before we proceed (thus missing a write/write conflict).
         Timer.Context acquireLocksTimer = getTimer("commitAcquireLocks").time();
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1323,47 +1323,57 @@ private void commitWrites(TransactionService transactionService) {
...
             // we won't block forever waiting for the unreadable timestamp to advance past the
-            // scrub timestamp (same as the hard delete transaction's start timestamp)
+            // scrub timestamp (same as the hard delete transaction's start timestamp).
+            // May not need to be here specifically, but this is a very cheap operation - scheduling another thread
+            // might well cost more.
             long microsForPunch = runAndReportTimeAndGetDurationMicros(
                     () -> cleaner.punch(commitTimestamp),
                     "microsForPunch");
 
+            // Serializable transactions need to check their reads haven't changed, by reading again at commitTs + 1.
```
sandorw: As we discussed this morning, mention that this needs to happen before checking lock validity. I don't think we need to sketch out the entire issue, might be enough to just say that thorough sweep is the problem case.

jeremyk-91:
I've added this as incorrect-F under the out of order tx protocol doc.

Will sketch something out in the OSS docs as well. The comments here are meant to be quick intuitions only

PR #3338, opened 2018-07-03
Don't run in-process compaction from sweep CLI.

gsheasby:
Where useful, this operation should be handled by the background compactor instead.

**Do not merge** without internal product approval.

**Goals (and why)**:
- Manage compaction 
- Fix an error that occurred in internal product testing

- Delete an if block

1. Should we keep this CLI around at all?
2. Does this break any internal product workflow? @mswintermeyer 
3. Just the minor existential concern that happens when you change some code functionality and don't break any unit tests 😞 

**Where should we start reviewing?**: SweepCommand.java

**Priority (whenever / two weeks / yesterday)**: ASAP, the error blocks rollout of the background compactor to places that need it.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3338)

### DIFF
--- a/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/SweepCommand.java
+++ b/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/SweepCommand.java
@@ -19,7 +19,6 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


gsheasby:
As per chat conversation with @mswintermeyer, on the latest internal product version, deployments can use sweep priority overrides and MultiSweep instead of the CLI, and the background compaction thread should pick up these calls.

If we have to backport this fix to other LTS branches, then we should add an option to also run `compactInternally`, so that existing functionality is not removed for those branches.

Issue #2539, opened 2017-10-20
delete putUnlessExists generated methods


nziebart:
These have been deprecated for 2 months now, let's delete them

hsaraogi:
Lets check that noone uses it and then delete them.

PR #3340, opened 2018-07-03
Kill DataTable.putUnlessExists()

### LINKED ISSUE #2047: putUnlessExists in generated table code is misleading
Example of code as it is generated today
```
@Override
    public void putUnlessExists(Multimap<BucketsRow, ? extends BucketsNamedColumnValue<?>> rows) {
        Multimap<BucketsRow, BucketsNamedColumnValue<?>> existing = getRowsMultimap(rows.keySet());
        Multimap<BucketsRow, BucketsNamedColumnValue<?>> toPut = HashMultimap.create();
        for (Entry<BucketsRow, ? extends BucketsNamedColumnValue<?>> entry : rows.entries()) {
            if (!existing.containsEntry(entry.getKey(), entry.getValue())) {
                toPut.put(entry.getKey(), entry.getValue());
            }
        }
        put(toPut);
    }
```

Before looking at the implementation, I would have expected `putUnlessExists` of a `(row, col, val)` triple to write that triple if the cell `(row, col)` does not exist in the table. This is e.g., what `java.util.Map#putIfAbsent` does if you consider the cell to be a key.

Instead it writes the `(row, col, val)` triple if that entire triple does not exist in the table, allowing you to overwrite an existing triple with a different value.

Now, this is a perfectly reasonable method to have, and may indeed have been what the author intended, but it is very misleading given different behavior of similar functions like `java.util.Map#putIfAbsent`.

What I'd like to see happen is for a method that acts like `java.util.Map#putIfAbsent` to be added, and for the current `putUnlessExists` method to be renamed to something else (as a deliberate compilation break so clients notice that there are two reasonable behaviors).

Feel free to bikeshed, but reasonable names to me are `putUnlessCellExists` and `putUnlessCellAndValueExists`.

jeremyk-91:
**Goals (and why)**: Fix #2047 and #2539 

- Remove the API from the table renderer
- Remove the API from all generated code tables in AtlasDB, by regenerating all schemas

- I noticed that regenerateSchemas in AtlasDB doesn't actually regenerate all the schemas. I didn't fix this here as it would require a bit of gradle digging.
- This is obviously a dev break, but it's been deprecated for 10 months so there's nothing holding me back here.
- I checked the docs and I think all references to `putUnlessExists` there refer to the KVS API, not the table-level API, but could be worth a sanity check.

**Where should we start reviewing?**: `TableRenderer`

@tpetracca for SA

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3340)

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -53,6 +53,7 @@ develop
     *    - |devbreak|
          - The ``putUnlessExists`` API has been removed from AtlasDB tables, as it was misleading (it only did the put if the given row, column and value triple were already present, as opposed to the more intuitive condition of the row and column value pair being present).
            Please replace any uses of ``putUnlessExists`` with a get, check and put - these will still be transactional because of the AtlasDB transaction protocol.
+           Note that this is not the same as the KVS ``putUnlessExists`` API, which is still used by the transaction protocol.
            This API has already been deprecated for 11 months.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3340>`__)
 
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,8 +50,11 @@ develop
     *    - Type
          - Change
 
-    *    -
-         -
+    *    - |devbreak|
+         - The ``putUnlessExists`` API has been removed from AtlasDB tables, as it was misleading (it only did the put if the given row, column and value triple were already present, as opposed to the more intuitive condition of the row and column value pair being present).
+           Please replace any uses of ``putUnlessExists`` with a get, check and put - these will still be transactional because of the AtlasDB transaction protocol.
+           This API has already been deprecated for 11 months.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3340>`__)
 
 =======
 v0.94.0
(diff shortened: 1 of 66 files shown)


docs/source/release_notes/release-notes.rst
```diff
@@ -52,9 +52,9 @@ develop
 
     *    - |devbreak|
          - The ``putUnlessExists`` API has been removed from AtlasDB tables, as it was misleading (it only did the put if the given row, column and value triple were already present, as opposed to the more intuitive condition of the row and column value pair being present).
-           Please replace any uses of ``putUnlessExists`` with a get, check and put - these will still be transactional because of the AtlasDB transaction protocol.
+           Please replace any uses of the table-level ``putUnlessExists`` with a get, check and put if appropriate - these will still be transactional because of the AtlasDB transaction protocol.
            Note that this is not the same as the KVS ``putUnlessExists`` API, which is still used by the transaction protocol.
-           This API has already been deprecated for 11 months.
+           This API has already been deprecated for 10 months (since August 2017).
```
snegrus-adb: nit: the relative value will be wrong next month

jeremyk-91: I'll put the exact date first. I've changed the relative part to say `11 months from time of writing` (how long it has been in this state for is also important!)

PR #3341, opened 2018-07-03
[baseline] timestamp-impl + commons-executors

gsheasby:
**Goals (and why)**: Cleaning up baseline issues in a spare moment

**Implementation Description (bullets)**: - Baseline for timestamp-impl
- Baseline for commons-executors

**Concerns (what feedback would you like?)**: nothing spectacular. Keep in mind that we want to avoid breaking internal products with these changes.

**Where should we start reviewing?**: wherever, each change is pretty independent

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3341)

### DIFF
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
@@ -36,8 +36,9 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor.AbortPolicy;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.palantir.remoting3.tracing.Tracers;
 
@@ -50,6 +51,7 @@ import com.palantir.remoting3.tracing.Tracers;
  */
 @SuppressWarnings("checkstyle:AbbreviationAsWordInName")
 public final class PTExecutors {
+    private static Logger log = LoggerFactory.getLogger(PTExecutors.class);
 
     private static final String FILE_NAME_FOR_THIS_CLASS = PTExecutors.class.getSimpleName() + ".java";
 
@@ -557,8 +559,7 @@ public final class PTExecutors {
                 }
             }
         }
-        String errorMessage = "Can't figure out what name to use for this thread factory!";
-        Logger.getLogger(PTExecutors.class.getName()).log(Level.WARNING, errorMessage);
+        log.warn("Can't figure out what name to use for this thread factory!");
         return new NamedThreadFactory("Unnamed thread", isDaemon);
     }
 
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
@@ -40,6 +40,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.remoting3.tracing.Tracers;
 
 /**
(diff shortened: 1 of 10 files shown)


commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
```diff
@@ -432,10 +435,9 @@ public static ScheduledThreadPoolExecutor newScheduledThreadPoolExecutor(int cor
     @SuppressWarnings("DangerousThreadPoolExecutorUsage")
     public static ScheduledThreadPoolExecutor newScheduledThreadPoolExecutor(int corePoolSize,
             ThreadFactory threadFactory, RejectedExecutionHandler handler) {
-        if (corePoolSize == 0) {
-            corePoolSize = 1;
-        }
-        ScheduledThreadPoolExecutor ret = new ScheduledThreadPoolExecutor(corePoolSize, threadFactory, handler) {
+        int positiveCorePoolSize = corePoolSize > 0 ? corePoolSize : 1;
+        ScheduledThreadPoolExecutor ret = new ScheduledThreadPoolExecutor(positiveCorePoolSize, threadFactory,
+                handler) {
```
jeremyk-91: I think this does actually change behaviour (if someone passed in negative values before, the `ScheduledThreadPoolExecutor` constructor would throw, while now we'll silently set it to 1) - given how widely used this class is I'd prefer for us not to change this too much.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
```diff
@@ -490,41 +492,24 @@ public static Runnable wrap(final Runnable runnable) {
         return new ForwardingRunnableFuture<>(wrappedRunnable, rf);
     }
 
-    private static Runnable wrapRunnable(Runnable runnable) {
+    /**
+     * Wraps the given {@code Callable} so that {@link ExecutorInheritableThreadLocal} variables are
+     * propagated through.
+     */
```
jeremyk-91: 👍

timestamp-impl/src/test/java/com/palantir/timestamp/PersistentTimestampServiceMockingTest.java
```diff
@@ -35,23 +30,22 @@
...
     private static final long INITIAL_TIMESTAMP = 12345L;
     private static final long TIMESTAMP = 100 * 1000;
-    private static final TimestampRange SINGLE_TIMESTAMP_RANGE = TimestampRange.createInclusiveRange(TIMESTAMP, TIMESTAMP);
+    private static final TimestampRange SINGLE_TIMESTAMP_RANGE = TimestampRange.createInclusiveRange(TIMESTAMP,
+            TIMESTAMP);
 
     private static final TimestampRange RANGE = TimestampRange.createInclusiveRange(100, 200);
 
     private PersistentTimestamp timestamp = mock(PersistentTimestamp.class);
-    private ExecutorService executor = Executors.newSingleThreadExecutor();
```
jeremyk-91: Nice. I'm surprised this executor still existed at all - but I guess this is why we have baseline

gsheasby: This wasn't baseline - I randomly found some unused code while fixing other baseline issues with that file, and followed the [boy scout rule](https://deviq.com/boy-scout-rule/).

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
gsheasby: Fair, I'll fix this.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
gsheasby: Huh, this was just a method reordering, but it seems the methods are rather similar 😂

timestamp-impl/src/test/java/com/palantir/timestamp/PersistentTimestampServiceMockingTest.java
jeremyk-91: Ah right. I always thought 'variable that was never used' was a baseline rule (it's probably just one of the intellij checks)

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
jeremyk-91: Yep. Realised that after the fact - but still, good that what this method does is explained

PR #3342, opened 2018-07-03
[TS] Write to sweep queue at start timestamp

gmaretic:
**Goals (and why)**:
Do not write to sweepable cells and sweepable timestamps at timestamp 0, use the transaction's start timestamp instead. Also, use multiPut, and ensure references are written first.

Is always reading at `Long.MAX_VALUE` reasonable?

@j-baker @ashrayjain for SA

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
@@ -25,6 +25,7 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
 import com.palantir.atlasdb.keyvalue.api.BatchColumnRangeSelection;
 import com.palantir.atlasdb.keyvalue.api.Cell;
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
```diff
@@ -36,8 +36,7 @@
     public static final int MAX_CELLS_DEDICATED = 100_000;
     public static final int SWEEP_BATCH_SIZE = MAX_CELLS_DEDICATED;
     public static final int BATCH_SIZE_KVS = 1000;
-    public static final long READ_TS = 1L;
-    public static final long WRITE_TS = 0L;
+    public static final long READ_TS = Long.MAX_VALUE;
```
jeremyk-91: This should be safe, since `(timestampPartition, timestampModulus, writeIndex)` which feature in the row/column components is a unique triple, so nothing should ever get written more than once

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
```diff
@@ -52,17 +53,35 @@ public KvsSweepQueueWriter(KeyValueService kvs, TableReference tableRef, WriteIn
...
+        });
+
+        partitionedWrites.keySet().stream()
+                .map(PartitionInfo::timestamp)
+                .mapToLong(x -> x)
+                .max()
+                .ifPresent(timestamp -> {
+                    write(referencesToDedicatedCells, timestamp);
+                    write(cellsToWrite, timestamp);
+                });
```
jeremyk-91: `max()` seems unnecessary given that right now each enqueue strictly consists of writes at the same timestamp, but post offline discussion there may be plans to do something different or cleverer later on.

Also: we may want to revisit this much later along the FFT project, if we're trying to make things faster. In particular, it looks like we now call multiPut twice instead of once.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepableTimestamps.java
```diff
@@ -41,6 +41,11 @@ public SweepableTimestamps(KeyValueService kvs, WriteInfoPartitioner partitioner
         super(kvs, TargetedSweepTableFactory.of().getSweepableTimestampsTable(null).getTableRef(), partitioner, null);
     }
 
+    @Override
+    Map<Cell, byte[]> populateReferences(PartitionInfo partitionInfo, List<WriteInfo> writes) {
+        return ImmutableMap.of();
+    }
```
jeremyk-91: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/KvsSweepQueueWriter.java
gmaretic: For the first point, you're right that all the timestamps should be the same. Maybe it's a bit of too defensive coding, but the contract of enqueue does not specify that all the write infos must have the same start timestmap.For the second point, this is because we don't want to write dedicated cells and then explode without writing the references, which can happen depending on how the writes get batched internally.

PR #3343, opened 2018-07-03
FFT - 1st ed., Part 7: One-Shot Start AtlasDB Transaction

jeremyk-91:
**Goals (and why)**: Make transactions faster ~because perf is awesome~ so that users have smoother experiences using AtlasDB. This PR specifically seeks to eliminate 1.5 round trips between AtlasDB and TimeLock when an Atlas user opens a transaction.

- Add a `start-atlasdb-transaction` endpoint on TimeLock that is basically (lock immutable timestamp + get fresh timestamp) _in that order_ (ordering is key).
- Implement an adapter for letting new versions of AtlasDB talk to old TimeLock Servers.

- I renamed the `LockImmutableTimestampRequest` to `IdentifiedTimeLockRequest` because it's used by both `lockImmutableTimestamp` and `startAtlasDbTransaction`. The serialization of the class has not changed (verified this by running a new client against an old server), but are there other ways this might break people (maybe if they use timelock directly for immutable timestamp work?)
- Is there an easier way to handle the `ImmutableTimestampBridgingTimelockService`?
- Is there a better name for that class? It's a monster
- The fresh timestamp used to be batched but now isn't any more - but I think this is fine as this does not increase the number of calls made to timelock in general.

**Where should we start reviewing?**: `AsyncTimelockResource`, probably.

**Priority (whenever / two weeks / yesterday)**: this week. I'd like to finish up the 'shuffle around the tx protocol' steps as quickly as possible - right now as far as I know there's only this and maybe running Pre-Commit Conditions and the final lock check in parallel.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3343)


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+    @After
+    public void verifyMocks() {
+        verifyNoMoreInteractions(timelockService);
+    }
+
+    @Test
+    public void defersToIndividualStepsIfStartTransactionNotAvailable() {
+        givenServerThrows404sOnStartTransaction();
+
+        whenStandardClientStartsTransactionNTimes(1);
```
gmaretic: nit: method name is awkward, maybe better just `whenStandardClientStartsTransactionTimes`? Also goes for all the subsequent methods

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+        when(timelockService.lockImmutableTimestamp(any())).thenReturn(IMMUTABLE_TIMESTAMP_RESPONSE);
+    }
+
+    @After
+    public void verifyMocks() {
+        verifyNoMoreInteractions(timelockService);
+    }
+
+    @Test
+    public void defersToIndividualStepsIfStartTransactionNotAvailable() {
```
gmaretic: Pretty cool way to write the tests. Very easy to read! 🏅

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+    public void passesThroughExceptionsWithOtherErrorCodes() {
+        givenServerThrowsOnStartTransaction(REMOTE_EXCEPTION_503);
+
+        whenStandardClientStartsTransactionNTimes(5);
+
+        thenStartTransactionIsCalledNTimes(5);
+        thenCallsFailedWithExceptions(Collections.nCopies(5, REMOTE_EXCEPTION_503));
+    }
+
+    @Test
```
gmaretic: Awesome tests! Maybe add one that checks that when we go from 503 to TRANSACTION_RESPONSE, we call the new endpoint only?

lock-api/src/main/java/com/palantir/lock/v2/IdentifiedTimeLockRequest.java
```diff
@@ -24,15 +24,15 @@
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 @Value.Immutable
-@JsonSerialize(as = ImmutableLockImmutableTimestampRequest.class)
-@JsonDeserialize(as = ImmutableLockImmutableTimestampRequest.class)
-public interface LockImmutableTimestampRequest {
+@JsonSerialize(as = ImmutableIdentifiedTimeLockRequest.class)
+@JsonDeserialize(as = ImmutableIdentifiedTimeLockRequest.class)
+public interface IdentifiedTimeLockRequest {
```
gmaretic: Why the rename? Will the old TimeLock be able to deserialize this as a `LockImmutableTimestampRequest`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -142,19 +142,21 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
         try {
-            LockToken immutableTsLock = immutableTsResponse.getLock();
-            long immutableTs = immutableTsResponse.getImmutableTimestamp();
+            LockToken immutableTsLock = transactionResponse.immutableTimestamp().getLock();
+            long immutableTs = transactionResponse.immutableTimestamp().getImmutableTimestamp();
             recordImmutableTimestamp(immutableTs);
-            Supplier<Long> startTimestampSupplier = getStartTimestampSupplier();
+
+            cleaner.punch(transactionResponse.freshTimestamp());
+            Supplier<Long> startTimestampSupplier = Suppliers.ofInstance(transactionResponse.freshTimestamp());
```
gmaretic: The ordering here looks good. Nothing really changed except that we now get the fresh timestamp and punch before the transaction even starts.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -137,9 +138,7 @@ public void shouldConflictIfImmutableTimestampLockExpiresEvenIfNoWritesOnThoroug
                 MultiTableSweepQueueWriter.NO_OP,
                 MoreExecutors.newDirectExecutorService());
 
-        when(timelock.getFreshTimestamp()).thenReturn(1L);
-        when(timelock.lockImmutableTimestamp(any())).thenReturn(
-                LockImmutableTimestampResponse.of(2L, LockToken.of(UUID.randomUUID())));
+        setupTimeLock(timelock);
```
gmaretic: All of the setup above is also duplicated, so setup should take no arguments and return a `TransactionManager` instead.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
+         - TimeLock Server now exposes a ``startAtlasDbTransaction`` endpoint which locks an immutable timestamp and then gets a fresh timestamp (in a single round-trip call); new TimeLock clients call this endpoint.
+           This saves an estimated one TimeLock round-trip of latency when committing a transaction.
```
gmaretic: when starting?

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
```diff
@@ -76,12 +77,19 @@ public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
     }
 
     @Override
-    public LockImmutableTimestampResponse lockImmutableTimestamp(LockImmutableTimestampRequest request) {
+    public LockImmutableTimestampResponse lockImmutableTimestamp(IdentifiedTimeLockRequest request) {
```
gmaretic: Is this not an API break?

lock-api/src/main/java/com/palantir/lock/v2/LockImmutableTimestampResponse.java
```diff
@@ -21,7 +21,7 @@
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
-// TODO(nziebart): also return a start timestamp here?
+// TODO(jkong): Can this be deprecated?
```
gmaretic: Kind of strange to deprecate when it's used in StartAtlasDbTransactionResponse

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockServiceTest.java
jeremyk-91: Alright, will remove the `N` before `Times` in most of these

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockServiceTest.java
jeremyk-91: Yep, makes sense - will add this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
jeremyk-91: Yep. It _does_ shift earlier, but I think that's fine. Punching is async and we punch again later; there aren't any significant steps we'll change the order of if I'm understanding this correctly.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
jeremyk-91: Well spotted! Changed.

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
jeremyk-91: The serialized form of `LockImmutableTimestampRequest` and `IdentifiedTimeLockRequest` are the same. We don't use the type name (de)serializer, so the serialized form of *both* of these objects is along the lines of

```
{"requestId":"ed4b106f-8718-4858-a402-d99415124cfe"}
```

(I've also tested this with an old timelock to verify things are compatible.)

lock-api/src/main/java/com/palantir/lock/v2/IdentifiedTimeLockRequest.java
jeremyk-91: I renamed this because calling an endpoint called `startAtlasDbTransaction` (for that matter, anything not named `lockImmutableTimestamp` or something close) with a `LockImmutableTimestampRequest` seemed awkward to me.

Deserialization: yep, that's a valid concern - it does deserialize. I've added a deserialization test as well.

lock-api/src/main/java/com/palantir/lock/v2/LockImmutableTimestampResponse.java
jeremyk-91: Fair enough - I think I should put this on the immutable timestamp _endpoint_, not this object!

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -657,7 +658,16 @@ static LockAndTimestampServices createLockAndTimestampServices(
...
-                withRefreshingLockService(lockAndTimestampServices));
+                withRefreshingLockService(
+                        withBridgingTimelockService(lockAndTimestampServices)));
+    }
+
+    private static LockAndTimestampServices withBridgingTimelockService(
+            LockAndTimestampServices lockAndTimestampServices) {
+        return ImmutableLockAndTimestampServices.builder()
+                .from(lockAndTimestampServices)
+                .timelock(ImmutableTimestampBridgingTimeLockService.create(lockAndTimestampServices.timelock()))
```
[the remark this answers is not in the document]
jeremyk-91: Yep. The raw services method is now private.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockService.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.common.util.concurrent.RateLimiter;
+import com.palantir.atlasdb.http.errors.AtlasDbRemoteException;
+import com.palantir.lock.v2.AutoDelegate_TimelockService;
+import com.palantir.lock.v2.IdentifiedTimeLockRequest;
+import com.palantir.lock.v2.StartAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.processors.AutoDelegate;
+
+/**
+ * This class serves as a bridge to allow newer Atlas clients to maintain compatibility with TimeLock servers that
```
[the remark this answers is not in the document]
jeremyk-91: Yep, sticking with `TimeLock` (changed the reference in paragraph 2)

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockService.java
```diff
@@ -0,0 +1,101 @@
...
+    public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLockRequest request) {
+        if (shouldUseStartAtlasDbTransactionEndpoint()) {
+            try {
+                StartAtlasDbTransactionResponse response = delegate.startAtlasDbTransaction(request);
+                serverSupportsStartAtlasDbTransaction = true;
+                return response;
+            } catch (AtlasDbRemoteException remoteException) {
+                if (remoteException.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
+                    // The server 404'd which almost certainly means it doesn't have our new APIs yet
+                    serverSupportsStartAtlasDbTransaction = false;
```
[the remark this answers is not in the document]
jeremyk-91: If TimeLock goes down I don't think you'll get an `AtlasDbRemoteException` at all (connection-related exceptions don't create `RemoteException`s). If something inside TimeLock e.g. Paxos breaks down you'll get a 5xx.

gmaretic:
Is the deserialization test you're referring to the main method in `IdentifiedTimeLockRequest`? Seems like you forgot to commit the test in while leaving a debugging method around. I removed it, and am generally OK with merging this in, but would be nice to have that test

jeremyk-91:
@gmaretic Ah I forgot to commit the deserialization test :( will push it

jeremyk-91:
@gmaretic would be good to have a final quick look before merge, thanks!

gmaretic:
Looks good! Thanks for the test!

PR #3345, opened 2018-07-03
Call async initialization callback after initialization

pkoenig10:
**Goals (and why)**:
Update the initialization status before running the async initialization callback. Clients need a way to perform initialization tasks once the transaction manager is initialized. Unfortunately in the current implementation, clients may not be able to call transaction manager methods and thus don't have a hook for post-initialization tasks.

@gmaretic What was the motivation for implementing it in the other order in #3011 (it appears deliberate)?

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3345)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
@@ -149,8 +149,8 @@ public class SerializableTransactionManager extends SnapshotTransactionManager {
 
         private void runCallback() {
             try {
-                callback.runWithRetry(txManager);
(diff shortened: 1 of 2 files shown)


pkoenig10:
I have a service with components that require a transaction manager but are created synchronously on startup. These components need to do some additional initialization once the transaction manager is initialized.

It seems like the correct solution is the use the delegate transaction manager. However for some of these components using this delegate transaction manager is not feasible.

For example, one initialization task can only be safely performed once the store is able to write data. However the store needs to be instantiated synchronously and thus uses the wrapped transaction manager.

gmaretic:
I'm going to close this out as we don't to change this behaviour. Tracking better documentation in https://github.com/palantir/atlasdb/issues/3353.

Happy to follow-up with you offline on how to make ti work for your use-case

PR #3346, opened 2018-07-04
randomize backoff period for conservative strategy

gokcanozakdag:
**Goals (and why)**:
Randomize the back off period for conservative strategy. I wanted to randomize it without changing expected value. Behavior is somewhat similar to http-remoting now.

| numRetries | conservative_expected(ms) | randomized_expected(ms) | randomized_min(ms) | randomized_max(ms)
|------------|------------------------|------------------------|-----------------|------------------|
|1 |1000 |1000 |500 |1500 |
|2 |2000 |2000 |1000 |3000 |
|3 |4000 |4000 |2000 |6000 |
|4 |8000 |8000 |4000 |12000 |
|5 |16000 |16000 |8000 | 24000 |
|6 |30000 |30000 |15000 |45000 |

Multiplying the value with a double coeff changing between 0.5 and 1.5.

`CassandraRequestExceptionHandler`

This week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3346)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
@@ -56,6 +56,27 @@ class CassandraRequestExceptionHandler {
         this.blacklist = blacklist;
     }
 
+    public static void main(String[] args) {
+        long[] vals = new long[10];
+        long[] min = new long[10];
+        long[] max = new long[10];
+        for (int i = 0; i < 10; i++) {
+            min[i] = Long.MAX_VALUE;
+        }
+        for (int j = 0; j < 1000; j++) {
+            for (int i = 1; i < 10; i++) {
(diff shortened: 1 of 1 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraRequestExceptionHandler.java
```diff
@@ -278,7 +278,8 @@ public boolean shouldBackoff(Exception ex) {
 
         @Override
         public long getBackoffPeriod(int numberOfAttempts) {
-            return Math.min(500 * (long) Math.pow(2, numberOfAttempts), MAX_BACKOFF);
+            double randomCoeff = ThreadLocalRandom.current().nextDouble() + 0.5;
```
tboam: Feels a bit odd that we specify the average maximum but don't really control the actual maximum - I think we really want to control the maximum and have that be 30.  You could either cap everything at 30s, or take 30s as the maximum, then calculate 2/3 of this and apply the +/-50% from this value.

gokcanozakdag:
@tboam changed the implementation to respect max_backoff. Do you think it is clear enough?

PR #3349, opened 2018-07-04
Feature/stop sweep priority writes

### LINKED ISSUE #3226: Stop writing to sweep.priority if targeted sweep is on
Targeted sweep will deal with the new data, and background sweep to eventually catch up and get to a state where it says NOTHING_TO_SWEEP (and it will stay there).

gsheasby:
**Goals (and why)**: Fixes #3226. We want background sweep to eventually catch up, so we can retire it.

* Simply don't wrap the kvs in SweepStatsKVS if TargetedSweep.enableSweepQueueWrites is set (it's an install parameter, so won't change after TM.create)

**Concerns (what feedback would you like?)**: Is it really this simple?

**Priority (whenever / two weeks / yesterday)**: before release?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3349)

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -273,10 +273,8 @@ public abstract class TransactionManagers {
         Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier =
                 () -> runtimeConfigSupplier().get().orElse(defaultRuntime);
 
-        QosClient qosClient = initializeCloseable(
-                () -> getQosClient(metricsManager,
-                        JavaSuppliers.compose(AtlasDbRuntimeConfig::qos, runtimeConfigSupplier)),
-                closeables);
+        QosClient qosClient = initializeCloseable(() -> getQosClient(metricsManager,
+                        JavaSuppliers.compose(AtlasDbRuntimeConfig::qos, runtimeConfigSupplier)), closeables);
 
         FreshTimestampSupplierAdapter adapter = new FreshTimestampSupplierAdapter();
         ServiceDiscoveringAtlasSupplier atlasFactory = new ServiceDiscoveringAtlasSupplier(metricsManager,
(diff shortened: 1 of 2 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -299,12 +297,16 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
             KeyValueService kvs = atlasFactory.getKeyValueService();
             kvs = ProfilingKeyValueService.create(kvs);
-            kvs = SweepStatsKeyValueService.create(kvs,
-                    new TimelockTimestampServiceAdapter(lockAndTimestampServices.timelock()),
-                    JavaSuppliers.compose(SweepConfig::writeThreshold, sweepConfig),
-                    JavaSuppliers.compose(SweepConfig::writeSizeThreshold, sweepConfig)
 
-            );
+            // If we are writing to the sweep queue, then we do not need to use SweepStatsKVS to record modifications.
+            if (!config.targetedSweep().enableSweepQueueWrites()) {
```
sandorw: Do we think this is sufficient? This means that writes are being persisted to the queue, but targeted sweep might not be enabled - we could check both. I can't think of a good reason to stay in that mode, unless we're addressing some critical bug that occurs when targeted sweep processes the queue. Even in that case, this would be a temporary state.

gsheasby: You're right that writing to the sweep queue but not processing it should be a temporary state - just like the state where we are writing sweep stats to `sweep.priority` (i.e. `enableSweepQueueWrites` is false), but sweep is currently disabled.

The large internal product enables and disables sweep on a daily basis due to banned hours config, and so might do the same with targeted sweep. It would seem silly to write sweep stats for the entire lifecycle of a KVS, just because it was started at a time when targeted sweep was turned off.

gsheasby: In fact, enabling sweep.priority writes in this case would actually be harmful, because the background sweeper would end up doing unnecessary work.

Issue #3234, opened 2018-06-04
Remove key skewness measure


sandorw:
It's unclear that this is providing value to us - the schema reviews we do internally catch unhashed keys now. With other internal developments, this doesn't seem like it will be useful anymore, or will provide false positives.

We have a lot of use cases that look like 'row (time bucket) -> dynamic columns', which look skewed since we write to a single time bucket for some time before moving to the next one. Basically, I think we're being careful about our keys and how well distributed they are, and won't miss the kind of issue the skew measure was meant to catch.

sandorw:
Also note that the results were often incorrect as per https://github.com/palantir/atlasdb/pull/3227

PR #3352, opened 2018-07-05
Remove token ranges skewness logger

sandorw:
**Goals (and why)**:
Fixes https://github.com/palantir/atlasdb/issues/3234

Expunge the class and all references from the record

Which sort of release note should it be? devbreak? changed? improved?

Not urgent, but it's an easy review

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3352)


sandorw:
Nice, didn't notice that there was more dead code 🎉

PR #3356, opened 2018-07-05
Less locking

j-baker:
Locks are generally quite hard to reason about, hence all the javadoc in this class and code to deal with the concurrent ways of accessing the set.

This PR changes to use an ArrayBlockingQueue, which I believe to be a better primitive here; e.g. it makes clear that all elements will be seen eventually, and doesn't have e.g. the fairness issues we see before.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3356)

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
@@ -16,16 +16,19 @@
 
 package com.palantir.lock.client;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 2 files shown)


lock-api/src/main/java/com/palantir/lock/client/AsyncTimeLockUnlocker.java
```diff
@@ -39,45 +42,41 @@
  *
  * Concurrency: We want to guarantee that a token T that is enqueued is included in some call to unlockOutstanding.
  * If T can pass the compareAndSet, then T itself is scheduled. If T does not, that means there is some other
- * thread that has scheduled the task, but the task has not retrieved the reference to the set of tokens to be
- * unlocked (because it re-sets unlockIsScheduled to false first, before extracting the reference to the set).
+ * thread that has scheduled the task, but the task has not yet retrieved the current contents of the queue.
  */
 public class AsyncTimeLockUnlocker implements AutoCloseable {
     private static final Logger log = LoggerFactory.getLogger(AsyncTimeLockUnlocker.class);
+    private static final int BACKPRESSURE = 1024;
```
jeremyk-91: :+1: nice, this is good in case timelock goes down

PR #3358, opened 2018-07-05
FFT - 1st ed., Part 8: Rollback Metrics

jeremyk-91:
- [x] Need to add changelog

**Goals (and why)**:
- Get a better idea of what percentage of transactions are failing. This is important if we pursue various methods of improving individual transaction performance that _could_ hurt overall transaction success rate (e.g. no use making transactions 2x faster, if the success rate drops to 1/3).

- Add `TransactionOutcomeMetrics` which is a wrapper around various meters in a tagged metrics registry.
- Log specific table references as tags (safety-screened, of course) in transaction conflicts, which may allow us to see if any tables in particular have high failure rates.

- There are some breaks in metric names for snapshot and serializable transaction conflicts. I think this is fine as we haven't really been relying on them in the past.
- I'm uncertain about test coverage. The core class is reasonably tested and I've injected it in a few places in the snapshot transaction tests. Serializable transaction tests are somewhat harder because of a dependency cycle between impl-shared and tests-shared.
- Rolling back other transactions is something I think would be useful to know about, though it's not really a transaction outcome per se - is it too out of place?
- Is the fancy `TransactionOutcomeMetricsAssert` too heavyweight?

**Where should we start reviewing?**: `TransactionOutcomeMetrics.java`

**Priority (whenever / two weeks / yesterday)**: lower side of 2 weeks

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3358)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetrics.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public void markReadWriteConflict(TableReference tableReference) {
+        getMeterForTable(READ_WRITE_CONFLICT, tableReference).mark();
+    }
+
+    public void markLocksExpired() {
+        getMeter(LOCKS_EXPIRED).mark();
+    }
+
+    public void markPutUnlessExistsFailed() {
```
gsheasby: Note for other reviewers: I initially thought this should also be per-table, but I infer that this is the putUnlessExists to the `_transactions` table.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
```diff
@@ -0,0 +1,140 @@
...
+        super(actual, TransactionOutcomeMetricsAssert.class);
+        taggedMetricRegistry = actual.metricsManager.getTaggedRegistry();
+    }
+
+    public static TransactionOutcomeMetricsAssert assertThat(TransactionOutcomeMetrics actual) {
+        return new TransactionOutcomeMetricsAssert(actual);
+    }
+
+    public TransactionOutcomeMetricsAssert hasSuccessfulCommits(long count) {
+        checkPresentAndCheckCount("successfulCommit", count);
```
gsheasby: `code.refactor.stability.Optional`: Consider making the outcomes into an enum, which would allow you to reference the string here rather than duplicating it.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
```diff
@@ -0,0 +1,140 @@
...
+        assertMetricNotExists(actual.getMetricName("readWriteConflict", getTableReferenceTags(tableReference)));
+        assertMetricNotExists(actual.getMetricName("writeWriteConflict", getTableReferenceTags(tableReference)));
+        return this;
+    }
+
+    private ImmutableMap<String, String> getTableReferenceTags(TableReference tableReference) {
+        return ImmutableMap.of("tableReference", tableReference.getQualifiedName());
+    }
+
+    private void checkPresentAndCheckCount(String name, long count) {
```
gsheasby: strengthening argument for the optional change above:
1 - `(read|write)WriteConflict` appears 3-4 times per operation
2 - these `String` parameters can become `TransactionOutcomeMetric` parameters

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
```diff
@@ -0,0 +1,140 @@
...
+import org.assertj.core.internal.Objects;
+
+import com.codahale.metrics.Metric;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.logging.LoggingArgs;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+public class TransactionOutcomeMetricsAssert extends
```
gsheasby: For the record, I'm happy that this class exists, but I'm not certain that it belongs in `src/main/java` rather than `src/test/java`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -220,6 +220,8 @@
     private final Timer.Context transactionTimerContext;
     protected final CommitProfileProcessor commitProfileProcessor;
 
+    protected final TransactionOutcomeMetrics transactionOutcomeMetrics;
```
gsheasby: nit: no reason why this field and the two above it can't belong in the main block of (mostly) `final` fields.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1217,6 +1220,7 @@ public void abort() {
                 if (hasWrites()) {
                     throwIfPreCommitRequirementsNotMet(null, getStartTimestamp());
                 }
+                transactionOutcomeMetrics.markAbort();
```
gsheasby: Query: Nothing will be recorded in metrics if the pre-commit requirements are not met. Should we record such cases as an `abort`, or a `preCommitCheckFailed`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetrics.java
jeremyk-91: Now that I've moved the string constants to be enumeration names, I can add some javadoc clarifying what the outcomes mean.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
jeremyk-91: Nice package naming trend 😄 This one is worth doing. I initially thought about this but didn't push on the basis that a tx can cause multiple outcomes, but on revisiting it that argument doesn't really make sense.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
jeremyk-91: Done as part of the Enum change.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TransactionOutcomeMetricsAssert.java
jeremyk-91: I'd agree if this was in any other project. This is the `tests-shared` project though, which is used by other tests like the snapshot tx test (that can't see tests-shared's tests, as it shouldn't), so I think it's acceptable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: Fair enough - indeed it looks weird that the volatile boolean is randomly sitting in the middle of final fields. Moved them up.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: I think it's not unreasonable to pick either. I'll mark them as a `preCommitCheckFailed`

jeremyk-91: We could in the future tag these failures with the context of the check failure (not doing this here though!)

PR #3359, opened 2018-07-06
add subtype name 'cassandra' to runtime config

gokcanozakdag:
**Goals (and why)**:
Currently we need to specify `type` in KvsRuntimeConfig as `CassandraKeyValueServiceRuntimeConfig`; which is a bit ugly, and different then what we have for KvsConfig.

Adding subtype name `cassandra` while keeping the old type name, for the sake of backwards compatibility.

`CassandraKeyValueServiceRuntimeConfig.java`

Whenever

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3359)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
```diff
@@ -28,12 +29,19 @@
 @AutoService(KeyValueServiceRuntimeConfig.class)
 @JsonDeserialize(as = ImmutableCassandraKeyValueServiceRuntimeConfig.class)
 @JsonSerialize(as = ImmutableCassandraKeyValueServiceRuntimeConfig.class)
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = CassandraKeyValueServiceRuntimeConfig.class,
```
gsheasby: Do we want to comment that this is deprecated, and add a release note?

PR #2647, opened 2017-11-07
Live Reloading the TimeLock Block, Part 3: Working with 0 Nodes

jeremyk-91:
- [x] Smoketest on internal shopping product

**Goals (and why)**:
Live reload the TimeLock cluster block, see #2622 and #2621.
This part is needed because previous work assumed that we would always have knowledge of at least 1 node, which is not necessarily a valid assumption in a K8s world.

**Implementation Description (bullets)**:
- Remove the min-size restriction of 1 in `ServerListConfig`.
- Re-enforce this min-size in the remote timestamp and lock blocks. Note that the leader block uses its own set of URIs.
- Create a failing proxy if we ever reload a timelock cluster size of 0.
- Add a bunch of tests.

**Concerns (what feedback would you like?)**:
- Technically a dev break, because users of the `ServerListConfig` object may have assumed that there was at least 1 node. Deserialization of the class appears unaffected, so probably not a user break.
- Is `ServiceNotAvailableException` the right thing to throw if making a request to a proxy with no hosts? Should I be throwing `AtlasDbRemoteException`? `IllegalStateException`?!

**Where should we start reviewing?**: `AtlasDbFeignTargetFactory`. Most of the rest is tests.

**Priority (whenever / two weeks / yesterday)**: two weeks?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/2647)

PR #3362, opened 2018-07-09
Make timelockRuntime block mandatory | Remove timelock block 

snegrus-adb:
- [x] make `timelockRuntime` block not optional
- [x] remove `serverList` field from `timelock` install block property from install block to simplify services intialization
- [x] remove `client` field from `timelock` install block because is deprecated
- [x] remove the docs about install config
- [ ] update release notes to add details about `isUsingTimelock` flag
- [ ] update docs about the flag above

**Goals (and why)**:
Timelock Auth feature will add a new field which needs `timelockRuntime` field to be mandatory. I am splitting the work since that PR is already quite bulky. Because this field is now always present the `timelock` block from the install section is not needed anymore and is removed. These changes are not separated because we could end up in a intermediate version with missing `timelock` field and `timelockRuntime` not being mandatory. Another reason is that the `TimeLockClientConfig` was deprecated last year.

I am removing the `Optional<...>` part of the `timelockRuntime` field type in `AtlasDbRuntimeConfig`, and making it having a default value. With the related changes.

Also removing `TimeLockClientConfig` and `timelock` block from config.

Are the proxies going to be created in the right way? Before this PR, if the `timelockRuntime` block is added, even with no servers, we try to create a proxy that just fails. With this PR we want to create a proxy for an external timelock only if there is a server in the list, but not when the block is present and empty. Am I missing anything?

This will probably break some usages where embedded is used and the block is missing, and where the `timelock` block is used to define things like `namespace` or `serverList`.

`AtlasDbRuntimeConfig.java` 

Soon.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3362)


atlasdb-config/src/main/java/com/palantir/atlasdb/config/ServerListConfigs.java
```diff
@@ -27,11 +26,13 @@ private ServerListConfigs() {
     }
 
     public static ServerListConfig parseInstallAndRuntimeConfigs(TimeLockClientConfig installClientConfig,
-            Supplier<Optional<TimeLockRuntimeConfig>> runtimeConfig,
+            Supplier<TimeLockRuntimeConfig> runtimeConfig,
             String namespace) {
-        ServerListConfig nonNamespacedConfig = runtimeConfig.get()
-                .map(TimeLockRuntimeConfig::serversList)
-                .orElse(installClientConfig.serversList());
+        ServerListConfig nonNamespacedConfig = installClientConfig.serversList();
```
gsheasby: This is a behaviour change. Previously, if we have 0 servers in the runtime block and servers in the install config, we picked the runtime block.

With the change of making the runtime block mandatory, there is now no distinction between "not actually configured block" and "configured block that has no servers at the moment".

We may want to deprecate the install config servers list, so that it's no longer valid to have a non-empty list of servers in the install config.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -605,8 +601,8 @@ private static PersistentLockService createAndRegisterPersistentLockService(
         return Callback.noOp();
     }
 
-    private static boolean isUsingTimeLock(AtlasDbConfig atlasDbConfig, AtlasDbRuntimeConfig runtimeConfig) {
-        return atlasDbConfig.timelock().isPresent() || runtimeConfig.timelockRuntime().isPresent();
+    private static boolean isUsingTimeLock(AtlasDbRuntimeConfig runtimeConfig) {
+        return runtimeConfig.timelockRuntime().serversList().hasAtLeastOneServer();
```
jeremyk-91: @SerialVelocity for SA - I have some concerns about using this as a proxy to determine whether we want to use timelock. 

See #2647 for context - there exists a case where we start up with zero servers because discovery can't find them yet in a k8s environment. `TransactionManagers` doesn't look like it handles this transition properly, since the raw services are only created once.

We could perhaps have a more explicit flag saying whether to use timelock or not - I don't have a very good solution offhand.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -746,17 +745,14 @@ private static LockAndTimestampServices createRawServicesFromTimeLock(
...
                 "Cannot create raw services from timelock with another source of timestamps/locks configured!");
-        TimeLockClientConfig clientConfig = config.timelock().orElse(ImmutableTimeLockClientConfig.builder().build());
-        String resolvedClient = OptionalResolver.resolve(clientConfig.client(), config.namespace());
-        return () -> ServerListConfigs.parseInstallAndRuntimeConfigs(
-                clientConfig,
-                () -> runtimeConfigSupplier.get().timelockRuntime(),
-                resolvedClient);
+        return () -> ServerListConfigs.addNamespaceToServerAddresses(
+                runtimeConfigSupplier.get().timelockRuntime().serversList(), config.namespace().get());
     }
```
jeremyk-91: You previously could specify a TimeLock client specifically (that was different from your namespace). That was done at the time namespaces were introduced to avoid a user-break in terms of config. Now that you're removing the timelock client configuration option, the following looks OK.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,14 @@ develop
     *    - Type
          - Change
 
+    *    - |userbreak|
+         - The ``timelockRuntime`` block is now mandatory. Before this it was optional.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3362>`__)
+
+    *    - |userbreak|
+         - The ``timelock`` block from the install config haa been removed. The block was deprecated in v0.67.0 (released on 15th of November 2017).
```
jeremyk-91: nit: has

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
```diff
@@ -36,27 +41,41 @@ private TimeLockTestUtils() {
...
     }
 
+    private static ServerListConfig getServerListConfig(List<String> serverUris) {
+        return ImmutableServerListConfig.builder()
+                .servers(serverUris)
+                .sslConfiguration(SSL_CONFIGURATION)
+                .build();
+    }
+
+    private static Supplier<Optional<AtlasDbRuntimeConfig>> getAtlasdbRuntimeConfigSupplier(List<String> serverUris) {
```
jeremyk-91: nit: `AtlasDbRuntimeConfig`

docs/source/services/timelock_service/installation.rst
```diff
@@ -79,12 +79,12 @@ Add client(s) to Timelock
...
 9. Configure each client to use Timelock.
    Detailed documentation is :ref:`here <timelock-client-configuration>`.
-   You must remove any ``leader``, ``timestamp``, or ``lock`` blocks; the ``timelock`` block to add looks like this:
+   You must remove any ``leader``, ``timestamp``, or ``lock`` blocks; the ``timelockRuntime`` block to add looks like this:
 
 .. code-block:: yaml
 
    atlasdb:
-      timelock:
+      timelockRuntime:
```
jeremyk-91: nit: The above block should be named `atlasdbRuntime`, or it should otherwise be clear that this is a part of runtime config and not install config.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -315,7 +309,8 @@ public void batchesRequestsIfBatchingEnabled() throws InterruptedException {
 
     @Test
     public void doesNotBatchRequestsIfBatchingNotEnabled() {
-        setUpTimeLockBlockInInstallConfig();
+        when(config.namespace()).thenReturn(Optional.of(CLIENT));
```
gsheasby: consider putting this `when` in setup rather than repeating it in several tests.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -422,12 +410,6 @@ public void timeLockMigrationReportsReadyIfMigrationNotNeeded() {
...
     }
 
-    @Test
-    public void usesTimeLockIfInstallConfigIsUnspecifiedButInitialRuntimeConfigContainsTimeLockBlock() {
-        setUpTimeLockBlockInRuntimeConfig();
-        verifyUsingTimeLockByGettingAFreshTimestamp();
-    }
-
     @Test
     public void throwsIfInstallConfigHasLeaderBlockButInitialRuntimeConfigContainsTimeLockBlock() throws IOException {
```
gsheasby: Would like to know what happens if leader block is there and timelock block has zero servers.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,14 @@ develop
     *    - Type
          - Change
 
+    *    - |userbreak|
+         - The ``timelockRuntime`` block is now mandatory. Before this it was optional.
```
gsheasby: The two userbreaks can be collapsed into one IMHO. It would be good to link to the docs here too.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
gsheasby: nit: would prefer to have `getServerListConfig` below this method, for better top-down readability.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
snegrus-adb: The presence of the `timelockRuntime` block acted as that flag, before this change, right?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -746,17 +745,14 @@ private static LockAndTimestampServices createRawServicesFromTimeLock(
                 .withMigrator(migrator);
     }
 
+    //  what's the deal with all these clients
```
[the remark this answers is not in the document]
snegrus-adb: Yes, what's the deal with all these clients? It seems these addresses are used by the proxies to hit the timelock right ?

atlasdb-config/src/test/java/com/palantir/atlasdb/config/AtlasDbConfigTest.java
```diff
@@ -250,52 +193,11 @@ public void inMemoryConfigWorksWithNonTestNamespace() {
...
-                .namespace(Optional.empty())
-                .keyValueService(kvsConfig)
-                .timelock(TIMELOCK_CONFIG_WITH_OPTIONAL_EMPTY_CLIENT)
-                .build())
-                .isInstanceOf(IllegalStateException.class)
-                .satisfies((exception) ->
-                        assertThat(exception.getMessage(), containsString("TimeLock client should not be empty")));
-    }
-
-    // Documenting that for in-memory, we don't care what the timelock client is - it just has to be non-empty.
```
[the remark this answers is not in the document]
snegrus-adb: But there is a test `inMemoryConfigCanHaveEmptyNamespace` and one that `inMemoryConfigWorksWithNonTestNamespace`.

atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
```diff
@@ -281,16 +279,9 @@ private void checkLeaderAndTimelockBlocks() {
         if (leader().isPresent()) {
             Preconditions.checkState(areTimeAndLockConfigsAbsent(),
                     "If the leader block is present, then the lock and timestamp server blocks must both be absent.");
-            Preconditions.checkState(!timelock().isPresent(),
```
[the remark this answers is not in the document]
snegrus-adb: Can't test I think, but will fail when creating service from TransactionManagers.

atlasdb-config/src/test/java/com/palantir/atlasdb/config/AtlasDbConfigTest.java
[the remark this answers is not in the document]
gsheasby: ok, good 👍

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
gsheasby: OK, yes - that wasn't clear immediately, but the timelockRuntime block is there and empty.

atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
```diff
@@ -54,11 +54,10 @@
      * key value services. Currently, this only applies to external timelock services, and Cassandra KVS (where it is
      * used as the keyspace).
      *
-     * For backwards compatibility reasons, this is optional. If no namespace is specified:
-     *   - if using Cassandra, the keyspace must be explicitly specified.
-     *   - if using TimeLock, the client name must be explicitly specified.
+     * For backwards compatibility reasons, this is optional. If no namespace is specified, if using Cassandra,
```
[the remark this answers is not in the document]
sandorw: Haven't given it a ton of thought, but bundling config breaks seems reasonable.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: Yeah the corroboration check will catch this.

When we do this release we could call out that if your client was _not the same_ as your KVS keyspace or namespace, you need to contact the AtlasDB team. A special manual migration task may be needed when upgrading AtlasDB to a version that has this check.

snegrus-adb: Where should I add that announcment Should I add that in the release note?

snegrus-adb:
Not going to do this one, instead, I opened this PR to only remove the install block: https://github.com/palantir/atlasdb/pull/3394

PR #3364, opened 2018-07-09
[TS] Do not range scan punch table if timestamp is too far in the past

gmaretic:
**Goals (and why)**:
Fix the issue where targeted sweep metrics excessively range scan the punch table.

The issue should already have been fixed for most use cases, since we now do not default the last swept timestamp to -1, as we did in 0.93.0, but we now further reduce the danger by not range scanning if the timestamp could only have been punched more than a week ago.
Also, remove the misleading comment from `KeyValueServicePuncherStore` and add some docs on the issue.

Ideally, we'd do all the expensive computation for metrics asynchronously and cache instead of memoizing when the metric is requested, but that is a bit trickier to do and probably not necessary. Will open an issue to track.

KeyValueServicePuncherStore.java

Pretty high

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3364)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
@@ -23,6 +23,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.codahale.metrics.Gauge;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.atlasdb.AtlasDbMetricNames;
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStoreTest.java
```diff
@@ -109,13 +110,61 @@ public void returnsTimestampAssociatedWithGreatestPunchedTimeEvenIfItIsNotGreate
                 .isNotEqualTo(TIMESTAMP_2); // strictly speaking not needed but better for readability
     }
 
+    @Test
+    public void getMillisForTimestampIfNotPunchedBeforeWhenFirstPunchedBeforeHasLowerTimestampAndSamePunchEntry() {
```
sandorw: 😮 I'm not sure I know what this means

gmaretic:
If there are no more comments, feel free to merge and ship!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -135,7 +140,16 @@ private static Long estimateMillisSinceTs(Long timestamp, Clock clock, Function<
             if (timestamp == null) {
                 return null;
             }
-            return clock.getTimeMillis() - tsToMillis.apply(timestamp);
+            long timeBeforeRecomputing = System.currentTimeMillis();
+            long result = clock.getTimeMillis() - tsToMillis.apply(timestamp);
+
+            long timeTaken = System.currentTimeMillis() - timeBeforeRecomputing;
+            if (timeTaken > TimeUnit.SECONDS.toMillis(10)) {
```
sandorw: if this is to let us know if it should be async or not, should we drop them by a factor of 10?

gmaretic: If we ever get this, we should probably drop to one day or even 12 hours, yes. The numbers chosen were arbitrary, but as long as it takes less than a second I think we are fine

PR #3366, opened 2018-07-10
Fix internal bump

snegrus-adb:
**Goals (and why)**: Allow `BackgroundSweepThread` class to be used in tests. 

A createForTests method is added that can be used to create instances for tests

I verified in the tests where this is needed and these changes seems sufficient.

`BackgroundSweepThread.java` (the only file)

Asap

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3366)


gsheasby:
You need a `@SuppressWarnings("checkstyle:RegexpMultilineCheck")` on the VFT method - package private isn't visible enough for us.

PR #3374, opened 2018-07-11
Benchmark for ranged tombstones

gmaretic:
**Goals (and why)**:
Add a benchmark for ranged tombstones in anticipation of a DBKVS implementation that does not use the default one.

Have a single highly versioned cell and then delete all but last version.

Would be nice to get it in soon, so we get some data for comparison

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3374)


atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/RegeneratingTable.java
```diff
@@ -95,6 +100,32 @@ public void cleanup() throws Exception {
...
+        }
+
+        @Override
+        public Cell getTableCells() {
+            return cell;
+        }
+
+        private static Multimap<Cell, Value> getVersions() {
+            Multimap<Cell, Value> versions = ArrayListMultimap.create();
+            LongStream.range(1, CELL_VERSIONS + 1).boxed()
```
jeremyk-91: nit: `LongStream.rangeClosed(1, CELL_VERSIONS)` might be a little bit clearer

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/KvsDeleteBenchmarks.java
```diff
@@ -83,4 +90,11 @@ public Object allRangeDelete(ConsecutiveNarrowTable.RegeneratingCleanNarrowTable
         return doDeleteRange(table, 1);
     }
 
+    @Benchmark
+    @Threads(1)
+    @Warmup(time = 5, timeUnit = TimeUnit.SECONDS)
+    @Measurement(time = 30, timeUnit = TimeUnit.SECONDS)
```
jeremyk-91: On current postgres I get something like

```
KvsDeleteBenchmarks.deleteAllTimestamps  com.palantir.atlasdb.performance.backend.PostgresKeyValueServiceInstrumentation@192.168.99.100:5432  sample    78  202748.219 ± 12694.104  us/op
```

The count is a bit low. I guess it will improve by a lot once we have #3379 though

gmaretic: Yeah, I also expect it to run faster on a dedicated box with the postgres image

atlasdb-perf/src/main/java/com/palantir/atlasdb/performance/benchmarks/table/RegeneratingTable.java
gmaretic: 👍

PR #3377, opened 2018-07-12
Wait slightly longer for TM initialisation

gsheasby:
[no release notes]

**Goals (and why)**: The tests in SerializableTransactionManagerTest are flaky (see [build #11472](https://palantir:11472)) because on circle, it may take more than two seconds to initialise the TM.

**Concerns (what feedback would you like?)**: Does using `THREE` instead of `TWO` give us the right balance between (a) not failing when circle is slow, and (b) not taking too long to run locally (it's about 10 seconds longer to run locally)? Should we use `FlakeRule` instead/aswell?

**Where should we start reviewing?**: One file

**Priority (whenever / two weeks / yesterday)** whenever (ideally today, I guess)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3377)


gsheasby:
Closing and reopening to try and fix the CLA check

PR #3379, opened 2018-07-12
Dbkvs Ranged Tombstones Impl

gmaretic:
**Goals (and why)**:
Implement efficient ranged tombstones for Dbkvs, for targeted sweep.

Pretty straightforward.

There was decent test coverage, but added tests for edge cases.

Running benchmarks on postgres shows the expected huge perf improvement. Oracle does not seem to show improvement, so something may be wrong with the hinting in `OracleOverflowWriteTable`
 
Not super urgent, also want to get https://github.com/palantir/atlasdb/pull/3374 merged in before, so we can see the benchmarking delta

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3379)


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractKeyValueServiceTest.java
```diff
@@ -1240,32 +1238,63 @@ public void deleteTimestampRangesIncludingSentinelsDeletesMultipleColumnsAcrossM
...
+        keyValueService.addGarbageCollectionSentinelValues(TEST_TABLE, ImmutableSet.of(TEST_CELL));
+        keyValueService.putWithTimestamps(TEST_TABLE, ImmutableMultimap.of(
+                TEST_CELL, Value.create(value0_t0, minTs),
+                TEST_CELL, Value.create(value0_t1, TEST_TIMESTAMP - 1),
+                TEST_CELL, Value.create(value0_t5, TEST_TIMESTAMP)));
+
+        keyValueService.deleteAllTimestamps(TEST_TABLE, ImmutableMap.of(TEST_CELL, TEST_TIMESTAMP), true);
+
+        assertThat(getAllTimestampsForTestCell(), contains(TEST_TIMESTAMP));
+    }
```
jeremyk-91: OK - tests make sense.

tboam:
Do we understand why the results are bi-modal?  Is the DB caching the queries?  Is something in Java getting compiled and so runs faster?

jeremyk-91:
@tboam I don't think we understand it yet. I have some bandwidth to run a few experiments today. 

**Caching**: We can validate this - if there is caching and we run the benchmark for longer, we shoulud expect to see the average time and the proportion of queries that are in the 'slow' group to decrease.

**JIT compilation**: Possible, though as far as I know the time taken in Atlas code is very small (it builds the SQL query and sends it over - we don't actually receive much other than the number of deleted elements (`9999`). I've also run the raw query a couple times in SQLPlus and have seen both fast and slow queries (e.g. <20ms to >2.7s).

gmaretic:
@berler do you know why we would be seeing this behaviour on Oracle? Or do you have a suggestion on how to do things differently?

gmaretic:
Should we merge this in as-is and reevaluate if needed later?

PR #3386, opened 2018-07-18
Fix Benchmarks by dagger using a synchronous unlocker

gmaretic:
The asynchronous unlocker can cause a race condition in sweep benchmarks because we run sweep immediately after committing the transactions, so every once in a while the sweep timestamp is too low and we end up failing an assertion because we sweep the wrong number of things.

The fix is to use a synchronous unlocker in the dagger module which benchmarks use

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3386)

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
@@ -44,21 +44,23 @@ public class TimeLockClient implements AutoCloseable, TimelockService {
 
     private final TimelockService delegate;
     private final LockRefresher lockRefresher;
-    private final AsyncTimeLockUnlocker asyncUnlocker;
+    private final TimeLockUnlocker unlocker;
 
     public static TimeLockClient createDefault(TimelockService timelockService) {
-        ScheduledExecutorService refreshExecutor = createSingleThreadScheduledExecutor("refresh");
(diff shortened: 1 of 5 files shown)


lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
```diff
@@ -44,21 +44,23 @@
...
-        ScheduledExecutorService refreshExecutor = createSingleThreadScheduledExecutor("refresh");
-        LockRefresher lockRefresher = new LockRefresher(refreshExecutor, timelockService, REFRESH_INTERVAL_MILLIS);
         ExecutorService asyncUnlockExecutor = createSingleThreadScheduledExecutor("async-unlock");
         AsyncTimeLockUnlocker asyncUnlocker = new AsyncTimeLockUnlocker(timelockService, asyncUnlockExecutor);
-        return new TimeLockClient(timelockService, lockRefresher, asyncUnlocker);
+        return new TimeLockClient(timelockService, createLockRefresher(timelockService), asyncUnlocker);
+    }
+
+    public static TimeLockClient withSynchronousUnlocker(TimelockService timelockService) {
+        return new TimeLockClient(timelockService, createLockRefresher(timelockService), timelockService::unlock);
```
jeremyk-91: Fair enough - pretty clever (maybe a bit too clever, but I think it's fine)

PR #3391, opened 2018-07-19
Update gradle wrapper

gmaretic:
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3391)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -69,7 +69,9 @@ def get_last_release_from_gradle():
   We parse the output of ./gradlew printLastVersion in order to get the current
   version of the repository, and then split off the commit hash. We are assuming
   that this file is two directories up from the codebase root, and the version
-  is the line after ':printVersion' in the stdout of the gradle command.
+  is the first expression matching the format of number.number.number[.-+alphanum]*
+  in the stdout of the gradle command. This is a superset of valid semver strings
+  but we do not always conform to semver, e.g., "0.94.0.dirty".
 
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -69,22 +69,24 @@ def get_last_release_from_gradle():
...
-  import subprocess, os
+  import subprocess, os, re
   conf_py_directory = os.path.dirname(os.path.realpath(__file__))
   codebase_root = os.path.join(conf_py_directory, '..', '..')
   p = subprocess.Popen(['./gradlew', 'printLastVersion'],
                        cwd=codebase_root, stdout=subprocess.PIPE)
   stdout = p.communicate()[0]
-  lines = stdout.decode().split('\n')
-  return lines[lines.index(':printLastVersion') + 1]
+  matcher = re.search('\d+\.\d+\.\d+[\.\-\+\w]*', stdout)
```
jeremyk-91: tested:
```
1.5
1.5.7
1.5.7-rc2
1.5.7.dirty
1.5.7-1-g123cab
1.5.7.dirty-9-g123456
```

work as expected.

A bit unfortunate we might get a false positive (e.g. ip address), but given this  is just a script, fine

sandorw: I believe that I usually see the `.dirty` on the end.

jeremyk-91: Fair point - yeah it should be the other way round. In any case I'm confident that case is correct.

PR #3394, opened 2018-07-19
Remove timelock install block

snegrus-adb:
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3394)


sandorw:
Ran a quick experiment at https://github.palantir.build/foundry/salt/pull/1272.

We may not be able to expect a build failure from removing the block; some teams may get a runtime error upon upgrade. Teams which have the common startup check (either in-memory KVS or using an install or runtime timelock block) will get a compile error, but some teams may not have this check (e.g. Salt IIRC).

sandorw:
We may also get runtime upgrade failures for anyone that has overridden their service to be HA, including the addition of an install block. For example (haven't dug into how it interacts with the product's config itself): https://github.palantir.build/deployment-cdconfig/papaya/pull/1739/files

sandorw:
Two edge cases:
- We have lots of places where we test with in-memory and not with actual Timelock, which means that a broken Timelock install config becomes a runtime error rather than failing the build. Some teams will avoid this if they have the common check for `in-memory || install.timelock`.
- Anywhere where we've overridden the config to specify Timelock usage might break, since it could be adding the install block.

sandorw:
Closing the PR. The branch is protected for when we revisit. @felixdesouza for SA

PR #3399, opened 2018-07-23
[TS] Targeted Sweep Outcome Metrics

gmaretic:
**Goals (and why)**:
Better overview of what's happening

Mostly just adapting from the legacy metrics

Generic test that there are no conflicts between the metrics and no conflicts with the existing legacy metrics.

Did not use all of the outcomes because not all of them make sense. Are all of the used ones sensible?

`SweepOutcomeMetrics.java`

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3399)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepBatch.java
```diff
@@ -30,6 +30,10 @@
     List<WriteInfo> writes();
     long lastSweptTimestamp();
 
+    default boolean hasNothingToSweep() {
```
tboam: nit: You can standardise the code a bit by calling this `isEmpty` and just deferring to `writes().isEmpty()`.  I think this reads slightly nicer when it is used too.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/SweepOutcomeMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+        return metrics;
+    }
+
+    public static SweepOutcomeMetrics registerTargeted(MetricsManager metricsManager) {
+        SweepOutcomeMetrics metrics = new SweepOutcomeMetrics();
+        metrics.registerMetric(metricsManager, TARGETED_OUTCOMES, TargetedSweepMetrics.class);
+        return metrics;
+    }
+
+    private <T> void registerMetric(MetricsManager manager, List<SweepOutcome> outcomes, Class<T> forClass) {
```
tboam: Do we have to register metrics on a class name?  We're finding this is quite brittle and expensive.  If we refactor some code we can inadvertently break our monitors and dashboards.

gmaretic: Offline: decided not to change this at the moment, but should consider in the future.

PR #3407, opened 2018-07-24
Cleanup AutoDelegate uses

gmaretic:
**Goals (and why)**:
Clean up uses of AutoDelegate to make things easier to follow

AutoDelegate annotations are now placed on the class we are extending. Also fix up STM situation.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3407)


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -23,8 +23,10 @@
 import com.palantir.lock.LockRequest;
 import com.palantir.lock.LockService;
 import com.palantir.lock.v2.TimelockService;
+import com.palantir.processors.AutoDelegate;
 import com.palantir.timestamp.TimestampService;
 
+@AutoDelegate(typeToExtend = TransactionManager.class)
```
sandorw: Sounds good, I like this solution much more than having to put it on just one extending class, although it does read strangely since I'm used to the other way.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,7 @@ develop
     *    - Type
          - Change
 
+=======
```
sandorw: What's this separator doing?

gmaretic: The release notes got messed up, so just reset it to the develop version

PR #3410, opened 2018-07-25
More efficient scans for sweep priority

sandorw:
**Goals (and why)**:
The existing implementation scans over all versions of each row on Cassandra, which places a lot of memory pressure on the DB.

* Scan instead on a single column to fetch the rows, then directly fetch all of the columns

Existing test coverage mainly concerns the correctness of the store implementation and doesn't hit scale, which is fine. We would ideally get some signal that this change improves the range scan behavior, but we know that the same pattern works elsewhere and has resulted in significant perf gains.

A better approach would be to know all the table names up front, but this is tricky for the few cases where people do dynamic table creation at runtime. This may be enough of an improvement that we don't need to look at the direct table access solution.

Should we move the extra logic from `SweepHistoryProvider` into the store?

`SweepPriorityStoreImpl`

Soon

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3410)


atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/SweepStatsKeyValueService.java
```diff
@@ -84,8 +82,7 @@
 
     private final Multiset<TableReference> writesByTable = ConcurrentHashMultiset.create();
 
-    private final Set<TableReference> clearedTables = Collections.newSetFromMap(
-            new ConcurrentHashMap<TableReference, Boolean>());
+    private final Set<TableReference> clearedTables = Sets.newConcurrentHashSet();
```
sandorw: Exact same thing underneath the hood at the moment, but we could pick up future improvements automatically, and this prevents someone else from digging in to the same thing.

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/generated/SweepPriorityTable.java
```diff
@@ -1054,20 +1054,69 @@ public void delete(Iterable<SweepPriorityRow> rows) {
...
     }
 
-    public BatchingVisitableView<SweepPriorityRowResult> getAllRowsUnordered() {
-        return getAllRowsUnordered(allColumns);
-    }
-
-    public BatchingVisitableView<SweepPriorityRowResult> getAllRowsUnordered(ColumnSelection columns) {
-        return BatchingVisitables.transform(t.getRange(tableRef, RangeRequest.builder().retainColumns(columns).build()),
-                new Function<RowResult<byte[]>, SweepPriorityRowResult>() {
+    public BatchingVisitableView<SweepPriorityRowResult> getRange(RangeRequest range) {
```
sandorw: Checked in generated code :'(

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/SweepHistoryProvider.java
```diff
@@ -19,16 +19,23 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import com.palantir.atlasdb.keyvalue.api.RangeRequest;
 import com.palantir.atlasdb.schema.generated.SweepPriorityTable;
+import com.palantir.atlasdb.schema.generated.SweepPriorityTable.SweepPriorityNamedColumn;
 import com.palantir.atlasdb.schema.generated.SweepTableFactory;
 import com.palantir.atlasdb.transaction.api.Transaction;
 
 class SweepHistoryProvider {
```
sandorw: Should we move this into the store class?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityStoreImpl.java
```diff
@@ -101,7 +107,18 @@ private void tryInitialize() {
...
+        // Load a single column first for each row. This is a much more efficient query on Cassandra
+        // than the full table scan that occurs otherwise.
+        List<SweepPriorityRowResult> rows = table.getRange(RangeRequest.builder()
+                .retainColumns(SweepPriorityTable.getColumnSelection(SweepPriorityNamedColumn.CELLS_DELETED))
+                .batchHint(READ_BATCH_SIZE)
+                .build())
+                .immutableCopy();
+
+        // Fetch all columns for the above rows directly
+        return Lists.transform(table.getRows(Lists.transform(rows, SweepPriorityRowResult::getRowName)),
```
jeremyk-91: There is a small trade-off here in that we now make 2 requests to the underlying KVS, so this might not be faster for small tables, which we'd hope is usually the case (the product that spawned the ticket that needed this fix is an exception). It's probably fine though since a little bit slower on something that's already fast is usually OK

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/SweepHistoryProvider.java
```diff
@@ -19,16 +19,23 @@
...
+
     Map<String, Long> getHistory(Transaction tx) {
         Map<String, Long> tableToLastTimeSwept = new HashMap<>();
         SweepPriorityTable sweepPriorityTable = SweepTableFactory.of().getSweepPriorityTable(tx);
-        sweepPriorityTable.getAllRowsUnordered(SweepPriorityTable.getColumnSelection(
-                SweepPriorityTable.SweepPriorityNamedColumn.LAST_SWEEP_TIME))
+        sweepPriorityTable.getRange(RangeRequest.builder()
+                .retainColumns(SweepPriorityTable.getColumnSelection(SweepPriorityNamedColumn.LAST_SWEEP_TIME))
+                .batchHint(READ_BATCH_SIZE)
+                .build())
```
jeremyk-91: 👍 for batching

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/compact/SweepHistoryProvider.java
jeremyk-91: This is used by the `CompactPriorityCalculator` (which is used in Oracle at least to determine whether to try to run shrink space etc. on tables) - the idea was that you want to compact something only after you've swept it since the last compaction, because you'll have tombstones that you know can be dropped (based on the number of cells deleted).

If the only user of the sweep history was the background sweeper I'd say it probably should be merged, but given this usage it may be cleaner this way.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityStoreImpl.java
```diff
@@ -101,7 +107,18 @@ private void tryInitialize() {
 
     private List<SweepPriority> loadPriorities(Transaction tx) {
         SweepPriorityTable table = sweepTableFactory.getSweepPriorityTable(tx);
-        return table.getAllRowsUnordered().transform(SweepPriorityStoreImpl::hydrate).immutableCopy();
+
+        // Load a single column first for each row. This is a much more efficient query on Cassandra
+        // than the full table scan that occurs otherwise.
+        List<SweepPriorityRowResult> rows = table.getRange(RangeRequest.builder()
+                .retainColumns(SweepPriorityTable.getColumnSelection(SweepPriorityNamedColumn.CELLS_DELETED))
```
jeremyk-91: This is probably a good call in terms of optimisation, as it's almost certainly the smallest one.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityStoreImpl.java
sandorw: Yep, and reads from this table shouldn't ever be blocking or performance critical.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/SweepPriorityStoreImpl.java
sandorw: That was my guess, although it's probably not a big difference overall.

PR #3412, opened 2018-07-25
Bump default max number of clients to 200

samrogerson:
**Goals (and why)**: It's not uncommon to exhaust the number of clients in previous default configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3412)


jeremyk-91:
Hey @samrogerson, 

Thanks for the contribution! Unfortunately I think we were working on this in parallel. Closing out as this change is already in #3413.

I realise I said 200 on the ticket, but that's probably a bit conservative, even - the DOS attacks we're worried about are more for O(tens of thousands) of clients.

PR #3413, opened 2018-07-25
[TimeLock] Bump Maximum Clients & Saturation Metrics

jeremyk-91:
**Goals (and why)**:
- TimeLock has a max number of clients, which used to be 100 by default. However, internal usage of Atlas has grown to the point where 100 isn't enough.
- Also, TimeLock doesn't have a good way of telling you you're about to breach the threshold until you do, and then it hard-fails.

- Bump client numbers to 500
- Add metrics for the number of active clients and maximum number of clients

- The number is referenced as a constant (thankfully) so it shouldn't affect existing tests.
- Metrics is a new feature. I've written a pair of tests that verify that the metrics are updated when clients are registered or the config changes respectively.

- Nothing in particular. Is the live reloading correct?
- The tests are a little janky
- Is 500 too high?

**Where should we start reviewing?**: TimeLockResource.java

**Priority (whenever / two weeks / yesterday)**: today. It would have been nice to have caught this before it actually became a problem.

@samrogerson for SA - @tboam and I were actually discussing this this morning!

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3413)

### DIFF
--- a/timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimeLockResourceTest.java
+++ b/timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimeLockResourceTest.java
@@ -25,7 +25,6 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 
 import java.util.Optional;
-import java.util.StringJoiner;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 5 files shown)


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimeLockResourceTest.java
```diff
@@ -113,4 +156,24 @@ private String uniqueClient() {
...
+
+    private void assertMaxClientsIs(int expected) {
+        assertThat(getGaugeValueForTimeLockResource(TimeLockResource.MAX_CLIENTS))
+                .isEqualTo(expected);
+    }
+
+    private int getGaugeValueForTimeLockResource(String gaugeName) {
+        Object value = Optional.ofNullable(metricsManager.getRegistry()
+                .getGauges()
+                .get(new StringJoiner(".").add(TimeLockResource.class.getCanonicalName())
```
tboam: Surely String concatenation is cleaner than StringJoiner in this instance?

jeremyk-91: Yeah, that's true 🤦‍♂️ 
I think when I wrote the first part I was thinking I might need to string a bunch of pieces together, but `getCanonicalName()` already has most of the things in order for you.

jeremyk-91:
I've filed an internal issue to track setting up a monitor once this metric starts going out to the field

j-baker:
internally, this should probably be a health check and not a monitor - easier to manage

PR #3414, opened 2018-07-25
optional validation of immutable ts on read operations

gokcanozakdag:
**Goals (and why)**:
Addressing this closed pr: https://github.com/palantir/atlasdb/pull/2996
Making lock validations optional for read operations.

Adding a configurable flag called `validateLocksOnReads` to TransactionManagers. It is set to `true` by default to conserve previous behavior. Validation after a read operation is done only if table is thoroughly swept and this flag is set to true. Validation at commit time is done if table is thoroughly swept, independent of `validateLocksOnReads` set to true or false. 

Adding two more test cases.
- First test confirms that setting `validateLocksOnReads` to true causes validation to run after a read operation, even if the transaction is not committed. 
- Second test case sets `validateLocksOnReads` to false. This time immutable ts locks are not checked after a read operation; but it is checked when we commit the transaction.

I also tried to refactor tests a bit by creating a private method to generate SnapshotTransaction, as it is constructor is taking large number of parameters, and making it hard to read the test case. It still doesn't look great, maybe we can add a builder to be used only in tests?

- Setting `validateLocksOnReads` to false will cause giving up on idempotent read operations. Clients can have two different reads for same value in a transaction. This cannot cause a correctness problem, as transaction will fail at commit time; the way the transaction fails will change.
- This adds yet another parameter to constructor of SnapshotTransaction, and TransactionManager :(

`SnapshotTransaction.java`

1 week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3414)


sandorw:
If we want to do that, we should wrap the `task.execute(txn)` call in `AbstractTransactionManager`, since we don't want to catch exceptions from the commit phase - this change only affects behavior during the task anyway, since it affects reads within the transaction. This way, if your code throws an IAE/ISE/NPE due to inconsistencies in the data it read because the lock checks were removed, it can be caught, the locks checked, and treated as a retriable exception instead.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -747,13 +750,17 @@ private static boolean isSingleton(Iterable<?> elements) {
         };
     }
 
-    private void validateExternalAndCommitLocksIfNecessary(TableReference tableRef, long timestamp) {
-        if (!isValidationNecessary(tableRef)) {
+    private void validateExternalAndCommitLocksOnReadIfNecessary(TableReference tableRef, long timestamp) {
```
jeremyk-91: nit: can we rename this to `validatePreCommitRequirementsOnReadIfNecessary`? Mainly because this checks the pre-commit conditions too, which may not be 'external locks'

PR #3420, opened 2018-07-26
Changes to targeted sweep for large internal product

gmaretic:
**Goals (and why)**:
These changes should make it easier to wire up targeted sweep in our large internal product

TargetedSweeper now implements BackgroundSweeper, and SweepQueue implements MultiTableSweepQueueWriter, so we don't have to jump through hoops.

Existing tests should be fine.

Are these classes slowly but surely turning into a mess?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3420)


gmaretic:
@jeremyk-91 I'd like to get these in today, so lmk if you don't have bandwidth

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
```diff
@@ -67,6 +68,15 @@ public static SweepQueue create(TargetedSweepMetrics metrics, KeyValueService kv
         return new SweepQueue(cells, timestamps, progress, deleter, cleaner, shards, metrics);
     }
 
+    /**
+     * Creates an instance of SweepQueue that should be used only for writing to the sweep queue and does not implement
+     * the deleting logic.
+     */
+    public static MultiTableSweepQueueWriter createWriter(TargetedSweepMetrics metrics, KeyValueService kvs,
+            Supplier<Integer> shardsConfig) {
+        return create(metrics, kvs, shardsConfig, new TargetedSweepFollower(ImmutableList.of(), null));
```
jeremyk-91: This is safe (because the TM in the follower should only be invoked if a cleanup task is actually invoked, and there are none), though would be nice if we could make this clearer.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -113,29 +126,25 @@ public static TargetedSweeper createUninitializedForTest(Supplier<Integer> shard
...
         isInitialized = true;
     }
 
     @Override
-    public void initialize(TransactionManager txManager) {
-        initialize(SpecialTimestampsSupplier.create(txManager),
-                txManager.getTimelockService(),
-                txManager.getKeyValueService(),
-                new TargetedSweepFollower(followers, txManager));
+    public void runInBackground() {
```
jeremyk-91: nit: does it make sense to add javadocs for this?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -163,11 +172,16 @@ public void sweepNextBatch(ShardAndStrategy shardStrategy) {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
         conservativeScheduler.close();
         thoroughScheduler.close();
```
jeremyk-91: random thought: if the conservative scheduler throws a runtime exception then the thorough scheduler is never closed. Not a regression, and non-blocking.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -49,7 +48,7 @@
 import com.palantir.logsafe.SafeArg;
 
 @SuppressWarnings({"FinalClass", "Not final for mocking in tests"})
-public class TargetedSweeper implements MultiTableSweepQueueWriter {
+public class TargetedSweeper implements MultiTableSweepQueueWriter, BackgroundSweeper {
```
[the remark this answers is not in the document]
gmaretic: Good point, I think we should do this, but in a separate PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
gmaretic: Cleaned it up, lmk what you think

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
gmaretic: Will fix in a separate PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
```diff
@@ -35,46 +34,35 @@
 
 public final class SweepQueue implements MultiTableSweepQueueWriter {
     private static final Logger log = LoggerFactory.getLogger(SweepQueue.class);
-    private final SweepableCells sweepableCells;
-    private final SweepableTimestamps sweepableTimestamps;
     private final ShardProgress progress;
+    private final SweepQueueWriter writer;
+    private final SweepQueueReader reader;
     private final SweepQueueDeleter deleter;
     private final SweepQueueCleaner cleaner;
```
jeremyk-91: nice, I like the separation here

Issue #3422, opened 2018-07-26
Stream store cleanup tasks unnecessarily demand all rows


jeremyk-91:
```
    @Override
    public boolean cellsCleanedUp(Transaction t, Set<Cell> cells) {
        ValueStreamIdxTable usersIndex = tables.getValueStreamIdxTable(t);
        Set<ValueStreamIdxTable.ValueStreamIdxRow> rows = Sets.newHashSetWithExpectedSize(cells.size());
        for (Cell cell : cells) {
            rows.add(ValueStreamIdxTable.ValueStreamIdxRow.of((Long) ValueType.VAR_LONG.convertToJava(cell.getRowName(), 0)));
        }
        Multimap<ValueStreamIdxTable.ValueStreamIdxRow, ValueStreamIdxTable.ValueStreamIdxColumnValue> rowsInDb = usersIndex.getRowsMultimap(rows);
        Set<Long> toDelete = Sets.newHashSetWithExpectedSize(rows.size() - rowsInDb.keySet().size());
        for (ValueStreamIdxTable.ValueStreamIdxRow rowToDelete : Sets.difference(rows, rowsInDb.keySet())) {
            toDelete.add(rowToDelete.getId());
        }
        ValueStreamStore.of(tables).deleteStreams(t, toDelete);
        return false;
    }
```

We only use existence, so a suitably deployed getRowsColumnRange should be enough, rather than getting the world.

PR #3423, opened 2018-07-27
getRowsColumnRange for IndexCleanupTasks

jeremyk-91:
**Goals (and why)**:
- Stream store index cleanup tasks read the whole row, which can be wide in certain cases, to check if _any_ reference to the stream remains (which gates deleting the actual stream blocks). This is inefficient; reading one column should suffice.

- Use `getRowsColumnRange()` instead of `getRowsMultimap()` to load one column per row
- Regenerate schemas.

- This code path is tested through `TargetedSweepEteTest` which checks deletes are correctly propagated.
- @snegrus-adb and I were working on an ETE test for the multitenancy, but that proved to be quite unwieldy.

- Is this actually more performant?

**Where should we start reviewing?**: Pick any one of the index cleanup tasks, which shows what the renderer is trying to do.

**Priority (whenever / two weeks / yesterday)**: Whenever. The heart of the issue that spawned this was with a downstream product abusing the stream store; this change makes Atlas more resilient to badness and possibly more performant in certain internal use cases as well.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3423)


sandorw:
Fixes https://github.com/palantir/atlasdb/issues/3422

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -684,9 +686,26 @@ private void cellsCleanedUp() {
...
+
+                    line("BatchColumnRangeSelection oneColumn = BatchColumnRangeSelection.create(");
+                    line("        PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 1);");
+                    line("Map<", StreamIdxRow, ", BatchingVisitable<", StreamIdxColumnValue, ">> existentRows");
+                    line("        = usersIndex.getRowsColumnRange(rows, oneColumn);");
+
+                    line("Set<", StreamIdxRow, "> rowsInDb = Sets.newHashSetWithExpectedSize(cells.size());");
+
+                    line("for (Map.Entry<", StreamIdxRow, ", BatchingVisitable<", StreamIdxColumnValue, ">> rowVisitable");
+                    line("        : existentRows.entrySet())"); {
```
tpetracca: nit: can we wrap the for loop content with brackets?

tpetracca: also not sure if this is affecting spacing, but the next line isn't being indented like I'd expect it to

jeremyk-91: Good call - the indentation is taken care of by the braces in the raw code-strings being there or not there. I left this out by accident (but the tests still passed, because the scope of the for loop happened to be 1 statement. New version should be OK (see generated code)

sandorw:
@jeremyk-91 What's the status for this? In particular, do we need to investigate the performance impact before merging? Does this need an additional review?

jeremyk-91:
Ugh, very old PR. I'm not aware of any obstacles to merging this. At the time of writing this PR Stefan and I ran a bunch of perf experiments - there was a little bit of overhead in the 1-ref case, and a substantial perf gain in the many-refs case.

I think this can be merged unless anyone has strong objections.

PR #3429, opened 2018-07-31
Increase sweep sleep if nothing to sweep

tboam:
**Goals (and why)**: Reduce calls to sweep priority and other legacy sweep activity if there's nothing to do.  With the move to targeted sweep products will reach a point where there's nothing left for sweep to do.

**Implementation Description (bullets)**: When we get the status NOTHING_TO_SWEEP we wait longer between iterations.

**Testing (What was existing testing like?  What have you done to improve it?)**: No tests for this, but is a small change.

**Concerns (what feedback would you like?)**: Is 10 minutes the right amount of time.  I wanted to balance not looping too often with not having to wait too long if something changes (e.g. someone unblacklists a table).

**Where should we start reviewing?**: atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java

**Priority (whenever / two weeks / yesterday)**: Whenever

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3429)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
@@ -219,7 +219,7 @@ public class BackgroundSweepThread implements Runnable {
     }
 
     private long getBackoffTimeWhenNothingToSweep() {
-        return 5 * getBackoffTimeWhenSweepHasNotRun(); // 10 minutes by default
+        return TimeUnit.MINUTES.toMillis(10);
     }
 
     @VisibleForTesting
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
@@ -16,6 +16,7 @@
 
 package com.palantir.atlasdb.sweep;
 
+import java.time.Duration;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
@@ -103,7 +104,7 @@ public class BackgroundSweepThread implements Runnable {
         try (SingleLockService locks = createSweepLocks()) {
             // Wait a while before starting so short lived clis don't try to sweep.
             waitUntilSpecificTableSweeperIsInitialized();
-            sleepForMillis(getBackoffTimeWhenSweepHasNotRun());
+            sleepFor(getBackoffTimeWhenSweepHasNotRun());
             log.info("Starting background sweeper with thread index {}", SafeArg.of("threadIndex", threadIndex));
             while (true) {
                 // InterruptedException might be wrapped in RuntimeException (i.e. AtlasDbDependencyException),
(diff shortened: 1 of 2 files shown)


docs/source/release_notes/release-notes.rst
```diff
@@ -53,6 +53,12 @@ develop
     *    -
          -
 
+    *    - |improved|
+         - Sequential sweep now sleeps longer between iterations if there was nothing to sweep.
+           Previously we would sleep for 2 minutes between runs, but it is unlikely that anything has changed dramatically in 2 minutes so we sleep for longer to prevent scanning the sweep priority table too often.  Going forward the most likely explanation for there being nothing to sweep is that we have switched to targeted sweep.
+           We don't stop completely or sleep for too long just in case configuration changes and a table is eligible to sweep again.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/XXXX>`__)
```
sandorw: 3429

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -213,7 +215,11 @@ private SweepOutcome grabLocksAndRun(SingleLockService locks) throws Interrupted
     }
 
     private long getBackoffTimeWhenSweepHasNotRun() {
-        return 20 * (1000 + sweepPauseMillis.get());
+        return 20 * (1000 + sweepPauseMillis.get()); // 2 minutes by default
```
sandorw: I've always been slightly confused by this - is this only used for the pause before we start sweep? If so, it might be better to defined the backoff time you're adding completely separately from the definition here, unless it should always be 5x this one.

tboam: The sweep pause millis is the pause between each iteration of sweep, we have a small pause (5 seconds by default) to prevent us hammering the database and we have increased this when a particular DB was under load or decreased it when we knew we had a lot of work to do.

The maths here is to ensure that we always pause for some amount of time when there's a problem in the hope it will clear.

This being 20x the sweep pause, and the new delay being 5x are kind of arbitrary but give sensible values for the range we expect for the sweep pause.  I can add further configuration values or just hardcode the new sleep to 10 minutes.

sandorw: I think hardcoding makes sense - the two don't seem linked to each other. We shouldn't need configuration for this. Legacy sweep has always been quite async and the sweep priority scans should be cheap enough that they are not a problem (or we should fix them further).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/BackgroundSweepThread.java
```diff
@@ -176,11 +177,13 @@ private void updateMetrics(SweepOutcome outcome) {
     }
 
     private void sleepUntilNextRun(SweepOutcome outcome) throws InterruptedException {
-        long sleepDurationMillis = getBackoffTimeWhenSweepHasNotRun();
+        Duration sleepDuration = getBackoffTimeWhenSweepHasNotRun();
```
sandorw: nit: might be cleaner as a switch statement rather than splitting out the functions

PR #3431, opened 2018-08-01
Expose Timelock client numbers so they can be used by health checks

tboam:
**Goals (and why)**: We want to create a health check for when the number of clients gets close to the configured maximum.  Exposing the values on the TimeLockAgent makes this easier.

**Implementation Description (bullets)**: Add getters, these are currently unused but will be used internally.

**Testing (What was existing testing like?  What have you done to improve it?)**: Adding getters so don't see value in testing.

**Concerns (what feedback would you like?)**: None

**Where should we start reviewing?**: timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java

**Priority (whenever / two weeks / yesterday)**: Not urgent.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3431)

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java
@@ -89,8 +89,12 @@ public class TimeLockResource {
         return servicesByNamespace.computeIfAbsent(namespace, this::createNewClient);
     }
 
-    public int numberOfClients() {
+    public int getNumberOfActiveClients() {
         return servicesByNamespace.size();
+    }
+
+    public int getMaxNumberOfClients() {
+        return maxNumberOfClients.get();
     }
 
     private TimeLockServices createNewClient(String namespace) {
@@ -99,7 +103,7 @@ public class TimeLockResource {
                         + "used.",
                 PaxosTimeLockConstants.LEADER_ELECTION_NAMESPACE);
 
-        if (numberOfClients() >= maxNumberOfClients.get()) {
+        if (getNumberOfActiveClients() >= maxNumberOfClients.get()) {
             log.error(
                     "Unable to create timelock services for client {}, as it would exceed the maximum number of "
                             + "allowed clients ({}). If this is intentional, the maximum number of clients can be "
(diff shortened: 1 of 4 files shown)


tboam:
Assigning Jeremy as he laid the groundwork for this change.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java
```diff
@@ -89,17 +89,21 @@ TimeLockServices getOrCreateServices(String namespace) {
...
+    }
+
     private TimeLockServices createNewClient(String namespace) {
         Preconditions.checkArgument(!namespace.equals(PaxosTimeLockConstants.LEADER_ELECTION_NAMESPACE),
                 "The client name '%s' is reserved for the leader election service, and may not be "
                         + "used.",
                 PaxosTimeLockConstants.LEADER_ELECTION_NAMESPACE);
 
-        if (numberOfClients() >= maxNumberOfClients.get()) {
+        if (getNumberOfActiveClients() >= maxNumberOfClients.get()) {
```
jeremyk-91: nit: We should standardise on using the public methods here (so we should use `getMaxNumberOfClients()`)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java
```diff
@@ -89,17 +89,21 @@ TimeLockServices getOrCreateServices(String namespace) {
...
             log.error(
                     "Unable to create timelock services for client {}, as it would exceed the maximum number of "
                             + "allowed clients ({}). If this is intentional, the maximum number of clients can be "
```
jeremyk-91: Not a part of the original change - but could we change the `maxNumberOfClients` in the `SafeArg` below to be `getMaxNumberOfClients()`? Right now it looks like we log the supplier, which isn't helpful

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockResource.java
```diff
@@ -89,17 +89,21 @@ TimeLockServices getOrCreateServices(String namespace) {
         return servicesByNamespace.computeIfAbsent(namespace, this::createNewClient);
     }
 
-    public int numberOfClients() {
+    public int getNumberOfActiveClients() {
```
jeremyk-91: Hmm, I think this broke some of the timelock tests. Should be an easy fix though

tboam: D'oh.  This is now fixed.

Issue #3426, opened 2018-07-30
Conflict in getRows implementation that uses an ImmutableSortedMap.builder


jeremyk-91:
For internal users: please see https://pl.ntr/tM for more detail.

There is a scenario in filtering the row results for `getRows` which can cause an exception. We inadvertently broke behaviour here, because a `SortedMap` (SM) supports being put to multiple times, but an `ImmutableSortedMap.Builder` (ISM.B) does not.

i.e.
```java
map.put(k1, v1);
map.put(k1, v2); // throws with an ISM.B
// SM has state {k1: v2}
```

So somewhere in that path we have the same key being put more than once.

PR #3445, opened 2018-08-13
Filter the keys added to results in getRows

snegrus-adb:
**Goals (and why)**: Fix https://github.com/palantir/atlasdb/issues/3426

**Implementation Description (bullets)**: Keep the keys added to results and filter on return.

**Testing (What was existing testing like?  What have you done to improve it?)**: not sure.

**Concerns (what feedback would you like?)**: is this the best solution?

**Where should we start reviewing?**: single file.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3445)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1121,7 +1121,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             }
         }
         Set<Cell> keysAddedToResults = keysAddedBuilder.build();
-        count.set(keysAddedToResults.size());
+        count.addAndGet(keysAddedToResults.size());
 
         if (!keysToDelete.isEmpty()) {
             // if we can't roll back the failed transactions, we should just try again
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1126,19 +1126,23 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
         if (!keysToDelete.isEmpty()) {
             // if we can't roll back the failed transactions, we should just try again
             if (!rollbackFailedTransactions(tableRef, keysToDelete, commitTimestamps, defaultTransactionService)) {
-                rawResults.keySet().removeAll(keysAddedToResults);
-                return rawResults;
+                return getRemainingResults(rawResults, keysAddedToResults);
             }
         }
 
         if (!keysToReload.isEmpty()) {
             Map<Cell, Value> nextRawResults = keyValueService.get(tableRef, keysToReload);
             validateExternalAndCommitLocksIfNecessary(tableRef, getStartTimestamp());
-            nextRawResults.keySet().removeAll(keysAddedToResults);
-            return nextRawResults;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1114,24 +1114,27 @@ private int estimateSize(List<RowResult<Value>> rangeRows) {
...
-                        found++;
                         results.put(key, transformer.apply(value));
+                        keysAddedBuilder.add(key);
                     }
                 }
             }
         }
-        count.set(found);
+        Set<Cell> keysAddedToResults = keysAddedBuilder.build();
+        count.set(keysAddedToResults.size());
```
jeremyk-91: This will change the size of `count` across iterations, because we're modifying `rawResults` across calls.
I think `count.accumulateAndGet(c -> c + keysAddedToResults.size())` should work.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1114,24 +1114,27 @@ private int estimateSize(List<RowResult<Value>> rangeRows) {
...
             }
         }
-        count.set(found);
+        Set<Cell> keysAddedToResults = keysAddedBuilder.build();
+        count.addAndGet(keysAddedToResults.size());
 
         if (!keysToDelete.isEmpty()) {
             // if we can't roll back the failed transactions, we should just try again
             if (!rollbackFailedTransactions(tableRef, keysToDelete, commitTimestamps, defaultTransactionService)) {
+                rawResults.keySet().removeAll(keysAddedToResults);
```
jeremyk-91: So... this is a little tricky, because it seems to me at least that `rawResults` may be an ImmutableMap and so a removeAll operation wouldn't be supported.

https://github.com/palantir/atlasdb/blob/8f526dd10790601199292f948dff3507d5830b68/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java#L433

passed into getRowColumnRangePostFiltered
passed into this.

---

One way is to return a view of the map by doing filterKeys on the key not being in this set. If we need to retry a lot that may not be such a good idea. Alternatives could be
1. take a mutable copy at all call-sites (which is fine, because this method is private)
2. make a copy on each iteration

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1114,24 +1114,27 @@ private int estimateSize(List<RowResult<Value>> rangeRows) {
...
                 return rawResults;
             }
         }
 
         if (!keysToReload.isEmpty()) {
             Map<Cell, Value> nextRawResults = keyValueService.get(tableRef, keysToReload);
             validateExternalAndCommitLocksIfNecessary(tableRef, getStartTimestamp());
+            nextRawResults.keySet().removeAll(keysAddedToResults);
```
jeremyk-91: Again, gotta be careful here - `kvs.get(...)` might return an immutable map (it does on Cassandra KVS!)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
snegrus-adb: Yeah, you are right, if that is hitting my changes then the call will fail. Why then, are we using the `Map` interface in all the places? Should we replace that with `ImmutableMap` where it may be the case?

Issue #3436, opened 2018-08-03
TimeLock: Request payloads can become massive in high-concurrency environments


jeremyk-91:
An internal framework we use implements a size limit on request payloads.

In PDS-69183 we observed that it is possible for this limit to be breached in the presence of many locks being taken out, because the lock refreshing service refreshes all active tokens in one shot. We could potentially refresh the locks in batches if the size grows beyond this point.

PR #3450, opened 2018-08-16
Batch lock refreshes

gmaretic:
**Goals (and why)**:
Fix https://github.com/palantir/atlasdb/issues/3436

Batch refresh calls in batches of 750K

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3450)


sandorw:
Are these payloads getting compressed?

gmaretic:
Added a test, although kind of flaky. Also, reduced the batch size to have a 10% buffer since we are approximating things

PR #3457, opened 2018-08-17
[no release notes] fix flaking test

gokcanozakdag:
**Goals (and why)**:
To prevent a test flake in timelockServer integration test. I believe the flake happens because of the assumption that difference between two successive getFreshTimestamp() calls should be strictly 1; but actually this can not be guaranteed. If client calls getFreshTimestamp(), and server receives the rpc, it will increment the timestamp by one, and respond to the client; but if the client cannot receive the response it will retry by calling getFreshTimestamp() again and the returned result will be bigger than the previous ts by 2.

I think the code is easier to understand, but I will try explaining 😄 

Rather than calling getFreshTimestamp() twice, and asserting the difference is equal to one; first client calls getFreshTimestamps() with a range that is bigger than the number of retries can be made on client side (100 this case). If the request of first client affects second client's timestamp; then the second call of second client should return a timestamp that is bigger than the first timestamp of second client at least by 100. 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3457)

### DIFF
--- a/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
@@ -326,11 +326,14 @@ public class PaxosTimeLockServerIntegrationTest {
         long firstServiceFirstTimestamp = timestampService1.getFreshTimestamp();
         long secondServiceFirstTimestamp = timestampService2.getFreshTimestamp();
 
-        long firstServiceSecondTimestamp = timestampService1.getFreshTimestamp();
+        long firstServiceSecondTimestamp = timestampService1
+                .getFreshTimestamps(100)
+                .getUpperBound();
+
         long secondServiceSecondTimestamp = timestampService2.getFreshTimestamp();
 
-        assertEquals(firstServiceFirstTimestamp + 1, firstServiceSecondTimestamp);
-        assertEquals(secondServiceFirstTimestamp + 1, secondServiceSecondTimestamp);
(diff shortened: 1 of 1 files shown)


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -326,11 +326,14 @@ public void timestampServiceShouldRespectDistinctClientsWhenIssuingTimestamps()
...
-        long firstServiceSecondTimestamp = timestampService1.getFreshTimestamp();
+        long firstServiceSecondTimestamp = timestampService1
+                .getFreshTimestamps(100)
+                .getUpperBound();
+
         long secondServiceSecondTimestamp = timestampService2.getFreshTimestamp();
 
-        assertEquals(firstServiceFirstTimestamp + 1, firstServiceSecondTimestamp);
-        assertEquals(secondServiceFirstTimestamp + 1, secondServiceSecondTimestamp);
+        assertThat(firstServiceSecondTimestamp - firstServiceFirstTimestamp).isGreaterThanOrEqualTo(100L);
```
jeremyk-91: Technically speaking this can fail because getFreshTimestamps might not return the full 100 timestamps (the API doesn't guarantee that). Maybe this is too much but we could do something like

```
TimestampRange tr = timestampService1.getFreshTimestamps(100);
while (tr.size() != 100) {
  tr = timestampService1.getFreshTimestamps(100);
}
long firstServiceSecondTimestamp = tr.getUpperBound();
```

gokcanozakdag: that is right. Used the already existing `getFortyTwoFreshTimestamps()` instead, that should address the problem.

Issue #2577, opened 2017-10-27
SlowLockLogger should log at WARNING rather than INFO


aggie05engineer:
From PDS-59157 - we should log lines like:

`com.palantir.lock.impl.SlowLockLogger "Blocked for {} ms to acquire lock {} {}.`

at the `WARN` level vs the `INFO` level. When locks are taking longer than anticipated this isn't informational - something is extending past thresholds.

hsaraogi:
We cant tell whats a typical long period of time for the product using the lock service for their own service.

We should add logging at the client side as the client has more context on what is a long blocking time for the lock request.

aggie05engineer:
I'm not really advocating an increase in thresholds here. 10s is likely long enough for the various reasons you'd want to take a lock in production.

tboam:
PR up #2828

tboam:
Closing as this PR merged, the next release of Atlas (0.73, probably week 1 2018) will include this.

PR #3458, opened 2018-08-17
Reduce noisy logging levels and remove unused logger

tboam:
**Goals (and why)**: reduce the amount of noise in Atlas logs.

**Implementation Description (bullets)**: reduce logging level.  delete unused logger

**Concerns (what feedback would you like?)**: Is it ok to make these info?  Do they really indicate a potential service problem?

**Where should we start reviewing?**: Only 2 classes


tboam:
Decided to put the logging level back up for slow locks.  If a product is hitting this it can indicate a problem.  IF this is intended behaviour for a product then they can configure the threshold to reduce the logging.  See here for further discussion of setting this to warn initially: https://github.com/palantir/atlasdb/issues/2577

Issue #3440, opened 2018-08-09
HeaderAccessUtils is wrong if content types have additional specifications


jeremyk-91:
```
< HTTP/1.1 200 OK
< Date: Thu, 09 Aug 2018 13:15:05 GMT
< Content-Type: text/plain;charset=iso-8859-1
< Vary: Accept-Encoding, User-Agent
< Content-Length: 36
< Server: Jetty(9.4.8.v20171121)
<
* Connection #0 to host [redacted] left intact
```

This is a text plain but we try to decode it as JSON (we should check starts-with on the value, not equality).

PR #3459, opened 2018-08-21
Fix value check in HeaderAccessUtils

snegrus-adb:
**Goals (and why)**: fixes https://github.com/palantir/atlasdb/issues/3440

**Implementation Description (bullets)**: use startsWith instead of contains for collection's items

**Testing (What was existing testing like?  What have you done to improve it?)**: added one test for that specific case

**Concerns (what feedback would you like?)**: would this make the code path slower?

**Priority (whenever / two weeks / yesterday)**: not urgent

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3459)


sandorw:
Can we use https://github.com/palantir/http-remoting/blob/develop/jaxrs-clients/src/main/java/feign/TextDelegateDecoder.java? Ideally we wouldn't need this logic specifically in AtlasDB at all.

docs/source/release_notes/release-notes.rst
```diff
@@ -55,6 +55,11 @@ develop
            Previously, trying to refresh a larger number of locks could trigger the 50MB limit in payload size.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3450>`__)
 
+    *    - |fixed|
+         - Requests for which headers' values contain some metadata after the value are parsed correctly.
```
sandorw: Can you rephrase this as 'Now correctly parse ...' for clarity?

snegrus-adb:
@sandorw , I've updated the class used with the one linked and updated the RN. I think more things could be replaced with the ones from http-remoting but that could be a separate PR. For `HeaderAccessUtils.java` the methods are slightly different. We have a method that checks if a given header has a given value but in the library it just checks if the header is present or returns its collection of values.

atlasdb-commons/src/main/java/com/palantir/common/remoting/HeaderAccessUtils.java
```diff
@@ -48,7 +48,12 @@ public static boolean shortcircuitingCaseInsensitiveContainsEntry(
             String value) {
         for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
             if (header.equalsIgnoreCase(entry.getKey())) {
-                return entry.getValue().contains(value);
+                for (String headerValue: entry.getValue()) {
```
sandorw: This method should be unused now aside from tests/benchmarks. Let's remove it.

atlasdb-commons/src/test/java/com/palantir/remoting/HeaderAccessUtilsTest.java
```diff
@@ -97,6 +101,15 @@ public void caseInsensitiveGetShortcircuits() {
         assertEquals(VALUE_2, HeaderAccessUtils.shortcircuitingCaseInsensitiveGet(testMap, KEY_2.toUpperCase()));
     }
 
+    @Test
```
sandorw: Shouldn't need this test, we can remove this method. We can probably drop the javax.ws.rs-api import as well, since I think this is the only place it's used.

atlasdb-commons/src/test/java/com/palantir/remoting/HeaderAccessUtilsTest.java
```diff
@@ -97,6 +101,15 @@ public void caseInsensitiveGetShortcircuits() {
...
+    public void parseContentTypePlainText() {
+        Map<String, Collection<String>> testMap = Maps.newLinkedHashMap();
+        testMap.put("Content-Type", ImmutableList.of("text/plain;charset=iso-8859-1"));
```
sandorw: nits, if we were going to keep this test:
* We imported HttpHeaders.CONTENT_TYPE, but still use the magic string "Content-Type" here
* Normal var naming convention in the next line is that all caps = static vars, and we should avoid naming a var the same thing as a static variable we're using

atlasdb-commons/src/main/java/com/palantir/common/remoting/HeaderAccessUtils.java
sandorw: I think `shortcircuitingCaseInsensitiveContainsEntry` is completely unused now?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/TextDelegateDecoder.java
```diff
@@ -1,71 +0,0 @@
...
-    private final Decoder stringDecoder;
-
-    public TextDelegateDecoder(Decoder delegate) {
-        this.delegate = delegate;
-        this.stringDecoder = new StringDecoder();
-    }
-
-    @Override
-    public Object decode(Response response, Type type) throws IOException, FeignException {
-        if (HeaderAccessUtils.shortcircuitingCaseInsensitiveContainsEntry(
```
sandorw: Please also remove `HeaderAccessUtils.shortcircuitingCaseInsensitiveContainsEntry`, since AFAICT it is now unused outside of tests and benchmarks. Please remove those tests and benchmarks as well.

atlasdb-feign/build.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
   explicitShadow project(':lock-api')
   explicitShadow group: 'com.squareup.okhttp3', name: 'okhttp'
   explicitShadow group: 'com.palantir.remoting-api', name: 'service-config'
+  explicitShadow group: 'com.palantir.remoting3', name: 'jaxrs-clients'
```
sandorw: I want to revisit if we need to shadow (and generally how much shadowing we need to do), but that's out of scope for this PR.

PR #3460, opened 2018-08-21
Throw specific exception when transaction is committed

ellisjoe:
**Goals (and why)**:
We've run into issues where concurrent reads/writes operating on the same transaction, at which point on of the operations fails and aborts the transaction, causing the other reads/writes to throw due to the transaction no longer being uncommitted. We'd like to handle this case explicitly and therefore need an specific exception to catch.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3460)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommittedTransactionException.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommittedTransactionException.java
@@ -16,7 +16,7 @@
 
 package com.palantir.atlasdb.transaction.impl;
 
-public final class CommittedTransactionException extends RuntimeException {
+public final class CommittedTransactionException extends IllegalStateException {
     public CommittedTransactionException() {
         super("Transaction must be uncommitted");
     }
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -304,8 +304,7 @@ protected void checkGetPreconditions(TableReference tableRef) {
             throw new TransactionFailedRetriableException("Transaction timed out.");
         }
         Preconditions.checkArgument(allowHiddenTableAccess || !AtlasDbConstants.hiddenTables.contains(tableRef));
-        Preconditions.checkState(state.get() == State.UNCOMMITTED || state.get() == State.COMMITTING,
-                "Transaction must be uncommitted.");
+        ensureUncommitted();
```
ellisjoe: this is a behavior change, though it doesn't seem necessary to support reads on `COMMITTING` transactions.

sandorw: This seems like a good behavior change, though it might affect our big internal product. Especially when serializable transactions are in the mix, we should not allow reads during the commit step.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/CommittedTransactionException.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.impl;
+
+public final class CommittedTransactionException extends RuntimeException {
+    public CommittedTransactionException() {
+        super("Transaction must not be committed");
```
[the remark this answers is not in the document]
ellisjoe: done

PR #3461, opened 2018-08-21
Circle2, Part 1: Going in Endless Circles (Initial Port)

jeremyk-91:
**Goals (and why)**:
- We need to move to Circle2 by the end of this month. See [CircleCI blog](https://circleci.com/blog/sunsetting-1-0/).

- Perform a basic translation of our Circle 1 circle.yml into a Circle 2 configuration.
  - This uses a three-step workflow (build, check, publish).
  - We save some $$ here in that we only run one node when building, and then distribute the class files to all nodes to run the tests.
- Drive by fixes of quite a few bugs that were causing problems building in Circle 2, or otherwise found:
  - **Buildscript bugs**
    - We assumed that `python`, `pip` and the Python `DateUtil` module were already installed on containers we run the builds on. We now install them if they aren't already installed.
    - Anyone who attempted to run the Jepsen tests and parsing script who didn't have `DateUtil` would also have been failed out, but we now install it if necessary.
    - We assumed that `rsync` was already installed on containers we run the builds on; we now install it if it wasn't already installed.
    - Our `rsync` command is technically unsafe if you have projects that are two or more directories deep (i.e. `./timelock-api` is OK, but if we had `./timelock/timelock-api` as a project that would break). For now I've removed the profile client from the checkstyle reports, though we should consider whether we actually want that piece of code hanging around.
  - **Test infrastructure**
    - The one node down Cassandra test was often too eager to create a Cassandra KVS and provide it to the Schema Mutation Lock Releasing Rule (which the Flake Rule needs to know about). I've changed this to lazily use a Supplier.
- Disabled the TimeLock migration tests for now, as they rely on mounting volumes which is not. A Part 2 of this feature could involve separating TimeLock migration into a separate job in our workflow, that runs on a `machine` executor, and/or changing the TimeLock migration test to run suitable Docker-Exec commands as opposed to manipulating the mounted TemporaryFolder.

- There isn't very much testing of our build infrastructure. Many of the "bugs" found were difficult to test because in practice the containers we got in Circle 1 don't put us in those circumstances.
- **Timelock Migration test is ignored**; we should get a fix out soon and I think I'll learn a bit more about circle 2 by doing that. The value of that test is somewhat lower now as well as fewer people are actively migrating, though could become significant again when $LARGE_INTERNAL_PRODUCT upgrades. Also, the test does cover some of the fail-safe functionality.

- There are probably some steps that can be improved or optimised.
- There are a bunch of hacks, as I was attempting to simply get things into a workable state (there's quite a lot on my plate this week, between this and other stability and resourcing work)
- The Cassandra multinode test is still flaky in spite of the improvement to the flake rule to lazily initialize the key-value service; sometimes it seems to take longer than three minutes. Moving to a machine executor _might_ help.
- Is disabling the timelock migration test reasonable to do?
- The deploy part of this is unfortunately untested.

**Where should we start reviewing?**: `.circleci/config.yml`. I'm happy to answer questions on the Circle docs; it's also probably worthwhile checking them to validate some of the constructs I've used.

**Priority (whenever / two weeks / yesterday)**: This week. 31 Aug is a hard deadline but we should really have this in well before that, so that we can proceed with some of the further improvements and that we'll have enough time to test publishing and deploying.


.circleci/config.yml
```diff
@@ -0,0 +1,139 @@
...
+          key: gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+          paths:
+            - ~/.gradle/wrapper
+      - save_cache:
+          key: gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}
+          paths:
+            - ~/.gradle/caches
+      - run: git tag -l | xargs git tag -d && git fetch -t # ensure all tags are fetched and up-to-date
+      - run:
+          command: ./scripts/circle-ci/stop-services.sh
```
iamdanfox: this is unnecessary now because the `circleci/openjdk:8u171-jdk-node-browsers` docker image shouldn't have random background stuff running.

jeremyk-91: Yep, alright. I'll remove this

iamdanfox:
@jeremyk-91 I think the CircleCI 1.0 deadline only applies to open source projects. https://circleci.com/sunset1-0/

However, it seems that atlasdb stopped using the open source CircleCI around seven months ago: https://circleci.com/gh/palantir/atlasdb.  Instead, the `ci/circleci-internal` thing seems to suggest that CI is done on our enterprise CircleCI?

jeremyk-91:
Correct, atlasdb is built internally so my concerns about the hard deadline are overblown. I think it's still worth finishing up now that the work has been done (admittedly had I known a week ago it wasn't required we might have prioritised things differently).

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+
+    steps:
+      - attach_workspace: { at: . }
+      - restore_cache:
+          key: gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+      - restore_cache:
+          key: gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}
+      - setup_remote_docker
+      - run: ./scripts/circle-ci/install-docker.sh
+      - run: sudo wget --retry-connrefused --waitretry=1 --read-timeout 20 --timeout 15 -t 10 -q -O /usr/local/bin/docker-compose https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m`
```
iamdanfox: @jmcampanini FYSA I think it would be cool if there was a variant of our internal `ubuntu-jdk` docker image that could be used externally so atlasdb doesn't have to do all this installing.

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+    parallelism: 8
+    docker:
+      - image: circleci/openjdk:8u171-jdk-node-browsers
+    resource_class: xlarge
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.console=plain -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false
+      _JAVA_OPTIONS: "-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -verbose:gc -XX:+PrintGC -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:-TraceClassUnloading -Xloggc:build-%t-%p.gc.log"
+      TERM: dumb
+      CIRCLE_TEST_REPORTS: /tmp/test-reports
+      CIRCLE_ARTIFACTS: /tmp/artifacts
```
iamdanfox: If possible, I'd recommend using exactly the same values that we use internally for this:

```
# Setup default locations for Circle test reports + artifacts
ENV CIRCLE_TEST_REPORTS=/home/circleci/junit CIRCLE_ARTIFACTS=/home/circleci/artifacts
RUN mkdir -p /home/circleci/junit && mkdir -p /home/circleci/artifacts
```

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+          command: find . -type f -regex ".*/build/test-results/TEST-.*\.xml" -exec cp {} --parents $CIRCLE_TEST_REPORTS/junit/ \;
+          when: always
+      - store_test_results:
+          path: /tmp/test-reports/ # Circle doesn't do substitution in this path
+      - run: mkdir -p $CIRCLE_ARTIFACTS/checkstyle $CIRCLE_ARTIFACTS/findbugs
+      - run: if [ -d docs/build/html ]; then cp -r docs/build/html $CIRCLE_ARTIFACTS/docs; fi
+      - run: sudo apt-get install rsync
+      - run: find . -type d -regex ".*/build/reports/checkstyle" | (grep -v profile-client || true) | sed 's#\./\(\(.*\)/build/reports/checkstyle\)#rsync -uav \1/ $CIRCLE_ARTIFACTS/checkstyle/\2#' | bash
+      - run: find . -type d -regex ".*/build/reports/findbugs" | (grep -v profile-client || true) | sed 's#\./\(\(.*\)/build/reports/findbugs\)#rsync -uav \1/ $CIRCLE_ARTIFACTS/findbugs/\2#' | bash
+      - run: rsync -uav "build/reports/profile" "$CIRCLE_ARTIFACTS/"
```
iamdanfox: If you use `com.palantir.baseline-circleci` then you don't need this step.

https://github.com/palantir/gradle-baseline#circleci-plugin-compalantirbaseline-circleci

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
...
+          paths: [ . ]
+
+  deploy:
+    docker:
+      - image: circleci/openjdk:8u171-jdk-node-browsers
+    resource_class: xlarge
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.console=plain -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false
+      _JAVA_OPTIONS: "-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -verbose:gc -XX:+PrintGC -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:-TraceClassUnloading -Xloggc:build-%t-%p.gc.log"
+      TERM: dumb
```
iamdanfox: I think you only need one of `-Dorg.gradle.console=plain` and `TERM: dumb`

.circleci/config.yml
```diff
@@ -0,0 +1,136 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: circleci/openjdk:8u171-jdk-node-browsers
```
iamdanfox: Interestingly, it seems `circleci/openjdk:8u181-node` is 696MB while the `*-8u171-jdk-node-browsers` series is more like 898MB.  Could you use the smaller one?

jeremyk-91: Sure. I'm pretty sure we don't actually use the browsers.

.circleci/config.yml
jeremyk-91: Got it. I'll remove `TERM: dumb`

.circleci/config.yml
jeremyk-91: Yep, will standardise on this. I don't think we depend on the directories being in any specific place

.circleci/config.yml
snegrus-adb: nit: would be feasible to use a Yaml pointer (I think they are called) (https://coderwall.com/p/wqgm7g/using-pointers-to-merge-elements-in-yaml-file) to reuse these configs? Not sure if it's possible to merge pointers between them for the particular cases.

iamdanfox: I know lots of people use yaml pointers now, but I think they're actually net negative for this use-case.  They introduce a bit more complexity (which nobody can write correctly off the top of their head, and nobody is 100% sure what they do just by eyeballing) just to eliminate one or two lines.  In practise, I think the duplication and super easy to read YAML is preferrable.

In order to make CircleCI config shorter, I've just used a denser syntax: 

```yaml
  compile:
    docker: [{ image: 'circleci/openjdk:...' }]
    resource_class: medium+
    steps:
```

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+          key: gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+          paths:
+            - ~/.gradle/wrapper
+      - save_cache:
+          key: gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}
+          paths:
+            - ~/.gradle/caches
+      - run: git tag -l | xargs git tag -d && git fetch -t # ensure all tags are fetched and up-to-date
+      - run: ./scripts/circle-ci/download-moreutils.sh
+      - run: ./scripts/time-cmd.sh ./gradlew --profile --parallel --stacktrace classes testClasses
```
iamdanfox: I'd recommend doing `resolveConfigurations classes testClasses` in one go.

atlasdb-ete-tests/build.gradle
```diff
@@ -52,6 +52,10 @@ task longTest(type: Test) {
 task timeLockTest(type: Test) {
     dependsOn prepareForEteTests, ':timelock-server-distribution:dockerTag'
     include '**/*TimeLock*.class'
+
+    // TODO (jkong): Re-enable migration tests once a new story for handling volumes-in-circle-2 is available.
```
iamdanfox: We use the machine executor internally to do dockerized testing with volumes

jeremyk-91: Yep, I've seen this in $INTERNAL_SECURITY_PRODUCT as well. Thanks

PR #3462, opened 2018-08-22
Allow multiple registration of TS metrics

gmaretic:
**Goals (and why)**:
Currently, if we register another instance of TS metrics, it will deregister the previous instance. This is fine if you do not have a separate writer and a sweeper, but gets messy if you want to have that, as the large internal products does.
 
Instead of bulldozing over existing metrics, use registerOrGet. Needed to hack around a bit to put the additional information necessary (that was previously part of the class) into the metrics themselves so that information can be shared across instances of TSMetrics.

Added a couple of tests to verify it behaves as expected.

Is there a nicer way of doing this?

TSMetrics

ASAP, because we need metrics to work correctly when we deploy with the large internal product.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3462)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
@@ -81,11 +81,15 @@ public class MetricsManager {
         registerMetric(clazz, metricName, (Metric) gauge);
     }
 
+    public void registerMetric(Class clazz, String metricName, Metric metric) {
+        registerMetricWithFqn(MetricRegistry.name(clazz, metricName), metric);
+    }
+
     /**
      * Add a new gauge metric of the given name.
      *
      * If the metric already exists, this will REPLACE it with a new metric.
-     * Consider using {@link MetricsManager#registerIfNotExists} instead.
+     * Consider using {@link MetricsManager#registerOrGet} instead.
      */
     public void registerMetric(Class clazz, String metricName, Gauge gauge, Map<String, String> tag) {
         MetricName metricToAdd = MetricName.builder()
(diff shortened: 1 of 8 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -101,33 +105,25 @@ public void registerMetric(Class clazz, String metricName, Gauge gauge, Map<Stri
...
 
         try {
-            taggedMetricRegistry.gauge(metricToAdd, gauge);
+            return taggedMetricRegistry.gauge(metricToAdd, gauge);
         } catch (IllegalArgumentException ex) {
             log.warn("Tried to add a gauge to a metric name {} that has non-gauge metrics associated with it."
                     + " This indicates a product bug.",
                     SafeArg.of("metricName", metricName),
                     ex);
+            return gauge;
```
snegrus-adb: Why return if is not the one registered?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -116,45 +117,54 @@ private static Long minimum(Collection<Long> currentValues) {
...
+            AggregatingVersionedSupplier<Long> lastSweptTsSupplier = new AggregatingVersionedSupplier<>(
+                    TargetedSweepMetrics::minimum, recomputeMillis);
+            lastSweptTs = register(AtlasDbMetricNames.LAST_SWEPT_TS,
+                    new AggregatingVersionedMetric<>(lastSweptTsSupplier), tag);
 
-            lastSweptTsSupplier = new AggregatingVersionedSupplier<>(TargetedSweepMetrics::minimum, recomputeMillis);
             Supplier<Long> millisSinceLastSweptTs = new CachedComposedSupplier<>(
-                    lastSweptTs -> estimateMillisSinceTs(lastSweptTs, wallClock, tsToMillis),
-                    lastSweptTsSupplier);
+                    swept -> estimateMillisSinceTs(swept, wallClock, tsToMillis), lastSweptTs::getVersionedValue);
```
snegrus-adb: nit: I know there are many `lastSwept*` variables but if possible could this be named different than `swept`, maybe `sweptTs`

atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
gmaretic: The alternative is to throw a RTE, or return an Optional, but since the old behaviour was to only log at warn and ignore, this seemed to be more consistent. Not sure if we want to hard fail when we hit this

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TimestampedAccumulatingValueMetric.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Long getValue() {
+        return delegate.getValue();
+    }
+
+    public void setValue(Long newValue, long timestamp) {
+        forwardTimestampTo(timestamp);
+        delegate.setValue(newValue);
+    }
```
jeremyk-91: Discussed offline: race condition here exists, but is acceptable as we aren't relying on the precise values for correctness and this is a reasonable approximation

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetricsTest.java
```diff
@@ -197,19 +197,49 @@ public void millisSinceLastSweptUpdatesAsClockUpdatesAfterWaiting() {
...
         puncherStore.put(0, 50);
-        assertThat(metricsManager).hasMillisSinceLastSweptConservativeEqualTo(50L);
+        assertThat(metricsManager).hasMillisSinceLastSweptConservativeEqualTo(clockTime - 50);
 
+        waitForProgressToRecompute();
         clockTime += 1;
-        assertThat(metricsManager).hasMillisSinceLastSweptConservativeEqualTo(50L);
-
-        clockTime += 100;
-        assertThat(metricsManager).hasMillisSinceLastSweptConservativeEqualTo(50L);
```
jeremyk-91: Seems like we lost the old test here - would it be feasible to re-add it?

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TimestampedAccumulatingValueMetric.java
```diff
@@ -0,0 +1,57 @@
...
+        delegate.accumulateValue(newValue);
+    }
+
+    private void forwardTimestampTo(long timestamp) {
+        latestTimestamp.accumulateAndGet(timestamp, Long::max);
+    }
+
+    public long getLatestTimestamp() {
+        return latestTimestamp.get();
+    }
```
jeremyk-91: A small amount of Javadoc on what this does might be useful for helping others understand the system later on.

PR #3463, opened 2018-08-22
[no release notes] Lower CachingTransaction shortcut logic logging to debug

mswintermeyer:
Large internal product just increased logging to info by default. Given this code will shortcut the call anyways, there's no need for so much log spew.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3463)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
@@ -71,7 +71,7 @@ public class CachingTransaction extends ForwardingTransaction {
     public SortedMap<byte[], RowResult<byte[]>> getRows(TableReference tableRef, Iterable<byte[]> rows,
                                                         ColumnSelection columnSelection) {
         if (Iterables.isEmpty(rows)) {
-            log.info("Attempted getRows on '{}' table and {} with empty rows argument", tableRef, columnSelection);
+            log.debug("Attempted getRows on '{}' table and {} with empty rows argument", tableRef, columnSelection);
             return AbstractTransaction.EMPTY_SORTED_ROWS;
         }
 
@@ -114,7 +114,7 @@ public class CachingTransaction extends ForwardingTransaction {
     @Override
     public Map<Cell, byte[]> get(TableReference tableRef, Set<Cell> cells) {
         if (cells.isEmpty()) {
-            log.info("Attempted get on '{}' table with empty cells argument", tableRef);
+            log.debug("Attempted get on '{}' table with empty cells argument", tableRef);
             return ImmutableMap.of();
         }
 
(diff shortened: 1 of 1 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
```diff
@@ -114,7 +118,11 @@ public Transaction delegate() {
     @Override
     public Map<Cell, byte[]> get(TableReference tableRef, Set<Cell> cells) {
         if (cells.isEmpty()) {
-            log.info("Attempted get on '{}' table with empty cells argument", tableRef);
+            if (log.isTraceEnabled()) {
+                log.trace(String.format("Attempted get on %s table with empty cells argument", tableRef), new Exception());
```
cakofony: String.format isn't necessary, the logger knows what to do with params + throwable:
`log.trace("Attempted get on {} table with empty cells argument", tableRef, new Exception())`

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
```diff
@@ -114,7 +118,11 @@ public Transaction delegate() {
...
+            } else if (log.isDebugEnabled()) {
+                log.debug("Attempted get on '{}' table with empty cells argument", tableRef);
```
cakofony: I'd prefer to leave this one at info level so that we can find future instances where we call these methods unnecessarily.

Minor preference to always log this one, regardless of whether or not trace is enabled, it's a bit funky that enabling trace will disable info logging.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
```diff
@@ -71,7 +71,11 @@ public Transaction delegate() {
     public SortedMap<byte[], RowResult<byte[]>> getRows(TableReference tableRef, Iterable<byte[]> rows,
                                                         ColumnSelection columnSelection) {
         if (Iterables.isEmpty(rows)) {
-            log.info("Attempted getRows on '{}' table and {} with empty rows argument", tableRef, columnSelection);
+            if (log.isTraceEnabled()) {
```
sandorw: Do we need the explicit conditions here? I'm not familiar with all of our logging libraries.

cakofony: `new Exception()` is expensive, so we want to make sure trace is enabled prior to allocating it, and filling in the call stack.

sandorw: Ah, I missed the exception in the log message. I don't like injecting the exception here. Is this just to force a stack trace that points to where the problem occurred?

schlosna: I'd actually propose to remove the `Attempted get*` logging completely at this point. If one calls with an empty rows/cells the empty result should be expected and I don't think this logging has ever actually been valuable.

sandorw: Agreed, I would be surprised if anyone used this.

cakofony: Works for me.

PR #3465, opened 2018-08-23
Improve visibility of inter-lock comms errors

stefank-palantir:
Make it easier to identify when remote exceptions have prevented quorum
by logging the exception at ERROR rather than WARN.

**Goals (and why)**:

We hit an issue in the field recently (PDS-71980) where a non-timelock lock server in a three-node cluster hit a thread leak. We suspect, but can't fully prove, that `PaxosLeaderElectionService.blockOnBecomingLeader` got stuck in a tight loop due to remote exceptions causing `PaxosQuorumChecker.collectResponses` to repeatedly immediately bail out with `NO_QUORUM`, leaking executor threads for the requests that didn't fail out.

However, WARN logging from external dependencies doesn't get picked up by default in the product in question, so we don't have any record of what the error might have been. I think remote exceptions probably deserve to be logged at ERROR anyway, so I think the correct minimal change here is to bump up the log level in Atlas.

I'll submit a separate PR for ensuring `blockOnBecomingLeader` backs off rather than tight-looping when quorum isn't achieved; this one is just for the logging change.

* Bump a couple of log lines to ERROR that were previously at WARN.

* N/A

* If you feel strongly that this log shouldn't be at ERROR, I guess that's the only real point of contention. 

* It's a very short diff.

* No specific deadline, but I'd hate to hit this again without the logging. Order of a week or so?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3465)

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -172,7 +172,7 @@ public final class PaxosQuorumChecker {
                     if (onlyLogOnQuorumFailure) {
                         toLog.add(e.getCause());
                     } else {
-                        log.warn(PAXOS_MESSAGE_ERROR, e.getCause());
+                        log.error(PAXOS_MESSAGE_ERROR, e.getCause());
                     }
                 }
             }
@@ -187,7 +187,7 @@ public final class PaxosQuorumChecker {
                     interrupted = true;
                     break;
                 } catch (ExecutionException e) {
-                    log.warn(PAXOS_MESSAGE_ERROR, e.getCause());
+                    log.error(PAXOS_MESSAGE_ERROR, e.getCause());
                 }
             }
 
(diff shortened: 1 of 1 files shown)


tboam:
I agree with Carter here: the principle we've been aiming for on develop is that something being logged at ERROR means a serious problem and someone should get paged.  WARN means something bad might be about to happen, or there could be something worth investigating at a later date.  We're not there yet, but this is the goal.

In this instance it seems like it would be fine for a few Exceptions to happen, particularly if we are still able to achieve quorum.  In fact in a HA world (this code is used in Timelock) it should be fine for another node to be uncontactable for short periods.

I'd prefer we push the product in question to allow WARN level logging from this class.  If this isn't possible, or isn't achievable in the timeframe needed I'm happy to merge this to 0.61 but wouldn't merge it into develop and we may need to back this change out once the investigation is complete.

stefank-palantir:
Yep, I agree on reflection that it's better to change the logging in the product rather than in atlas.

Feel free to close this out. Thanks Tom!

PR #3469, opened 2018-08-24
Fix CqlQuery leak and make it (more) foolproof

jeremyk-91:
**Goals (and why)**:
- Internal ref PDS-73313

- Make CqlQuery foolproof
- Fix the specific leak
- Fix other benign but incorrect usages

- Non-existent for CqlQuery; now present including checks on log levels
- New methods are tested

- The unlogged batch for timestamp service is snowflaky right now
- Are the safety levels correct?
- Did I break anything when moving CqlQuery to use immutables? This isn't the smallest change but I want to improve that class.

**Where should we start reviewing?**: CqlQuery

**Priority (whenever / two weeks / yesterday)**: several weeks ago 🔥 🔥 🔥

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3469)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTimestampUtils.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTimestampUtils.java
@@ -47,6 +47,7 @@ import com.palantir.atlasdb.table.description.NamedColumnDescription;
 import com.palantir.atlasdb.table.description.TableMetadata;
 import com.palantir.atlasdb.table.description.ValueType;
 import com.palantir.atlasdb.transaction.api.ConflictHandler;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.util.Pair;
 
 public final class CassandraTimestampUtils {
@@ -93,7 +94,13 @@ public final class CassandraTimestampUtils {
             builder.append(constructCheckAndSetQuery(columnName, expected, target));
         });
         builder.append("APPLY BATCH;");
-        return new CqlQuery(builder.toString());
+
+        // This looks awkward. However, we know that all expressions in this String pertain to timestamps and known
+        // table references, hence this is actually safe. Doing this quickly owing to priority.
+        // TODO (jkong): Build up a query by passing around legitimate formats and args.
+        return CqlQuery.builder()
+                .safeQueryFormat(builder.toString())
+                .build();
     }
 
     public static boolean isValidTimestampData(byte[] data) {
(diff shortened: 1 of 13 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTimestampUtils.java
```diff
@@ -93,20 +94,29 @@ public static CqlQuery constructCheckAndSetMultipleQuery(Map<String, Pair<byte[]
             builder.append(constructCheckAndSetQuery(columnName, expected, target));
         });
         builder.append("APPLY BATCH;");
-        return new CqlQuery(builder.toString());
+
+        // This looks awkward. However, we know that all expressions in this String pertain to timestamps and known
```
gmaretic: OK, as long as we actually fix this soon. OK with this to get out of p0 situation

gmaretic:
Also logging levels look correct

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/SchemaMutationLockTables.java
```diff
@@ -100,10 +100,13 @@ private void createTableWithCustomId(TableReference tableRef) throws TException
...
                 + "    AND id = '%s'";
-        CqlQuery query = new CqlQuery(createTableStatement,
-                SafeArg.of("keyspace", keyspace),
-                SafeArg.of("internalTableName", internalTableName),
-                SafeArg.of("cfId", uuid));
+        CqlQuery query = CqlQuery.builder()
+                .safeQueryFormat(createTableStatement)
+                .addArgs(
+                        SafeArg.of("keyspace", keyspace),
+                        SafeArg.of("internalTableName", internalTableName),
```
tpetracca: any reason we assume safe here and don't use LoggingArgs like we do elsewhere?

jeremyk-91: I think this is because the log arbitrator works on the basis of schema metadata, while the lock tables are an internal table that isn't a part of a product's atlas schema.

It would probably be nice to push this logic into a part of the log arbitrator so that it handles these cases properly, though

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTimestampUtils.java
jeremyk-91: This is addressed in #3472

PR #3473, opened 2018-08-24
Fix upgrading shards issue

gmaretic:
**Goals (and why)**:
Fix a bug where we would repeatedly fail to CAS to update the number of shards from the default value causing targeted sweep to error out.

This would only hit the early adopters as in the very first release with TS we were persisting the default number of shards (1) due to defaulting a missing entry to -1:
https://github.com/palantir/atlasdb/blob/0bd6dbd4969bc2e57ebd0df0a6b92774e4d4da83/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java#L118

However, this could hit other users as well if we ever changed the default value.
 
If we determine that we are changing from the default value, we check again to make sure if there is an existing entry. Could have avoided this extra lookup at the cost of making the code harder to read, but decided its not worth it.

The new test, which fails on develop.

Are there any other corner cases we might have missed?

Small

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3473)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
@@ -39,11 +39,11 @@ import com.palantir.util.PersistableBoolean;
 
 public class ShardProgress {
     private static final Logger log = LoggerFactory.getLogger(ShardProgress.class);
-    private static final TableReference TABLE_REF = TargetedSweepTableFactory.of()
+    static final TableReference TABLE_REF = TargetedSweepTableFactory.of()
             .getSweepShardProgressTable(null).getTableRef();
 
     private static final int SHARD_COUNT_INDEX = -1;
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ShardProgress.java
```diff
@@ -126,29 +126,49 @@ private long increaseValueFromToAtLeast(ShardAndStrategy shardAndStrategy, long
...
+                    .map(persistedValue -> createSingleCellRequest(shardAndStrategy, persistedValue, colValNew))
+                    .orElse(createNewCellRequest(shardAndStrategy, colValNew));
+        } else {
+            return createSingleCellRequest(shardAndStrategy, oldVal, colValNew);
         }
-        byte[] colValOld = SweepShardProgressTable.Value.of(oldVal).persistValue();
+    }
+
+    private boolean isDefaultValue(ShardAndStrategy shardAndStrategy, long oldVal) {
+        return oldVal == SweepQueueUtils.INITIAL_TIMESTAMP
```
sandorw: Comment here about the bug? Comparing the number of shards to a timestamp reads strangely

PR #3478, opened 2018-08-28
Cassandra KVS Startup Metadata Check Must Be Case-Insensitive

jeremyk-91:
- [x] Add a test that captures precisely this

**Goals (and why)**:
- Fix PDS-73442 which has caused some alarm internally
- In recent versions of Atlas the Cassandra KVS writes metadata as lower-case to allow for faster look-ups, but I think we missed this call-site

- Have a fallback mechanism for looking up table metadata given a map of available table metadata, and use it

- Not great which is why we didn't catch this. I plan to add a test tomorrow.

- This is unlikely to be the most performant way to implement this method (it is worst-case quadratic; a linear solution exists by pre-processing the map and deciding on a standard form). Is this a problem?

**Where should we start reviewing?**: CassandraKeyValueServiceImpl

**Priority (whenever / two weeks / yesterday)**: tomorrow ideally

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3478)


clockfort:
fix lgtm, though definitely wants a test

PR #3484, opened 2018-08-31
Circle2, Part 2: Sharpening the Circular Saw

jeremyk-91:
**Goals (and why)**:
- Make builds faster

- Reshuffle some of the tasks that were not on the fastest executors

- The tests haven't changed :)

- It is possible to do even better by fragmenting the startup independence tests. I avoided that for simplicity here.

**Where should we start reviewing?**: run-circle-tests.sh

**Priority (whenever / two weeks / yesterday)**: whenever, also blocked on #3461

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3484)

### DIFF
--- a/scripts/circle-ci/run-circle-tests.sh
+++ b/scripts/circle-ci/run-circle-tests.sh
@@ -11,17 +11,17 @@ function checkDocsBuild {
     make html
 }
 
-CONTAINER_1=(':atlasdb-cassandra-integration-tests:check' ':atlasdb-cassandra-integration-tests:memorySensitiveTest')
+CONTAINER_1=(':atlasdb-cassandra-integration-tests:check' ':atlasdb-cassandra:check')
 
-CONTAINER_2=(':atlasdb-ete-tests:check')
+CONTAINER_2=(':atlasdb-ete-tests:check' ':atlasdb-ete-test-utils:check')
 
-CONTAINER_3=(':atlasdb-perf:postgresBenchmarkTest')
+CONTAINER_3=(':atlasdb-dbkvs:check'  ':atlasdb-perf:postgresBenchmarkTest')
 
-CONTAINER_4=(':atlasdb-dbkvs:check' ':atlasdb-cassandra-multinode-tests:check' ':atlasdb-impl-shared:check' ':atlasdb-cassandra-integration-tests:longTest')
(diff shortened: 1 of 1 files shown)


scripts/circle-ci/run-circle-tests.sh
```diff
@@ -11,17 +11,17 @@ function checkDocsBuild {
...
-CONTAINER_3=(':atlasdb-perf:postgresBenchmarkTest')
+CONTAINER_3=(':atlasdb-dbkvs:check' ':atlasdb-perf:postgresBenchmarkTest' ':atlasdb-cassandra:check')
 
-CONTAINER_4=(':atlasdb-dbkvs:check' ':atlasdb-cassandra-multinode-tests:check' ':atlasdb-impl-shared:check' ':atlasdb-cassandra-integration-tests:longTest')
+CONTAINER_4=(':atlasdb-cassandra-multinode-tests:check' ':atlasdb-impl-shared:check' ':atlasdb-cassandra-integration-tests:longTest' ':atlasdb-tests-shared:check' ':atlasdb-perf:check')
 
-CONTAINER_5=(':atlasdb-ete-tests:longTest' ':lock-impl:check' ':atlasdb-dbkvs-tests:check' ':atlasdb-tests-shared:check' ':atlasdb-perf:check')
+CONTAINER_5=(':atlasdb-cassandra-integration-tests:memorySensitiveTest' ':atlasdb-ete-tests:longTest' ':lock-impl:check' ':atlasdb-dbkvs-tests:check')
 
-CONTAINER_6=(':atlasdb-ete-tests:startupIndependenceTest' ':atlasdb-ete-test-utils:check' ':atlasdb-cassandra:check' ':atlasdb-api:check' ':atlasdb-jepsen-tests:check' ':atlasdb-cli:check')
```
sandorw: The last three here were dropped

jeremyk-91:
Yep, the "dropped" checks were intentional. Container 0 handles all the tasks that weren't explicitly called out to be run on any other container, so by removing them from the overrides it actually means they're being run on container 0.

We could do some more precise fine-tuning, but given that Startup Independence is the longest test by quite a margin fine-tuning wouldn't actually release the build time very much.

scripts/circle-ci/run-circle-tests.sh
jeremyk-91: (indeed, good spot - as mentioned below, 0 runs all the rest of the tasks)

PR #3488, opened 2018-09-05
Improved visibility into cassandra load

tboam:
**Goals (and why)**: We want to understand what is causing high call rates to the _transactions table.

**Implementation Description (bullets)**:  Add logging when we look up a large number of transactions at once (this threshold is configurable.  Tag some Meters with the table name so we can understand which tables are returning a large number of transactions to look up.

**Testing (What was existing testing like?  What have you done to improve it?)**: Logging and metrics so haven't tested it.

**Concerns (what feedback would you like?)**: Have I done this correctly?  Is the default threshold correct?

**Where should we start reviewing?**: SnapshotTransaction.java

**Priority (whenever / two weeks / yesterday)**: Ideally this week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3488)


atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -193,6 +189,12 @@ private synchronized Meter registerOrGetMeter(String fullyQualifiedMeterName) {
         return meter;
     }
 
+    public synchronized Meter registerOrGetTaggedMeter(Class clazz, String metricName, Map<String, String> tags) {
+        MetricName name = getTaggedMetricName(clazz, metricName, tags);
+        // TODO(tboam): do we need to keep a seprate record of registeredTaggedMetrics so we can clean those up too?
```
tboam: `deregisterMetrics` removes untagged metrics but I don't see the same for tagged metrics.  does this need to be added?

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
```diff
@@ -80,6 +80,7 @@ private AtlasDbConstants() {
     public static final long MAX_TS = Long.MAX_VALUE;
 
     public static final long DEFAULT_TRANSACTION_LOCK_ACQUIRE_TIMEOUT_MS = 60_000;
+    public static final int THRESHOLD_FOR_LOGGING_LARGE_NUMBER_OF_TRANSACTION_LOOKUPS = 10_000;
```
tboam: Is this the right threshold?  I could set it ridiculously high so users have to 'enable' it by setting a sensible value in config.

atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -127,20 +121,22 @@ public Gauge registerOrGet(Class clazz, String metricName, Gauge gauge, Map<Stri
...
+                .safeName(MetricRegistry.name(clazz, metricName))
+                .safeTags(tags)
+                .build();
+    }
+
+    public Map<String, String> getTableNameTagFor(TableReference tableRef) {
+        String tableName = tableRef == null ? "unknown" : tableRef.getTablename();
         if (!isSafeToLog.test(tableRef)) {
-            tableName = "unsafeTable_" + obfuscate(tableRef);
+            tableName = "unsafeTable";
```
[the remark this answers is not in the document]
tboam: Nullable - Done.

Most tables are Safe for logging now so I see less need for this slightly hacky code.

tboam:
TaggedMetrics are now cleaned up too

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
tboam: Added a Histogram

gokcanozakdag:
Did a second review, looks good 👍 @tboam

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
tpetracca: it also may be interesting as a follow up for this to have the theshold be configurable/overridable per table.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
```diff
@@ -32,4 +32,8 @@ public long getLockAcquireTimeoutMillis() {
         return AtlasDbConstants.DEFAULT_TRANSACTION_LOCK_ACQUIRE_TIMEOUT_MS;
     }
 
+    @Value.Default
+    public int getThresholdForLoggingLargeNumberOfTransactionLookups() {
+        return AtlasDbConstants.THRESHOLD_FOR_LOGGING_LARGE_NUMBER_OF_TRANSACTION_LOOKUPS;
```
tpetracca: As mentioned above it would be cool for this to be a map of table-name to value or something like that (where an empty map implies all tables use the default.

PR #3490, opened 2018-09-05
Pick up rebranded remoting

qinfchen:
**Goals (and why)**:
pick the latest rebranded remoting

there are several changes made to http-remoting as part of the rebranding
- `http-remoting` is renamed to `conjure-java-runtime` with the new group name ` com.palantir.conjure.java.runtime`
- `http-remoting` is renamed to `conjure-java-runtime-api`
- `http-remoting`'s tracing has been moved to [tracing java](https://github.com/palantir/tracing-java) with new group name `com.palantir.tracing`

- Need to work with atlasdb team to make the changes do not introduce potentially break and there is  safe migration path for atlasdb to transition from remoting3 to conjure-java-runtime
- tracing is thread local, renaming tracing potentially would break all projects' tracing, this is blocked on https://github.com/palantir/tritium/pull/100 to ensure that remoting3 and tracing use the same Tracer
- if atlasdb throws remoting3 exceptions, atlasdb needs to updated to handle both old and new exceptions. 
- remoting3 is already on jackson 2.9.5 but atlasdb is on jackson 2.6.7, and the new ObjectMappers require [MapperFeature.ALLOW_COERCION_OF_SCALARS](https://github.com/palantir/http-remoting/blob/conjure/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java#L110) which is only available in 2.9.x

version.props

two weeks


sandorw:
Let's revisit this next week with infra people in town.

qinfchen:
Makes sense. This pr is also pretty out of date

Issue #3390, opened 2018-07-19
Cassandra NodeDownTestSuite tests are difficult to reason about


jeremyk-91:
Mainly because we have tests that have side effects. The test table is set up once at the start with some notional data, and then various tests run that may affect the contents of the table.

gmaretic:
The refactor made them run in different keyspaces, which solves most of the side effects complexity. Probably won't invest any more time in simplyfing things further

PR #3480, opened 2018-08-30
Follow up all cassandra schema changes by more permissive wait

gmaretic:
**Goals (and why)**:
Fixes:
1. https://github.com/palantir/atlasdb/issues/3367
2. https://github.com/palantir/atlasdb/issues/3390
3. https://github.com/palantir/atlasdb/issues/3398
4. Correct wrong javadocs in CKVSImpl.
5. Make the behaviour with respect to retrying logic deterministic when fewer than a quorum of nodes are reachable.
6. Replace table truncation custom retry logic with runWithRetry.

**Implementation Description (bullets)**:
1. All operations that do C* schema mutations now only require a quorum of nodes to succeed. In case there is a follow-up action, as in the case of deleting metadata entries when we drop tables, we make a best-effort attempt and do not throw on failure. Also change the implementation of getAllTableNames to require that we have a consistent view across a quorum of nodes instead of just getting whatever the first node we can talk to tells us.
2. Basically the changes in https://github.com/palantir/atlasdb/pull/3495 with also fixing the clean cassandra lock state test to actually create two lock tables and delete one, and wiring in a custom implementation of the conservative retrying logic that does not back off to make it fast.
5. If we reach the retrying limit, throw a specific exception instead of throwing the last encountered exception. All the encountered exceptions are added as suppressed exceptions. Also removed the logic depending on the thrown exception being TTransportException since that was nondeterministic anyway.

**Testing (What was existing testing like?  What have you done to improve it?)**:
We test that the behaviour is as expected in case of a quorum and less than a quorum of nodes up. In particular, we also change that the cassandra schema did not change in any of the calls where fewer than a quorum of nodes were up.

**Where should we start reviewing?**:
Maybe start with tests that should document the behaviour and make sure it's what we actually want.

**Priority (whenever / two weeks / yesterday)**:
Ideally today or tomorrow morning, but it's quite a lot

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3480)

PR #3495, opened 2018-09-07
Refactor Cassandra multinode tests

gmaretic:
**Goals (and why)**:
Need to write some new tests for the quorum schema changes, but the existing tests were in bad shape. Also, fix https://github.com/palantir/atlasdb/issues/3390

Each test class gets its own namespace so there is no pollution across tests. Initialization needs to be done before the cluster is degraded, so we use a bit of reflection to do that. 

Along the way noticed that table truncation uses its own retrying strategy for no discernable reason, so refactored that as well -- any concerns there? Also, removed the flake retrying -- if it still flakes,  might be fixed in the follow-up PR, and if not we can reintroduce it (but the old way we did it was broken). 

Not critical but would be nice

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3495)


sandorw:
@gmaretic What's the status on this?

gmaretic:
These are subsumed by https://github.com/palantir/atlasdb/pull/3480

Issue #3489, opened 2018-09-05
[TimeLock] OkHttp calls should be cancelled with Call.cancel(), not just Thread.interrupt()


jeremyk-91:
Internal reference: PDS-73051

If a TimeLock server is unresponsive but still listening on a port (e.g. if it is undergoing severe GC pressure, is out of memory, or didn't start or stop properly) then under high load it is possible for the TimeLock leader to accumulate a large number of threads and eventually OOM on being unable to allocate new threads.

Internal metrics revealed that this is caused by the leader election executor (which is unbounded and makes heavy use of PaxosQuorumChecker(PQC)). PQC attempts to talk to a quorum of nodes, and interrupts other threads to cancel requests once it has successfully got a consensus. However, this interruption mechanism doesn't look like it works well with OkHttp - the maintainers have suggested that `Call.cancel()` should be used instead.

This manifests in part as threads taking some time to finish once cancelled. This can be problematic on internal stacks where the latest Paxos sequence number endpoints can be queried at 8,000 QPS or higher; if the numbers in our test (40% after 3 seconds, 80% after 2 seconds, 90% after 1 second) scale, then the equilibrium population of threads here could be in the tens of thousands which is certainly not intentional, and in most cases unreasonable. Using `Call.cancel()`, threads clear up relatively much more quickly.

Separately, smart batching of Paxos sequence checks can be a good thing to do.

### Reproduction
In this test I patch TimeLock to sleep for one hour on any inbound request. I then call the TimeLock server using a separate Java client.

```
    private static void adhcTest() {
        System.out.println("Pingable Leader Test.");
        System.out.println("Initial thread count: " + EX.getActiveCount());
        for (int it = 0; it < 1; it++) {
            PingableLeader pl = AtlasDbHttpClients.createProxy(
                    new MetricRegistry(),
                    Optional.empty(),
                    "http://localhost:8421",
                    PingableLeader.class);
            List<Future<Boolean>> futures = IntStream.range(0, 3_000)
                    .boxed()
                    .map(unused -> EX.submit(pl::ping))
                    .collect(Collectors.toList());
            System.out.println("Thread count: " + EX.getActiveCount() + ". Cancelling...");
            futures.forEach(f -> f.cancel(true));
            System.out.println("All cancelled.");
            System.out.println("Thread count: " + EX.getActiveCount() + ". Waiting 10s.");
            for (int i = 0; i <= 10; i++) {
                System.out.println(
[...]

PR #3496, opened 2018-09-07
Add backoff when leadership check hits NO_QUORUM

stefank-palantir:
When calling determineLeadershipStatus during blockOnBecomingLeader results in a NO_QUORUM response, wait for a randomly chosen period before re-requesting.

**Goals (and why)**:

We recently encountered an issue where a deployment on legacy clustered lock server hit a thread leak – two of the three lock servers spun up to 30k+ threads and then OOMed. See PDS-71980.

I think it's possible for this to happen as follows:
* blockOnBecomingLeader calls determineLeadershipStatus
* Ultimately this calls PaxosLatestRoundVerifierImpl.isLatestRound
* This calls PaxosQuorumChecker.collectResponses, which spins up a new thread on the executor service for each verifier to be contacted.
* collectResponses blocks until either a quorum of results is obtained, or enough failures occur that no quorum can be achieved.
* In the latter case, we try to reap any outstanding request threads by interrupting them, but this is no guarantee that the interrupt is actually respected (in the case of PDS-71980 the many threads were blocking on obtaining a socket from a socket pool).
* These failures can potentially occur rapidly – if the nack occurs because we received a serialised exception from the remote server, this short-circuit might happen right away.
* This results in a NO_QUORUM return value within blockOnBecomingLeader. At present, when this happens we immediately `continue` in the while loop and call determineLeadershipStatus again.
* So I think potentially we could be hard looping through this code path and leaking request threads each round. At the least, in this case we'd be spamming the other nodes in the cluster.
* So I think backing off makes sense.

* Add a noQuorumMaxDelay parameter to PaxosLeadershipElectionService. 
* Default to 2s but allow this to be configured in PaxosLeadershipElectionServiceBuilder.
* When we hit a NO_QUORUM response, wait for a uniform period between 0s and noQuorumMaxDelay.

* I haven't added any tests here. I'm not sure there's an easy way to test this, and it feels relatively trivial. Happy to add some UTs if the reviewer disagrees and has suggestions for sensible tests.

* Is there anywhere else I should pipe the noQuorumMaxDelay value through from? Or is the builder the right entry point?

* It's pretty short – I think it's self explanatory.

* A week or so should be fine :) Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3496)

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -66,6 +67,7 @@ import com.palantir.paxos.PaxosValue;
  */
 public class PaxosLeaderElectionService implements PingableLeader, LeaderElectionService {
     private static final Logger log = LoggerFactory.getLogger(PaxosLeaderElectionService.class);
+    static final int DEFAULT_NO_QUORUM_MAX_DELAY_MS = 2000;
 
     private final ReentrantLock lock;
     private final CoalescingPaxosLatestRoundVerifier latestRoundVerifier;
@@ -79,6 +81,7 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
 
     final long updatePollingRateInMs;
     final long randomWaitBeforeProposingLeadership;
+    final long noQuorumMaxDelay;
     final long leaderPingResponseWaitMs;
 
     final ExecutorService executor;
@@ -98,8 +101,9 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
                                       long randomWaitBeforeProposingLeadership,
                                       long leaderPingResponseWaitMs) {
         this(proposer, knowledge, potentialLeadersToHosts, acceptors, learners, executor,
-                updatePollingWaitInMs, randomWaitBeforeProposingLeadership, leaderPingResponseWaitMs,
-                PaxosLeaderElectionEventRecorder.NO_OP);
+                updatePollingWaitInMs, randomWaitBeforeProposingLeadership,
+                DEFAULT_NO_QUORUM_MAX_DELAY_MS,
+                leaderPingResponseWaitMs, PaxosLeaderElectionEventRecorder.NO_OP);
     }
 
     PaxosLeaderElectionService(PaxosProposer proposer,
@@ -110,6 +114,7 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
             ExecutorService executor,
             long updatePollingWaitInMs,
             long randomWaitBeforeProposingLeadership,
+            long noQuorumMaxDelay,
             long leaderPingResponseWaitMs,
             PaxosLeaderElectionEventRecorder eventRecorder) {
         this.proposer = proposer;
@@ -121,6 +126,7 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
         this.executor = executor;
         this.updatePollingRateInMs = updatePollingWaitInMs;
         this.randomWaitBeforeProposingLeadership = randomWaitBeforeProposingLeadership;
+        this.noQuorumMaxDelay = noQuorumMaxDelay;
         this.leaderPingResponseWaitMs = leaderPingResponseWaitMs;
         lock = new ReentrantLock();
         this.eventRecorder = eventRecorder;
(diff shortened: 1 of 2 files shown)


tboam:
Jeremy was investigating a similar issue of thread explosion causing OOMs. Would like to check if this is the same problem and whether you've come to the same conclusion.

jeremyk-91:
Hey @stefank-palantir, this looks great! Yep, immediate retry looks bogus - I'll have a closer look on Monday when I get in.

@tboam I think this is related, though not the same. #3489 is on a slightly different path - that ticket was "I have quorum but the last node is very slow", while this is "I have no quorum". One part of the root cause is similar - as Stefan mentions "this is no guarantee that the interrupt is actually respected", and the experiments on that ticket establish that the interrupt is not respected. 

When we do have a quorum we also try to cancel our requests on the other nodes via interrupts, which can be a problem if we have a deluge of calls (and we did).

jeremyk-91:
(This should go on develop at some point, too.)

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -141,6 +145,8 @@ public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
                     return new PaxosLeadershipToken(greatestLearned);
                 } else if (leadingStatus == StillLeadingStatus.NO_QUORUM) {
                     // If we don't have quorum we should just retry our calls.
+                    long backoffTime = (long) (noQuorumMaxDelay * Math.random());
+                    log.debug("Waiting for [{}] ms before rerequesting leadership status", SafeArg.of("waitTimeMs", backoffTime));
```
schlosna: Where does it actually backoff?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -98,18 +100,19 @@ public PaxosLeaderElectionService(PaxosProposer proposer,
                                       long randomWaitBeforeProposingLeadership,
                                       long leaderPingResponseWaitMs) {
         this(proposer, knowledge, potentialLeadersToHosts, acceptors, learners, executor,
-                updatePollingWaitInMs, randomWaitBeforeProposingLeadership, leaderPingResponseWaitMs,
-                PaxosLeaderElectionEventRecorder.NO_OP);
+                updatePollingWaitInMs, randomWaitBeforeProposingLeadership, ,
+                DEFAULT_NO_QUORUM_MAX_DELAY_MS,
+                leaderPingResponseWaitMs, PaxosLeaderElectionEventRecorder.NO_OP);
```
schlosna: Why the double comma?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
stefank-palantir: ... whoops. 

Sorry, that's a pretty big slip. Fixed, thanks for spotting >_>

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
stefank-palantir: Argument mismatch; fixed. Thanks!

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionServiceBuilder.java
```diff
@@ -84,6 +85,11 @@ public PaxosLeaderElectionServiceBuilder leaderPingResponseWaitMs(long leaderPin
...
     }
 
+    public PaxosLeaderElectionServiceBuilder noQuorumMaxDelayMs(long noQuorumMaxDelayMs) {
+        this.noQuorumMaxDelayMs = noQuorumMaxDelayMs;
+        return this;
+    }
+
     public PaxosLeaderElectionServiceBuilder eventRecorder(PaxosLeaderElectionEventRecorder eventRecorder) {
         this.eventRecorder = eventRecorder;
         return this;
```
schlosna: I can't comment on the below `build()`, but do you need to pipe the `noQuorumMaxDelayMs` into the `PaxosLeaderElectionService` constructor?

stefank-palantir: I agree that I do. Fixed, thanks.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -141,6 +146,9 @@ public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
                     return new PaxosLeadershipToken(greatestLearned);
                 } else if (leadingStatus == StillLeadingStatus.NO_QUORUM) {
                     // If we don't have quorum we should just retry our calls.
+                    long backoffTime = (long) (noQuorumMaxDelay * Math.random());
```
jeremyk-91: A little nitpicky: I'd probably write this differently, since if I'm not wrong this might be called on multiple threads, in which case the calls to Math.random() would contend. I'd prefer `ThreadLocalRandom.current().nextLong(0, noQuorumMaxDelay)` here.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionServiceBuilder.java
```diff
@@ -99,6 +105,7 @@ public PaxosLeaderElectionService build() {
                 executor,
                 pingRateMs,
                 randomWaitBeforeProposingLeadershipMs,
+                noQuorumMaxDelayMs,
```
jeremyk-91: OK. Could we do a preconditions-check that this is non-negative? Otherwise if someone gives a negative value we'll only fail when we try to generate the range / in current impl, when we sleep for negative time.

gmaretic:
I opened #3796 to trigger the circle build

PR #3498, opened 2018-09-09
Timelock safety checks

j-baker:
**Goals (and why)**: Timelock has some sharp edges. Let's dull them.

1. Make it impossible to overflow. You can deadlock yourself, but that is fixable. Overflow is not.
2. Have users indicate whether this is a new instantiation of the service. It's not cool for timelock to create the data dir if it's not a new service. It won't deal with every case, but it seems like it'd provide a lot of real world mileage.

This class is only used internally and so can only be tested internally. It should need a default config change, but nothing more.

Whether there are easier ways to achieve the same effect.

Probably worth getting this in pretty fast.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3498)

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
+         - A few timelock ops edge cases have been removed. Timelock users must now indicate whether they are booting their
+           servers for the first time or subsequent times, to avoid the situation where a timelock node becomes newly
+           misconfigured and thinks it is booting up for the first time again.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3498>`__)
+
     *    - |fixed|
          - The Jepsen tests no longer assume that users have installed Python or DateUtil, and will install these itself if needed.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3461>`__)
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -53,7 +53,9 @@ develop
     *    - |improved|
          - A few timelock ops edge cases have been removed. Timelock users must now indicate whether they are booting their
            servers for the first time or subsequent times, to avoid the situation where a timelock node becomes newly
-           misconfigured and thinks it is booting up for the first time again.
+           misconfigured and thinks it is booting up for the first time again. Additionally, timestamps no longer overflow
+           when they hit Long.MAX_VALUE; this would only happen due to a bug, but at least now the DB will become read only
+           and not corrupt.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3498>`__)
 
     *    - |fixed|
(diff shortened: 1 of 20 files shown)


timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
```diff
@@ -27,8 +27,31 @@ default File dataDirectory() {
...
+                            + "be set to false for safety reasons.");
+        }
+
+        if (!isNewService() && !dataDirectory().isDirectory()) {
+            throw new IllegalArgumentException(
+                    "If this is a new timelock service, please configure paxos.is-new-service to true for the first "
+                            + "startup only of each node. \n\nOtherwise, the timelock data directory appears to no "
+                            + "longer exist. If you are trying to move the nodes on your timelock cluster or add new "
+                            + "nodes, you have likely already made a mistake by this point. This is a non-trivial "
+                            + "operation, please be careful and do not be afraid to seek out help.");
```
tboam: I think we can be more direct here and instruct people to speak to the Atlas team before proceeding.

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
```diff
@@ -27,8 +27,31 @@ default File dataDirectory() {
         return new File("var/data/paxos");
     }
 
+    /**
+     * Set to true if this is a new stack. Otherwise, set to false.
+     */
+    @JsonProperty("is-new-service")
+    boolean isNewService();
```
gokcanozakdag: what would be the default value here?

jeremyk-91: There is no default value, and we'll throw an ISE if we try and build an object without it.

This is probably reasonable - in internal distributions we can specify it explicitly (probably `false`), and users can override it as needed using standard internal protocols.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
+         - A few timelock ops edge cases have been removed. Timelock users must now indicate whether they are booting their
+           servers for the first time or subsequent times, to avoid the situation where a timelock node becomes newly
+           misconfigured and thinks it is booting up for the first time again.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3498>`__)
+
```
jeremyk-91: nit: The overflow thing is a win and maybe worth calling out (you could say that it's unlikely anyone actually hits this, to be fair, but I think it's still nice)

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
j-baker: i've cited 'appropriate approvals' - in short these messages with a 'who' on the end have typically ended up fairly out of date...

timelock-agent/src/test/java/com/palantir/timelock/config/PaxosInstallConfigurationTest.java
```diff
@@ -28,45 +27,65 @@
...
+    }
+
+    @Test
+    public void throwsIfConfiguredToBeExistingServiceWithoutDirectory() {
+        File mockFile = getMockFileWith(false, true);
+
+        assertThatThrownBy(ImmutablePaxosInstallConfiguration.builder()
+                .dataDirectory(mockFile)
+                .isNewService(false)
+                ::build).isInstanceOf(IllegalArgumentException.class);
```
jeremyk-91: nit: we can probably extract a method for the assertion?

PR #3500, opened 2018-09-10
default oracle config migration state to FINISHED to match docs

jochs:
**Goals (and why)**:

Introduce a default for OracleDdlConfig's overflowMigrationState which matches the documentation specification for new clients ([here](https://github.com/palantir/atlasdb/blob/f61dde689afb75281b97299751359f973f3ed259/docs/source/configuration/key_value_service_configs/oracle_key_value_service_config.rst#configuration-parameters)).

This will keep new clients from needing to specify something that only matters for legacy clients, and will decrease confusion around Oracle config overrides.

Is there concern around old clients upgrading and inheriting the incorrect value? My thought is that there's been sufficient time to upgrade/I'm not sure if anybody relies on this anymore.

Eventually

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3500)


sandorw:
@jochs what drove this? I don't know the correct behavior here without digging, and wouldn't know, for example, whether the docs are even correct here.

Issue #2559, opened 2017-10-25
AtlasConsole getRange() can OOM on large tables


dxiao:
When running large range reads, Atlas Console OOMs. Atlas Console records every put/get and stores it in a list, which can be retrieved by calling currentTransaction() and it only cleared in a few places (not in the middle of a big range read for example)

The entire point of console is that we can answer questions about our DB, so making full table scans OOM is a huge problem.

internal ticket 98743

tboam:
@dxiao we're not going to have time to work on this for several weeks, is this something you want to take a look at?

stefank-palantir:
Bump - it's a bit painful to have to deliver bespoke CLIs to diagnose DB issues. Would be way better to get this in Atlas if possible.

PR #3502, opened 2018-09-11
Issue 2559 - AtlasConsole getRange can OOM

stefank-palantir:
**Goals (and why)**:

This PR is a speculative fix to https://github.com/palantir/atlasdb/issues/2559 – table(xyz).getRange() can OOM AtlasConsole.

I have a theory that this is issue is caused because the default batch size of queries
is 2000, which feels a little wide. It's also unfortunate that the batch size isn't configurable.

This commit brings that default down to 100, and adds a configurable batch size
for when we want to trade memory for performance. In future the user can simply pass `batch_size` into the range map.

* Table.getRange actually already passes all the range parameters through to atlas.
* batch_size is a parameter that Atlas will expect and handle.
* However the existing Table.getRange code will convert all range parameters to lists, because `start` and `end` need to be lists but AtlasConsole allows the user to give a singleton.
* This doesn't make sense for `batch_size` – so if the user provides it, atlas console will crash because Atlas doesn't want a list.
* I've added a list of parameters which `Table.getRange` _should_ uplift to lists. The rest will be passed unaltered.
* I've also modified `Table.getRange` to pass a default batch size of 100, which feel safer.

*** I haven't tested this at all. I would really appreciate a conversation around how best to test this, either in code or on a test rig. :) ***

Primarily testing, or maybe alternative methods for providing a batch size if folks have better suggestions.

It's a one-class change

At your convenience.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3502)

### DIFF
--- a/atlasdb-console/src/main/groovy/com/palantir/atlasdb/console/module/Table.groovy
+++ b/atlasdb-console/src/main/groovy/com/palantir/atlasdb/console/module/Table.groovy
@@ -30,6 +30,8 @@ class Table {
     boolean mutationsEnabled
 
     private static int DEFAULT_JOIN_BATCH_SIZE = 10000;
+    private static final int DEFAULT_RANGE_BATCH_SIZE = 100;
+    private static final String[] DEFAULT_LIST_QUERY_PARAMS = ["start", "end"];
 
     Table(String name, AtlasConsoleServiceWrapper service, boolean mutationsEnabled) {
         this.name = name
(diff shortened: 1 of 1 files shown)


stefank-palantir:
*Let's not merge yet* – per the initial PR comment I haven't done any testing. I just wanted to get some code up for discussion.

I'll follow up with you guys out of band about what sensible ways to test this look like – I'm not sure how you would usually go about manual testing so would welcome some pointers :)

sandorw:
@stefank-palantir what's the status of the testing you were going to do?

stefank-palantir:
Still on me, sorry. Hoping to get to it in the next couple of weeks.

sandorw:
Closing since this hasn't gotten any further response

PR #3505, opened 2018-09-13
Distinct Bounded Thread Pools for PaxosLeaderElectionService

jeremyk-91:
**Goals (and why)**:
- #3489 proved to be very difficult to fix directly. This is a workaround which reasonably caps the number of excess threads, and as far as I know still allows the remaining two nodes that are fast to communicate and make Paxos-y decisions without too much of a perf hit.

- Implement a `MultiplexingCompletionService` which is basically an `ExecutorCompletionService` but has multiple sources feeding into the one queue. (You can do this directly with `ExecutorCompletionService` and manipulating the blocking queue, but I thought it worthwhile to keep the complexity in one place).
- Switch PaxosLeaderElectionService to use bounded executors.

- Leader election is already tested a fair bit.
- The multiplexing completion service is tested quite a bit, using `DeterministicScheduler` for the most part, but also with a few tests that check that we are able to deal with high load if using bounded pools.

- For now I picked a core pool size of 1 and max of 100; this is based on the idea that metrics tend to indicate general activity of <10 under normal circumstances. Should these be configurable from the beginning?

**Where should we start reviewing?**: MultiplexingCompletionService

**Priority (whenever / two weeks / yesterday)**: one week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3505)


atlasdb-commons/src/main/java/com/palantir/common/concurrent/MultiplexingCompletionService.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Submits a task to be run on a specific executor.
+     *
+     * @param key to identify which executor the task should be run on
+     * @param task to be run on the relevant executor
+     * @return future associated with submitting the task to the correct executor
+     *
+     * @throws IllegalStateException if the key provided is not associated with any executor
+     */
+    public Future<V> submit(T key, Callable<V> task) {
```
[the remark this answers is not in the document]
gokcanozakdag: great, thanks 👍

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -148,21 +152,29 @@ public static LocalPaxosServices createInstrumentedLocalServices(
...
-                        .setNameFormat("atlas-leaders-election-%d")
+        // TODO (jkong): Make the limits configurable.
+        // Current use cases tend to have not more than 10 inflight tasks under normal circumstances.
+        Function<String, ExecutorService> leaderElectionExecutor = (useCase) -> new InstrumentedExecutorService(
+                PTExecutors.newThreadPoolExecutor(
+                        1,
+                        100,
+                        5000,
+                        TimeUnit.MILLISECONDS,
+                        new LinkedBlockingQueue<>(),
```
jeremyk-91: actually, the work queue needs to be bounded, otherwise _that_ will become incredibly long and we could waste heap/oom

schlosna: The queue used here has to be bounded for the pool to ever increase thread count as new threads are only created when queue.offer returns false

jeremyk-91: Yep. I've switched to use SynchronousQueue which was more in line with how I imagined this would work (also don't want to cause perf/latency hit under low loads)

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -148,21 +152,29 @@ public static LocalPaxosServices createInstrumentedLocalServices(
...
                 leaderUuid, proposerExecutorService));
 
-        InstrumentedExecutorService leaderElectionExecutor = new InstrumentedExecutorService(
-                PTExecutors.newCachedThreadPool(new ThreadFactoryBuilder()
-                        .setNameFormat("atlas-leaders-election-%d")
+        // TODO (jkong): Make the limits configurable.
+        // Current use cases tend to have not more than 10 inflight tasks under normal circumstances.
+        Function<String, ExecutorService> leaderElectionExecutor = (useCase) -> new InstrumentedExecutorService(
+                PTExecutors.newThreadPoolExecutor(
+                        1,
```
schlosna: probably want core pool size to be the number of other potential leaders

jeremyk-91: Yep, that sounds right!

PR #3508, opened 2018-09-14
Targeted sweeper ignores queued writes where the table has been deleted.

dxiao:
**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3508)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
@@ -71,7 +71,8 @@ public class SweepQueueDeleter {
                         });
             } catch (IllegalArgumentException e) {
                 // hack because InMemoryKVS doesn't throw a TableMappingNotFoundException...
-                if (e.getCause() instanceof TableMappingNotFoundException || e.getMessage().endsWith("does not exist")) {
+                if (e.getCause() instanceof TableMappingNotFoundException
+                        || e.getMessage().endsWith("does not exist")) {
                     log.debug("Could not resolve full name for table reference {}, "
                             + "assuming table has been deleted and therefore relevant cells as well.",
                             LoggingArgs.tableRef(entry.getKey()));
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
@@ -21,13 +21,20 @@ import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.collect.Iterables;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.impl.TableMappingNotFoundException;
+import com.palantir.atlasdb.logging.LoggingArgs;
 import com.palantir.atlasdb.sweep.Sweeper;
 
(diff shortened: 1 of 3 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweepTest.java
```diff
@@ -194,6 +194,18 @@ public void sweepDeletesAllIfLatestCommittedTombstoneInThorough() {
         assertNoEntryForCellInKvs(TABLE_THOR, TEST_CELL);
     }
 
+    @Test
+    public void sweepThrowsAwayWritesForDroppedTables() {
+        useOneSweepQueueShard();
+        writeInTransactionAndGetStartTimestamp(SINGLE_WRITE);
+        Long startTimestamp = writeInTransactionAndGetStartTimestamp(SINGLE_WRITE);
+
+        keyValueService.dropTable(TABLE_CONS);
```
dxiao: fun story - you can't put this test in TargetedSweeperTest because the kvs in that one has automaticallyCreateTables = true, so this is useless.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
```diff
@@ -48,19 +55,32 @@
...
+                                kvs.addGarbageCollectionSentinelValues(entry.getKey(),
+                                        maxTimestampByCellPartition.keySet());
+                                kvs.deleteAllTimestamps(entry.getKey(), maxTimestampByCellPartition, false);
+                            } else {
+                                kvs.deleteAllTimestamps(entry.getKey(), maxTimestampByCellPartition, true);
+                            }
+                        });
+            } catch (IllegalArgumentException e) {
+                // hack because InMemoryKVS doesn't throw a TableMappingNotFoundException...
+                if (e.getCause() instanceof TableMappingNotFoundException
```
tboam: Is there any other reason that a `TableMapingNotFoundryException` could be thrown?  Could a table be renamed and the name in the TS queue is now out of date?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
```diff
@@ -48,19 +55,32 @@
...
+                        || e.getMessage().endsWith("does not exist")) {
+                    log.debug("Could not resolve full name for table reference {}, "
```
tboam: Could log at a higher priority here.

I feel slightly nervous about just dropping the items, but we don't have a concept of a Dead Letter Queue for storing failed messages.  Is it worth logging the cells here?  Am I being over cautious here?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
dxiao: We allow for renames?

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,10 @@ develop
     *    - Type
          - Change
 
+    *    - |fixed|
+         - Targeted sweep no longer chokes if the table for which a write it is trying to sweep no longer exists.
```
sandorw: -> "Targeted sweep no longer chokes if a table in the queue no longer exists" is clearer I think

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
sandorw: There's some stuff around tablename encoding or case (see for example https://github.com/palantir/atlasdb/pull/3278), but I would expect all of it to be encapsulated in the KVS and for any `TableMappingNotFoundException`s to be legitimate (or at least, I would hope, given the few related things I've seen...). Oracle has some sort of table name mapping stuff as well, but it again is all underneath the KVS layer.

PR #3509, opened 2018-09-14
Gozakdag/leadership observer

gokcanozakdag:
**Goals (and why)**:
We observed that "isCurrentSuspectedLeader" tag is stuck to be 'true', even after losing leadership. This pr mainly aims to fix this issue.

In a timelock cluster of A, B, C; let's pick A as the leader first. In that case, any request coming to timelockServices in A will be instrumented, causing to publish a metric with `isCurrentSuspectedLeader` set to `true`. If A loses leadership to B, B will start to publish those metrics, but A will continue to emit those metrics with `isCurrentSuspectedLeader`, and it will not set it to false (as we are only applying the tag function if request is successful, and requests coming to a follower will not succeed). Even if it was set to false, we would end up having two separate tags, one set to true, one set to false. So the solution is to deregister those metrics from the node, when it loses leadership.

This pr implements a `LeadershipObserver`, which is attached to `PaxosLeaderElectionEventsRecorder`. Any class having access to this observer can now submit tasks that will be executed on either losing or gaining leadership. Tasks are executed async in a separate thread; as I did not want to block paxos processes in case of having a long running task submitted to `LeadershipObserver`.

Currently we are only submitting the task that deregisters the leadership metrics when leadership is lost.

- Unit tests for `AsyncLeadershipObserver`.
- Addition to unit tests for LeadershipEventRecorder to make sure observer is notified.

- Is the executorService used correctly? Is it an overkill to use separate thread for the observer.
- Can I have the same functionality with less change on public API's?

`LeadershipObserver`

A week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3509)

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
@@ -17,24 +17,35 @@
 package com.palantir.leader;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.palantir.common.concurrent.PTExecutors;
 
 public class AsyncLeadershipObserver implements LeadershipObserver {
 
-    private ExecutorService executorService = PTExecutors.newSingleThreadExecutor(true);
-    private List<Runnable> leaderTasks = new ArrayList<>();
-    private List<Runnable> followerTasks = new ArrayList<>();
+    private final ExecutorService executorService;
+    private final List<Runnable> leaderTasks = Collections.synchronizedList(new ArrayList<>());
+    private final List<Runnable> followerTasks = Collections.synchronizedList(new ArrayList<>());
+
+    public static AsyncLeadershipObserver create() {
+        return new AsyncLeadershipObserver(PTExecutors.newSingleThreadExecutor(true));
+    }
+
+    @VisibleForTesting
+    AsyncLeadershipObserver(ExecutorService executorService) {
+        this.executorService = executorService;
+    }
 
     @Override
-    public void gainedLeadership() {
+    public synchronized void gainedLeadership() {
         executorService.execute(this::executeLeaderTasks);
     }
 
     @Override
-    public void lostLeadership() {
+    public synchronized void lostLeadership() {
         executorService.execute(this::executeFollowerTasks);
     }
 
--- a/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
@@ -30,6 +30,10 @@ public class AsyncLeadershipObserver implements LeadershipObserver {
     private final List<Runnable> followerTasks = new CopyOnWriteArrayList<>();
 
     public static AsyncLeadershipObserver create() {
+        /**
+         * This executor should be kept single threaded to ensure ordered execution of tasks
+         * which is guaranteed by the API contract of {@link LeadershipObserver}
+         */
         return new AsyncLeadershipObserver(PTExecutors.newSingleThreadExecutor(true));
     }
 
--- a/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
@@ -16,9 +16,8 @@
 
 package com.palantir.leader;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -27,8 +26,8 @@ import com.palantir.common.concurrent.PTExecutors;
 public class AsyncLeadershipObserver implements LeadershipObserver {
 
     private final ExecutorService executorService;
-    private final List<Runnable> leaderTasks = Collections.synchronizedList(new ArrayList<>());
-    private final List<Runnable> followerTasks = Collections.synchronizedList(new ArrayList<>());
+    private final List<Runnable> leaderTasks = new CopyOnWriteArrayList<>();
+    private final List<Runnable> followerTasks = new CopyOnWriteArrayList<>();
 
     public static AsyncLeadershipObserver create() {
         return new AsyncLeadershipObserver(PTExecutors.newSingleThreadExecutor(true));
@@ -41,12 +40,12 @@ public class AsyncLeadershipObserver implements LeadershipObserver {
 
     @Override
     public synchronized void gainedLeadership() {
-        executorService.execute(this::executeLeaderTasks);
+        executorService.execute(() -> leaderTasks.forEach(Runnable::run));
     }
 
     @Override
     public synchronized void lostLeadership() {
-        executorService.execute(this::executeFollowerTasks);
+        executorService.execute(() -> followerTasks.forEach(Runnable::run));
     }
 
     @Override
@@ -57,13 +56,5 @@ public class AsyncLeadershipObserver implements LeadershipObserver {
     @Override
     public synchronized void executeWhenLostLeadership(Runnable task) {
         followerTasks.add(task);
-    }
-
-    private void executeFollowerTasks() {
-        followerTasks.forEach(Runnable::run);
-    }
-
-    private void executeLeaderTasks() {
-        leaderTasks.forEach(Runnable::run);
     }
 }
--- a/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
@@ -24,7 +24,7 @@ import com.palantir.common.concurrent.PTExecutors;
 
 public class AsyncLeadershipObserver implements LeadershipObserver {
 
-    private ExecutorService executorService = PTExecutors.newSingleThreadExecutor();
+    private ExecutorService executorService = PTExecutors.newSingleThreadExecutor(true);
     private List<Runnable> leaderTasks = new ArrayList<>();
     private List<Runnable> followerTasks = new ArrayList<>();
 
--- a/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
```diff
@@ -119,7 +125,19 @@ private AsyncTimelockService createRawAsyncTimelockService(
...
                 metricRegistry, serviceClass, service, MetricRegistry.name(serviceClass),
                 context -> ImmutableMap.of(
-                        "client", client,
-                        "isCurrentSuspectedLeader", String.valueOf(leadershipCreator.isCurrentSuspectedLeader())));
+                        AtlasDbMetricNames.TAG_CLIENT, client,
+                        AtlasDbMetricNames.TAG_CURRENT_SUSPECTED_LEADER, String.valueOf(
+                                leadershipCreator.isCurrentSuspectedLeader())));
+    }
+
+    private void deregisterLeaderMetrics() {
```
tboam: I think this should be pushed down into the `MetricsManager` class, it stores an additional list of metrics which should be kept up to date.  Have a `deregisterTaggedMetrics(Predicate<MetricName> filterCondition)` method.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeadershipEventRecorder.java
```diff
@@ -27,6 +30,7 @@
 
     private final String leaderId;
     private final LeadershipEvents events;
+    private final List<LeadershipObserver> leadershipObservers;
```
tboam: Seems a little odd that this has a `List<LeadershipObserver>` and the only implementation of that contains a `List<Runnable>`.  I'd try to only have one level of multiplicity.

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
```diff
@@ -84,6 +88,8 @@ public TimeLockServices createTimeLockServices(
                         : JavaSuppliers.compose(NonTransactionalLockService::new, rawLockServiceSupplier),
                 client);
 
+        leadershipCreator.executeWhenLostLeadership(this::deregisterLeaderMetrics);
```
tboam: Do we need a similar method when we gain leadership to deregister any metrics with `isCurrentSuspectedLeader=false`?

leader-election-impl/src/main/java/com/palantir/leader/LeadershipObserver.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.leader;
+
+public interface LeadershipObserver {
```
tboam: I don't really like being able to change the list of tasks during the lifecycle of this object.  I'd rather the `Runnable`s were passed in at construction time and then couldn't be changed.

leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+
+import com.palantir.common.concurrent.PTExecutors;
+
+public class AsyncLeadershipObserver implements LeadershipObserver {
+
+    private ExecutorService executorService = PTExecutors.newSingleThreadExecutor(true);
```
tboam: It is key that the tasks are carried out in order.  The `newSingleThreadExecutor` guarantees this so this is correct, but it could be worth adding a comment to make this clear to future readers.

leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
```diff
@@ -0,0 +1,58 @@
...
+        followerTasks.add(task);
+    }
+
+    private void executeFollowerTasks() {
+        followerTasks.forEach(Runnable::run);
+    }
+
+    private void executeLeaderTasks() {
+        leaderTasks.forEach(Runnable::run);
+    }
```
jeremyk-91: I think we should use a SynchronizedList here. The main issue is one of visibility; if in one thread I register a callback and later in another thread I call executeLeaderTasks(), if I'm not wrong it is actually permissible to do nothing; there isn't a happens-before relationship between the read of `leaderTasks` when scheduling them, and the write the callback made. 

Also, if someone registers a callback while we're iterating I think it's possible to get `ConcurrentModificationException` here.

leader-election-impl/src/main/java/com/palantir/leader/LeadershipObserver.java
```diff
@@ -0,0 +1,25 @@
...
+    void gainedLeadership();
+    void lostLeadership();
+
+    void executeWhenGainedLeadership(Runnable task);
+    void executeWhenLostLeadership(Runnable task);
```
jeremyk-91: I think for this class in general, we should be more explicit (javadoc?) about what guarantees we are giving the user especially if there are multiple runnables submitted. We can be explicit that we are not guaranteeing much (for example, even though right now it looks like the callbacks are executed in order sequentially, we should be explicit as to whether we allow users of this to assume this behaviour.)

leader-election-impl/src/test/java/com/palantir/leader/AsyncLeadershipObserverTest.java
```diff
@@ -0,0 +1,76 @@
...
+        verify(followerTask, times(0)).run();
+    }
+
+    @Test
+    public void doNotRunLeaderTasksAfterLosingLeadership() {
+        leadershipObserver.executeWhenGainedLeadership(leaderTask);
+        loseLeadership();
+
+        verify(leaderTask, times(0)).run();
+    }
```
jeremyk-91: There's a trick that can be useful for condensing these tests down a little. Mockito has a `verifyNoMoreInteractions(Object...)` method that checks that nothing else has run on the mocks you give to it. You could then call that for both the leader and follower tasks on each test.

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
jeremyk-91: It could be useful. I don't think we're likely to watch the `false` metrics as closely (e.g. we're unlikely to use them for monitors), but it would be good to eliminate potentially confusing information.

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
gokcanozakdag: done!

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeadershipEventRecorder.java
gokcanozakdag: Makes sense to me; I changed it to keep an Optional<LeadershipObserver> rather than keeping a list of observers.

leader-election-impl/src/main/java/com/palantir/leader/LeadershipObserver.java
gokcanozakdag: It is a bit hard to implement as construction of the observer and attaching the tasks takes place on different levels. I changed setters in observer to be synchronized, and also used a synchronized list to keep `Runnable`s to avoid race conditions.

leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
gokcanozakdag: Changed the implementation to use synchronized methods, and also utilize a synchronized list. Also to fix the tests, I used a `DeterministicScheduler` and waited until tasks are finished.

leader-election-impl/src/main/java/com/palantir/leader/LeadershipObserver.java
gokcanozakdag: Added javadoc 👍

leader-election-impl/src/test/java/com/palantir/leader/AsyncLeadershipObserverTest.java
gokcanozakdag: Thanks!

leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
jeremyk-91: If I'm not wrong the private methods still need to be synchronized, because the calls to `gainedLeadership` and `lostLeadership` submit things to an executor which might iterate over the list without holding the lock for the list (so you could still get a ConcurrentModificationException; a synchronized list doesn't protect us against that).

Since we don't register too many things it might be better to just use a `CopyOnWriteArrayList` - sorry for bad/ambiguous advice above.

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
```diff
@@ -84,13 +87,25 @@ public TimeLockServices createTimeLockServices(
...
                 asyncTimelockService,
                 lockService,
                 asyncOrLegacyTimelockService,
                 asyncTimelockService);
     }
 
+    private static Predicate<MetricName> withTagIsCurrentSuspectedLeader(boolean currentLeader) {
+        return metricName -> metricName.safeTags().containsKey(AtlasDbMetricNames.TAG_CURRENT_SUSPECTED_LEADER)
+                && metricName.safeTags().get(AtlasDbMetricNames.TAG_CURRENT_SUSPECTED_LEADER)
+                        .equals(String.valueOf(currentLeader));
```
jeremyk-91: nit: this is `Optional.ofNullable(metricName.safeTags().get(ADMN.TAG_CURRENT_SUSPECTED_LEADER)).filter(x -> x.equals(String.valueOf(currentLeader)).isPresent();` though I'll leave this as personal preference

leader-election-impl/src/main/java/com/palantir/leader/AsyncLeadershipObserver.java
```diff
@@ -0,0 +1,60 @@
...
+import com.palantir.common.concurrent.PTExecutors;
+
+public class AsyncLeadershipObserver implements LeadershipObserver {
+
+    private final ExecutorService executorService;
+    private final List<Runnable> leaderTasks = new CopyOnWriteArrayList<>();
+    private final List<Runnable> followerTasks = new CopyOnWriteArrayList<>();
+
+    public static AsyncLeadershipObserver create() {
+        return new AsyncLeadershipObserver(PTExecutors.newSingleThreadExecutor(true));
```
sandorw: Might be worth a comment since expanding this beyond one thread no longer guarantees ordered execution.

PR #3512, opened 2018-09-18
Shuffle Sweep Priority Overrides

### LINKED ISSUE #3511: Background Sweeper Priority Overrides Are Broken For Multiple Priority Tables
We accidentally repeatedly pick the first table, because of a bug in how we've implemented this where we iterate through a list in order and always pick the first one when deciding the next table to sweep.

jeremyk-91:
**Goals (and why)**:
- Address a bug found in PDS-74117
- Fix #3511 

- Shuffle the list of tables before trying to lock them.

- Existing testing did not have any tests that cover interactions between _multiple_ calls to the priority table. This change adds one test that used to fail but now passes.

- Shuffle is not an efficient algorithm (O(priority list size) instead of likely O(min(priority list size, number of sweep threads))) but I think it's fine as the priority list should be small anyway

**Where should we start reviewing?**: NextTableToSweepProviderTest

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3512)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
@@ -87,7 +87,7 @@ public class NextTableToSweepProvider {
 
         List<TableReference> tablesOrderedByPriority = orderTablesByPriority(tablesWithNonZeroPriority);
 
-        Optional<TableToSweep> chosenTable = attemptToChooseTableFromList(tablesOrderedByPriority,
+        Optional<TableToSweep> chosenTable = attemptToChooseTableFromPrioritisedList(tablesOrderedByPriority,
                 "it has a high priority score");
 
         return logDecision(chosenTable, scores, overrideConfig);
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/priority/NextTableToSweepProvider.java
```diff
@@ -96,6 +97,12 @@ public static NextTableToSweepProvider create(KeyValueService kvs, LockService l
         List<TableReference> priorityTableRefs = overrideConfig.priorityTablesAsList().stream()
                 .map(TableReference::createFromFullyQualifiedName)
                 .collect(Collectors.toList());
+
+        // If there are multiple priority tables, we don't want to consistently use the same ordering.
+        // It is true that this operation is O(list length) while an O(min(list length, cluster size)) algorithm
+        // exists, but the priority table reference list is expected to be small.
+        Collections.shuffle(priorityTableRefs);
+
         return attemptToChooseTableFromList(priorityTableRefs, "it is on the sweep priority list");
```
tboam: Detail: I would rename this method to "attemptToChooseTableFromPrioritisedList" to make this sort of bug a little clearer.

PR #3516, opened 2018-09-20
Add another fix for targeted sweep when another service has deleted t…

dxiao:
…he table.

And therefore the TableMapper has a reference to the table name, but the
underlying kvs can't actually find it.

**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3516)


dxiao:
Expedited review would be nice - I'd like to get this in ASAP so I can test on latest and verify this is the last thing that's preventing thorough sweep from being happy before monday when we branch.

PR #3520, opened 2018-09-21
The targeted sweep queue no longer contains json encoded data

j-baker:
This leads to additional space being used than necessary. In practice as
I understand Atlas, this is safe to be rolling upgraded around; older
nodes will start failing to sweep but once the cluster is upgraded
everything will be fine.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3520)


j-baker:
(3x when cells are small. It basically kills 80 bytes constant overhead, and then 1/3 the length of your cell (so between 0 and 1000 bytes depending)

PR #3524, opened 2018-09-24
Expose TimestampManagementService in TransactionManagers

JacekLach:
**Goals (and why)**:

Live-migration library needs to be able to fast-forward timestamps when it commits a migration, so that it can invalidate previous leases and abort long-running transactions that are currently in progress.

- Each `AtlasDbFactory` can now convert its own `TimestampService` into a `TimestampManagementService`. In every current case, it's just a cast.
- `TransactionManagers` wires the `TimestampManagementService` the same way it does for `TimestampService`, in each configuration option.

WIP :P

~~
- There's are very suspicious `time.get()`s in https://github.com/palantir/atlasdb/blob/aad75bc57db14f3358ec9ed943dbc05b3e289fe7/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L836-L840 and https://github.com/palantir/atlasdb/blob/aad75bc57db14f3358ec9ed943dbc05b3e289fe7/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L948-L949 . They cause more `Timestamp service fetched for a second time` errors when running tests. I think what I need to do here is wrap `time` into a memoizing supplier, but I'd like confirmation on that, because I'm not sure what the impact of running multiple timestamp service classes is.
~~

^ those are now memoized. Unclear how to verify this via testing.

`TransactionManagers` and the `*AtlasDbFactory` classes is the bulk of the change

This week would be good, as this is somewhat blocking for live-migration

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3524)

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -28,6 +28,7 @@ import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.qos.FakeQosClient;
 import com.palantir.atlasdb.qos.QosClient;
 import com.palantir.atlasdb.util.MetricsManager;
+import com.palantir.timestamp.TimestampManagementService;
 import com.palantir.timestamp.TimestampService;
 import com.palantir.timestamp.TimestampStoreInvalidator;
 
@@ -88,6 +89,8 @@ public interface AtlasDbFactory {
             KeyValueService rawKvs,
             Optional<TableReference> timestampTable,
             boolean initializeAsync);
+
+    TimestampManagementService getTimestampManagementService(TimestampService timestampService);
 
     default TimestampStoreInvalidator createTimestampStoreInvalidator(KeyValueService rawKvs) {
         return () -> {
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -90,7 +90,7 @@ public interface AtlasDbFactory {
             Optional<TableReference> timestampTable,
             boolean initializeAsync);
 
-    TimestampManagementService getTimestampManagementService(TimestampService timestampService);
+    TimestampManagementService createTimestampManagementService(TimestampService timestampService);
 
     default TimestampStoreInvalidator createTimestampStoreInvalidator(KeyValueService rawKvs) {
         return () -> {
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -22,6 +22,7 @@ import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Preconditions;
 import com.palantir.atlasdb.config.LeaderConfig;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
@@ -90,7 +91,15 @@ public interface AtlasDbFactory {
             Optional<TableReference> timestampTable,
             boolean initializeAsync);
 
-    TimestampManagementService createTimestampManagementService(TimestampService timestampService);
+    static TimestampManagementService createTimestampManagementService(TimestampService timestampService) {
+        Preconditions.checkArgument(timestampService instanceof TimestampManagementService,
+                "TimestampManagementService must be created based on result of createTimestampService call."
+                        + "\nExpected a TS implementing TimestampManagementService, got %s",
+                timestampService.getClass());
+
+        return (TimestampManagementService) timestampService;
+
+    }
 
     default TimestampStoreInvalidator createTimestampStoreInvalidator(KeyValueService rawKvs) {
         return () -> {
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
@@ -24,6 +24,7 @@ import com.palantir.lock.LockRequest;
 import com.palantir.lock.LockService;
 import com.palantir.lock.v2.TimelockService;
 import com.palantir.processors.AutoDelegate;
+import com.palantir.timestamp.TimestampManagementService;
 import com.palantir.timestamp.TimestampService;
 
 @AutoDelegate(typeToExtend = TransactionManager.class)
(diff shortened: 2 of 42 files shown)


JacekLach:
@sandorw this is probably good for review; I'm fighting circle but the core changes should be stable

JacekLach:
Hints on how to write tests for this would be welcome; afaik none of the `*AtlasDbFactoryTest`s actually exercise the `createTimestampService` etc codepaths, so I can't just extend those to verify that timestamp management service creation works

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -273,9 +279,30 @@ public static TransactionManager create(MetricsManager metricsManager,
                 : transactionManager;
     }
 
+
+    public static SerializableTransactionManager createForTest(MetricsManager metricsManager,
```
sandorw: Would rather avoid additional constructors if we don't strongly need them. In this case, the caller can just use the other constructor and list the timestamp service twice.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
```diff
@@ -48,10 +50,30 @@
     private final Map<TableReference, ConflictHandler> conflictHandlerOverrides = new HashMap<>();
     private Optional<Long> unreadableTs = Optional.empty();
 
+
+    /**
+     * Utility constructor for when we know we have an InMemoryTimestampService.
+     */
+    public TestTransactionManagerImpl(MetricsManager metricsManager,
```
sandorw: I feel less strongly about this one, since it's test code, but maybe we don't need this constructor either, given that we already have several and this one isn't meaningfully different.

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
```diff
@@ -89,6 +90,8 @@ TimestampService createTimestampService(
             Optional<TableReference> timestampTable,
             boolean initializeAsync);
 
+    TimestampManagementService getTimestampManagementService(TimestampService timestampService);
```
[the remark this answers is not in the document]
JacekLach: So this was actually intentional - since you're not creating a new access point, but instead converting the existing one. I.e. get timestamp management service from this given timestamp service, rather than create a new one.

Happy to change if you think `create` is a better name anyway,

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
JacekLach: That was my initial approach, will revert to it. I introduced this to minimise the diff, but it's not used in many places so happy to make invocations a bit more verbose.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -257,6 +258,14 @@ default boolean isInitialized() {
      */
     TimestampService getTimestampService();
 
+    /**
+     * The timestamp management service is used by libraries providing additional functionality
+     * around AtlasDB. End-user clients probably should not require it.
+     *
+     * @return the timestamp management service for this transaction manager
+     */
```
jeremyk-91: :+1: thanks!

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
```diff
@@ -139,6 +141,15 @@ public TimestampService createTimestampService(
                 initializeAsync);
     }
 
+    @Override
+    public TimestampManagementService createTimestampManagementService(TimestampService timestampService) {
+        Preconditions.checkArgument(timestampService instanceof PersistentTimestampService,
+                "TimestampManagementService must be created based on result of createTimestampService call."
+                        + "\nExpected a PersistentTimestampServiceImpl, got %s", timestampService.getClass());
```
jeremyk-91: nit: the Precondition message should be updated because we're now checking for `PersistentTimestampService`

Also, is there any reason why we check for `PersistentTimestampService` specifically, instead of just the `TimestampManagementService` we want?

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/DbAtlasDbFactory.java
```diff
@@ -96,6 +98,15 @@ public TimestampService createTimestampService(
         return PersistentTimestampServiceImpl.create(createTimestampBoundStore(timestampTable, dbkvs));
     }
 
+    @Override
+    public TimestampManagementService createTimestampManagementService(TimestampService timestampService) {
+        Preconditions.checkArgument(timestampService instanceof PersistentTimestampService,
+                "TimestampManagementService must be created based on result of createTimestampService call."
+                        + "\nExpected a PersistentTimestampServiceImpl, got %s", timestampService.getClass());
```
jeremyk-91: nit: as above, probably should update the preconditions message; also not sure if we should just check if it's an instance of TMS as opposed to PTS specifically

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,11 @@ develop
     *    - Type
          - Change
 
+    *    - |new|
+         - ``TransactionManager``s now expose a ``TimestampManagementService``, allowing clients to fast-forward timestamps when necessary.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3524>`__)
```
jeremyk-91: Probably worth a bit of a warning, maybe along the lines of "this feature is only intended for libraries that directly extend AtlasDB functionality; it is very unlikely that AtlasDB users in general should directly use the timestamp management service"?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
JacekLach: Hm, that's actually fair, and this could become a default method at that point.

I initially wrote these thinking that different (cassandra/db/etc) factories would have different types for the `TimestampService`, but that proved not to be the case.

JacekLach:
@jeremyk-91 Any advice in where in atlasdb-ete-tests such a test would fit? I read through some of the test suites, but some of htem I have trouble following - they seem to make requests to a test service instead of talking to transaction managers?

The closest one seems to be https://github.com/palantir/atlasdb/blob/91fc1ef5f1a9c0b374eca928e61fd78ab4aa4df2/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/QosCassandraEteTestSetup.java#L57-L59 as it actually builds a TM; should I extend that?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
JacekLach: Done, made the method static and just check if it's castable.

JacekLach:
Cool, I think this is all ready.

j-baker:
Hey - I'm fairly un-pumped about this PR. Does it not just assert that every TimestampService must always be an instance of TimestampManagementService too? Does this not just solidify obvious badness?

j-baker:
This looks fine and is exactly how I'd implement this :)

PR #3525, opened 2018-09-26
Upgrade dependency on rescue to 4.x.x

ashrayjain:
CC @tpetracca
In preparation for the 4.x release=

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3525)


sandorw:
Do we want to commit to any code changes before implicitly dropping support for embedded here?

PR #3526, opened 2018-09-27
_transactions2, Part 3: Specific Targets in Timestamp Ranges

jeremyk-91:
**Goals (and why)**:
- Build out a utility for identifying which timestamp in a range is the one we may be interested in as part of `_transactions2`
- Also have something easy to review :)

- Implement a utility class `TimestampRanges` that allows this functionality

- This is new code, and tests have been added that cover a variety of cases.

- There is a bit of math. It's tested, but worth checking that the reasoning makes sense.
- Is there a more performant way to perform this calculation?

**Where should we start reviewing?**: `TimestampRangesTest` probably

**Priority (whenever / two weeks / yesterday)**: this week / early next

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3526)


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertThatThrownBy(() -> TimestampRanges.getTimestampMatchingModulus(SEVENTY_THREE_TO_EIGHTY_TWO, 2, 2))
+                .isInstanceOf(IllegalArgumentException.class);
+        assertThatThrownBy(() -> TimestampRanges.getTimestampMatchingModulus(SEVENTY_THREE_TO_EIGHTY_TWO, 3, 2))
+                .isInstanceOf(IllegalArgumentException.class);
+    }
+
+    @Test
+    public void throwsIfTimestampRangeDoesNotContainAnyValuesMatchingModulus() {
+        TimestampRange oneTimestamp = TimestampRange.createInclusiveRange(77, 77);
+        assertThat(TimestampRanges.getTimestampMatchingModulus(oneTimestamp, 6, 10)).isNotPresent();
```
felixdesouza: this doesn't check that we're throwing?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
```diff
@@ -0,0 +1,86 @@
...
+                .satisfies(optionalLong -> {
+                    long value = optionalLong.getAsLong();
+                    assertThat(value).isIn(74L, 76L, 78L, 80L, 82L);
+                });
+    }
+
+    @Test
+    public void throwsIfResidueExceedsOrEqualsModulus() {
+        assertThatThrownBy(() -> TimestampRanges.getTimestampMatchingModulus(SEVENTY_THREE_TO_EIGHTY_TWO, 2, 2))
+                .isInstanceOf(IllegalArgumentException.class);
```
felixdesouza: assert the message that you expect in the case that this fails with a IAE for other reasons

I typically find:

```java
assertThatExceptionOfType(Illegal...)
   .isThrownBy()
   .withMessage{Containing}(...);
```

reads better, in later versions of assertJ, you can even do 
```java
assertThatIllegalArgumentException()
    .isThrownBy(...)
    ...
```

as a convenience.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRanges.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * If the timestamp range does not contain a timestamp matching the criteria, returns empty.
+     *
+     * @param range timestamp range to extract a single timestamp from
+     * @param residue desired residue class of the timestamp returned
+     * @param modulus modulus used to partition numbers into residue classes
+     * @return a timestamp in the given range in the relevant residue class modulo modulus
+     * @throws IllegalArgumentException if residue >= modulus
+     */
+    public static OptionalLong getTimestampMatchingModulus(TimestampRange range, int residue, int modulus) {
```
felixdesouza: why not just put this on the `TimestampRange`? Similar to how guava's `Range` works.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
jeremyk-91: Right. This is an artifact of when I previously had the API throw if nothing was found - before I realised that there are legitimate cases where `timestampService.getFreshTimestamps(N)` might actually return you nothing that matches a given residue modulo N - will rename the test.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
jeremyk-91: We should discuss this as a team - I don't think we normally do this (because I don't think we consider exception _messages_ to be a part of our public API).

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRanges.java
jeremyk-91: Fair enough, will do.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
jeremyk-91: (offline discussion - yep, they aren't a part of our public API, but is worth testing for the reasons mentioned!)

timestamp-api/src/test/java/com/palantir/timestamp/TimestampRangeTest.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
```
felixdesouza: - [ ] License

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRanges.java
jeremyk-91: I've moved this off the `TimestampRange` and back to the static class. Reason is that `TimestampRange` is in `timestamp-api` which doesn't have Guava and it doesn't feel right to force users of the timestamp API to have a Guava dependency - and I'd prefer not to copy related `LongMath` and `Preconditions` stuff into our project.

timestamp-api/src/main/java/com/palantir/timestamp/TimestampRange.java
```diff
@@ -77,4 +79,35 @@ public long size() {
...
+     *
+     * If the timestamp range does not contain a timestamp matching the criteria, returns empty.
+     *
+     * @param residue desired residue class of the timestamp returned
+     * @param modulus modulus used to partition numbers into residue classes
+     * @return a timestamp in the given range in the relevant residue class modulo modulus
+     * @throws IllegalArgumentException if residue >= modulus
+     */
+    public OptionalLong getTimestampMatchingModulus(int residue, int modulus) {
+        Preconditions.checkArgument(residue < modulus,
```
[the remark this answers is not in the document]
jeremyk-91: Yep, I've added support for negatives. Guava has some decent utilities.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRanges.java
felixdesouza: I'm not sure I buy this, there aren't non-atlas consumers of `timestamp-api` and at the very least we're not encouraging this. So if we assume everyone who has this has atlas-client, then they're already pulling in Guava (albeit transitively). So not really sure why it matters here. e.g. `qos-service-api` pulls in an entire remoting3 which I would argue is a lot more heavy handed than guava (and *does* pull in guava anyway). 

It does seem like the best place to put it and isn't just a mere utility, but I won't push too much.

jeremyk-91: There _are_ non-Atlas consumers of `timestamp-api` though (check large internal project). The actual server in said non-Atlas consumer does pull in Guava on its own, but IMO that's a separate concern and I'm not sure we should force that.

Regarding QoS, I agree that Remoting3 is a lot more heavy handed than Guava, but I'd be inclined to agree with the "everyone who has this has atlas-client" statement a lot more strongly if you're a QoS user. (I agree with the logic, just not so much with the premise in this case)

PR #3529, opened 2018-09-27
_transactions2, Part 5: Better Balanced Partitioning

jeremyk-91:
**Goals (and why)**:
- Partition things somewhat sensibly; try as far as possible to avoid nodes actually acquiring the same moduli in the presence of restarts.

- Implement `DistributingModulusGenerator`, which takes a range of residues (e.g. 0-15 inclusive), and supports two operations: getting a residue (which in practice is the residue with the fewest users), and releasing a residue.
- Implement a loading cache around the modulus generator as a `CachingPartitionAllocator`. This is important because the DMG maintains no knowledge of what IDs it was called with.

New classes are tested.

- Is it worth doing some internal repartitioning (e.g. internally, by periodically flushing the cache - or doing something clever in unmarkResidue also possibly invalidating certain keys in the cache)? One issue with the current algorithm is that you may get into an imbalanced state if you have an influx of clients and then some leave.

**Where should we start reviewing?**: `DistributingModulusGeneratorTest`

**Priority (whenever / two weeks / yesterday)**: one week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3529)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+public class DistributingModulusGenerator {
+    private final SortedSet<ReferenceCountedResidue> referenceCounts;
+
+    @VisibleForTesting
+    DistributingModulusGenerator(SortedSet<ReferenceCountedResidue> referenceCounts) {
+        this.referenceCounts = referenceCounts;
+    }
+
+    public static DistributingModulusGenerator create(int modulus) {
+        SortedSet<ReferenceCountedResidue> referenceCounts
```
felixdesouza: Adding the map should be trivial no? Might as well include it imo 🤷‍♂️ 

```java
referenceCounts.stream().collect(toMap(ReferenceCountedResidue::residue, Function.identity());
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+ * residues).
+ *
+ * Unmarking residues may not be performant if the number of residues used is large.
+ */
+public class DistributingModulusGenerator {
+    private final SortedSet<ReferenceCountedResidue> referenceCounts;
+
+    @VisibleForTesting
+    DistributingModulusGenerator(SortedSet<ReferenceCountedResidue> referenceCounts) {
+        this.referenceCounts = referenceCounts;
```
felixdesouza: whilst this is `@VisibleForTesting` you're not actually using it, you're using `create` below. I think the constructor should take the `modulus` as opposed to an arbitrary `SortedSet` i.e. you could write a test with a sorted set sorted arbitrarily as opposed to how you actually want it to be sorted => impl is in the test.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.Test;
+
+import com.google.common.collect.Iterables;
+
+public class DistributingModulusGeneratorTest {
+    private DistributingModulusGenerator generator;
+
+    @Test
+    public void doesNotReuseResiduesIfNotNeeded() {
+        givenGeneratorHasModulus(3);
```
gmaretic: This sounds like an assertion. Maybe rename to `setupGeneratorWithModulus`?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+        requestResiduesAndAssertBalanced(3);
+        requestResiduesAndAssertBalanced(3);
+        requestResiduesAndAssertBalanced(3);
+
+        unmarkResidue(2);
+        unmarkResidue(2);
+        unmarkResidue(2);
+        List<Integer> additionalResponses = requestResidues(6);
```
gmaretic: do `requestResidues(3)` and then you can do the following check and `requestResiduesAndAssertBalanced(3);` without having to sublist

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+        assertThat(additionalResponses.subList(0, 3)).containsExactly(2, 2, 2);
+        assertListBalanced(additionalResponses.subList(3, 6), 3);
+    }
+
+    private void givenGeneratorHasModulus(int modulus) {
```
gmaretic: Only ever calling this with modulus 3. Either add tests with a different number or just hardcode the generator

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private List<Integer> requestResidues(int times) {
+        return IntStream.range(0, times)
+                .map(unused -> generator.getAndMarkResidue())
+                .boxed()
+                .collect(Collectors.toList());
+    }
+
+    private void requestResiduesAndAssertBalanced(int times) {
```
gmaretic: If you hardcode generator, then this should also be hardcoded for 3

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampServiceTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.After;
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.atlasdb.timelock.paxos.ManagedTimestampService;
+import com.palantir.atlasdb.timelock.transaction.client.NumericPartitionAllocator;
+import com.palantir.timestamp.TimestampRange;
+
+@SuppressWarnings("unchecked") // Mocks of parameterised types
+public class DelegatingClientAwareManagedTimestampServiceTest {
```
gmaretic: Skipping as the other PR has more up to date version

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/TimestampRangesTest.java
```diff
@@ -0,0 +1,86 @@
...
+package com.palantir.atlasdb.timelock.transaction.timestamp;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.Test;
+
+import com.palantir.timestamp.TimestampRange;
+
+public class TimestampRangesTest {
```
gmaretic: Skipping as the other PR has more up to date version

jeremyk-91:
NOTE: I've done the rebase, but will go through the comments tomorrow. So at this point there is nothing new to review yet

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+                        referenceCountedResidue.residue());
+                referenceCounts.remove(referenceCountedResidue);
+                referenceCounts.add(referenceCountedResidue.unmark());
+                return;
+            }
+        }
+        throw new IllegalStateException("Residue " + residue + " was not found when unmarking!");
+    }
+
+    @Value.Immutable
```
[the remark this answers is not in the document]
jeremyk-91: Yep, agree that the current implementation isn't the most performant, though currently given small number of objects I think this should be fine.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+
+        @Value.Parameter
+        int residue();
+
+        @Value.Check
+        default void check() {
+            Preconditions.checkState(references() >= 0, "Reference counts must not be negative");
+        }
+
+        default ReferenceCountedResidue mark() {
```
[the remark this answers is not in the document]
jeremyk-91: thank @felixdesouza for this one :)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
jeremyk-91: Added tests with a different number (along with a check that the modulus parameter is positive)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
jeremyk-91: (Didn't hardcode the generator.)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,164 @@
...
+import com.google.common.collect.Iterables;
+
+public class DistributingModulusGeneratorTest {
+    private DistributingModulusGenerator generator;
+
+    @Test
+    public void doesNotReuseResiduesIfNotNeeded() {
+        setupGeneratorWithModulus(3);
+
+        requestResiduesAndAssertBalanced(3);
```
felixdesouza: I would argue being slightly more verbose in your tests. Reading this, and looking at the test name, I don't really have an idea what it is doing.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private void requestResiduesAndAssertBalanced(int times) {
+        assertListBalanced(requestResidues(times), times);
+    }
+
+    /**
+     * Asserts that, starting from the beginning of the list, each full modulus number of elements has the integers
+     * from 0 to (modulus - 1) once each. No guarantees are made on the remainder of the list.
+     */
+    private void assertListBalanced(List<Integer> responses, int modulus) {
```
felixdesouza: See above comment about being verbose, i'm not a huge fan of "computations" in tests, as a reviewer I have to effectively run through in my head what potential outputs this should have. This is compounded by javadoc explaining the method.

Also, `static`.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
felixdesouza: is it merely coincidence that `times` and `modulus` are equal?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/CachingPartitionAllocatorTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void unmarksResidueIfClientGoesAway() {
+        List<Integer> firstResponse = allocator.getRelevantModuli(KEY);
+
+        time.addAndGet(5_000_000L);
+        allocator.loadingCache.cleanUp();
+        scheduler.runUntilIdle();
+
+        allocator.getRelevantModuli(KEY);
```
gmaretic: This should force eviction from the cache, so if you call `scheduler.runUntilIdle()` after this line, the test shouldn't flake and you can make @felixdesouza happy by hiding the loadingCache.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @Test
+    public void canUseResidueOne() {
+        setupGeneratorWithModulus(1);
+
+        IntStream.range(0, 50).forEach(unused -> requestResiduesAndAssertBalanced(1));
+    }
+
+    @Test
+    public void throwsIfCreatedWithNonPositiveModuli() {
+        assertThatThrownBy(() -> setupGeneratorWithModulus(0))
```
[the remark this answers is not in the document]
gmaretic: This is open for discussion. I think that X results in Y exception is more consistent with the usual format of assertThat(X) satisfies Y.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
gmaretic: The verifying method only works if times is a multiple of modulus

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/CachingPartitionAllocatorTest.java
jeremyk-91: Yep, this does force eviction (though it relies on an implementation detail of the cache!) - I will switch to this.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Test
+    public void canUnmarkMarkedResidues() {
+        setupGeneratorWithModulus(3);
+
+        requestResidues(3);
+        unmarkResidue(0);
+        unmarkResidue(1);
+        unmarkResidue(2);
+        // Not throwing is a success!
```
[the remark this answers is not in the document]
jeremyk-91: Yep, this is good! The method name makes my skin crawl (`exception` is countable, so this reads as wrong to me - it feels like it should be `doesNotThrowAnyExceptions`) but agree this is clearer

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/client/DistributingModulusGeneratorTest.java
jeremyk-91: I've removed this code since the imbalanced case wasn't actually ever asserted (i.e. we didn't need to say that [0, 1, 2, {one of them}] needs to pass anywhere)

gmaretic:
LGTM, @felixdesouza ?

Issue #3519, opened 2018-09-21
StartupIndependenceMultipleCassandraEteTest flakes/takes long time to execute


gokcanozakdag:
`StartupIndependenceMultipleCassandraEteTest` is generally the longest running test (takes around 4 minutes); and even worse, flakes frequently and requiring rebuilding all the project. Not sure what would be the best solution here, wanted to create an issue to track it.

PR #3530, opened 2018-09-28
Fix startup independence tests

gmaretic:
**Goals (and why)**:
Address https://github.com/palantir/atlasdb/issues/3519

Replaced existing tests with tests that TM gets eventually initialized when KVS gets initialized async, and that it gets synchronously initialized when KVS is ready using an "asynchonously initializable" implementation of in memory KVS. Then, we also test that Cassandra KVS gets asynchronously initialized when cassandra is started after creation.
Also, found and fixed a bug in the initialization of TM where the TM was not actually ready as soon as it was returned.

New tests

ASAP, as it makes builds faster and less flaky

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3530)


atlasdb-commons/src/main/java/com/palantir/async/initializer/AsyncInitializer.java
```diff
@@ -115,8 +115,7 @@ void assertNeverCalledInitialize() {
         }
     }
 
-    // Not final for tests.
```
felixdesouza: :heart:

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/memory/InMemoryAsyncAtlasDbConfig.java
```diff
@@ -0,0 +1,59 @@
...
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.google.auto.service.AutoService;
+import com.palantir.atlasdb.spi.KeyValueServiceConfig;
+
+@JsonTypeName(InMemoryAsyncAtlasDbConfig.TYPE)
+@AutoService(KeyValueServiceConfig.class)
+public final class InMemoryAsyncAtlasDbConfig implements KeyValueServiceConfig {
```
felixdesouza: immutables?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -67,7 +67,11 @@
             this.initializationPrerequisite = initializationPrerequisite;
             this.callback = callBack;
             this.executorService = initializer;
-            scheduleInitializationCheckAndCallback();
+            if (isInitializedInternal()) {
```
felixdesouza: what is this doing? extract to a method?

as a side note: `scheduleInitializationCheckAndCallback` keeps scheduling itself, which is unnecessary since the executor service is a `ScheduledExecutorService`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/memory/InMemoryAsyncAtlasDbFactoryTest.java
```diff
@@ -0,0 +1,92 @@
...
+        KeyValueService kvs = createRawKeyValueService(false);
+        assertThat(kvs.isInitialized()).isTrue();
+        assertThat(kvs.getAllTableNames()).isEmpty();
+    }
+
+    @Test
+    public void asyncInitKvsAsynchronous() {
+        KeyValueService kvs = createRawKeyValueService(true);
+        assertThat(kvs.isInitialized()).isFalse();
+        assertThatThrownBy(kvs::getAllTableNames).isInstanceOf(NotInitializedException.class);
```
felixdesouza: isn't this flaky?

felixdesouza: is there a way to insert some latch or some barrier? although goes back to my first point, not immediately useful 🤷‍♂️  should be inside the `AsyncInitializer` test

timestamp-api/src/main/java/com/palantir/timestamp/InMemoryTimestampService.java
```diff
@@ -54,4 +54,9 @@ public void fastForwardTimestamp(long currentTimestamp) {
     public String ping() {
         return PING_RESPONSE;
     }
+
+    @Override
+    public long getUpperLimitTimestampToHandOutInclusive() {
```
felixdesouza: if going to be unimplemented, why change superinterface out of curiosity?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
felixdesouza: also this block exists elsewhere

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceAsyncInitializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+                CassandraKeyValueServiceRuntimeConfig::getDefault,
+                CassandraContainer.LEADER_CONFIG,
+                CassandraMutationTimestampProviders.legacyModeForTestsOnly(),
+                true,
+                FakeQosClient.INSTANCE);
+
+        assertThat(asyncInitializedKvs.isInitialized()).isFalse();
+
+        startCassandra();
+
```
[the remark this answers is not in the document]
gmaretic: The KVS is created before the container is even started, we test that the KVS eventually gets initialized after we start the container.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/memory/InMemoryAsyncAtlasDbConfig.java
gmaretic: Just used InMemoryAtlasDbConfig as template. If anything, this should really be a singleton enum?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
gmaretic: You need to give it a period if you want it repeated, which we could do but then we'd need to cancel it when it's successful so this looks nicer. Extracted duplicated code.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/memory/InMemoryAsyncAtlasDbFactoryTest.java
gmaretic: It's not really that flaky as the KVS is going to take at least one second to initialize.

timestamp-api/src/main/java/com/palantir/timestamp/InMemoryTimestampService.java
gmaretic: Just because I needed a single interface to use with @AutoDelegate for AsyncInitializeableInMemoryTimestampService

gmaretic:
Also, decoupled existing tests by having them use separate keyspaces

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraSweepProgressStoreIntegrationTest.java
```diff
@@ -24,14 +24,16 @@
...
     @ClassRule
     public static final Containers CONTAINERS = new Containers(CassandraBackgroundSweeperIntegrationTest.class)
-            .with(new CassandraContainer());
+            .with(container);
 
     @Override
     protected KeyValueService getKeyValueService() {
         return CassandraKeyValueServiceImpl.createForTesting(
-                CassandraContainer.KVS_CONFIG,
+                container.getConfig(),
```
j-baker: this code snippet is the same in all of these tests. Have you thought about just making a new CassandraContainers rule which wraps the Containers rule and just contains a CassandraKeyValueServiceImpl?

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTargetedSweepIntegrationTest.java
```diff
@@ -40,23 +40,21 @@
...
     public static final Containers CONTAINERS =
-            new Containers(CassandraTargetedSweepIntegrationTest.class).with(new CassandraContainer());
-
-    private final MetricsManager metricsManager = MetricsManagers.createForTests();
+            new Containers(CassandraTargetedSweepIntegrationTest.class).with(container);
 
     @Rule
     public final RuleChain ruleChain = SchemaMutationLockReleasingRule.createChainedReleaseAndRetry(
-            getKeyValueService(), CassandraContainer.KVS_CONFIG);
+            getKeyValueService(), container.getConfig());
```
j-baker: what's the SchemaMutationLockReleasingRule?

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/InMemoryKeyValueService.java
```diff
@@ -86,12 +142,20 @@ public InMemoryKeyValueService(boolean createTablesAutomatically) {
...
+    public InMemoryKeyValueService(boolean createTablesAutomatically, ExecutorService executor) {
         super(executor);
         this.createTablesAutomatically = createTablesAutomatically;
     }
 
+    /**
+     * Creates an InMemoryKvs that can be asynchronously initialized. If initializeAsync is true, the returned Kvs will
+     * initially be uninitialized, and will only succeed at the second attempt at initialization.
+     */
+    public static KeyValueService create(boolean createTablesAutomatically, boolean initializeAsync) {
```
j-baker: we shouldn't need this... The InMemoryKvs should never need initializing

j-baker: it's a bunch of hashmaps

j-baker: if we want to have behaviour like this in our tests, we should just have a test kvs which does this; the InMemoryKvs is closer to prod code and so shouldn't really contain extraneous stuff

atlasdb-commons/src/main/java/com/palantir/async/initializer/Callback.java
```diff
@@ -75,6 +75,24 @@ public void runWithRetry(R resource) {
         }
     }
 
+    /**
+     * Try init() once only, performing any necessary cleanup
+     */
+    public boolean runOnceOnly(R resource) {
+        try {
+            lock.lock();
```
j-baker: any chance we could get rid of the 'lock' and instead use synchronized here? In particular Java locks need to be used in the pattern:

```
lock.lock();
try {
    // do stuff
} finally {
    lock.unlock();
}
```

rather than the current one; this one is broken because the lock.lock() call may fail, and then you're unlocking a lock you never had :(

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/memory/InMemoryAsyncAtlasDbFactory.java
```diff
@@ -0,0 +1,72 @@
...
+import com.palantir.atlasdb.spi.AtlasDbFactory;
+import com.palantir.atlasdb.spi.KeyValueServiceConfig;
+import com.palantir.atlasdb.spi.KeyValueServiceRuntimeConfig;
+import com.palantir.atlasdb.util.MetricsManager;
+import com.palantir.atlasdb.versions.AtlasDbVersion;
+import com.palantir.timestamp.InMemoryTimestampService;
+import com.palantir.timestamp.TimestampService;
+
+@AutoService(AtlasDbFactory.class)
+public class InMemoryAsyncAtlasDbFactory implements AtlasDbFactory {
```
j-baker: this should also exist only in test code! It's not valuable for a consumer of Atlas to have this code around

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AsyncInitializeableInMemoryTimestampService.java
```diff
@@ -0,0 +1,68 @@
...
+package com.palantir.atlasdb.keyvalue.impl;
+
+import com.google.common.base.Preconditions;
+import com.palantir.async.initializer.AsyncInitializer;
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.timestamp.AutoDelegate_PersistentTimestampService;
+import com.palantir.timestamp.InMemoryTimestampService;
+import com.palantir.timestamp.PersistentTimestampService;
+
+public final class AsyncInitializeableInMemoryTimestampService extends AsyncInitializer
```
j-baker: likewise - this should be in a test package

timestamp-api/src/main/java/com/palantir/timestamp/InMemoryTimestampService.java
j-baker: I agree with felix - this interface just isn't implemented!

j-baker: like, if you need something for use in testing, then by all means add some small piece of test-only code which encapsulate the hacks you need to write your test - but we shouldn't compromise our exported code (by e.g. incorrectly implementing interfaces) in order to make testing easier - it implies that the decomp is incorrect.

The Atlas code should make sense as something someone can read, and we shouldn't have to say 'oh this class doesn't actually make sense as anything you'd want to run, but it's easier to write this test (that no longer exists because someone refactored it away) if it works this way'.

gmaretic:
Since this PR has exploded in size, I'll leave further fixes for a follow-up PR: in many places we unnecessarily start new CKVSs and close them for each test, which ends up costing us a few minutes in test runtime. Also, we need to rebalance the containers, since this PR effectively frees the slowest container up.

What this PR introduces:
- much faster and less flaky Async initialization tests
- less sketchy synchronization for AsyncInitilializer
- new Cassandra ClassRules that, among other things, decouple test classes by running each in a separate keyspace

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/InMemoryKeyValueService.java
```diff
@@ -78,6 +78,8 @@
  */
 @ThreadSafe
 public class InMemoryKeyValueService extends AbstractKeyValueService {
+
```
gokcanozakdag: nit

PR #3532, opened 2018-09-29
Fix issue where TargetedSweep couldn't sweep tables with no namespace

dxiao:
Like tables from a certain large project. :P

**Goals (and why)**:

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3532)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
@@ -56,13 +56,14 @@ public abstract class WriteReference implements Persistable {
             .registerModule(new Jdk8Module())
             .registerModule(new AfterburnerModule());
 
+    @SuppressWarnings("deprecated")
     public static final Hydrator<WriteReference> BYTES_HYDRATOR = input -> {
         if (input[0] != prefix[0]) {
             return decodeLegacy(input);
         }
         int offset = 1;
         String tableReferenceString = EncodingUtils.decodeVarString(input, offset);
-        TableReference tableReference = TableReference.createFromFullyQualifiedName(tableReferenceString);
+        TableReference tableReference = TableReference.createUnsafe(tableReferenceString);
         offset += EncodingUtils.sizeOfVarString(tableReferenceString);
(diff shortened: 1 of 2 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReference.java
```diff
@@ -56,13 +56,14 @@ public CellReference cellReference() {
...
 
+    @SuppressWarnings("deprecated")
     public static final Hydrator<WriteReference> BYTES_HYDRATOR = input -> {
         if (input[0] != prefix[0]) {
             return decodeLegacy(input);
         }
         int offset = 1;
         String tableReferenceString = EncodingUtils.decodeVarString(input, offset);
-        TableReference tableReference = TableReference.createFromFullyQualifiedName(tableReferenceString);
+        TableReference tableReference = TableReference.createUnsafe(tableReferenceString);
```
j-baker: I'm not even sure this is right. How about 'fromString'? This class is full of juicy code, e.g.

createUnsafe vs fromString vs fromInternalTableName all having slightly different implementations

j-baker: Do we need to worry about fromInternalTableName and the curse of __ -> ., basically?

j-baker: turns out no, that's a Cassandra thing

j-baker: but still - maybe fromString, since it's not deprecated?

dxiao:
Also, can somebody explain to me what is going on with the test failure? 
```
[00:05:15] Warning, treated as error:
[00:05:15] /home/circleci/project/docs/source/release_notes/release-notes.rst::Anonymous hyperlink mismatch: 1 references but 0 targets.
[00:05:15] See "backrefs" attribute for IDs.
[00:05:15] Makefile:56: recipe for target 'html' failed
```

j-baker:
Fixed in #3533

PR #3533, opened 2018-09-30
A dictionary for table names for use in targeted sweep

j-baker:
Even now, the overheads of targeted sweep can be relatively high on
systems with many indices and high ingestion rate (notably the lack of
block compression and the comparatively small disk sizes bite us).

This will require care wrt our internal backup infrastructure. I'm going
to coordinate this, but it adds an extra 'must repair these tables in C*
post restore' because this is one of those tables where even a small
inconsistency can become catastrophic.

Benchmarking the storage from this we can see that the improvement is
significant. At the moment in our benchmarking, the average
row-column-value total overhead is 80 bytes over the internal workload.

However, we can account for at most 19 bytes in the row and column, so
there must be on average 61 bytes in the value. Moreover, some samples
imply that a good chunk of the more written-to-tables have #bytes used
values >= 30 characters, which would imply that we could plausibly
see another step change improvement here.

Note: PR looks big. A lot of it is just changes to codegen; total diff is a _reasonable_ 800-lines-ish.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3533)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/id/SweepTableIndices.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private int loadUncached(TableReference table) {
+        while (true) {
+            Optional<SweepTableIdentifier> identifier = namesToIds.currentMapping(table);
+            if (identifier.isPresent() && !identifier.get().isPending()) {
+                return identifier.get().identifier();
+            }
+            log.info("Assigning table {} an identifier", LoggingArgs.tableRef(table));
+            SweepTableIdentifier afterPendingPut = namesToIds.storeAsPending(table, idToNames.getNextId());
```
sandorw: This reads a bit gross if we have to go through the loop more than once. This cas will fail and return the most recently written ID, which was inserted by the other `storeAsPending` call. My first thought was that the logic was incorrect, but since it doubles as a read, it works.

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
```diff
@@ -33,27 +35,48 @@
...
+        Schema schema = newSchemaObject();
+        addSweepStateTables(schema);
+        addTableIdentifierTables(schema);
+
+        schema.validate();
+        return schema;
+    }
+
+    /**
+     * This is visible for the ETE tests, which must not truncate the table identifier tables or risk cache
```
sandorw: Maybe worth clarifying that the table mappings can be persistent without any issues?

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
```diff
@@ -81,19 +104,45 @@ private static Schema generateSchema() {
...
+            conflictHandler(ConflictHandler.IGNORE_ALL);
+        }});
+
+        schema.addTableDefinition("sweepIdToName", new TableDefinition() {{
+            allSafeForLoggingByDefault();
+            rowName();
+            hashFirstRowComponent();
+            rowComponent("singleton", ValueType.STRING);
+            dynamicColumns();
+            columnComponent("tableId", ValueType.VAR_LONG, ValueByteOrder.DESCENDING);
```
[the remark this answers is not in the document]
j-baker: I want it to be dynamic columns because the alternative is to do a range scan, and I'd like Atlas to never rely on those for new features.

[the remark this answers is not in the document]
j-baker: and yeah, will add some comments.

j-baker:
note: do not merge this; there's a rescue PR internally that I'd like merged first.

atlasdb-client/src/main/java/com/palantir/atlasdb/schema/TargetedSweepSchema.java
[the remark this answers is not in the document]
sandorw: Makes sense - asked this before I got to the part where we scan for latest ID.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/id/SweepTableIndices.java
j-baker: I made it more clear. We certainly have to pick between a PUE or a CAS here

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/WriteReferencePersister.java
```diff
@@ -0,0 +1,96 @@
...
+                } catch (IOException e) {
+                    throw new RuntimeException("Exception hydrating object.");
+                }
+            }
+
+            @Override
+            public WriteReference visitTableNameAsStringBinary(byte[] ref) {
+                int offset = 1;
+                String tableReferenceString = EncodingUtils.decodeVarString(ref, offset);
+                TableReference tableReference = TableReference.createFromFullyQualifiedName(tableReferenceString);
```
dxiao: oh c'mon, I thought we fixed that...

PR #3535, opened 2018-10-02
Fds/test coverage 3148

### LINKED ISSUE #3148: Add test coverage for Serializable range request conflict detection
See https://github.com/palantir/atlasdb/pull/3147 - we obviously have no test coverage for this and no one has hit this use case, since the code has been broken for ~2 years.

felixdesouza:
**Goals (and why)**:
Fixes #3148. If the fix to #3148 is reverted, the added test will fail. Also cleaned up some things in this class and neighbouring classes.

Opted to use `Multimap` as opposed to a map with a collection value. Cleaner semantics and less null checks, happy to revert if necessary.

Added a test that reads from a second row. Should we amend all our tests to read from at least 2 rows/cols?

Conversions to `computeIfAbsent`, `Stream`-ifying, etc.

`AbstractSerializationTransactionTest`

Whenever.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3535)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -30,6 +30,8 @@ import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
 
+import javax.annotation.Nullable;
+
 import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -43,6 +45,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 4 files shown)


felixdesouza:
[no release notes]

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -253,23 +256,13 @@ public SerializableTransaction(MetricsManager metricsManager,
...
 
     private ConcurrentNavigableMap<Cell, byte[]> getReadsForTable(TableReference table) {
-        ConcurrentNavigableMap<Cell, byte[]> reads = readsByTable.get(table);
-        if (reads == null) {
-            ConcurrentNavigableMap<Cell, byte[]> newMap = new ConcurrentSkipListMap<>();
-            readsByTable.putIfAbsent(table, newMap);
-            reads = readsByTable.get(table);
-        }
-        return reads;
+        return readsByTable.computeIfAbsent(table, $ -> new ConcurrentSkipListMap<>());
```
j-baker: probably best to not use '$' as a variable name; it's disallowed by our checkstyle but unenforced due to a checkstyle bug. Rather, prefer something like 'unused'.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1460,14 +1446,7 @@ protected void throwIfReadWriteConflictForSerializable(long commitTimestamp) {
...
     private boolean hasWrites() {
-        boolean hasWrites = false;
-        for (SortedMap<?, ?> map : writesByTable.values()) {
-            if (!map.isEmpty()) {
-                hasWrites = true;
-                break;
-            }
-        }
-        return hasWrites;
+        return writesByTable.values().stream().anyMatch(writesForTable -> !writesForTable.isEmpty());
```
felixdesouza: are we opposed to doing `hasWrites` as a `volatile boolean` similar to `hasReads`?

sandorw: Not opposed if you think that's clearer, although this is also sufficient. There's already a lot of state being tracked in this class.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -702,6 +702,25 @@ public void testMultipleReadsToSameColumnRange() {
         t1.commit();
     }
 
+    @Test
+    public void testMultipleReadsToSameColumnRangeAcrossRows() {
+        byte[] row1 = PtBytes.toBytes("row1");
+        byte[] row2 = PtBytes.toBytes("row2");
+
+        Transaction transaction = startTransaction();
+        Map<byte[], BatchingVisitable<Map.Entry<Cell, byte[]>>> columnRange =
```
sandorw: Share the `BatchColumnRangeSelection` to make it clearer it's the same and prevent someone from breaking the test by updating one and not the other.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
felixdesouza: I shall leave it, agree with the state tracking bit.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
felixdesouza: good catch

PR #3542, opened 2018-10-02
Excavator/3528

felixdesouza:
Ported from #3528

Review, commit by commit.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3542)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CheckAndSetRequest.java
```diff
@@ -35,20 +40,33 @@
 
     public abstract Cell cell();
 
-    public abstract Optional<byte[]> oldValue();
+    @Nullable
+    @JsonProperty("oldValue")
+    protected abstract byte[] oldValueNullable();
```
jeremyk-91: It looks like this is safe as far as serialization is concerned (https://github.com/FasterXML/jackson-modules-java8/tree/master/datatypes), so this should be fine.

qos-service-impl/src/test/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricMeasurementsLoaderTest.java
```diff
@@ -69,21 +67,6 @@ public void setup() {
...
-        scheduledExecutorService.tick(4, TimeUnit.SECONDS);
-        verify(HEALTH_METRIC_SUPPLIER, times(2)).get();
-
-        scheduledExecutorService.tick(5, TimeUnit.SECONDS);
-        verify(HEALTH_METRIC_SUPPLIER, times(3)).get();
-
-        scheduledExecutorService.tick(5, TimeUnit.SECONDS);
-        verify(HEALTH_METRIC_SUPPLIER, times(4)).get();
-    }
-
```
jeremyk-91: It seems reasonable to have _a test_ that checks that with the passage of time we hit the supplier again (this specific one might be way too strong) though - seems like current set of tests would admit a solution that caches a loaded set of values and never hits the supplier again, unless I'm missing something

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -111,7 +111,7 @@ public void initialize(TransactionManager txManager) {
         runInBackground();
     }
 
-    public void initializeWithoutRunning(TransactionManager txManager) {
+    private void initializeWithoutRunning(TransactionManager txManager) {
```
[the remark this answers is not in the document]
felixdesouza: I don't even recall changing this, good catch.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CellsSweeperShould.java
```diff
@@ -125,14 +126,13 @@ public void releaseTheBackupLockAfterDeleteAndAddingSentinels() {
...
-        } catch (Exception e) {
-            // expected
-        }
+    public void releaseTheBackupLockIfDeleteFails() {
+        doThrow(new RuntimeException("something bad happened"))
+                .when(mockKvs).delete(TABLE_REFERENCE, SINGLE_CELL_TS_PAIR);
+
+        assertThatExceptionOfType(RuntimeException.class)
+                .isThrownBy(() -> sweeper.sweepCells(TABLE_REFERENCE, SINGLE_CELL_TS_PAIR, ImmutableSet.of()))
+                .withMessage("something bad happened");
```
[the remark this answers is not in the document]
felixdesouza: This is a direct translation of the previous test. Lines 129-130 aren't valid in Mockito 2 since `KeyValueService.delete` does *not* throw any checked exceptions. That said, I have to change the assertion, and I'm switching to using the assertJ style.

You're right that the exception doesn't matter, which is why I've picked a `RuntimeException`. But I want to know that I'm not throwing elsewhere from where I'm doing the `doThrow` hence why I'm asserting on the message => I know that a bad thing happening in delete still doesn't matter.

qos-service-impl/src/test/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricMeasurementsLoaderTest.java
felixdesouza: I deleted it because the test seemed pretty wild, how does it know that it has to be 5 seconds that passes? It doesn't even check that if you call it smaller than 5 seconds you'd get the same thing/or the count wouldn't go up. 

Ideally the impl would use something like Guava `AbstractScheduledService` and we just test the `runOneIteration` method.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
```diff
@@ -29,9 +32,23 @@
...
-    public abstract Optional<byte[]> getPreviousStartRow();
+    @Nullable
+    @JsonProperty("previousStartRow")
+    protected abstract byte[] getPreviousStartRowNullable();
+
+    @Nullable
+    @JsonProperty("nextStartRow")
+    protected abstract byte[] getNextStartRowNullable();
+
+    @JsonIgnore
```
sandorw: I don't think this would work if we persisted these anywhere and needed to deserialize them, but it looks like we don't.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueService.java
```diff
@@ -382,7 +381,7 @@ public boolean supportsCheckAndSet() {
     @Override
     public void checkAndSet(CheckAndSetRequest checkAndSetRequest) {
         try {
-            CheckAndSetRequest request = ImmutableCheckAndSetRequest.builder()
```
sandorw: Boo

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CellsSweeperShould.java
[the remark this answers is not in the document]
sandorw: Jeremy's concern might have been that the exception gets wrapped at some point, although it looks like that's not the case here.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/SweepResults.java
felixdesouza: why not?

sandorw: Nevermind, missed the new JsonProperty keys

qos-service-impl/src/test/java/com/palantir/atlasdb/qos/ratelimit/CassandraMetricMeasurementsLoaderTest.java
jeremyk-91: The test as written is not good (for the reasons you mention), but I think the concern that the current set of tests admits a clearly deficient solution still holds (hence 'a test' as opposed to 'this test').

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/CellsSweeperShould.java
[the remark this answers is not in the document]
jeremyk-91: Exactly - my concern was if the exception gets wrapped failing seems harsh.

PR #3549, opened 2018-10-03
kvs-slow-log -> kvs-slow-log-2

fsamuel-bs:
**Goals (and why)**: Work around https://github.com/palantir/atlasdb/pull/3469.

**Implementation Description (bullets)**: Modifying `kvs-slow-log` to `kvs-slow-log-2`.

**Testing (What was existing testing like?  What have you done to improve it?)**: -

**Concerns (what feedback would you like?)**: Should I name it something else?

**Where should we start reviewing?**: +1 -1

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3549)


tboam:
Seems sensible if we can't unblacklist this yet - I'll check on this.  Note that we might need to add this new origin to log whitelists in certain places.

PR #3557, opened 2018-10-05
LockServiceStateLogger: Finding the next lock

jeremyk-91:
**Goals (and why)**:
- The lock service state logger should be able to tell, for each outstanding request, what the next lock it needs to acquire is.
- Runtime performance of the lock service should not be affected if no dumps are taken.

- Expose an ability on a `KnownClientLock` to check if a lock is being held by an individual client in the read or write mode.
- The algorithm iterates through the locks in order, stopping at the first lock that we do not hold. That is reported as the next lock.
- We also report progress information (e.g. this is lock number 55 of 77).

Sample output could look like this:

```
# WARNING: For LockClient.ANONYMOUS this method may not return accurate results.
---
SynthesizedLockRequestState:
    Client B:
    -   nextLock: Lock-1
        numLocksAcquired: 0
        requests:
        -   blockingDuration: 1000
            blockingMode: BLOCK_UNTIL_TIMEOUT
            clientId: Client B
            creatingThread: main
            lockCount: 1
            lockDescriptor: Lock-1
            lockGroupBehavior: LOCK_ALL_OR_NONE
            lockMode: WRITE
            lockTimeout: 120000
            versionId: null
        totalNumLocks: 1
```

Lock service state logger is tested, I added a test for the new file.

- This approach seems pretty complicated. An easier approach is to maintain state for each request in a map and update it as we go, but that violates my second goal of not normally adding overhead. Is there a way to manage the complexity here, or is my second goal misguided?
- This method does not work for ANONYMOUS lock clients. Internal use cases we were debugging use named lock clients, so this should be fine, but is this a broader problem?
- Are there any threading / synchronization landmines I'm stepping on? There are race conditions everywhere, but I think this is still useful - if there's a legit deadlock things should settle.
- Do we leak lock descriptors anywhere in the synthesized file? I don't think so, but should be worth a check.

**Where should we start reviewing?**: LockServiceStateDebugger

**Priority (whenever / two weeks / yesterday)**: early next week. I'm not sure it's as critical for the PDS ticket any more (as you can reconstruct this information from what's already available if I'm not wrong), but this may make things easier.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3557)

### DIFF
--- a/lock-impl/src/main/java/com/palantir/lock/impl/KnownClientLock.java
+++ b/lock-impl/src/main/java/com/palantir/lock/impl/KnownClientLock.java
@@ -92,4 +92,9 @@ public interface KnownClientLock {
      *         anonymous or if this is a read lock instead of a write lock.
      */
     void unlockAndFreeze();
+
+    /**
+     * Returns true if and only if the lock is believed to be held by this lock client.
+     */
+    boolean isHeld();
 }
--- a/lock-impl/src/main/java/com/palantir/lock/impl/LockServerLock.java
+++ b/lock-impl/src/main/java/com/palantir/lock/impl/LockServerLock.java
@@ -146,6 +146,11 @@ public class LockServerLock implements ClientAwareReadWriteLock {
         }
 
         @Override
+        public boolean isHeld() {
+            return sync.holdsReadLock(clientIndex);
+        }
+
+        @Override
         public String toString() {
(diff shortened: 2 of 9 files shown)


jeremyk-91:
Discussed offline - yes, because (1) this is called from CLIs, and (2) I'm not too concerned about this as it's not a regression (existing code can already throw a CME; practice here is just to retry)

PR #3558, opened 2018-10-05
Replacing with Apache 2.0

felixdesouza:
Replacing with Apache 2.0

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3558)


felixdesouza:
For PRs merged after this if they're open then they'd need to rebase (if the files are new, I can go through and flag these), if it's created after this PR merges then their IDE should cover this.

As for PRs merged before, not much has merged since I started working on this (Friday afternoon), so it should be okay.

I will make the eclipse changes shortly, good catch!

PR #3523, opened 2018-09-24
Autobatching subproject + use it

j-baker:
Have a simple generalised mechanism for autobatching and use it for
unlocking and timestamp fetching.

Should lead to (in general) much simpler code.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3523)

PR #3568, opened 2018-10-10
corroborating timestamp service

gokcanozakdag:
**Goals (and why)**:
Adding runtime checks for timestamps going back in time. Aims to prevent data corruption caused by incorrect timelock migrations.

Comparing the received timestamp with a in-memory stored previous version. To ensure thread safety, we compare the received timestamp with a snapshot of global lower bound taken before getFreshTimestamp() request.

Testing the corroborating timestamp service to check if it actually fails in case of going back in time.

- Used -1L for starting value of the lower bound; can it cause some problems?
- Is there any concurrency related issue I am missing?

`CorroboratingTimestampService`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3568)


timestamp-client/src/test/java/com/palantir/timestamp/CorroboratingTimestampServiceTests.java
```diff
@@ -0,0 +1,107 @@
...
+        rawService = new GoBackInTimeTimestampService();
+        corroboratingTimestampService = new CorroboratingTimestampService(rawService);
+    }
+
+    @Test
+    public void getFreshTimestampShouldFailIfGoesBackInTime() {
+        getIndividualTimestamps(20);
+        goBackInTimeFor(10);
+
+        assertThatThrownBy(corroboratingTimestampService::getFreshTimestamp).isInstanceOf(AssertionError.class);
```
felixdesouza: assert message please

timestamp-client/src/test/java/com/palantir/timestamp/CorroboratingTimestampServiceTests.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private void goBackInTimeFor(int numberOfTimestamps) {
+        rawService.goBackInTime(numberOfTimestamps);
+    }
+
+    static class GoBackInTimeTimestampService implements TimestampService, TimestampManagementService {
+        private final AtomicLong counter = new AtomicLong(0);
+
+        public void goBackInTime(long time) {
+            counter.getAndUpdate(x -> {
```
felixdesouza: `counter.decrementAndGet()`?

timestamp-client/src/main/java/com/palantir/timestamp/CorroboratingTimestampService.java
```diff
@@ -0,0 +1,70 @@
...
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A timestamp service decorator for introducing runtime validity checks on received timestamps.
+ */
+@ThreadSafe
+public class CorroboratingTimestampService implements TimestampService {
+
+    private final TimestampService delegate;
```
felixdesouza: Could we just put this inside the actual impl? It does feel auxiliary like tracing/stats collection, since it's critical to correctness. Plus it does mean less wrapper/delegate things.

timestamp-client/src/test/java/com/palantir/timestamp/CorroboratingTimestampServiceTests.java
felixdesouza: or just 
```java
x -> Math.max(0, x - time)
```

timestamp-client/src/main/java/com/palantir/timestamp/CorroboratingTimestampService.java
jeremyk-91: (I think we mean _doesn't_ feel auxiliary right?)

The impl is, at the root, a Feign proxy so we'll need to delegate at some point - though we could see if there's a way to consolidate things

timestamp-client/src/main/java/com/palantir/timestamp/CorroboratingTimestampService.java
```diff
@@ -0,0 +1,70 @@
...
+        TimestampRange timestampRange = delegate.getFreshTimestamps(numTimestampsRequested);
+
+        if (timestampRange.getLowerBound() <= snapshot) {
+            throw clocksWentBackwards(snapshot, timestampRange.getLowerBound());
+        }
+        updateLowerBound(timestampRange.getUpperBound());
+        return timestampRange;
+    }
+
+    private synchronized void updateLowerBound(long freshTimestamp) {
```
jeremyk-91: feels like we should use `AtomicLong` as opposed to the volatile+synchronized combo here

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+        if (freshTimestamp <= lowerBound) {
+            throw clocksWentBackwards(timestampLowerBound, freshTimestamp);
+        }
+    }
+
+    private synchronized void updateLowerBound(long freshTimestamp) {
+        lowerBound = Math.max(freshTimestamp, lowerBound);
+    }
+
+    private AssertionError clocksWentBackwards(long timestampLowerBound, long freshTimestamp) {
```
felixdesouza: nit: `static`

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public TimelockService delegate() {
+        return delegate;
+    }
+
+    @Override
+    public long getFreshTimestamp() {
+        return checkAndUpdate(delegate::getFreshTimestamp, x -> x, x -> x);
```
felixdesouza: not really sure we need the functional approach, it (imo) unnecessarily complicates `checkAndUpdate`. It should be as simple as:
```java
public long getFreshTimestamp() {
  long delegateTimestamp = delegate.getFreshTimestamp();
  checkAndUpdate(delegateTimestamp, delegateTimestamp);
  return delegateTimestamp;
}

public TimestampRange getFreshTimestamps(numRequested) {
  TimestampRange range = delegate.getFreshTimestamps(numRequested);
  checkAndUpdate(range.getLowerBound(), range.getUpperBound());
  return range;
}

public StartAtlasDbTransactionResponse startAtlasDbTransaction(request) {
  StartAtlaqsDbTransactionResponse response = delegate.startAtlasDbTransaction(request);
  long timestamp = response.freshTimestamp();
  checkAndUpdate(timestamp, timestamp);
  return response;
}
```

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockServiceTest.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Test
+    public void getFreshTimestampShouldFail() {
+        getIndividualTimestamps(20);
+        goBackInTimeFor(10);
+
+        assertThatThrownBy(timelockService::getFreshTimestamp)
+                .isInstanceOf(AssertionError.class)
+                .hasMessage(generateClocksWentBackwardsMessage(20, 11));
```
felixdesouza: let's not do this, just do a `hasMessageContaining`/`hasMessageStartingWith` instead of exact match, I have no idea just looking at this what this message is supposed to be.

felixdesouza: I prefer to avoid computation in tests, since it's primarily what is read in reviews and I don't want to compute things myself.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLockRequest request) {
+        return checkAndUpdate(() -> delegate.startAtlasDbTransaction(request),
+                StartAtlasDbTransactionResponse::freshTimestamp,
+                StartAtlasDbTransactionResponse::freshTimestamp);
+    }
+
+    private <T> T checkAndUpdate(Supplier<T> timestampContainerSupplier,
```
felixdesouza: `checkAndUpdate` -> `checkAndUpdateLowerBound`

felixdesouza:
also, is there a way to annotate why something is thread safe, similar to `@GuardedBy` or something?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * A timelock service decorator for introducing runtime validity checks on received timestamps.
+ */
+public class TimestampCorroboratingTimelockService implements AutoDelegate_TimelockService {
+    static final String CLOCKS_WENT_BACKWARDS_MESSAGE =
+            "Expected timestamp to be greater than %s, but a fresh timestamp was %s!";
+
+    private final TimelockService delegate;
+    private volatile long lowerBound = Long.MIN_VALUE;
```
j-baker: please use a final LongAccumulator for this (new LongAccumulator(Math::max, Long.MIN_VALUE)).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+            throw clocksWentBackwards(timestampLowerBound, freshTimestamp);
+        }
+    }
+
+    private synchronized void updateLowerBound(long freshTimestamp) {
+        lowerBound = Math.max(freshTimestamp, lowerBound);
+    }
+
+    private AssertionError clocksWentBackwards(long timestampLowerBound, long freshTimestamp) {
+        return new AssertionError(String.format(CLOCKS_WENT_BACKWARDS_MESSAGE, timestampLowerBound, freshTimestamp));
```
j-baker: this isn't really an AssertionError; this has to be an IllegalStateException, I think. Like, Assertion error implies bug in code; this could be bug in deployment.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
j-baker: basically you should never need 'synchronized' or 'volatile' in this class

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
j-baker: That implementation is very racey, Gokcan's is not!

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+import com.palantir.lock.v2.AutoDelegate_TimelockService;
+import com.palantir.lock.v2.IdentifiedTimeLockRequest;
+import com.palantir.lock.v2.StartAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.timestamp.TimestampRange;
+
+/**
+ * A timelock service decorator for introducing runtime validity checks on received timestamps.
+ */
+public class TimestampCorroboratingTimelockService implements AutoDelegate_TimelockService {
```
j-baker: I don't think this should be a forwarding timelock service; at most it should be a TimestampService (which is the perfect interface for this)

j-baker: in fact: even this on its own is weird, because in practice with https://github.com/palantir/atlasdb/pull/3523 we'll never end up calling some of the methods

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
j-baker: in fact; make it a SafeIllegalStateException and add safe args with the params you care about

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+        timelockService.startAtlasDbTransaction(IDENTIFIED_TIME_LOCK_REQUEST);
+        goBackInTimeFor(1);
+
+        assertThatThrownBy(() -> timelockService.startAtlasDbTransaction(IDENTIFIED_TIME_LOCK_REQUEST))
+                .isInstanceOf(AssertionError.class)
+                .hasMessage(generateClocksWentBackwardsMessage(1, 1));
+    }
+
+    private static String generateClocksWentBackwardsMessage(long lowerBound, long freshTimestamp) {
+        return String.format(TimestampCorroboratingTimelockService.CLOCKS_WENT_BACKWARDS_MESSAGE,
```
j-baker: this isn't what you want. In particular, the reason you assert on messages is to make sure the result was right, whereas this does not achieve that.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+import com.palantir.lock.v2.AutoDelegate_TimelockService;
+import com.palantir.lock.v2.IdentifiedTimeLockRequest;
+import com.palantir.lock.v2.LockImmutableTimestampResponse;
+import com.palantir.lock.v2.StartAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.timestamp.TimestampManagementService;
+import com.palantir.timestamp.TimestampRange;
+import com.palantir.timestamp.TimestampService;
+
+public class TimestampCorroboratingTimelockServiceTest {
```
j-baker: these tests feel overengineered... you should be able to test all this appropriately in <100 lines

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+import com.palantir.lock.v2.AutoDelegate_TimelockService;
+import com.palantir.lock.v2.IdentifiedTimeLockRequest;
+import com.palantir.lock.v2.StartAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.timestamp.TimestampRange;
+
+/**
+ * A timelock service decorator for introducing runtime validity checks on received timestamps.
+ */
+public final class TimestampCorroboratingTimelockService implements AutoDelegate_TimelockService {
```
j-baker: Copied from a now-dead comment

I don't think this should be a forwarding timelock service; at most it should be a TimestampService (which is the perfect interface for this)

In fact even on this own it's weird; with #3523 we'll never end up calling some of the methods and we never care about any thread safety because there's only one caller of timelock at a time

j-baker: For example, atop that PR, this change would be

```java
     @VisibleForTesting
     static Consumer<List<BatchElement<Integer, TimestampRange>>> consumer(TimestampService delegate) {
-        return batch -> {
-            long totalTimestamps = batch.stream().mapToLong(BatchElement::argument).reduce(0, Math::addExact);
-            long startInclusive = 0;
-            long endExclusive = 0;
-            for (BatchElement<Integer, TimestampRange> element : batch) {
-                int timestampsRequired = element.argument();
-                if (element.argument() <= endExclusive - startInclusive) {
-                    element.result().set(createExclusiveRange(startInclusive, startInclusive + timestampsRequired));
-                    startInclusive += timestampsRequired;
-                    totalTimestamps -= timestampsRequired;
+        return new Consumer<List<BatchElement<Integer, TimestampRange>>>() {
+            private long maxTimestampSeen = Long.MIN_VALUE;
+
+            @Override
+            public void accept(List<BatchElement<Integer, TimestampRange>> batch) {
+                long totalTimestamps = batch.stream().mapToLong(BatchElement::argument).reduce(0, Math::addExact);
+                long startInclusive = 0;
+                long endExclusive = 0;
+                for (BatchElement<Integer, TimestampRange> element : batch) {
+                    int timestampsRequired = element.argument();
+                    if (element.argument() <= endExclusive - startInclusive) {
+                        element.result().set(createExclusiveRange(startInclusive, startInclusive + timestampsRequired));
+                        startInclusive += timestampsRequired;
+                        totalTimestamps -= timestampsRequired;
+                    } else {
+                        TimestampRange requested = getFreshTimestamps(Ints.saturatedCast(totalTimestamps));
+                        startInclusive = requested.getLowerBound();
+                        endExclusive = Math.addExact(requested.getUpperBound(), 1);
+                        int toTake = Math.min(Ints.checkedCast(endExclusive - startInclusive), timestampsRequired);
+                        element.result().set(createExclusiveRange(startInclusive, startInclusive + toTake));
+                        startInclusive += toTake;
+                        totalTimestamps -= timestampsRequired;
+                    }
+                }
+            }
+
+            private TimestampRange getFreshTimestamps(int timestamps) {
+                TimestampRange range = getFreshTimestampsFromDelegate(delegate, timestamps);
+                if (maxTimestampSeen > range.getLowerBound()) {
+                    throw new RuntimeException("error");
                 } else {
-                    TimestampRange requested = getFreshTimestampsFromDelegate(
-                            delegate, Ints.saturatedCast(totalTimestamps));
-                    startInclusive = requested.getLowerBound();
-                    endExclusive = Math.addExact(requested.getUpperBound(), 1);
-                    int toTake = Math.min(Ints.checkedCast(endExclusive - startInclusive), timestampsRequired);
-                    element.result().set(createExclusiveRange(startInclusive, startInclusive + toTake));
-                    startInclusive += toTake;
-                    totalTimestamps -= timestampsRequired;
+                    maxTimestampSeen = range.getUpperBound();
                 }
             }
         };
```

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
+        return checkAndUpdate(() -> delegate.getFreshTimestamps(numTimestampsRequested),
+                TimestampRange::getLowerBound,
+                TimestampRange::getUpperBound);
+    }
+
+    @Override
```
jeremyk-91: Minor completeness issue - it seems strange to not do this for `lockImmutableTimestamp`. No Atlas transaction actually uses that so I don't regard this as blocking, but it _does_ mean that the order in which you wrap stuff suddenly becomes important because if you put the Bridging one around this and you are using an old timelock the check goes away

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
jeremyk-91: (also, thinking about it - do we still need the install time check?)

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
jeremyk-91: I think why we did it at this level was because we wanted to check the timestamps in 'compound ops' (lockImmutableTimestamp, startAtlasDbTransaction) which don't quite exist in TimestampService. Those methods make one RPC to timelock, so they don't actually call the underlying timestamp methods on the client side.

The compound ops aren't batched yet, though of course they could be batched in the future if we need to control timelock load

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
j-baker: Gokcan's takes a snapshot of the lower bound before it calls the server. Yours will fail because two concurrent requests will race with each other (A updates the lower bound, then B does).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
j-baker: we should merge the install check into this one; instead of initializing at Long.MIN_VALUE, initialize with the punch table one

gokcanozakdag: I will address this with a separate pr shortly, as I am not sure if the end result will be cleaner than the current state.

gokcanozakdag: #3594

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
felixdesouza: is it worth doing the batching work in a follow up (i.e. after this PR) to avoid this undesirable middle state?

jeremyk-91: It can be the case that the delegate gets called twice (or for that matter, O(num requests in batch) times), because of the contract of `TimestampService#getFreshTimestamps()` - returns at least 1, but not necessarily the number you ask for.

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,10 @@ develop
     *    - Type
          - Change
 
+    *    - |improved|
+         - Introduced runtime checks on client side for timestamps retrieved from timelock; aiming to prevent data corruption in case of timestamps going back in time, possibly caused by a misconducted timelock migration. This is a best effort for catching abnormalities on timestamps at runtime, and do not provide absolute protection.
```
jeremyk-91: A bit of pedantry/grammar-related nits. I'd suggest rewriting this a little to split the sentences, though non-blocking:

Introduced runtime checks on _the_ client side for timestamps retrieved from timelock. _This aims_ to prevent data corruption _if_ timestamps _go_ back in time, possibly caused by a misconducted timelock migration. This is a best effort for catching abnormalities on timestamps at runtime, and _does_ not provide absolute protection.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
felixdesouza: I see! Thanks!

Issue #2933, opened 2018-01-29
Products should be able to specify a set of tables to truncate


hsaraogi:
This would allow cleaning up of cruft after migrations and unsupported tables.

sandorw:
Possible impl skeleton:
- Provide set of tables as an optional argument to `TransactionManagers` builder
- On start-up, do a best effort drop of those tables (fails if not all C* nodes online, so best effort is key)

Might be an interesting edge case around how we define the tables to be deleted, since we likely don't want them included in the schema definition at the point we want to drop them. We certainly don't want to recreate the tables after cleaning them up, for example.

tpetracca:
Can we just infer them? Look at the schema provided in code, look at what tables exist in the keyspace, calculate the diff.

sandorw:
Yes, that looks feasible. `KVS.getAllTableNames` fetches all non-KVS table names, including Atlas internal tables. Probably need to make sure that there's no cases where this deletes tables by accident, but I can't think of a good reason for removing a table from the schema without expecting it to be deleted.

tpetracca:
There's probably some edges here around dynamically created tables in PG land that we probably need to protect against.

sandorw:
Fair point, that's one argument for the more intentional version of explicitly listing tables.

clockfort:
Also probably issues around migration rollback if we immediately truncate the unlisted table

Issue #2918, opened 2018-01-25
Create a set of DEAD_TABLES


gsheasby:
After various AtlasDB schema changes, we have a few dead tables that are currently just `HIDDEN_TABLES`. Namely, `_scrub`, `_sweep_progress`, `_sweep_progress1_5`, and `_sweep_progress2`.

We should be treating these slightly differently, and can maintain a set of `DEAD_TABLES`, and then follow these rules:
- Do not sweep a dead table
- If there are dead tables present on startup, delete them

fsamuel-bs:
Also consider allowing a product to pass in `DEAD_TABLES`, to augment a hardcoded list. This would allow products to delete old tables which they've migrated off of and do not care anymore. We need to make clear that the service could not rollback to the old version anymore.

We need to clarify what "deleting a table" means in this case. Dropping the table would change the Cassandra schema which leaves us open to Cassandra schema consensus issues, while just truncating would delete all the data, while leaving the table behind. Would like to hear @tpetracca and @clockfort opinions here.

clockfort:
truncation also requires all nodes to be up and agree on the exact point in time a table is truncated

sandorw:
See also https://github.com/palantir/atlasdb/issues/2933

PR #3569, opened 2018-10-10
Truncate old tables

felixdesouza:
**Goals (and why)**:
Addresses #2933, but *not* #2918 for reasons stated below.

It makes the most sense when writing a schema and adding a *new* table to then add the deprecated table in the same breath => it is a method on `Schema`. 

`TransactionManagers` blast radius is huge, and doesn't make sense to integration test, so just added a test to `TransactionManagersInitializer`.

* I haven't addressed indexes, because it's not entirely trivial to get the name for an index as I believe we also need to add the suffix. An approach could include just adding both suffixes (for additive and cell referencing) provided the underlying kvs ignores missing tables.
* As discussed with @sandorw, the onus is on developers to ensure that they are not running conflicting versions in a multi-node configuration. If Node 1 is running a version that reads/writes from table x, and node 2 is running a version that considers table x deprecated, the developer must ensure that these configurations can't run together. This is similar to our [stance on table creation](https://github.com/palantir/atlasdb/blob/develop/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schemas.java#L102). 
* Following on from above, we're not attempting to drop atlas internal tables, because that would then push compatible ranges onto the consumer (despite their own schemas being self consistent). This can be handled at a later stage, when we have some way of "scheduling" tasks past a timestamp checkpoint.

Anywhere

Whenever, but let's not let it linger.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3569)

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schema.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schema.java
@@ -15,6 +15,12 @@
  */
 package com.palantir.atlasdb.table.description;
 
+import static java.util.stream.Collectors.toSet;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static com.google.common.collect.Maps.immutableEntry;
+import static com.palantir.logsafe.Preconditions.checkState;
+
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
@@ -38,6 +44,7 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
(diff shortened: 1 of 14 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/Schema.java
```diff
@@ -201,6 +215,24 @@ private void validateTableNameLength(String idxName) {
...
 
+    /**
+     * Registers old tables referred to by {@code deprecatedTableNames} to be dropped from backing store on startup.
+     *
+     * If running in a multi-node configuration, you must ensure that tables added in one version are not being
+     * read/written to in the other running version to avoid consistency problems.
+     *
+     * @param deprecatedTableNames old tables that should be dropped
+     */
+    public void addDeprecatedTables(String... deprecatedTableNames) {
```
sandorw: 👍 to putting this in the schema, good idea

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -345,25 +343,26 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
                         JavaSuppliers.compose(AtlasDbRuntimeConfig::targetedSweep, runtimeConfigSupplier)),
                 closeables);
 
-        Callback<TransactionManager> callbacks = new Callback.CallChain<>(ImmutableList.of(
-                timelockConsistencyCheckCallback(config, runtimeConfigSupplier.get(), lockAndTimestampServices),
+        Callback<TransactionManager> callbacks = new Callback.CallChain<>(
+                timelockConsistencyCheckCallback(config(), runtimeConfigSupplier.get(), lockAndTimestampServices),
                 targetedSweep.singleAttemptCallback(),
-                asyncInitializationCallback()));
+                new DeprecatedTablesCleaner(schemas()),
```
sandorw: Where, if anywhere, did we call `validate` on schemas before? We now additionally call it here, which is likely fine - just trying to understand where we're going to fail if we somehow sneak in an invalid schema.

PR #3577, opened 2018-10-13
Fix the stupid sweep things.

dxiao:
Unfortunately, it looks like there's no good way to
1. test for writes to empty namespace tables enqueued for targeted sweep
   before the table mapping migration.
2. test for oracle or postgres missing tables in this repo
So I guess I'm going to have to just fix this and hope

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
@@ -30,6 +30,7 @@ import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.keyvalue.impl.TableMappingNotFoundException;
 import com.palantir.atlasdb.logging.LoggingArgs;
 import com.palantir.atlasdb.sweep.Sweeper;
+import com.palantir.exception.PalantirSqlException;
 
 public class SweepQueueDeleter {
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
```diff
@@ -68,12 +69,11 @@ public void sweep(Collection<WriteInfo> writes, Sweeper sweeper) {
                                 kvs.deleteAllTimestamps(entry.getKey(), maxTimestampByCellPartition, true);
                             }
                         });
-            } catch (IllegalArgumentException e) {
+            } catch (IllegalArgumentException | PalantirSqlException e) {
```
sandorw: Did you track down why you ran into a `PalantirSqlException`? I'd still like to clean this up, so it would be good to note where this was thrown.

gmaretic:
If you want a postgres test, we can do the equivalent of CassandraTargetedSweepIntegrationTest and add it there

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
dxiao: Yeah, it was a OracleSqlException which got wrapped by a VerboseSQLException which got wrapped by a PalantirSqlException. The illegal argument exception was another way this could happen, through the table remapping

dxiao:
I think that would be good, yes.

Updated the release notes, whomever sees this first, please merge for me :)

PR #3579, opened 2018-10-16
Prevent AutoDelegate AutoFootShooting

gmaretic:
**Goals (and why)**:
Remove the `typeToExtend` parameter and extend the class that is annotated. This definitely prevents people from accidentally extending the same class multiple times.

Existing tests pass

This prevents extending external classes, but we have no such usages in atlasdb.

not urgent but small

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3579)


sandorw:
This will presumably break internal users of this, so be ready to fix them.

PR #3581, opened 2018-10-16
Targeted sweep now immediately cleans up dedicated rows

j-baker:
Don't dilly dally waiting for dedicated rows to be cleaned up; instead
just delete them immediately after sweeping them (since once we start
loading the contents of a dedicated row, we just delete all the
contents).

Notable gaps: Testing. I'm not sure I'm convinced it works. Is there
actually a test that the dedicated rows get cleaned up @gmaretic?
I couldn't really find one.

Making progress here is dependent on David's internal testing.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3581)

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
@@ -66,9 +66,15 @@ import com.palantir.atlasdb.cleaner.PuncherStore;
 import com.palantir.atlasdb.encoding.PtBytes;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.api.RangeRequest;
+import com.palantir.atlasdb.keyvalue.api.RowResult;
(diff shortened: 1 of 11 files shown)


j-baker:
Basically it feels like wrt the testing that I have tests that the data gets cleaned up, and tests that the row gets cleaned up if you deliberately try to clean it up, but no test that the dedicated rows get passed back and then get cleaned up.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -773,7 +773,7 @@ public void stopReadingEarlyWhenEncounteringEntryKnownToBeCommittedAfterSweepTs(
 
         // we now read all to the end
         sweepQueue.sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
-        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(4 + writesInDedicated + 3 + writesInDedicated + 2);
+        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(4 + 3 + writesInDedicated + 2);
```
j-baker: was this a test bug @gmaretic?

gmaretic:
Looks good now, checkstyle is sad though

PR #3586, opened 2018-10-18
Remove QoS service

gmaretic:
**Goals (and why)**:
Reduce unnecessary complexity

Minimal change that should catch all of the Qos things. There is a follow-up of fixing exceptions thrown by CKVSImpl, but that's out of scope and also nontrivial.
Hardwired the Qos metrics into the client created by CassandraClientFactory, removed the estimated and rate limiting metrics as those do not make sense in the new world.

As far as I'm aware, no one actually used this feature, but I might be wrong.
Do we need tests for metrics?

whenever

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3586)


sandorw:
I want to review this before it goes in - there's a few things we added that would be nice to keep (metrics for example).

j-baker:
Please make sure we didn't kill QosMetrics!

gmaretic:
Waiting on @sandorw to review

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/ThriftQueryWeighers.java
```diff
@@ -124,64 +98,24 @@ public QueryWeight weighFailure(Exception error, long timeTakenNanos) {
...
-    private static <T> QosClient.QueryWeigher<T> writeWeigher(long numRows, Supplier<Long> bytesWritten) {
-        Supplier<Long> weight = Suppliers.memoize(() -> safeGetNumBytesOrDefault(bytesWritten))::get;
+        return new QueryWeigher<T>() {
 
-        return new QosClient.QueryWeigher<T>() {
             @Override
-            public QueryWeight estimate() {
+            public QueryWeight weighSuccess(T result, long timeTakenNanos) {
                 return ImmutableQueryWeight.builder()
                         .from(DEFAULT_ESTIMATED_WEIGHT)
```
sandorw: It looks like we can remove `DEFAULT_ESTIMATED_WEIGHT` entirely, since we're setting all the attributes here and overriding the default parameters anyway. This is the only remaining use of the variable.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/qos/ThriftQueryWeighersTest.java
```diff
@@ -1,214 +0,0 @@
...
-import org.apache.cassandra.thrift.Mutation;
-import org.junit.Test;
-
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.palantir.atlasdb.qos.ImmutableQueryWeight;
-import com.palantir.atlasdb.qos.QosClient;
-import com.palantir.atlasdb.qos.QueryWeight;
-
-public class ThriftQueryWeighersTest {
```
sandorw: Any functionality here that needs to be tested outside of `QosCassandraClientTest`?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TodoEteTest.java
```diff
@@ -65,8 +65,8 @@ public void shouldSweepStreamIndices() {
         assertThat(valueSweep.getCellTsPairsExamined(), equalTo(9L));
         assertThat(valueSweep.getStaleValuesDeleted(), equalTo(4L));
 
-        // Stores five larger streams, which take 19 blocks each
-        StreamTestUtils.storeFiveStreams(todoClient, 7654321);
+        // Stores five larger streams, which take 3 blocks each
```
sandorw: ???

PR #3589, opened 2018-10-18
More HikariCP changes

jeremyk-91:
Fixing one bug and implementing one FR on top of the original change #3585

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3589)


jeremyk-91:
Thanks @schlosna!

PR #3592, opened 2018-10-19
Increase SnapshotTransaction tombstone exception's debugging info

mswintermeyer:
**Goals (and why)**:

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1089,7 +1089,8 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                                 + " This can be caused by hard delete transactions using the type "
                                 + TransactionType.AGGRESSIVE_HARD_DELETE
                                 + ". It can also be caused by transactions taking too long, or"
-                                + " its locks expired. Retrying it should work.");
+                                + " its locks expired. Retrying it should work."
+                                + " Table " + tableRef + ", cell " + key + ".");
                     default:
                         throw new IllegalStateException("Invalid read sentinel behavior " + getReadSentinelBehavior());
                 }


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1089,7 +1089,8 @@ private int estimateSize(List<RowResult<Value>> rangeRows) {
                                 + " This can be caused by hard delete transactions using the type "
                                 + TransactionType.AGGRESSIVE_HARD_DELETE
                                 + ". It can also be caused by transactions taking too long, or"
-                                + " its locks expired. Retrying it should work.");
+                                + " its locks expired. Retrying it should work."
+                                + " Table " + tableRef + ", cell " + key + ".");
```
felixdesouza: `TableRef`'s are potentially unsafe to log depending on schema. `key` is certainly unsafe. If this is strictly necessary, then I believe we want to have `TransactionFailedRetriableException` implement `SafeLoggable` and you pass the `key` and the `tableRef` as unsafe args.

mswintermeyer: When I was debugging an issue with sweep, I just patched atlas with a locally built patch of this. It was on a table without any sensitive info so I forgot what the various values could be. Makes sense that this can't go in as-is. I don't know too much about SafeLoggable / how to use it. Is there a way to configure it in-the-moment to be able to see "potentially-unsafe" info when it's necessary and manually verified as "actually-safe" to do so?

felixdesouza: Sorry I've been on holiday and I'm now just catching up on emails. I'm not sure what this means:

> Is there a way to configure it in-the-moment to be able to see "potentially-unsafe" info when it's necessary and manually verified as "actually-safe" to do so?

PR #3594, opened 2018-10-19
move timestamp checks to one place

gokcanozakdag:
**Goals (and why)**:
Move timestamp checks to one place, reduce complexity (hopefully)

- Add `validateWithConservativeLowerBound` to `TimestampCorroboratingTimelockService`, which takes a lower bound supplier and compares the result with a fresh timestamp.
- Use unreadable timestamp as conservative lower bound (that is not change)
- Register `validateWithConservativeLowerBound` as a `LambdaCallback`

Does it decrease the code complexity?

`TimestampCorroboratingTimelockService`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3594)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -295,7 +294,7 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         com.google.common.base.Supplier<TimestampManagementService> managementSupplier =
                 Suppliers.compose(atlasFactory::getTimestampManagementService, timestampSupplier);
 
-        LockAndTimestampServices lockAndTimestampServices = createLockAndTimestampServices(
+        LockAndTimestampServices rawLockAndTimestampServices = createLockAndTimestampServices(
```
jeremyk-91: nit: this isn't quite the case, as the lock and timestamp services returned here already have the bridging and lock refreshing clients in place.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -305,6 +304,10 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                 managementSupplier,
                 atlasFactory.getTimestampStoreInvalidator(),
                 userAgent());
+        TimestampCorroboratingTimelockService corroboratingTimelockService = TimestampCorroboratingTimelockService
+                .create(rawLockAndTimestampServices.timelock());
+        LockAndTimestampServices lockAndTimestampServices =
+                withCorroboratingTimestampService(rawLockAndTimestampServices, corroboratingTimelockService);
```
jeremyk-91: Non-actionable: Technically there is a change in behaviour here, because we've changed the wrapping order - in particular, before this change the check is instrumented (so will show up as part of the time taken for a getFreshTimestamps) while now only the inner call is instrumented. I think this is fine.

sandorw:
@gokcanozakdag what's the status on this PR?

gokcanozakdag:
@sandorw I will try to merge it today after addressing comments

PR #3596, opened 2018-10-21
Distinct Bounded Thread Pools for PaxosLeaderElectionService (with few modifications)

gokcanozakdag:
**Goals (and why)**:
Re-merging capacity limiter pr(#3505 ), with few modifications. 
- **Why we do we have #3505 at first place?** We have realized that using an unbounded thread pool for leader election service ends up creating large number of threads, causing service to either OOM or slow down drastically because of heap pressure. Refer to #3505 and #3489 for a more detailed explanation.
- **Why did we revert #3505?** Using a strictly bounded executor service caused `RejectedExecutionException` in case of attempts to submit more tasks than thread pool allows. Combined with a bug in `AwaitingLeadershipProxy`; this caused leader to respond with `NotCurrentLeaderException` for some clients (refer to internal ticket pds-75127).
- **Why can we merge it now?** 
  - Bug in the `AwaitingLeadershipProxy` is fixed in a separate pr here: #3582 .
  - A few modifications are made on the initial pr, aiming to prevent `RejectedExecutionException`, and continue to function normally if quorum of healthy nodes exist.

Modifications on #3505:
- Constructing `PaxosLatestRoundVerifier` with a distinct thread pool for each host, rather than using same thread pool for every host.
- `PaxosQuorumChecker` swallows `RejectedExecutionException`. Reason is:
  - Suppose we have clients for hosts A,B and C, with a dedicated thread pool for each. Suppose thread pool for C has reached its max (possibly because of an abnormality in connection to C). Calling `collectQuroumResponses` on `PaxosQuorumChecker` will attempt to submit the specified task for all three hosts, and throws `RejectedExecutionException` because of C. Swallowing the exception thrown by thread pool C would make it possible to achieve quorum as we still have A and B.

`MultiplexingCompletionService`; but if you are already familiar with #3505 , you can review the diff, which are modifications mainly in `PaxosLatestRoundVerifierImp` and `PaxosQuorumChecker`

This week?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3596)


atlasdb-commons/src/main/java/com/palantir/common/concurrent/MultiplexingCompletionService.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the BSD-3 License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://opensource.org/licenses/BSD-3-Clause
```
felixdesouza: license.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -118,23 +120,87 @@ public PaxosLeaderElectionService(PaxosProposer proposer,
...
+                        paxosLearners::get,
+                        index -> executorServiceFactory.apply("knowledge-update-" + index)));
+    }
+
+    private Map<PaxosAcceptor, ExecutorService> createLatestRoundVerifierExecutors(
+            List<PaxosAcceptor> paxosAcceptors,
+            Function<String, ExecutorService> executorServiceFactory) {
+        Map<PaxosAcceptor, ExecutorService> executors = Maps.newHashMap();
+        for (int i = 0; i < paxosAcceptors.size(); i++) {
+            executors.put(paxosAcceptors.get(i), executorServiceFactory.apply("lates-round-verifier-" + i));
```
felixdesouza: typo: `lates` -> `latest`

felixdesouza: it would be pretty nice for the `PaxosAcceptor` and `PaxosLearner` had uuid's similar to `PingableLeader`.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -194,14 +260,13 @@ private boolean pingLeader() {
         }
         final PingableLeader leader = maybeLeader.get();
 
-        CompletionService<Boolean> pingCompletionService = new ExecutorCompletionService<Boolean>(
-                executor);
+        MultiplexingCompletionService<PingableLeader, Boolean> multiplexingCompletionService
```
felixdesouza: why do we always create a new one? don't we lose progress between calls then?

felixdesouza: + aren't we then creating a lot of thread pools?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -258,15 +323,14 @@ boolean getAndRecordLeaderPingResult(@Nullable Future<Boolean> pingFuture) throw
...
+                = MultiplexingCompletionService.create(leaderPingExecutors);
 
         // kick off requests to get leader uuids
         List<Future<Entry<String, PingableLeader>>> allFutures = Lists.newArrayList();
         for (final PingableLeader potentialLeader : otherPotentialLeadersToHosts.keySet()) {
-            allFutures.add(pingService.submit(() -> new AbstractMap.SimpleEntry<>(
-                    potentialLeader.getUUID(),
-                    potentialLeader)));
+            allFutures.add(pingService.submit(potentialLeader,
+                    () -> new AbstractMap.SimpleEntry<>(potentialLeader.getUUID(), potentialLeader)));
```
felixdesouza: nit: `Maps.immutableEntry()`

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -258,15 +323,14 @@ boolean getAndRecordLeaderPingResult(@Nullable Future<Boolean> pingFuture) throw
     }
 
     private Optional<PingableLeader> getSuspectedLeaderOverNetwork(String uuid) {
-        CompletionService<Entry<String, PingableLeader>> pingService = new ExecutorCompletionService<>(
-                executor);
+        MultiplexingCompletionService<PingableLeader, Entry<String, PingableLeader>> pingService
```
felixdesouza: also curious about why we need to create this every time?

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLatestRoundVerifierImpl.java
```diff
@@ -30,12 +31,12 @@
...
     private final ImmutableList<PaxosAcceptor> acceptors;
     private final int quorumSize;
-    private final ExecutorService executor;
+    private final Map<PaxosAcceptor, ExecutorService> executor;
     private final Supplier<Boolean> onlyLogOnQuorumFailure;
 
     public PaxosLatestRoundVerifierImpl(
             List<PaxosAcceptor> acceptors, int quorumSize,
-            ExecutorService executor, Supplier<Boolean> onlyLogOnQuorumFailure) {
+            Map<PaxosAcceptor, ExecutorService> executor, Supplier<Boolean> onlyLogOnQuorumFailure) {
```
felixdesouza: `executors` or `executorsByAcceptors` or something.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -75,7 +80,7 @@ private PaxosQuorumChecker() {
             ImmutableList<SERVICE> remotes,
             final Function<SERVICE, RESPONSE> request,
             int quorumSize,
-            Executor executor,
+            ExecutorService executor,
```
felixdesouza: nit: `executor` -> `executorService`, here and in the rest of the file.

felixdesouza: dear god, how is this class just static, it's more than just utilities! *weeps

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
gokcanozakdag: This something I am also concerned with. We won't be losing progress, but not sure about the resource cost of creating this object again and again. I run a few load tests to see if it has an effect on perf, and it looks fine - but I would also want to hear @jeremyk-91 's thoughts here.

jeremyk-91: Regarding thread pools, we aren't re-creating them; the same set of executors (and hence same underlying thread pools) are being used even in separate instances of the MCS.

Think of MCS as a bridge that sends a task to some executors, and then has a queue where it collects the responses - the executors themselves aren't a part of an MCS. There might be some overhead for creating MCS and its completion queue, but the underlying executors are the same (thanks for load testing @gokcanozakdag!)

We have to create a separate MCS per request because responses across requests aren't generally interchangeable (e.g. a `getSuspectedLeaderOverNetwork()` invoked with different string arguments). For ping(), it may be possible to re-use responses in some way - though I didn't intend that to be part of this PR.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
jeremyk-91: (As mentioned above, responses to separate calls to `getSuspectedLeaderOverNetwork` aren't generally interchangeable, so we need a fresh one.)

atlasdb-commons/src/main/java/com/palantir/common/concurrent/MultiplexingCompletionService.java
```diff
@@ -0,0 +1,101 @@
...
+    private final ImmutableMap<K, ExecutorService> executors;
+    private final BlockingQueue<Future<V>> taskQueue;
+
+    private MultiplexingCompletionService(
+            ImmutableMap<K, ExecutorService> executors, BlockingQueue<Future<V>> taskQueue) {
+        this.executors = executors;
+        this.taskQueue = taskQueue;
+    }
+
+    public static <T, V> MultiplexingCompletionService<T, V> create(
```
jeremyk-91: nit: if using `<K, V>` then the params here should probably be `<K, V>` too

atlasdb-commons/src/test/java/com/palantir/common/concurrent/MultiplexingCompletionServiceTest.java
```diff
@@ -0,0 +1,182 @@
...
+        assertThatThrownBy(() -> boundedService.submit(KEY_1, sleepForFiveSeconds))
+                .isInstanceOf(RejectedExecutionException.class);
+    }
+
+    @Test
+    public void resilientToLargeNumberOfRequests() throws InterruptedException, ExecutionException {
+        MultiplexingCompletionService<String, Integer> boundedService = MultiplexingCompletionService.create(
+                ImmutableMap.of(KEY_1, createBoundedExecutor(2), KEY_2, createBoundedExecutor(2)));
+
+        for (int i = 0; i < 50_000; i++) {
```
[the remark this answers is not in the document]
jeremyk-91: Yep, 3 will do

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -236,7 +286,7 @@ private PaxosQuorumChecker() {
             }
         }, OUTSTANDING_REQUEST_CANCELLATION_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
 
-        if (log.isDebugEnabled() && shouldLogCancellationStatus()) {
+        if (log.isDebugEnabled() && shouldLogException()) {
```
jeremyk-91: nit: this one isn't really an exception though. `shouldLogDiagnosticInformation()` maybe?

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -116,23 +159,30 @@ private PaxosQuorumChecker() {
...
 
         // kick off all the requests
         List<Future<RESPONSE>> allFutures = Lists.newArrayList();
         for (final SERVICE remote : remotes) {
-            allFutures.add(responseCompletionService.submit(() -> request.apply(remote)));
+            try {
+                allFutures.add(responseCompletionService.submit(remote, () -> request.apply(remote)));
+            } catch (RejectedExecutionException e) {
+                if (shouldLogException()) {
+                    log.warn("Quorum checker executor rejected task", e);
```
jeremyk-91: Non-blocking: could be nice to log some information about the executor as an unsafe arg for debugging purposes. Probably not this PR though as it's a bit large already

atlasdb-commons/src/main/java/com/palantir/common/concurrent/MultiplexingCompletionService.java
gokcanozakdag: thanks! missed that one

PR #3597, opened 2018-10-22
_transactions2, Part 7: Burnin' Bridges / StartIdentifiedAtlasDbTransaction

jeremyk-91:
**Goals (and why)**:
- Actually use the new timelock endpoint, unblocking general transactions2 work

- Switch to using the new timelock endpoint.
- 🔥 **[USER BREAK] Remove the immutable timestamp bridging service**. I discussed with @sandorw and this does introduce a break... but the alternative is to have something that switches between _three_ possible ways of starting transactions which I don't really like.

- Not too much. The main thing here is whether using the new endpoint still delivers transactions respecting the timestamp guarantee (which the ETEs and integ tests should cover).

- Do you agree that burning bridges is reasonable and/or the right thing to do here?
- Is my argument about the existing tests sufficiently testing this sound?

**Where should we start reviewing?**: Commit by commit, ideally.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3597)


docs/source/release_notes/release-notes.rst
```diff
@@ -89,6 +89,19 @@ develop
          - Fixed a bug where ``AwaitingLeadershipProxy`` stops trying to gain leadership, causing client calls to leader to throw ``NotCurrentLeaderException``.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3582>`__)
 
+    *    - |userbreak|
+         - As part of preparatory work to migrate to a new transactions table, this version of AtlasDB and all versions going forward expect to be using a version of TimeLock that supports the ``startIdentifiedAtlasDbTransaction`` endpoint.
+           Support for previous versions of TimeLock has been dropped; please update your TimeLock server.
```
felixdesouza: is there a transitive dependency on the timelock server that we can add?

felixdesouza: sls-wise, that is

jeremyk-91: (discussed offline - this can be done internally through a plugin, though we still rely on users being up-to-date with that plugin)

sandorw: Should we consider https://github.com/palantir/atlasdb/blob/cdeb25712e51490fe9b0cc9b0497cf383660b8ce/atlasdb-cassandra/build.gradle#L86?

jeremyk-91: Ah, good call. I wasn't aware we _declared_ these in OSS, will add.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/ImmutableTimestampBridgingTimeLockService.java
```diff
@@ -1,98 +0,0 @@
...
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.RateLimiter;
-import com.palantir.atlasdb.http.errors.AtlasDbRemoteException;
-import com.palantir.lock.v2.AutoDelegate_TimelockService;
-import com.palantir.lock.v2.IdentifiedTimeLockRequest;
-import com.palantir.lock.v2.StartAtlasDbTransactionResponse;
-import com.palantir.lock.v2.TimelockService;
-
-/**
- * This class serves as a bridge to allow newer Atlas clients to maintain compatibility with TimeLock servers that
```
gmaretic: I think we've maintained this functionality long enough 👍

atlasdb-config/build.gradle
```diff
@@ -47,3 +48,12 @@ dependencies {
     // Needed for Jersey Response-based tests
     testCompile group: 'org.glassfish.jersey.core', name: 'jersey-common'
 }
+
+recommendedProductDependencies {
```
gmaretic: What does this do exactly?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -151,15 +153,17 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
             long immutableTs = transactionResponse.immutableTimestamp().getImmutableTimestamp();
             recordImmutableTimestamp(immutableTs);
 
-            cleaner.punch(transactionResponse.freshTimestamp());
-            Supplier<Long> startTimestampSupplier = Suppliers.ofInstance(transactionResponse.freshTimestamp());
+            cleaner.punch(transactionResponse.startTimestampAndPartition().timestamp());
+            Supplier<Long> startTimestampSupplier = Suppliers.ofInstance(
+                    transactionResponse.startTimestampAndPartition().timestamp());
 
             SnapshotTransaction transaction = createTransaction(immutableTs, startTimestampSupplier,
```
gmaretic: This and a couple of tests seem to be the only places the method is called, including large internal product. In each instance we have a supplier of instance, so why don't we change it to just a `long`? As a bonus, you can reorder the statements above to then `cleaner.punch(startTs)`

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -193,10 +194,13 @@ private TransactionManager setupTransactionManager() {
...
-        when(timelock.startAtlasDbTransaction(any())).thenReturn(
-                StartAtlasDbTransactionResponse.of(
-                        LockImmutableTimestampResponse.of(2L, LockToken.of(UUID.randomUUID())), 1L));
+        when(timelock.startIdentifiedAtlasDbTransaction(any())).thenReturn(
+                StartIdentifiedAtlasDbTransactionResponse.of(
+                        LockImmutableTimestampResponse.of(2L, LockToken.of(UUID.randomUUID())),
+                        TimestampAndPartition.of(1L, 1)
+                ));
         TRM.registerTransactionManager(txnManagerWithMocks);
+
```
gmaretic: nit: newline?

atlasdb-config/build.gradle
jeremyk-91: So if you're shipping a distribution and you include a dependency on this package and you have Gradle Preflight, then you'll fail to build. If you notice on `timelock-server` I've added an ignore on this dependency; if I didn't have that, the build would fail.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
jeremyk-91: Yep, will remove this. Must have accidentally added it

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
jeremyk-91: Discussed offline - we can push this one layer in, but unfortunately in `SnapshotTransaction` you still need the supplier since read only transactions grab a timestamp lazily

PR #3599, opened 2018-10-23
[WIP][DO NOT MERGE] Compare against existing client pool when refreshing pool

amandachow:
**Goals (and why)**:

Refactor of refreshPool() method in CassandraClientPoolImpl where servers added/removed are identified by comparing against servers in the current token range with the servers in the current (un-updated) client pool. This allows decommissioned c* nodes to be removed from the client pool without restarting the client.

The current behavior is:
- If auto-refresh is on, find all servers in the config or token range. Add servers not already in the client pool.
- If auto-refresh is off, remove servers in the client pool that are not in the config.

Behavior with this PR is:
- If auto-refresh is off:
	- Add all servers from the config to the client pool, if not there already
	- Remove any servers in the client pool that aren't in the config
- If auto-refresh is on:
	- Add to client pool all servers from the current token range that aren't already in the client pool
	- Remove from client pool any servers that aren't in the current token range.

Tested on an old branch I had: https://github.com/amandachow/atlasdb/tree/achow/refresh-remove-nodes
What I did was:
- Try a migration workflow to reproduce existing issue:
	- Create c* nodes 1, 2, 3
	- Add to cluster c* nodes 4, 5, 6
	- Decom 1, 2, 3 -- At this stage atlas client starts spewing errors
- Do same workflow with client running my branch (built locally)

Started a new branch (this one) because the change was so simple and my branch was so old/full of annoying logging I added, haven't done the same workflow with this branch but it doesn't look like this function has been touched since then anyway

Not sure how to approach writing tests. Also not sure if my understanding of the current behavior is correct. Also would like to walk through weird edge cases (such as create nodes 1, 2, 3, 4, then remove node 4, then spin up a new node on same host as node 4, or start atlasdb client with nodes 1, 2, 3, 4 in the config when actually the token range is 1, 2, 3, etc etc), especially around initialization

CassandraClientPoolImpl.refreshPool()

Two weeks

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3599)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -256,17 +256,19 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
     private synchronized void refreshPool() {
         blacklist.checkAndUpdate(cassandra.getPools());
 
-        Set<InetSocketAddress> serversToAdd = Sets.newHashSet(config.servers());
-        Set<InetSocketAddress> serversToRemove = ImmutableSet.of();
+        Set<InetSocketAddress> serversToRemove = Sets.newHashSet();
+        Set<InetSocketAddress> serversToAdd = Sets.newHashSet();
+        Set<InetSocketAddress> currentServers = cassandra.refreshTokenRanges();
 
         if (config.autoRefreshNodes()) {
-            serversToAdd.addAll(cassandra.refreshTokenRanges());
+            serversToRemove.addAll(Sets.difference(cassandra.getPools().keySet(), currentServers));
+            serversToAdd.addAll(Sets.difference(currentServers, cassandra.getPools().keySet()));
         }
 
-        serversToAdd = Sets.difference(serversToAdd, cassandra.getPools().keySet());
-
+        //what does this mean?????
         if (!config.autoRefreshNodes()) { // (we would just add them back in)
-            serversToRemove = Sets.difference(cassandra.getPools().keySet(), config.servers());
+            serversToAdd.addAll(Sets.difference(config.servers(), cassandra.getPools().keySet()));
+            serversToRemove.addAll(Sets.difference(cassandra.getPools().keySet(), config.servers()));
         }
 
         serversToAdd.forEach(cassandra::addPool);
@@ -274,9 +276,6 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
 
         if (!(serversToAdd.isEmpty() && serversToRemove.isEmpty())) { // if we made any changes
             sanityCheckRingConsistency();
-            if (!config.autoRefreshNodes()) { // grab new token mapping, if we didn't already do this before
-                cassandra.refreshTokenRanges();
-            }
         }
 
         log.debug("Cassandra pool refresh added hosts {}, removed hosts {}.",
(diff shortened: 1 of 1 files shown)


amandachow:
@tpetracca

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -256,29 +256,29 @@ int getMaxTriesTotal() {
...
 
-        Set<InetSocketAddress> serversToAdd = Sets.newHashSet(config.servers());
-        Set<InetSocketAddress> serversToRemove = ImmutableSet.of();
+        Set<InetSocketAddress> serversToRemove = Sets.newHashSet();
+        Set<InetSocketAddress> serversToAdd = Sets.newHashSet();
+        Set<InetSocketAddress> currentServers = cassandra.refreshTokenRanges();
 
         if (config.autoRefreshNodes()) {
-            serversToAdd.addAll(cassandra.refreshTokenRanges());
+            serversToRemove.addAll(Sets.difference(cassandra.getPools().keySet(), currentServers));
```
gmaretic: I don't know why we weren't doing this before tbh

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -256,29 +256,29 @@ int getMaxTriesTotal() {
...
+            serversToAdd.addAll(Sets.difference(currentServers, cassandra.getPools().keySet()));
         }
 
-        serversToAdd = Sets.difference(serversToAdd, cassandra.getPools().keySet());
-
+        //what does this mean?????
```
gmaretic: No idea what the comment was trying to convey, the whole logic here seems to just be a convoluted way of keeping the cassandra pool == config.servers, but we should never be removing anything since nothing else is modifying them and the config is fixed

amandachow:
I can remove the option to disallow autorefresh but I don't think it's necessary since the behavior (when disallowing refresh) shouldn't have changed here.

Currently it guarantees that at the end of this method, the client pool exactly matches the servers listed in config:
- Ensures all config servers are part of client pool https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java#L257
- Ensures all client pool servers are part of config https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java#L267

My change should do the same thing:
- Ensures all config servers are part of client pool https://github.com/palantir/atlasdb/blob/328c0115a4d5716eb6a0da35091bfb810130523e/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java#L270
- Ensures all client pool servers are part of config https://github.com/palantir/atlasdb/blob/328c0115a4d5716eb6a0da35091bfb810130523e/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java#L271

I agree with your code comment that shouldn't have been necessary because if autorefresh is off then it must have been off since the start (since can't live reload config), meaning it should've have initiated the client pool with the config servers then not changed it. Not sure if the code is just unnecessary and convoluted to begin with or if there's some undocumented edge case it was trying to guard against

gmaretic:
This has been merged in #3661

PR #3602, opened 2018-10-23
Change TransactionManagers.builder

felixdesouza:
Two changes:
1) `TransactionManagers.builder()` is no longer a staged builder, so no longer returns a `ConfigStage`. 
2) `TransactionManagers.Builder` now has a `serializable(Configurator<T extends TransactionManager> configurator)` method allowing alternate forms of creating transaction managers e.g. delegating to a library a way to finalise creation of a `TransactionManager` without allowing the consumer to intervene.

**Goals (and why)**:
A library needs to wrap a `TransactionManager` in order to provide safety guarantees. However, a non-wrapped `TransactionManager` cannot run with a wrapped `TransactionManager`. Goal is to force the consumer to pick one and not ever be in a state where they have both.

**Concerns**:
It does have a dev break, but since it's a builder, it's usually used fluently, so it should be fine, unless people are storing intermediate builders into variables (in my mind shouldn't be necessary, so I'm okay with the break here).

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3602)

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -215,7 +215,7 @@ public abstract class TransactionManagers {
     }
 
     public static Builder builder() {
-        return ImmutableTransactionManagers.builder();
+        return new Builder();
     }
 
     @VisibleForTesting
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -148,7 +148,6 @@ import com.palantir.util.JavaSuppliers;
 import com.palantir.util.OptionalResolver;
 
 @Value.Immutable
-@Value.Style(stagedBuilder = true)
 public abstract class TransactionManagers {
     private static final int LOGGING_INTERVAL = 60;
     private static final Logger log = LoggerFactory.getLogger(TransactionManagers.class);
@@ -205,7 +204,17 @@ public abstract class TransactionManagers {
         return Callback.noOp();
     }
 
-    public static ImmutableTransactionManagers.ConfigBuildStage builder() {
+    public interface Configurator<T extends TransactionManager> {
+        Supplier<T> configure(Builder builder);
+    }
+
+    public static class Builder extends ImmutableTransactionManagers.Builder {
+        <T extends TransactionManager> Supplier<T> serializable(Configurator<T> configurator) {
+            return configurator.configure(this);
+        }
+    }
+
+    public static Builder builder() {
         return ImmutableTransactionManagers.builder();
     }
 
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -252,7 +252,7 @@ public abstract class TransactionManagers {
     }
 
     @JsonIgnore
-    @Value.Derived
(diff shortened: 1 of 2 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -205,8 +204,18 @@ boolean validateLocksOnReads() {
         return Callback.noOp();
     }
 
-    public static ImmutableTransactionManagers.ConfigBuildStage builder() {
-        return ImmutableTransactionManagers.builder();
+    public interface Configurator<T extends TransactionManager> {
+        Supplier<T> configure(Builder builder);
```
JacekLach: Hm, why enforce that this must be a `Supplier<T>`?

I'd just do

```
public interface Configurator<T> {
  T configure(Builder builder);
}
```

since the end user will know what T is, and therefore get back exactly the specific type, whether it be `Supplier<TM>` or `Future<TM>` etc

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -205,8 +204,18 @@ boolean validateLocksOnReads() {
...
+    }
+
+    public static class Builder extends ImmutableTransactionManagers.Builder {
+        <T extends TransactionManager> Supplier<T> serializable(Configurator<T> configurator) {
+            return configurator.configure(this);
```
JacekLach: We could prevent an off scenario by invalidating the builder after calling configurator:

```
public static class Builder extends ImmutableTransactionManagers.Builder {
        private boolean valid = true;
        <T extends TransactionManager> Supplier<T> serializable(Configurator<T> configurator) {
            T value = configurator.configure(this);
            this.valid = false
            return value;
         }

        @Override
        public ImmutableTransactionManagers build() {
          if (!valid) { throw new InvalidStateException blah }
          return super.build();
        }
```

JacekLach: But maybe not important since we already allow the footgun of:

```
TM.Builder builder = ...
tm1 = builder.build().serializable();
tm2 = builder.build().serializable();
```

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -148,7 +148,6 @@
 import com.palantir.util.OptionalResolver;
 
 @Value.Immutable
-@Value.Style(stagedBuilder = true)
```
j-baker: fwiw, the staged builder bit was _great_; meant it was basically impossible to construct the txn manager wrong

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -243,7 +252,7 @@ public static TransactionManager createInMemory(Set<Schema> schemas) {
     }
 
     @JsonIgnore
-    @Value.Derived
+    @Value.Lazy
```
j-baker: read the method body! you can't make this change

felixdesouza: The method body is effectively 600-700 lines of wiring, what are you referring to exactly? It's also weird that just building a `TransactionManagers` starts kicking off everything which is what `@Value.Derived` does. Prior to the change you've suggested in this branch, it meant we couldn't complete it another way i.e. pass in the `Configurator` to the overloaded `serializable` method, because all the initialisation would have been kicked off by the time you've called `.build()`. Less of an issue with your suggested changes however.

PR #3604, opened 2018-10-23
Prevent raw->overflow conversions

sandorw:
**Goals (and why)**:
Throw for schema changes that would otherwise fail at runtime when the overflow column doesn't exist.

Detect cases where we are updating table metadata due to a schema change and ensure that the overflow column is in place if it should be. If it isn't, we throw, although we might try to fix the problem at some point in the future instead.

No tests included here. There is an open PR on an internal testing repo with added tests for this coverage. They pass when I run them locally on top of this commit. Shortly after this merges, we can merge the tests in the internal repo.

Any edge cases I'm missing? Code style nits?

`OracleDdlTable`

Would be nice to merge this week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3604)


atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -85,19 +86,26 @@ public static OracleDdlTable create(
...
         }
 
-        boolean needsOverflow = false;
-        TableMetadata metadata = TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(tableMetadata);
-        if (metadata != null) {
-            needsOverflow = metadata.getColumns().getMaxValueSize() > AtlasDbConstants.ORACLE_OVERFLOW_THRESHOLD;
+        String shortTableName = createTable(needsOverflow);
+
+        if (needsOverflow && !overflowColumnExists(shortTableName)) {
+            throwForMissingOverflowTable();
```
jeremyk-91: This part is to handle races between multiple nodes in an upgrade right?

sandorw: It could handle races or simply startup failures. `createTable` is the first modification we make, and the last one is to insert the table into the metadata table, so any failure in between would leave us in an inconsistent state that wouldn't be caught by the first `throwForMissingOverflowTable`.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/oracle/OracleDdlTable.java
```diff
@@ -107,7 +115,31 @@ public void create(byte[] tableMetadata) {
...
 
-    private void createTable(boolean needsOverflow) {
+    private boolean needsOverflow(byte[] tableMetadata) {
+        boolean needsOverflow = false;
+        TableMetadata metadata = TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(tableMetadata);
+        if (metadata != null) {
+            needsOverflow = metadata.getColumns().getMaxValueSize() > AtlasDbConstants.ORACLE_OVERFLOW_THRESHOLD;
+        }
+        return needsOverflow;
+    }
```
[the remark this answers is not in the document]
sandorw: Done 👍

PR #3605, opened 2018-10-23
Remove the Schema Metadata Service

jeremyk-91:
**Goals (and why)**:
- Disable the schema metadata service, as it was causing support issues like PDS-76193.

- Remove the code for SMS.
- Leave the schema metadata table in AtlasDbConstants as a hidden table; this must remain since it was atomic and attempts to BG sweep it will likely explode.

Nothing in particular. The service was producing schema metadata, but this metadata was never consumed.

- Is this the right decision, as opposed to merely disabling registering the SMS in TransactionManagers.create()? I think so as we aren't pursuing this short term and it is easily recoverable if we want to get some of this stuff back.
- Did I miss anything?

**Where should we start reviewing?**: `TransactionManagersInitializer`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3605)


felixdesouza:
wow! here I only thought that schema metadata service was like 150 lines aha

gmaretic:
The flake you got should be fixed on develop

PR #3606, opened 2018-10-23
less tablescans for grabbing CKVS tablemetadata (#3278)

clockfort:
backport for customer issue 76375

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3606)


clockfort:
deciding not to do the backport based on other missing code from this old a branch, that I don't want to also backport, that interacts poorly with this patch

PR #3608, opened 2018-10-24
[_transactions2] Coordination, Part 1: Coordination Store

jeremyk-91:
**Goals (and why)**:
- Unblock transactions2 work which involves knowledge of when to move from transactions1 to transactions2.

- Implement a coordination store, which stores a _sequence_. Sequences use one row and dynamic columns to store a pair `(k, v)` where `v` regularly changes but `k` does not. It is intended that `k` is the metadata map and `v` is the validity as far as transactions2 is concerned. The main reason for this is because you cannot do check-and-set in Oracle if the value is large enough that it goes into an overflow table.
- [x] TODO at time of writing: Capture coordination store schema in a JavaDoc comment.

This is mainly new code, which is tested.
I refactored the CheckAndSetResult to be more general, but the Cassandra use cases should already be tested by the CAS tests in AbstractKVS testing, plus Cassandra integration/ETEs as those use the transactions table which uses CAS.

- I don't like the massive pile of metadata at the top of the KVS coordination store, but I think it's important to specify that it has a SweepStrategy.NOTHING.
- Is the schema good / are there alternative schemas I should have picked?
- Are there any other lists of 'special non-transactional tables' I need to add this one to?

**Where should we start reviewing?**: `CoordinationStore` and `KeyValueServiceCoordinationStoreTest`

**Priority (whenever / two weeks / yesterday)**: this week? or early next week

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3608)


atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,181 @@
...
+            0,
+            false,
+            TableMetadataPersistence.SweepStrategy.NOTHING, // we do our own cleanup
+            false,
+            TableMetadataPersistence.LogSafety.SAFE);
+    private static final long WRITE_TIMESTAMP = 0L;
+    private static final ObjectMapper OBJECT_MAPPER = ObjectMappers.newServerObjectMapper();
+
+    private final KeyValueService kvs;
+    private final byte[] coordinationSequence;
```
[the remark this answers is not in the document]
jeremyk-91: It is the row key used in the table - for transactions2 we only plan on using one sequence, but if we want more sequences in the future we don't want to use additional tables since that has some other overheads.

(This will be in the TODO item I mention above on javadoc of the schema)

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/SequenceAndBound.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+/**
+ * A pair of a sequence number and a bound on the validity of the value associated with that sequence number.
+ */
+@Value.Immutable
+@JsonSerialize(as = ImmutableSequenceAndBound.class)
+@JsonDeserialize(as = ImmutableSequenceAndBound.class)
```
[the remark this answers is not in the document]
jeremyk-91: IIRC default is fail on unknown properties, which I _think_ is fine for the usage here since it's just which entry is the relevant one + the validity bound. We'll want ignore unknown on the entry object.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,181 @@
...
+
+    public KeyValueServiceCoordinationStore(KeyValueService kvs, byte[] coordinationSequence) {
+        this.kvs = kvs;
+        kvs.createTable(AtlasDbConstants.COORDINATION_TABLE, COORDINATION_TABLE_METADATA.persistToBytes());
+
+        this.coordinationSequence = coordinationSequence;
+    }
+
+    @Override
+    public Optional<byte[]> getValue(long sequenceNumber) {
```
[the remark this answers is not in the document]
felixdesouza: fair enough, I'm not fussed which one, it was more of the equality thing, and delegating to something that already existed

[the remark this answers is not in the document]
felixdesouza: although it's exposing okio everywhere, I suppose it's fine since it's internal

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,201 @@
...
+        return coordinationStore;
+    }
+
+    @Override
+    public Optional<ByteString> getValue(long sequenceNumber) {
+        Preconditions.checkState(
+                sequenceNumber > 0,
+                "Only positive sequence numbers are supported, but found %s",
+                sequenceNumber);
+        Cell targetCell = getCellForSequence(sequenceNumber);
```
gmaretic: Extract a method that takes a sequence number and returns Optional<Value> to get rid of duplication in getCoordinationValue?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,201 @@
...
+                = kvs.get(AtlasDbConstants.COORDINATION_TABLE, ImmutableMap.of(targetCell, Long.MAX_VALUE));
+        return Optional.ofNullable(response.get(targetCell))
+                .map(Value::getContents)
+                .map(KeyValueServiceCoordinationStore::deserializeSequenceAndBound);
+    }
+
+    @Override
+    public CheckAndSetResult<SequenceAndBound> checkAndSetCoordinationValue(
+            Optional<SequenceAndBound> oldValue, SequenceAndBound newValue) {
+        CheckAndSetRequest request = oldValue
```
gmaretic: I'm fine either way, but could be easier to read using CheckAndSetRequest.Builder directly

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Attempts to atomically update the {@link SequenceAndBound} associated with the relevant
+     * {@link CoordinationService}.
+     *
+     * @param oldValue old value of the sequence and bound
+     * @param newValue new value of the sequence and bound
+     * @return a {@link CheckAndSetResult} indicating success or failure of the operation, and the current value
+     */
+    CheckAndSetResult<SequenceAndBound> checkAndSetCoordinationValue(
+            Optional<SequenceAndBound> oldValue, SequenceAndBound newValue);
```
gmaretic: @SuppressWarnings("OptionalUsedAsFieldOrParameterType")

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
```diff
@@ -0,0 +1,64 @@
...
+    Optional<ByteString> getValue(long sequenceNumber);
+
+    /**
+     * Stores a value for a given sequence number.
+     * Sequence numbers are expected to be strictly positive.
+     *
+     * @param sequenceNumber sequence number to store a value for
+     * @param value value to be stored
+     */
+    void putValue(long sequenceNumber, ByteString value);
```
gmaretic: We should be clear about repeated puts for the same sequenceNumber with different values not being supported. Probably fine to say it's nondeterministic bit idempotent and leave it at that?

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
```diff
@@ -91,6 +92,7 @@ private AtlasDbConstants() {
             PARTITION_MAP_TABLE,
             PERSISTED_LOCKS_TABLE,
             SWEEP_PROGRESS_TABLE,
+            COORDINATION_TABLE,
```
sandorw: There's also the `ATOMIC_TABLES` set, although it's missing some other tables too (targeted sweep table ref dictionary).

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
sandorw: Does it need to be nondeterministic? Can we just fail instead?

edit: I see that we can't, since the table is non-transactional and the values are potentially too big to CAS. Might still be worth best effort protection, since a race is pretty unlikely? Also unclear to me that not failing is acceptable, since this could cause corruption.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+
+import com.palantir.atlasdb.keyvalue.impl.CheckAndSetResult;
+
+import okio.ByteString;
+
+/**
+ * A {@link CoordinationStore} stores data that a {@link CoordinationService} may use.
+ */
+public interface CoordinationStore {
```
sandorw: Alternately, we could hide the sequence number from users of the store - putting a value does the equivalent of `putValue` followed by `checkAndSetCoordinationValue`. As it is, it's not immediately clear what the sequence number represents, and it doesn't look like it's meaningful outside of being used in `getValue`. If `getCoordinationValue` returned a `ByteStreamAndBound`, would that remove the need? One downside would be that the client can't cache values of known sequence numbers, but we could put that logic in the store (alongside disallowing updates to existing sequence numbers, which we can easily do internally).

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,201 @@
...
+import com.palantir.remoting3.ext.jackson.ObjectMappers;
+
+import okio.ByteString;
+
+/**
+ * An implementation of {@link CoordinationStore} that persists its state in a {@link KeyValueService}.
+ *
+ * Specifically, a {@link KeyValueServiceCoordinationStore} stores its data in a single row in the
+ * {@link AtlasDbConstants#COORDINATION_TABLE}. This specific row is identified by the store's
+ * {@link KeyValueServiceCoordinationStore#coordinationSequence}; the reason for doing this is that it allows
```
sandorw: This is pretty confusing - is there any relation between `CoordinationStore#sequenceNumber` and `KeyValueServiceCoordinationStore#coordinationSequence`?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,201 @@
...
+    private final byte[] coordinationSequence;
+
+    private KeyValueServiceCoordinationStore(KeyValueService kvs, byte[] coordinationSequence) {
+        this.kvs = kvs;
+        this.coordinationSequence = coordinationSequence;
+    }
+
+    public static CoordinationStore create(KeyValueService kvs, byte[] coordinationSequence) {
+        CoordinationStore coordinationStore = new KeyValueServiceCoordinationStore(kvs, coordinationSequence);
+        kvs.createTable(AtlasDbConstants.COORDINATION_TABLE, COORDINATION_TABLE_METADATA.persistToBytes());
```
sandorw: Not sure when we'll create this - will this fit in with out current async init code?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -0,0 +1,201 @@
...
+                            .build())),
+            new ColumnMetadataDescription(new DynamicColumnDescription(
+                    NameMetadataDescription.create(
+                            ImmutableList.of(
+                                    new NameComponentDescription.Builder()
+                                            .componentName("sequenceNumber")
+                                            .logSafety(TableMetadataPersistence.LogSafety.SAFE)
+                                            .type(ValueType.VAR_LONG)
+                                            .build())),
+                    ColumnValueDescription.forType(ValueType.VAR_LONG))),
```
sandorw: Should this be `BLOB`? Sequence values certainly seem to be blobs, and the special sequence number 0 is a `SequenceAndBound`.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
gmaretic: As long as you call this from something that is being async initialized, yes

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
jeremyk-91: @sandorw I think I will change this to failing. It _is_ possible to fail, actually (to be clear: the issue blocks CAS, but _not_ PUE as far as I know - I will confirm this with a test).

Thinking about this, values shouldn't be written too often (the bound is written a lot, but you don't write a new value unless you must), so it's probably fair to incur the cost of one round of Cassandra's Paxos when doing this

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
```diff
@@ -0,0 +1,64 @@
...
+     * Sequence numbers are expected to be strictly positive.
+     *
+     * @param sequenceNumber sequence number to read a value for
+     * @return value stored at that sequence number; empty if none is present
+     */
+    Optional<ByteString> getValue(long sequenceNumber);
+
+    /**
+     * Stores a value for a given sequence number.
+     * Sequence numbers are expected to be strictly positive.
```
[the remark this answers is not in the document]
jeremyk-91: (we're changing this API post discussions with @sandorw)

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
felixdesouza: Do we actually flag this in our repo currently?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
jeremyk-91: That is _much_ better actually, it becomes

```
        CheckAndSetRequest request = new CheckAndSetRequest.Builder()
                .table(AtlasDbConstants.COORDINATION_TABLE)
                .cell(getCoordinationValueCell())
                .oldValue(oldValue.map(KeyValueServiceCoordinationStore::serializeSequenceAndBound))
                .newValue(serializeSequenceAndBound(newValue))
                .build();
```

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
jeremyk-91: It does come up in IntelliJ I think, which might have been bothering @gmaretic

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
jeremyk-91: No, there isn't a relation. I think I'll rename this to `coordinationRow` as the old "coordinationSequence" was the row in the table we'd use to write the dynamic columns.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
jeremyk-91: Discussed offline - schema changed to `ValueAndBound<T>`. The idea makes sense and contains, though I chose to use `T` rather than `ByteStreamAndBound` as having users provide transforms from `ByteStream -> ByteStream` seemed a bit excessive.

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbConstants.java
jeremyk-91: Added to `ATOMIC_TABLES` - I'm not aware if it's particularly significant, but indeed this table would be unhappy on a non-pUE supporting KVS

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -172,21 +206,31 @@ public void putValue(long sequenceNumber, ByteString value) {
     }
 
     private static SequenceAndBound deserializeSequenceAndBound(byte[] contents) {
+        return deserializeData(contents, SequenceAndBound.class, "coordination sequence and bound");
+    }
+
+    private static <T> T deserializeData(byte[] data, Class<T> clazz, String safeDescriptionOfItemToDeserialize) {
```
sandorw: This is static, so it can't use the class variable anyway, but should we rename the class var to `valueClass` for clarity?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -172,21 +206,31 @@ public void putValue(long sequenceNumber, ByteString value) {
...
         try {
-            return OBJECT_MAPPER.readValue(contents, SequenceAndBound.class);
+            return OBJECT_MAPPER.readValue(data, clazz);
```
sandorw: Should probably let the user inject an object mapper that will work

atlasdb-coordination-impl/src/test/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStoreTest.java
```diff
@@ -70,22 +88,18 @@ public void throwsIfAttemptingToGetAtNegativeSequenceNumber() {
...
-        assertThat(coordinationStore.getCoordinationValue()).isEmpty();
-    }
-
-    @Test
-    public void storesAndRetrievesCoordinationValues() {
-        assertThat(coordinationStore.checkAndSetCoordinationValue(Optional.empty(), SEQUENCE_AND_BOUND_1))
-                .isEqualTo(ImmutableCheckAndSetResult.of(true, ImmutableList.of(SEQUENCE_AND_BOUND_1)));
-        assertThat(coordinationStore.getCoordinationValue())
-                .contains(SEQUENCE_AND_BOUND_1);
+    public void throwsIfAttemptingToPutTwice() {
```
sandorw: 🏅

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractKeyValueServiceTest.java
```diff
@@ -1506,6 +1506,15 @@ public void testCheckAndSetAndBackAgain() {
         verifyCheckAndSet(TEST_CELL, val(0, 0));
     }
 
+    @Test
+    public void testCheckAndSetLargeValue() {
+        byte[] megabyteValue = new byte[1048756];
+        CheckAndSetRequest request = CheckAndSetRequest.newCell(TEST_TABLE, TEST_CELL, megabyteValue);
+
+        keyValueService.checkAndSet(request);
```
sandorw: Should we be testing `putUnlessExists` instead or in addition?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -69,96 +71,128 @@
...
+        return extractRelevantValues(targetValue, newBound, casResult);
+    }
+
+    private CheckAndSetResult<ValueAndBound<T>> extractRelevantValues(T targetValue, long newBound,
+            CheckAndSetResult<SequenceAndBound> casResult) {
+        if (casResult.successful()) {
+            return CheckAndSetResult.of(true, ImmutableList.of(ValueAndBound.of(Optional.of(targetValue), newBound)));
+        }
+        return CheckAndSetResult.of(
+                casResult.successful(),
```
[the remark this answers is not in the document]
jeremyk-91: Ah, yeah. The main point of this method is 'avoid doing another DB read if you know you were successful' - agree this would be clearer as `false`.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -69,96 +71,128 @@
...
+                casResult.existingValues()
+                        .stream()
+                        .map(value -> ValueAndBound.of(getValue(value.sequence()), value.bound()))
+                        .collect(Collectors.toList()));
+    }
+
+    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
```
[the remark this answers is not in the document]
jeremyk-91: Good point - this used to take an Optional<ValueAndBound> but now doesn't. I'll remove this

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -69,96 +71,128 @@
...
                 "Only positive sequence numbers are supported, but found %s",
                 sequenceNumber);
-        kvs.put(AtlasDbConstants.COORDINATION_TABLE,
-                ImmutableMap.of(getCellForSequence(sequenceNumber), value.toByteArray()),
-                WRITE_TIMESTAMP);
+        try {
+            kvs.putUnlessExists(AtlasDbConstants.COORDINATION_TABLE,
+                    ImmutableMap.of(getCellForSequence(sequenceNumber), serializeData(value, VALUE_DESCRIPTION)));
+        } catch (KeyAlreadyExistsException e) {
+            throw new IllegalStateException("The coordination store failed a putUnlessExists. This is unexpected"
```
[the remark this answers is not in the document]
jeremyk-91: That's true. Actually the way we use sequence numbers to get the new bound is kind of dependent on timestamps. I think we can accept this leak - and I've made the error message emphasise the severity more.

PR #3613, opened 2018-10-25
Don't retry schema mutation requests...

j-baker:
This is a little silly. We shouldn't be retrying schema mutation
requests because they re-introduce the race condition the schema
mutation lock is supposed to solve in the presence of timeouts.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3613)

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -1504,7 +1504,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     }
 
     private void createTablesInternal(final Map<TableReference, byte[]> tableNamesToTableMetadata) throws Exception {
-        clientPool.runWithRetry(client -> {
+        clientPool.run(client -> {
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1504,7 +1504,7 @@ private void runSchemaMutationTask(Runnable task) {
     }
 
     private void createTablesInternal(final Map<TableReference, byte[]> tableNamesToTableMetadata) throws Exception {
-        clientPool.runWithRetry(client -> {
+        clientPool.run(client -> {
```
j-baker: oh even this doesn't quite fix it

j-baker: we have to leak the schema mutation lock deliberately if we throw

gmaretic:
Yeah, this won't work. What would work is: acquire lock, wait for schemas to settle, apply mutation talking to a node that is known to be in this good state, wait for schemas to settle. Also, we need to fix how quorum is calculated for waitForSchemaVersions, since doing n/2 + 1 is not correct :/
Alternatively, we can use cql and provide a cfId ourselves, but that gets tricky if we want to be able to recreate dropped tables

gmaretic:
Doing this in a different way

PR #3614, opened 2018-10-26
Do not break cassandra when creating and dropping tables

gmaretic:
**Goals (and why)**:
See internal ticket 76619. Prevent breaking your cassandra schema that can happen when a task requiring schema mutation locks is retried. 

This is a bit heavy-handed because it forces all nodes to be available before we allow table creation and dropping, but that is comparable to previous behaviour.

Existing tests have been updated to reflect the sotw

The following could have happened in atlas 0.107.0 and before, which was a bug, but may have increased availability:
We have quorum up, try to create table -> this requests the schema mutation then fails on waiting for versions to settle
We restart atlas, still only have quorum but now the table exists and we happily start since we only need a quorum of nodes to agree on schema versions if others are unavailable.

Now, we will never issue the request in the first place, so you will not be able to start atlas if you need to create a table and not all nodes are up. We should follow up next week with a fix where we supply the cfId ourselves as @j-baker suggested.

Small

Today & release

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3614)


gmaretic:
Made obsolete by https://github.com/palantir/atlasdb/pull/3620

PR #3619, opened 2018-10-30
[_transactions2] Coordination, Part 2: Coordination Service

jeremyk-91:
**Goals (and why)**:
- Provide a mechanism to support the main workflows the coordination service needs for _transactions2 - "is this value valid at this timestamp?" and "advance the state machine, considering both value and timestamp validity"
- Do the above efficiently, and without the user needing to be aware of the sequence-number layer that we need in the `CoordinationStore`

- Introduce a `CoordinationService` API for clients to use which supports the two workflows
- Implement `CoordinationServiceImpl`, which provides mechanisms for interfacing with a CoordinationStore and suitably transforming the currently active bound.

This is new code. I have added tests.

- Is the `CoordinationService` API too difficult to reason about? I'm not too happy with it, but have been thinking about it for a while and struggled to come up with something better that doesn't expose internals.
- Are there multithreading issues or race conditions I might have missed out on? The coordination service needs to work with high concurrency and across multiple nodes as well.
- Does it make sense to have the caching layer for the CoordinationService to be where it is, or should it be a separate decorator on top of the impl that knows how to talk to the DB?

**Where should we start reviewing?**: `CoordinationService` (interface), the guarantees are a little tricky.

**Priority (whenever / two weeks / yesterday)**: this week, ideally.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3619)

### DIFF
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.coordination;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * A {@link CoordinationService} is used to agree on values being relevant or correct at a given timestamp.
+ * The sequence of values being agreed should evolve in a backwards consistent manner; that is, if I read a value
+ * for some timestamp TS and make a decision at TS, all future values written to the {@link CoordinationService} should
+ * also ensure that no decision that would be inconsistent with the decision made at TS would be made at TS. In general,
+ * this means that one should avoid changes that affect behaviour of values below the validity bound.
+ *
+ * It is the responsibility of the caller to preserve the above property.
+ */
+public interface CoordinationService<T> {
+    /**
+     * Returns a value that the coordination service has agreed is appropriate at the provided timestamp.
+     *
+     * The value returned by this method may change over time. However, values returned by this method would
+     * be consistent at this timestamp - that is, the service should never return values for the same timestamp
+     * that would be contradictory or result in different operation for uses at this timestamp.
+     *
+     * This operation will return an empty Optional if no value has been agreed upon yet, or if the current value agreed
+     * by the coordination service is not valid at the provided timestamp.
+     *
+     * @param timestamp timestamp to retrieve the coordinated value for
+     * @return value associated with that timestamp
+     */
+    Optional<ValueAndBound<T>> getValueForTimestamp(long timestamp);
+
+    /**
+     * Updates the value that this coordination service stores, marking this value as valid up to a certain timestamp.
+     *
+     * Evolutions of the value must be compatible in terms of backwards consistency as defined in the class docs.
+     *
+     * @param transform transformation to apply to the existing value and bound the coordination service agrees on
+     * @return true if and only if the transformation was applied
+     */
+    boolean tryTransformCurrentValue(Function<Optional<ValueAndBound<T>>, ValueAndBound<T>> transform);
+}
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
@@ -46,8 +46,13 @@ public interface CoordinationService<T> {
 
     /**
      * Updates the value that this coordination service stores, marking this value as valid up to a certain timestamp.
+     * It may transpire that the value in the coordination service is already acceptable - if so, the provided
+     * transformation should simply return the original input. If the transformation returns a value with a bound
+     * less than or equal to the bound of the input, the results of the transformation will not be stored.
      *
      * Evolutions of the value must be compatible in terms of backwards consistency as defined in the class docs.
+     *
+     * The {@link ValueAndBound} returned by the transform must contain a value.
      *
      * @param transform transformation to apply to the existing value and bound the coordination service agrees on
      * @return true if and only if the transformation was applied
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
@@ -23,12 +23,16 @@ import com.palantir.atlasdb.keyvalue.impl.CheckAndSetResult;
 
 /**
  * A {@link CoordinationService} is used to agree on values being relevant or correct at a given timestamp.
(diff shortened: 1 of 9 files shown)


atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+
+import com.palantir.atlasdb.keyvalue.impl.CheckAndSetResult;
+
+/**
+ * A {@link CoordinationService} is used to agree on values being relevant or correct at a given timestamp.
+ * The sequence of values being agreed should evolve in a backwards consistent manner; that is, if I read a value
+ * for some timestamp TS and make a decision at TS, all future values written to the {@link CoordinationService} should
+ * also ensure that no decision that would be inconsistent with the decision made at TS would be made at TS. In general,
```
sandorw: The idea makes sense, but this sentence should be rephrased to be clearer.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
```diff
@@ -0,0 +1,64 @@
...
+ * this means that one should avoid changes that affect behaviour of values below the validity bound.
+ *
+ * It is the responsibility of the caller to preserve the above property.
```
sandorw: How is this done? I'd assume that the service/store layer enforce that you can't change behavior for some old bound, so it's up to clients of this service to respect the behavior communicated for each bound.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public interface CoordinationService<T> {
+    /**
+     * Returns a value that the coordination service has agreed is appropriate at the provided timestamp.
+     *
+     * The value returned by this method may change over time. However, values returned by this method would
```
sandorw: May change here means that we may add some new feature that needs to change some behavior, but it would be applied in a way such that any value read here would be consistent with the behavior we were already using?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
```diff
@@ -0,0 +1,64 @@
...
+     * be consistent at this timestamp - that is, the service should never return values for the same timestamp
+     * that would be contradictory or result in different operation for uses at this timestamp.
+     *
+     * This operation will return an empty Optional if no value has been agreed upon yet, or if the current value agreed
+     * by the coordination service is not valid at the provided timestamp.
+     *
+     * @param timestamp timestamp to retrieve the coordinated value for
+     * @return value associated with that timestamp
+     */
+    Optional<ValueAndBound<T>> getValueForTimestamp(long timestamp);
```
sandorw: Which timestamp is this? I assume that, from a transaction's perspective, it must use consistent behavior across its start and commit timestamps?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
```diff
@@ -0,0 +1,64 @@
...
+     * less than or equal to the bound of the input, the results of the transformation will not be stored.
+     *
+     * Evolutions of the value must be compatible in terms of backwards consistency as defined in the class docs.
+     *
+     * The {@link ValueAndBound} returned by the transform must contain a value.
+     *
+     * @param transform transformation to apply to the existing value and bound the coordination service agrees on
+     * @return a {@link CheckAndSetResult} indicating whether the transform was applied and the current value
+     */
+    CheckAndSetResult<ValueAndBound<T>> tryTransformCurrentValue(
```
sandorw: Would it be better if the bounds management was handled internally? Clients would propose behavior, and internally we would make it apply at the next reasonable opportunity (= some future timestamp).

atlasdb-coordination-impl/src/test/java/com/palantir/atlasdb/coordination/keyvalue/CoordinationServiceImplTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final ImmutableValueAndBound<String> STRING_AND_ONE_HUNDRED
+            = ImmutableValueAndBound.of(Optional.of("string"), 100);
+    private static final ImmutableValueAndBound<String> OTHER_STRING_AND_ONE_THOUSAND
+            = ImmutableValueAndBound.of(Optional.of("otherstring"), 1000);
+    private static final ImmutableValueAndBound<String> ANOTHER_STRING_AND_ONE_HUNDRED
+            = ImmutableValueAndBound.of(Optional.of("anotherstring"), 100);
+
+    private final AtomicLong timestamp = new AtomicLong();
+    private final KeyValueService kvs = new InMemoryKeyValueService(true);
+    private final CoordinationStore coordinationStore = spy(KeyValueServiceCoordinationStore.create(
```
sandorw: Why not just use a mock based on the interface?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
jeremyk-91: (offline discussion: I'll attempt to do this. This still needs to be a function, but it will probably be enough to have this be a transform on just the value, rather than including the bounds)

jeremyk-91:
I will rebase on Part 1 soon, as that's close to merge.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
jeremyk-91: Yep, that's exactly the idea I was going for. Changes are applied in a way that any values for behaviour at timestamps less than the previous validity bound are preserved.

Within the context of _transactions2, the idea is that you have a `RangeMap` of timestamps to transaction schema you should be using. This is initially `(0, infinity) => 1`, but once installed at some validity `V` (where the previous validity was `V'`) this looks like `(0, V'] => 1, (V', infinity) => 2`

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
jeremyk-91: This depends on the feature. If I'm not wrong, if we want to use this for some feature where the start or commit timestamp alone decides which behaviour we want (_transactions2 is decided based on start timestamp) then this works as long as we're consistent. I haven't thought about ensuring consistency across both, but presumably a feature needing that could check at the end after getting the commit timestamp and throw if the value has changed then.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
sandorw: In practice, though, any client should treat it as though the bound is `(0, V'] => 1, (V', bound) => 2`, right? It would obviously be dangerous to assume any behavior extends to infinity

jeremyk-91:
(edit: I will rebase on develop. There have been substantial changes in that the underlying Store API has changed.)

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationService.java
jeremyk-91: Right, though generally I think we assume `bound` itself is good - so the part between `[bound+1, infinity)` should not be relied on to stay the same.

jeremyk-91: The main reason I use infinity is that if this turns out to be a part of some Atlas internal schema (so we have something like `{"transactions": "(0, inf) => 1", "otherThing": 1}`) and we want to update the `otherThing` key, we don't have to think about logic for transactions directly.

Issue #3612, opened 2018-10-25
We cannot retry tasks that mutate cassandra schema on other clients


gmaretic:
e.g. if `runSchemaMutationTask(() -> schemaMutationLock.runWithLock(() -> createTablesInternal(tablesToActuallyCreate)));` fails in waiting for schema versions, we will retry potentially on a different node and create the table twice

PR #3620, opened 2018-10-30
Use CQL to create tables

gmaretic:
**Goals (and why)**:
Fix https://github.com/palantir/atlasdb/issues/3612 (also see internal ticket 76619).

Whenever a table is to be created, a cfId is deterministically generated for it from the keyspace and the table, and used in the CQL query. This ensures that even if multiple nodes simultaneously try to create the table on different Cassandra nodes, we do not end up in an inconsistent state.
When a table is to be dropped, we first truncate the table, so that if it gets recreated later, entries do not come back to life.

Existing tests, also did manual testing in Cassandra.

We need all nodes to be up to drop tables, as we need to be able to truncate. Is the table creation correct (I did manual testing to check metadata in C* and looked consistent to me, but doesn't hurt to take another look)?

CassandraTableCreator

ASAP

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3620)


gmaretic:
To elaborate on the race condition, the least complicated one I was able to come up with is:
1. T1 calls createTable(X)
2. T2 calls createTable(X)
3. T1 gets UUID1
4. T2 gets UUID1
5. T1 creates table X in Cassandra
6. T3 calls dropTable(X)
7. T3 invalidates UUID for table X
8. T3 drops table X
9. T4 calls createTable(X) (alternatively this happens in step 3. but the thread blocks before getting UUID)
10. T4 gets UUID2
Now T2 and T4 try to create table X with different UUIDs, if they happen to talk to different Cassandra nodes before they can communicate the table creation to each other, schemas diverge

The simple way to avoid this is to use a deterministic UUID based on keyspace and table name, but then we have to truncate the table before it gets dropped. This means we cannot drop tables unless we have all Cassandra nodes up.

Another way to fix it is to keep the schema mutation lock and do these mutations while holding the lock (we have to be more careful as the old implementation had more bugs than just the bad retrying). However, we then still need to get schema agreement on all nodes before we can release the schema mutation lock and otherwise leak it on purpose to avoid a bad race condition

gmaretic:
Update: decided to go with option 2 from the comment above.
Table creation uses a deterministically generated UUID, table deletion truncates before drop.

gmaretic:
@tpetracca @amandachow can we get someone from cassops to review this as well?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,137 @@
...
+
+        settings.forEach((option, value) -> queryBuilder.append("AND " + option + " = " + value + " "));
+
+        queryBuilder.append("AND CLUSTERING ORDER BY (" + COLUMN + " ASC, " + TIMESTAMP + " ASC) ");
+
+        return CqlQuery.builder()
+                .safeQueryFormat(queryBuilder.toString())
+                .addArgs(
+                        SafeArg.of("keyspace", config.getKeyspaceOrThrow()),
+                        SafeArg.of("internalTableName", CassandraKeyValueServiceImpl.internalTableName(tableRef)),
```
jeremyk-91: note: We cannot guarantee this is safe (it is an arbitrary table name after all)

But given internal logging practices I'm tempted to say we should consider allowing TRs within the context of Cassandra to be assumed safe - let's discuss tomorrow when the team is in.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,137 @@
...
+        settings.put("id", "'%s'");
+        settings.put("populate_io_cache_on_flush", cachePriorityIsHottest(tableMetadata));
+        settings.put("speculative_retry", "'NONE'");
+
+        queryBuilder.append("CREATE TABLE IF NOT EXISTS \"%s\".\"%s\""
+                + " ( " + ROW + " blob, " + COLUMN + " blob, " + TIMESTAMP + " bigint, " + VALUE + " blob, "
+                + "PRIMARY KEY (" + ROW + ", " + COLUMN + ", " + TIMESTAMP + ")) "
+                + "WITH COMPACT STORAGE ");
+
+        settings.forEach((option, value) -> queryBuilder.append("AND " + option + " = " + value + " "));
```
jeremyk-91: Eventually we should have a separate object that builds up these options, taking care of the valid keys as well as the formatting of the values to be in single quotes / be legitimate objects. I realise this is high priority, so not blocking.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableTruncator.java
```diff
@@ -0,0 +1,75 @@
...
+                        client.truncate(CassandraKeyValueServiceImpl.internalTableName(tableRef));
+                        return true;
+                    });
+                }
+                return null;
+            }
+
+            @Override
+            public String toString() {
+                return "truncateTables(" + tablesToTruncate.size() + " tables)";
```
jeremyk-91: nit: if (but only if) we take tables within C* as safe, then this might as well say what the table names are

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,13 @@ develop
     *    - Type
          - Change
 
+    *    - |fixed|
+         - Cassandra KVS no longer uses the schema mutation lock and instead creates tables using an id deterministically generated from the Cassandra keyspace and the table name.
+           As part of this change, table deletion now truncated the table before dropping it in Cassandra, therefore requiring all Cassandra nodes to be available to drop tables.
```
jeremyk-91: nit: ...now truncates the table...

atlasdb-cassandra-multinode-tests/src/test/java/com/palantir/cassandra/multinode/OneNodeDownTableManipulationTest.java
```diff
@@ -57,21 +57,15 @@ public void canCreateTables() {
...
     @Test
-    public void canDropTables() {
-        getTestKvs().dropTables(ImmutableSet.of(TABLE_TO_DROP_2));
-
-        assertThat(getTestKvs().getAllTableNames()).doesNotContain(TABLE_TO_DROP_2);
-        assertKvsReturnsEmptyMetadata(TABLE_TO_DROP_2);
-        assertCassandraSchemaChanged();
+    public void cannotDropTables() {
+        assertThrowsAtlasDbDependencyExceptionAndDoesNotChangeCassandraSchema(() ->
+                getTestKvs().truncateTable(TABLE_TO_DROP_2));
```
jeremyk-91: Is this intentional? It seems like these tests should still call drop directly - especially since there are tables to be truncated in other tests.

atlasdb-cassandra-multinode-tests/src/test/java/com/palantir/cassandra/multinode/TwoNodesDownTableManipulationTest.java
```diff
@@ -31,35 +29,10 @@ void testSetup(CassandraKeyValueService kvs) {
...
-        assertThrowsAtlasDbDependencyExceptionAndDoesNotChangeCassandraSchema(() ->
-                getTestKvs().truncateTable(TEST_TABLE));
-
-    }
-
-    @Test
-    public void truncateTablesThrowsAndDoesNotChangeCassandraSchema() {
-        assertThrowsAtlasDbDependencyExceptionAndDoesNotChangeCassandraSchema(() ->
-                getTestKvs().truncateTables(ImmutableSet.of(TEST_TABLE)));
-    }
```
jeremyk-91: I guess the rationale for removing these tests here is "if we can't do it with 1 node down, we almost certainly can't do it with 2 down"?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -384,4 +385,12 @@ public static boolean isEmptyOrInvalidMetadata(byte[] metadata) {
         return false;
     }
 
+    public static TableMetadata getMetadataOrDefaultToGeneric(byte[] metadata) {
+        if (metadata == null || Arrays.equals(metadata, AtlasDbConstants.EMPTY_TABLE_METADATA)) {
+            return TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(AtlasDbConstants.GENERIC_TABLE_METADATA);
+        } else {
+            return TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(metadata);
+        }
```
jeremyk-91: nit: `else` is unnecessary

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,137 @@
...
+        queryBuilder.append("AND CLUSTERING ORDER BY (" + COLUMN + " ASC, " + TIMESTAMP + " ASC) ");
+
+        return CqlQuery.builder()
+                .safeQueryFormat(queryBuilder.toString())
+                .addArgs(
+                        SafeArg.of("keyspace", config.getKeyspaceOrThrow()),
+                        SafeArg.of("internalTableName", CassandraKeyValueServiceImpl.internalTableName(tableRef)),
+                        SafeArg.of("cfId", getUuidForTable(tableRef)))
+                .build();
+    }
```
jeremyk-91: I defer to Cassops - but base Atlas explicitly sets fields to default values in `ColumnFamilyDefinitions` which we aren't setting here, like `dclocal_read_repair_chance`, `min_index_interval` etc. Is it safe now to not explicitly set these?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,137 @@
...
+    private String getCompression(int explicitCompressionBlockSizeKb) {
+        int chunkLength = explicitCompressionBlockSizeKb != 0 ? explicitCompressionBlockSizeKb
+                : AtlasDbConstants.MINIMUM_COMPRESSION_BLOCK_SIZE_KB;
+        return "{'chunk_length_kb': '" + chunkLength + "', "
+                + "'sstable_compression': '" + CassandraConstants.DEFAULT_COMPRESSION_TYPE + "'}";
+    }
+
+    private String cachePriorityIsHottest(TableMetadata tableMetadata) {
+        return Boolean.toString(tableMetadata.getCachePriority()
+                .equals(TableMetadataPersistence.CachePriority.HOTTEST));
```
jeremyk-91: nit: `CachePriority` is an enum, so I'd use `==` rather than `.equals()`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableDropper.java
```diff
@@ -39,68 +38,68 @@
...
+                                    .entrySet(),
                             ts));
         } catch (Exception e) {
             throw Throwables.unwrapAndThrowAtlasDbDependencyException(e);
         }
 
         try {
-            new CellRangeDeleter(clientPool, wrappingQueryRunner, deleteConsistency, no -> System.currentTimeMillis())
+            new CellRangeDeleter(clientPool, wrappingQueryRunner, CassandraKeyValueServiceImpl.DELETE_CONSISTENCY,
+                    no -> System.currentTimeMillis())
```
jeremyk-91: 😄

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableDropper.java
```diff
@@ -39,68 +38,68 @@
...
+                    CassandraVerifier.sanityCheckTableName(table);
+                    if (existingTables.contains(table)) {
+                        cassandraTableTruncator.runTruncateOnClient(ImmutableSet.of(table), client);
+                        client.system_drop_column_family(CassandraKeyValueServiceImpl.internalTableName(table));
+                        updateMetadataForTable(table);
+                    } else {
+                        log.warn("Ignored call to drop a table ({}) that did not exist.", LoggingArgs.tableRef(table));
+                    }
+                }
+                CassandraKeyValueServices.waitForSchemaVersions(config, client, "after dropping the column "
```
tpetracca: I wonder if we should be wrapping every drop/create call with this (both sides of it and at the individual resource level). So something more like:
```
for (TableReference table: tablesToDrop) {
    // wait for schema versions
    // drop table
    // wait for schema versions
}
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableTruncator.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private FunctionCheckedException<CassandraClient, Void, TException> truncationFunction(
+            Collection<TableReference> tablesToTruncate) {
+        return new FunctionCheckedException<CassandraClient, Void, TException>() {
+            @Override
+            public Void apply(CassandraClient client) throws TException {
+                for (TableReference tableRef : tablesToTruncate) {
+                    queryRunner.run(client, tableRef, () -> {
+                        client.truncate(CassandraKeyValueServiceImpl.internalTableName(tableRef));
```
tpetracca: might be nice if we had some error handling here to log useful messages, like that this is failing (or maybe we don't even attempt to run it) if all nodes are not up

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
```diff
@@ -342,17 +342,4 @@ private static void checkRfsMatchConfig(KsDef ks, CassandraKeyValueServiceConfig
             }
         }
     }
-
-    static final FunctionCheckedException<CassandraClient, Boolean, UnsupportedOperationException>
```
tpetracca: can we actually delete this? aren't there other tables we CAS?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
tpetracca: The datastax driver does all of this for you! And has built in schema version agreement checking!
https://docs.datastax.com/en/latest-java-driver-api/com/datastax/driver/core/schemabuilder/SchemaBuilder.html

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,137 @@
...
+        this.config = config;
+    }
+
+    void createTables(Map<TableReference, byte[]> tableRefToMetadata) {
+        try {
+            clientPool.runWithRetry(client -> {
+                for (Map.Entry<TableReference, byte[]> entry : tableRefToMetadata.entrySet()) {
+                    createTable(entry.getKey(), entry.getValue(), client);
+                }
+                CassandraKeyValueServices.waitForSchemaVersions(config, client, "after adding the column family for "
```
tpetracca: mentioned in the delete section as well; I think we should be wrapping all calls:
```
for (table in tables) {
    // wait for schema versions
    // create table (or do any schema thing)
    // wait for schema versions
}
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
tpetracca: I get actually using the driver to do the call (and thus getting the free schema version checking) might be hard, but might at least be able to use the SchemaBuilder class to generate the `CREATE TABLE` statement here more easily.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
tpetracca: +1 to assume safe

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableDropper.java
```diff
@@ -39,68 +38,68 @@
...
-                        CassandraKeyValueServices.waitForSchemaVersions(config, client, "after dropping the column "
-                                + "family for tables " + tablesToDrop + " in a call to drop tables");
-                        return null;
-                    });
+                for (TableReference table : tablesToDrop) {
+                    CassandraVerifier.sanityCheckTableName(table);
+                    if (existingTables.contains(table)) {
+                        cassandraTableTruncator.runTruncateOnClient(ImmutableSet.of(table), client);
+                        client.system_drop_column_family(CassandraKeyValueServiceImpl.internalTableName(table));
+                        updateMetadataForTable(table);
```
sandorw: What is this metadata used for? Is it correct to update the metadata after truncating and deleting the table?

gmaretic: Bad naming, renamed to make clear

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableTruncator.java
gmaretic: I have this as a follow-up task. In general when we retry 6 times we end up with several different causes so it's not always easy to figure out the "root cause"

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableTruncator.java
gmaretic: 👍

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
gmaretic: We do use CAS elsewhere, but we don't actually check that it is supported and CassandraKvsImpl.supportsCheckAndSet() returns true regardless of what happens during intitialisation. We should then either wire it up correctly or just remove it since no one is afaik on a version of Cassandra that doesn't support it.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableDropper.java
gmaretic: To elaborate, we don't use this metadata to determine if a table exists for future deletions and creations, so having some potentially stale metadata is fine.

atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/AtlasCli.java
```diff
@@ -45,7 +44,6 @@ private AtlasCli() {}
                 .withCommand(Help.class)
                 .withCommand(SweepCommand.class)
                 .withCommand(KvsMigrationCommand.class)
-                .withCommand(CleanCassLocksStateCommand.class)
```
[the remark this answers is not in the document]
gmaretic: Let me know if you have details :)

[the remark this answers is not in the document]
sandorw: Unfortunately, probably can't remove the code for cleaning up the lock state until we know that everyone is on a newer version of Atlas such that it can no longer occur.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,134 @@
...
+                .compressionOptions(getCompression(tableMetadata.getExplicitCompressionBlockSizeKB()))
+                .dcLocalReadRepairChance(0.1)
+                .gcGraceSeconds(config.gcGraceSeconds())
+                .minIndexInterval(128)
+                .maxIndexInterval(2048)
+                .populateIOCacheOnFlush(tableMetadata.getCachePriority() == CachePriority.HOTTEST)
+                .speculativeRetry(SchemaBuilder.noSpeculativeRetry())
+                .clusteringOrder(COLUMN, SchemaBuilder.Direction.ASC)
+                .clusteringOrder(TIMESTAMP, SchemaBuilder.Direction.ASC)
+                .buildInternal();
```
gmaretic: Also the following things are explicitly set to defaults in the thrift implementation that I was unable to set here. Is this something that we must do @tpetracca ?
cf.setTriggers(ImmutableList.of());
cf.setCells_per_row_to_cache("0");
cf.setColumn_metadata(ImmutableList.of());
cf.setKey_validation_class("org.apache.cassandra.db.marshal.BytesType");
cf.setDefault_validation_class("org.apache.cassandra.db.marshal.BytesType");

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
tpetracca: yea it was mostly a concern back when we had PG impls using cassandra 1.2, but my guess is it's fine to remove at this point

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -61,71 +64,65 @@ void createTables(Map<TableReference, byte[]> tableRefToMetadata) {
...
+        String query = SchemaBuilder.createTable(keyspace, internalTableName)
+                .ifNotExists()
+                .addPartitionKey(ROW, DataType.blob())
+                .addClusteringColumn(COLUMN, DataType.blob())
+                .addClusteringColumn(TIMESTAMP, DataType.bigint())
+                .addColumn(VALUE, DataType.blob())
+                .withOptions()
+                .bloomFilterFPChance(falsePositive(tableMetadata.hasNegativeLookups(), appendHeavyReadLight))
+                .caching(SchemaBuilder.Caching.KEYS_ONLY)
+                .comment("")
```
tpetracca: I think we can delete this

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
tpetracca: You should be all good:
- triggers are their own kind of statement `CREATE TRIGGER` so there's no concept of it in a createTable statement in CQL
- we're not caching rows anyway
- the bottom 3 are all covered in a different way by the top section here (the row, column1, column2, value names and their types)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
```diff
@@ -0,0 +1,134 @@
...
+        if (appendHeavyAndReadLight) {
+            return negativeLookups ? CassandraConstants.NEGATIVE_LOOKUPS_SIZE_TIERED_BLOOM_FILTER_FP_CHANCE
+                    : CassandraConstants.DEFAULT_SIZE_TIERED_COMPACTION_BLOOM_FILTER_FP_CHANCE;
+        }
+        return negativeLookups ? CassandraConstants.NEGATIVE_LOOKUPS_BLOOM_FILTER_FP_CHANCE
+                : CassandraConstants.DEFAULT_LEVELED_COMPACTION_BLOOM_FILTER_FP_CHANCE;
+    }
+
+    private CompactionOptions<?> getCompaction(boolean appendHeavyReadLight) {
+        return appendHeavyReadLight ? SchemaBuilder.sizedTieredStategy().minThreshold(4).maxThreshold(32)
```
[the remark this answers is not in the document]
tpetracca: We deleted the 80MB thing a while ago (this comment should have been deleted then). We just use the cassandra default for leveled which is 160MB.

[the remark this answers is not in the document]
tpetracca: so you're all good here.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
gmaretic: I included it for completeness sake, but I guess it is slightly overkill :)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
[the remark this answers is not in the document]
gmaretic: 👍

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraTableCreator.java
gmaretic: Thanks for confirming, I assumed that was the case since the comparator worked that way

PR #3624, opened 2018-11-01
Make TableMetadata use Immutables

gmaretic:
**Goals (and why)**:
Avoid using a constructor with 10 arguments to specify sweep strategy, avoid having to correctly maintain defaults in several classes, make it easier to write and understand table metadata.

Straightforward

There were some existing tests that no longer make much sense, there are incidental tests in a few places, should we have more?

This will be a devbreak for large internal product, but should be easy to fix.
I specifically did not touch the defaults in TableDefinition even though it probably makes sense to also have those use default values from TableMetadata for all properties that are not set. Should we do it?

TableMetadata

Whenever

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3624)


gmaretic:
@jeremyk-91 Ping on this

atlasdb-client/src/test/java/com/palantir/atlasdb/logging/SafeLoggableDataUtilsTest.java
```diff
@@ -54,21 +53,22 @@
...
+
+    private static final TableMetadata TABLE_METADATA_2 = TableMetadata.builder()
+            .rowMetadata(NAME_METADATA_DESCRIPTION)
+            .columns(COLUMN_METADATA_DESCRIPTION)
+            .nameLogSafety(LogSafety.SAFE)
+            .build();
+    private static final TableMetadata TABLE_METADATA_DYNAMIC_COLUMNS = TableMetadata.builder()
+            .rowMetadata(NAME_METADATA_DESCRIPTION)
+            .nameLogSafety(LogSafety.SAFE)
+            .build();
```
jeremyk-91: 🎉 much better than the original

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/SimpleKvsTimestampBoundStore.java
```diff
@@ -1,106 +0,0 @@
...
-import com.palantir.atlasdb.table.description.NameComponentDescription;
-import com.palantir.atlasdb.table.description.NameMetadataDescription;
-import com.palantir.atlasdb.table.description.NamedColumnDescription;
-import com.palantir.atlasdb.table.description.TableMetadata;
-import com.palantir.atlasdb.table.description.ValueType;
-import com.palantir.atlasdb.transaction.api.ConflictHandler;
-import com.palantir.timestamp.MultipleRunningTimestampServiceError;
-import com.palantir.timestamp.TimestampBoundStore;
-
-public class SimpleKvsTimestampBoundStore implements TimestampBoundStore {
```
jeremyk-91: Checked internally, don't think this is used anywhere. (I'm surprised this even exists)

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -27,145 +27,69 @@
...
+    public boolean appendHeavyAndReadLight() {
+        return false;
     }
 
-    public ConflictHandler getConflictHandler() {
-        return conflictHandler;
+    @Value.Default
+    public LogSafety nameLogSafety() {
+        return LogSafety.UNSAFE;
     }
```
jeremyk-91: Debatable, and certainly non-blocking: We could probably have gotten away with leaving the method names as they were, and using `@Value.Style(get = {"get*", "is*"})`

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -27,145 +27,69 @@
...
 
-    public ColumnMetadataDescription getColumns() {
-        return columns;
+    @Value.Default
+    public SweepStrategy sweepStrategy() {
+        return SweepStrategy.CONSERVATIVE;
     }
 
-    public boolean isWriteWriteConflict() {
-        return conflictHandler != ConflictHandler.IGNORE_ALL;
```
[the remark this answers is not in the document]
gmaretic: It's not used anywhere in our codebase or in large internal product, and it is just brittle logic

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
gmaretic: As much as I dislike the style, minimizing the devbreak makes sense 👍

gmaretic:
@jeremyk-91 I merged with develop, so make sure to double check that the metadata in KeyValueServiceCoordinationStore is correct

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -27,145 +27,69 @@
...
-    final ConflictHandler conflictHandler;
-    final CachePriority cachePriority;
-    final boolean rangeScanAllowed;
-    final int explicitCompressionBlockSizeKB;
-    final boolean negativeLookups;
-    final SweepStrategy sweepStrategy;
-    final boolean appendHeavyAndReadLight;
-    final LogSafety nameLogSafety;
+@Value.Immutable
+public abstract class TableMetadata implements Persistable {
```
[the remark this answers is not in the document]
gmaretic: 👍

gmaretic:
This has been smoke tested by verifying that the persisted versions of generic table metadata, kvs coordination store, and kvs punch store are identical to the old versions.

At this point I don't see a compelling reason not to merge it in

gmaretic:
@felixdesouza @jeremyk-91 (whoever has time to take a look)

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ColumnMetadataDescription.java
```diff
@@ -48,6 +50,25 @@ public ColumnMetadataDescription(DynamicColumnDescription dynamicColumn) {
         this.dynamicColumn = dynamicColumn;
     }
 
+    public static ColumnMetadataDescription singleNamed(String shortName, String longName, ValueType valueType) {
```
felixdesouza: nit, probably later if you're feeling bothered, it may be nice to have these on the builder? 

so you have something on the builder so you can do this:
```java
TableMetadata.internal()
  .namedColumn("shortName", "longName", ValueType.BLOB)
  .dynamicColumn("name", ValueType.BLOB, ValueType.SHA256)
  // etc
```

```java
class Builder extends ImmutableBuilder {
  public Builder namedColumn(String name, String longName, ValueType valueType) {
    // contents on the method here
  }
}
```

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -185,149 +110,64 @@ public LogSafety getNameLogSafety() {
...
-                explicitCompressionBlockSizeKB,
-                negativeLookups,
-                sweepStrategy,
-                appendHeavyAndReadLight,
-                nameLogSafety);
+        return builder.build();
     }
 
     @Override
     public String toString() {
```
felixdesouza: you should be able to delete this?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServiceScrubberStore.java
```diff
@@ -94,25 +92,15 @@ private KeyValueServiceScrubberStore(KeyValueService keyValueService) {
...
-                                new NameComponentDescription.Builder()
-                                        .componentName("col")
-                                        .type(ValueType.BLOB)
-                                        .build())),
-                        ColumnValueDescription.forType(ValueType.VAR_LONG))),
-                ConflictHandler.IGNORE_ALL,
-                TableMetadataPersistence.LogSafety.SAFE);
+        TableMetadata scrubTableMeta = TableMetadata.internal()
+                .rowMetadata(NameMetadataDescription.create("row", ValueType.BLOB))
+                .columns(new ColumnMetadataDescription(new DynamicColumnDescription(
```
felixdesouza: don't you want to use the utility methods you've defined?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -33,31 +33,22 @@
...
-                    new NameComponentDescription.Builder().componentName("table_name").type(ValueType.STRING).build())),
-            new ColumnMetadataDescription(ImmutableList.of(
-                    new NamedColumnDescription(
-                            AtlasDbConstants.NAMESPACE_SHORT_COLUMN_NAME,
-                            "short_name",
-                            ColumnValueDescription.forType(ValueType.STRING)))),
-            ConflictHandler.IGNORE_ALL,
-            TableMetadataPersistence.LogSafety.SAFE);
+    public static final TableMetadata NAMESPACE_TABLE_METADATA = TableMetadata.internal()
+            .rowMetadata(NameMetadataDescription.create(ImmutableList.of(
```
felixdesouza: nit: reuse static methods/builder here

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
gmaretic: I think the generated one is slightly different, but we should be able to live with that userbreak

PR #3627, opened 2018-11-02
Fix deprecated OKHttp builder call

mrowbotham:
**Goals (and why)**:

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.java#L81 is thrown on startup with Java9+, for things which use timelock. 

https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.Builder.html#sslSocketFactory-javax.net.ssl.SSLSocketFactory-

Goal is to remove the call to deprecated `public OkHttpClient.Builder sslSocketFactory(SSLSocketFactory sslSocketFactory)` 

Changed Optional<SSLSocketFactory> to Optional<TrustContext>.

Ran all existing tests.

Is there any extra risk from also passing a X509TrustManager to the OKHttpClient.Builder ?
Is there any worry about changing the Optional<SSLSocketFactory> to Optional<TrustContext> ?
Would you also like a refactor to remove Optionals as method params?

Wherever you like.

Would like to get this deployed within a week.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3627)


mrowbotham:
A little help here with failing tests would be cool too, seemed to work on initial PR creation then break after merging in develop. Could be related to something i've done. Can't make it fail locally.

PR #3644, opened 2018-11-19
Clean up TimeLockClient

zakmagnus:
This is a bug - the actual timestamp-and-lock close method was not getting registered as a close callback of its transaction manager. That close method closes `TimeLockClient`. Failing to do this causes thread leaks, especially in test scenarios where many transaction managers can be created and destroyed. I'm surprised the type checker allowed this bug to exist.

This week, please


sandorw:
@j-baker as a follow-up, should we instead be injecting an executor and shutting it down? Right now, closing the services using the autobatcher can still lead a thread running autobatcher work.

PR #3648, opened 2018-11-21
[PDS-75716] Finer grained DbKVS connection manager profiling

jeremyk-91:
**Goals (and why)**: 
See PDS-75716; this aims to help some investigation concerning slow execution of queries.

- Time creating a statement and executing it separately.

- admittedly nonexistent 😞 

- Can creating the statement ever be slow? My guess would be yes based on reading the JavaDoc (it can throw a SQLException "if a database access error occurs"), but not sure.

**Priority (whenever / two weeks / yesterday)**: P2; this week/early next.

### DIFF
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
@@ -368,7 +368,7 @@ public class HikariCPConnectionManager extends BaseConnectionManager {
     private class ConnectionAcquisitionProfiler {
         private static final String PROFILING_MESSAGE_FORMAT = "[{}] Waited {}ms for connection"
                 + " (acquisition: {}, statement creation: {}, execution: {})";
-        
+
         private final Stopwatch globalStopwatch;
         private final Stopwatch trialStopwatch;
 
@@ -413,7 +413,7 @@ public class HikariCPConnectionManager extends BaseConnectionManager {
                         queryExecutionMillis);
                 logPoolStats();
             } else {
-                log.warn(PROFILING_MESSAGE_FORMAT,
+                log.debug(PROFILING_MESSAGE_FORMAT,
                         connConfig.getConnectionPoolName(),
                         elapsedMillis,
                         acquisitionMillis,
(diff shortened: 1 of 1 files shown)


jeremyk-91:
Hmm, this is failing on `Failed to download "https://artifactory.palantir.build/artifactory/internal-dist-release/com/palantir/assetserver/asset-server/1.0.2/asset-server-1.0.2.sls.tgz!asset-server-1.0.2/deployment/configuration.yml" for "asset-server": Failed to get "https://artifactory.palantir.build/artifactory/internal-dist-release/com/palantir/assetserver/asset-server/1.0.2/asset-server-1.0.2.sls.tgz!asset-server-1.0.2/deployment/configuration.yml" --> "/tmp/130879099/asset-server.configuration.yml": Failed to get "https://artifactory.palantir.build/artifactory/internal-dist-release/com/palantir/assetserver/asset-server/1.0.2/asset-server-1.0.2.sls.tgz!asset-server-1.0.2/deployment/configuration.yml": bad response code: 404` which I suspect is blacklisting. I'll have a look later today.

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
```diff
@@ -211,8 +211,13 @@ private void testDataSource(HikariDataSource dataSourcePool) throws SQLException
      * @throws SQLException if connection is invalid and cannot be used.
      */
     private void testConnection(Connection connection) throws SQLException {
-        try (Statement stmt = connection.createStatement();
-                ResultSet rs = stmt.executeQuery(connConfig.getTestQuery())) {
+        testConnection(connection, Optional.empty());
```
dxiao: confused - what's the point of splitting out profiling into two steps if the profiler is never actually called because you're passing in Optional.empty()? seems like you need to pipe this one level higher.

dxiao:
Otherwise, this seems like it will present reasonable logging

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
jeremyk-91: Testing the connection is done in two places - `acquirePooledConnection` and `testDataSource`, but we only have been doing the profiling in the first case. In the second case we don't have a profiler object, which is why we need a version that takes `Optional`. 

That said, I forgot to update the call inside `acquirePooledConnection` to use the profiler, which is now done - good catch there.

jeremyk-91:
The original ticket was deprioritised!

PR #3649, opened 2018-11-21
Fix getMetadataForTables() and add tests documenting odd Cassandra behaviour

gmaretic:
**Goals (and why)**:
This addresses strangeness with the above mentioned KVS method on Cassandra. In particular, don't show metadata for tables that were dropped, don't show metadata for tables that don't exist in cassandra, and use the capitalization that is used in Cassandra. For more details, see https://github.com/palantir/atlasdb/pull/3650 and https://github.com/palantir/atlasdb/pull/3656 .

This is technically a userbreak, so marked it as such.


atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTestUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static void insertGenericMetadataIntoLegacyCell(KeyValueService kvs, TableReference tableRef, byte[] data) {
+        Cell legacyMetadataCell = CassandraKeyValueServices.getOldMetadataCell(tableRef);
+        kvs.put(AtlasDbConstants.DEFAULT_METADATA_TABLE, ImmutableMap.of(legacyMetadataCell, data),
+                System.currentTimeMillis());
+    }
+
+    // notably, this metadata is different from the default AtlasDbConstants.GENERIC_TABLE_METADATA
+    // to make sure the tests are actually exercising the correct retrieval codepaths
+    public static byte[] originalMetadata() {
```
jeremyk-91: OK - seems like this should be a constant, but this is fine

docs/source/release_notes/release-notes.rst
```diff
@@ -50,6 +50,12 @@ develop
     *    - Type
          - Change
 
+    *    - |fixed| |userbreak|
+         - Cassandra KVS `getMetadataForTables` method now does not contain entries for tables that do not exist in Cassandra.
+           Previously, when a table was dropped, an empty byte array would be written into the _metadata table to mark it as deleted.
+           Furthermore, the table reference keys of the returned map have capitalisation matching the table names in Cassandra.
```
jeremyk-91: So one possible concern about this change is that now when we drop a table we need to call `deleteAllMetadataRows`, which triggers a range scan on _metadata across the ASCII range from its upper to lower versions. Should we call this out here, along with 'please contact the AtlasDB team if this is an issue' or something like that?

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableManipulationIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+        assertThat(kvs.getMetadataForTables()).isEmpty();
+        assertThat(kvs.getMetadataForTable(LOWER_UPPER)).isEmpty();
+    }
+
+    @Test
+    public void nonExistentTablesWithMetadataDoNotAppearInGetMetadataForTables() {
+        insertMetadataIntoNewCell(LOWER_UPPER);
+
+        assertThat(kvs.getMetadataForTables()).isEmpty();
+        assertThat(kvs.getMetadataForTable(LOWER_UPPER)).contains(AtlasDbConstants.GENERIC_TABLE_METADATA);
```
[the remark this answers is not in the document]
gmaretic: Fair

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -316,30 +320,53 @@ private static Column createColumnAtSpecificCassandraTimestamp(Cell cell, Value
...
     }
 
     @SuppressWarnings("checkstyle:RegexpSinglelineJava")
-    static TableReference tableReferenceFromBytes(byte[] name) {
-        return TableReference.createUnsafe(new String(name, Charset.defaultCharset()));
+    static byte[] upperCaseTableReferenceToBytes(TableReference tableRef) {
+        return tableRef.getQualifiedName().toUpperCase().getBytes(Charset.defaultCharset());
+    }
+
+    @SuppressWarnings("checkstyle:RegexpSinglelineJava")
```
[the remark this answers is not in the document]
gmaretic: I guess we just need to suppress for the usage of Charset.defaultCharset(), but I just went with what the other methods do

docs/source/release_notes/release-notes.rst
gmaretic: Makes sense. I don't think there should be cases where this comes back to bite us, but makes sense to call it out.

Issue #3625, opened 2018-11-01
CKVS can return metadata for deleted tables


clockfort:
CKVS (and maybe other KVSs, I didn't check) can return previously deleted tables in a call to KVS#getMetadataForTables() (the one that gets all of the metadata with no filtering) 

This is because CKVS (and I think all of the other KVSs) record table deletion with a table metadata insert of (tablename, fakeCol, current wallclock, byte[0]), and it happens to not filter out non-live tables whose most recent cell's value is a byte[0] in this particular codepath.
There is also some oddness around tables with missing metadata that are also defaulted to byte[0] when returning.

https://github.com/palantir/atlasdb/blob/abf54b5c158bdf1794f1c545c0374cf2053898e4/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java#L1602-L1646

@gmaretic do you want to look into this as part of your table metadata fixes, or should I put something in (after your thing merges)?

gmaretic:
Do we want to just put a ranged tombstone across the entire Long value range in CassandraTableDropper for both the row corresponding to table name and the one corresponding to its canonical form? Putting the empty byte array is then redundant unless we specifically lose a node after dropping the table and before updating metadata.

I want to look into some other weirdness around metadata, my favourite being Cassandra being case sensitive while AtlasDB trying to enforce case insensitivity -- try to call CKVS.createTables({AA, aa}) for fun results...

Anyway, I'm OOTO for almost two weeks staring today, so feel free to fix things :)

clockfort:
The ranged tombstone sounds good to me.

To be clear, this isn't a sweep-managed table, so currently we leak all the historical table schema changes ad infinitum, but I've actually been (ab)using that to do some debug work for internal ticket PDS-76535, but such debugging workflows should hopefully never be necessary in a future where table metadata stuff isn't a hot mess.

gmaretic:
Fixed in https://github.com/palantir/atlasdb/pull/3649

PR #3650, opened 2018-11-22
Delete metadata properly when dropping tables

gmaretic:
**Goals (and why)**:
Fix https://github.com/palantir/atlasdb/issues/3625

Instead of writing an empty byte array to the metadata cell, just nuke both the current and the legacy metadata cells using a ranged timestamp.

Diff shows that the behaviour is now as expected, added a test showing that it works for tables that have metadata in the legacy cell as well.

Technically introduces a race condition if we lose a node after truncating the table and before deleting the metadata where before we would have placed the empty byte array before erroring out.

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraStrangeBehaviourTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraStrangeBehaviourTest.java
@@ -82,11 +82,20 @@ public class CassandraStrangeBehaviourTest {
     }
 
     @Test
-    public void droppedTableMetadataShowsUpInGetMetadataForTables() {
-        kvs.createTable(LOWER_LOWER, AtlasDbConstants.GENERIC_TABLE_METADATA);
-        kvs.dropTable(LOWER_LOWER);
+    public void droppedTableMetadataDoesNotShowUpInGetMetadataForTables() {
+        kvs.createTable(LOWER_UPPER, AtlasDbConstants.GENERIC_TABLE_METADATA);
+        kvs.dropTable(LOWER_UPPER);
 
-        assertThat(kvs.getMetadataForTables().keySet()).contains(LOWER_LOWER);
+        assertThat(kvs.getMetadataForTables()).isEmpty();
+    }
+
+    @Test
+    public void droppedTableMetadataDoesNotShowUpInGetMetadataForLegacyTables() {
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraStrangeBehaviourTest.java
```diff
@@ -121,4 +130,21 @@ private void createTablesIgnoringException() {
...
+                true);
+
+        Cell legacyMetadataCell = CassandraKeyValueServices.getOldMetadataCell(tableRef);
+        kvs.put(AtlasDbConstants.DEFAULT_METADATA_TABLE,
+                ImmutableMap.of(legacyMetadataCell, AtlasDbConstants.GENERIC_TABLE_METADATA),
+                System.currentTimeMillis());
+
+        assertThat(kvs.getMetadataForTable(LOWER_UPPER)).isNotEmpty();
+        assertThat(kvs.getMetadataForTables()).isNotEmpty();
+    }
```
jeremyk-91: There is one sad thing we can do

```java
    @Test
    public void legacyCellWithDifferentCapitalisationBreaksTheWorld() {
        Cell legacyMetadataCell = CassandraKeyValueServices.getOldMetadataCell(LOWER_UPPER);
        kvs.put(AtlasDbConstants.DEFAULT_METADATA_TABLE,
                ImmutableMap.of(legacyMetadataCell, AtlasDbConstants.GENERIC_TABLE_METADATA),
                System.currentTimeMillis());
        kvs.createTable(LOWER_LOWER, AtlasDbConstants.GENERIC_TABLE_METADATA);
        kvs.dropTable(LOWER_LOWER);

        assertThat(kvs.getMetadataForTable(LOWER_LOWER)).isNotEmpty(); // wat
    }
```

but the methods I'm aware of to fix this require considering O(2^n) elements, so there's probably nothing much reasonable we can do

gmaretic: Arguably, as long as there are entries there we should really clean up. I can follow up later

docs/source/release_notes/release-notes.rst
```diff
@@ -50,8 +50,12 @@ develop
     *    - Type
          - Change
 
-    *    -
-         -
+    *    - |fixed|
+         - Cassandra KVS `getAllTimestamps` method no longer returns entries for tables that were dropped.
```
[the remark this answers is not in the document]
gmaretic: oops 🗡

PR #3651, opened 2018-11-22
Gozakdag/commit time read validation removal

gokcanozakdag:
**Goals (and why)**:
One less RPC to timelock. (Affects only read-only transactions on thorough swept tables)

If we are checking validity of immutable timestamp lock after every read, then at commit time we can guarantee that we have a consistent snapshot view without checking immutable timestamp lock once more.

Using existing tests.

Is there an edge case that I am missing?

`SnapshotTransaction.java`

Early next week?


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -755,10 +755,14 @@ private void validatePreCommitRequirementsOnReadIfNecessary(TableReference table
     }
 
     private boolean isValidationNecessaryOnReads(TableReference tableRef) {
-        return validateLocksOnReads && isValidationNecessary(tableRef);
+        return validateLocksOnReads && isThoroughSwept(tableRef);
```
sandorw: Not sure if we've set a convention on this, but `isThoroughlySwept` reads better

sandorw:
@jeremyk-91 for another set of eyes to confirm that we're not violating anything here

PR #3652, opened 2018-11-23
Fix postgres utility table creation race condition

gmaretic:
**Goals (and why)**:
Fix bug that was causing TodoEteTest flakes.

Due to a race condition, it was possible to insert multiple entries into the dual table at initialization time (see https://www.postgresql.org/docs/7.2/xact-read-committed.html). This would permanently break some functionality such as cause an infinite loop when we try to write a garbage deletion sentinel.
The simple fix is to simply create the dual table with a pk constraint so that we cannot end up in this situation. Kept the old query not to break existing dual tables without the constraint (everyone who had the table correctly initialized the first time atlas started on postgres is fine).
 
Some manual tests, but as long as existing tests pass, there shouldn't be anything to worry about.

What am I doing with my life?

Fixes really annoying flake, so asap

### DIFF
--- a/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresTableInitializer.java
+++ b/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresTableInitializer.java
@@ -35,12 +35,13 @@ public class PostgresTableInitializer implements DbTableInitializer {
     @Override
     public void createUtilityTables() {
         executeIgnoringError(
-                "CREATE TABLE dual (id BIGINT)",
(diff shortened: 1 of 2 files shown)


atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresTableInitializer.java
```diff
@@ -35,12 +35,13 @@ public PostgresTableInitializer(ConnectionSupplier conns) {
     @Override
     public void createUtilityTables() {
         executeIgnoringError(
-                "CREATE TABLE dual (id BIGINT)",
+                "CREATE TABLE dual (id BIGINT, CONSTRAINT pk_dual PRIMARY KEY (id))",
```
jeremyk-91: nit: probably use `PrimaryKeyConstraintNames.get("dual")`?

PR #3653, opened 2018-11-26
Remove usages of Runtime#addShutdownHook to cleanup resources.

jkozlowski:
**Goals (and why)**:

Remove a memory leak due to usages of Runtime#addShutdownHook.

* Use TransactionManager#registerClosingCallback to shutdown PersistentLockManager instead of Runtime#addShutdownHook.
* Add checkstyle check to prohibit this use in the future.

I created and shutdown TransactionManagers in a loop in a small (500M) JVM. Pre-fix it was able to fill up OldGen pretty quickly and start CMSing all the time, with this fix it seems stable.

The only thing that the old code was doing was making sure that PersistentLockManager gets closed after BackgroundSweeperImpl. I believe this invariant should hold with the new code.

TransactionManagers

High, caused production problems.

### DIFF
--- a/.baseline/checkstyle/checkstyle.xml
+++ b/.baseline/checkstyle/checkstyle.xml
@@ -423,5 +423,11 @@
         </module>
 
         <!-- Stricter checks end -->
+
+        <module name="RegexpSinglelineJava">
+            <property name="id" value="DoNotUseShutdownHooks"/>
+            <property name="format" value=".*addShutdownHook\("/>
+            <property name="message" value="Shutdown hooks should not be used in AtlasDB, it is a library. Products using AtlasDB might be creating and destroying many instances of AtlasDB, therefore relying on shutdown hooks for cleaning up resources can lead to memory leaks."/>
+        </module>
     </module>
 </module>
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
@@ -180,7 +180,7 @@ public class PersistentLockManager implements AutoCloseable {
     }
 
     private void logFailedShutdown(Exception exception) {
-        log.warn("An exception occurred while shutting down. This means that we had the backup lock out when"
+        log.warn("An exception occurred while shutting down. This means that we had the backup lock out when "
                         + "the shutdown was triggered, but failed to release it. If this is the case, sweep "
                         + "or backup may fail to take out the lock in future. If this happens consistently, "
                         + "consult the following documentation on how to release the dead lock: "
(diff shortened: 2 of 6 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
```diff
@@ -172,4 +178,13 @@ private void waitForRetry() {
             throw new RuntimeException(e);
         }
     }
+
+    private void logFailedShutdown(Exception exception) {
+        log.warn("An exception occurred while shutting down. This means that we had the backup lock out when"
+                        + "the shutdown was triggered, but failed to release it. If this is the case, sweep "
```
jeremyk-91: nit: missing space between the strings

docs/source/release_notes/release-notes.rst
```diff
@@ -50,8 +50,9 @@ develop
     *    - Type
          - Change
 
-    *    -
-         -
+    *    - |fixed|
+         - Remove a memory leak due to usages of Runtime#addShutdownHook to cleanup resources.
```
jeremyk-91: Probably obvious, but may be worth saying this only affects users if they use >1 instance of AtlasDB per JVM.

jkozlowski: Done.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/PersistentLockManager.java
jkozlowski: Well spotted! Fixed

j-baker:
does this make Atlas more likely to leak the backup lock?

.baseline/checkstyle/checkstyle.xml
```diff
@@ -423,5 +423,11 @@
         </module>
 
         <!-- Stricter checks end -->
+
+        <module name="RegexpSinglelineJava">
```
j-baker: would you be opposed to contributing this to baseline? internally, in most cases you want this to be a managed thing rather than registering a shutdown hook

jkozlowski:
@j-baker I think you are right, well spotted: that would explain why it was a shutdownHook. We might need to bring this behaviour back, but clearly deregister it on TransactionManager#close.

But it still feels a bit wrong to have that in a library.

jkozlowski:
I threw something together quickly https://github.com/palantir/atlasdb/pull/3655

.baseline/checkstyle/checkstyle.xml
jkozlowski: here it goes: https://github.com/palantir/gradle-baseline/pull/473

PR #3655, opened 2018-11-26
Quick attempt at shutdown hooks.

jkozlowski:
**Goals (and why)**:

This might be buggy, but wanted to throw something together before I leave.

* I'm wondering though how important these are vs. hooking into the web server hooks to actually shutdown cleanly. This kind of functionality should really live there and TransactionManager should really be just another thing that needs to be shutdown cleanly. So maybe we should instead make sure TransactionManager can be cleanly hooked into the server's lifetime: make some sort of a dev break that forces you to hook into some sort of lifecycle manager.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/RuntimeBackedShutdownHookRunner.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/RuntimeBackedShutdownHookRunner.java
@@ -16,6 +16,7 @@
 
 package com.palantir.atlasdb.factory;
 
+@SuppressWarnings("DoNotUseShutdownHooks")
 enum RuntimeBackedShutdownHookRunner implements ShutdownHookRunner {
 
     INSTANCE;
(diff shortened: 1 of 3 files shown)


j-baker:
Not sure I think we should be doing this; rather I think we should be just removing the sweep lock

jkozlowski:
@j-baker that's an option too, do we have that on the docket?

sandorw:
That requires rolling out Rescue 4.x, since it removes the need for the backup lock.

jkozlowski:
Ok, are we good to close this then and agree that the state of the world is now OK with my previous change? Or is there some followup to be had

sandorw:
It sounds like it's a combination of the two. Rescue 4.x requires Timelock and no longer deals with the backup lock at all. Rescue 3.x will fallback to taking the backup lock if the service doesn't use Timelock.

sandorw:
Closing out, we'll follow-up once we can require Rescue 4.x.

sandorw:
Rescue now takes the immutable timestamp instead of the backup lock (except for embedded timestamp/lock services, which are still supported on 3.x), which is the behavior we need to require across the board before removing the lock.

sandorw:
Nope, it's a normal Timelock lock, so it's not persistent and will eventually timeout and expire.

jkozlowski:
Awesome, makes sense.

PR #3663, opened 2018-11-28
[_transactions2] Coordination, Part 4: Reuse Existing Values in the Coordination Store

jeremyk-91:
**Goals (and why)**:
- Most of the time, coordination involves perpetuating existing values. When this happens, we don't want to have to re-write existing values that are already stored there, to avoid unnecessarily bloating disk.

- If a transform transforms a valueAndBound to the value that's present, advance the bound, but don't actually write a new value. Previously if the latest version was `v3` and bound was `10M`, we might have done a CAS to `(v4, 15M)` while now we do it to `(v3, 15M)`.

`valuePreservingTransformationsDoNotWriteTheSameValueAgain` is a new test that verifies this behaviour (and would have failed on the previous branch).
Other tests should sanity check that things aren't horribly broken.

- There's an extra object equality check here on the objects stored in the coordination store. Is this reasonable?
- Might this optimisation be broken in a way I haven't thought of?

**Where should we start reviewing?**: KVSCoordinationStoreTest

### DIFF
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
@@ -186,7 +186,8 @@ public final class KeyValueServiceCoordinationStore<T> implements CoordinationSt
         }
     }
 
-    private Optional<SequenceAndBound> getCoordinationValue() {
+    @VisibleForTesting
+    Optional<SequenceAndBound> getCoordinationValue() {
         return readFromCoordinationTable(getCoordinationValueCell())
                 .map(Value::getContents)
                 .map(this::deserializeSequenceAndBound);
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
@@ -162,8 +162,8 @@ public final class KeyValueServiceCoordinationStore<T> implements CoordinationSt
             Optional<SequenceAndBound> coordinationValue,
             Optional<T> extantValue,
             T targetValue) {
-        return coordinationValue.isPresent() &&
-                extantValue.map(presentValue -> presentValue.equals(targetValue)).orElse(false);
+        return coordinationValue.isPresent()
+                && extantValue.map(presentValue -> presentValue.equals(targetValue)).orElse(false);
     }
 
     private CheckAndSetResult<ValueAndBound<T>> extractRelevantValues(T targetValue, long newBound,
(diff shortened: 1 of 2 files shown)


atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -125,18 +125,45 @@ private KeyValueServiceCoordinationStore(
...
+    private SequenceAndBound determineNewSequenceAndBound(
+            Optional<SequenceAndBound> coordinationValue,
+            ValueAndBound<T> extantValueAndBound,
+            T targetValue) {
+        long sequenceNumber;
+        long newBound;
+        if (shouldReuseExtantValue(coordinationValue, extantValueAndBound.value(), targetValue)) {
+            // Safe as we're only on this branch if the value is present
+            sequenceNumber = coordinationValue.get().sequence();
+            newBound = getNewBound(sequenceNumberSupplier.getAsLong());
```
sandorw: nit: maybe update the argument of `getNewBound` to not take a sequence number, since we're clearly not passing in the sequence number here. I guess this is just some blurring of sequence numbers and timestamps that's hard to resolve.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -125,18 +125,45 @@ private KeyValueServiceCoordinationStore(
     @Override
     public CheckAndSetResult<ValueAndBound<T>> transformAgreedValue(Function<ValueAndBound<T>, T> transform) {
         Optional<SequenceAndBound> coordinationValue = getCoordinationValue();
-        T targetValue = transform.apply(
-                ValueAndBound.of(coordinationValue.flatMap(
-                        sequenceAndBound -> getValue(sequenceAndBound.sequence())),
-                        coordinationValue.map(SequenceAndBound::bound).orElse(SequenceAndBound.INVALID_BOUND)));
+        ValueAndBound<T> extantValueAndBound = ValueAndBound.of(coordinationValue.flatMap(
```
sandorw: This new logic looks like it should be fine. Whether we reuse a value or not, we still advance the bound, and the check and set request will still conflict appropriately since it includes the bound. Values are immutable if, so no issues there with reuse.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
jeremyk-91: Yep, that's a good call. I've updated the argument to `pointInTime` rather than `sequenceNumber`. Would it make sense to have a separate `boundSupplier` (potentially taking an extra timelock rpc?)

sandorw: I'm not too concerned, since this code won't get changed much (I hope), as long as it's clear to someone reading the code what we're doing and that it's correct - this line looked a bit fishy to me at first glance.

jeremyk-91: OK. Yeah, `sequenceNumber` for that method is definitely misleading after this PR. I'll leave it as `pointInTime` for now.

PR #3665, opened 2018-11-29
Fix quorum calculation for waiting on schema versions

gmaretic:
**Goals (and why)**:
After we do Cassandra schema mutations, we wait for the nodes to agree on schema versions. Ideally, we'd want all nodes to agree, but in order to be HA, we allow some nodes to be unreachable.

Historically, this was hardcoded to allow at most one node to be unreachable, and later changed to less than half. The first is very restrictive, and the latter doesn't give us the guarantees we want. This PR is to check that the available nodes guarantee write consistency, i.e. EACH_QUORUM. 

If we hit the interesting case of some nodes being unreachable and others on the same version, we request the necessary information from Cassandra and then check for each TokenRange that each datacenter has a quorum of available nodes.

Added a few tests for the different cases.

Is this actually what we want? Is this a reasonable way to do it? I'd like your feedback, @tpetracca @amandachow @ashrayjain

Probably from the tests

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
@@ -89,9 +89,6 @@ public final class CassandraKeyValueServices {
         long sleepTime = INITIAL_SLEEP_TIME;
         Map<String, List<String>> versions;
         do {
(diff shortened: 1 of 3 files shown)


amandachow:
Wanted to flag another quorum calculation, this is used in health checks: https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java#L1750

gmaretic:
Done differently in https://github.com/palantir/atlasdb/pull/3754

PR #3666, opened 2018-11-29
Move javax.anntations-api dependency to atlasdb-api

pkoenig10:
**Goals (and why)**:
Only `atlasdb-config` is needed to generate schemas. The generated schema classes contain the `Generated` annotation but the dependency is missing from `atlasdb-config`.

cc @robert3005


pkoenig10:
Closing in favor of #3677

PR #3667, opened 2018-12-03
Fix warning from errorprone in generated code.

jkozlowski:
**Goals (and why)**:

This warning is blocking upgrading to newer gradle-baseline in an internal product. For some reason errorprone is not recognising the @Generated annotation on atlasdb generated code; that is to be figured out, but fixing the warning is lowest lift fix.


jkozlowski:
Can I have a release please?

PR #3670, opened 2018-12-04
@AutoDelegate: Support Interface Statics and Generics

jeremyk-91:
**Goals (and why)**:
- Add support for static methods in interfaces
- Add basic support for generic types
- These features are generally useful; the latter is needed for the coordination store in transactions2

- Filter static methods out when trying to pick which methods we want to generate a copy of
- Add correct type parameter elements to the javapoet builder when generating an AutoDelegate class

There were some pretty good tests, but none for these specific behaviours. I've added new ones for these.

- Some more complex interactions with generics, particularly around subclassing don't work correctly. For example, if I have `interface One<A, B>` and `interface Two<C> extends One<Integer, C>`, methods from `One` generated in `AutoDelegate_Two` reference `A` and `B` which clearly don't exist. Even worse, if I have `interface One<A, B>` and `interface Two<B, A> extends One<A, B>`, the parameters for the `One` methods work opposite of the way they should.

**Where should we start reviewing?**: `GenericsTester` for an overview of what's broken, then `AutoDelegateProcessor`

### DIFF
--- a/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
+++ b/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
@@ -38,7 +38,6 @@ public class AutoDelegateGenericTests {
         Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_GenericsTester.class);
         Set<String> originalMethods = TestingUtils.extractNonStaticMethods(GenericsTester.class);
 
-        generatedMethods.removeAll(originalMethods);
         assertThat(Sets.difference(generatedMethods, originalMethods)).containsOnly("delegate");
     }
 }
--- a/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
+++ b/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
@@ -1,0 +1,53 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.processors;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import java.lang.reflect.Modifier;
+import java.util.Set;
+
+import org.junit.Test;
+
+public class AutoDelegateGenericTests {
+    @Test
+    public void generatedInterfaceHasInterfaceMethods() {
+        Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_GenericsTester.class);
+        Set<String> originalMethods = extractNonStaticMethods(GenericsTester.class);
+
+        assertThat(generatedMethods, hasItems(originalMethods.toArray(new String[0])));
+    }
+
+    @Test
+    public void generatedInterfaceHasDelegateMethod() {
+        Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_GenericsTester.class);
+        Set<String> originalMethods = extractNonStaticMethods(GenericsTester.class);
+
+        generatedMethods.removeAll(originalMethods);
+        assertThat(generatedMethods.size(), is(1));
+        assertThat(generatedMethods.iterator().next(), containsString("delegate"));
+    }
+
+    private Set<String> extractNonStaticMethods(Class klass) {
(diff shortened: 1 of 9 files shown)


atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.junit.Test;
+
+public class AutoDelegateGenericTests {
+    @Test
+    public void generatedInterfaceHasInterfaceMethods() {
+        Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_GenericsTester.class);
+        Set<String> originalMethods = extractNonStaticMethods(GenericsTester.class);
+
+        assertThat(generatedMethods, hasItems(originalMethods.toArray(new String[0])));
```
felixdesouza: please use assertJ for this:

```java
assertThat(generatedMethods.contains(originalMethods);
```

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test
+    public void generatedInterfaceHasDelegateMethod() {
+        Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_GenericsTester.class);
+        Set<String> originalMethods = extractNonStaticMethods(GenericsTester.class);
+
+        generatedMethods.removeAll(originalMethods);
```
felixdesouza: similarly here:
```java
assertThat(Sets.difference(generatedMethods, originalMethods))
    .containsOnly("delegate");
```

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
```diff
@@ -96,4 +107,10 @@ public void generatedInterfaceCallsMethodOnDelegate() {
         instanceOfInterface.methodWithReturnType();
         verify(mockImpl, times(1)).methodWithReturnType();
     }
+
+    private Set<String> extractNonStaticMethods(Class klass) {
```
felixdesouza: nit: `static`

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
```diff
@@ -0,0 +1,53 @@
...
+        assertThat(generatedMethods.size(), is(1));
+        assertThat(generatedMethods.iterator().next(), containsString("delegate"));
+    }
+
+    private Set<String> extractNonStaticMethods(Class klass) {
```
felixdesouza: nit: `static`

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
```diff
@@ -63,26 +65,35 @@ public void packagePrivateInterfacesGeneratePackagePrivateInterfaces() {
...
         assertThat(generatedMethods.size(), is(1));
         assertThat(generatedMethods.iterator().next(), containsString("delegate"));
     }
 
+    @Test
+    public void generatedInterfaceDoesNotHaveStaticMethods() {
+        Set<String> generatedMethods = TestingUtils.extractMethods(AutoDelegate_TestInterface.class);
+        Set<String> originalStaticMethods = TestingUtils.extractMethodsSatisfyingPredicate(TestInterface.class,
+                method -> Modifier.isStatic(method.getModifiers()));
+
```
felixdesouza: `assertThat(Sets.intersection(generatedMethods, originalMethods)).isEmpty()`?

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
felixdesouza: perhaps move this to `TestingUtils`?

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
jeremyk-91: Ah, ok. I copied this from the other classes (where I'd prefer to stick with old-style as I'm not feeling up to changing all the old tests to use AssertJ), but no reason to hang on to this for new classes.

jeremyk-91: nit: we want `containsAll`, `contains` is varargs on elements

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateGenericTests.java
jeremyk-91: Yep, gets rid of the duplication as well

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
jeremyk-91: Yep, that works. I'll stick with old-style here as having both junit and assertj styles in the same file gets annoying for imports.

atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
jeremyk-91: (This method was moved to `TestingUtils`, and is static there.)

PR #3674, opened 2018-12-05
depend on sls-cass 3.24

tpetracca:
CC @leonz @j-baker


felixdesouza:
Can you retrieve the version of cassandra in code? Might be worth failing on creation if version is less than expected (for those who don't have the gradle constraint stuff).

sandorw:
Can follow-up on additional safeguards later if needed, going to merge to unblock.

PR #3675, opened 2018-12-05
thrift pue support logic

tpetracca:
CC @j-baker 

@jeremyk-91 not sure how you intended on actually wiring up the bits that decide if you use transactions2 or not, but in theory you should just be able to take advantage of the method I added here.

edit: And as discussed offline, it's worth noting that the code before this PR is written in a way that won't break with clients that provide a version with the `-pt` in it.


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraApiVersionTest.java
```diff
@@ -62,4 +62,15 @@
...
     }
 
+    @Test public void
+    version_20_1_0_does_not_support_pue() {
+        CassandraApiVersion version = new CassandraApiVersion("20.1.0");
+        assertThat(version.supportsThriftPutUnlessExists(), is(false));
+    }
+
+    @Test public void
+    version_20_1_0_pt0_supports_cas() {
```
jeremyk-91: nit: `supports_pue`?

PR #3678, opened 2018-12-06
[_transactions2] Coordination, Part 5: Async Initializable Coordination Store

jeremyk-91:
**Goals (and why)**:
- Make the coordination store async initializable
- Transactions2 / Transactions Service eventually needs to follow the rules of async initialization.

- Implement standard async initialization primitives for `KeyValueServiceCoordinationStore`

There is one new test that handles the case where the KVS isn't ready initially, but becomes ready

**Where should we start reviewing?**: `KeyValueServiceCoordinationStore.java`

**Priority (whenever / two weeks / yesterday)**: This week, hopefully? If not early next

### DIFF
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
@@ -20,11 +20,23 @@ import java.util.Optional;
 import java.util.function.Function;
 
 import com.palantir.atlasdb.keyvalue.impl.CheckAndSetResult;
+import com.palantir.processors.AutoDelegate;
 
 /**
  * A {@link CoordinationStore} stores data that a {@link CoordinationService} may use.
  */
+@AutoDelegate
 public interface CoordinationStore<T> {
+    /**
+     * Coordination stores may require asynchronous initialization, if dependencies aren't initially available.
+     *
+     * @return true iff the coordination store is ready to service requests
+     */
+
+    default boolean isInitialized() {
+        return true;
+    }
+
     /**
      * Gets the value stored in this {@link CoordinationStore}. This value may not be the most recent value; however,
      * it is guaranteed that any value returned by this method will be at least as current as any value returned by
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
@@ -32,7 +32,6 @@ public interface CoordinationStore<T> {
      *
      * @return true iff the coordination store is ready to service requests
      */
-
     default boolean isInitialized() {
         return true;
     }
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
@@ -31,7 +31,9 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.palantir.async.initializer.AsyncInitializer;
 import com.palantir.atlasdb.AtlasDbConstants;
+import com.palantir.atlasdb.coordination.AutoDelegate_CoordinationStore;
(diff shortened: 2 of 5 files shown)


atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
```diff
@@ -20,11 +20,23 @@
...
  * A {@link CoordinationStore} stores data that a {@link CoordinationService} may use.
  */
+@AutoDelegate
 public interface CoordinationStore<T> {
+    /**
+     * Coordination stores may require asynchronous initialization, if dependencies aren't initially available.
+     *
+     * @return true iff the coordination store is ready to service requests
+     */
+
```
sandorw: nit: empty line

atlasdb-coordination-impl/src/test/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStoreTest.java
```diff
@@ -159,10 +169,31 @@ public void multipleStoresCanCoexist() {
...
+                .createTable(any(), any());
+
+        CoordinationStore<String> asyncStore = KeyValueServiceCoordinationStore.create(
+                ObjectMappers.newServerObjectMapper(),
+                kvs,
+                COORDINATION_ROW,
+                timestampSequence::incrementAndGet,
+                String.class,
+                true);
+        Awaitility.waitAtMost(Duration.ONE_MINUTE)
```
sandorw: Is this how we generally test this elsewhere? The alternative would be to just call `tryInitialize` directly, assuming the general async initialization stuff is tested independently.

What's the default wait period for async init? It looks like the default behavior is to wait 10 seconds.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -319,4 +332,21 @@ private static TableMetadata getCoordinationTableMetadata() {
                 false,
                 TableMetadataPersistence.LogSafety.SAFE);
     }
+
+    private class InitializingWrapper extends AsyncInitializer implements AutoDelegate_CoordinationStore<T> {
```
gmaretic: The convention we used is to put this as the first item in the class

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
gmaretic: nit: newline

jeremyk-91: Good catch, removed.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
jeremyk-91: Alright, have brought it to the front.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationStore.java
jeremyk-91: Yep, fixed - interesting that this rendered separately from Sandor's comment

atlasdb-coordination-impl/src/test/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStoreTest.java
jeremyk-91: Right, the default wait period is 10s. I agree that it's probably better to hit `tryInitialize` for this test, to keep test duration short and avoid waiting 10s - also avoid failures if something else in async init like the standard wait period changes.

Not aware of how this is tested elsewhere apart from integration tests - e.g. `LockStoreImpl`, another use of this, is untested.

jeremyk-91: Post offline discussion - Not sure if this test brings much value, as the framework is tested elsewhere, and this is a straightforward usage (no cleanup tasks), so removing this.

Also taking out Awaitility as we don't use it elsewhere

PR #3679, opened 2018-12-06
[_transactions2] Coordination, Part 6: Semantics Change in TransactionSchemaManager

jeremyk-91:
**Goals (and why)**:
- Get round a problem where proposing a value needs a timestamp, but embedded services with a leader block cannot give you timestamps until initialization is complete.

- Change the behaviour of `perpetuateExistingState` and `installNewVersion` in `TransactionSchemaManager` to allow planting of the default value. 

The integration test that used to use the initializer now relies on behaviour we have (`TransactionSchemaManagerIntegrationTest`).

- Is returning booleans for whether the version was installed reasonable?
- Are there any errors in constructing the result?

**Where should we start reviewing?**: `TransactionSchemaManager.java`

Also, note that merging into develop (and _not_ atop `jkong/coord-5-1` is intentional).


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/InternalSchemaMetadata.java
```diff
@@ -28,9 +31,20 @@
 @JsonSerialize(as = ImmutableInternalSchemaMetadata.class)
 @JsonDeserialize(as = ImmutableInternalSchemaMetadata.class)
 public interface InternalSchemaMetadata {
-    TimestampPartitioningMap<Integer> timestampToTransactionsTableSchemaVersion();
+    // Warning: Do not change this without a migration
+    byte[] DEFAULT_METADATA_COORDINATION_KEY = PtBytes.toBytes("m");
```
gmaretic: Put this in `AtlasDbConstants`? Also not sure about the comment, since we have constants like this elsewhere that shouldn't be changed 😢

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
```diff
@@ -45,33 +48,49 @@ public int getTransactionsSchemaVersion(long timestamp) {
...
-            return InternalSchemaMetadata.builder()
-                    .from(internalSchemaMetadata)
-                    .timestampToTransactionsTableSchemaVersion(
-                            installNewVersionInMap(
-                                    internalSchemaMetadata.timestampToTransactionsTableSchemaVersion(),
-                                    valueAndBound.bound() + 1,
-                                    newVersion))
-                    .build();
-        });
+        InternalSchemaMetadata internalSchemaMetadata = valueAndBound.value().get();
```
gmaretic: InternalSchemaMetadata internalSchemaMetadata = valueAndBound.value().orElse(InternalSchemaMetadata.defaultValue())

jeremyk-91: Yep, that's better. This does change behaviour (in that on the first update you'd previously just install default, while now you install default _and_ install a new version at the right place), but I think it's a good change.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/InternalSchemaMetadata.java
jeremyk-91: That's fair. I've bumped it to `AtlasDbConstants`.

I think with some of these it's obvious that you can't change them without blowing stuff up (e.g. tablereferences) hence comments aren't needed. In any case I think this one is pretty obviously dangerous too, so removed it.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
jeremyk-91: Subsequent offline discussion - unfortunately doesn't work because the transform on its first go has no idea what to set the new version of the bound to be. This can be addressed by calling the timestamp service directly, or making transforms a bi-function that also has a timestamp that is the bound we'll propose our value at. For now we'll just use the old semantics...

PR #3680, opened 2018-12-06
[_transactions2] Part 8: Split Key Delegating Transaction Service

jeremyk-91:
**Goals (and why)**:
- Have a transaction service that knows how to deal with multiple (or unknown!) transaction schema versions

- Implement `SplitKeyDelegatingTransactionService<T>` that has a function from timestamp to some key, and a mapping of key to transaction services.

New code is tested.

- Would it be easier to just have a single `Function<Long, TransactionService>` to map timestamps to services?

**Where should we start reviewing?**: There's one class and its tests - pick one.

**Priority (whenever / two weeks / yesterday)**: next week?


jeremyk-91:
This isn't the right branch, this is a testing branch for wiring everything up

PR #3682, opened 2018-12-07
(KVS migration) Fast forward source KVS timestamps as well

gmaretic:
**Goals (and why)**:
Avoid specific issue if we tablesplitting kvs for atomic tables, attempt a migration, and then decide to instead keep using the old KVS. If we are not on timelock, we can end up reusing timestamps that are already precommitted in the _transactions table.

Maybe this is splitting hairs, but it's simple enough to fix

### DIFF
--- a/docs/source/cluster_management/kvs-migration.rst
+++ b/docs/source/cluster_management/kvs-migration.rst
@@ -43,6 +43,12 @@ Then, we immediately insert an entry into the transactions table of the target K
     If you are using TimeLock, the TimeLock server must be running in order to do the migration.
     Otherwise, you must use the ``--offline`` flag, which will remove the leader block from your configuration for the purposes of migration.
 
+.. warning::
+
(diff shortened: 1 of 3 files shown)


docs/source/cluster_management/kvs-migration.rst
```diff
@@ -43,6 +43,12 @@ Then, we immediately insert an entry into the transactions table of the target K
     If you are using TimeLock, the TimeLock server must be running in order to do the migration.
     Otherwise, you must use the ``--offline`` flag, which will remove the leader block from your configuration for the purposes of migration.
 
+.. warning::
+
+    After attempting a migration, the source KVS is likely going to be behind in timestamps than the target KVS.
```
sandorw: 'behind the target KVS in timestamps'

gmaretic:
I'm going to close it out because there is still a possible corner case we can hit, so I'd prefer to not give a false sense of security

PR #3686, opened 2018-12-11
[_transactions2] Coordination, Part 7: Closing the Circuit

jeremyk-91:
- [x] smoke test with internal blob service

**Goals (and why)**:
- Prepare for releasing a version of AtlasDB which is aware of the coordination table and _transactions2, and knows it needs to read from this table when committing transactions. This version of AtlasDB can be rolling-upgraded from previous versions of AtlasDB, and supports rolling-upgrades to a version of AtlasDB where we wish to install _transactions2.
- Fix a bug with the old TransactionsSchemaManager when encountering sweep sentinels.

- Wire up the coordination service as implemented in #3619 and #3679 to actually be used by an AtlasDB transaction manager. This covers tests as well.
- Fix a bug with sweep sentinels in the `SplitKeyDelegatingTransactionService` and `TransactionSchemaManager` - these would previously have caused an infinite loop, since we'd have consistently tried to advance the bound in hope of getting an agreement on `-1` which is useless.

- Many existing integration tests (e.g. anything extending from `TransactionTestSetup`, and the `TargetedSweeperTest`) have been rigged up to use the new transaction service - the exceptions are tests which don't have a proper timestamp service.
- The new behaviour pertaining to sweep sentinels has been added as an explicit case to `TransactionSchemaManagerIntegrationTest`.

- The plumbing around Targeted Sweeper and the queues is a little messy; I chose to do this rather than expose a `getTransactionService()` method on a `TransactionManager`. It's perfectly fine that these are distinct objects (as they would be across multiple JVMs/nodes talking to the same keyspace, anyway)
- Are there any other unexpected pitfalls?

**Where should we start reviewing?**: `TransactionManagers.java`

**Priority (whenever / two weeks / yesterday)**: This week, ideally. This (plus a subsequent release) is the last piece to our December goals for _transactions2.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
```diff
@@ -163,15 +172,31 @@ private SweepQueueFactory(
...
+                TimelockService timelock,
+                Supplier<Integer> shardsConfig,
+                TransactionService transaction) {
             Schemas.createTablesAndIndexes(TargetedSweepSchema.INSTANCE.getLatestSchema(), kvs);
             ShardProgress shardProgress = new ShardProgress(kvs);
             Supplier<Integer> shards = createProgressUpdatingSupplier(shardsConfig, shardProgress,
                     SweepQueueUtils.REFRESH_TIME);
             WriteInfoPartitioner partitioner = new WriteInfoPartitioner(kvs, shards);
-            SweepableCells cells = new SweepableCells(kvs, partitioner, metrics);
+
```
gmaretic: nit: newline

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -114,6 +117,11 @@ public void initializeWithoutRunning(TransactionManager txManager) {
         initializeWithoutRunning(SpecialTimestampsSupplier.create(txManager),
                 txManager.getTimelockService(),
                 txManager.getKeyValueService(),
+                TransactionServices.createTransactionService(txManager.getKeyValueService(),
+                        CoordinationServices.createDefault(
+                                txManager.getKeyValueService(),
+                                txManager.getTimestampService(),
+                                false)),
```
gmaretic: discussed offline: does not need to be async, since it will be tried when the requirements are ready

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/TransactionServices.java
```diff
@@ -15,15 +15,48 @@
...
+import com.palantir.atlasdb.internalschema.InternalSchemaMetadata;
+import com.palantir.atlasdb.internalschema.TransactionSchemaManager;
+import com.palantir.atlasdb.internalschema.persistence.CoordinationServices;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.timestamp.TimestampService;
 
 public final class TransactionServices {
     private TransactionServices() {
-        // Uility class
+        // Utility class
```
gmaretic: 🎉

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/cleaner/Helper.java
```diff
@@ -1,68 +0,0 @@
...
-import com.google.common.collect.Sets;
-import com.palantir.atlasdb.keyvalue.api.Cell;
-import com.palantir.atlasdb.keyvalue.api.KeyValueService;
-import com.palantir.atlasdb.keyvalue.api.TableReference;
-import com.palantir.atlasdb.keyvalue.api.Value;
-import com.palantir.atlasdb.keyvalue.impl.InMemoryKeyValueService;
-import com.palantir.atlasdb.transaction.service.TransactionService;
-import com.palantir.atlasdb.transaction.service.TransactionServices;
-
-public final class Helper {
```
gmaretic: Best class ever 🏅

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/TransactionServices.java
```diff
@@ -15,15 +15,48 @@
...
     }
 
     public static TransactionService createTransactionService(
-            KeyValueService keyValueService) {
+            KeyValueService keyValueService, CoordinationService<InternalSchemaMetadata> coordinationService) {
```
sandorw: Will need to coordinate releases with other internal services/libraries that need a `TransactionService`

jeremyk-91: Good point. I will mark this as a dev break (which it is)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
jeremyk-91: Removed the newline.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/TransactionServices.java
jeremyk-91: Also, if people are doing this then it probably makes sense for the `TransactionManager` to expose the `TransactionService` it uses

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
```diff
@@ -87,9 +92,16 @@ public boolean tryInstallNewTransactionsSchemaVersion(int newVersion) {
     private InternalSchemaMetadata installNewVersionInMapOrDefault(int newVersion,
             ValueAndBound<InternalSchemaMetadata> valueAndBound) {
         if (!valueAndBound.value().isPresent()) {
+            log.info("Attempting to install a new transactions schema version {}, but no past data was found,"
```
gmaretic: Make it a warn, and say this should happen only once on first startup?

jeremyk-91: Yep, done. The wording is unfortunately a little more guarded, because of the possibility of a race condition.

PR #3687, opened 2018-12-11
ping leader with try

gokcanozakdag:
**Goals (and why)**:
Leader ping should not fail if one of the hosts in timelock cluster responds slow - causing thread pool to fill.

Wrap submit calls to try-catch.

- Is the log message sensible?

`PaxosLeaderElectionService`


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -288,8 +292,15 @@ boolean getAndRecordLeaderPingResult(@Nullable Future<Boolean> pingFuture) throw
         // kick off requests to get leader uuids
         List<Future<Entry<String, PingableLeader>>> allFutures = Lists.newArrayList();
         for (final PingableLeader potentialLeader : otherPotentialLeadersToHosts.keySet()) {
-            allFutures.add(pingService.submit(potentialLeader,
-                    () -> Maps.immutableEntry(potentialLeader.getUUID(), potentialLeader)));
+            try {
+                allFutures.add(pingService.submit(potentialLeader,
+                        () -> Maps.immutableEntry(potentialLeader.getUUID(), potentialLeader)));
```
j-baker: this feels like it treats the symptom rather than the cause. Instead of allowing the thread pool to get wrecked, and then backing off, maybe we should be setting some timeouts elsewhere?

PR #3700, opened 2018-12-20
A partitioning timelock service to reduce the request sizes going to timelock

j-baker:
Occasionally the clients of TimelockService make requests which are larger than we'd reasonably expect the server to process, clients should automatically batch their requests to avoid making such large requests.


lock-api/src/main/java/com/palantir/lock/client/PartitioningTimelockService.java
```diff
@@ -0,0 +1,176 @@
...
+                }
+
+                LockRequest batchedRequest = ImmutableLockRequest.builder()
+                        .from(request)
+                        .acquireTimeoutMs(acquireTimeout.get().toMillis())
+                        .lockDescriptors(batch)
+                        .build();
+                LockResponse response = timelockService.lock(batchedRequest);
+                if (!response.wasSuccessful()) {
+                    return descriptors.failure();
```
markelliot: propagate the lock response?

markelliot: ie copy the failure mode

lock-api/src/main/java/com/palantir/lock/client/PartitioningTimelockService.java
```diff
@@ -0,0 +1,172 @@
...
+    public Set<LockToken> unlock(Set<LockToken> unenrichedTokens) {
+        Set<LockToken> tokens = addProxies(unenrichedTokens);
+        if (tokens.size() <= UNLOCK_BATCH_THRESHOLD) {
+            return timelockService.unlock(tokens);
+        }
+        return Streams.stream(Iterables.partition(tokens, UNLOCK_BATCH_THRESHOLD))
+                .map(ImmutableSet::copyOf)
+                .map(timelockService::unlock)
+                .flatMap(Collection::stream)
+                .collect(toSet());
```
markelliot: the result of this contains the proxy tokens and the tokens they proxy, but should probably only contain the proxy tokens.

j-baker: yep makes sense

lock-api/src/main/java/com/palantir/lock/client/PartitioningTimelockService.java
j-baker: yep - there is no other failure mode than 'timed out'.

lock-api/src/main/java/com/palantir/lock/client/PartitioningTimelockService.java
```diff
@@ -0,0 +1,206 @@
...
+                LockResponse response = timelockService.lock(batchedRequest);
+                if (!response.wasSuccessful()) {
+                    return descriptors.failure(response);
+                }
+                descriptors.add(response.getToken());
+            }
+        } catch (RuntimeException e) {
+            throw descriptors.failure(e);
+        }
+        return descriptors.success();
```
jeremyk-91: It's possible I'm being overly pedantic/worrying too much about the semantics here. My understanding of the guarantees of the lock service was that if you get a successful response, then there existed a point in time (not necessarily when the method returns) where you had all the locks. Currently this is true on the timelock server side, if I read the code correctly (notice that while a request is in flight, `HeldLocksCollection#shouldRemove()` will never be true).

Here if we have 2 batches and a long GC between batch 1 and 2, we might lose the locks in batch 1 which someone else might acquire but still return success, meaning we never held all of them simultaneously. I don't think this affects correctness of the Atlas protocol in that we don't commit unless we refresh our token successfully later on. 

I'm not sure how/if this will affect broken workflows where people rely on the transaction task in a `runTaskWithRetry(...)` not running without being able to acquire all the locks for an instant, though it's arguable (and I'd agree) that that's not something we should concern ourselves with. 

That said, I'm not sure I have a better solution - one way is to immediately refresh the tokens and confirm all are there before returning OK, but that's unperformant for Atlas write transactions which are going to check these tokens later anyway. Alternatively, if there's a way to clarify that this `TimelockService` might not exactly have the same semantics as a non-batching one, that would work for me too.

@sandorw for a sanity check on reasoning above?

sandorw:
Will be doing something different for the time being - just going to throw client side if requests are this large.

PR #3701, opened 2018-12-22
Implement leadership leases in timelock

j-baker:
I was feeling bored. Think this is a fairly simple (probably correct)
implementation. Would like to get feedback, and ultimately get this
released internally within the first couple weeks of q1.

Live migration onto and off of it is easily possible

**Goals (and why)**:

- Dramatically reduce timelock costs, which appear to be around 10-40ms of quorum checking on a busy service.
- Keep it out of the Paxos code, since this can be modular
- Keep it understandable; e.g. isolate nanotime complexities

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
@@ -19,6 +19,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Spliterator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
 
@@ -85,9 +88,9 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
          * result.
          */
         if (delegate() instanceof List) {
-            return of(Lists.transform(castAsList(), function));
+            return of(Lists.transform((List<T>) delegate(), function));
         } else if (delegate() instanceof Collection) {
-            return of(Collections2.transform(castAsCollection(), function));
+            return of(Collections2.transform((Collection<T>) delegate(), function));
         }
         return of(Iterables.transform(delegate(), function));
     }
@@ -118,7 +121,7 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
              * Use the more efficient Lists.partition which utilizes sublists
              * without allocating new lists for the returned partitions.
              */
-            return of(Lists.partition(castAsList(), size));
+            return of(Lists.partition((List<T>) delegate(), size));
         }
 
         return of(Iterables.partition(castAsIterable(), size));
@@ -213,6 +216,23 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
     }
 
     @Override
+    @SuppressWarnings("unchecked")
+    public Spliterator<T> spliterator() {
+        return (Spliterator<T>) delegate().spliterator();
+    }
+
+    /**
+     * @return a sequential {@code Stream} with the contents as its source.
+     */
+    @SuppressWarnings("unchecked")
+    public Stream<T> stream() {
+        if (delegate() instanceof Collection) {
+            return ((Collection<T>) delegate()).stream();
+        }
+        return StreamSupport.stream(spliterator(), false);
+    }
+
+    @Override
     public String toString() {
         return Iterables.toString(delegate());
     }
@@ -224,34 +244,6 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
     @SuppressWarnings("unchecked")
     private Iterable<T> castAsIterable() {
         return (Iterable<T>)delegate();
-    }
-
-    /**
-     * Casts the {@link #delegate()} as a {@link Collection} of T to avoid type
-     * warnings due to type erasure. This is safe if the delegate has already
-     * been checked to be an instance of {@link Collection}, but will throw if
-     * not a {@link Collection}
-     *
-     * @throws ClassCastException if the {@link #delegate()} is not a
-     *         {@link Collection}
-     */
-    @SuppressWarnings("unchecked")
-    private Collection<T> castAsCollection() {
-        return (Collection<T>) delegate();
-    }
-
-    /**
-     * Casts the {@link #delegate()} as a {@link List} of T to avoid type
-     * warnings due to type erasure. This is safe if the delegate has already
-     * been checked to be an instance of {@link List}, but will throw if
-     * not a {@link List}
-     *
-     * @throws ClassCastException if the {@link #delegate()} is not a
-     *         {@link List}
-     */
-    @SuppressWarnings("unchecked")
-    private List<T> castAsList() {
-        return (List<T>) delegate();
     }
 
 }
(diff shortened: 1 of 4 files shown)


leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
```diff
@@ -0,0 +1,102 @@
...
+        return new LeadershipLeaseLeaderElectionService(delegate, Duration.ZERO, Duration.ofSeconds(2));
+    }
+
+    public static LeaderElectionService wrap(LeaderElectionService delegate) {
+        return new LeadershipLeaseLeaderElectionService(delegate, Duration.ofSeconds(1), Duration.ofSeconds(2));
+    }
+
+    @Override
+    public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
+        return leaseTokens.get(delegate.blockOnBecomingLeader());
```
j-baker: this needs to handle interruption properly

leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
```diff
@@ -0,0 +1,102 @@
...
+        this.leaseExpiry = leaseExpiry;
+    }
+
+    // this method exists in order to provide a rolling migration mechanism for migrating to schema leases
+    public static LeaderElectionService wrapWithoutLeasing(LeaderElectionService delegate) {
+        return new LeadershipLeaseLeaderElectionService(delegate, Duration.ZERO, Duration.ofSeconds(2));
+    }
+
+    public static LeaderElectionService wrap(LeaderElectionService delegate) {
+        return new LeadershipLeaseLeaderElectionService(delegate, Duration.ofSeconds(1), Duration.ofSeconds(2));
```
j-baker: these constants should probably be shared

leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
+        return leaseTokens.get(delegate.blockOnBecomingLeader());
+    }
+
+    @Override
+    public Optional<LeadershipToken> getCurrentTokenIfLeading() {
+        return delegate.getCurrentTokenIfLeading().map(leaseTokens::get);
```
j-baker: it feels a bit goofy for this to be blocking in the way it is. Should probably have the cache contain an Entry class which only becomes valid past a point (+ tests).

leader-election-impl/src/main/java/com/palantir/leader/lease/ClockReversalDetector.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Suppliers;
+
+final class ClockReversalDetector {
```
j-baker: tempted to remove this and instead put a reversal check in the actual token class

leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
```diff
@@ -0,0 +1,102 @@
...
+            Duration leaseRefresh,
+            Duration leaseExpiry) {
+        this.delegate = delegate;
+        this.leaseRefresh = leaseRefresh;
+        this.leaseExpiry = leaseExpiry;
+    }
+
+    // this method exists in order to provide a rolling migration mechanism for migrating to schema leases
+    public static LeaderElectionService wrapWithoutLeasing(LeaderElectionService delegate) {
+        return new LeadershipLeaseLeaderElectionService(delegate, Duration.ZERO, Duration.ofSeconds(2));
```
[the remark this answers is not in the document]
j-baker: Discussed this with @gokcanozakdag and @jeremky-91 a few days ago. Right now, the cost of a leader election is:

1. Time each node will wait between pings before telling the leader that it is not the leader: 5 seconds
2. Time each node will wait before proposing itself as the leader: random duration between 0 and 5 seconds
3. Paxos round cost (4 RTTs), <100ms.

So this looks roughly like adding an expected leadership change delay of 25%.

leader-election-impl/src/test/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionServiceTests.java
```diff
@@ -0,0 +1,96 @@
...
+    @Test
+    public void testDelaysAcquiringLeadership_ifLeading() throws Exception {
+        testDelaysAcquiringLeadership(() -> leased.getCurrentTokenIfLeading().get());
+    }
+
+    private void testDelaysAcquiringLeadership(Callable<LeadershipToken> tokenSupplier) throws Exception {
+        long before = System.nanoTime();
+        tokenSupplier.call();
+        long after = System.nanoTime();
+        assertThat(after - before).isGreaterThan(TimeUnit.MILLISECONDS.toNanos(100));
```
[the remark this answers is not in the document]
j-baker: done

leader-election-impl/src/main/java/com/palantir/leader/lease/ClockReversalDetector.java
j-baker: so i'm actually just going to remove this and replace with an implementation of the clock skew monitor; but this can go into a FLUP PR.

leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
j-baker: fixed by just being like getIfPresent

leader-election-impl/src/main/java/com/palantir/leader/lease/NanoTime.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(time);
+    }
+
+    @Override
+    public int compareTo(NanoTime other) {
+        if (other.time == time) {
+            return 0;
+        } else if (time - other.time > 0) { // this is the critical bit; now can wrap according to javadoc
```
j-baker: this means that we can't lease for longer than 292 years...

leader-election-impl/src/main/java/com/palantir/leader/lease/LeadershipLeaseLeaderElectionService.java
gokcanozakdag: If this method returns Optional.empty, subsequent calls will always return Optional.empty unless `blockOnBecomingLeader()` is called in between. I have checked the usage in `AwaitingLeadershipProxy` and the contract stated in the `LeaderElectionService` interface, and that is inline with this implementation.

leader-election-impl/src/main/java/com/palantir/leader/lease/ClockReversalDetector.java
gokcanozakdag: we will probably need to have a ReversalDetector on client side for the client lock leases. I can take care of removing this class in that PR.

PR #3703, opened 2019-01-04
Add stream() to IterableView

schlosna:
**Goals (and why)**:

Allow simple use of `IterableView` as a `Stream`, optimized if the delegate is a `Collection`.

Override `IterableView.spliterator()` to allow for potential optimizations based on delegate characteristics (e.g. if the delegate is sized).

Added unit test coverage as there was no explicit coverage for `IterableView`.

`IterableView.stream()`

Whenever, nice to have for use in a downstream project to simplify & optimize stream usage.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
@@ -88,9 +88,9 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
          * result.
          */
         if (delegate() instanceof List) {
-            return of(Lists.transform(castAsList(), function));
+            return of(Lists.transform((List<T>) delegate(), function));
         } else if (delegate() instanceof Collection) {
-            return of(Collections2.transform(castAsCollection(), function));
+            return of(Collections2.transform((Collection<T>) delegate(), function));
         }
         return of(Iterables.transform(delegate(), function));
     }
@@ -121,7 +121,7 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
              * Use the more efficient Lists.partition which utilizes sublists
              * without allocating new lists for the returned partitions.
              */
-            return of(Lists.partition(castAsList(), size));
+            return of(Lists.partition((List<T>) delegate(), size));
         }
 
         return of(Iterables.partition(castAsIterable(), size));
@@ -244,34 +244,6 @@ public abstract class IterableView<T> extends ForwardingObject implements Iterab
     @SuppressWarnings("unchecked")
     private Iterable<T> castAsIterable() {
         return (Iterable<T>)delegate();
-    }
-
-    /**
-     * Casts the {@link #delegate()} as a {@link Collection} of T to avoid type
-     * warnings due to type erasure. This is safe if the delegate has already
-     * been checked to be an instance of {@link Collection}, but will throw if
-     * not a {@link Collection}
-     *
-     * @throws ClassCastException if the {@link #delegate()} is not a
-     *         {@link Collection}
-     */
-    @SuppressWarnings("unchecked")
-    private Collection<T> castAsCollection() {
-        return (Collection<T>) delegate();
-    }
-
-    /**
-     * Casts the {@link #delegate()} as a {@link List} of T to avoid type
-     * warnings due to type erasure. This is safe if the delegate has already
-     * been checked to be an instance of {@link List}, but will throw if
-     * not a {@link List}
-     *
-     * @throws ClassCastException if the {@link #delegate()} is not a
-     *         {@link List}
-     */
-    @SuppressWarnings("unchecked")
-    private List<T> castAsList() {
-        return (List<T>) delegate();
     }
 
 }
--- a/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
@@ -19,6 +19,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Spliterator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
```diff
@@ -212,6 +215,23 @@ public T getFirst() {
...
+
+    /**
+     * @return a sequential {@code Stream} with the contents as its source.
+     */
+    @SuppressWarnings("unchecked")
+    public Stream<T> stream() {
+        if (delegate() instanceof Collection) {
+            return ((Collection<T>) delegate()).stream();
+        }
+        return StreamSupport.stream(spliterator(), false);
```
j-baker: Given we have Guava here, please can we just do `return Streams.stream(delegate())` which is functionally equivalent?

atlasdb-commons/src/test/java/com/palantir/common/collect/IterableViewTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void listPartitionSpliterator() {
+        IterableView<List<String>> view = IterableView.of(ImmutableList.of("a", "b", "c")).partition(1);
+        Spliterator<List<String>> spliterator = view.spliterator();
+        assertThat(spliterator.estimateSize()).isEqualTo(3);
+        int characteristics = spliterator.characteristics();
+        assertThat(characteristics).isNotEqualTo(0);
+        assertThat(characteristics & Spliterator.CONCURRENT).isEqualTo(0);
+        assertThat(characteristics & Spliterator.DISTINCT).isEqualTo(0);
+        assertThat(characteristics & Spliterator.IMMUTABLE).isEqualTo(0);
```
j-baker: These tests seem fragile to changes elsewhere. For example, it would be super reasonable for Guava's Iterables.partition to claim that it's non-null (always) and immutable (given immutable input), and for that to change in some later version of Guava.

I would propose that you test this with a single test, which simply validates that the spliterator characteristics are identical if you call delegate().spliterator() vs spliterator()?

atlasdb-commons/src/test/java/com/palantir/common/collect/IterableViewTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertThat(characteristics & Spliterator.SORTED).isEqualTo(0);
+        assertThat(characteristics & Spliterator.SUBSIZED).isEqualTo(Spliterator.SUBSIZED);
+    }
+
+    @Test
+    public void setSpliterator() {
+        Spliterator<String> spliterator = IterableView.of(ImmutableSortedSet.of("a", "b")).spliterator();
+        assertThat(spliterator.estimateSize()).isEqualTo(2);
+        int characteristics = spliterator.characteristics();
+        assertThat(characteristics).isNotEqualTo(0);
```
j-baker: would be tempted to remove these isNotZero style asserts; they're redundant given the below constraints

atlasdb-commons/src/test/java/com/palantir/common/collect/IterableViewTest.java
j-baker: also, this is almost surely an oversight, but you aren't testing that the spliterators actually contain the right elements

atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
j-baker: I _think_ your unchecked casts are safe, having gone through all the cases. Annoyingly the ones below in castAsList and castAsCollection are not, since

```
List<Long> longs = Lists.newArrayList(1L);
List<Number> numbers = IterableView.of(longs).castAsList();
numbers.add(1.0);
assertThat(longs.stream().allMatch(Long.class::isInstance)).isTrue();
```

will sadly compile.

Do you fancy quickly changing them to wrap with Collections.unmodifiableCollection and Collections.unmodifiableList while you're modifying this class?

j-baker: bugs beget bugs etc; I'm ok with classes suppressing warnings iff every usage in that class is safe :)

atlasdb-commons/src/test/java/com/palantir/common/collect/IterableViewTest.java
schlosna: I'll trim them down to check that the returned spliterator has `SIZED` and `SUBSIZED` characteristics which are what we want to ensure that we can inherit sized optimizations that you don't get with a standard iterator based spliterator `java.util.Spliterators#spliteratorUnknownSize(java.util.Iterator<? extends T>, int)` from `java.lang.Iterable#spliterator`.

schlosna: I can't just assert that the spliterator's characteristics are identical to the delegate's as that doesn't always hold, e.g. for the transformed case of ImmutableList delegate, Guava does not currently propagate `IMMUTABLE` and `NONNULL` for a `com.google.common.collect.Lists$TransformingRandomAccessList`'s `spliterator`.

```
spliterator = {java.util.Spliterators$IteratorSpliterator@1192} 
 collection = {com.google.common.collect.Lists$TransformingRandomAccessList@1196}  size = 3
 it = {com.google.common.collect.Lists$TransformingRandomAccessList$1@1197} 
 characteristics = 16464 (0x4050)
 est = 3 (0x3)
 batch = 0 (0x0)
delegateSpliterator = {java.util.Spliterators$ArraySpliterator@1193} 
 array = {java.lang.Object[3]@1195} 
 index = 0 (0x0)
 fence = 3 (0x3)
 characteristics = 17744 (0x4550)
```

atlasdb-commons/src/test/java/com/palantir/common/collect/IterableViewTest.java
schlosna: actually left them in as the failure message if this assumption fails is more useful than the true/false of characteristic check.

j-baker: fair

atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java
j-baker: lame, wish guava had a better type here (streams aren't mutable so you should be able to make a ? extends T a stream of T

schlosna:
Thanks @j-baker !

PR #3704, opened 2019-01-07
[_transactions2] Coordination, Part 11b: PreStartTimestampHandler

jeremyk-91:
**Goals (and why)**: See #3695. The previous PRs caused issues because they would throw an exception when attempting to call `putUnlessExists(X, Y)` on the transaction service, where `X` is non-positive. However, there are cases where this can occur in normal operation.

In particular, the change would have broken sweep for any thoroughly swept table that used to be conservatively swept and had sentinels placed, or any thoroughly swept table after a Hard Delete.

- Negative timestamps are treated by the split key delegating service as normal timestamps, rather than specifically filtering them out.
- The `PreStartTimestampHandler` puts all negative timestamps on transactions 1, otherwise delegating to the regular `TransactionsSchemaManager`.

For reference, the legacy behaviour is to treat negative timestamps the same as positive timestamps, so something like

```
transactions.putUnlessExists(-5, 579);
assertThat(transactions.get(-5)).isEqualTo(579);
```

is expected to pass.

Updated old tests to reflect the new way we handle negatives, and included a broader test for the legacy behaviour in `TransactionServicesTest`.

- Eventually we probably want to drop support for transactions1; at that time, how should we handle this? Coordination Service doesn't like negative numbers.

**Where should we start reviewing?**: PreStartTimestampHandler

**Priority (whenever / two weeks / yesterday)**: ideally by tomorrow if that's reasonable, as this is needed to upgrade AtlasDB on the large internal product.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
@@ -27,7 +27,6 @@ import javax.annotation.CheckForNull;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
-import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.keyvalue.api.KeyAlreadyExistsException;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
@@ -37,7 +36,10 @@ import com.palantir.logsafe.exceptions.SafeIllegalStateException;
  * on which timestamps are requested. This class preserves the {@link TransactionService} guarantees regardless of
  * which underlying service is contacted.
  *
- * The timestamp service will throw an exception if the timestamp-to-service-key function returns a key which is
+ * The timestampToServiceKey function is expected to handle all possible long arguments (including negative values).
+ * The function may return null; if it does, then values written at that timestamp are considered to be uncommitted.
+ *
+ * The transaction service will also throw an exception if the timestamp-to-service-key function returns a key which is
  * not in the keyedServices map.
  *
  * Service keys are expected to be safe for logging.
@@ -56,9 +58,6 @@ public class SplitKeyDelegatingTransactionService<T> implements TransactionServi
     @CheckForNull
     @Override
     public Long get(long startTimestamp) {
-        if (startTimestamp < AtlasDbConstants.STARTING_TS) {
-            return null;
-        }
         return getServiceForTimestamp(startTimestamp).map(service -> service.get(startTimestamp)).orElse(null);
     }
 
@@ -66,9 +65,6 @@ public class SplitKeyDelegatingTransactionService<T> implements TransactionServi
     public Map<Long, Long> get(Iterable<Long> startTimestamps) {
         Multimap<T, Long> queryMap = HashMultimap.create();
         for (Long startTimestamp : startTimestamps) {
-            if (startTimestamp < AtlasDbConstants.STARTING_TS) {
-                continue;
-            }
             T mappedValue = timestampToServiceKey.apply(startTimestamp);
             if (mappedValue != null) {
                 queryMap.put(mappedValue, startTimestamp);
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
@@ -37,7 +37,8 @@ import com.palantir.logsafe.exceptions.SafeIllegalStateException;
  * which underlying service is contacted.
  *
  * The timestampToServiceKey function is expected to handle all possible long arguments (including negative values).
- * The function may return null; if it does, then values written at that timestamp are considered to be uncommitted.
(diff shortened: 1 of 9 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
```diff
@@ -37,7 +36,11 @@
  * on which timestamps are requested. This class preserves the {@link TransactionService} guarantees regardless of
  * which underlying service is contacted.
  *
- * The timestamp service will throw an exception if the timestamp-to-service-key function returns a key which is
+ * The timestampToServiceKey function is expected to handle all possible long arguments (including negative values).
+ * The function may return null; if it does, then for reads, values written at that timestamp are considered to be
+ * uncommitted. The transaction service will throw if a write is attempted at such a timestamp.
```
sandorw: Not sure I follow this comment - isn't this natural if I haven't yet committed my transaction?

jeremyk-91: This comment was to explain what the null values from #3694 mean, actually - since we have to allow for transaction managers that don't have access to a timestamp service. In this case, they are allowed to read, but may encounter values where coordination has not been achieved on the relevant start timestamp. These haven't committed (since to commit we need to putUnlessExists into the transactions service, which will achieve coordination on the start timestamp if needed), so the read-only services are allowed to recognise them as uncommitted.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/TransactionServices.java
```diff
@@ -42,10 +42,10 @@ private static TransactionService createSplitKeyTransactionService(
             KeyValueService keyValueService,
             CoordinationService<InternalSchemaMetadata> coordinationService) {
         TransactionSchemaManager transactionSchemaManager = new TransactionSchemaManager(coordinationService);
+        int versionOne = 1;
         return new SplitKeyDelegatingTransactionService<>(
-                transactionSchemaManager::getTransactionsSchemaVersion,
-                ImmutableMap.of(1, createV1TransactionService(keyValueService))
-        );
+                new PreStartTimestampHandler<>(versionOne, transactionSchemaManager::getTransactionsSchemaVersion),
+                ImmutableMap.of(versionOne, createV1TransactionService(keyValueService)));
```
jeremyk-91: - [ ] Add some explanation as to what is going on here / make this more descriptive

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SplitKeyDelegatingTransactionService.java
sandorw: Thanks - this was due to my misunderstanding of what this layer was doing (serving the version of the store, not the commit timestamp directly).

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractGetCandidateCellsForSweepingTest.java
```diff
@@ -179,6 +180,17 @@ public void startFromGivenRowThorough() {
                 .containsExactly(cell(2, 1), cell(2, 2), cell(3, 1), cell(3, 2));
     }
 
+    @Test
+    public void considersSentinelsForThorough() {
+        new TestDataBuilder()
+                .put(1, 1, -1L)
+                .put(1, 1, 5L)
+                .store();
+        CandidateCellForSweeping candidateCellForSweeping = Iterables.getOnlyElement(
```
sandorw: I don't think there's ever a case where a sentinel could be written but not 'committed' - is this correct? If that's the only case where this issue comes up, should we instead special case sentinels in the targeted sweep code to avoid the `TransactionService` lookups?

jeremyk-91: This is true in standard Atlas usage. I guess the broader question is whether we want to break the behaviour of a `TransactionService`, which previously accepted negative numbers. This may be reasonable, since internal tests and large internal product tests haven't failed otherwise

jeremyk-91: Writing of sentinels is nontransactionally done in sweep - it's done before deleting stuff, so there's no guarantee that we don't attempt to put sentinels multiple times, but it's guaranteed to happen at least once.

jeremyk-91: So I think whether we want to continue to support or be robust to negative accesses on transactions service is still a bit of an open question. I'm not currently aware of any other places in the codebase where we access negative timestamps, and as mentioned the tests are favourable.

I've fixed up the sweeper to behave properly here, its behaviour on encountering sentinels in the past was pretty silly, anyway:

- it sees `-1` as a normal timestamp
- so it queries the db for `-1`, and finds that the "transaction" at `-1` either has been "rolled back" with `-1` or hasn't been
- if it hasn't, it tries to rollback the `-1` transaction by putting `-1, -1` into the transactions1 table
- then it reads that cell as corresponding to an uncommitted transaction

Now we filter out `-1` starts, and put them back later on (if it's thorough and we need to consider them)

jeremyk-91:
block on #3709

PR #3392, opened 2018-07-19
FFT - _transactions2, Part 2: Preparing _transactions2 (no migration/usage yet!)

jeremyk-91:
**Goals (and why)**:
- Provide an implementation of the tickets algorithm for _transactions2, which avoids hotspotting
- Provide an implementation of a service-splitting transaction service, which will be essential for _transactions2

**Implementation Description (bullets)**:
- Refactor the _transactions1 transaction service to implement various translation methods of an abstract class
- Implement the tickets algorithm in _transactions2
- Implement a service-splitting service that switches on a long-predicate and hits the correct underlying service

**Testing (What was existing testing like?  What have you done to improve it?)**:
- Transactions1 testing is very well covered by e.g. `AbstractKeyValueServiceTransactionIntegrationTest`
- Additional testing was added for the ticket encoding/decoding methods, and the dynamic splitter

**Concerns (what feedback would you like?)**:
- Is the test coverage for _transactions2 good enough?
- The guarantee on the dynamic splitter's long predicate is a little strange, but I think we need that. Does it make sense?
- We need to be careful with the `TransactionKvsWrapper` as we move to transactions2 as well. It's used by large internal product.

**Where should we start reviewing?**: AbstractKeyValueServiceBackedTransactionService

**Priority (whenever / two weeks / yesterday)**: Early next week

This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/3392)

PR #3706, opened 2019-01-08
[_transactions2] Part 9: Write-Batching Transaction Service

jeremyk-91:
- [x] Improve testing of `WriteBatchingTransactionService`
- [x] Add docs detailing semantics of `WriteBatchingTransactionService`
- [x] Firm up contract of `TimestampEncodingStrategy`

**Goals (and why)**:
- Batch writes to the transaction service; eventually this means we will make few calls to `putUnlessExistsMultiple` (along the lines of RBTS and its `getFreshTimestamps`). This significantly reduces the Paxos overhead we will incur in Cassandra, thus improving performance.

- Introduce a notion of an encoding strategy for transaction timestamps - this is based loosely on the design in #3392 but uses composition instead, and also only has a v1 implementation here
- Refactor `SimpleTransactionService` to use a `V1EncodingStrategy`.
- Introduce a write batching transaction service that uses `DisruptorAutobatcher` to batch calls to `putUnlessExists`.

- Existing behaviour of the `SimpleTransactionService` is tested by the existing integration tests, like `CassandraKeyValueServiceTransactionIntegrationTest`. 
- The new `WriteBatchingTransactionService` has a few simple tests, though that's why this is a WIP as I want to write a few more. Nonetheless this is in a state that can plausibly be reviewed.

- Is there a way to keep `putUnlessExistsMultiple` to be mainly for internal usage?
- Will the `EncodingTransactionService` cause problems? It's not a very good abstraction but to be fair the `KeyAlreadyExistsExceptions` we're throwing already spit out details of how we encode values.

**Where should we start reviewing?**: `TimestampEncodingStrategy`


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/WriteBatchingTransactionService.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Semantics for batch processing:
+     *
+     * - If there are multiple requests to {@link TransactionService#putUnlessExists(long, long)} with the same
+     *   start timestamp, only the first of these will actually be attempted; the remainder will fail with a
+     *   {@link SafeIllegalArgumentException}, which is fine (following the contract it means the put may or may not
+     *   have happened).
+     * - If a {@link KeyAlreadyExistsException} is thrown, we fail out requests for keys present in the
```
gmaretic: I'd give a bit more detail on why we're retrying here, and possibly why we did not opt for alternative approaches

jeremyk-91:
Offline discussion. This has some idiosyncrasies on transactions1; let's suppose you have a batch of timestamps `[t1, t2, ... tX, ..., tN]` where `tX` is a transaction that has been read rollbacked. Now we fire off a `putUnlessExists` of `[t1, t2 ... tX-1, tX+1, ..., tN]`, but the first `X-1` of these, of course, were committed already, so we make another `X-1` futile requests.

Also, we need to think about how to handle key-value-services that throw `KeyAlreadyExistsException` but don't return the keys.

This is safer on transactions2 since we should get all existent columns when a `putUnlessExists` returns. It's possible that in the worst case we still need `O(N)` roundtrips (consider what happens if between each db call exactly one of the transactions with those values gets scheduled for rollback), and we might need a bit of handling to deal nicely with cases where the row you're writing to changes.

gmaretic:
I pushed the tests that caught this

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/service/WriteBatchingTransactionServiceTest.java
```diff
@@ -129,12 +130,15 @@ public void throwsOnUnspecifiedKeyAlreadyExistsExceptions() {
...
-                ENCODING_STRATEGY.encodeStartTimestampAsCell(3L)));
-        KeyAlreadyExistsException newExistingKeyException = new KeyAlreadyExistsException("boo", ImmutableList.of(
-                ENCODING_STRATEGY.encodeStartTimestampAsCell(2L)));
-
-        doThrow(originalException).doThrow(originalException)
+        KeyAlreadyExistsException originalException = new KeyAlreadyExistsException(
+                "boo",
+                ImmutableList.of(ENCODING_STRATEGY.encodeStartTimestampAsCell(3L)),
+                ImmutableList.of(ENCODING_STRATEGY.encodeStartTimestampAsCell(2L)));
+        KeyAlreadyExistsException newExistingKeyException = new KeyAlreadyExistsException(
```
gmaretic: This is now never going to be thrown because the list of timestamps will be empty (once success, one failure). Should probably add a third timestamp that neither succeeded nor failed on the first attempt and verify we retry for that one

jeremyk-91: Yep, makes sense - I added a third `elementToRetry`

jeremyk-91:
Needs NRN and a missed nit on `V1EncodingStrategy` now

gmaretic:
This is good to go whenever we are ready. @jeremyk-91 said he wanted to release first; I don't know if there are any other blockers

PR #3707, opened 2019-01-09
[_transactions2] Part 10: Tickets Encoding Strategy

jeremyk-91:
**Goals (and why)**:
- Implement the tickets encoding strategy, which is needed as a part of transactions2

- Implement the key functions (converting a timestamp to a cell, and a commit timestamp to a value, and vice versa)
- Reference the number of partitions from `TransactionConstants`

New code, new tests.

- What else is needed in terms of the tests?
- I revised the quantum down to 25M from 400M, proportional with the reduction in NP from 256 to 16. Is this reasonable?

**Where should we start reviewing?**: `TicketsEncodingStrategy`

**Priority (whenever / two weeks / yesterday)**: this week / early next week


jeremyk-91:
While we're here, I'd like to gauge appetite for introducing a hard limit on the gap between start and commit timestamp for transactions2, perhaps at `50,000,000` - the idea being that that makes clean-transaction-range type operations more tractable while still being resilient to a burst of leader elections.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/encoding/TicketsEncodingStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+        long columnComponent = decodeColumnName(cell.getColumnName());
+
+        return (rowComponent / ROWS_PER_QUANTUM) * PARTITIONING_QUANTUM
+                + columnComponent * ROWS_PER_QUANTUM
+                + rowComponent % ROWS_PER_QUANTUM;
+    }
+
+    @Override
+    public byte[] encodeCommitTimestampAsValue(long startTimestamp, long commitTimestamp) {
+        return TransactionConstants.getValueForTimestamp(commitTimestamp - startTimestamp);
```
jeremyk-91: No good - watch how negatives are handled

sandorw:
Is that number motivated by discussion with @tpetracca ? I'm not overly concerned with a lower number either, since normally a leader election would lose you your locks and prevent committing anyway (so you require leader elections between lock checks and the PUE).

jeremyk-91:
I don't think we discussed the precise number - the original idea I had in mind was just 'higher than leader election'. You're right, the leader election jump will only affect transactions where elections take place between lock checks and PUE.

We discussed this separately and I think we're not likely to lump this change to be part of transactions2; it might use the `CoordinationService` we've developed so far, but will probably be done as a separate layer. I think we want the limit to be configurable (and of course we need to remember a history of what value the limit has been in the past).

PR #3709, opened 2019-01-11
[_transactions2] Coordination, Part 11c: Formal Ban on Negative Timestamps

jeremyk-91:
**Goals (and why)**:
- Make a decision that simplifies #3704
- Formalise some of the assumptions we have about timestamps that aren't quite documented

- Firm up javadocs on `TransactionService` and `TimestampService` and a few implementations, to make it clear that negative numbers are a bad idea
- Remove the precheck on timestamps before submitting them to the transaction service
- Clean up the tests to reflect that we don't expect

Not much. This change strengthens some assumptions made in #3704, and thus removes tests that handled cases outside these assumptions.

- Is this a reasonable thing to do?

**Where should we start reviewing?**: TimestampService.java

**Priority (whenever / two weeks / yesterday)**: Today would be nice, if not early next week


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/service/TransactionService.java
```diff
@@ -34,6 +35,16 @@
  */
 public interface TransactionService {
     @CheckForNull
+    /**
+     * Gets the commit timestamp associated with a given start timestamp.
+     * This may be cached on the client-side, if desired.
```
sandorw: Non-null responses may be cached

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/PreStartHandlingTransactionService.java
```diff
@@ -0,0 +1,84 @@
...
+        Map<Long, Long> result = Maps.newHashMap();
+        result.putAll(delegate.get(classifiedTimestamps.get(true)));
+        result.putAll(Maps.uniqueIndex(classifiedTimestamps.get(false), unused -> AtlasDbConstants.STARTING_TS - 1));
+        return result;
+    }
+
+    @Override
+    public void putUnlessExists(long startTimestamp, long commitTimestamp) throws KeyAlreadyExistsException {
+        if (!isTimestampValid(startTimestamp)) {
+            throw new SafeIllegalStateException("Attempted to putUnlessExists({}, {}) which is unexpected.",
```
sandorw: 'disallowed' over 'unexpected' to be clearer that this is a bad thing to do, not just a case we didn't consider

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/PreStartHandlingTransactionService.java
```diff
@@ -0,0 +1,84 @@
...
+                .collect(Collectors.partitioningBy(PreStartHandlingTransactionService::isTimestampValid));
+
+        Map<Long, Long> result = Maps.newHashMap();
+        result.putAll(delegate.get(classifiedTimestamps.get(true)));
+        result.putAll(Maps.uniqueIndex(classifiedTimestamps.get(false), unused -> AtlasDbConstants.STARTING_TS - 1));
+        return result;
+    }
+
+    @Override
+    public void putUnlessExists(long startTimestamp, long commitTimestamp) throws KeyAlreadyExistsException {
```
sandorw: Add some simple tests for the new behavior, I don't see that at the moment

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/service/TransactionService.java
jeremyk-91: Yep, this is important! Fixed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/PreStartHandlingTransactionService.java
jeremyk-91: :+1: changed to disallowed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/PreStartHandlingTransactionService.java
jeremyk-91: Yep, will do - thanks for flagging this.

PR #3710, opened 2019-01-14
add debug/stats CLI for targeted sweep queue

clockfort:
Adds a debug/statistics gathering CLI useful for users (me) looking to rewrite their schema to work better with targeted sweep

e.g. finding good candidates for tables to possibly rewrite the products' interactions with in order to move these tables to non-conservative (liberal? 😉 ) sweep strategies

This is in order to further reduce the amount of bloat that heavy writes into conservative tables cause in environments with large transaction timeouts, where they have to stick around for a while in the queue.

Extremely specifically, the number one thing I want is less pressure on various oracle high water marks for people with oracle who don't have EE edition and the online compaction it offers. Tables like the sweep queue table end up growing to enormous sizes during initial day-1 life-of-a-stack ingests where most of the initial data comes in, and the space can never be freed, as there's no way a manual compaction would ever be able to grab a table lock on a table like the targeted sweep queue, which is constantly receiving new writes every atlas txn.

### DIFF
--- a/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/debug/ComputeSweepStats.java
+++ b/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/debug/ComputeSweepStats.java
@@ -1,0 +1,95 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.cli.command.debug;
+
+import java.util.Map;
+
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+import com.palantir.atlasdb.cli.command.SingleBackendCommand;
+import com.palantir.atlasdb.cli.output.OutputPrinter;
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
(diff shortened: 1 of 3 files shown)


atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/debug/ComputeSweepStats.java
```diff
@@ -0,0 +1,95 @@
...
+                            updateStats(tableRef, metadata.conservative());
+                        });
+                    });
+            return 0;
+        });
+        printStats();
+        return 0;
+    }
+
+    private void updateStats(TableReference tableRef, boolean isConservative) {
```
sandorw: Note that dedicated rows will not have a real value for the `WriteReference` in the non-dedicated row. I've not checked what is written, though.

sandorw:
On the topic of raising the high water mark, we made some improvements late last year to more quickly delete data from bursty write workflows (= clean up dedicated rows faster than we were before). After that, I don't think there's a ton of room to tune table design to targeted sweep:

- (most important) use thorough sweep
- keep the immutable timestamp moving forward globally
- batch things in reasonably sized transactions

The size of the queue is roughly the number of cells written by transactions in the period that targeted sweep is behind (immutable ts moving + the time to do the work).

clockfort:
> On the topic of raising the high water mark, we made some improvements late last year to more quickly delete data from bursty write workflows (= clean up dedicated rows faster than we were before). After that, I don't think there's a ton of room to tune table design to targeted sweep:
> 
> * (most important) use thorough sweep

Yeah, this cli is to try and find good candidates for tables to move to thorough sweeping.

> * keep the immutable timestamp moving forward globally

I also just committed a healthcheck for this in our product yesterday.

> * batch things in reasonably sized transactions

Right now it looks like our worst-sized txns are just lazy people doing horrible single transaction migration jobs. Our product is also getting a framework similar to the one you wrote that lets people write migrations and deal with a live transition period of two active schemas more easily.

> The size of the queue is roughly the number of cells written by transactions in the period that targeted sweep is behind (immutable ts moving + the time to do the work).

Yeah, the problem I'm seeing is that stacks tend to grow hundreds of GB on their first day alive and the HWMs get stuck real big.

gmaretic:
For one time isolated ingests of new data with a single version, you could disable the targeted sweep queue and avoid writing into it altogether. I.e. if you are doing a massive ingest in a single transaction, and not overwriting anything

gmaretic:
Is this something that we still want?

clockfort:
¯\_(ツ)_/¯

doesn't matter to me, I used it for my own purposes already in my own project.

PR #3713, opened 2019-01-15
[_transactions2] Part 12: Finer-grained CAS compatibility levels for KVSes

jeremyk-91:
Collaboration with @gmaretic 

**Goals (and why)**:
- Our strategy for handling batches in #3706 is significantly better with knowledge of _which_ keys failed in a given batched putUnlessExists operation. This isn't supported by all KVS (in particular DbKVS doesn't currently support it), and implementing it in DbKVS seems nontrivial and a bit hacky as it looks like we'll have to parse exception messages.
- Have a clear way of ensuring that batching is only done with services that are friendly to it.

- Define 3 possible levels of CAS support: none, supports with no detail on failure besides "I failed", and supports with details on failures (i.e. "I failed because these rows are there" for PUE, "I failed and the current value is X" for CAS).
- Have existing KVSes give their level of support (Cassandra has support-with-failures, Memory and DbKVS are support-no-detail-on-failures).

Mostly new code, which is tested. We added tests for the CAS Compatibility method, as well as for KVSes where the delegation is not completely straightforward.

- This is another devbreak 😞 but I think doing this this way is better than automatically inferring a `SUPPORTED_NO_DETAIL` from the already existing `supportsCheckAndSet()` method.

**Where should we start reviewing?**: CheckAndSetCompatibility

**Priority (whenever / two weeks / yesterday)**: tomorrow? This blocks #3706

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CheckAndSetCompatibility.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CheckAndSetCompatibility.java
@@ -16,6 +16,10 @@
 
 package com.palantir.atlasdb.keyvalue.api;
 
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 /**
  * Indicates whether a {@link KeyValueService} supports check and set (CAS) and put unless exists (PUE) operations, and
  * if so the granularity with which it can provide feedback.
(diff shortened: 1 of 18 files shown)


jeremyk-91:
Sanity checked the high-level idea with @sandorw - it's possible we may not need this in the future after dbkvs and in-memory play catch-up, and there are probably other ways to control what KVSes this is created with, but this can be used to ensure that no one accidentally tries to create things in the bad state and is easy to reverse.

PR #3715, opened 2019-01-17
fix javadoc for WRITE_WRITE handler

laflechejonathan:
**Goals (and why)**:
Currently, the description for WRITE_WRITE conflict handling makes it sound like the same behaviour as WRITE_WRITE_CELL. `s/cell/row` gives a more accurate description.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
@@ -23,7 +23,7 @@ public enum ConflictHandler {
     IGNORE_ALL(false, false, false, false),
 
     /**
-     * If two transactions concurrently write to the same cell the one that commits later will
+     * If two transactions concurrently write to the same row the one that commits later will
      * throw a {@link TransactionConflictException} and should have details about how this happened.
      */
     RETRY_ON_WRITE_WRITE(false, true, true, false),


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -23,7 +23,7 @@
     IGNORE_ALL(false, false, false, false),
 
     /**
-     * If two transactions concurrently write to the same cell the one that commits later will
+     * If two transactions concurrently write to the same row the one that commits later will
```
j-baker: This is wrong; Atlas will always retry on the cell cahnging not the row

j-baker: Basically, there are two things at play here:

1. The granularity of your locking: do you lock on the row or the cell?
2. The types of conflicts you will see.

Your comment affects 2, but write write and write write cell are identical from that perspective.

laflechejonathan: not sure I understand the distinction.

To make this concrete, say I have:

```
row("id", uuid)
column("value_a", int)
column("value_b", int)

Transaction X
put(id=1, value_a=4)

Transaction Y
put(id=1, value_b=10)
```

What is the behaviour of WRITE_WRITE and WRITE_WRITE_CELL? Do neither of them conflict?

sandorw: WRITE_WRITE: X and Y block on each other via a row lock but don't/can't conflict (conflict checks are always at the cell level)

WRITE_WRITE_CELL: X and Y can run concurrently and can't conflict

laflechejonathan:
Interesting, I'll take another stab at fixing these docs when I get a chance

laflechejonathan:
In what circumstance can write_write encounter conflicts then? Never?

j-baker:
Atlas has a notion of a row and a notion of a cell. The difference between WRITE_WRITE and WRITE_WRITE_CELL is the locking granularity.

To give an example, if I have a table with row component A and columns B and C, with WRITE_WRITE then a transaction which writes to column B will take a lock on A when committing, and a transaction which writes to C will also take a lock on A. There are no conflicts that can happen here; you wrote to different things. But, if the writes happen concurrently, you will see _contention_ for the same lock.

The advantage is that if you write to 10 columns in that row, you only need to take out one lock.

If you have WRITE_WRITE_CELL meanwhile, the lock when writing will be on (A, B) and (A, C) - there is no contention.

The conflicts will be the same, though.

PR #3723, opened 2019-01-23
client-side lock leases

gokcanozakdag:
**Goals (and why)**:
Faster transactions with lower Timelock load.

Timelock now responds with a lease period to lock requests. Client caches the received locks, and checks validity by checking the expiration date before asking it to Timelock.
To reduce the amount of code change, and keep lock leases transparent from clients, I have split the client side and server side `TimelockService` interface.

Currently basic unit tests for `LeasingTimelockClient` and `LockLeaseManager`. I will add more comprehensive tests.


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -792,8 +794,13 @@ private static LockAndTimestampServices getLockAndTimestampServices(
...
         LockService lockService = new ServiceCreator<>(metricsManager, LockService.class, userAgent)
                 .applyDynamic(timelockServerListConfig);
-        TimelockService timelockService = new ServiceCreator<>(metricsManager, TimelockService.class, userAgent)
-                .applyDynamic(timelockServerListConfig);
+
+        TimelockServerInterface rawTimelockService = new ServiceCreator<>(metricsManager,
+                TimelockServerInterface.class,
+                userAgent).applyDynamic(timelockServerListConfig);
+
+        TimelockService timelockService = LeasingTimelockClient.create(rawTimelockService);
```
gokcanozakdag: I will open a separate pr to actually use the leasing client. So we can merge this first, release Timelock, then merge the pr that uses the leasing client and declares a dependency to correct version of Timelock.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -792,8 +794,13 @@ private static LockAndTimestampServices getLockAndTimestampServices(
             String userAgent) {
         LockService lockService = new ServiceCreator<>(metricsManager, LockService.class, userAgent)
                 .applyDynamic(timelockServerListConfig);
-        TimelockService timelockService = new ServiceCreator<>(metricsManager, TimelockService.class, userAgent)
-                .applyDynamic(timelockServerListConfig);
+
+        TimelockServerInterface rawTimelockService = new ServiceCreator<>(metricsManager,
```
j-baker: maybe worth calling this TimelockRpcs or something like that? TimelockServerInterface seems a bit fixable...

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLockRequest request) {
+        return delegate.startAtlasDbTransaction(request);
+    }
+
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction(
+            StartIdentifiedAtlasDbTransactionRequest request) {
+        long startTime = System.nanoTime();
```
j-baker: please can you use/factor out the NanoTime class I use in the server side part? NanoTime comparisons are tricky and I'd like to centralize it to one place.

j-baker: (basically: the numberline can wrap in nanotime, this makes things tricky)

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public long getImmutableTimestamp() {
+        return delegate.getImmutableTimestamp();
+    }
+
+    @Override
+    public LockResponse lock(LockRequest request) {
+        long startTime = System.nanoTime();
```
j-baker: same here; NanoTimes, plz

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,154 @@
...
+        LeasableLockResponse leasableResponse = delegate.leasableLock(request);
```
j-baker: again, this could do better probably. at a minimum, could just call it lock and have it take a LeaseableLockRequest (which has a `@JsonValue` of LockRequest)

lock-api/src/main/java/com/palantir/lock/client/LockLeaseManager.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.annotations.VisibleForTesting;
+import com.palantir.lock.v2.LockToken;
+
+public class LockLeaseManager {
+    private final Supplier<Long> clock;
```
j-baker: `Supplier<NanoTime>`

lock-api/src/main/java/com/palantir/lock/client/LockLeaseManager.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static LockLeaseManager create() {
+        return new LockLeaseManager(System::nanoTime, Duration.ofSeconds(1));
+    }
+
+    @VisibleForTesting
+    LockLeaseManager(Supplier<Long> clock, Duration leaseExpiry) {
+        this.clock = clock;
+        this.leasedTokens = Caffeine.newBuilder()
+                .expireAfterWrite(leaseExpiry.toNanos(), TimeUnit.NANOSECONDS)
```
j-baker: I wouldn't use the Duration for this; as we discussed last week you probably want .weakKeys() (https://www.baeldung.com/java-weak-reference)

j-baker: actually, for the perfect solution you probably want Caffeine's programmable '.expiry' method

j-baker: so you can couple it to the actual expiry time

lock-api/src/main/java/com/palantir/lock/client/LockLeaseManager.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void updateLease(LockToken lockToken, long expiry) {
+        leasedTokens.put(lockToken, expiry);
+    }
+
+    public void invalidate(LockToken lockToken) {
+        leasedTokens.invalidate(lockToken);
+    }
+
+    public boolean isValid(LockToken lockToken) {
```
j-baker: i got this far in the review :) rest is for tomorrow

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public Set<LockToken> refreshLockLeases(Set<LockToken> tokens) {
+        Set<LockToken> validByLease = tokens.stream()
+                .filter(lockLeaseManager::isValid)
+                .collect(Collectors.toSet());
+
+        Set<LockToken> toRefresh = Sets.difference(tokens, validByLease);
+
+        long startTime = System.nanoTime();
+        LeasableRefreshLockResponse refreshLockResponse = delegate.leasableRefreshLockLeases(toRefresh);
```
sandorw: Some tokens that were valid due to leases may have expired by this point, given the potential length of the RPC, but that doesn't matter given our protocol (one could always have lost the locks immediately after having verified that they were held)

lock-api/src/main/java/com/palantir/lock/client/LockLeaseManager.java
gokcanozakdag: Actually simply calling .expireAfterWrite and .weakKeys on builder combines these two - I did a simple test to verify that.

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public LockResponse lock(LockRequest request) {
+        long startTime = System.nanoTime();
+        LeasableLockResponse leasableResponse = delegate.leasableLock(request);
+
+        LockResponse lockResponse = leasableResponse.getLockResponse();
+        Optional<Duration> leasePeriod = leasableResponse.getLeasePeriod();
+
+        if (lockResponse.wasSuccessful()) {
```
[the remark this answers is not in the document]
gokcanozakdag: you are right... we will need to keep both the links from real token to generated one and generated to real one. That would at least require one more weakHashMap, I believe. There may be some tricks to avoid that though, let's sync on that offline.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -110,9 +110,9 @@
     private static final long EMBEDDED_BOUND = 3;
 
     private static final String TIMELOCK_SERVICE_FRESH_TIMESTAMP_METRIC =
-            MetricRegistry.name(TimelockService.class, "getFreshTimestamp");
+            MetricRegistry.name(TimelockRpcClient.class, "getFreshTimestamp");
```
gokcanozakdag: we will be breaking the metrics dashboard for client side - shall we put extra effort to keep metric name same by differing the class name and metric name? or would it cause more problems?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbFeignTargetFactory.java
```diff
@@ -49,7 +51,9 @@
             10_000, 65_000);
 
     private static final ObjectMapper mapper = new ObjectMapper()
-            .registerModule(new Jdk8Module());
+            .registerModule(new Jdk8Module())
+            .registerModule(new JavaTimeModule())
+            .registerModule(new GuavaModule());
```
gokcanozakdag: JavaTimeModule is required for `Duration` in `Lease`, and GuavaModule is needed to deserialize immutableSet in `LeasableRefreshLockResponse`

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public LockImmutableTimestampResponse lockImmutableTimestamp(IdentifiedTimeLockRequest request) {
+        return delegate.lockImmutableTimestamp(request);
+    }
+
+    @Override
+    public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLockRequest request) {
+        return delegate.startAtlasDbTransaction(request);
+    }
```
gokcanozakdag: We need to either remove lockImmutableTimestamp and startAtlasDbTransaction or return leasedLocks to prevent class cast exceptions

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LeaseExpirationTimer.java
```diff
@@ -15,26 +15,33 @@
...
 
 public class LeaseExpirationTimer {
 
-    public static final long LEASE_TIMEOUT_MILLIS = 20_000;
+    public static final Duration LEASE_TIMEOUT = Duration.ofMillis(20_000);
 
-    private volatile long lastRefreshTimeMillis;
-    private final Clock clock;
+    private final AtomicReference<NanoTime> lastRefreshTime;
+    private final Supplier<NanoTime> clock;
```
gokcanozakdag: using nanotime here also, to have the same clock source for both publish lease contracts and expiring them.

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
```diff
@@ -14,19 +14,25 @@
...
 import com.google.common.annotations.VisibleForTesting;
 
 public final class NanoTime implements Comparable<NanoTime> {
+    @JsonProperty("time")
     private final long time;
 
-    @VisibleForTesting
-    NanoTime(long time) {
+    @JsonCreator
+    public NanoTime(@JsonProperty("time") long time) {
```
j-baker: I would omit the `@JsonProperty("time")` and just always serialize a nanotime as a long?

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
```diff
@@ -14,19 +14,25 @@
...
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.annotations.VisibleForTesting;
 
 public final class NanoTime implements Comparable<NanoTime> {
+    @JsonProperty("time")
```
j-baker: `@JsonValue`?

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
j-baker: I somehow get the impression that prod code should only be creating this by calling NanoTime.now(). How about letting this be called only via a method `createForTests`?

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
j-baker: nah; we can roll this out real quick and then it'll be the case for the whole fleet

j-baker: but i guess, given that this is just a string... why change it?

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+import com.palantir.lock.v2.IdentifiedTime;
+import com.palantir.lock.v2.Lease;
+import com.palantir.lock.v2.LockToken;
+
+public final class LeasedLockToken implements LockToken {
+    private final LockToken serverToken;
+    private final LockToken clientToken;
+
+    private Lease lease;
+    private boolean inValidated = false;
```
j-baker: invalidated is probably just all lower case?

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+        inValidated = true;
+    }
+
+    @Override
+    public UUID getRequestId() {
+        return clientToken.getRequestId();
+    }
+
+    @Override
+    public int hashCode() {
```
j-baker: could probably do with a toString as well...

j-baker: but here: I somehow get the impression that all you care about is identity and so implementing hashCode and equals is unnecessary? Like, pretty sure that given the code that exists here (which ensures that clientToken is unique per object) your equals method is equivalent to just ==?

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+package com.palantir.lock.client;
+
+import java.util.Objects;
+import java.util.UUID;
+
+import com.palantir.lock.v2.IdentifiedTime;
+import com.palantir.lock.v2.Lease;
+import com.palantir.lock.v2.LockToken;
+
+public final class LeasedLockToken implements LockToken {
```
j-baker: think this class should try to be package private. Like, you shouldn't even really be able to reference LeasedLockToken outside of this package

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+    private final LockToken serverToken;
+    private final LockToken clientToken;
+
+    private Lease lease;
```
j-baker: can we add `@GuardedBy("this")` annotations to these two fields? It'll cause Error-prone to throw if we do an unsafe access to them.

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
j-baker: as in, something is invalidated, not in-validated

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Objects;
+import java.util.UUID;
+
+import com.palantir.lock.v2.IdentifiedTime;
+import com.palantir.lock.v2.Lease;
+import com.palantir.lock.v2.LockToken;
+
+public final class LeasedLockToken implements LockToken {
+    private final LockToken serverToken;
+    private final LockToken clientToken;
```
j-baker: I'd probably just have this be `UUID requestId`, it's not like it's ever a lock token that you can use as a lock token...

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+    LockToken serverToken() {
+        return serverToken;
+    }
+
+    synchronized Lease getLease() {
+        return lease;
+    }
+
+    synchronized void updateLease(Lease newLease) {
+        this.lease = newLease;
```
j-baker: there's probably some precondition where like, we should only be updating the lease if it's advantageous to do so? E.g. if someone tries to update it into the past, that's probably not great

j-baker: also, we should probably do the quick check that the lease we're updating it to has the same leadership time as the lease we updated from.

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
j-baker: can't we write lockImmutableTimestamp in terms of startIdentifiedAtlasDbTransaction?

j-baker: removing the legacy startAtlasDbTransaction method seems sensible; we don't ever use it and it's clearly scoped to AtlasDB; but locking the immutable timetamp is something people sometimes do

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,160 @@
...
+        return ImmutableLockResponse.of(Optional.of(leasedLockToken));
+    }
+
+    @Override
+    public WaitForLocksResponse waitForLocks(WaitForLocksRequest request) {
+        return delegate.waitForLocks(request);
+    }
+
+    @Override
+    public Set<LockToken> refreshLockLeases(Set<LockToken> tokens) {
```
j-baker: better call this 'uncastedTokens'; the variable name 'tokens' doesn't play well with 'allTokens', doesn't really describe why they're different.

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,160 @@
...
+        IdentifiedTime identifiedTime = delegate.getLeaderTime();
+        Set<LeasedLockToken> allTokens = tokens.stream()
```
j-baker: You shouldn't need to copy this set in order to do this. Rather, I'd do something like:

```
checkArgument(tokens.stream().allMatch(token -> token instanceOf LeasedLockToken), ...);
Set<LeasedLockToken> tokens = (Set<LeasedLockToken>) (Set<?>) tokens;
```

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,160 @@
...
+
+            refreshedTokens.forEach(t -> t.updateLease(lease));
+        }
+
+        return Sets.union(refreshedTokens, validByLease);
+    }
+
+    @Override
+    public Set<LockToken> unlock(Set<LockToken> tokens) {
+        Set<LeasedLockToken> leasedLockTokens = tokens.stream()
```
j-baker: this snippet feels like it should be a static method that you factor out and can use here and in the above method, like 'private static Set<LeasedLockToken> leasedTokens(Set<LockToken> tokens)`

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,160 @@
...
+                .filter(token -> token.isValid(identifiedTime))
+                .collect(Collectors.toSet());
+
+        Set<LeasedLockToken> toRefresh = Sets.difference(allTokens, validByLease);
+
+        Set<LeasedLockToken> refreshedTokens = ImmutableSet.of();
+
+        if (!toRefresh.isEmpty()) {
+            LeasableRefreshLockResponse refreshLockResponse = delegate.leasableRefreshLockLeases(
+                    toRefresh.stream().map(LeasedLockToken::serverToken).collect(Collectors.toSet()));
```
j-baker: this snippet could again be a method `serverTokens(Set<LeasedLockToken> leasedTokens)`

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * A version of {@link TimelockService#unlock(Set)} where one does not need to know whether the locks associated
+     * with the provided tokens were successfully unlocked or not.
+     *
+     * In some implementations, this may be more performant than a standard unlock.
+     *
+     * @param tokens Tokens for which associated locks should be unlocked.
+     */
+    default void tryUnlock(Set<LockToken> tokens) {
```
j-baker: we should kill this method now; it's not implemented here

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
```diff
@@ -0,0 +1,153 @@
...
+import com.palantir.lock.v2.WaitForLocksRequest;
+import com.palantir.lock.v2.WaitForLocksResponse;
+import com.palantir.logsafe.Safe;
+import com.palantir.processors.AutoDelegate;
+import com.palantir.timestamp.TimestampRange;
+
+@Path("/timelock")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@AutoDelegate
```
j-baker: think we don't want AutoDelegate for this anymore

lock-api/src/main/java/com/palantir/lock/v2/LeasableStartIdentifiedAtlasDbTransactionResponse.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableLeasableStartIdentifiedAtlasDbTransactionResponse.class)
+@JsonDeserialize(as = ImmutableLeasableStartIdentifiedAtlasDbTransactionResponse.class)
+public interface LeasableStartIdentifiedAtlasDbTransactionResponse {
+    @Value.Parameter
+    StartIdentifiedAtlasDbTransactionResponse getStartTransactionResponse();
```
j-baker: I wouldn't reference the old response here (it sets a dubious precedent around wrapping; the V10 response shouldn't contain all V9, V8, ... responses. Also, it's a bunch of JSON bloat (startTransactionResponse is like 30 bytes of text. I'd prefer just listing the fields that show up.

lock-api/src/main/java/com/palantir/lock/v2/Lease.java
```diff
@@ -0,0 +1,60 @@
...
+@JsonDeserialize(as = ImmutableLease.class)
+public abstract class Lease {
+    @Value.Parameter
+    public abstract UUID leaseOwnerId();
+
+    @Value.Parameter
+    public abstract NanoTime startTime();
+
+    @Value.Parameter
+    public abstract Duration period();
```
j-baker: 'period' isn't quite right here. I'd be tempted to go for something like 'validity'?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java
```diff
@@ -128,12 +135,52 @@ public void waitForLocks(@Suspended final AsyncResponse response, WaitForLocksRe
...
+    @POST
+    @Path("leasable-refresh-locks")
+    public LeasableRefreshLockResponse leasableRefreshLockLeases(Set<LockToken> tokens) {
+        return timelock.leasableRefreshLockLeases(tokens);
+    }
+
+    @POST
+    @Path("leasable-lock")
+    public void leasableLock(@Suspended final AsyncResponse response, LockRequest request) {
+        IdentifiedTime identifiedStartTime = timelock.identifiedTime();
```
j-baker: recently you made some perf improvements specifically around avoiding multiple calls to the AsyncTimelockService, because you have to do a leadership check every time. This feels like it avoids that :P

j-baker: I also don't think this is the right time to take this time; at a minimum what if it takes a few seconds to actually wait for the locks to be released. I expect that this time should be calculated as part of constructing the HeldLocksCollection and merely accessed here? Might need some refactoring, maybe split the server side changes into a separate PR since this one is already big? Also, because they have to be released separately this PR cannot be merged as-is; it requires the server side changes to be merged, released internally, and then a version number updated with the client side changes; I know you know this but just wanted to make more argument for server side changes being released separately.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockService.java
```diff
@@ -50,4 +53,11 @@
     StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction(
             StartIdentifiedAtlasDbTransactionRequest request);
 
+    LeasableRefreshLockResponse leasableRefreshLockLeases(Set<LockToken> tokens);
+
+    LeasableStartIdentifiedAtlasDbTransactionResponse leasableStartIdentifiedAtlasDbTransaction(
```
j-baker: Within the AsyncTimelockService, we shouldn't have to maintain backcompat; this means that these methods should ideally be replacing some current implementations (other than IdentifiedTime, which is new).

Put differently, refreshLockLeases should be defined in terms of leasableRefreshLockLeases, and probably only at the AsyncTimelockResource level (not here).

Same for the startTransaction one; there should simply not be a non-leaseable startIdentifiedAtlasDbTransaction at this part of timelock; it should be stripped out on the way to the client at the level above.

The method here should just be 'refreshLockLeases'.

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * @param tokens Tokens for which associated locks should be unlocked.
+     */
+    default void tryUnlock(Set<LockToken> tokens) {
+        unlock(tokens);
+    }
+
+    @GET
+    @Path("leader-time")
+    IdentifiedTime getLeaderTime();
```
j-baker: it's worth a test that this changes every leadership election (on the same server), because this is part of the contract of timelock

j-baker: as in, if leadership goes A -> B -> A, A should have a different id the second time.

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+    @Test
+    public void lockResponeHasCorrectLeasedLock() {
+        LockRequest lockRequest = mock(LockRequest.class);
+        Lease lease = getLease();
+        when(timelockService.leasableLock(lockRequest)).thenReturn(
+                LeasableLockResponse.of(LOCK_RESPONSE, lease));
+
+        LockResponse clientResponse = timelockClient.lock(lockRequest);
+        LeasedLockToken leasedLockToken = (LeasedLockToken) clientResponse.getToken();
+        assertEquals(LOCK_TOKEN, leasedLockToken.serverToken());
```
j-baker: you should never use assertEquals, it's too error-prone (people flip expected and actual all the time). Prefer using assertThat() every time.

j-baker: (so that's a comment about every assertEquals in these files)

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+    @Test
+    public void delegatesStartAtlasdbTransactionRequest() {
+        StartIdentifiedAtlasDbTransactionRequest request = mock(StartIdentifiedAtlasDbTransactionRequest.class);
+        StartIdentifiedAtlasDbTransactionResponse response = startTransactionResponseWith(LOCK_TOKEN);
+        when(timelockService.leasableStartIdentifiedAtlasDbTransaction(request)).thenReturn(
+                LeasableStartIdentifiedAtlasDbTransactionResponse.of(response, getLease()));
+
+        timelockClient.startIdentifiedAtlasDbTransaction(request);
+
+        verify(timelockService).leasableStartIdentifiedAtlasDbTransaction(request);
```
j-baker: these kinds of verifications are largely fragile; they don't actually really test anything (you don't care that the method was called, you care if it returns the right response) and easily break.

In this case, you should be delegating that the result is as you expect, rather than verifying that the method was right

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+        StartIdentifiedAtlasDbTransactionResponse clientResponse =
+                timelockClient.startIdentifiedAtlasDbTransaction(request);
+
+        LeasedLockToken leasedLock = (LeasedLockToken) clientResponse.immutableTimestamp().getLock();
+        assertEquals(leasedLock.serverToken(), LOCK_TOKEN);
+        assertEquals(lease, leasedLock.getLease());
+    }
+
+    @Test
```
j-baker: you should have tests for the 'I asked for locks and didn't get them' cases

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Before
+    public void setUp() {
+        when(timelockService.getLeaderTime()).thenReturn(IdentifiedTime.of(UUID.randomUUID(), NanoTime.now()));
+        timelockClient = LeasingTimelockClient.create(timelockService);
+    }
+
+    @Test
+    public void delegatesLockRequest() {
+        LockRequest lockRequest = mock(LockRequest.class);
```
j-baker: don't use the `mock()` static method, instead add `@RunWith(MockitoJUnitRunner.class)` to the whole class and create your mocks as fields, like

```
@Mock private LockRequest lockRequest;
```

why? because Mockito will tell you if you're mocking things unnecessarily (i.e. that don't get called), which is usually a sign of bugs.

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+    private TimelockRpcClient timelockService = mock(TimelockRpcClient.class);
+    private TimelockService timelockClient;
+
+    private static final UUID LEADER_ID = UUID.randomUUID();
+
+    private static final LockToken LOCK_TOKEN = LockToken.of(UUID.randomUUID());
+    private static final LockResponse LOCK_RESPONSE = LockResponse.successful(LOCK_TOKEN);
+
+    @Before
+    public void setUp() {
```
j-baker: prefer 'before()' to 'setUp' as your method name; Baseline actually blacklists 'setUp'

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+        when(timelockService.getLeaderTime()).thenReturn(IdentifiedTime.of(UUID.randomUUID(), NanoTime.now()));
```
j-baker: I don't think you'll have a good time with NanoTime.now(); it will surely cause you issues when it comes to invalidation, right? As in, I think in these tests you have no test for a lock lease which was ever valid, your tests all test lock leases that are either always valid or never valid, there's no notion of expiry.

These kinds of tests (while they are seemingly reasonable) are the sorts of things which lead to untested bugs in general.

In cases like this, I recommend factoring out your time source (into a `Supplier<NanoTime>` most likely) and testing that explicitly.

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+        when(timelockService.leasableStartIdentifiedAtlasDbTransaction(request)).thenReturn(
+                LeasableStartIdentifiedAtlasDbTransactionResponse.of(response, getLease()));
+
+        timelockClient.startIdentifiedAtlasDbTransaction(request);
+
+        verify(timelockService).leasableStartIdentifiedAtlasDbTransaction(request);
+    }
+
+    @Test
+    public void startAtlasdbTransactionResponseHasCorrectLeasedLock() {
```
j-baker: this test seems to make the test above redundant (delegatesStartAtlasdbTransactionRequest seems pointless if you have this test)

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertValid(refreshedLeasedLockToken);
+        assertValid(leasedLockToken);
+        assertEquals(leasedLockToken, refreshedLeasedLockToken);
+    }
+
+    private StartIdentifiedAtlasDbTransactionResponse startTransactionResponseWith(LockToken lockToken) {
+        return StartIdentifiedAtlasDbTransactionResponse.of(
+                LockImmutableTimestampResponse.of(1L, lockToken),
+                mock(TimestampAndPartition.class)
+        );
```
j-baker: prefer not putting braces on the next line; style within AtlasDB (and at Palantir in general) to close braces all at once.

j-baker: same in the other cases

PR #3724, opened 2019-01-25
[no changelog] Remove deprecated commons-lang 2.6

cakofony:
**Goals (and why)**:

commons-lang 2.6 is deprecated and should not be used. Functionality is replaced by commons-lang3.


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/TypeAndName.java
```diff
@@ -15,7 +15,7 @@
  */
 package com.palantir.atlasdb.table.description.render;
 
-import org.apache.commons.lang.Validate;
+import org.apache.commons.lang3.Validate;
```
cakofony: Oops, thanks!

PR #3728, opened 2019-01-31
[_transactions2] Part 17: Wire up TicketsEncodingStrategy into STS and TtransactionServices

gmaretic:
**Goals (and why)**:
First part of wiring things up


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SimpleTransactionService.java
```diff
@@ -24,15 +24,15 @@
...
 
 public class SimpleTransactionService implements EncodingTransactionService {
     private final KeyValueService keyValueService;
-    private final TimestampEncodingStrategy encodingStrategy = V1EncodingStrategy.INSTANCE;
+    private final TimestampEncodingStrategy encodingStrategy;
 
-    public SimpleTransactionService(KeyValueService keyValueService) {
+    public SimpleTransactionService(KeyValueService keyValueService, TimestampEncodingStrategy encodingStrategy) {
         this.keyValueService = keyValueService;
+        this.encodingStrategy = encodingStrategy;
```
jeremyk-91: Discussed offline - I think we probably want the table to put stuff into to be injected into the constructor as well, otherwise we need to ensure that no conflict is possible between all encoding strategies used on a given transaction table

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/SimpleTransactionService.java
```diff
@@ -24,15 +24,15 @@
...
-import com.palantir.atlasdb.transaction.encoding.V1EncodingStrategy;
 import com.palantir.atlasdb.transaction.impl.TransactionConstants;
 
 public class SimpleTransactionService implements EncodingTransactionService {
     private final KeyValueService keyValueService;
-    private final TimestampEncodingStrategy encodingStrategy = V1EncodingStrategy.INSTANCE;
+    private final TimestampEncodingStrategy encodingStrategy;
 
-    public SimpleTransactionService(KeyValueService keyValueService) {
+    public SimpleTransactionService(KeyValueService keyValueService, TimestampEncodingStrategy encodingStrategy) {
```
jeremyk-91: This will require us to update the LIP - it's fine, but we should be aware of this.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/service/TransactionServicesTest.java
```diff
@@ -18,38 +18,141 @@
...
 import com.palantir.atlasdb.keyvalue.impl.InMemoryKeyValueService;
+import com.palantir.atlasdb.transaction.encoding.TicketsEncodingStrategy;
+import com.palantir.atlasdb.transaction.encoding.V1EncodingStrategy;
+import com.palantir.atlasdb.transaction.impl.TransactionConstants;
+import com.palantir.atlasdb.transaction.impl.TransactionTables;
 import com.palantir.timestamp.InMemoryTimestampService;
+import com.palantir.timestamp.TimestampManagementService;
+import com.palantir.timestamp.TimestampService;
 
 public class TransactionServicesTest {
```
jeremyk-91: It looks like we may have some use for the parameterized runner, or some other way to try and avoid some of the duplication. This isn't blocking as it's a test, but could be worth looking at if we have a bit of bandwidth.

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/InMemoryKeyValueService.java
```diff
@@ -402,18 +404,27 @@ private void putInternal(
...
 
-            Key key = getKey(table, e.getKey(), timestamp);
+            Key key = getKey(table, entry.getKey(), timestamp);
             byte[] oldContents = putIfAbsent(table, key, contents);
             if (oldContents != null && (doNotOverwriteWithSameValue || !Arrays.equals(oldContents, contents))) {
-                throw new KeyAlreadyExistsException("We already have a value for this timestamp");
+                throw new KeyAlreadyExistsException("We already have a value for this timestamp",
+                        ImmutableList.of(entry.getKey()), knownSuccessfullyCommittedKeys);
             }
+            knownSuccessfullyCommittedKeys.add(entry.getKey());
```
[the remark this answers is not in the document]
gmaretic: Makes sense, but let's not go into too much effort sicne we're pressed for time

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/service/TransactionServicesTest.java
gmaretic: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/TransactionServices.java
```diff
@@ -45,17 +46,24 @@ private static TransactionService createSplitKeyTransactionService(
         return new PreStartHandlingTransactionService(
                 new SplitKeyDelegatingTransactionService<>(
                         transactionSchemaManager::getTransactionsSchemaVersion,
-                        ImmutableMap.of(1, createV1TransactionService(keyValueService))));
+                        ImmutableMap.of(
+                                1, createV1TransactionService(keyValueService),
+                                2, createV2TransactionService(keyValueService))));
```
jeremyk-91: non-urgent: magic numbers should probably be a constant, though I'm not sure where to put them (maybe `TransactionConstants`, or we can try something more clever)

atlasdb-service/src/main/java/com/palantir/atlasdb/transaction/service/PaxosTransactionService.java
```diff
@@ -67,4 +67,9 @@ public void putUnlessExists(long startTimestamp, long commitTimestamp) throws Ke
         }
     }
 
+    @Override
+    public void close() {
+        // do nothing
+    }
+
```
jeremyk-91: incoming merge conflict 🎉

Issue #3662, opened 2018-11-28
FailoverFeignTarget throws without trying all available hosts


gokcanozakdag:
There exists such cases that FeignClient throws without trying all available hosts. A case such that:
- We have 3 timelock nodes: A, B and C
  - B is down, and connections throws `SocketTimeoutException
  - C is the leader
- Assume the list of servers are ordered exactly in the order of A, B, and C
1. `FailoverFeignTarget` first tries hitting A, gets a `NotCurrentLeaderException`
2. It fail overs to B, connection throws `SocketTimeoutException` after pre-determined timeout period
3. If timeout value is bigger than 10 seconds, then `FailoverFeignTarget.checkAndHandleFailure` throws because of `FastFailover` - without retrying the request on C.

This behaviour becomes particularly problematic if the client is in the middle of initializing `TransactionManager`, as `ping()` to `TimestampManager` will consistently fail. We have an internal support ticket related to this issue: PDS-78055

I have also opened a branch with a test reproducing the described case above: https://github.com/palantir/atlasdb/tree/gozakdag/fast-failover-bug

PR #3730, opened 2019-02-02
try all available servers before failing

gokcanozakdag:
**Goals (and why)**:
fixing this issue: https://github.com/palantir/atlasdb/issues/3662

Before failing because of timeout, check if all available servers are tried.

Passing a long supplier to feignTarget constructor, and using the supplier to control time, rather than using thread.sleep - to avoid long running tests.

This causes to fail slower - am I missing any edge cases that can cause more problems?

### DIFF
--- a/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
+++ b/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
@@ -146,7 +146,7 @@ public class FailoverFeignTarget<T> implements Target<T>, Retryer {
         final long currentTime = currentTimeMillis.get();
         boolean failedDueToFastFailover = fastFailoverStartTime != 0
                 && (currentTime - fastFailoverStartTime) > fastFailoverTimeoutMillis
-                && numSwitches.get() > servers.size();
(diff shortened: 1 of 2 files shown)


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
```diff
@@ -134,11 +143,11 @@ private void failoverToNextNode(ExceptionRetryBehaviour retryBehaviour) {
 
     private void checkAndHandleFailure(RetryableException ex) {
         final long fastFailoverStartTime = startTimeOfFastFailover.get();
-        final long currentTime = System.currentTimeMillis();
+        final long currentTime = currentTimeMillis.get();
         boolean failedDueToFastFailover = fastFailoverStartTime != 0
-                && (currentTime - fastFailoverStartTime) > fastFailoverTimeoutMillis;
+                && (currentTime - fastFailoverStartTime) > fastFailoverTimeoutMillis
+                && numSwitches.get() > servers.size() - 1; //try all the available servers before failing
```
gokcanozakdag: we need to change the logic a bit more, currently it is setting numSwitches to 0 after switching to a node - which is a bit weird

gokcanozakdag:
similar pr: #3752

PR #3733, opened 2019-02-04
[_transactions2] Part 19: Coalesce range-adjacent equivalent entries in TimestampPartitioningMaps

jeremyk-91:
**Goals (and why)**:
- Write as little to the coordination table as possible, avoiding unnecessary overheads.
- In particular, `[1000, inf)=2` and installing 2 at timestamp 2000 used to give us `[1000, 2000)=2, [2000, inf)=2` which is unnecessary for our purposes.

- Only produce a new entry if we actually need to. 

- It did cover functional characteristics (timestamps are returned correctly) but didn't cover this size-based one. I've added a test that would have failed under the old implementation.

- RangeMap has a `putCoalescing` but that requires mutability, would a better solution have been to create a mutable rangemap and eventually return it?

**Where should we start reviewing?**: `TimestampPartitioningMap`

**Priority (whenever / two weeks / yesterday)**: ASAP

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/internalschema/TimestampPartitioningMapTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/internalschema/TimestampPartitioningMapTest.java
@@ -30,6 +30,7 @@ public class TimestampPartitioningMapTest {
 
     private static final long TIMESTAMP_1 = 77L;
     private static final long TIMESTAMP_2 = 777L;
+    private static final long TIMESTAMP_3 = 7777L;
 
     @Test
     public void throwsIfInitialMapIsEmpty() {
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/internalschema/TimestampPartitioningMapTest.java
```diff
@@ -85,6 +86,22 @@ public void supportsRevertingVersions() {
         assertThat(newMap.getValueForTimestamp(TIMESTAMP_2)).isEqualTo(1);
     }
 
+    @Test
+    public void coalescesRangesOnEquivalentValues() {
```
gmaretic: 👍

PR #3734, opened 2019-02-05
[_transactions2] Part 20: Better Logging for TransactionSchemaManager#tryInstall

jeremyk-91:
**Goals (and why)**:
- Give more visible feedback on what happened when a new transaction schema version was attempted to be installed.

- Add more descriptive log messages when attempting to do an install, depending on the circumstances.

- No tests in the past. I have added a few for the logic used to determine the log messages.

- Are the messages reasonably clear and understandable?

TransactionSchemaManager.java

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
@@ -80,7 +80,8 @@ public class TransactionSchemaManager {
     public boolean tryInstallNewTransactionsSchemaVersion(int newVersion) {
         CheckAndSetResult<ValueAndBound<InternalSchemaMetadata>> transformResult = tryInstallNewVersion(newVersion);
 
-        Map.Entry<Range<Long>, Integer> finalVersion = getRangeAtBoundThreshold(transformResult);
+        Map.Entry<Range<Long>, Integer> finalVersion = getRangeAtBoundThreshold(
+                Iterables.getOnlyElement(transformResult.existingValues()));
         long newVersionTimestampThreshold = finalVersion.getKey().lowerEndpoint();
 
         if (transformResult.successful() && finalVersion.getValue() == newVersion) {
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
```diff
@@ -78,16 +78,52 @@ public int getTransactionsSchemaVersion(long timestamp) {
...
-                        .orElseThrow(() -> new SafeIllegalStateException("Unexpectedly found no value in store"))
-                        .timestampToTransactionsTableSchemaVersion()
-                        .getValueForTimestamp(valueAndBound.bound()))
-                .collect(Collectors.toList());
-        return Iterables.getOnlyElement(presentVersionPeakValidity) == newVersion;
+        CheckAndSetResult<ValueAndBound<InternalSchemaMetadata>> transformResult = tryInstallNewVersion(newVersion);
+
+        Map.Entry<Range<Long>, Integer> finalVersion = getRangeAtBoundThreshold(
+                Iterables.getOnlyElement(transformResult.existingValues()));
+        long newVersionTimestampThreshold = finalVersion.getKey().lowerEndpoint();
```
jeremyk-91: probably should be `finalVersionTimestampThreshold`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/TransactionSchemaManager.java
```diff
@@ -78,16 +78,52 @@ public int getTransactionsSchemaVersion(long timestamp) {
...
-                .map(valueAndBound -> valueAndBound.value()
-                        .orElseThrow(() -> new SafeIllegalStateException("Unexpectedly found no value in store"))
-                        .timestampToTransactionsTableSchemaVersion()
-                        .getValueForTimestamp(valueAndBound.bound()))
-                .collect(Collectors.toList());
-        return Iterables.getOnlyElement(presentVersionPeakValidity) == newVersion;
+        CheckAndSetResult<ValueAndBound<InternalSchemaMetadata>> transformResult = tryInstallNewVersion(newVersion);
+
+        Map.Entry<Range<Long>, Integer> finalVersion = getRangeAtBoundThreshold(
+                Iterables.getOnlyElement(transformResult.existingValues()));
```
gmaretic: Just an observation that it's strange to return a CAS result that has a list of existing values, when we know it is always going to be a single value. Probably OK, but maybe it's worth having a special superclass of CheckAndSetResult<ValueAndBound<T>> (or even not generic) and have a method on it that returns the single existing value.

jeremyk-91: Yep, that makes sense. I've checked and neither CASR nor VAB are serialized, so changing this in the future at least right now probably won't be too bad.

PR #3735, opened 2019-02-06
client lock leases pt1

gokcanozakdag:
**Goals (and why)**:
Separating #3723 into smaller ones.
This pr introduces the interfaces used by lock-lease mechanism.

Separating timelock client and service interfaces - so that we can add-remove functionality on server side without worrying about client side.

Also introducing leasable responses, where an additional `Lease` object is stored in addition to original response. So that clients can know how long they can trust to a lock.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,19 +14,21 @@
  * limitations under the License.
  */
 
-package com.palantir.leader.lease;
+package com.palantir.common.time;
 
 import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.locks.LockSupport;
 
-import com.google.common.annotations.VisibleForTesting;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
 
 public final class NanoTime implements Comparable<NanoTime> {
+    @JsonValue
     private final long time;
 
-    @VisibleForTesting
-    NanoTime(long time) {
+    @JsonCreator
+    public NanoTime(long time) {
         this.time = time;
     }
 
--- a/atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
@@ -28,12 +28,16 @@ public final class NanoTime implements Comparable<NanoTime> {
     private final long time;
 
     @JsonCreator
-    public NanoTime(long time) {
+    private NanoTime(long time) {
         this.time = time;
     }
 
     public static NanoTime now() {
         return new NanoTime(System.nanoTime());
+    }
+
+    public static NanoTime createForTests(long time) {
+        return new NanoTime(time);
     }
 
     /**
--- a/lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
@@ -16,30 +16,34 @@
 
 package com.palantir.lock.client;
 
-import java.util.Objects;
 import java.util.UUID;
+
+import javax.annotation.concurrent.GuardedBy;
 
 import com.google.common.base.Preconditions;
 import com.palantir.lock.v2.IdentifiedTime;
 import com.palantir.lock.v2.Lease;
 import com.palantir.lock.v2.LockToken;
 
-public final class LeasedLockToken implements LockToken {
+final class LeasedLockToken implements LockToken {
     private final LockToken serverToken;
-    private final LockToken clientToken;
+    private final UUID requestId;
 
+    @GuardedBy("this")
     private Lease lease;
-    private boolean inValidated = false;
+
+    @GuardedBy("this")
+    private boolean invalidated = false;
 
     static LeasedLockToken of(LockToken serverToken, Lease lease) {
         return new LeasedLockToken(serverToken,
-                LockToken.of(UUID.randomUUID()),
+                UUID.randomUUID(),
                 lease);
     }
 
-    private LeasedLockToken(LockToken serverToken, LockToken clientToken, Lease lease) {
+    private LeasedLockToken(LockToken serverToken, UUID requestId, Lease lease) {
         this.serverToken = serverToken;
-        this.clientToken = clientToken;
+        this.requestId = requestId;
         this.lease = lease;
     }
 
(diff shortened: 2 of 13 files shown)


lock-api/src/test/java/com/palantir/lock/client/LeasedLockTokenTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test
+    public void shouldCreateValidTokensUntilExpiry() {
+        LeasedLockToken token = LeasedLockToken.of(LOCK_TOKEN, Lease.of(getIdentifiedTime(), Duration.ofSeconds(1)));
+        assertThat(token.isValid(getIdentifiedTime())).isTrue();
+    }
+
+    @Test
+    public void shouldBeInvalidAfterExpiry() throws Exception {
+        LeasedLockToken token = LeasedLockToken.of(LOCK_TOKEN, Lease.of(getIdentifiedTime(), Duration.ofMillis(10)));
+        Thread.sleep(15);
```
felixdesouza: not a fan of the `Thread.sleep(...)` or relying on two invocations of `now` within the same test

```java
NanoTime now = getIdentifiedTime(); // make this static too
LeaseLockToken token = LLT.of(...)
assertThat(token.isValid(now.plus(Duration.ofSeconds(5))).isTrue(); 
assertThat(token.isValid(now.plus(Duration.ofSeconds(15))).isFalse();
```

I think you can probably delete the above test: `shouldCreateValidTokensUntilExpiry`.

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
```diff
@@ -0,0 +1,149 @@
...
+    @Path("refresh-locks-v2")
+    LeasableRefreshLockResponse refreshLockLeasesV2(Set<LockToken> tokens);
+
+    @POST
+    @Path("lock-v2")
+    LeasableLockResponse lockV2(LockRequest request);
+
+    @POST
+    @Path("start-atlasdb-transaction-v3")
+    LeasableStartIdentifiedAtlasDbTransactionResponse startAtlasDbTransactionV3(
```
felixdesouza: is adding a new endpoint strictly necessary? can `LeasableStartIdentifiedAtlasDbTransactionResponse` extend `StartIdentifiedAtlasDbTransactionResponse`? If that were the case, then unknown properties will be ignored on old clients, and you can just change the type to include new fields.

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,79 @@
...
+    LockToken serverToken() {
+        return serverToken;
+    }
+
+    synchronized Lease getLease() {
+        return lease;
+    }
+
+    synchronized void updateLease(Lease newLease) {
+        Preconditions.checkArgument(newLease.leaseOwnerId().equals(lease.leaseOwnerId()),
```
felixdesouza: if a leased lock token is invalidated, should updating the lease throw? or make it valid again? as it stands, update will succeed, but the lock token will always remain invalid, so it's a bit weird imo

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
```diff
@@ -14,26 +14,32 @@
...
+    @JsonCreator
+    private NanoTime(long time) {
         this.time = time;
     }
 
     public static NanoTime now() {
         return new NanoTime(System.nanoTime());
     }
 
+    public static NanoTime createForTests(long time) {
```
felixdesouza: *sniff*, what was wrong with the previous package-private constructor?

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
```diff
@@ -14,26 +14,32 @@
...
+import com.fasterxml.jackson.annotation.JsonValue;
 
 public final class NanoTime implements Comparable<NanoTime> {
+    @JsonValue
     private final long time;
 
-    @VisibleForTesting
-    NanoTime(long time) {
+    @JsonCreator
+    private NanoTime(long time) {
```
felixdesouza: nit: `time` -> `nanos`

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
```diff
@@ -14,26 +14,32 @@
...
 
 import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.locks.LockSupport;
 
-import com.google.common.annotations.VisibleForTesting;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
 
 public final class NanoTime implements Comparable<NanoTime> {
```
felixdesouza: out of curiosity, how come this class isn't immutables? but all the other classes are?

j-baker: yeahhhh, this should use immutables... sorry, my bad

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
j-baker: > unknown properties will be ignored on old clients

you sure about that? 🤣

felixdesouza: hoping/praying? 😅

lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.UUID;
+
+import javax.annotation.concurrent.GuardedBy;
+
+import com.google.common.base.Preconditions;
+import com.palantir.lock.v2.IdentifiedTime;
+import com.palantir.lock.v2.Lease;
+import com.palantir.lock.v2.LockToken;
+
+final class LeasedLockToken implements LockToken {
```
j-baker: I don't think that this class should be in this PR. It's unused, so it's hard for a reviewer to understand the context of how it should be used in the future.

j-baker: Generally, a PR should try to do one thing; which means that it can be considered in its entirety. IMO the purpose of this PR is to split out the TimelockRpcClient from the TimelockService (which enables a second PR to add the new endpoints and test them on the server and a third PR which starts using them on the client, as we discussed).

lock-api/src/main/java/com/palantir/lock/client/TimelockRpcClient.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param tokens Tokens for which associated locks should be unlocked.
+     * @return Tokens for which associated locks were unlocked
+     */
+    @POST
+    @Path("unlock")
+    Set<LockToken> unlock(Set<LockToken> tokens);
+
+    @GET
+    @Path("leader-time")
+    IdentifiedTime getLeaderTime();
```
j-baker: I would probably not add the new endpoints in this PR, because they won't be implemented on the server and so will lead to slightly weird code for people in the meantime?

lock-api/src/test/java/com/palantir/lock/client/LeasedLockTokenTest.java
j-baker: Instead of using NanoTime.now() anywhere, I'd use NanoTime.createForTests everywhere. It's great when tests are deterministic, and time is a nice new way to inject non-determinism.

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
j-baker: (no need to do this in this PR; can be a little change on its own)

atlasdb-commons/src/main/java/com/palantir/common/time/NanoTime.java
j-baker: useful to be accessed from outside the package, sadly. The alternative is a test dependency in the same package, but it's not like gradle/idea make this easy.

C++ friends would be nice here...

gokcanozakdag:
that makes sense to me @j-baker . Just opened #3737 addressing those points, rather than modifying this one.

PR #3742, opened 2019-02-06
use nanoTime while reaping locks

gokcanozakdag:
**Goals (and why)**:
We are going to be using `System.nanoTime()` as the reference time for lock-leases; and we need to use the same reference clock while reaping the locks.

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LeaseExpirationTimer.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LeaseExpirationTimer.java
@@ -25,16 +25,16 @@ public class LeaseExpirationTimer {
 
     public static final Duration LEASE_TIMEOUT = Duration.ofMillis(20_000);
 
-    private final AtomicReference<NanoTime> lastRefreshTime;
+    private volatile NanoTime lastRefreshTime;
     private final Supplier<NanoTime> clock;
 
     public LeaseExpirationTimer(Supplier<NanoTime> clock) {
         this.clock = clock;
-        this.lastRefreshTime = new AtomicReference<>(clock.get());
+        this.lastRefreshTime = clock.get();
     }
 
     public void refresh() {
-        lastRefreshTime.set(clock.get());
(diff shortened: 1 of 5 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LeaseExpirationTimer.java
```diff
@@ -15,26 +15,33 @@
...
-    private volatile long lastRefreshTimeMillis;
-    private final Clock clock;
+    private final AtomicReference<NanoTime> lastRefreshTime;
+    private final Supplier<NanoTime> clock;
 
-    public LeaseExpirationTimer(Clock clock) {
+    public LeaseExpirationTimer(Supplier<NanoTime> clock) {
         this.clock = clock;
-        this.lastRefreshTimeMillis = clock.getTimeMillis();
+        this.lastRefreshTime = new AtomicReference<>(clock.get());
```
j-baker: given that this uses no atomic operations (cas, etc) might be worth just leaving this as volatile?

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/LeaseExpirationTimerTest.java
```diff
@@ -44,24 +45,24 @@ public void isNotImmediatelyExpired() {
...
-        mockOffsetFromStartTime(LeaseExpirationTimer.LEASE_TIMEOUT_MILLIS * 2);
+        mockOffsetFromStartTime(LeaseExpirationTimer.LEASE_TIMEOUT.toNanos() * 2);
         timer.refresh();
 
         assertThat(timer.isExpired()).isFalse();
     }
 
     private void mockOffsetFromStartTime(long offset) {
-        when(clock.getTimeMillis()).thenReturn(START_TIME_MILLIS + offset);
+        currentTimeNanos = START_TIME_NANOS + offset;
```
j-baker: maybe change this method to be called setTime?

PR #3745, opened 2019-02-07
do not create paxos dir on config deserialization

tpetracca:
**Goals (and why)**: see internal issue 287 for timelock

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -83,9 +83,10 @@ develop
            Previously, these were stored as separate entries, meaning that unnecessary values may have been written to the coordination store; this does not affect correctness, but is unperformant.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3733>`__)
 
-    *    - |changed|
-         - Timelock will now no longer create it's high level paxos directory at configuration de-serialization time. Instead it waits until creating each individual learner or acceptor log directory,
-           allowing timelock to rely more accurately on directory existence as a proxy for said timelock node being new or not.
+    *    - |changed| |improved|
+         - TimeLock will now no longer create its high level paxos directory at configuration de-serialization time.
+           Instead it waits until creating each individual learner or acceptor log directory, allowing timelock to rely more accurately on directory existence as a proxy for said timelock node being new or not.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3745>`__)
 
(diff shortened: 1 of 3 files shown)


felixdesouza:
somewhat confused, when will this get created if not from here?

tpetracca:
Each directory within it (acceptor and learner) will get created when it's cooresponding `PaxosStateLogImpl` is initialized as seen here: https://github.com/palantir/atlasdb/blob/develop/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java#L88

tpetracca:
and note that this codepath should not affect the only other upstream caller of `PaxosInstallConfiguration.dataDirectory` because the validation done [here](https://github.com/palantir/atlasdb/blob/develop/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java#L88-L91) is on the sub-directories which this deleted code didn't create anyway (this check doesn't throw today because of the line I linked in the previous comment).

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
```diff
@@ -62,8 +62,5 @@ default void check() {
                             + "nodes, you have likely already made a mistake by this point. This is a non-trivial "
                             + "operation, please be careful and make sure that you have appropriate approvals.");
         }
-
-        Preconditions.checkArgument(dataDirectory().mkdirs() || dataDirectory().isDirectory(),
```
jeremyk-91: Checkstyle is failing because we're still importing `Preconditions`.

docs/source/release_notes/release-notes.rst
```diff
@@ -83,6 +83,10 @@ develop
            Previously, these were stored as separate entries, meaning that unnecessary values may have been written to the coordination store; this does not affect correctness, but is unperformant.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3733>`__)
 
+    *    - |changed|
+         - Timelock will now no longer create it's high level paxos directory at configuration de-serialization time. Instead it waits until creating each individual learner or acceptor log directory,
+           allowing timelock to rely more accurately on directory existence as a proxy for said timelock node being new or not.
```
jeremyk-91: probably should be `|improved|`, also we normally put the PR number. I'll change this

PR #3746, opened 2019-02-08
enable leasable locking

gokcanozakdag:
**Goals (and why)**:
- Changing AsyncLockService interface to support leasable locking. ~~This pr does not add/remove any timelock endpoints - but makes it much easier to add new leasable endpoints~~.

- Adding leasable endpoints to timelock resource to enable leasing locks on client side.


lock-api/src/main/java/com/palantir/lock/v2/LeasableLockResponse.java
```diff
@@ -0,0 +1,65 @@
...
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableLeasableLockResponse.class)
+@JsonDeserialize(as = ImmutableLeasableLockResponse.class)
+public abstract class LeasableLockResponse {
+    @Value.Parameter
+    public abstract Optional<LockToken> getTokenOrEmpty();
+
+    @Value.Parameter
+    public abstract Optional<Lease> getLeaseOrEmpty();
```
gokcanozakdag: `Optional<Lease>` is a bit ugly, but the other option was to add an invalid lease to timed-out responses. I picked this option.

lock-api/src/main/java/com/palantir/lock/v2/IdentifiedTime.java
```diff
@@ -0,0 +1,40 @@
...
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.common.time.NanoTime;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableIdentifiedTime.class)
+@JsonDeserialize(as = ImmutableIdentifiedTime.class)
+public interface IdentifiedTime {
```
j-baker: maybe call this LeaderTime? and leaderId as the message?

j-baker: ok so if I were you, I'd make this be an abstract class, have clockId be leaderId, and then have a method `boolean isComparableWith(LeaderTime leaderTime)` if that makes sense? Basically push the notion of leader time compatibility into this class

lock-api/src/main/java/com/palantir/lock/v2/LeasableLockResponse.java
j-baker: Two other options are to have an explicit token which contains both LockToken and Lease.

Another option is to have a union type for this; LeasableLockResponse

```
@JsonTypeInfo(type=PROPERTY, property="successful")
@JsonSubtypes(@Type(Successful.class), @Type(Unsuccessful.class)
public interface LeasableLockResponse {
    <T> T accept(Visitor<T> visitor);

    @JsonTypeName("success")
    @Value.Immutable
    @JsonDeserialize(as = ...)
    interface Successful extends LeasableLockResponse {
        LockToken getToken();
    }    

    interface Visitor<T> {
         T visit(Successful success);
         T visit(Unsuccessful failure);
    }
}
```

and then you can just visit whichever response you want; avoids having methods which might throw or whatever if it's in the wrong state?

lock-api/src/main/java/com/palantir/lock/v2/Lease.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.common.time.NanoTime;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableLease.class)
+@JsonDeserialize(as = ImmutableLease.class)
+public abstract class Lease {
+    @Value.Parameter
+    public abstract UUID leaseOwnerId();
```
j-baker: This should probably be `LeaderTime`?

j-baker: (aka merge those two objects?)

lock-api/src/main/java/com/palantir/lock/v2/LeaseContract.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.palantir.lock.v2;
+
+import java.time.Duration;
+
+public final class LeaseContract {
+    private LeaseContract() {}
+
+    public static final Duration LEASE_PERIOD = Duration.ofSeconds(10);
```
j-baker: this should really be in a server-only class because it's not really API

j-baker: it should probably also document when this can be changed (whenever!)

j-baker: I'd be tempted to also make this be a teeny bit smaller? Like, 8 seconds? So we _probably_ don't miss _two_ iterations of the lock refresher?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java
```diff
@@ -125,9 +131,45 @@ public void waitForLocks(@Suspended final AsyncResponse response, WaitForLocksRe
...
+                response.resume(LeasableLockResponse.timedOut());
+            } else {
+                response.resume(LeasableLockResponse.successful(result.get().token(), result.get().lease()));
+            }
+        });
+    }
+
+    @POST
+    @Path("start-atlasdb-transaction-v3")
+    public LeasableStartAtlasDbTransactionResponse startAtlasDbTransactionV3(
```
j-baker: probably worth grouping the logic in this class? so like, all the startAtlasDbTransactions end up near each other?

Also, I'd be tempted to call the old ones deprecatedStartAtlasDbTransaction? rather than versioning these method calls?

lock-api/src/main/java/com/palantir/lock/v2/LeasableLockResponse.java
felixdesouza: +1 on union type

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/AsyncLockService.java
```diff
@@ -39,6 +45,7 @@
     private final HeldLocksCollection heldLocks;
     private final AwaitedLocksCollection awaitedLocks;
     private final ImmutableTimestampTracker immutableTsTracker;
+    private final UUID serviceId;
```
j-baker: at this point I've seen enough UUIDs in this PR that I'd be tempted to have a wrapper type LeadershipId? which just wraps a UUID and has a random() method? Generally prefer wrapping types, just because they make clear what the thing's meant to be in all contexts.

lock-api/src/main/java/com/palantir/lock/v2/LeasableLockResponse.java
gokcanozakdag: going with union type

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/AsyncLockServiceEteTest.java
```diff
@@ -119,8 +117,8 @@ public void waitingRequestGetsTheLockAfterItIsUnlockedWithMultipleLocks() {
     public void requestsAreIdempotentDuringAcquisitionPhase() {
         LockToken currentHolder = lockSynchronously(REQUEST_1, LOCK_A);
 
-        AsyncResult<LockToken> tokenResult = lock(REQUEST_2, LOCK_A);
-        AsyncResult<LockToken> duplicateResult = lock(REQUEST_2, LOCK_A);
+        AsyncResult<LeasableLockToken> tokenResult = lock(REQUEST_2, LOCK_A);
```
j-baker: as noted before; I think you can do with some tests that validate some stuff to do with leasing; that a lock comes back with a lease, that the lease times make sense, that the lease duration makes sense, etc.

j-baker: Like, we currently have a random constant; if that constant were > 20 seconds there'd be a correctness issue here; we should have a test that would catch that correctness issue.

lock-api/src/test/java/com/palantir/lock/v2/LeasableLockResponseTest.java
```diff
@@ -0,0 +1,87 @@
...
+            Lease.of(LeaderTime.of(LeadershipId.random(), NanoTime.createForTests(0)),
+                    Duration.ofSeconds(1));
+
+    private static final String SUCCESSFUL_LOCK_RESPONSE = "{"
+            + "\"type\":\"success\","
+            + "\"token\":{\"requestId\":\"803a09b5-97cb-4034-b35e-1db299b93a7e\"},"
+            + "\"lease\":{"
+                + "\"leaderTime\":{\"leadershipId\":{\"id\":\"2600e8ef-cdb5-441e-a235-475465b7b7fa\"},"
+                + "\"currentTime\":{\"time\":0}},"
+                + "\"validity\":1.000000000}}";
```
j-baker: to me, we should have it so

```json
{"type": "success", "token": {...}, "lease": {"leaderTime": {"id": "uuid", "currentTime":0, "validity":1.000}
```

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/AsyncLockServiceEteTest.java
```diff
@@ -276,14 +275,26 @@ public void timedOutRequestDoesNotHoldLocks() {
...
         assertThat(request2.isFailed()).isTrue();
         assertThat(request2.getError()).isInstanceOf(NotCurrentLeaderException.class);
     }
 
+    @Test
+    public void leaseShouldExpireBeforeReapingLocks() {
+        LeasableLockToken result = lock(REQUEST_1, LOCK_A).get();
+        assertThat(result.lease().isValid(service.identifiedTime())).isTrue();
+
+        waitForTimeout(TimeLimit.of(
```
j-baker: in most tests; recommend stubbing out time rather than sleeping - it makes the tests unnecessarily long-running

j-baker:
some of my comments have become hidden by github after recent changes; think they're all still valid

lock-api/src/main/java/com/palantir/lock/v2/LockResponseV2.java
```diff
@@ -0,0 +1,88 @@
...
+    interface Successful extends LockResponseV2 {
+        LockToken getToken();
+        Lease getLease();
+
+        default <T> T accept(Visitor<T> visitor) {
+            return visitor.visit(this);
+        }
+    }
+
+    @Value.Immutable
```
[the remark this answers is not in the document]
gokcanozakdag: nice!

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollection.java
```diff
@@ -68,23 +83,34 @@ public void failAllOutstandingRequestsWithNotCurrentLeaderException() {
...
+    private Leased<Set<LockToken>> filter(Set<LockToken> tokens, Predicate<HeldLocks> predicate) {
         Set<LockToken> filtered = Sets.newHashSetWithExpectedSize(tokens.size());
+        Optional<NanoTime> minStartTime = Optional.empty();
 
         for (LockToken token : tokens) {
             AsyncResult<HeldLocks> lockResult = heldLocksById.get(token.getRequestId());
             if (lockResult != null && lockResult.test(predicate)) {
                 filtered.add(token);
+                NanoTime lastRefreshTime = lockResult.get().lastRefreshTime();
+                minStartTime = minStartTime.map(t -> lastRefreshTime.isBefore(t) ? lastRefreshTime : t);
```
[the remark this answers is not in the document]
gokcanozakdag: fixed this one. Returned lease will always start _before_ the oldest lease in the returned set of locks - that is start of the lease won't be equal to min refresh time, but less than that. I believe that is fine though.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollection.java
```diff
@@ -68,23 +83,34 @@ public void failAllOutstandingRequestsWithNotCurrentLeaderException() {
...
+    }
+
     private boolean shouldRemove(AsyncResult<HeldLocks> lockResult) {
         return lockResult.isFailed()
                 || lockResult.isTimedOut()
                 || lockResult.test(HeldLocks::unlockIfExpired);
     }
 
-    private Set<LockToken> filter(Set<LockToken> tokens, Predicate<HeldLocks> predicate) {
+    private Leased<Set<LockToken>> filter(Set<LockToken> tokens, Predicate<HeldLocks> predicate) {
```
j-baker: given this is also used in the unlock method, it's probably wrong to return leased here. I think this should just return `Set<LockToken>` and we can wrap with the lease in refresh.

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -179,4 +330,12 @@ private LockToken mockHeldLocksForNewRequest(Consumer<HeldLocks> mockApplier) {
         return request;
     }
 
+    private class Clock implements Supplier<NanoTime> {
+        private Supplier<NanoTime> supplier;
```
j-baker: I would default this thing to NanoTime::now

j-baker: or otherwise new AtomicInteger(0)::incrementAndGet

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -21,21 +21,37 @@
...
+import com.palantir.lock.v2.LeadershipId;
+import com.palantir.lock.v2.Lease;
 import com.palantir.lock.v2.LockToken;
 
 public class HeldLocksCollectionTest {
 
     private static final UUID REQUEST_ID = UUID.randomUUID();
+    private static final UUID REQUEST_ID_2 = UUID.randomUUID();
+
+    private Clock clock = new Clock();
```
j-baker: `Supplier<NanoTime> time` instead of clock

private final LeaderClock leaderClock = new LeaderClock(LeadershipId.random(), () -> time.get());

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+    @Test
+    public void locksShouldBeReapedAndLeaseShouldBeInvalidatedAfterReapPeriod() {
+        AsyncResult<HeldLocks> result = new AsyncResult<>();
+        AsyncResult<Leased<LockToken>> asyncResult =
+                heldLocksCollection.getExistingOrAcquire(REQUEST_ID, () -> result);
+        result.complete(heldLocksForId(REQUEST_ID));
+
+        Lease lease = asyncResult.get().lease();
+
+        setTimeStatic(clock.get().plus(LockLeaseContract.SERVER_LEASE_TIMEOUT).plus(Duration.ofNanos(1)));
```
j-baker: setTime

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -21,21 +21,37 @@
...
 
     private static final UUID REQUEST_ID = UUID.randomUUID();
+    private static final UUID REQUEST_ID_2 = UUID.randomUUID();
+
+    private Clock clock = new Clock();
+    private LeaderClock leaderClock = new LeaderClock(LeadershipId.random(), clock);
+    private final HeldLocksCollection heldLocksCollection = new HeldLocksCollection(leaderClock);
+
+    @Before
+    public void setUp() {
```
j-baker: this is redundant; junit will instantiate a new instance each time so you can just instantiate in the constructor

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+    public void lockLeasesAreValidUntilExpiry() {
+        AsyncResult<HeldLocks> result = new AsyncResult<>();
+        AsyncResult<Leased<LockToken>> asyncResult =
+                heldLocksCollection.getExistingOrAcquire(REQUEST_ID, () -> result);
+        result.complete(heldLocksForId(REQUEST_ID));
+
+        Lease lease = asyncResult.get().lease();
+        assertThat(lease.isValid(leaderClock.time())).isTrue();
+
+        setTimeStatic(clock.get().plus(LockLeaseContract.CLIENT_LEASE_TIMEOUT.minus(Duration.ofNanos(1))));
```
j-baker: advance(Duration)

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
         assertThat(actual).isEqualTo(expected);
     }
 
+    @Test
+    public void lockLeasesAreValidUntilExpiry() {
```
j-baker: setTime(NanoTime.createForTests(123));

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+        Lease lease = asyncResult.get().lease();
+
+        setTimeStatic(clock.get().plus(LockLeaseContract.SERVER_LEASE_TIMEOUT).plus(Duration.ofNanos(1)));
+        assertThat(lease.isValid(leaderClock.time())).isFalse();
+        assertUnlocked(REQUEST_ID);
+    }
+
+    @Test
+    public void refreshShouldReturnAllLockedTokens() {
+        setTimeDynamic(NanoTime::now);
```
j-baker: omit

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+
+        Lease lease = asyncResult.get().lease();
+
+        setTimeStatic(clock.get().plus(LockLeaseContract.SERVER_LEASE_TIMEOUT).plus(Duration.ofNanos(1)));
+        assertThat(lease.isValid(leaderClock.time())).isFalse();
+        assertUnlocked(REQUEST_ID);
+    }
+
+    @Test
+    public void refreshShouldReturnAllLockedTokens() {
```
j-baker: we can drop this test, if there's already a test for the case where we didn't refresh all the tests

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+
+        Set<LockToken> tokens = ImmutableSet.of(t1, t2);
+
+        Set<LockToken> refreshResult = heldLocksCollection.refresh(tokens).value();
+        assertThat(refreshResult).containsExactlyInAnyOrder(t1, t2);
+    }
+
+    @Test
+    public void leaseShouldStartBeforeRefreshTime() {
+        setTimeDynamic(NanoTime::now);
```
j-baker: set the supplier to new AtomicLong(123)::getAndIncrement

timelock-server/src/test/java/com/palantir/atlasdb/timelock/lock/HeldLocksCollectionTest.java
```diff
@@ -111,6 +128,106 @@ public void unlockReturnsSubsetOfUnlockedLocks() {
...
+    @Test
+    public void emptyRefreshResponse() {
+        setTimeDynamic(NanoTime::now);
+
+        LockToken t1 = LockToken.of(UUID.randomUUID());
+        LockToken t2 = LockToken.of(UUID.randomUUID());
+        Leased<Set<LockToken>> refreshResult = heldLocksCollection.refresh(ImmutableSet.of(t1, t2));
+
+        assertThat(refreshResult.value()).isEmpty();
+        assertThat(refreshResult.lease().leaderTime().currentTime()).isLessThan(NanoTime.now());
```
j-baker: isLessThanOrEqualTo?

lock-api/src/main/java/com/palantir/lock/v2/Lease.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Value.Parameter
+    abstract Duration validity();
+
+    public boolean isValid(LeaderTime currentLeaderTime) {
+        return leaderTime().isComparableWith(currentLeaderTime)
+                && currentLeaderTime.currentTime().isBefore(expiry());
+    }
+
+    private NanoTime expiry() {
```
j-baker: probably worth having a speculative `@JsonIgnore` here?

PR #3752, opened 2019-02-11
[PDS-82684] Feign Target Edge Case Sadness

jeremyk-91:
**Goals (and why)**:
- Internal ref. PDS-82684
- This fixes an edge case where timelock / something pointed to by an AtlasDbHttpClient can throw even when the cluster is good. Specifically, suppose you have three nodes `n1`, `n2` and `n3`, where `n3` is the leader and `n2` takes a long time to respond and eventually fails the request. Then, we will try `n1` and obtain that it is not the leader which starts a ten-second clock on fast failover exceptions. We then try `n2`, which after a long time tosses out a socket timeout exception. The ten-second clock has run out, so we fail, rather than trying `n3`.

- Require additionally that we have tried all nodes at least once before failing out.

- I added two tests that would have failed under the old model.

- Are there concurrency issues I'm missing out? It seems the purpose of using `AtomicLong` and `AtomicReference` is for visibility, not concurrency; each request gets its own copy of the target (looking at the javadoc for `Retryer`).
- Is this bad for performance? I think it's fine, as it is no worse than what we already have today, and also once we figure out which node the leader is, we'll stay pinned to that node until there is a leader election or network failure there.
- There is a similar bug for clusters of size 15 and above, though timelock generally isn't run in a 15-node configuration so we haven't run into that.

**Where should we start reviewing?**: FailoverFeignTarget

### DIFF
--- a/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
+++ b/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
@@ -28,6 +28,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.palantir.logsafe.SafeArg;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import feign.Client;
@@ -61,10 +62,6 @@ public class FailoverFeignTarget<T> implements Target<T>, Retryer {
 
     private final ThreadLocal<Integer> mostRecentServerIndex = new ThreadLocal<>();
 
-    public FailoverFeignTarget(Collection<String> servers, Class<T> type) {
-        this(servers, DEFAULT_MAX_BACKOFF_MILLIS, type);
-    }
-
     public FailoverFeignTarget(Collection<String> servers, int maxBackoffMillis, Class<T> type) {
         Preconditions.checkArgument(maxBackoffMillis > 0);
         this.servers = ImmutableList.copyOf(ImmutableSet.copyOf(servers));
@@ -72,7 +69,7 @@ public class FailoverFeignTarget<T> implements Target<T>, Retryer {
         this.maxBackoffMillis = maxBackoffMillis;
     }
 
-    public void sucessfulCall() {
+    private void successfulCall() {
         numSwitches.set(0);
         failuresSinceLastSwitch.set(0);
         startTimeOfFastFailover.set(0);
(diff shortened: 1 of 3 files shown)


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
```diff
@@ -57,25 +62,30 @@
 
     private final AtomicLong failuresSinceLastSwitch = new AtomicLong();
     private final AtomicLong numSwitches = new AtomicLong();
-    private final AtomicLong startTimeOfFastFailover = new AtomicLong();
+    private final AtomicReference<Optional<FastFailoverState>> fastFailoverState
```
jkozlowski: In your PR you say that each request gets a copy of FailoverFeignTarget, I don't see that? I think we absolutely need the atomics here.

The concurrency setup in this class is a bit special but it doesn't seem immediately off.

jkozlowski:
Thinking about it a bit more, this kinda nullifies the "fast-failover" idea: the way I understand "fast-failover" is that you want to fail after 10 seconds because you have some sort of deadline to fulfill and you don't care if you get something slightly wrong, because being late is worse than wrong.

Therefore, I would question the idea of fast-failover: do we actually want it? If we do, we should keep this as is, but turn it off in the migrator which clearly cares that you try the whole cluster.

This is my slight outsider perspective.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
```diff
@@ -57,25 +62,30 @@
...
+            = new AtomicReference<>(Optional.empty());
 
     private final ThreadLocal<Integer> mostRecentServerIndex = new ThreadLocal<>();
 
-    public FailoverFeignTarget(Collection<String> servers, Class<T> type) {
-        this(servers, DEFAULT_MAX_BACKOFF_MILLIS, type);
-    }
+    private final LongSupplier clock;
```
gokcanozakdag: we have a `Clock` object that can be used here instead of LongSupplier

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
jeremyk-91: I said this because FailoverFeignTarget implements Retryer, where the javadocs say

```
 * Cloned for each invocation to {@link Client#execute(Request, feign.Request.Options)}.
 * Implementations may keep state to determine if retry operations should continue or not.
```

Though I don't understand feign internals sufficiently well to be 100% sure that removing this is safe.

jeremyk-91: We spoke offline - it's not actually safe to remove this. Clone() is called, but our impl of clone() does a return this;

jeremyk-91:
Caught up offline. This "fast failover" stuff was originally written just for the PaxosLeaderElectionService - I interpreted it to mean "don't retry on me, go to the next node" (compare with retrying 2X on most other kinds of failures); in practice this only happens for 503s. So we're "fast" in terms of failing over to other nodes, though I don't think we're necessarily fast in serving the whole request or failing (right now if we have a connection timeout of 60 seconds we still take up to 70 to fail, rather than 10).

Will swap out `LongSupplier` for `Clock`

jkozlowski:
Ping on this: should we merge?

PR #3755, opened 2019-02-12
move timelock dep to lock-api

tpetracca:
atlasdb-config has a dep on lock-api so this isn't a break. It just makes sure people who are only dep'ing on lock-api and not atlasdb-config can detect constraints.


sandorw:
Is this the right final place for this to be? It's confusing the more we move it around, so just asking if this is the right project building block where no one would leave this out or rely specifically on a more granular package.

tpetracca:
@sandorw mostly this feels like the right place because it's where the `TimeLockClient` class lives, which is (I assume) the smallest reasonable public API for timelock.

PR #3756, opened 2019-02-13
Fix getRowColumnRange NPEs

### LINKED ISSUE #3743: NPE when calling getRowsColumnRange
This is on 0.116.1.

```
java.lang.NullPointerException: {throwable0_message}
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl$3$1.apply(CassandraKeyValueServiceImpl.java:864)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl$3$1.apply(CassandraKeyValueServiceImpl.java:835)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPoolingContainer.runWithGoodResource(CassandraClientPoolingContainer.java:129)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPoolingContainer.runWithPooledResource(CassandraClientPoolingContainer.java:105)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPoolImpl.runWithPooledResourceRecordingMetrics(CassandraClientPoolImpl.java:447)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPoolImpl.runWithRetryOnHost(CassandraClientPoolImpl.java:394)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl$3.page(CassandraKeyValueServiceImpl.java:832)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl$3.getFirstPage(CassandraKeyValueServiceImpl.java:821)
	at com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServiceImpl$3.getFirstPage(CassandraKeyValueServiceImpl.java:818)
	at com.palantir.util.paging.AbstractPagingIterable$PagingIterator.computeNext(AbstractPagingIterable.java:60)
	at com.google.common.collect.AbstractIterator.tryToComputeNext(AbstractIterator.java:140)
	at com.google.common.collect.AbstractIterator.hasNext(AbstractIterator.java:135)
	at com.palantir.common.base.ClosableIterators$1.hasNext(ClosableIterators.java:32)
	at com.google.common.collect.Iterators$ConcatenatedIterator.hasNext(Iterators.java:1331)
	at com.palantir.atlasdb.keyvalue.impl.LocalRowColumnRangeIterator.hasNext(LocalRowColumnRangeIterator.java:34)
	at com.google.common.collect.Iterators$ConcatenatedIterator.hasNext(Iterators.java:1331)
	at com.palantir.atlasdb.keyvalue.impl.LocalRowColumnRangeIterator.hasNext(LocalRowColumnRangeIterator.java:34)
	at com.palantir.atlasdb.transaction.impl.SnapshotTransaction.getRowsColumnRange(SnapshotTransaction.java:380)
	at com.palantir.atlasdb.transaction.impl.ReadTransaction.getRowsColumnRange(ReadTransaction.java:115)
        ...
```

sandorw:
**Goals (and why)**:
Fixes #3743 

Handle the case where all results for the specified row had ts > startTs, which causes them to be filtered out and the returned results to be null. In this case, we should return an empty page and be careful to specify the correct behavior for whether or not there are potentially more results on subsequent pages. To avoid infinite paging loops, we must also extract the last column returned in the current page.

The next paging request will specify a start range using both the start column and the start timestamp, which means that we'll only return valid results or move on to the next column.

Added several new test cases, all of which NPEd without the fix. Several tests cover additional edge cases where one might retrieve an entire page of results with ts > startTs, at which point it's important to continue paging, as there may be real results that are visible, or other edge cases in the paging logic.

Is the paging logic correct? Any more edge cases to test?

CKVSI

Soon, active support issue

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -864,9 +864,10 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
                         RowColumnRangeExtractor extractor = new RowColumnRangeExtractor(metricsManager);
                         extractor.extractResults(ImmutableList.of(row), results, startTs);
                         RowColumnRangeExtractor.RowColumnRangeResult decoded = extractor.getRowColumnRangeResult();
-                        //if (decoded.getResults().isEmpty()) {
-                        //    return SimpleTokenBackedResultsPage.create(startCol, ImmutableList.of(), false);
-                        //}
+                        if (decoded.getResults().isEmpty()) {
+                            return SimpleTokenBackedResultsPage.create(startCol, ImmutableList.of(),
+                                    values.size() >= batchColumnRangeSelection.getBatchHint());
+                        }
                         Map<Cell, Value> ret = decoded.getResults().get(row);
 
                         ColumnOrSuperColumn lastColumn = values.get(values.size() - 1);
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -865,18 +865,10 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
                         extractor.extractResults(ImmutableList.of(row), results, startTs);
                         RowColumnRangeExtractor.RowColumnRangeResult decoded = extractor.getRowColumnRangeResult();
                         if (decoded.getResults().isEmpty()) {
-                            if (values.size() == 1) {
-                                throw new RuntimeException("Found 1 result");
-                            }
-                            if (values.size() == 2) {
-                                throw new RuntimeException("Found 2 results");
-                            }
-                            if (values.size() > 2) {
-                                throw new RuntimeException("Found >2 results");
-                            }
-                            // Need to move startCol forward
-                            // What if all we read were versions of the same row?
-                            return SimpleTokenBackedResultsPage.create(startCol, ImmutableList.of(),
+                            // All returned results were at ts > startTs
+                            // Start the next page from the last column seen to ensure forward progress
+                            byte[] lastColumn = extractor.getLastColumnForRow(row).get();
+                            return SimpleTokenBackedResultsPage.create(lastColumn, ImmutableList.of(),
                                     values.size() >= batchColumnRangeSelection.getBatchHint());
                         }
                         Map<Cell, Value> ret = decoded.getResults().get(row);
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -865,6 +865,17 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
                         extractor.extractResults(ImmutableList.of(row), results, startTs);
                         RowColumnRangeExtractor.RowColumnRangeResult decoded = extractor.getRowColumnRangeResult();
                         if (decoded.getResults().isEmpty()) {
+                            if (values.size() == 1) {
+                                throw new RuntimeException("Found 1 result");
+                            }
+                            if (values.size() == 2) {
+                                throw new RuntimeException("Found 2 results");
+                            }
+                            if (values.size() > 2) {
+                                throw new RuntimeException("Found >2 results");
+                            }
+                            // Need to move startCol forward
+                            // What if all we read were versions of the same row?
                             return SimpleTokenBackedResultsPage.create(startCol, ImmutableList.of(),
                                     values.size() >= batchColumnRangeSelection.getBatchHint());
                         }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -21,7 +21,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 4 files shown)


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
```diff
@@ -258,6 +261,60 @@ public void testKeyValueEmptyRange() {
...
+    public void testRowsColumnRange_forwardProgressAfterValidResult() {
+        putDirect("row1", "col1", "v1a", 1);
+        putDirect("row1", "col1", "v1b", 2);
+        putDirect("row1", "col1", "v1c", 3);
+        putDirect("row1", "col1", "v1d", 4);
+        putDirect("row1", "col1", "v1e", 5);
+        putDirect("row1", "col2", "v2", 5);
+
+        byte[] rowBytes = PtBytes.toBytes("row1");
+        RowColumnRangeIterator iterator = keyValueService.getRowsColumnRange(
```
sandorw: I expected this test to fail, since we page through 1 result at a time, and the startCol will never move beyond `col1`, since each successive query should restart its search from `startTs = 6`.

However, this hits the pre-existing code path, which moves you onto the next lexicographic column name since we already found a result for startCol.

A result at `ts = 7` for `row1` would still give us forward progress, since the next range scan would start from the concrete row `row1` and `startTs = 6`, which would find `(col1, v1e)` and move us on to the next potential column.

sandorw:
@hsaraogi for SA

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -863,6 +863,13 @@ public String toString() {
                         RowColumnRangeExtractor extractor = new RowColumnRangeExtractor(metricsManager);
                         extractor.extractResults(ImmutableList.of(row), results, startTs);
                         RowColumnRangeExtractor.RowColumnRangeResult decoded = extractor.getRowColumnRangeResult();
+                        if (decoded.getResults().isEmpty()) {
+                            // All returned results were at ts > startTs
+                            // Start the next page from the last column seen to ensure forward progress
+                            byte[] lastColumn = extractor.getLastColumnForRow(row).get();
```
sandorw: `.get()` here should be safe since we are collecting results only for a single row, and we have already confirmed that results were returned - there must be at least one column for which results were returned.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RowColumnRangeExtractor.java
```diff
@@ -125,4 +124,9 @@ public RowColumnRangeResult getRowColumnRangeResult() {
         return new RowColumnRangeResult(collector, rowsToLastCompositeColumns, emptyRows, rowsToRawColumnCount);
     }
 
+    public Optional<byte[]> getLastColumnForRow(byte[] row) {
+        return Optional.ofNullable(rowsToLastCompositeColumns.get(row))
+                .map(CassandraKeyValueServices::decomposeName)
+                .map(Pair::getLhSide);
```
jeremyk-91: Yep, looks right

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
jeremyk-91: It's probably worth adding the test for having some results after the start, to confirm we pick the right one. I didn't see a test for this case (at least, didn't see one near the tests you've added)

```
    @Test
    public void testRowsColumnRange_abortsCorrectlyHalfway() {
        putDirect("row1", "col1", "v1a", 3);
        putDirect("row1", "col1", "v1b", 4);
        putDirect("row1", "col1", "v1c", 5);
        putDirect("row1", "col1", "v1d", 7);
        putDirect("row1", "col1", "v1e", 8);
        putDirect("row1", "col2", "v2", 5);

        byte[] rowBytes = PtBytes.toBytes("row1");
        RowColumnRangeIterator iterator = keyValueService.getRowsColumnRange(
                TEST_TABLE,
                ImmutableList.of(rowBytes),
                new ColumnRangeSelection(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY),
                1,
                6);
        assertThat(PtBytes.toBytes("v1c")).isEqualTo(iterator.next().getValue().getContents());
        assertThat(PtBytes.toBytes("v2")).isEqualTo(iterator.next().getValue().getContents());
        assertThat(iterator.hasNext()).isFalse();
    }
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -863,6 +863,13 @@ public String toString() {
...
+                            return SimpleTokenBackedResultsPage.create(lastColumn, ImmutableList.of(),
+                                    values.size() >= batchColumnRangeSelection.getBatchHint());
```
j-baker: I think this code can be simplified. Isn't this appropriately fixed by line 873 just being a getOrDefault(row, Collections.emptyMap()); // row might not exist if all values are at a newer timestamp

j-baker: line 876 does the same thing as your new method in RowColumnRangeExtractor, after all.

sandorw: Yes, you're correct, can swap to that and remove the added code in the extractor.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -863,8 +863,9 @@ public String toString() {
                         RowColumnRangeExtractor extractor = new RowColumnRangeExtractor(metricsManager);
                         extractor.extractResults(ImmutableList.of(row), results, startTs);
                         RowColumnRangeExtractor.RowColumnRangeResult decoded = extractor.getRowColumnRangeResult();
-                        Map<Cell, Value> ret = decoded.getResults().get(row);
 
+                        // May be empty if all results are at ts > startTs
+                        Map<Cell, Value> ret = decoded.getResults().getOrDefault(row, Maps.newLinkedHashMap());
```
j-baker: let's just make the return type be map...

PR #3757, opened 2019-02-13
Expose lightweight function for fetching rows in range

sandorw:
**Goals (and why)**:
Wanted to start discussion about potentially exposing this endpoint. The PR is not complete - possibly wanting a test, better javadoc, and the API may not be quite what we want.

Expose a new method on the KVS to allow use of some C* code we wrote to make legacy sweep better.

Is this a reasonable method to expose on the KVS? Especially since it's only implemented for Cassandra currently.

The use case is to range scan over the rows in a table for a data migration. Only a subset of the rows need to be migrated, but some rows can be quite large, so we'd like to avoid scanning over all of the data. As a follow-up question, since this endpoint isn't transactional, will this workflow work?

- Update write workflows to respect the migration (live writes enforce expected behavior)
- Non-transactionally scan a range of the table with this endpoint
- Transactionally migrate the data in those rows

If data disappears in the meantime, that's fine, since we'll just fail to read it transactionally. If rows are added after the fact, we should be safe as long as concurrent writes enforce the intended behavior and don't need to be re-migrated.

Lastly, any complaints about the API as is, assuming we want to expose it?


sandorw:
@liliyang for SA

felixdesouza:
it might be worth implementing this for `InMemoryKeyValueService`? So you can iterate quickly without having to spin up cassandra.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1214,6 +1215,21 @@ CfDef getCfForTable(TableReference tableRef, byte[] rawMetadata, int gcGraceSeco
...
+    @Override
+    public List<byte[]> getRowKeysInRange(TableReference tableRef, byte[] startRowInclusive, int batchSize) {
+        RowGetter rowGetter = new RowGetter(clientPool, queryRunner, ConsistencyLevel.QUORUM, tableRef);
+        GetCellTimestamps getCellTimestamps = new GetCellTimestamps(
+                newInstrumentedCqlExecutor(),
+                rowGetter,
+                tableRef,
+                startRowInclusive,
+                batchSize,
+                config);
```
jeremyk-91: It's an unfortunate consequence of coupling between sweep and `GetCellTimestamps`, but internally this has a `rowGetter.getRows("getCandidateCellsForSweeping", ...)` call which will likely be confusing e.g. for jstacks or the kvs-slow-log, as that name is used to identify what the KVS is trying to do.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -426,6 +426,17 @@ void deleteAllTimestamps(TableReference tableRef,
...
             CandidateCellForSweepingRequest request);
 
+    /**
+     * Scans a range of a table and returns the row keys in that range.
+     * <p>
+     * This endpoint specifically avoids paging through each full row.
+     * <p>
+     * The batch size limits the number of results from each request.
+     */
+    default List<byte[]> getRowKeysInRange(TableReference tableRef, byte[] startRowInclusive, int batchSize) {
```
jeremyk-91: From an API standpoint, I'd expect this to probably have the same interface as `getRange`, which takes a `RangeRequest` object (rather than always being end-unbounded). I see that the internal APIs e.g. for `GetCellTimestamps` are similarly always end-unbounded; these were in the past only used for Sweep, so end-unbounded made sense then, though if we're opening this up I don't know if it would be useful to allow end bounds. For the specific use case mentioned it's probably fine.

If I'm reading the impl correctly this actually returns the results of just one batch, so I think the `batchSize` should probably be something stronger, like `maxResults`

j-baker:
Main thing for me is: Not sure we should be providing this as functionality; all the use cases for it are things we don't want people to do; the current case is very very special and it seems reasonable for them to reach in and find the cassandra key value service at the moment.
Also, I heard the data they need can be found elsewhere?
Also, it seems wrong to not implement the method in some of our supported key-value-services; it removes the write-once-run-anywhere-maybe-slowly claim of Atlas (after all, this internal project supports postgres, and you can't test without the inmemorykvs).

sandorw:
Thanks for the discussion - I think it's easier to just expose this on CKVS for the moment - we can always support it more formally on other KVSs if it turns out that we need to. Waiting for the moment on some digging through C* code to see if there's another way to do this which avoids Cassandra needing to scan the data, even if it doesn't return it.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
gmaretic: How do we feel about not adding more methods that use `byte[]`? It's kind of awkward to mix and match `ByteString` (or whatever alternative) and `byte[]` in the same interface, but not convinced that perpetuating the use of byte arrays in newly added methods is great either

felixdesouza: it might be possible we could probably generate overloads to keep backcompat or something, where the base interface will be in terms of `ByteString`s and then the public facing type: `KeyValueService` will keep the `byte[]` stuff around. 

On the other hand, are there really that many places that will use a `byte[]` directly? We should be able to make most of these changes.

> N.B. this is for api surfaces, internals can change at a later cadence I suppose.

liliyang:
Hi @sandorw, what's the status on this work?

sandorw:
It stalled a bit last week due to some other last minute tickets that needed work. I'll be on support this week, so I'll coordinate someone else to push this through.

gmaretic:
Update:

We considered implementing this through CQL "SELECT DISTINCT key" which seemed to do exactly what we wanted, but I after reading through the Cassandra code and running some benchmarks, I'm pretty convinced the above implementation is basically the same except that SELECT DISTINCT makes sure that the row contains a live value. This does come at a cost, since the distinct statement was significantly slower in benchmarks.

Even with wide rows, very high memory hogging, and a hundred clients making requests, the test Cassandra did not roll over and was in fact quite performant. It would be great if we could get @tpetracca  to actually verify that we do not in fact scan the entire table under the hood if that is indeed the case though.

Added a couple of integration tests and a few ete tests as proof of concept how this can be used given a TM.

gmaretic:
@sandorw I hijacked this PR, so now you can review it :)

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -426,6 +426,7 @@ void deleteAllTimestamps(TableReference tableRef,
             TableReference tableRef,
             CandidateCellForSweepingRequest request);
 
+
```
sandorw: There must be balance. If you fix whitespace in one place, you must break it in another.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
sandorw: @gmaretic - worth incorporating this feedback. I think taking an end byte[] is a good idea. Taking a `RangeRequest` unfortunately allows selecting columns to retain (ignored) and keeps the `batchHint` issue.

sandorw: I see you already swapped to `maxResults` - we can either also include the upper bound or can create a wrapper object to encapsulate them.

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/rows/SimpleGetRowKeysResource.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private final CassandraKeyValueService kvs;
+
+    public SimpleGetRowKeysResource(KeyValueService wrappedKvs) {
+        this.kvs = unwrapCassandraKvs(wrappedKvs);
+    }
+
+    @Override
+    public void resetTable() {
+        kvs.createTable(TABLE_REFERENCE,
```
sandorw: Why create the table here?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraGetRowKeysEteTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public void canGetRowKeysOnEmptyTable() {
+        assertThat(client.getRowKeys(0, 10)).isEmpty();
+    }
+
+    @Test
+    public void getRowKeysReturnsOneResultPerRowStartingWithStartRow() {
+        generateCellsWithNumberedRows(3).forEach(client::insertCell);
+        generateCellsWithNumberedRows(5).forEach(client::insertCell);
+        generateCellsWithNumberedRows(4).forEach(client::insertCell);
+        assertListsContainsExactlyRowsFromTo(client.getRowKeys(1, 10), 1, 5);
```
sandorw: Move to batch size of 5? If this was broken, we're currently only 1 result off from not returning 5 given the number of results per row key.

docs/source/release_notes/release-notes.rst
```diff
@@ -163,6 +163,11 @@ v0.118.0
     *    - Type
          - Change
 
+    *    - |improved|
+         - ``CassandraKeyValueServiceImpl`` now exposes a lightweight method for obtaining row keys.
+           If you believe you need to use this method, you should reach out to the AtlasDB team first to asses your options.
```
sandorw: nit: assess

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -358,6 +365,78 @@ public void upgradeFromOlderInternalSchemaDoesNotErrorOnTablesWithOldMetadata()
         verify(logger, never()).error(anyString(), any(Object.class));
     }
 
+    @Test
```
sandorw: I think we should consolidate the tests here and just check the unwrapping logic in the the other test class

gmaretic:
Final update:
Dug deep into Cassandra code with @ashrayjain and confirmed that Cassandra skips reading even a single column from sstables on this call. This is consistent with the benchmarking and stress testing, and means we can use this regardless of how wide our rows and how large our cells are. It does mean we will potentially return rows that only have tombstones, but that is a fair tradeoff for how lightweight this operation is.

Addressed comments above and added some more tests

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceGetRowKeysInRangeTest.java
```diff
@@ -0,0 +1,127 @@
...
+    public void initialize() {
+        kvs = CASSANDRA.getDefaultKvs();
+        kvs.createTable(GET_ROW_KEYS_TABLE, NO_SWEEP_METADATA);
+    }
+
+    @After
+    public void cleanup() {
+        kvs.truncateTable(GET_ROW_KEYS_TABLE);
+    }
+
```
sandorw: I think the case missed here is the empty range case from the other class.

sandorw:
Concerned about the unwrap functionality working correctly? Not sure if we test that anywhere.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceGetRowKeysInRangeTest.java
gmaretic: sure, added this in

gmaretic:
@sandorw I opened an issue about it. We verified that it works manually, so not that worried, but we probably should add in a test so that it doesn't break. Outside of the scope of this PR though

sandorw:
Can't approve this since I started it, but 👍

PR #3758, opened 2019-02-14
move timelock clientId information to timelock client

gokcanozakdag:
**Goals (and why)**:
Now that we have separate interfaces for Timelock client, and actual server; we can generate the client id inside timelock client, rather than passing it as a parameter.

This pr will help the client-side lock lease implementation, as we can easily convert lockImmutableTimestamp request into a startIdentifiedAtlasDbTransaction, and get the lease information.

I have also removed the deprecated `startAtlasDbTransaction` on client side, as it is a pretty AtlasDb internal method, and removing it won't affect server side, so why not.

Create a clientId in `DefaultTimelockService`, and use it while calling `startIdentifiedAtlasDbTransaction`.

`DefaultTimelockService`


lock-api/src/main/java/com/palantir/lock/v2/DefaultTimelockService.java
```diff
@@ -17,18 +17,25 @@
...
+        this.clientId = clientId;
     }
 
     public static TimelockService create(TimelockRpcClient timelockRpcClient) {
-        return new DefaultTimelockService(timelockRpcClient);
+        return new DefaultTimelockService(timelockRpcClient, UUID.randomUUID());
+    }
+
+    public static TimelockService create(TimelockRpcClient timelockRpcClient, UUID clientId) {
+        return new DefaultTimelockService(timelockRpcClient, clientId);
```
jeremyk-91: This one is only used in the timelock integration tests (in fact when I first looked at it I wondered why it existed in the first place). I think if we change the testable timelock cluster to use an RPC client (see comment there) we can remove this method.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -195,7 +197,8 @@ public WaitForLocksResponse waitForLocks(WaitForLocksRequest request) {
 
     public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction(
             StartIdentifiedAtlasDbTransactionRequest request) {
-        return timelockService().startIdentifiedAtlasDbTransaction(request);
+        return timelockService(defaultClient, request.requestorId())
+                .startIdentifiedAtlasDbTransaction(ImmutableIdentifiedTimeLockRequest.of(request.requestId()));
     }
```
jeremyk-91: If within this class we expose the RPC client, the implementation can be a little simpler:

```
    private TimelockRpcClient rpcClient(String client) {
        return proxies.failoverForClient(client, TimelockRpcClient.class);
    }

    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction(
            StartIdentifiedAtlasDbTransactionRequest request) {
        return rpcClient(defaultClient).startIdentifiedAtlasDbTransaction(request);
    }
```

(and you won't need the timelockService(String, UUID) method nor the DefaultTimelockService factory that takes a UUID)

PR #3760, opened 2019-02-15
client side lock lease changes

gokcanozakdag:
**Goals (and why)**:
Client side changes for leasable locking. Opening this pr against server-side branch, to make it easier to see the diff. Should be merged to develop, after merging server-side changes (and declaring a version dependency)


lock-api/src/main/java/com/palantir/lock/client/LeasedLockToken.java
```diff
@@ -0,0 +1,79 @@
...
+    LockToken serverToken() {
+        return serverToken;
+    }
+
+    synchronized Lease getLease() {
+        return lease;
+    }
+
+    synchronized void updateLease(Lease newLease) {
+        Preconditions.checkArgument(lease.leaderTime().isComparableWith(newLease.leaderTime()),
```
j-baker: probably worth using safe-logging preconditions and having args for the incorrect leader ids

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,178 @@
...
+    private final TimelockRpcClient delegate;
+    private final UUID clientId;
+
+    private LeasingTimelockClient(TimelockRpcClient timelockRpcClient, UUID clientId) {
+        this.delegate = timelockRpcClient;
+        this.clientId = clientId;
+    }
+
+    public static LeasingTimelockClient create(TimelockRpcClient timelockRpcClient) {
+        return new LeasingTimelockClient(timelockRpcClient, UUID.randomUUID());
```
j-baker: `return create(timelockRpcClient, UUID.randomUUID());`

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,178 @@
...
+                "All lock tokens should be an instance of LeasedLockToken");
+        return (Set<LeasedLockToken>) (Set<?>) tokens;
+    }
+
+    private static Set<LockToken> serverTokens(Set<LeasedLockToken> leasedTokens) {
+        return leasedTokens.stream()
+                .map(LeasedLockToken::serverToken)
+                .collect(Collectors.toSet());
+    }
+
```
j-baker: can omit this newline

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,178 @@
...
+        Set<LeasedLockToken> refreshedTokens = leasedTokens.stream()
+                .filter(t -> refreshLockResponse.refreshedTokens().contains(t.serverToken()))
+                .collect(Collectors.toSet());
+
+        refreshedTokens.forEach(t -> t.updateLease(lease));
+        return refreshedTokens;
+    }
+
+    private static Set<LeasedLockToken> leasedTokens(Set<LockToken> tokens) {
```
j-baker: `@SuppressWarnings("unchecked")`

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
+        return delegate.getFreshTimestamps(numTimestampsRequested);
+    }
+
+    @Override
+    public LockImmutableTimestampResponse lockImmutableTimestamp(IdentifiedTimeLockRequest request) {
+        StartAtlasDbTransactionResponseV3 leasableResponse =
```
j-baker: would just call this `response`

lock-api/src/main/java/com/palantir/lock/client/LeasingTimelockClient.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return ImmutableLockImmutableTimestampResponse.of(
+                leasableResponse.immutableTimestamp().getImmutableTimestamp(),
+                LeasedLockToken.of(leasableResponse.immutableTimestamp().getLock(), leasableResponse.getLease()));
+    }
+
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction(
+            IdentifiedTimeLockRequest request) {
+        StartAtlasDbTransactionResponseV3 leasableResponse =
```
j-baker: same here, `response`

lock-api/src/test/java/com/palantir/lock/client/LeasedLockTokenTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void shouldBeInvalidAfterInvalidation() {
+        LeasedLockToken token = LeasedLockToken.of(LOCK_TOKEN, Lease.of(getIdentifiedTime(), Duration.ofSeconds(1)));
+        token.inValidate();
+        assertThat(token.isValid(getIdentifiedTime())).isFalse();
+    }
+
+    private LeaderTime getIdentifiedTime() {
+        return LeaderTime.of(LEADER_ID, NanoTime.now());
```
j-baker: we should not be using stubbed out time and sleeps; NanoTime has a createForTests method on it!

lock-api/src/test/java/com/palantir/lock/client/LeasedLockTokenTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.junit.Test;
+
+import com.palantir.common.time.NanoTime;
+import com.palantir.lock.v2.LeaderTime;
+import com.palantir.lock.v2.LeadershipId;
+import com.palantir.lock.v2.Lease;
+import com.palantir.lock.v2.LockToken;
+
+public class LeasedLockTokenTest {
```
j-baker: think there's a fair bit of behaviour in LeasedLockToken which is not tested here; stuff like 'we shouldn't allow an update from a different user', 'we shouldn't allow it to be updated into the past', 'it can be valid past refresh', etc. This test class doesn't really test the contract of the leased lock token.

j-baker: there's stuff around the surrogate id not being the same as the server's lock token

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,226 @@
...
+        timelockService = LeasingTimelockClient.create(timelockRpcClient, SERVICE_ID);
+    }
+
+    @Test
+    public void delegatesLockRequest() {
+        when(timelockRpcClient.lock(lockRequest)).thenReturn(
+                LockResponseV2.successful(LOCK_TOKEN, getLease()));
+
+        timelockService.lock(lockRequest);
+
```
j-baker: whitespace :)

j-baker: also you don't actually verify anything in this test

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,226 @@
...
+    private static final UUID SERVICE_ID = UUID.randomUUID();
+    private StartIdentifiedAtlasDbTransactionRequest startTxnRequest =
+            StartIdentifiedAtlasDbTransactionRequest.createForRequestor(SERVICE_ID);
+
+    private static final LockToken LOCK_TOKEN = LockToken.of(UUID.randomUUID());
+    private static final LockResponse LOCK_RESPONSE = LockResponse.successful(LOCK_TOKEN);
+
+    @Before
+    public void before() {
+        when(timelockRpcClient.getLeaderTime()).thenReturn(LeaderTime.of(LeadershipId.random(), NanoTime.now()));
```
j-baker: red flag: the fact you're not ever changing the time implies that this testing is insufficient. The purpose of writing unit tests like this is to codify the assumptions you're making about how this code should work; the time being static here is a huge red flag to me because it means that all these tests are time invariant (which means that there's no leasing being tested).

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,231 @@
...
+    private static final LeadershipId LEADER_ID = LeadershipId.random();
+    private static final UUID SERVICE_ID = UUID.randomUUID();
+    private StartIdentifiedAtlasDbTransactionRequest startTxnRequest =
+            StartIdentifiedAtlasDbTransactionRequest.createForRequestor(SERVICE_ID);
+
+    private static final LockToken LOCK_TOKEN = LockToken.of(UUID.randomUUID());
+    private static final LockResponse LOCK_RESPONSE = LockResponse.successful(LOCK_TOKEN);
+
+    private TimelockService timelockService;
+    private Supplier<NanoTime> time = NanoTime::now;
```
j-baker: I don't think we should be using real time anywhere here; it just makes the tests unpredictable, and makes it tricky to figure out when you intend time to pass!

j-baker: (the other time was a bit different since it was a bit less unit-test-y)...

lock-api/src/test/java/com/palantir/lock/client/LeasingTimelockClientTest.java
```diff
@@ -0,0 +1,231 @@
...
+    public void unlockShouldCallRemoteServer_validLeases() {
+        LeasedLockToken token = LeasedLockToken.of(LOCK_TOKEN, getLease());
+        assertValid(token);
+        timelockService.unlock(ImmutableSet.of(token));
+
+        verify(timelockRpcClient).unlock(ImmutableSet.of(token.serverToken()));
+    }
+
+    @Test
+    public void unlockShouldCallRemoteServer_inValidLeases() {
```
j-baker: invalid...

PR #3763, opened 2019-02-19
Newer version of jetty alpn agent

j-baker:


j-baker:
This is needed for an internal setup which wants to use a newer JDK version

PR #3764, opened 2019-02-20
Conjure

### LINKED ISSUE #3697: Support conjure-java-runtime
http-remoting has become conjure-java-runtime. Up-to-date consumers of conjure-java-runtime that also use atlas have to do backwards conversions to configure atlas using the classes they have on hand. Atlas should natively support conjure-java-runtime so consumers don't have to do these conversions.

tpetracca:
**Goals (and why)**:
Resolves #3697 
Remove all [un-necessary] references/dependies on remoting, in particular remoting-api and remoting3.

Looked for all references to remoting-api and remoting3 and replaced them with their modern equivalents in `com.palantir.tracing`, `com.palantir.conjure.java.api`, and `com.palantir.conjure.java.runtime`
Also this involves bumping the tracing and tritium versions.

Didn't do anything

My understanding is that we're currently relying on error handling semantics in remoting2 and so this PR has attempted to not touch any remoting2 dependencies.

various `build.gradle` files and `versions.props` I suppose.

High priority. Internal consumers are currently blocked on the fact that the remoting3 version AtlasDb is using is blacklisted, and we want to unblock them picking up a release with #3763 ASAP.


sandorw:
I'm working on the remaining bits here to get things good. They should land tomorrow.

sandorw:
Going to put the alpn changes here as well, since that branch seems to be broken by conflicts with version 2.0.7, which as best I can tell might be coming from remoting.

timelock-server/build.gradle
```diff
@@ -54,5 +57,9 @@ dependencies {
     testCompile group: 'com.squareup.okhttp3', name: 'okhttp'
 
     integTestCompile group: 'io.dropwizard', name: 'dropwizard-testing'
+    integTestCompile group: 'com.palantir.remoting3', name: 'http2-agent'
```
sandorw: I'm considering just pulling the file in rather than maintaining the extra small remoting3 dependency.

timelock.props
```diff
@@ -2,8 +2,6 @@
 # This is supported by Dropwizard starting v1.0.0.
 # http://www.dropwizard.io/1.0.6/docs/manual/configuration.html#http-2-over-tls
 ch.qos.logback:logback-classic = 1.1.7
-com.fasterxml.jackson.*:* = 2.6.7
```
sandorw: I think most of this file should go. Also, lost significant time today since I didn't know this file existed and was confused by the jackson issues :(

sandorw:
Actually, should be able to get this in without the further alpn changes, so we'll rebase once this is fixed.

I think the remaining failures are because we only produce remoting2 style errors from our special exception handlers on Timelock - default exceptions not caught by them won't be parsable by the client. Also, this change is already in the wild, so the fix for now might just be to @ignore the tests and fix them in the near future by ripping out all of this custom remoting logic.

timelock.props
tpetracca: We ran into all the same problems! I was so confused until I read dependencyInsight more clearly and realized this file existed.

+1 on killing this. The dropwizard-jackson downgrade is also very strange.

timelock-server/build.gradle
```diff
@@ -54,5 +54,6 @@ dependencies {
     testCompile group: 'com.squareup.okhttp3', name: 'okhttp'
 
     integTestCompile group: 'io.dropwizard', name: 'dropwizard-testing'
-    integTestCompile group: 'com.palantir.remoting3', name: 'http2-agent'
+    integTestCompile group: 'com.palantir.conjure.java.runtime', name: 'jetty-http2-agent'
+    integTestCompile group: 'org.mortbay.jetty.alpn', name: 'jetty-alpn-agent'
```
tpetracca: this as an artifact of my debugging and I think we should be able to get rid of it

sandorw:
Should be generally good for review. I lightly smoke tested internally and the platform is working against Timelock built on this version.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/Http2Agent.java
```diff
@@ -0,0 +1,58 @@
...
+            AgentLoader.loadAgentClass(Http2Agent.class.getName(), "");
+            hasBeenInstalled = true;
+        } catch (Exception e) {
+            log.warn("Unable to dynamically install jetty-alpn-agent via ea-agent-loader, proceeding anyway...", e);
+        }
+    }
+
+    /** Agent entry-point. */
+    public static void agentmain(String args, Instrumentation inst) throws Exception {
+        Premain.premain(args, inst);
```
jeremyk-91: This is analogous to that found in http-remoting2, so should be correct.

PR #3765, opened 2019-02-20
move requestId to identifiedLockRequest

gokcanozakdag:
**Goals (and why)**:
Currently successive calls on lock with same set of lock descriptors succeeds as long as you have the same `requestId`. As a result, an Atlas client can create a `LockRequest` object, and call `lock()` multiple times, and use any of the returned lockTokens (they should all behave the same). This is a guarantee that I believe we don't want to support, and supporting it will be much harder when we start to use leasedLockTokens.

Rather than having requestId in `LockRequest`, this information is created just before sending the request to timelock, making it transparent to the user of `TimelockService`. So if a user calls `lock()` multiple times with same set of lock descriptors, only one will succeed.

Added a serde test, that uses serialized version of old `LockRequest` creating `IdentifiedLockRequest` to ensure back compat.

I wanted to have `IdentifiedLockRequest` to be package private, to prevent confusing Atlas clients; but this will be used both on client side and server side - so seems like it needs to be public.


lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+
+public class IdentifiedLockRequestTest {
+    private static final String SERIALIZED_LOCK_REQUEST = "{"
+            + "\"requestId\":\"885afd9c-de62-44ff-a517-5db14b71bfaa\","
+            + "\"lockDescriptors\":[{\"bytes\":\"Zm9v\"}],"
+            + "\"acquireTimeoutMs\":123,"
+            + "\"clientDescription\":\"Thread: main\"}";
+
+    private static final ObjectMapper mapper = new ObjectMapper()
```
felixdesouza: you should use the mapper you expect to be used, otherwise this test doesn't really mean anything. Also how much do you care about serde? Is there some backcompat concern that you're trying to uphold? If so, I'd adjust the test name to indicate that.

lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
```diff
@@ -0,0 +1,43 @@
...
+            .registerModule(new Jdk8Module())
+            .registerModule(new GuavaModule());
+
+    @Test
+    public void serializationDeserialization() throws Exception {
+        IdentifiedLockRequest request = mapper.readValue(SERIALIZED_LOCK_REQUEST, IdentifiedLockRequest.class);
+        assertThat(mapper.writeValueAsString(request)).isEqualTo(SERIALIZED_LOCK_REQUEST);
+    }
```
jeremyk-91: I realise the implementation works, but I'd prefer we test this a different way, because comparing string equality imposes an ordering on the fields, which I don't think is guaranteed by an ObjectMapper without specific annotations (e.g. http://fasterxml.github.io/jackson-annotations/javadoc/2.3.0/com/fasterxml/jackson/annotation/JsonPropertyOrder.html)

lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
jeremyk-91: This is a backcompat test I think (specifically, that an old `LockRequest` serializes to the new `IdentifiedLockRequest`)

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -297,18 +299,35 @@ public void testCurrentTimeMillis() throws InterruptedException {
...
+        ));
 
-        assertThat(response.join()).isEqualTo(duplicateResponse.join());
+        cluster.timelockRpcClient().unlock(ImmutableSet.of(token));
+
+        assertThat(response).isEqualTo(duplicateResponse);
+    }
+
+    @Test
+    public void lockSuccedsOnlyOnce() {
```
jeremyk-91: nit :`Succeeds`

lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
gokcanozakdag: this is a backcompat test; I can rename this to make it more clear.
On object mapper part, I think we actually create object mappers and register modules on server and client side separately - there isn't one single ObjectMapper template used everywhere. This may be something we want to fix though (probably in a separate PR)

felixdesouza: if so, then I'd recommend serialising with the client object mapper we use, and then deserialising with the server object mapper we use, as this object mapper is quite far from what we use and not representative if we're doing a back compat test.

lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
felixdesouza: why do we care about serialising the old thing? we're not doing that anymore right? Isn't it sufficient to just check that the `SERIALIZED_LOCK_REQUEST` *deserialises* to something that we can understand and expect? Unless I'm mistaken

gokcanozakdag: problem is you need to upgrade atlas versions on both the client and the timelock server, which is not possible. So we need to make sure we can serialize and deserialize from old to new version (old atlas client, new timelock server), and vice versa (old timelock server, new atlas client)

lock-api/src/test/java/com/palantir/lock/client/IdentifiedLockRequestTest.java
gokcanozakdag: problem is we don't have a well defined client and server object mappers. It is being created like this: https://github.com/palantir/atlasdb/blob/develop/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbFeignTargetFactory.java#L53 (so currently there is no way to ensure we are using the same object mapper in the tests here)

I still think this is something we need to refactor though.

lock-api/src/main/java/com/palantir/lock/client/IdentifiedLockRequest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Value.Parameter
+    Optional<String> getClientDescription();
+
+    static IdentifiedLockRequest of(Set<LockDescriptor> lockDescriptors, long acquireTimeoutMs) {
+        return ImmutableIdentifiedLockRequest.of(
+                UUID.randomUUID(),
+                lockDescriptors,
+                acquireTimeoutMs,
+                Optional.of("Thread: " + Thread.currentThread().getName()));
```
j-baker: do we use this information anywhere? this stuff isn't obviously safe, so it's also not obviously useful to us

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -297,18 +299,35 @@ public void testCurrentTimeMillis() throws InterruptedException {
...
     @Test
-    public void lockRequestsAreIdempotent() {
-        LockToken token = cluster.lock(requestFor(LOCK_A)).getToken();
+    public void lockRequestsToRpcClientAreIdempotent() {
+        IdentifiedLockRequest request = IdentifiedLockRequest.from(requestFor(LOCK_A));
 
-        LockRequest request = requestFor(LOCK_A);
-        CompletableFuture<LockResponse> response = cluster.lockAsync(request);
-        CompletableFuture<LockResponse> duplicateResponse = cluster.lockAsync(request);
+        LockResponseV2 response = cluster.timelockRpcClient().lock(request);
```
j-baker: the previous test was more interesting for our purposes; one tests whether if i lock something that's already locked that it hands back the same token, the other makes sure that when a _different_ holder of the lock unlocks it, that both requestors acquire successfully. Basically, I think it's a less obvious constraint to enforce, seems like an obvious bug to accidentally add, and so I prefer the old test

lock-api/src/main/java/com/palantir/lock/client/IdentifiedLockRequest.java
gokcanozakdag: I agree, I think it makes sense to have Optional.empty if clientDescription is not supplied.

PR #3767, opened 2019-02-21
Fix Fragile Timelock Test

gmaretic:
**Goals (and why)**:
Fix the interference between `AsyncTimelockServiceIntegrationTest` and `AsyncTimelockServiceTransactionIntegrationTest` where they are talking to timelock as the same agent. When tests are executed in parallel, it's possible for one to start a transaction, messing up the immutable timestamp test in the other test class.

This could have caused the test failure in run #15635.

Change the config of the transaction manager to avoid the clash

Are there any other concerns I may have missed?


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -74,7 +74,10 @@ public TestableTimelockCluster(String baseUri, String defaultClient, String... c
     }
 
     public void waitUntilLeaderIsElected() {
```
felixdesouza: I think we should get rid of this, and force everyone to specify an agent.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
```diff
@@ -40,7 +42,7 @@ static TransactionManager createTransactionManager(TestableTimelockCluster clust
                 .map(server -> server.serverHolder().getTimelockUri())
                 .collect(Collectors.toList());
         AtlasDbConfig config = ImmutableAtlasDbConfig.builder()
-                .namespace("test")
```
felixdesouza: ha, agent smith ;)

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
```diff
@@ -31,6 +31,8 @@
 import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
 
 public final class TimeLockTestUtils {
+    public static final String AGENT = "smith";
```
felixdesouza: I don't recommend making this public. Basically, we want every invocation to create a corresponding timelock cluster or whatever to return a separate namespaced instance.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
gmaretic: Went the other way so you cannot specify the default agent, which is a random uuid

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
gmaretic: I'm so cool

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
gmaretic: Should be good now?

gmaretic:
@felixdesouza now generating a random default agent

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -54,15 +55,14 @@
 
     private final TemporaryFolder temporaryFolder = new TemporaryFolder();
 
-    private final String defaultClient;
+    private final String defaultClient = UUID.randomUUID().toString();
```
felixdesouza: nit, let's not call this `defaultClient`, since there's no way for you to change it ;)

`client` will do

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -74,26 +74,30 @@ public TestableTimelockCluster(String baseUri, String defaultClient, String... c
...
 
-    public void waitUntilReadyToServeClients(List<String> clients) {
+    private void waitUntilReadyToServeClients(List<String> clients) {
         Awaitility.await()
                 .atMost(60, TimeUnit.SECONDS)
                 .pollInterval(500, TimeUnit.MILLISECONDS)
                 .until(() -> {
                     try {
                         clients.forEach(client -> timelockServiceForClient(client).getFreshTimestamp());
+                        timelockServiceForClient(defaultClient).getFreshTimestamp();
```
jeremyk-91: nit: seems a bit unexpected to me that `waitUntilReadyToServeClients(clients)` also talks to the default client, I'd suggest either having the factories pass `ImmutableList.of(defaultClient)` and [defaultClient + additionalClients] (don't remember efficient syntax offhand), or keep the parameter named `additionalClients` here

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceTransactionIntegrationTest.java
```diff
@@ -53,6 +53,7 @@
     private static final TableReference TABLE = TableReference.create(Namespace.create("test"), "test");
     private static final byte[] DATA = "foo".getBytes();
     private static final Cell CELL = Cell.create("bar".getBytes(), "baz".getBytes());
+    private static final String AGENT = "smith";
```
jeremyk-91: !

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
```diff
@@ -51,7 +56,7 @@ static TransactionManager createTransactionManager(TestableTimelockCluster clust
                 .build();
         return TransactionManagers.builder()
                 .config(config)
-                .userAgent("test")
```
jeremyk-91: I don't think this was meant to be connected with the `"test"` for the namespace, but sure

PR #3769, opened 2019-02-21
Remove request id from lockImmutableTimestamp and startAtlasTxn from service interface

gokcanozakdag:
**Goals (and why)**:
Similar to #3765, preventing clients to call `lockImmutableTimestamp()` and `startIdentifiedAtlasDbTransaction()` with the same `IdentifiedTimelockRequest`.

Remove the requestId information from `TimelockService` interface, and create it just before calling `TimelockRpcClient`

I have removed the `startIdentifiedAtlasDbTransaction(IdentifiedTimeLockRequest request)` all together as I don't think it is being used outside of Atlas, but did not remove `lockImmutableTimestamp(IdentifiedTimeLockRequest request)`. I only deprecated the method, and added a default implementation that redirects to `lockImmutableTimestamp()`.


lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -40,10 +40,18 @@ default boolean isInitialized() {
 
     TimestampRange getFreshTimestamps(@Safe @QueryParam("number") int numTimestampsRequested);
 
+    /**
+     * @deprecated use {@link #lockImmutableTimestamp()} instead.
+     */
+    @Deprecated
+    default LockImmutableTimestampResponse lockImmutableTimestamp(IdentifiedTimeLockRequest request) {
+        return lockImmutableTimestamp();
+    }
```
jeremyk-91: I'm not sure if it's better if we keep this method around or just force a dev break on `lockImmutableTimestamp` users, since the semantics of this are different from the previous method.

gokcanozakdag: I am more inclined towards forcing a dev break - given that this should have a very narrow use.

j-baker: agreed

PR #3774, opened 2019-02-25
[_transactions2] Part 22: Reverse Bits, not Bytes

jeremyk-91:
**Goals (and why)**:
- Ensure better distribution of timestamps by reversing the bits in the row key as opposed to bytes. This went under the radar because nodes contribute multiple tokens in consistent hashing, but I felt was against the spirit of the tickets algorithm.
- Previously, the active rows `1` through `16` would be written to the bytes `00` through `0f`; we'd then go through `10` through `1f` and so on, so the bits that were changing were actually 5th through 8th rather than the highest order ones.

- Reverse bits rather than bytes, via a new `EncodingUtils` method.

- Added new tests for the `EncodingUtils` method, and a test for `TicketsEncodingStrategy` that adjacent rows are distributed across byte-space. We previously had a test that checked that cells were distributed across rows (but these rows would have been not that far away from each other in byte-space).

- Is there an easier way to implement the `EncodingUtils` method? A cursory search didn't reveal anything.
- Breaks wire format, but no one should actually be using transactions 2 yet so I think this is a good time to do this.

**Where should we start reviewing?**: EncodingUtils

**Priority (whenever / two weeks / yesterday)**: this week, ideally tomorrow.

@amandachow for SA

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/ptobject/EncodingUtils.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/ptobject/EncodingUtils.java
@@ -464,4 +464,21 @@ public class EncodingUtils {
             return ArrayUtils.addAll(new byte[] { 1 }, PtBytes.toBytes(Long.MIN_VALUE ^ value));
         }
     }
+
+    public static byte[] reverseBits(byte[] value) {
+        ArrayUtils.reverse(value);
+        for (int index = 0; index < value.length; index++) {
+            value[index] = reverseBitsInByte(value[index]);
+        }
+        return value;
+    }
+
+    private static byte reverseBitsInByte(byte value) {
+        byte result = 0;
+        for (int index = Byte.SIZE - 1; index >= 0; index--) {
+            result += (value & 1) << index;
+            value >>= 1;
+        }
+        return result;
+    }
 }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/ptobject/EncodingUtils.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/ptobject/EncodingUtils.java
@@ -464,21 +464,4 @@ public class EncodingUtils {
             return ArrayUtils.addAll(new byte[] { 1 }, PtBytes.toBytes(Long.MIN_VALUE ^ value));
         }
     }
-
-    public static byte[] reverseBits(byte[] value) {
-        ArrayUtils.reverse(value);
-        for (int index = 0; index < value.length; index++) {
-            value[index] = reverseBitsInByte(value[index]);
-        }
-        return value;
-    }
-
-    private static byte reverseBitsInByte(byte value) {
-        byte result = 0;
-        for (int index = Byte.SIZE - 1; index >= 0; index--) {
-            result += (value & 1) << index;
-            value >>= 1;
-        }
-        return result;
-    }
 }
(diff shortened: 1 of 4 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/encoding/TicketsEncodingStrategy.java
```diff
@@ -33,7 +32,7 @@
  * least significant bits of the timestamp and using that as the row number. For example, we would store timestamps
  * 1, ROW_PER_QUANTUM + 1, 2 * ROW_PER_QUANTUM + 1 etc. in the same row.
  *
- * We store the row name as a little-endian representation of the row number to ensure even distribution in key-value
+ * We store the row name as a bit-wise reversed version the row number to ensure even distribution in key-value
```
jeremyk-91: oops: `bit-wise reversed version of`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/encoding/TicketsEncodingStrategy.java
```diff
@@ -91,7 +90,7 @@ public long decodeValueAsCommitTimestamp(long startTimestamp, byte[] value) {
         long row = (startTimestamp / PARTITIONING_QUANTUM) * ROWS_PER_QUANTUM
                 + (startTimestamp % PARTITIONING_QUANTUM) % ROWS_PER_QUANTUM;
         byte[] rowName = ValueType.VAR_LONG.convertFromJava(row);
-        ArrayUtils.reverse(rowName);
+        EncodingUtils.reverseBits(rowName);
```
j-baker: I might be minded to use a fixed long for this (since after all if it's going in the row key, it won't show up in the index summary much) and then just use Long.reverse

j-baker: then it's just Long.reverse(PtBytes.toBytes(row)) and Long.reverse(PtBytes.fromBytes(row))

j-baker: sorry, realised that first comment was bad. The row key is a bit irrelevant for dynamic columns rows in terms of bytes, because it's only stored every few kB data; so the difference between this is well well under 1%.

atlasdb-client/src/main/java/com/palantir/atlasdb/ptobject/EncodingUtils.java
```diff
@@ -464,4 +464,21 @@ public static Long decodeFlippedNullableFixedLong(byte[] value, int offset) {
...
+        ArrayUtils.reverse(value);
+        for (int index = 0; index < value.length; index++) {
+            value[index] = reverseBitsInByte(value[index]);
+        }
+        return value;
+    }
+
+    private static byte reverseBitsInByte(byte value) {
+        byte result = 0;
+        for (int index = Byte.SIZE - 1; index >= 0; index--) {
```
j-baker: these methods often turn out to be weirdly slow (if you load 1k timestamps out of Cassandra, you'll need to spend legit a few million timestamps (1000 * probably 4 cycles per loop iteration * Long.SIZE)) just decoding.

j-baker: ok so that math was clearly very blatantly wrong

j-baker: but it's still millions of cycles a second on some of our larger stacks. Basically, in cases like this it's worth making a `byte[]` which just does the mapping for you for all byte values.

jeremyk-91: Ah, yeah a lookup table makes sense since there are only 256 possibilities - though I agree on `Long.reverse`, I was looking for a library method to do this stuff but had no idea that was a thing

PR #3775, opened 2019-02-26
a proof of concept implementation for shared start txn calls

gokcanozakdag:
**Goals (and why)**:
To demonstrate one of the ways of implementing shared start txn calls.

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/HeldImmutableTsToken.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/HeldImmutableTsToken.java
@@ -1,0 +1,55 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import java.util.UUID;
+
+import com.palantir.lock.v2.LockToken;
+
+class HeldImmutableTsToken implements LockToken {
+    private UUID requestId;
+    private SharedImmutableTsToken sharedImmutableTsToken;
+    private volatile boolean invalidated;
+
+    @Override
+    public UUID getRequestId() {
+        return requestId;
+    }
+
+    //adding without marking?
+    HeldImmutableTsToken(SharedImmutableTsToken sharedImmutableTsToken, UUID id) {
+        this.requestId = id;
+        this.sharedImmutableTsToken = sharedImmutableTsToken;
+        this.invalidated = false;
+    }
+
+    static HeldImmutableTsToken of(SharedImmutableTsToken sharedImmutableTsToken) {
+        sharedImmutableTsToken.mark();
+        return new HeldImmutableTsToken(sharedImmutableTsToken, UUID.randomUUID());
+    }
+
+    synchronized void invalidate() {
+        if (!invalidated) {
+            invalidated = true;
+            sharedImmutableTsToken.unmark();
+        }
+    }
+
+    SharedImmutableTsToken sharedToken() {
+        return sharedImmutableTsToken;
+    }
+}
--- a/lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockClient.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockClient.java
@@ -1,0 +1,98 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import java.util.Set;
+
+import com.palantir.lock.v2.LockImmutableTimestampResponse;
+import com.palantir.lock.v2.LockRequest;
+import com.palantir.lock.v2.LockResponse;
+import com.palantir.lock.v2.LockToken;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockRpcClient;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.lock.v2.WaitForLocksRequest;
+import com.palantir.lock.v2.WaitForLocksResponse;
+import com.palantir.timestamp.TimestampRange;
+
+public class RemoteTimelockClient implements TimelockService {
+    private LeasingTimelockClient leasingTimelockClient;
+    private TimelockTransactionService timelockTransactionService;
+    private TimelockRpcClient rawClient;
+
+    RemoteTimelockClient(TimelockRpcClient rawClient) {
+        this.rawClient = rawClient;
+        this.leasingTimelockClient = LeasingTimelockClient.create(rawClient);
+        this.timelockTransactionService = new SimpleTimelockTransactionService(leasingTimelockClient);
+    }
+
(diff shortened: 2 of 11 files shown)


lock-api/src/main/java/com/palantir/lock/client/TimelockTransactionService.java
```diff
@@ -0,0 +1,33 @@
...
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+
+//better name
+
+/**
+ * responsible for managing shared start txn calls
+ */
+public interface TimelockTransactionService {
+    StartIdentifiedAtlasDbTransactionResponse startTransaction(); //should coalesce concurrent calls
+    Set<LockToken> reduceForRefresh(Set<LockToken> tokens);
```
gokcanozakdag: these are just static methods, probably should be moved to somewhere else

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
```diff
@@ -0,0 +1,83 @@
...
+public class SimpleTimelockTransactionService implements TimelockTransactionService {
+    private LeasingTimelockClient delegate;
+
+    public SimpleTimelockTransactionService(LeasingTimelockClient delegate) {
+        this.delegate = delegate;
+    }
+
+
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponse startTransaction() {
```
gokcanozakdag: simpleTxnService just delegates the startTxn call (and wraps it) without coalescing. CoalescingTxnService will be using AutoBatcher to batch these calls into a single startTxns call (this endpoint doesn't exist right now)

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockClient.java
```diff
@@ -0,0 +1,98 @@
...
+        return leasingTimelockClient.lock(request);
+    }
+
+    @Override
+    public WaitForLocksResponse waitForLocks(WaitForLocksRequest request) {
+        return rawClient.waitForLocks(request);
+    }
+
+    @Override
+    public Set<LockToken> refreshLockLeases(Set<LockToken> tokens) {
```
j-baker: i think these two methods are both critically broken

lock-api/src/main/java/com/palantir/lock/client/SharedImmutableTsToken.java
```diff
@@ -0,0 +1,53 @@
...
+        //precondition
+        reference--;
+    }
+
+    synchronized boolean dereferenced() {
+        return reference == 0;
+    }
+
+    LockToken token() {
+        return token;
```
j-baker: this should probably throw if the lock is already reference counted out

lock-api/src/main/java/com/palantir/lock/client/SharedImmutableTsToken.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.UUID;
+
+import com.palantir.lock.v2.LockToken;
+
+class SharedImmutableTsToken implements LockToken {
+    private LockToken token;
+    private volatile int reference;
+
+    SharedImmutableTsToken(LockToken token) {
```
j-baker: think it's worth this taking the reference count in its constructor and only allowing it to be decremented. It means you can't get to zero and then go back up to 1

lock-api/src/main/java/com/palantir/lock/client/SharedImmutableTsToken.java
```diff
@@ -0,0 +1,53 @@
...
+    synchronized void mark() {
+        reference++;
+    }
+
+    synchronized void unmark() {
+        //precondition
+        reference--;
+    }
+
+    synchronized boolean dereferenced() {
```
j-baker: this returns the wrong result if you call unmark too many times

lock-api/src/main/java/com/palantir/lock/client/SharedImmutableTsToken.java
```diff
@@ -0,0 +1,53 @@
...
+    SharedImmutableTsToken(LockToken token) {
+        this.token = token;
+        reference = 0;
+    }
+
+    synchronized void mark() {
+        reference++;
+    }
+
+    synchronized void unmark() {
```
j-baker: probably call this release

lock-api/src/main/java/com/palantir/lock/client/SharedImmutableTsToken.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.palantir.lock.client;
+
+import java.util.UUID;
+
+import com.palantir.lock.v2.LockToken;
+
+class SharedImmutableTsToken implements LockToken {
+    private LockToken token;
+    private volatile int reference;
```
j-baker: this doesn't need to be volatile

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public Set<LockToken> reduceForUnlock(Set<LockToken> tokens) {
+        Set<HeldImmutableTsToken> immutableTsTokens = filterImmutableTsTokens(tokens);
+        immutableTsTokens.forEach(HeldImmutableTsToken::invalidate);
+        Set<LockToken> toUnlock;
+        synchronized (this) {
+            toUnlock = immutableTsTokens.stream().map(h -> h.sharedToken())
+                    .filter(SharedImmutableTsToken::dereferenced)
```
j-baker: don't think there's any place these ever get dereferenced?

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private Set<HeldImmutableTsToken> filterImmutableTsTokens(Set<LockToken> tokens) {
+        return tokens.stream().filter(t -> t instanceof HeldImmutableTsToken)
+                .map(t -> (HeldImmutableTsToken) t)
+                .collect(Collectors.toSet());
+    }
+
+    private Set<LockToken> filterOutImmutableTsTokens(Set<LockToken> tokens) {
+        return tokens.stream().filter(t -> !(t instanceof HeldImmutableTsToken))
```
j-baker: probably worth having a function boolean isImmutableLockToken?

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockClient.java
j-baker: worth thinking through what the contract of these methods is

lock-api/src/main/java/com/palantir/lock/client/HeldImmutableTsToken.java
```diff
@@ -0,0 +1,55 @@
...
+package com.palantir.lock.client;
+
+import java.util.UUID;
+
+import com.palantir.lock.v2.LockToken;
+
+class HeldImmutableTsToken implements LockToken {
+    private UUID requestId;
+    private SharedImmutableTsToken sharedImmutableTsToken;
+    private volatile boolean invalidated;
```
j-baker: probably doesn't need to be invalidated

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
j-baker: ah, i found it

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
```diff
@@ -0,0 +1,83 @@
...
+        Set<HeldImmutableTsToken> immutableTsTokens = filterImmutableTsTokens(tokens);
+        Set<LockToken> reducedImmutableTsTokens = immutableTsTokens.stream()
+                .map(h -> h.sharedToken().token())
+                .collect(Collectors.toSet());
+
+        return Sets.union(filterOutImmutableTsTokens(tokens), reducedImmutableTsTokens);
+    }
+
+    @Override
+    public Set<LockToken> reduceForUnlock(Set<LockToken> tokens) {
```
j-baker: This code is very... I don't know the right word. To me, it does a bit too much reaching inside the state of other code in order to figure out what to do; it means that code that should be owned somewhere else is owned here.

As an example, why doesn't unlock on HeldImmutableTsToken return an `Optional<LockToken>` which is the LockToken which should be passed to the underlying lock service? and it does that only the last time? In that case this class doesn't have to be aware of the shared tokens, and the synchronized can vanish :)

lock-api/src/main/java/com/palantir/lock/client/SimpleTimelockTransactionService.java
```diff
@@ -0,0 +1,83 @@
...
+        Set<HeldImmutableTsToken> immutableTsTokens = filterImmutableTsTokens(tokens);
+        immutableTsTokens.forEach(HeldImmutableTsToken::invalidate);
+        Set<LockToken> toUnlock;
+        synchronized (this) {
```
j-baker: red flag for you: a synchronized call which doesn't wrap calls to multiple fields probably implies you're synchronizing in the wrong place.

PR #3776, opened 2019-02-26
[_transactions2] Part 23: Tuning Tables for Cassandra

jeremyk-91:
> I can't give you the heart you think you gave me;
It's time to say goodbye, to tuning tables

(with apologies)

**Goals (and why)**:
- Provide good Cassandra-level table settings to maintain read performance as we move from transactions1 to transactions2.

- The core of this change is adding a boolean `denselyAccessedWideRows()` setting to table metadata. This is not exposed in schema definition.
- If this setting is specified, it currently does nothing for In-Memory and DBKVS. However, tables created in Cassandra will have the following parameters changed, based on benchmarking:
  - `bloom_filter_fp_chance` is set to 0.0001 (default: 0.1). This uses more memory, but avoids disk seeks when reading from SSTables
  - `min_index_interval` is set to 1 (default: 256), allowing for fewer disk seeks when reading from disk
  - `chunk_length_kb` in compression is set to 64 (default: 4), allowing for smaller data overall; by design, the data written by transactions2 compresses down very well

Not much previous testing. I added tests for some refactors of the cf definitions, and an integration test overall that the table was created with the correct constants if we subsequently get the `CfDef` from Cassandra.

- Are the tests too awkward/strange to work with?
- Possibly not in scope here, but are there better ways to enforce consistency across `ColumnFamilyDefinitions` and `CassandraTableCreator`?
- More importantly, are there any changes apart from to those two classes mentioned above that need to be added?

**Where should we start reviewing?**: CassandraTableCreator

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableCreationIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableCreationIntegrationTest.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.atlasdb.keyvalue.cassandra;
 
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 
@@ -25,6 +26,9 @@ import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 
+import org.apache.cassandra.thrift.CfDef;
+import org.apache.cassandra.thrift.KsDef;
+import org.apache.thrift.TException;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
@@ -33,16 +37,19 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.MoreCollectors;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cassandra.ImmutableCassandraKeyValueServiceConfig;
 import com.palantir.atlasdb.containers.CassandraResource;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.keyvalue.impl.AbstractKeyValueService;
 import com.palantir.atlasdb.protos.generated.TableMetadataPersistence;
 import com.palantir.atlasdb.table.description.TableDefinition;
 import com.palantir.atlasdb.table.description.ValueType;
 import com.palantir.atlasdb.transaction.api.ConflictHandler;
+import com.palantir.atlasdb.transaction.impl.TransactionConstants;
 
 public class CassandraKeyValueServiceTableCreationIntegrationTest {
     private static final TableReference GOOD_TABLE = TableReference.createFromFullyQualifiedName("foo.bar");
@@ -164,6 +171,29 @@ public class CassandraKeyValueServiceTableCreationIntegrationTest {
         assertThat(initialMetadata, is(existingMetadata));
 
         kvs.dropTable(caseSensitiveTable);
+    }
+
+    @Test
+    public void testCreateTransactions2TableProducesCorrectMetadata() throws TException {
+        kvs.createTable(
+                TransactionConstants.TRANSACTIONS2_TABLE,
+                TransactionConstants.TRANSACTIONS2_TABLE_METADATA.persistToBytes());
+
+        KsDef ksDef = kvs.getClientPool()
+                .run(client -> client.describe_keyspace(CASSANDRA.getConfig().getKeyspaceOrThrow()));
+        CfDef transactions2CfDef = ksDef.cf_defs.stream()
+                .filter(cfDef -> cfDef.name.equals(
+                        AbstractKeyValueService.internalTableName(TransactionConstants.TRANSACTIONS2_TABLE)))
+                .collect(MoreCollectors.onlyElement());
+        assertThat(transactions2CfDef.bloom_filter_fp_chance, equalTo(
+                CassandraConstants.DENSELY_ACCESSED_WIDE_ROWS_BLOOM_FILTER_FP_CHANCE));
+        assertThat(transactions2CfDef.min_index_interval,
+                equalTo(CassandraConstants.DENSELY_ACCESSED_WIDE_ROWS_INDEX_INTERVAL));
+        assertThat(transactions2CfDef.max_index_interval,
+                equalTo(CassandraConstants.DENSELY_ACCESSED_WIDE_ROWS_INDEX_INTERVAL));
+        assertThat(transactions2CfDef.compression_options.get(CassandraConstants.CFDEF_COMPRESSION_CHUNK_LENGTH_KEY),
+                equalTo(String.valueOf(
+                        TransactionConstants.TRANSACTIONS2_TABLE_METADATA.getExplicitCompressionBlockSizeKB())));
     }
 
     private static CassandraKeyValueService kvsWithSchemaMutationTimeout(int millis) {
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
@@ -16,6 +16,7 @@
 package com.palantir.atlasdb.keyvalue.cassandra;
 
 import java.util.Map;
+import java.util.Optional;
 
 import org.apache.cassandra.thrift.CfDef;
 import org.slf4j.Logger;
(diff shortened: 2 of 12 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -83,6 +83,15 @@ public LogSafety getNameLogSafety() {
         return LogSafety.UNSAFE;
     }
 
+    /**
+     * @return whether the table has densely accessed wide rows. This helps identify tables which could benefit from
+     * setting database tuning parameters to values more in line with such workflows, like the _transactions2 table.
+     */
+    @Value.Default
+    public boolean hasDenselyAccessedWideRows() {
+        return false;
```
felixdesouza: out of curiosity, if we were to expose this in the schema, is it safe to change after the fact? i.e. with a preexisting table?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -213,6 +203,18 @@ static boolean isMatchingCf(CfDef clientSide, CfDef clusterSide) {
                     clusterSide.isSetPopulate_io_cache_on_flush());
             return false;
         }
+        if (clientSide.min_index_interval != clusterSide.min_index_interval) {
```
felixdesouza: do you not want to return false in both of these cases?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -115,11 +82,34 @@ static CfDef getCfDef(String keyspace, TableReference tableRef, int gcGraceSecon
...
+                CassandraConstants.DEFAULT_COMPRESSION_TYPE);
+
+        int explicitCompressionBlockSizeKb = tableMetadata.map(TableMetadata::getExplicitCompressionBlockSizeKB)
+                .orElse(0);
+        int actualCompressionBlockSizeKb = explicitCompressionBlockSizeKb == 0
+                ? AtlasDbConstants.MINIMUM_COMPRESSION_BLOCK_SIZE_KB
+                : explicitCompressionBlockSizeKb;
+        compressionOptions.put(CassandraConstants.CFDEF_COMPRESSION_CHUNK_LENGTH_KEY,
+                Integer.toString(actualCompressionBlockSizeKb));
+        return compressionOptions;
```
felixdesouza: can we put the map building code together?

e.g.

```java
int explicitCompressionBlockSizeKb = tableMetadata.map(TableMetadata::getExplicitCompressionBlockSizeKB)
                .orElse(0);
int actualCompressionBlockSizeKb = explicitCompressionBlockSizeKb == 0
                ? AtlasDbConstants.MINIMUM_COMPRESSION_BLOCK_SIZE_KB
                : explicitCompressionBlockSizeKb;
return ImmutableMap.<String, String>builder()
  .put(CassandraConstants.CFDEF_COMPRESSION_TYPE_KEY, CassandraConstants.DEFAULT_COMPRESSION_TYPE)
  .put(CassandraConstants.CFDEF_COMPRESSION_CHUNK_LENGTH_KEY, Integer.toString(actualCompressionBlockSizeKb))
  .build();
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -47,57 +48,23 @@ private ColumnFamilyDefinitions() {
...
+                : Optional.of(TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(rawMetadata));
 
-        if (negativeLookups) {
-            falsePositiveChance = CassandraConstants.NEGATIVE_LOOKUPS_BLOOM_FILTER_FP_CHANCE;
-        }
+        Map<String, String> compressionOptions = getCompressionOptions(tableMetadata);
+        cf.setCompression_options(compressionOptions);
 
-        if (appendHeavyAndReadLight) {
+        if (tableMetadata.map(TableMetadata::isAppendHeavyAndReadLight).orElse(false)) {
```
felixdesouza: I'm somewhat biased to not inlining these, and having readable variable names.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableCreationIntegrationTest.java
```diff
@@ -166,6 +173,29 @@ public void testGetMetadataCaseInsensitive() {
         kvs.dropTable(caseSensitiveTable);
     }
 
+    @Test
+    public void testCreateTransactions2TableProducesCorrectMetadata() throws TException {
```
felixdesouza: it's somewhat a weird test imo, what happens when you need to change it? you'll just edit the test. Happy to keep here, but unclear what value it has.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
felixdesouza: the same for all the `Optional.map(...).orElse(...)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
jeremyk-91: yep, I do - good spot!

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
jeremyk-91: Yes, though the changes to the bloom filters / index summary etc. won't be applied to existing SSTables. The next time cassandra compacts those tables, it will write them out in the new format (so if you want to force it to happen now, as we did in benchmarks, you would run `nodetool compact`)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
jeremyk-91: (ah, and the tests didn't catch these because they found _the first difference_ between the dense/wide-row tables and the other ones)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
jeremyk-91: OK, extracted each of the map/orElse expressions to a named variable. Makes sense

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableCreationIntegrationTest.java
jeremyk-91: This was meant to test the changes to `CassandraTableCreator` (to set the new options) which itself doesn't have tests right now. So if I say forgot to write line 91 (`SchemaBuilder....maxIndexInterval(...)`) for the table-creator or someone removes it later on, this test should catch that. We could conceivably check that the CQL query there is well formed, though I think an integration test would remain useful.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -83,6 +83,15 @@ public LogSafety getNameLogSafety() {
         return LogSafety.UNSAFE;
     }
 
+    /**
+     * @return whether the table has densely accessed wide rows. This helps identify tables which could benefit from
+     * setting database tuning parameters to values more in line with such workflows, like the _transactions2 table.
+     */
+    @Value.Default
+    public boolean hasDenselyAccessedWideRows() {
```
gmaretic: We should either not persist this when default, or make sure to change the CassandraKVSImpl table creation code which currently updates table metadata for a table if the entry persisted in the _metadata table differs from the new serialized table metadata. As is, this would make us rewrite table metadata for all tables in our schemas

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
felixdesouza: interesting, nice to know!

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableCreationIntegrationTest.java
felixdesouza: That's fair.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/ColumnFamilyDefinitions.java
```diff
@@ -58,7 +58,8 @@ static CfDef getCfDef(String keyspace, TableReference tableRef, int gcGraceSecon
         Map<String, String> compressionOptions = getCompressionOptions(tableMetadata);
         cf.setCompression_options(compressionOptions);
 
-        if (tableMetadata.map(TableMetadata::isAppendHeavyAndReadLight).orElse(false)) {
+        boolean appendHeavyAndReadLight = tableMetadata.isPresent() && tableMetadata.get().isAppendHeavyAndReadLight();
```
felixdesouza: nit: `tableMetadata.map(TableMetadata::isAppendHeavyAndReadLight).orElse(false)` ??

jeremyk-91: yeah, sure. Not sure why I changed this one back

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/TableMetadata.java
```diff
@@ -165,7 +165,9 @@ public Builder dynamicColumns(List<NameComponentDescription> components, ValueTy
         builder.setSweepStrategy(getSweepStrategy());
         builder.setAppendHeavyAndReadLight(isAppendHeavyAndReadLight());
         builder.setNameLogSafety(getNameLogSafety());
-        builder.setDenselyAccessedWideRows(hasDenselyAccessedWideRows());
+        if (hasDenselyAccessedWideRows()) {
+            builder.setDenselyAccessedWideRows(hasDenselyAccessedWideRows());
```
felixdesouza: out of curiosity, are you going to make the change to compare objects instead of byte arrays at a future point in time?

jeremyk-91: Discussed offline with @gmaretic, I think we'll do it now actually

jeremyk-91: Turns out this is not as trivial as I thought because of an edge case where if you want to write generic table metadata, you won't. I'll revert to the old method for now.

PR #3780, opened 2019-02-27
Replace commit profile processor with tracing [no release notes]

j-baker:
This was useful, but really should be traced instead. The advantages of
tracing are that it doesn't clutter the logs, is simpler code (methods
that return stuff are no longer overly complex), and provides the
information in a context sensitive fashion.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1370,104 +1365,75 @@ private void commitWrites(TransactionService transactionService) {
...
-                    .verifyUserPreCommitConditionMicros(microsForUserPreCommitCondition)
-                    .putCommitTimestampMicros(microsForPutCommitTs)
-                    .commitTimestamp(commitTimestamp)
-                    .totalCommitStageMicros(TimeUnit.NANOSECONDS.toMicros(commitStageTimer.stop()))
-                    .totalTimeSinceTransactionCreationMicros(microsSinceCreation)
-                    .build());
-        } finally {
-            long microsForPostCommitUnlock = runAndReportTimeAndGetDurationMicros(
-                    () -> timelockService.tryUnlock(ImmutableSet.of(commitLocksToken)), "postCommitUnlock");
+        timedAndTraced("commitStage", () -> {
```
jeremyk-91: One thing this changes, if I'm reading it correctly, is that unsuccessful operations now go into the timer metrics because of try-with-resources on the timer context. This is probably reasonable - not tracking these previously was probably unintentional.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1370,104 +1365,75 @@ private void commitWrites(TransactionService transactionService) {
...
+
+                long microsSinceCreation = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis() - timeCreated);
+                getTimer("commitTotalTimeSinceTxCreation").update(microsSinceCreation, TimeUnit.MICROSECONDS);
+                getHistogram(AtlasDbMetricNames.SNAPSHOT_TRANSACTION_BYTES_WRITTEN).update(byteCount.get());
+            } finally {
+                timedAndTraced("postCommitUnlock",
+                        () -> timelockService.tryUnlock(ImmutableSet.of(commitLocksToken)));
+            }
+        });
+    }
```
jeremyk-91: I've verified correspondence between the original flow of steps and what we have here, which looks right.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
j-baker: yep, happy with

PR #3782, opened 2019-02-27
introducing shared lock tokens

gokcanozakdag:
**Goals (and why)**:
Given that we would like to share immutable lock tokens in multiple concurrently running transactions, we need a class that encapsulates the contract of a shared lock token.

Multiple `SharedLockToken`s can point to same `LockToken`. Unlocking a shared lock token should only unlock the referenced token iff all references are already unlocked.


lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.palantir.lock.v2.LockToken;
+
+public class SharedLockTokenTest {
+    private static final LockToken LOCK_TOKEN = LockToken.of(UUID.randomUUID());
+
+    @Test
+    public void allSharedLocksHaveCorrectUnderlyingToken() {
+        List<LockToken> tokens = SharedLockToken.share(LOCK_TOKEN, 3);
+        convertToSharedTokens(tokens).forEach(t -> assertThat(t.referencedToken()).isEqualTo(LOCK_TOKEN));
```
felixdesouza: I'd prefer if you invert it to make the assertion clear:

```java
assertThat(convertToSharedTokens(tokens))
  .extracting(SharedLockToken::referencedToken)
  .containsOnly(LOCK_TOKEN);
```

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    public void allSharedTokensShouldHaveDifferentIds() {
+        List<LockToken> tokens = SharedLockToken.share(LOCK_TOKEN, 3);
+        assertThat(tokens.stream()
```
felixdesouza: similarly:

```java
assertThat(tokens)
  .extracting(LockToken::getRequestId)
  .doesNotHaveDuplicates()
  .hasSize(3); // the above line implies this line so happy to leave out
```

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+        assertThat(firstToken.unlock()).isEmpty();
+        assertThat(firstToken.unlock()).isEmpty();
+    }
+
+    @Test
+    public void shouldWorkAsExpectedIfThereIsOnlyOneSharedToken() {
+        List<LockToken> tokens = SharedLockToken.share(LOCK_TOKEN, 1);
+
+        assertThat(tokens.size()).isEqualTo(1);
```
felixdesouza: nit: `assertThat(tokens).hasSize(1)`

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+                .hasMessage("Reference count should be more than zero");
+    }
+
+    private List<SharedLockToken> convertToSharedTokens(List<LockToken> lockTokens) {
+        return lockTokens.stream()
+                .map(this::convertToSharedToken)
+                .collect(Collectors.toList());
+    }
+
+    private SharedLockToken convertToSharedToken(LockToken lockToken) {
```
felixdesouza: nit: `static`

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+        assertThat(sharedLockToken.unlock()).contains(LOCK_TOKEN);
+    }
+
+    @Test
+    public void shouldThrowIfCreatedWithLessThanOneReference() {
+        assertThatThrownBy(() -> SharedLockToken.share(LOCK_TOKEN, 0))
+                .hasMessage("Reference count should be more than zero");
+    }
+
+    private List<SharedLockToken> convertToSharedTokens(List<LockToken> lockTokens) {
```
felixdesouza: nit: `static`

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
```diff
@@ -0,0 +1,95 @@
...
+        return lockTokens.stream()
+                .map(this::convertToSharedToken)
```
felixdesouza: fun fact, you could do: `.map(SharedLockToken.class::cast)`

lock-api/src/test/java/com/palantir/lock/client/SharedLockTokenTest.java
felixdesouza: I don't think you ever need this, wherever you use this, you always have the list, and then you extract them element by element, and then you convert. Just use `convertToSharedTokens` and then extract element by element.

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private volatile boolean unlocked;
+
+    private SharedLockToken(ReferenceCounter referenceCounter) {
+        this.referenceCounter = referenceCounter;
+        this.requestId = UUID.randomUUID();
+        this.unlocked = false;
+    }
+
+    public static List<LockToken> share(LockToken token, int referenceCount) {
```
felixdesouza: being a bit pedantic here, but what happens if you share a `SharedLockToken`?

i.e. if I have raw lock token, share it 5 times, and then for one of the shared tokens I share that 3 times, but then I go and unlock the first 5 shared lock tokens. What "guarantees" do I have about holding a `SharedLockToken`?

Aside from guarantees or whatever, it feels like it should special case `SharedLockToken` and pipe through the underlying lock token + somehow use the same reference counter?

If this won't come up in actual use case, then probably not worth it, but in terms of what guarantees you get holding a `SharedLockToken`, something feels amiss. Happy to be wrong though.

gokcanozakdag: we won't be sharing a shared lock token, but I think we can explicitly check that. That is, throw if someone tries to create shared lock tokens from another shared lock token.

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
```diff
@@ -0,0 +1,91 @@
...
+        return referenceCounter.lockToken;
+    }
+
+    @Override
+    public UUID getRequestId() {
+        return requestId;
+    }
+
+    private static class ReferenceCounter {
+        private LockToken lockToken;
```
jeremyk-91: Seems a bit strange to me to have the reference counter keep track of the 'real' lock token as well - was there a reason to put this here as opposed to in the `SharedLockToken`?

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.stream.IntStream;
+
+import com.google.common.base.Preconditions;
+import com.palantir.lock.v2.LockToken;
+
+class SharedLockToken implements LockToken {
+    private final UUID requestId;
+    private final ReferenceCounter referenceCounter;
+
+    private volatile boolean unlocked;
```
jeremyk-91: nit: since `unlocked` is only read/written during construction with assignment to a final field and then in synchronized methods, I don't think this needs to be volatile, unless I'm missing something

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        synchronized int unmark() {
+            Preconditions.checkState(referenceCount >= 0, "Reference count can not go below zero!");
+            return referenceCount--;
+        }
+
+        synchronized boolean dereferenced() {
+            return referenceCount == 0;
+        }
```
jeremyk-91: I'm not sure if these need to be synchronized, as accesses should already be synchronized on the `SharedLockToken`'s lock

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
```diff
@@ -0,0 +1,91 @@
...
+        if (!unlocked) {
+            unlocked = true;
+            referenceCounter.unmark();
+        }
+
+        return referenceCounter.dereferenced() ? Optional.of(referenceCounter.lockToken) : Optional.empty();
+    }
+
+    public LockToken referencedToken() {
+        return referenceCounter.lockToken;
```
jeremyk-91: This one isn't synchronized - this is fine as it looks like the lock token does have safe publication (it happens-before the SharedLockToken constructors are called, and these have a memory barrier at their end since they write to a final field), and is effectively final. Though I think we should just make the lock token final, as in other comments.

lock-api/src/main/java/com/palantir/lock/client/SharedLockToken.java
gokcanozakdag: I agree, this doesn't need to be volatile.

lock-api/src/test/java/com/palantir/lock/client/LockTokenShareTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    public void shouldReturnUnderlyingTokenAfterAllReferencesAreUnlocked() {
+        List<LockTokenShare> tokens = getSharedLockTokens(LOCK_TOKEN, 2);
+        LockTokenShare firstToken = tokens.get(0);
+        LockTokenShare secondToken = tokens.get(1);
+
+        assertThat(firstToken.unlock()).isEmpty();
+        assertThat(secondToken.unlock()).contains(LOCK_TOKEN);
```
felixdesouza: reading this test again, I think we also want to have a third assert: (also added `as` descriptions to describe each state/assert, since it's "stateful")

```java
assertThat(firstToken.unlock())
  .as("no token as second token still has reference")
  .isEmpty();
assertThat(secondToken.unlock())
  .as("get lock token since both references are unlocked")
  .contains(LOCK_TOKEN);
assertThat(firstToken.unlock())
  .as("first token now has lock token since both references unlocked")
  .contains(LOCK_TOKEN);
```

lock-api/src/test/java/com/palantir/lock/client/LockTokenShareTest.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Test
+    public void callingUnlockOnSameTokenMultipleTimesShouldNotUnlockUnderlyingToken() {
+        List<LockTokenShare> tokens = getSharedLockTokens(LOCK_TOKEN, 2);
+
+        LockTokenShare firstToken = tokens.get(0);
+
+        assertThat(firstToken.unlock()).isEmpty();
+        assertThat(firstToken.unlock()).isEmpty();
```
felixdesouza: similarly with the above, make an assert with `secondToken.unlock` to show that you will get there eventually, but until then, you won't

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private static final class ReferenceCounter {
+        private int referenceCount;
+
+        private ReferenceCounter(int referenceCount) {
+            this.referenceCount = referenceCount;
+        }
+
+        synchronized void unmark() {
+            Preconditions.checkState(referenceCount >= 0, "Reference count can not go below zero!");
```
j-baker: this doesn't work; you can get to -1

j-baker: it has to be specifically greater than zero here

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -0,0 +1,97 @@
...
+    public synchronized Optional<LockToken> unlock() {
+        if (!unlocked) {
+            unlocked = true;
+            referenceCounter.unmark();
+        }
+
+        return referenceCounter.dereferenced() ? Optional.of(sharedLockToken) : Optional.empty();
+    }
+
+    public LockToken sharedLockToken() {
```
j-baker: We should probably throw here if we've already unlocked the underlying lock token?

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -0,0 +1,97 @@
...
+        this.unlocked = false;
+    }
+
+    /**
+     * Returns a list of {@link LockTokenShare}, that are referencing to same LockToken.
+     *
+     * Share should be called only once on same {@link LockTokenShare} instance, guarantees do not hold between two
+     * different lists of {@link LockTokenShare} generated by successive share calls on same {@link LockToken}.
+     */
+    public static List<LockToken> share(LockToken token, int referenceCount) {
```
j-baker: all the 'public' declarations in this class are redundant because the class is package private

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -0,0 +1,97 @@
...
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Unlocks shared token on client side - does not guarantee underlying token to be unlocked on server side.
+     *
+     * @return referenced shared lock token if all lock token shares are unlocked; that is lock token on server side is
+     * good to be unlocked.
+     */
+    public synchronized Optional<LockToken> unlock() {
```
j-baker: There's some concurrency weirdness here. I don't think it's a bug, but I think it's probably worth paying attention to?

Basically, because the reference counter is locked differently to this class, two different classes can unmark concurrently and then both return the shared lock token.

This just means that sometimes, two people will try to unlock the same lock. This just leads to some weirdness wrt the lock contract; if I call unlock and it returns my lock token, it means that I had the lock for the whole period; here there's a way to end up with that incorrectly not being the case.

I'd have the unmark method return true if we actually unlocked the thing (and do that once)

```
if (!unlocked) {
    unlocked = true;
    return referenceCounter.unmark() ? Optional.of(sharedLockToken) : Optional.empty();
}
return Optional.empty();
```

gokcanozakdag: as we need to call refresh in any case to ensure that we were actually holding the locks before returning when unlock is called, we stick to contract in this implementation also I believe.
One problem with not returning the referenced token when it is already unlocked here is that, if you call unlock on lockTokenShare but fail to call timelock server to unlock, on the second call you may think the lock is actually unlocked, and just skip it.

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
gokcanozakdag: will fix that in a separate pr

PR #3784, opened 2019-02-27
[_transactions2] [WIP] Part 25: The CellLoader Reloaded

jeremyk-91:
**Goals (and why)**:
- Achieve faster read performance by batching reads, both in the form of `_transactions2` and in general. For instance if you read one row with 25 static columns... you're actually making 25 parallel RPCs to Cassandra while after this change you'll make just one.

- Actually use the new Cassandra endpoint we're planning on introducing in `CellLoader`.

- WIP, **TODO**

- The commands are going to become larger, possibly because we encode the slice predicate a bunch of times. In general I think this is fine; in static columns we only need to pass the short-name, and in dynamic columns you need to relay the information anyway. Nonetheless, this could mean the size of our RPCs increases.
  - There is a mitigating factor: when benchmarking Transactions1 with the new CellLoader, its perf was largely unaffected even though all the columns being asked for were the same.
- The batch size is `50,000` right now, which seems a bit arbitrary - is this a reasonable value?
- I opted to "compile" the `Map<KeyPredicate, List<ColumnOrSuperColumn>>` into a `Map<ByteBuffer, List<ColumnOrSuperColumn>>` because the visitor framework is still used for that type in the context of `multiget`s. Is this worth a change?

**Where should we start reviewing?**: CellLoader

**Priority (whenever / two weeks / yesterday)**: this week, though this is blocked on quite a number of other issues.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
@@ -61,6 +61,13 @@ public interface CassandraClient {
             SlicePredicate predicate, ConsistencyLevel consistency_level)
             throws InvalidRequestException, UnavailableException, TimedOutException, org.apache.thrift.TException;
 
+    Map<KeyPredicate, List<ColumnOrSuperColumn>> multiget_multislice(
+            String kvsMethodName,
+            TableReference tableRef,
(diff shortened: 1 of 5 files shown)


felixdesouza:
It would be nice/pretty cool for posterity sake if you'd put the benchmark results (going on the assumption you're making these changes as a result of benchmarks) in the PR description or as a separate comment.

jeremyk-91:
Will re-work this entirely, seeing as there were some changes in the algorithm we're using for this.

PR #3786, opened 2019-02-27
update min postgres version to 9.5.2

clockfort:
Shire devs desperately want to use upserts

I believe we were the only ones pulling your version requirements back to only being 9.2+ ; I don't anticipate issues with foundry-peeps from this, as sls-postgres is already 9.6

If you find this change particularly contentious, that's fine, and I can duplicate this version check back into our codebase with the change.


sandorw:
Release notes didn't conflict but are definitely in the wrong place.

PR #3790, opened 2019-02-28
Don't include timestamp in span operation

pkoenig10:
**Goals (and why)**:
Tracing key value service calls is great for debugging. Often when debugging you you want to aggregate all traces of a particular span to see things like p95, p99, etc. To do these aggregations, calls with similar expected performance should have the same span operation.

Currently the tracing key value services creates span names that use the timestamp. This means that reads in different transactions will have different span operations, making it impossible to do the above-mentioned aggregations.

Removes the timestamp component from the tracing key value store span operations.


j-baker:
Otoh, I also found the timestamp useful just this week for debugging purposes.

j-baker:
Like, our other option would be to put it in the 'params' part of the trace, but we blackhole those.

gmaretic:
It's unclear if we actually want this

PR #3791, opened 2019-03-01
Implement ForwardingKVS using AutoDelegate

gmaretic:
**Goals (and why)**:
Reduce unnecessary complexity

Originally wanted to completely remove the class, but we do actually have to override the lovely `getDelegates()` method.
Specifically decided not to have several other KVS implementations extending ForwardingObject implement this since it would be too easy to miss updating them correctly when adding methods to the KVS interface.

Existing tests

None

Small

Low, but this is small


felixdesouza:
wow the confidence
screen shot 2019-03-01 at 14 24 02

PR #3797, opened 2019-03-04
Batched start transaction endpoint

gokcanozakdag:
**Goals (and why)**:
Adding a batched-start-transaction endpoint to be used for coalesced start transaction requests (and sharing immutable ts locks)

This endpoint will be returning a single immutable ts (and lock) to be shared between multiple transactions. It will also return a range of timestamps and a timestamp partition. Timestamp range and partition can be used to create multiple start timestamps on client side.

`AsyncTimelockResource`


lock-api/src/main/java/com/palantir/lock/v2/BatchedStartTransactionResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+    static BatchedStartTransactionResponse of(
+            LockImmutableTimestampResponse immutableTimestamp,
+            TimestampRangeAndPartition startTimestamp,
+            Lease lease) {
+        return ImmutableBatchedStartTransactionResponse.builder()
+                .immutableTimestamp(immutableTimestamp)
+                .startTimestamps(startTimestamp)
+                .lease(lease)
+                .build();
```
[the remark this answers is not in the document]
gokcanozakdag: sometimes it is easier to read when builder is used, but I don't think it is the case with this one. I will change it to use `of()`

jeremyk-91:
Looks good, I'm aware we're waiting on some changes to the timestamp range and partition before merge

timestamp-impl/src/test/java/com/palantir/timestamp/TimestampRangesTest.java
```diff
@@ -19,95 +19,135 @@
...
+
+    private OptionalLong getTimestampMatchingModulus(
+            TimestampRange range, int residue, int modulus, SearchOption searchOption) {
+        switch (searchOption) {
+            case LOWEST:
+                return TimestampRanges.getLowestTimestampMatchingModulus(range, residue, modulus);
+            case HIGHEST:
+                return TimestampRanges.getHighestTimestampMatchingModulus(range, residue, modulus);
+        }
+        throw new RuntimeException("unknown type");
```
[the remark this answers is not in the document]
j-baker: you actually don't want this to be a default case, jeremy; error prone can enforce that you handled all cases right up until you add it as a default case

lock-api/src/test/java/com/palantir/lock/v2/TimestampRangeAndPartitionTest.java
```diff
@@ -0,0 +1,52 @@
...
+        TimestampRangeAndPartition timestampRange = TimestampRangeAndPartition.of(
+                TimestampRange.createInclusiveRange(timestamp, timestamp),
+                PARTITION
+        );
+
+        assertThat(timestampRange.getStartTimestamps()).containsOnly(timestamp);
+    }
+
+    @Test
+    public void getTimestampsShouldYieldCorrectResult_multipleTimestamps() {
```
felixdesouza: what is a "correct result" vs an "incorrect result"? Not entirely sure what you're testing here.

lock-api/src/main/java/com/palantir/lock/v2/BatchedStartTransactionResponse.java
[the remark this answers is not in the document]
felixdesouza: I've never really found `@Value.Parameter` useful tbh, because we never really use the `of` method of the `Immutable*` class, so we're forced to define the `of` class inside our class. If we then call `Immutable*.of` then it just becomes like calling a big constructor. 

🤷‍♂️

lock-api/src/main/java/com/palantir/lock/v2/BatchedStartTransactionResponse.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableBatchedStartTransactionResponse.class)
+@JsonDeserialize(as = ImmutableBatchedStartTransactionResponse.class)
+public interface BatchedStartTransactionResponse {
+    @Value.Parameter
+    LockImmutableTimestampResponse immutableTimestamp();
+
+    @Value.Parameter
+    TimestampRangeAndPartition timestampRange();
```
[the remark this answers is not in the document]
gokcanozakdag: it works both with tx2 and tx1, tx2 requires partitioned timestamps, tx1 doesn't care if it is partitioned or not.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/BatchingTimestampService.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.transaction.timestamp;
+
+import java.util.UUID;
+
+import com.palantir.lock.v2.TimestampRangeAndPartition;
+
+public interface BatchingTimestampService extends ClientAwareManagedTimestampService {
```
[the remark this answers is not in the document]
gokcanozakdag: problem is `AsyncTimelockService` implements `ClientAwareManagedTimestampService` and I don't want to add getFreshTimestampsForClient to `AsyncTimelockService` interface. I don't like this solution also, but don't have much alternatives :/

lock-api/src/main/java/com/palantir/lock/client/SharedConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+public final class SharedConstants {
+    private SharedConstants() {/* */}
+
+    // DO NOT change without a transactions table migration!
+    public static final int TRANSACTION_NUM_PARTITIONS = 16;
```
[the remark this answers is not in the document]
gokcanozakdag: that would be less error prone actually, I agree; @jeremyk-91 what do you think?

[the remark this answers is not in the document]
jeremyk-91: No strong opinions here, though either way I think the `V2` is important

lock-api/src/main/java/com/palantir/lock/v2/BatchedStartTransactionResponse.java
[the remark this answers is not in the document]
jeremyk-91: Not sure if it's really that we don't use `Immutable*.of` actually. Agree that calling ImmutableX.of() when X.of() exists is strange, though we aren't always consistent with declaring that

```
$ find . | grep "\.java" | grep -v "generated" | xargs cat | grep "Immutable.*\.of(" | grep -vE "Immutable(Sorted)?(Set|Map|List|Multimap)" | grep -o "Immutable[a-zA-Z0-9]*" | sort | uniq -c
  16 ImmutableCallbackInvocation
   3 ImmutableCheckAndSetResult
   1 ImmutableDedicatedRows
   2 ImmutableFastFailoverState
   3 ImmutableIdentifiedLockRequest
   1 ImmutableIdentifiedSystemTime
   4 ImmutableIdentifiedTimeLockRequest
   1 ImmutableLeaderTime
   1 ImmutableLeadershipId
   2 ImmutableLeadershipState
   1 ImmutableLease
   3 ImmutableLockImmutableTimestampResponse
   3 ImmutableLockRequest
   3 ImmutableLockRequestSpecification
   3 ImmutableLockResponse
   2 ImmutableLockToken
   1 ImmutableLockWithClient
   8 ImmutableMaskedValue
   1 ImmutableOrderedLocks
   2 ImmutableOverflowValue
   2 ImmutablePaxosLong
   1 ImmutablePaxosValueResponse
   2 ImmutablePersistentLockId
   1 ImmutableRangeAndValue
   6 ImmutableRangeMap
   2 ImmutableReferenceCountedResidue
   2 ImmutableRefreshLockResponseV2
   2 ImmutableRequestInfo
  11 ImmutableSequenceAndBound
   1 ImmutableSimpleLockRequest
   1 ImmutableSimpleTokenInfo
   2 ImmutableStartAtlasDbTransactionResponse
   1 ImmutableStartAtlasDbTransactionResponseV3
   2 ImmutableStartIdentifiedAtlasDbTransactionRequest
   3 ImmutableStartIdentifiedAtlasDbTransactionResponse
   1 ImmutableStoredWriteReference
   2 ImmutableSweepTableIdentifier
   1 ImmutableTimeLimit
   3 ImmutableTimeLimiterConfiguration
   1 ImmutableTimestamp
   2 ImmutableTimestampAndPartition
   1 ImmutableTimestampClientConfig
   2 ImmutableTimestampPair
   2 ImmutableTimestampPartitioningMap
  11 ImmutableTimestampResponse
   9 ImmutableTodo
   1 ImmutableTsLock
   3 ImmutableUnsuccessful
   2 ImmutableValueAndBound
   3 ImmutableWaitForLocksRequest
   3 ImmutableWaitForLocksResponse
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/BatchingTimestampService.java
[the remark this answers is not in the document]
felixdesouza: I'm not sure I follow this one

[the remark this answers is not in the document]
felixdesouza: Basically what I'm proposing is to merge `BatchingTimestampService` into `ClientAwareManagedTimestampService`, you're already providing the implementation. 

Structure surrounding the indirection with `AsyncTimelockService`, `AsyncTimelockServiceImpl`, `AsyncTimelockResource` (which I think is what is biting you here) are somewhat unrelated and should be resolved at a later date imo.

That aside, it breaks the symmetry with `TimestampService`.

lock-api/src/main/java/com/palantir/lock/client/SharedConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+public final class SharedConstants {
```
j-baker: what does sharedconstants mean here? When should I be using it?

lock-api/src/main/java/com/palantir/lock/v2/BatchedStartTransactionRequest.java
```diff
@@ -0,0 +1,46 @@
...
+@JsonDeserialize(as = ImmutableBatchedStartTransactionRequest.class)
+public interface BatchedStartTransactionRequest {
+    @Value.Parameter
+    UUID requestId();
+
+    @Value.Parameter
+    UUID requestorId();
+
+    @Value.Parameter
+    int batchSize();
```
j-baker: maybe numTransactions?

j-baker: doesn't really describe what the batch is of otherwise

lock-api/src/main/java/com/palantir/lock/v2/TimestampRangeAndPartition.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.lock.client.SharedConstants;
+import com.palantir.timestamp.TimestampRange;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableTimestampRangeAndPartition.class)
+@JsonDeserialize(as = ImmutableTimestampRangeAndPartition.class)
+public interface TimestampRangeAndPartition {
+    @Value.Parameter
+    TimestampRange range();
```
j-baker: this feels like a weird shape of response, right? It's a timestamp range where you've handed back an actual range, but you're only supposed to use a certain subset of them.

lock-api/src/main/java/com/palantir/lock/v2/TimestampRangeAndPartition.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.palantir.lock.v2;
+
+import org.immutables.value.Value;
+
+import com.palantir.timestamp.TimestampRange;
+
+@Value.Immutable
+public interface TimestampRangeAndPartition {
```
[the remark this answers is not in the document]
j-baker: I'd like to follow up on this; I think that this class will only really make sense if it's documented correctly both in explicit comments and actually the code.

So, instead of returning a range and a modulus indicating which in the range are valid mod 16, maybe we should instead be returning:

1. Start
2. Interval
3. Count

indicating the beginning timestamp, the difference between each timestamp, and the number of timestamps returned?

In Jeremy's case, you'd then be returning(0, 16, 2) which is to me much more fully specified?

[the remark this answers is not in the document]
j-baker: it also means the client doesn't have to know about how many partitions we have, which is also much nicer; it means we can change it down the line and don't need SharedConstants.

lock-api/src/main/java/com/palantir/lock/v2/TimestampRangeAndPartition.java
```diff
@@ -0,0 +1,49 @@
...
+@JsonDeserialize(as = ImmutableTimestampRangeAndPartition.class)
+public interface TimestampRangeAndPartition {
+    @Value.Parameter
+    TimestampRange range();
+
+    @Value.Parameter
+    int partition();
+
+    @JsonIgnore
+    default long[] getStartTimestamps() {
```
j-baker: If I asked for 10k transactions with a partition of 0, it seems weird that i'd be iterating through 160k things, rather than just 10k. For your current code, you could e.g. have a range between 0 and (upper - lower)/num partitions and then add lower and multiply by num partitions for the result...

but I'd probably just return a LongStream and compute it by being like

```
return LongStream.range(0, count).map(item -> base + item * interval);
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampService.java
```diff
@@ -58,15 +60,29 @@ public static ClientAwareManagedTimestampService createDefault(ManagedTimestampS
...
+        return TimestampAndPartition.of(
+                timestampRangeAndPartition.range().getLowerBound(),
+                timestampRangeAndPartition.partition());
+    }
+
+    @Override
+    public TimestampRangeAndPartition getFreshTimestampsForClient(UUID clientIdentifier, int numTransactionsRequested) {
         while (true) {
-            TimestampRange timestampRange = delegate.getFreshTimestamps(NUM_PARTITIONS);
+            TimestampRange timestampRange = delegate.getFreshTimestamps(NUM_PARTITIONS * numTransactionsRequested);
```
j-baker: does this correctly handle the case where you asked for more than 10k timestamps and we didn't hand you 10k timestamps?

j-baker: I guess yes.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -342,6 +347,43 @@ public void distinctClientsStillShareTheSameSequenceOfTimestamps() {
         assertThat(temporalSequence).isSorted();
     }
 
+    @Test
+    public void temporalOrderingIsPreservedForBatchedStartTransactionRequests() {
```
j-baker: so here, we're asserting that we got the stuff back in sorted order. But, that's not just what we do here; the contract also says that there must be an immutable lock and the results must be separated by the modulus. Fancy testing that?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampService.java
jeremyk-91: I think so (and also the cases where you ask for `N < 10000` and get `M < N`, which IIRC can happen as leaders just return what they have rather than do more Paxos if they get to the end of their range) - the general idea is if you have something, _anything_ usable then return that, otherwise try again.

lock-api/src/main/java/com/palantir/lock/v2/TimestampRangeAndPartition.java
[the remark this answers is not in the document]
jeremyk-91: Yep, I think I prefer this specific triple format over trying to reuse `TimestampRange`

lock-api/src/main/java/com/palantir/lock/v2/TimestampRangeAndPartition.java
jeremyk-91: doh, missed this - yeah probably best to directly calculate the usable timestamps

timestamp-impl/src/test/java/com/palantir/timestamp/TimestampRangesTest.java
```diff
@@ -19,95 +19,135 @@
...
+    public void canGetTimestampFromRangeIfItIsTheLowerBound(SearchOption searchOption) {
+        assertThat(searchOption.timestampMatchingModulus(SEVENTY_THREE_TO_EIGHTY_TWO, 3, 10))
                 .isPresent()
                 .hasValue(SEVENTY_THREE);
     }
 
-    @Test
-    public void canGetTimestampFromRangeIfItIsTheUpperBound() {
-        assertThat(TimestampRanges.getTimestampMatchingModulus(SEVENTY_THREE_TO_EIGHTY_TWO, 2, 10))
+    @Theory
```
j-baker: fwiw, I've found that theories make it very hard to actually debug test failures. If you fail on only one case, it's tricky to figure out the exact failure mode.

timestamp-impl/src/main/java/com/palantir/timestamp/TimestampRanges.java
```diff
@@ -61,4 +53,37 @@ public static OptionalLong getTimestampMatchingModulus(TimestampRange range, int
                 ? OptionalLong.of(candidate)
                 : OptionalLong.empty();
     }
+
+    /**
+     * Returns the highest timestamp in the provided timestamp range that has the provided residue class modulo the
+     * provided modulus.
+     *
+     * If the timestamp range does not contain a timestamp matching the criteria, returns empty.
```
j-baker: i think this feels weird from a contract perspective. We never call this method in a way where it could return empty... this validation is all essentially redundant.

Maybe just have this method actually return the response object? like something that returns the combination of (base, count, multiple)?

timestamp-impl/src/test/java/com/palantir/timestamp/TimestampRangesTest.java
gokcanozakdag: removing theories, as we don't need to use two separate search options anymore

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/ClientAwareManagedTimestampService.java
```diff
@@ -38,4 +39,6 @@
      * @return a suitable timestamp
      */
     TimestampAndPartition getFreshTimestampForClient(UUID clientIdentifier);
```
j-baker: As noted in the leased locks thing; we should be killing this method at this time (since it can be written in terms of the new one :) )

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceImpl.java
```diff
@@ -124,21 +119,42 @@ public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLoc
     @Override
     public StartAtlasDbTransactionResponseV3 startIdentifiedAtlasDbTransaction(
             StartIdentifiedAtlasDbTransactionRequest request) {
+        Leased<LockImmutableTimestampResponse> leasedLockImmutableTimestampResponse =
```
j-baker: this method should be written _in terms_ of the other one, as in the other PRs

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceImpl.java
```diff
@@ -124,21 +119,42 @@ public StartAtlasDbTransactionResponse startAtlasDbTransaction(IdentifiedTimeLoc
...
+                timestampService.getFreshTimestampForClient(request.requestorId());
+
+        return StartAtlasDbTransactionResponseV3.of(
+                leasedLockImmutableTimestampResponse.value(),
+                timestampAndPartition,
+                leasedLockImmutableTimestampResponse.lease());
+    }
+
+    @Override
+    public BatchedStartTransactionResponse batchedStartTransaction(BatchedStartTransactionRequest request) {
```
j-baker: in line with comments on the old PR - think this isn't a new 'batched' version of starting transaction, I think it's actually startAtlasDbTransactionV4

j-baker: (which might well lead to some class renames here)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampService.java
j-baker: I don't think your IIRC is quite right; looking at the Paxos bound it will increase the bound before handing you anything back.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampService.java
```diff
@@ -58,15 +58,24 @@ public static ClientAwareManagedTimestampService createDefault(ManagedTimestampS
 
     @Override
     public TimestampAndPartition getFreshTimestampForClient(UUID clientIdentifier) {
+        long partitionedTimestamp = getFreshTimestampsForClient(clientIdentifier, 1).start();
```
j-baker: yeah see this is what we want :)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampService.java
```diff
@@ -84,4 +93,8 @@ public TimestampAndPartition getFreshTimestampForClient(UUID clientIdentifier) {
     public ManagedTimestampService delegate() {
         return delegate;
     }
+
+    private static int calculatePartition(long timestamp) {
+        return (int) timestamp % NUM_PARTITIONS;
```
j-baker: I don't think this is what you meant to write. This is equivalent to

```java
((int) timestamp) % NUM_PARTITIONS
```

which is equivalent to

```java
(timestamp % 2^32) % NUM_PARTITIONS
```

and if NUM_PARTITIONS is a power of 2 then we're fine here, but if e.g. `NUM_PARTITIONS` is 3 then `calculatePartition(2^32) == 0` whereas `2^32 mod 3 = 1`.

In this case, because it's a power of 2 it's fine, but it's not documented it _has_ to be a power of 2. I would write this code as

```
return Ints.checkedCast(timestamp % NUM_PARTITIONS);
```

because I think it follows the order of precedence that you want?

j-baker:
```
|  Welcome to JShell -- Version 11.0.1
|  For an introduction type: /help intro

jshell> var x = (1L << 32)
x ==> 4294967296

jshell> var mod = 3
mod ==> 3

jshell> (int) x % mod
$3 ==> 0

jshell> (int) (x % mod)
$4 ==> 1

jshell>
```

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampServiceTest.java
```diff
@@ -114,6 +114,37 @@ public void makesRequestsAgainIfTimestampRangeDoesNotIncludeCorrectModuli() {
...
         verify(timestamps, times(3)).getFreshTimestamps(anyInt());
+    }
+
+    @Test
+    public void batchedTimestampCallShouldReturnAtLeastOneUsableTimestamp() {
+        when(allocator.getRelevantModuli(UUID_ONE)).thenReturn(RESIDUE_ONE);
+        when(timestamps.getFreshTimestamps(anyInt()))
+                .thenReturn(TIMESTAMP_RANGE);
+
+        assertThat(service.getFreshTimestampsForClient(UUID_ONE, 2).stream())
```
j-baker: is there a containsExactly method here? `assertThat(service.getFreshTimestampsForClient(UUID_ONE, 2).stream()).containsExactly(RESIDUE_ONE_TIMESTAMP_IN_RANGE.timstamp());`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/DelegatingClientAwareManagedTimestampServiceTest.java
```diff
@@ -114,6 +114,37 @@ public void makesRequestsAgainIfTimestampRangeDoesNotIncludeCorrectModuli() {
 
         verify(allocator, times(3)).getRelevantModuli(UUID_TWO);
         verify(timestamps, times(3)).getFreshTimestamps(anyInt());
+    }
```
j-baker: can probably kill these above tests if the old methods vanish as well :)

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -342,6 +348,56 @@ public void distinctClientsStillShareTheSameSequenceOfTimestamps() {
...
+    }
+
+    private List<Long> getSortedBatchedStartTimestamps(UUID requestorUuid, int numRequestedTimestamps) {
+        BatchedStartTransactionRequest request = BatchedStartTransactionRequest.createForRequestor(
+                requestorUuid,
+                numRequestedTimestamps);
+        BatchedStartTransactionResponse response = CLUSTER.timelockRpcClient().batchedStartTransaction(request);
+        return response.timestamps().stream()
+                .boxed()
+                .sorted() // we do not guarantee sorted order for same batch
```
j-baker: we can probably just kill these two lines; the wire format means we effectively guarantee sort order...

lock-api/src/main/java/com/palantir/lock/v2/TimelockRpcClient.java
```diff
@@ -57,27 +55,19 @@
...
     @POST
-    @Path("start-atlasdb-transaction")
-    StartAtlasDbTransactionResponse deprecatedStartAtlasDbTransaction(IdentifiedTimeLockRequest request);
-
-    @POST
-    @Path("start-identified-atlasdb-transaction")
-    StartIdentifiedAtlasDbTransactionResponse deprecatedStartIdentifiedAtlasDbTransaction(
+    @Path("start-atlasdb-transaction-v3")
+    StartAtlasDbTransactionResponseV3 deprecatedStartTransaction(
             StartIdentifiedAtlasDbTransactionRequest request);
```
gokcanozakdag: removing deprecated methods from rpcClient while keeping endpoints on the server (planing to remove those also with 1.x release)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java
```diff
@@ -79,22 +81,36 @@ public LockImmutableTimestampResponse lockImmutableTimestamp(IdentifiedTimeLockR
 
     @POST
     @Path("start-atlasdb-transaction")
-    public StartAtlasDbTransactionResponse deprecatedStartAtlasDbTransaction(IdentifiedTimeLockRequest request) {
-        return timelock.startAtlasDbTransaction(request);
+    public StartAtlasDbTransactionResponse deprecatedStartTransactionV1(IdentifiedTimeLockRequest request) {
+        return timelock.deprecatedStartTransaction(request);
     }
```
gokcanozakdag: wanted to rename these also as it was getting hard to follow

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceImpl.java
```diff
@@ -115,30 +110,53 @@ public RefreshLockResponseV2 refreshLockLeases(Set<LockToken> tokens) {
...
+    public StartAtlasDbTransactionResponseV3 startTransaction(
             StartIdentifiedAtlasDbTransactionRequest request) {
+        StartTransactionResponseV4 startTransactionResponseV4 =
+                startTransactions(StartTransactionRequestV4.createForRequestor(request.requestorId(), 1));
+
+        return StartAtlasDbTransactionResponseV3.of(
+                startTransactionResponseV4.immutableTimestamp(),
+                getTimestampAndPartition(startTransactionResponseV4.timestamps()),
+                startTransactionResponseV4.lease()
+        );
```
j-baker: super nit: probably drop this on the same line for consistency with the rest of the codebase?

PR #3799, opened 2019-03-04
Add backoff when leadership check hits NO_QUORUM to develop

gmaretic:
**Goals (and why)**:
Port #3796 to develop
 
Does this still make sense? Did I mess something up in fixing conflicts?

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
@@ -74,7 +74,7 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
     static final int DEFAULT_NO_QUORUM_MAX_DELAY_MS = 2000;
 
     private final RateLimiter logRateLimiter = RateLimiter.create(LOG_RATE);
-    private final ReentrantLock lock;
+    private final ReentrantLock lock = new ReentrantLock();
     private final CoalescingPaxosLatestRoundVerifier latestRoundVerifier;
 
     final PaxosProposer proposer;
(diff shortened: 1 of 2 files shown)


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -106,8 +109,33 @@ public PaxosLeaderElectionService(PaxosProposer proposer,
...
+                executorServiceFactory,
+                updatePollingWaitInMs,
+                randomWaitBeforeProposingLeadership,
+                DEFAULT_NO_QUORUM_MAX_DELAY_MS,
+                leaderPingResponseWaitMs,
+                PaxosLeaderElectionEventRecorder.NO_OP,
+                onlyLogOnQuorumFailure);
+    }
+
+    PaxosLeaderElectionService(PaxosProposer proposer,
```
felixdesouza: why do we have this *second* constructor if we have a builder? :S

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -175,6 +204,10 @@ public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
                     return currentState.confirmedToken().get();
                 case NO_QUORUM:
                     // If we don't have quorum we should just retry our calls.
+                    long backoffTime = ThreadLocalRandom.current().nextLong(0, noQuorumMaxDelay + 1);
```
jeremyk-91: nit: This is just `nextLong(noQuorumMaxDelay + 1);`. Saves a bunch of tests, though to be fair this might get optimised out.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
jeremyk-91: Since we do need to pass in the value of `leaderPingResponseWaitMs`, it's needed if you don't want to break the API. An alternative (which I'd prefer) is that we _can_ remove the deprecated constructor and leave just this one; a quick check on large internal product did not reveal any uses.

felixdesouza: yeah, understandable, just feel like we should be a bit more aggressive with breaks to make our codebase a bit cleaner one constructor at a time aha.

gmaretic:
ping

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -84,6 +86,7 @@
 
     final long updatePollingRateInMs;
     final long randomWaitBeforeProposingLeadership;
+    final long noQuorumMaxDelay;
```
felixdesouza: what is the unit? 🤔

whilst you're here, `Duration` all the way!

gmaretic:
Not worth it to unravel the merge conflicts

Issue #3800, opened 2019-03-04
Table Remapping KVS can throw exceptions if doing getAllTableNames() and dropTables() in parallel


jeremyk-91:
getFullTableName#78 can fail, if you drop stuff and call getAllTableNames() in parallel.

```
  @Test
  public void reproduction() {
    ThreadPoolExecutor executor = injector.getInstance(SomeMagicServiceFactory.class)
        .newCachedThreadPool(0, 10000);
    for (int i = 0; i < 10000; i++) {
      String firstWorld = generateRandomTablesPrefixed();
      String secondWorld = generateRandomTablesPrefixed();
      final Set<Future<Void>> futures = new HashSet<>();
      futures.add(executor.submit(() -> {
        String worldPrefix = NAMESPACE.getName() + "." + firstWorld + "__";
        Set<TableReference> tablesToDrop = Sets.newHashSet();
        for (TableReference tableRef : kvs.getAllTableNames()) {
          if (tableRef.getQualifiedName().startsWith(worldPrefix)) {
            tablesToDrop.add(tableRef);
          }
        }
        kvs.dropTables(tablesToDrop);
        return null;
      }));
      futures.add(executor.submit(() -> {
        String worldPrefix = NAMESPACE.getName() + "." + secondWorld + "__";
        Set<TableReference> tablesToDrop = Sets.newHashSet();
        for (TableReference tableRef : kvs.getAllTableNames()) {
          if (tableRef.getQualifiedName().startsWith(worldPrefix)) {
            tablesToDrop.add(tableRef);
          }
        }
        kvs.dropTables(tablesToDrop);
        return null;
      }));
      Futures.waitForAllFuturesUnchecked(futures);
    }
  }
  private String generateRandomTablesPrefixed() {
    String worldName = UUID.randomUUID().toString().replace("-", "_");
    String tableName = UUID.randomUUID().toString().replace("-", "_");
    kvs.createTable(TableReference.create(NAMESPACE, worldName + "__" + tableName), new TableDefinition() {{
      javaTableName(tableName);
      rowName();
      rowComponent("row", ValueType.VAR_STRING);
      columns();
      column("col", "c", ValueType.FIXED_LONG);
      conflictHandler(ConflictHandler.RETRY_ON_WRITE_WRITE);
      rangeScanAllowed();
      negativeLookups();
      ignoreHotspottingChecks();
      explicitCompressionBlockSizeKB(64);
    }}.toTableMetadata().persistToBytes());
[...]

PR #3803, opened 2019-03-05
Fix Some TableMapper Race Conditions

jeremyk-91:
**Goals (and why)**:
- Address the most egregious instances of #3800 - such as a DDL operation on unrelated tables causing reads to fail out.

- Never flush the map entirely, only remove entries when things are dropped.
- Standardise throwing the `TableMappingNotFoundException` rather than sometimes using Validate and sometimes throwing that.

No previous tests. I added the test featured in #3800 plus one more.

- There is probably a whole litany of badness in these classes, and I think this improves the state of things. However, is there any other low-hanging fruit I've missed here?

**Where should we start reviewing?**: TableRemappingKeyValueServiceTest to get an understanding of what behaviour this tries to fix, then probably the TableMappingService


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private final KeyValueService kvs = TableRemappingKeyValueService.create(rawKvs, tableMapper);
+
+    @Test
+    public void canConcurrentlyDropAndCreateDisjointSetsOfTables() {
+        ExecutorService executor = Executors.newFixedThreadPool(NUM_THREADS);
+        for (int i = 0; i < NUM_ITERATIONS; i++) {
+            String firstWorld = generateRandomTablesPrefixed();
+            String secondWorld = generateRandomTablesPrefixed();
+            final Set<Future<Void>> futures = new HashSet<>();
+            futures.add(executor.submit(() -> dropTablesWithPrefix(firstWorld)));
```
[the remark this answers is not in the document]
felixdesouza: are we able to do some stuff with `CountdownLatches`? I've read the test a few times and not entirely sure I know what it's doing, plus there's no guarantee of a test run actually catching this right?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -110,14 +115,31 @@ private TableReference getAlreadyExistingMappedTableName(TableReference tableRef
...
+                            unused -> AtlasDbConstants.TRANSACTION_TS,
+                            MultimapBuilder.hashKeys().arrayListValues()::build));
+            kv.delete(AtlasDbConstants.NAMESPACE_TABLE, deletionMap);
+        }
+
+        // Need to invalidate the table refs in case we end up re-creating the same table again.
+        while (true) {
+            BiMap<TableReference, TableReference> existingMap = tableMap.get();
+            BiMap<TableReference, TableReference> newMap = HashBiMap.create(existingMap);
+            tableRefs.forEach(newMap::remove);
```
felixdesouza: This will drop updates from other threads no? We'd eventually get there I *think*.

`MapDifference` usually helps, but not sure how to deal with table removals without some sort of sentinel, and also don't want to keep them in the map long 🤷‍♂️

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void canConcurrentlyReadFromStableTablesWhileCreatingAndDroppingOtherTables() {
+        ExecutorService executor = Executors.newFixedThreadPool(NUM_THREADS);
+        kvs.createTable(DATA_TABLE_REF, AtlasDbConstants.GENERIC_TABLE_METADATA);
+        Cell testCell = Cell.create(PtBytes.toBytes("row"), PtBytes.toBytes("col"));
+        byte[] value = PtBytes.toBytes("value");
+        long testTimestamp = 1L;
+        kvs.put(DATA_TABLE_REF, ImmutableMap.of(testCell, value), testTimestamp);
+        for (int i = 0; i < NUM_ITERATIONS; i++) {
```
felixdesouza: I have no idea what is going on here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -110,14 +115,31 @@ private TableReference getAlreadyExistingMappedTableName(TableReference tableRef
...
+
+    @Override
+    public void removeTables(Set<TableReference> tableRefs) {
         if (kv.getAllTableNames().contains(AtlasDbConstants.NAMESPACE_TABLE)) {
-            kv.delete(AtlasDbConstants.NAMESPACE_TABLE, ImmutableMultimap.of(key, 0L));
+            Multimap<Cell, Long> deletionMap = tableRefs.stream()
+                    .map(tableRef -> Cell.create(getBytesForTableRef(tableRef),
+                            AtlasDbConstants.NAMESPACE_SHORT_COLUMN_BYTES))
+                    .collect(Multimaps.toMultimap(
+                            key -> key,
```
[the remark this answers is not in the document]
jeremyk-91: Yep, that's a fair observation

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
jeremyk-91: It will drop updates from other threads, and if I understand things correctly, we may even fail to eventually get the right answer. 

Basically with this PR I gave up any guarantees of consistency if you concurrently drop and create the same table, the main thing I wanted to do was to ensure that concurrent updates to table B don't affect table A.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractTableMappingService.java
```diff
@@ -70,14 +70,17 @@ protected void validateShortName(TableReference tableRef, TableReference shortNa
                 "Table mapper has an invalid table name for table reference %s: %s", tableRef, shortName);
     }
 
-    private TableReference getFullTableName(TableReference shortTableName) {
-        if (tableMap.get().containsValue(shortTableName)) {
-            return tableMap.get().inverse().get(shortTableName);
+    private TableReference getFullTableName(TableReference shortTableName) throws TableMappingNotFoundException {
+        BiMap<TableReference, TableReference> mapToConsider = tableMap.get();
```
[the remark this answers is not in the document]
jeremyk-91: You're right, if we wanted a consistent view. Probably easier to discuss offline tomorrow - this PR does not have as a goal making the concurrency correct in general (I mainly wanted to make it stable for tables where no DDL operations are ongoing), so many things might look broken.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractTableMappingService.java
```diff
@@ -102,7 +105,11 @@ private TableReference getFullTableName(TableReference shortTableName) {
             if (inputName.isFullyQualifiedName()) {
                 shortNameToFullTableName.put(inputName, inputName);
             } else if (tableMap.get().containsValue(inputName)) {
```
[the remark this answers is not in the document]
jeremyk-91: You're right, you can (but again, see point above on PR not aiming to achieve consistency)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueService.java
```diff
@@ -136,13 +137,9 @@ public void dropTables(Set<TableReference> tableRefs) {
...
 
         // We're purposely updating the table mappings after all drops are complete
-        for (TableReference tableRef : tableRefs) {
-            // Handles the edge case of deleting _namespace when clearing the kvs
-            if (tableRef.equals(AtlasDbConstants.NAMESPACE_TABLE)) {
-                break;
-            }
-            tableMapper.removeTable(tableRef);
-        }
+        tableMapper.removeTables(tableRefs.stream()
```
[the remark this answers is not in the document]
jeremyk-91: Alright, can do

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private final AtomicLong timestamp = new AtomicLong();
+    private final KeyValueService rawKvs = new InMemoryKeyValueService(true);
+    private final KvTableMappingService tableMapper = KvTableMappingService.create(rawKvs, timestamp::incrementAndGet);
+    private final KeyValueService kvs = TableRemappingKeyValueService.create(rawKvs, tableMapper);
+
+    @Test
+    public void canConcurrentlyDropAndCreateDisjointSetsOfTables() {
+        ExecutorService executor = Executors.newFixedThreadPool(NUM_THREADS);
+        for (int i = 0; i < NUM_ITERATIONS; i++) {
+            String firstWorld = generateRandomTablesPrefixed();
```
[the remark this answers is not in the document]
jeremyk-91: Yep, the creation isn't concurrent, only the drops are

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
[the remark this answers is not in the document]
jeremyk-91: 1. Yeah, that's right - will rename the test.
2. Possibly (I'm thinking you want a `CyclicBarrier` actually), if you wanted to enforce concurrency. There's no guarantee, but in general with 1000 iterations without the fix you almost always get failures at iteration 10 or earlier

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
jeremyk-91: Similar to the first test, except that instead of dropping table A and dropping table B concurrently, this drops table A and _reads_ from table B concurrently. They go through different code-paths IIRC (and both fail without this change)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/AbstractTableMappingService.java
```diff
@@ -102,7 +105,11 @@ private TableReference getFullTableName(TableReference shortTableName) {
...
-                shortNameToFullTableName.put(inputName, getFullTableName(inputName));
+                try {
+                    shortNameToFullTableName.put(inputName, getFullTableName(inputName));
+                } catch (TableMappingNotFoundException e) {
+                    // Possible if the table was concurrently dropped. Continue.
```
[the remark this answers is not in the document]
jeremyk-91: Mainly for consistency reasons. This is used in `getAllTableNames()`, which does not guarantee atomicity (e.g. see `TableSplittingKeyValueService` implementation) and in `getMetadataForTables()` which is also non-atomic (e.g. Cassandra implementation). In particular if a table is dropped during these operations, it's usually allowed to either return or not return it here, which is what swallowing the exception will do.

jeremyk-91:
Internal product: I think not (though we should check), because this feature is used at Oracle deployments, and that does happen for products in the large internal product's ecosystem (that are not the large internal product itself).

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+                        Value.create(value, testTimestamp)));
+                return null;
+            }));
+            futures.forEach(Futures::getUnchecked);
+        }
+    }
+
+    private Void dropTablesWithPrefix(String prefix) {
+        String namespacedPrefix = NAMESPACE.getName() + "." + prefix + "__";
+        Set<TableReference> tablesToDrop = Sets.newHashSet();
```
gmaretic:
```
= kvs.getAllTableNames().stream()
                .filter(tableRef -> tableRef.getQualifiedName().startsWith(namespacedPrefix))
                .collect(Collectors.toSet());
```

for style points?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/TableRemappingKeyValueServiceTest.java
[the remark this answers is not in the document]
gmaretic: Still need to rename the test, as the name confused me as well, otherwise it's fine

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -36,70 +50,76 @@
...
         try {
-            kv.putUnlessExists(AtlasDbConstants.NAMESPACE_TABLE, ImmutableMap.of(key, value));
+            kvs.putUnlessExists(AtlasDbConstants.NAMESPACE_TABLE, ImmutableMap.of(key, value));
         } catch (KeyAlreadyExistsException e) {
             return getAlreadyExistingMappedTableName(tableRef);
         }
         return TableReference.createWithEmptyNamespace(shortName);
     }
 
+    private Cell getKeyCellForTable(TableReference tableRef) {
```
felixdesouza: static

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -110,35 +130,136 @@ private TableReference getAlreadyExistingMappedTableName(TableReference tableRef
...
+                "Table mapper has an invalid table name for table reference %s: %s", tableRef, shortName);
+    }
+
+
+    @Override
+    public <T> Map<TableReference, T> mapToShortTableNames(Map<TableReference, T> toMap)
+            throws TableMappingNotFoundException {
+        Map<TableReference, T> newMap = Maps.newHashMap();
+        for (Map.Entry<TableReference, T> e : toMap.entrySet()) {
+            newMap.put(getMappedTableName(e.getKey()), e.getValue());
```
felixdesouza: this is sketch, can we get the map, then get the things from it?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
```diff
@@ -110,35 +130,136 @@ private TableReference getAlreadyExistingMappedTableName(TableReference tableRef
...
     @Override
+    public Map<TableReference, TableReference> generateMapToFullTableNames(Set<TableReference> tableRefs) {
+        Map<TableReference, TableReference> shortNameToFullTableName = Maps.newHashMapWithExpectedSize(
+                tableRefs.size());
+        Set<TableReference> tablesToReload = Sets.newHashSet();
+        for (TableReference inputName : tableRefs) {
+            if (inputName.isFullyQualifiedName()) {
+                shortNameToFullTableName.put(inputName, inputName);
+            } else {
+                TableReference candidate = tableMap.get().inverse().get(inputName);
```
felixdesouza: similarly here

gmaretic: The fix here is simple, so although it adds some overhead, makes sense

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
gmaretic: The fix here would result in more unreadable code, so given that there is no contract about what should happen in presence of race conditions, I'm leaning to leaving this as is

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/KvTableMappingService.java
gmaretic: 👍

gmaretic:
Added unit tests. These also pass on develop, except for the two verifying the short table name validation. I believe this was a bug on develop since the validation is not done when we have to reload the table mapping from KVS, but it is done when it is not reloaded -- now we do it either way

PR #3807, opened 2019-03-07
Fix targeted sweep correctness bugs

j-baker:
Here, I fix two correctness bugs.

1. The abstract method is bogus. It doesn't guarantee that stuff is
deleted in the right order, and so isn't safe; when deleting cells you
HAVE to go from 0 upwards. Kill it with fire. I didn't write tests
for the fix, but it's JDBC only which isn't used in prod.
2. The Cassandra approach is unsafe. Even at delete consistency all, you
can't guarantee you deleted safely unless it _succeeds_. Turns out at
consistency all, if it didn't succeed you open yourself up to data
coming back from the dead. Which is what can happen here.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -332,14 +332,15 @@ public interface KeyValueService extends AutoCloseable {
     /**
      * For each cell, deletes all timestamps prior to the associated maximum timestamp. If this
      * operation fails, it's acceptable for this method to leave an inconsistent state, however
-     * implementations of this method <b>must</b> not leave a state where the timestamp listed
-     * has been inconsistently deleted and all the others have not been consistently deleted.
+     * implementations of this method <b>must</b> guarantee that, for each cell, if a value at the
+     * associated timestamp is inconsistently deleted, then all other values of that cell in the
+     * relevant range must have already been consistently deleted.
      *
      * @param tableRef the name of the table to delete the timestamps in.
      * @param rangesToDelete cells to be deleted, and the ranges of timestamps to delete for each cell
      */
     @Idempotent
-    void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete> rangesToDelete)
+    void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete> deletes)
             throws InsufficientConsistencyException;
 
     /**
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -337,7 +337,7 @@ public interface KeyValueService extends AutoCloseable {
      * relevant range must have already been consistently deleted.
      *
      * @param tableRef the name of the table to delete the timestamps in.
-     * @param rangesToDelete cells to be deleted, and the ranges of timestamps to delete for each cell
+     * @param deletes cells to be deleted, and the ranges of timestamps to delete for each cell
      */
     @Idempotent
     void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete> deletes)
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -330,17 +330,17 @@ public interface KeyValueService extends AutoCloseable {
     void deleteRange(TableReference tableRef, RangeRequest range);
 
     /**
-     * For each cell, deletes all timestamps prior to the associated maximum timestamp. Depending on the
-     * implementation, this may result in a range tombstone in the underlying KVS.
+     * For each cell, deletes all timestamps prior to the associated maximum timestamp. If this
+     * operation fails, it's acceptable for this method to leave an inconsistent state, however
+     * implementations of this method <b>must</b> not leave a state where the timestamp listed
+     * has been inconsistently deleted and all the others have not been consistently deleted.
      *
      * @param tableRef the name of the table to delete the timestamps in.
-     * @param maxTimestampExclusiveByCell exclusive maximum timestamp to delete for each cell.
-     * @param deleteSentinels if true, this method will also delete garbage collection sentinels.
+     * @param rangesToDelete cells to be deleted, and the ranges of timestamps to delete for each cell
      */
     @Idempotent
-    void deleteAllTimestamps(TableReference tableRef,
-            Map<Cell, Long> maxTimestampExclusiveByCell,
-            boolean deleteSentinels);
+    void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete> rangesToDelete)
+            throws InsufficientConsistencyException;
 
     /**
      * Truncate a table in the key-value store.
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TimestampRangeDelete.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TimestampRangeDelete.java
@@ -17,6 +17,7 @@
 package com.palantir.atlasdb.keyvalue.api;
 
 import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Lazy;
 
 @Immutable
 public interface TimestampRangeDelete {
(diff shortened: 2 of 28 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellRangeDeleter.java
```diff
@@ -108,29 +107,45 @@ public String toString() {
...
+     * if there are nodes down. This leads to a scenario where data can come back from the dead in some situations.
+     * This can only affect us if there is no covering tombstone, so we make sure that we've deleted all historical
+     * versions before deleting the latest cell (which is an Atlas level tombstone).
+     */
+    private void insertRangeTombstones(
+            CassandraClient client, Map<Cell, TimestampRangeDelete> deletes,
+            TableReference tableRef, long rangeTombstoneCassandraTs) throws TException {
+        insertTombstones(client, deletes, tableRef, rangeTombstoneCassandraTs, TimestampRangeDelete::timestamp);
+        insertTombstones(client, Maps.filterValues(deletes, TimestampRangeDelete::endInclusive),
+                tableRef, rangeTombstoneCassandraTs, delete -> delete.maxTimestampToDelete() + 1);
```
sandorw: Is this an exclusive bound (maxTimestampToDelete)? If so, can we make this clearer?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellRangeDeleter.java
```diff
@@ -108,29 +107,45 @@ public String toString() {
...
         wrappingQueryRunner.batchMutate("deleteAllTimestamps", client, ImmutableSet.of(tableRef), mutationMap,
                 deleteConsistency);
     }
 
-    private Mutation getMutation(Cell cell, long maxTimestampExclusive,
-            boolean deleteSentinel, long rangeTombstoneCassandraTimestamp) {
-        if (deleteSentinel) {
-            return Mutations.rangeTombstoneIncludingSentinelForColumn(cell.getColumnName(), maxTimestampExclusive,
+    private Mutation getMutation(Cell cell, TimestampRangeDelete delete,
+            long rangeTombstoneCassandraTimestamp, ToLongFunction<TimestampRangeDelete> exclusiveTimestampExtractor) {
```
sandorw: Looks like it's exclusive here, let's bubble it up further so that the logic is easier to check.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellRangeDeleter.java
```diff
@@ -108,29 +107,45 @@ public String toString() {
...
+        if (delete.deleteSentinels()) {
+            return Mutations.rangeTombstoneIncludingSentinelForColumn(
```
sandorw: Could push down the `TimestampRangeDelete` here given the existence of the `minTimestampToDelete` method.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/DbKvs.java
```diff
@@ -599,10 +600,9 @@ public void deleteRange(TableReference tableRef, RangeRequest range) {
     }
 
     @Override
-    public void deleteAllTimestamps(TableReference tableRef, Map<Cell, Long> maxTimestampExclusiveByCell,
-            boolean deleteSentinels) {
+    public void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete> maxTimestampByCell) {
```
sandorw: nit: rename maxTimestampsByCell

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueDeleter.java
```diff
@@ -57,20 +58,20 @@
...
                         .forEach(cells -> {
-                            Map<Cell, Long> maxTimestampByCellPartition = cells.stream()
+                            Map<Cell, TimestampRangeDelete> maxTimestampByCellPartition = cells.stream()
                                     .collect(Collectors.toMap(Function.identity(), entry.getValue()::get));
                             follower.run(entry.getKey(), maxTimestampByCellPartition.keySet());
                             if (sweeper.shouldAddSentinels()) {
                                 kvs.addGarbageCollectionSentinelValues(entry.getKey(),
                                         maxTimestampByCellPartition.keySet());
-                                kvs.deleteAllTimestamps(entry.getKey(), maxTimestampByCellPartition, false);
+                                kvs.deleteAllTimestamps(entry.getKey(), maxTimestampByCellPartition);
```
sandorw: nit: pull the shared call out of the if statement

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -330,17 +330,17 @@ default boolean supportsCheckAndSet() {
     void deleteRange(TableReference tableRef, RangeRequest range);
 
     /**
-     * For each cell, deletes all timestamps prior to the associated maximum timestamp. Depending on the
-     * implementation, this may result in a range tombstone in the underlying KVS.
+     * For each cell, deletes all timestamps prior to the associated maximum timestamp. If this
+     * operation fails, it's acceptable for this method to leave an inconsistent state, however
+     * implementations of this method <b>must</b> not leave a state where the timestamp listed
```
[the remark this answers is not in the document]
jeremyk-91: If I'm understanding this correctly, it's something like

```
implementations must guarantee that, for each cell, if a value at the associated timestamp is inconsistently deleted, then all other values of that cell in the range must have already been consistently deleted.
```

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/TimestampRangeDelete.java
```diff
@@ -0,0 +1,36 @@
...
+    boolean endInclusive();
+    boolean deleteSentinels();
+
+    default long maxTimestampToDelete() {
+        return timestamp() + (endInclusive() ? 0 : -1);
+    }
+
+    default long minTimestampToDelete() {
+        return deleteSentinels() ? Value.INVALID_VALUE_TIMESTAMP : Value.INVALID_VALUE_TIMESTAMP + 1;
+    }
```
jeremyk-91: nit: it looks a bit unexpected to not have `@Value.Lazy` annotations on these methods; as written I think we end up re-doing the calculations repeatedly (they're cheap, but still no reason to redo them).

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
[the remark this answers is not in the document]
sandorw: I think that is clearer together with the rest of the docs above.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellRangeDeleter.java
jeremyk-91: I've renamed the parameter in `insertTombstones` to be very clear that the function returns the _exclusive_ max timestamp to delete.

jeremyk-91:
@sandorw This should be ready for a second pass.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/Mutations.java
```diff
@@ -15,34 +15,34 @@
...
+import org.apache.cassandra.thrift.SlicePredicate;
+
+import com.palantir.atlasdb.keyvalue.api.TimestampRangeDelete;
 
 public final class Mutations {
 
     private Mutations() { }
 
-    public static Mutation rangeTombstoneForColumn(
+    public static Mutation fromTimestampRangeDelete(
```
sandorw: nit: I feel like this name could be clearer - `deleteTimestampRange`?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/thrift/Mutations.java
```diff
@@ -15,34 +15,34 @@
...
-        Deletion deletion = new Deletion()
-                .setTimestamp(timestampForRangeTombstone)
-                .setPredicate(SlicePredicates
-                        .rangeTombstoneIncludingSentinelForColumn(columnName, maxAtlasTimestampExclusive));
-
-        return new Mutation().setDeletion(deletion);
+    private static SlicePredicate getSlicePredicate(
+            byte[] columnName, TimestampRangeDelete delete, long maxExclusiveTimestamp) {
+        return delete.deleteSentinels()
+                ? SlicePredicates.rangeTombstoneIncludingSentinelForColumn(columnName, maxExclusiveTimestamp)
```
sandorw: I guess we still don't get to use the min timestamp function, but this is fine.

PR #3808, opened 2019-03-07
Simpler fix

j-baker:
Fix to bug 2 in https://github.com/palantir/atlasdb/pull/3807. Basically, instead of strengthening the contract of delete (it can delete in any order, but must externally appear as though it deleted in reverse order), have targeted sweep work with the current contract (indeterminate state if it fails).


sandorw:
Given that the KVS method exists, it seems that we might be better off fixing the logic there rather than special casing the targeted sweep logic. Anyone else using the KVS method could make the same mistake.

jeremyk-91:
Closing out as superseded by #3807 which was the approach we went with in the end.

PR #3811, opened 2019-03-07
Fix bug in PaxosQuorumChecker

gmaretic:
**Goals (and why)**:
When calling ``collectAsManyResponsesAsPossible`` with a node down, exceptions were not treated correctly and would result in blocking until timeout (which is set to 5 seconds in all usages).

Refactored to understand the code better, and made a change to check if we actually did get all of the responses in the loop condition.

Did not add any tests. Are there particularly concerning bits?

It's mostly a refactor, other than the bug fix. Did we introduce a change by accident?

PaxosResponses?

asap

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -49,6 +49,10 @@ develop
 
     *    - Type
          - Change
+
+    *    - |fixed|
+         - Fixed a bug in ``PaxosQuorumChecker`` causing a new timelock leader to block for 5 seconds before being able to serve requests if another node was unreachable.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3811>`__)
     
     *    - |improved|
          - ``CassandraKeyValueService`` now exposes a lightweight method for obtaining row keys.
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -179,7 +179,7 @@ public final class PaxosQuorumChecker {
                 shortcircuitIfQuorumImpossible);
         try {
             long deadline = System.nanoTime() + remoteRequestTimeout.toNanos();
-            while (receivedResponses.shouldProcessNextRequest()) {
+            while (receivedResponses.hasMoreRequests() && receivedResponses.shouldProcessNextRequest()) {
                 try {
                     Future<RESPONSE> responseFuture = responseCompletionService.poll(
                             deadline - System.nanoTime(),
@@ -203,7 +203,7 @@ public final class PaxosQuorumChecker {
                 Thread.currentThread().interrupt();
             }
 
-            if (receivedResponses.hasQuorum()) {
+            if (!receivedResponses.hasQuorum()) {
                 encounteredErrors.forEach(throwable -> log.warn(PAXOS_MESSAGE_ERROR, throwable));
             }
         }
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -178,7 +178,7 @@ public final class PaxosQuorumChecker {
         PaxosResponses<RESPONSE> receivedResponses = new PaxosResponses<>(remotes.size(), quorumSize,
                 shortcircuitIfQuorumImpossible);
         try {
-            long deadline = System.nanoTime() + remoteRequestTimeout.getNano();
+            long deadline = System.nanoTime() + remoteRequestTimeout.toNanos();
             while (receivedResponses.shouldProcessNextRequest()) {
                 try {
                     Future<RESPONSE> responseFuture = responseCompletionService.poll(
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -15,7 +15,7 @@
  */
 package com.palantir.paxos;
 
-import java.util.ArrayList;
+import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -33,9 +33,7 @@ import org.slf4j.LoggerFactory;
 import com.codahale.metrics.Meter;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.palantir.common.concurrent.MultiplexingCompletionService;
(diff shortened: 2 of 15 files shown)


docs/source/release_notes/release-notes.rst
```diff
@@ -49,6 +49,10 @@ develop
 
     *    - Type
          - Change
+
+    *    - |fixed|
+         - Fixed a bug in ``PaxosQuorumChecker`` causing a new timelock leader to block for 5 seconds before being able to serve requests if another node was unreachable.
```
felixdesouza: nit:
> block for 5 seconds (by default)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosTimestampBoundStore.java
```diff
@@ -208,17 +207,16 @@ SequenceAndBound forceAgreedState(long seq, @Nullable Long oldState) {
...
                 executor,
-                PaxosQuorumChecker.DEFAULT_REMOTE_REQUESTS_TIMEOUT_IN_SECONDS,
-                ONLY_LOG_ON_QUORUM_FAILURE);
-        if (responses.isEmpty()) {
+                PaxosQuorumChecker.DEFAULT_REMOTE_REQUESTS_TIMEOUT);
+        if (!responses.hasQuorum()) {
             return Optional.empty();
         }
-        return Optional.of(ImmutableSequenceAndBound.of(seq, responses.iterator().next().getValue()));
+        return Optional.of(ImmutableSequenceAndBound.of(seq, responses.getResponses().get(0).getValue()));
```
felixdesouza: `Iterables.getFirst`

leader-election-impl/src/main/java/com/palantir/paxos/PaxosResponses.java
```diff
@@ -13,26 +13,70 @@
...
  */
+
 package com.palantir.paxos;
 
-import com.google.common.base.Predicate;
-import com.palantir.paxos.persistence.generated.remoting.PaxosAcceptorPersistence;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PaxosResponses<T extends PaxosResponse> {
```
felixdesouza: I think this class should have a `.get` and a `.stream()`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosSynchronizer.java
```diff
@@ -61,12 +62,12 @@ public static void synchronizeLearner(PaxosLearner learnerToSynchronize,
...
         ExecutorService executor = PTExecutors.newCachedThreadPool();
-        List<PaxosValueResponse> responses = PaxosQuorumChecker.collectAsManyResponsesAsPossible(
+        PaxosResponses<PaxosValueResponse> responses = PaxosQuorumChecker.collectAsManyResponsesAsPossible(
                 ImmutableList.copyOf(paxosLearners),
                 learner -> ImmutablePaxosValueResponse.of(learner.getGreatestLearnedValue()),
                 executor,
-                PaxosQuorumChecker.DEFAULT_REMOTE_REQUESTS_TIMEOUT_IN_SECONDS);
-        return responses.stream()
+                PaxosQuorumChecker.DEFAULT_REMOTE_REQUESTS_TIMEOUT);
+        return responses.getResponses().stream()
```
felixdesouza: can revert this line with above comment

felixdesouza:
Looks good, would appreciate another set of eyes on it, since we paired on it.

docs/source/release_notes/release-notes.rst
gmaretic: This 5 seconds was hardcoded in

gmaretic:
@sandorw ? Since doesn't seem the others are around...

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -189,90 +173,50 @@ private PaxosQuorumChecker() {
...
+                shortcircuitIfQuorumImpossible);
         try {
-            long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(remoteRequestTimeoutInSec);
-            // handle responses
-            while (acksRecieved < quorumSize) {
+            long deadline = System.nanoTime() + remoteRequestTimeout.toNanos();
+            while (receivedResponses.shouldProcessNextRequest()) {
                 try {
-                    // check if quorum is impossible (nack quorum failure)
-                    if (shortcircuitIfQuorumImpossible && nacksRecieved > remotes.size() - quorumSize) {
```
sandorw: Why didn't this get us out of this logic before? Were we indefinitely retrying deeper down?

felixdesouza: `shortcircuitIfQuorumImpossible` was `false` and setting it to `true` isn't what we want

felixdesouza: and yes, we were retrying indefinitely

sandorw: Oops, yeah, this is a `&&`

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -189,90 +173,50 @@ private PaxosQuorumChecker() {
...
-        int nacksRecieved = 0;
-
+        PaxosResponses<RESPONSE> receivedResponses = new PaxosResponses<>(remotes.size(), quorumSize,
+                shortcircuitIfQuorumImpossible);
         try {
-            long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(remoteRequestTimeoutInSec);
-            // handle responses
-            while (acksRecieved < quorumSize) {
+            long deadline = System.nanoTime() + remoteRequestTimeout.toNanos();
+            while (receivedResponses.shouldProcessNextRequest()) {
```
sandorw: Can we make this clearer via `hasNextRequest` and `shouldProcessNextRequest` perhaps? We're hoping to exit here because there are no remaining requests, not because we choose not to process one.

felixdesouza:
Can do, but going from that PR to this would effectively be the same thing, how the fix is written changes entirely with the refactor

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -189,90 +173,50 @@ private PaxosQuorumChecker() {
...
             if (interrupted) {
                 Thread.currentThread().interrupt();
             }
 
-            if (onlyLogOnQuorumFailure && acksRecieved < quorumSize) {
-                for (Throwable throwable : toLog) {
-                    log.warn(PAXOS_MESSAGE_ERROR, throwable);
-                }
+            if (receivedResponses.hasQuorum()) {
+                encounteredErrors.forEach(throwable -> log.warn(PAXOS_MESSAGE_ERROR, throwable));
```
gokcanozakdag: I think we will be logging every failure even if we reach quorum. If one timelock node is in a bad shape, this would cause very large number of warns in service.log right?

gokcanozakdag: shouldn't this be `!receivedResponses.hasQuorum()` ?

sandorw:
Sure, don't split the refactors now.

PR #3818, opened 2019-03-08
PostgresVersionCheck now asserts as well as logs

mswintermeyer:
**Goals (and why)**:

### DIFF
--- a/atlasdb-dbkvs/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresVersionCheckTest.java
+++ b/atlasdb-dbkvs/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresVersionCheckTest.java
@@ -32,9 +32,9 @@ public class PostgresVersionCheckTest {
     @SuppressWarnings(value = "Slf4jConstantLogMessage")
     public void shouldLogErrorOn_9_2_24() {
         Logger log = Mockito.mock(Logger.class);
-        PostgresVersionCheck.checkDatabaseVersion("9.2.24", log);
         thrown.expect(AssertionError.class);
         thrown.expectMessage("The minimum supported version is ");
+        PostgresVersionCheck.checkDatabaseVersion("9.2.24", log);
         Mockito.verify(log).error(contains("The minimum supported version is {}"), Mockito.anyObject(),
                 Mockito.eq(PostgresVersionCheck.MIN_POSTGRES_VERSION));
         Mockito.verifyNoMoreInteractions(log);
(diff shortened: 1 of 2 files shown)


atlasdb-dbkvs/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresVersionCheckTest.java
```diff
@@ -32,6 +32,8 @@
     @SuppressWarnings(value = "Slf4jConstantLogMessage")
     public void shouldLogErrorOn_9_2_24() {
         Logger log = Mockito.mock(Logger.class);
+        thrown.expect(AssertionError.class);
+        thrown.expectMessage("The minimum supported version is ");
```
j-baker: this test doesn't seem right anymore. Lines 37-40 are now dead code, right?

mswintermeyer: assertAndLog is named oddly; the logging happens first and regardless of the assert failure. So checking the logging still passes, and it's useful for to check because assertions won't usually be enabled in an actual production environment.

jeremyk-91: I think James's point was that if an assertion error is thrown, we wouldn't run the rest of the body of the method?

I don't think OSS Atlas runs its tests with assertions enabled, so testing this might be a little less straightforward.

atlasdb-dbkvs/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/postgres/PostgresVersionCheckTest.java
```diff
@@ -32,6 +32,8 @@
...
         PostgresVersionCheck.checkDatabaseVersion("9.2.24", log);
         Mockito.verify(log).error(contains("The minimum supported version is {}"), Mockito.anyObject(),
                 Mockito.eq(PostgresVersionCheck.MIN_POSTGRES_VERSION));
```
jeremyk-91: Think we need to change this line. Using `assertAndLog` formats the string first, _and then_ later on calls `log.error("Assertion {} with exception ", message, t)` (see `AssertUtils:102`). So this call to verify will throw... but the test passes because the Mockito `ArgumentsAreDifferent` is an `AssertionError` and its string representation does contain the exception's message.

PR #3823, opened 2019-03-11
Better error message when writing too many cells

gmaretic:
**Goals (and why)**:
`java.lang.IllegalArgumentException: Dedicated row number must non-negative and strictly less than 64, but it is 64.` is not very informative when the real problem is that you're writing more than 6.4 million cells corresponding to the same shard in a single transaction.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
@@ -16,6 +16,8 @@
 package com.palantir.atlasdb.sweep.queue;
 
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueUtils.java
```diff
@@ -76,6 +81,20 @@ public static WriteInfo toWriteInfo(TableReference tableRef, Map.Entry<Cell, byt
         return WriteInfo.of(WriteReference.of(tableRef, cell, isTombstone), timestamp);
     }
 
+    public static void validateNumberOfCellsWritten(Collection<List<WriteInfo>> writes) {
+        writes.forEach(writesForShard -> {
+            int cellsWritten = writesForShard.size();
+            if (cellsWritten > MAX_CELLS_PER_SHARD) {
```
j-baker: this is precise, of course, but also weirdly probabilistic (fails based on hash function). Fancy changing it to just sum the total writes? Seems reasonable to say 'Atlas supports up to 6.4MM writes in a transaction'.

j-baker: otherwise you get this based on the murmur hash of your keys which is a bit weird

gmaretic: I generally agree with this, but are we sure we want to be this heavy handed? As it is, we always have an escape hatch where we can increase the sharding using the runtime config

sandorw: @gmaretic any resolution on this?

gmaretic: I'm going ahead and merge as is, since we couldn't get a conclusive ack that 6.4M is high enough. We can revisit at some later point if we really want to

PR #3824, opened 2019-03-12
Stop BG Sweep Priority Writes Only If Targeted Sweep Is Fully Enabled

jeremyk-91:
**Goals (and why)**:
- Address internal issues with storage on internal canine product.
- Ensure background sweep users (who have explicitly opted out of targeted sweep) don't accidentally find their background sweep to have stopped, if they are using default config for writing to the sweep queue.

- Stop writes to sweep.priority if and only if targeted sweep is fully enabled (that is, we are writing to the targeted sweep queue _and_ targeted sweep is on).
- Targeted sweep being enabled is a runtime config, but we take the first value on startup.

- Added tests as to whether the KVS contains at some level a `SweepStatsKVS` when it is initialized with various values of enableSweepQueueWrites/enabled for targeted sweep.

- Targeted Sweep's enabled flag is live reloadable, but in this implementation we take the first value on startup. This means that if enableSweepQueueWrites is on and we toggle targetedSweep's enabled flag, that does not change whether we're writing to the sweep priority table until the next restart. A solution could involve plumbing a flag on whether to actually write into the sweep stats KVS, but I don't think this is a significant overhead. Are there any easy solutions I've missed? I don't think this is important - am I wrong?
- The algorithm for checking whether we write to `sweep.priority` in the tests checks whether at some layer we have a `SweepStatsKeyValueService`. It is not resilient to cyclic dependencies, though I don't think we have any in normal creation.

**Where should we start reviewing?**: TransactionManagers.java


j-baker:
Why aren't we just turning targeted sweep on in that case?

j-baker:
Feels like the internal setup opted out of the default configuration and it didn't go well for them as a consequence.

At a minimum, they have unbounded storage; every write appends to a queue which can never be emptied.

gmaretic:
@j-baker we originally had separate configs because we wanted to be able to opt out of targeted sweep at runtime if it turns out to be necessary. At this point it may be reasonable to just have one install config and force people to sweep if they are writing to the queue

jeremyk-91:
Yep - if this pattern was common I'd have made this live-reloadable, but it seems rare to me (hence I went for this low-effort approach).

Yeah, a config with enable-writes on and targeted-sweep off is generally a bad idea if you have no intention of using targeted sweep. As @gmaretic mentions I think we did it this way because we wanted an escape hatch if targeted sweep behaved badly.

PR #3825, opened 2019-03-12
Use more up to date timestamp for correctness checks

sandorw:
**Goals (and why)**:
Tighten the constraints imposed by the `TimestampCorroborationConsistencyCheck`. It now uses a recent value from the puncher store rather than the unreadable timestamp (which is an hour old value). Previously, we couldn't catch any issues with the timestamp store if the stack had been used for less than an hour.

Call through to `KeyValueServicePuncherStore#get` rather than using `TransactionManager#getUnreadableTimestamp`.

No further tests added

Calling static methods on `KeyValueServicePuncherStore` is pretty gross here, as it's an implementation detail of the `Puncher`, but the relevant method isn't exposed on `Puncher`, nor is the `Puncher` exposed anywhere on `TransactionManager`. Furthermore, it means that different areas of the code would potentially break if we used a different `Puncher` implementation. 

However, I don't want to poke more holes through to the internals of TM to extract it, and we already call straight through in a similar manner in `TargetedSweepMetrics`.

Even worse - `KeyValueServicePuncherStore` is async initializable but also has static methods on it that only function correctly if it's been initialized. It's fine in this case since this is a callback that only occurs if the TM is initialized already (and therefore the `Puncher` is initialized as well).

Cleaning this up in a better way probably means breaking up `TransactionManager` into more reusable components rather than one giant class that does everything. 

`TransactionManagers`

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+import java.util.function.ToLongFunction;
 import java.util.stream.Collectors;
 
 import javax.ws.rs.ClientErrorException;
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -610,15 +614,19 @@ private static PersistentLockService createAndRegisterPersistentLockService(
...
+        // before you can return a TM, you need to return a TM to listen on ports, and you need to listen on
+        // ports in order to get a working timestamp service.
         if (isUsingTimeLock(atlasDbConfig, initialRuntimeConfig)) {
-            // Only do the consistency check if we're using TimeLock.
-            // This avoids a bootstrapping problem with leader-block services without async initialisation,
-            // where you need a working timestamp service to check consistency, you need to check consistency
-            // before you can return a TM, you need to return a TM to listen on ports, and you need to listen on
-            // ports in order to get a working timestamp service.
+            ToLongFunction<TransactionManager> conservativeBoundSupplier = txnManager -> {
+                Clock clock = GlobalClock.create(txnManager.getTimelockService());
```
jeremyk-91: Safe because this is a callback, by which point the puncher and timelock must already have been initialised.

PR #3827, opened 2019-03-12
[_transactions2] Part 27: :fire: CONTENTIOUS - Bound Timestamp Gaps in Transactions2 to 25,000,000

jeremyk-91:
**Goals (and why)**:
- Make clean-transactions-range a fast operation, by bounding commit timestamps to be 25,000,000 above start timestamps. Thus, if one needs to range scan transactions2 for _commit_ timestamps in the range `[L, U)`, it suffices to do a range query on _start_ timestamps in the range `[L-25,000,000, U)` rather than reading the history of the table up to `U`.
- Still allow moderately long running transactions to carry out the work they need to do.

- Implement a `GapLimitingTransactionService` that never putUnlessExists any `(start, commit)` pairs where commit is more than 25,000,000 above start.
- As a consequence, one only needs to look in the next block of 16 rows over when doing a range scan of commit timestamps.

- Added new tests for `GapLimitingTransactionService`

- Is this limit fair? I've timed how long it takes to burn through 25MM timestamps on some stacks (bearing in mind that they are already using startIdentifiedAtlasDbTransaction); feel free to approach me offline for details.
- There isn't really an escape strategy short of a patch if we need to force something through that consistently takes long enough that 25MM timestamps is a problem (though there won't be data corruption). Is this a concern?

**Where should we start reviewing?**: GapLimitingTransactionService.java

**Priority (whenever / two weeks / yesterday)**: this week. Kind-of blocks transactions2 rollouts.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/service/TransactionService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/service/TransactionService.java
@@ -48,6 +48,18 @@ public interface TransactionService extends AutoCloseable {
     @CheckForNull
     Long get(long startTimestamp);
 
+    /**
+     * Gets the commit timestamps associated with given start timestamps.
+     * For each individual timestamp, non-null responses may be cached on the client-side. Null responses must not be
+     * cached, as they could subsequently be updated.
+     *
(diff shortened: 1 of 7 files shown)


jeremyk-91:
Discussed offline - there is an easier way to achieve the effect this PR was aiming for, so we will not proceed with this.

PR #3830, opened 2019-03-14
0.39/better leader debugging

dxiao:
**Goals (and why)**:
Internal support ticket 79173

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -189,8 +189,15 @@ public final class PaxosQuorumChecker {
 
         } finally {
             // cancel pending futures (during failures)
+            boolean canceled = false;
             for (Future<RESPONSE> future : allFutures) {
-                future.cancel(true);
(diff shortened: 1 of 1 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -189,8 +189,15 @@ public RESPONSE call() throws Exception {
 
         } finally {
             // cancel pending futures (during failures)
+            boolean canceled = false;
             for (Future<RESPONSE> future : allFutures) {
-                future.cancel(false);
+                if (!future.isDone()) {
+                    future.cancel(true);
+                    canceled = true;
+                }
```
jeremyk-91: We might get some false positives this way I think, if the future completes between the check on `isDone()` and `cancel()` actually taking place. I _think_ it's probably enough to write

```
if (future.cancel(true)) { // safely returns false if the task has already completed
  canceled = true;
}
```

dxiao: oh! right, forgot that works.

PR #3835, opened 2019-03-18
coalescing start transaction calls into a single timelock rpc

gokcanozakdag:
**Goals (and why)**:
To reduce timelock load created by start transaction calls, as we are expecting an increase in start transaction calls after moving to thorough sweep (as read only transaction will be required to grab immutable ts locks)

Using `AutoBatcher` to coalesce calls into a single batched call. Using `LockTokenShare` to enable reference counting on immutable ts lock token.

Testing `CoalescingTransactionService` for both single and multi threaded calls.

I will try to refactor it a bit more to make it easier to follow.
We first need to release Timelock with Atlas 0.126.0, before merging this pr.

`CoalescingTransactionService`


lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -32,10 +36,12 @@
 public final class RemoteTimelockServiceAdapter implements TimelockService {
```
felixdesouza: I've found the logic here pretty hard to follow to be honest, haven't really gotten a good idea of what it should look like just yet, but yeah.

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -32,10 +36,12 @@
...
     private final LockLeaseService lockLeaseService;
     private final TimelockRpcClient timelockRpcClient;
+    private final CoalescingTransactionService transactionService;
 
     private RemoteTimelockServiceAdapter(TimelockRpcClient timelockRpcClient) {
         this.timelockRpcClient = timelockRpcClient;
         this.lockLeaseService = LockLeaseService.create(timelockRpcClient);
+        this.transactionService = CoalescingTransactionService.create(lockLeaseService);
```
felixdesouza: `startTransactionService`

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionServiceTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private static PartitionedTimestamps getPartitionedTimestamps(long startTs, int count) {
+        return ImmutablePartitionedTimestamps.builder()
+                .start(startTs)
+                .count(count)
+                .interval(NUM_PARTITIONS)
+                .build();
+    }
+
+    private static final class BlockingBatchedResponse implements Answer<StartTransactionResponseV4> {
```
felixdesouza: We have tests for autobatcher, I'm not sure it's useful to prove that it works everywhere we use it imo.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
+import com.palantir.common.base.Throwables;
+import com.palantir.lock.v2.LockImmutableTimestampResponse;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.StartTransactionResponseV4;
+import com.palantir.lock.v2.TimestampAndPartition;
+
+final class CoalescingTransactionService {
```
felixdesouza: The more I think about, the more I think that this class needn't exist. We should just autobatch by default inside `LockLeaseService`. The contract is the same, we don't need to test that it will work, since we have tests for autobatcher. Akin to caching.

Once you've moved out all the different misplaced pieces, it actually doesn't do much besides tunnel calls into autobatcher. The same way we wouldn't test a cache (unless that cache is doing different things such as in `CachingPartitionAllocator`), I don't think it's necessary to test autobatcher.

felixdesouza: This is a slightly wild view, so can do this last, but I think overall it does make things cleaner.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
```diff
@@ -0,0 +1,92 @@
...
+    private final DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher;
+
+    private CoalescingTransactionService(
+            DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher) {
```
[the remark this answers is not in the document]
j-baker: dunno about you, but generally I get the heebyjeebies when I see a constructor which actually does work. I end up struggling to think about initialization races, data sharing, etc etc, and then the second constructor must take care to make this work.

Much rather we just inject static state like this PR does currently, right?

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
+import com.palantir.common.base.Throwables;
+import com.palantir.lock.v2.LockImmutableTimestampResponse;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.StartTransactionResponseV4;
+import com.palantir.lock.v2.TimestampAndPartition;
+
+final class CoalescingTransactionService {
+    private final DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher;
```
j-baker: note: these autobatchers are closeable, and you have to close them. It won't work to just create them, you'll end up with a memory leak.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
```diff
@@ -0,0 +1,92 @@
...
+        List<StartIdentifiedAtlasDbTransactionResponse> result = new ArrayList<>();
+        while (result.size() < numberOfTransactions) {
+            result.addAll(split(lockLeaseService.startTransactions(numberOfTransactions - result.size())));
+        }
+        return result.subList(0, numberOfTransactions);
+    }
+
+    @VisibleForTesting
+    static List<StartIdentifiedAtlasDbTransactionResponse> split(StartTransactionResponseV4 batchedResponse) {
+        LockImmutableTimestampResponse immutableTsAndLock = batchedResponse.immutableTimestamp();
```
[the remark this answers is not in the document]
gokcanozakdag: hmm, actually moving split logic to StartTransactionReponse requires making LockTokenShare public rather than package private, which I really don't want to.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    private static List<StartIdentifiedAtlasDbTransactionResponse> startTransactions(
+            LockLeaseService lockLeaseService, int numberOfTransactions) {
+        List<StartIdentifiedAtlasDbTransactionResponse> result = new ArrayList<>();
+        while (result.size() < numberOfTransactions) {
+            result.addAll(split(lockLeaseService.startTransactions(numberOfTransactions - result.size())));
+        }
+        return result.subList(0, numberOfTransactions);
```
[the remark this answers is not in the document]
gokcanozakdag: not really, too much of paranoia

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -35,10 +35,12 @@
 import com.palantir.lock.v2.StartAtlasDbTransactionResponseV3;
 import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionRequest;
 import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.StartTransactionRequestV4;
+import com.palantir.lock.v2.StartTransactionResponseV4;
 import com.palantir.lock.v2.TimelockRpcClient;
 import com.palantir.logsafe.Preconditions;
 
-final class LockLeaseService {
+class LockLeaseService {
```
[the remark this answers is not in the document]
gokcanozakdag: mockito was complaining

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
gokcanozakdag: will add that after merging #3841

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionService.java
gokcanozakdag: I have concerns on adding the coalescing logic to lockLeaseService. I believe it would make it pretty hard to follow, but functionalities are complex enough to have its own place. Of course, this layered approach is not the only way of decomposing the functionality, but in any case I believe we need to separate them in some way. Tried something different here: #3844

PR #3837, opened 2019-03-18
refactor http clients

gokcanozakdag:
**Goals (and why)**:
Reduce cruft. Our static utility classes have large number of methods that is hard to maintain / modify.

Tried to remove as much code as possible without breaking public API's.

Some of the removed methods were public, and can be used by our large internal product. That will end up in a compile error that is easy to fix.

Probably there is still room to improve, but wanted to keep it incremental.


felixdesouza:
<3 please tick off the ones you've done on the document, not sure if it's everything.

PR #3839, opened 2019-03-18
Add some documentation inside AwaitingLeadershipProxy

felixdesouza:
Whilst trying to figure out where we can move things, I heeded the instruction to

> Do not modify, hide, or remove this line without considering impact on correctness

Whilst it is helpful to tell us to stay away, if we *do* need to change things, it's super annoying because there's no reasoning written down making my life harder. Adding it as I go along.

> NB: I'm not advocating needlessly commenting things everywhere, however it is not self-evident from the code. Since this class is heavily concurrency related, I'm not sure that it *is* possible for it to be that way, so a comment explaining correctness is warranted here imo.

Whenever


felixdesouza:
haha yeah unfortunately not

PR #3841, opened 2019-03-19
closeable timelock service

gokcanozakdag:
**Goals (and why)**:
Making timelock service closeable - so that we can register clean up tasks for RemoteTimelockServiceAdapter.


lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -78,4 +79,7 @@ default void tryUnlock(Set<LockToken> tokens) {
 
     long currentTimeMillis();
 
+    @Override
+    default void close() {}
```
felixdesouza: don't make this a default, otherwise we could end up skipping the close if we keep on wrapping.

felixdesouza: also, why do we have `TimeLockClient` and `RemoteTimelockServiceAdapter`, it's unclear what the distinction is between the two.

gokcanozakdag: TimelockClient takes a timelockService and adds few more higher level functionality, such as lock refreshing and autobatched unlocking. It is used for both embedded timelock and remote one.
`RemoteTimelockServiceAdapter` is only used for remote timelock, and converts `TimelockRpcClient` into `TimelockService`. Idea there is, timelock endpoints do not need to map one to one with timelock service methods because of perf optimisations. (Such as lock leases and batched start txn calls); and callers of `TimelockService` shouldn't care about these perf optimising endpoints, which are visible in `TimelockRpcClient` but not in `TimelockService`. Also these optimisations are only for remote timelock, not for embedded one.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TimestampDecoratingTimelockService.java
```diff
@@ -44,6 +44,11 @@ public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
         return decoratedTimestamps.getFreshTimestamps(numTimestampsRequested);
     }
 
+    @Override
+    public void close() {
+        delegate.close();
```
jeremyk-91: I think you need to close `decoratedTimestamps` if it is closeable - in practice this is going to be a `RequestBatchingTimestampService` which needs to be closed to shut down the disruptor stuff. 

It's a bit sad. Maybe this class can take a `CloseableTimestampService` instead?

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -24,7 +25,7 @@
 import com.palantir.timestamp.TimestampRange;
 
 @AutoDelegate
-public interface TimelockService {
+public interface TimelockService extends Closeable {
```
jeremyk-91: nit: I'd prefer `AutoCloseable` here. `Closeable` throws an `IOException` in its default signature, and the javadoc suggests that it tends to be used for streams. TimelockService is certainly a source/destination of data, but I wouldn't quite think of it as a stream.

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -176,6 +176,9 @@ public long currentTimeMillis() {
         return lockService.currentTimeMillis();
     }
 
+    @Override
+    public void close() {}
```
jeremyk-91: confirming: are we sure we don't want this to close the lock and timestamp services?

jeremyk-91: I suspect at least lock is closed separately, though (1) I would think close is idempotent, and (2) if we're not doing this I think it's worth a comment.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TimestampDecoratingTimelockService.java
gokcanozakdag: even better, I removed `TimestampDecoratingTimelockService`, apparently we don't use it anymore (I guess we were using it to make request batching runtime configurable?)

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
gokcanozakdag: none of the lock and timestamp services are closeable, so should be fine to do nothing here.

felixdesouza: that's not true! legacy timelock service can receive a local lock service - it's hidden behind layers and layers - and it has executors e.g. see `LockServiceImpl`

gokcanozakdag: I was trying to say `LockService` interface is not closeable, so `LegacyTimelockService` won't be able to close underlying services - and were not closing them before this pr.
I guess your concern is marking TimelockService as closeable would give the impression that closing it would actually close underlying resources also (which is a pretty valid assumption), and may cause bugs in future, even though it is not changing current behaviour. Is that right? @felixdesouza @jeremyk-91

felixdesouza: Yes. Part of me feels like if this is a layer on top, then we will never think to close the underlying layers. i.e. I think this change without doing due diligence on the lower layers does more harm than good in the long term.

In my mind, I think the correct approach is what you employed fairly recently: you have a client that's solely responsible for RPC, and it doesn't wrap anything. Then have a corresponding class that may or may not extend the RPC interface (don't care at this point), but it *is* closeable. Otherwise, this will be impossible to track imo.

If you have this model, we don't necessarily have infinite wrappers, we can just have one class that implements the non-rpc "client" and that can customise behaviour however it pleases. We don't have to worry and think: "oh what timestamp service do I actually have here, is the ordering correct?" etc.

As an aside/rambling at this point, the fact we have all of these interfaces are a lie, they're not necessarily interchangeable. I'd rather have something that combined this state and presented it as is, (it can do nesting/wrapping inside the class if it needs to).

gokcanozakdag: that makes sense, I guess marking `TimelockService` as closeable is not the correct approach here, given all we wanted was registering close callback of `RemoteTimelockServiceAdapter`. Opened another pr doing that: #3847

PR #3842, opened 2019-03-19
refactor timelock migrator

gokcanozakdag:
**Goals (and why)**:
Currently TimelockMigrator creates its own proxy to talk to remote timelock service, which is unnecessary and makes it hard to manage proxies. This pr aims to improve that.

Rather than creating its own proxy, now it gets a `TimelockManagementService`. It also makes tests easier to read.

Refactored existing tests to get rid of wiremock.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -788,12 +788,18 @@ public abstract class TransactionManagers {
             String userAgent) {
         Supplier<ServerListConfig> serverListConfigSupplier =
                 getServerListConfigSupplierForTimeLock(config, runtimeConfigSupplier);
-        TimeLockMigrator migrator =
-                TimeLockMigrator.create(metricsManager,
-                        serverListConfigSupplier, invalidator, userAgent, config.initializeAsync());
+
+        LockAndTimestampServices lockAndTimestampServices =
+                getLockAndTimestampServices(metricsManager, serverListConfigSupplier, userAgent);
+
+        TimeLockMigrator migrator = TimeLockMigrator.create(
+                lockAndTimestampServices.timestampManagement(),
+                invalidator,
+                config.initializeAsync());
         migrator.migrate(); // This can proceed async if config.initializeAsync() was set
+
         return ImmutableLockAndTimestampServices.copyOf(
-                getLockAndTimestampServices(metricsManager, serverListConfigSupplier, userAgent))
+                lockAndTimestampServices)
                 .withMigrator(migrator);
     }
 
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -798,8 +798,7 @@ public abstract class TransactionManagers {
                 config.initializeAsync());
         migrator.migrate(); // This can proceed async if config.initializeAsync() was set
 
-        return ImmutableLockAndTimestampServices.copyOf(
-                lockAndTimestampServices)
+        return ImmutableLockAndTimestampServices.copyOf(lockAndTimestampServices)
                 .withMigrator(migrator);
     }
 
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/startup/TimeLockMigratorTest.java
```diff
@@ -16,104 +16,53 @@
...
     public void invalidationDoesNotProceedIfTimelockPingUnsuccessful() {
-        wireMockRule.stubFor(PING_MAPPING.willReturn(aResponse().withStatus(500)));
+        when(timestampManagementService.ping()).thenThrow(EXCEPTION);
 
-        TimeLockMigrator migrator =
-                TimeLockMigrator.create(
-                        MetricsManagers.createForTests(),
-                        timelockConfig.toNamespacedServerList(), invalidator, USER_AGENT);
+        TimeLockMigrator migrator = TimeLockMigrator.create(timestampManagementService, invalidator);
         assertThatThrownBy(migrator::migrate).isInstanceOf(AtlasDbDependencyException.class);
```
jeremyk-91: nit: `hasCause(EXCEPTION)` as well?

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/startup/TimeLockMigratorTest.java
```diff
@@ -122,73 +71,47 @@ public void invalidationDoesNotProceedIfTimelockPingUnsuccessful() {
...
         when(invalidator.backupAndInvalidate()).thenThrow(new IllegalStateException());
 
-        TimeLockMigrator migrator =
-                TimeLockMigrator.create(
-                        MetricsManagers.createForTests(),
-                        timelockConfig.toNamespacedServerList(), invalidator, USER_AGENT);
+        TimeLockMigrator migrator = TimeLockMigrator.create(timestampManagementService, invalidator);
         assertThatThrownBy(migrator::migrate).isInstanceOf(IllegalStateException.class);
-        wireMockRule.verify(0, postRequestedFor(urlEqualTo(TEST_ENDPOINT)));
+        verify(timestampManagementService, times(0)).fastForwardTimestamp(anyLong());
```
jeremyk-91: nit: usually we use `never()` rather than `times(0)` for readability - they're equivalent

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/startup/TimeLockMigratorTest.java
```diff
@@ -122,73 +71,47 @@ public void invalidationDoesNotProceedIfTimelockPingUnsuccessful() {
...
-                        MetricsManagers.createForTests(),
-                        () -> timelockConfig.toNamespacedServerList(), invalidator, USER_AGENT, true);
+        TimeLockMigrator migrator = TimeLockMigrator.create(timestampManagementService, invalidator, true);
         migrator.migrate();
 
-        Awaitility.await()
-                .atMost(30, TimeUnit.SECONDS)
-                .pollInterval(1, TimeUnit.SECONDS)
-                .until(migrator::isInitialized);
+        waitUntilInitialized(migrator);
```
jeremyk-91: Thanks, this is much better than the original!  🎉

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -788,12 +788,17 @@ private static LockAndTimestampServices createRawServicesFromTimeLock(
...
+
+        TimeLockMigrator migrator = TimeLockMigrator.create(
+                lockAndTimestampServices.timestampManagement(),
+                invalidator,
+                config.initializeAsync());
         migrator.migrate(); // This can proceed async if config.initializeAsync() was set
-        return ImmutableLockAndTimestampServices.copyOf(
-                getLockAndTimestampServices(metricsManager, serverListConfigSupplier, userAgent))
+
+        return ImmutableLockAndTimestampServices.copyOf(lockAndTimestampServices)
```
jeremyk-91: This changes the order the objects are created in, so the lock and timestamp services come before the migrator. `getLockAndTimestampServices` involves a bunch of proxies and/or thin wrappers above them (e.g. timelock-timestamp service adapter, leasing timelock client). It doesn't look like these are/can be accessed statically, so this should be fine.

PR #3844, opened 2019-03-20
coalescing start txn - alternative impl

gokcanozakdag:
**Goals (and why)**:
Original pr: #3835 is hard to follow. This is an attempt to make it cleaner.

I think one problem with the original pr was having multiple classes being aware of LockTokenShare concept, so this one reduces it to a single class.
I believe the other problem was convoluted use of `LockLeaseService`, `TimelockRpcClient` and `CoalescingTransactionService` inside `RemoteTimelockServiceAdapter`, this pr tries to make it a bit clear what classes are responsible for what.


lock-api/src/main/java/com/palantir/lock/client/TransactionCoalescingLockDecoratorService.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private TransactionCoalescingLockDecoratorService(
+            DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher,
+            LockDecoratorService delegate) {
+        this.autobatcher = autobatcher;
+        this.delegate = delegate;
+    }
+
+    static TransactionCoalescingLockDecoratorService create(LockDecoratorService delegate) {
+        return new TransactionCoalescingLockDecoratorService(DisruptorAutobatcher.create(batch -> {
```
sandorw: nit: probably more readable to break out this function

lock-api/src/main/java/com/palantir/lock/client/LockDecoratorService.java
```diff
@@ -0,0 +1,35 @@
...
+import com.palantir.lock.v2.LockRequest;
+import com.palantir.lock.v2.LockResponse;
+import com.palantir.lock.v2.LockToken;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.StartTransactionResponseV4;
+
+interface LockDecoratorService {
+    LockImmutableTimestampResponse lockImmutableTimestamp();
+    StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction();
+    StartTransactionResponseV4 startTransactions(int numTransactions);
```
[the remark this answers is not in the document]
sandorw: Discussed a bit in person, but the abstraction feels off here - there are two things implementing this interface. This endpoint only exists because one of them wants it to exist, not because any client ever calls it. As a result, it's an internal detail, but still has to be exposed on the interface, but we don't expect anyone to call it.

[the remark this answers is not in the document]
gokcanozakdag: I think the abstraction was not intuitive enough, so I am removing it all together.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,204 @@
...
+            for (int i = 0; i < numTransactions; i++) {
+                batch.get(i).result().set(startTransactionResponses.get(i));
+            }
+        };
+    }
+
+    private static List<StartIdentifiedAtlasDbTransactionResponse> getStartTransactionResponses(
+            LockLeaseService lockLeaseService, int numberOfTransactions) {
+        List<StartIdentifiedAtlasDbTransactionResponse> result = new ArrayList<>();
+        while (result.size() < numberOfTransactions) {
```
sandorw: Seems slightly subtle that a single batch request may return fewer transactions than requested. Is this the right layer to handle that vs. `LockLeaseService`? Similar to `TimestampService#getFreshTimestamps`, let's document appropriately (on `TimelockService`?) that this is the case.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,204 @@
...
+                .collect(Collectors.toSet());
+        Set<LockToken> resultLockTokens = lockTokens.stream()
+                .filter(refreshedTokens::contains)
+                .collect(Collectors.toSet());
+
+        return Sets.union(resultLockTokenShares, resultLockTokens);
+    }
+
+    Set<LockToken> unlock(Set<LockToken> tokens) {
+        Set<LockToken> lockTokens = filterOutTokenShares(tokens);
```
sandorw: nit: keep the next line together for consistent style

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,204 @@
...
+        return Sets.union(resultLockTokenShares, resultLockTokens);
+    }
+
+    @Override
+    public void close() {
+        autobatcher.close();
+    }
+
+    private Set<LockToken> refreshLockLeasesInternal(Set<LockToken> toRefresh) {
+        if (!toRefresh.isEmpty()) {
```
sandorw: Can we push these checks down to `LockLeaseService` to simplify the code here? `LockLeaseService#refreshTokens` has an analogous shortcut already.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,204 @@
...
+        Set<LockTokenShare> lockTokenShares = filterLockTokenShares(tokens);
+        Set<LockToken> referencedTokens = lockTokenShares.stream()
+                .map(LockTokenShare::sharedLockToken)
+                .collect(Collectors.toSet());
+
+        Set<LockToken> toUnlock = reduceForUnlock(lockTokenShares);
+        Set<LockToken> toRefresh = Sets.difference(referencedTokens, toUnlock);
+
+        //calling refresh to check validity of tokens that won't be unlocked
+        Set<LockToken> refreshed = refreshLockLeasesInternal(toRefresh);
```
sandorw: I don't immediately follow why we need to refresh, then unlock. Can you explain?

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
gokcanozakdag: I think `TimelockRpcClient` is the right place to document this, as `TimelockService` actually do not have a `startTransactions` method - and shouldn't have as it requires specific logic to handle sharing immutable timestamp lock token.
I am a bit concerned about moving the handling less-than-expected case to a lower layer, as this is the only place that requires this guarantee. We may end up using some other delegate at some point (rather than using lockLeaseService), in that case, we will need to make sure that new thing handles it correctly.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
gokcanozakdag: added a javadoc, tried to explain there. Does that make sense?

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -63,20 +64,21 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
...
 
-    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        StartAtlasDbTransactionResponseV3 response = delegate.deprecatedStartTransaction(
-                StartIdentifiedAtlasDbTransactionRequest.createForRequestor(clientId));
+    StartTransactionResponseV4 startTransactions(int batchSize) {
+        StartTransactionRequestV4 request = StartTransactionRequestV4.createForRequestor(clientId, batchSize);
+        StartTransactionResponseV4 response = delegate.startTransactions(request);
 
-        Lease lease = response.getLease();
+        Lease lease = response.lease();
```
felixdesouza: still confused between lease and lock token share and why they're separate, but that can wait

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+    @Before
+    public void before() {
+        coalescingTransactionStarter = CoalescingTransactionStarter.create(lockLeaseService);
+    }
+
+    @Test
+    public void splitShouldYieldCorrectStartTransactionResponses_singleTransaction() {
+        StartTransactionResponseV4 batchedResponse = getStartTransactionResponse(10, 1);
+
+        assertThat(CoalescingTransactionStarter.split(batchedResponse))
```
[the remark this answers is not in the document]
felixdesouza: basically, these two test methods aren't useful, you're testing this in the ones below.

Remove these and make `split` private again.

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Test
+    public void shouldDeriveStartTransactionResponseFromBatchedResponse_multipleTransactions() {
+        StartTransactionResponseV4 batchResponse = getStartTransactionResponse(40, 3);
+        when(lockLeaseService.startTransactions(3))
+                .thenReturn(batchResponse);
+
+        List<StartIdentifiedAtlasDbTransactionResponse> responses = requestBatches(3);
+        assertThatStartTransactionResponsesAreUnique(responses);
+        assertThat(responses)
```
felixdesouza: nit:
```java
assertThat(responses)
  .satisfies(CoalescingTransactionsStarterTest::responsesAreUnique)
  .hasSize(3)
  .allSatisfy(startTransactionResponse -> derivableFromBatchedResponse(...));
```

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    private List<StartIdentifiedAtlasDbTransactionResponse> requestBatches(int size) {
+        List<BatchElement<Void, StartIdentifiedAtlasDbTransactionResponse>> elements = IntStream.range(0, size)
+                .mapToObj(unused -> ImmutableTestBatchElement.builder()
+                        .argument(null)
+                        .result(SettableFuture.create())
+                        .build())
+                .collect(toList());
+        CoalescingTransactionStarter.consumer(lockLeaseService).accept(elements);
```
felixdesouza: this feels quite iffy, wondering whether there is a lighter approach to "mocking"

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+                .extracting(response -> response.immutableTimestamp().getLock().getRequestId())
+                .doesNotHaveDuplicates();
+
+        assertThat(responses)
+                .as("Each response should have a different start timestamp")
+                .extracting(response -> response.startTimestampAndPartition().timestamp())
+                .doesNotHaveDuplicates();
+    }
+
+    private void assertDerivableFromBatchedResponse(
```
felixdesouza: nit: static

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+                .mapToObj(unused -> ImmutableTestBatchElement.builder()
+                        .argument(null)
+                        .result(SettableFuture.create())
+                        .build())
+                .collect(toList());
+        CoalescingTransactionStarter.consumer(lockLeaseService).accept(elements);
+        return Futures.getUnchecked(Futures.allAsList(Lists.transform(elements, BatchElement::result)));
+    }
+
+    private void assertThatStartTransactionResponsesAreUnique(
```
felixdesouza: nit: static

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,201 @@
...
+                .start(startTs)
+                .count(count)
+                .interval(NUM_PARTITIONS)
+                .build();
+    }
+
+    @Value.Immutable
+    interface TestBatchElement extends BatchElement<Void, StartIdentifiedAtlasDbTransactionResponse> {
+        @Override
+        @Nullable Void argument();
```
felixdesouza: is this only here so you can specify `@Nullable`?

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
felixdesouza: it feels like you want to control the batch size given to the autobatcher, or that the consumer interface needs to be clearer i.e.:

`Function<List<Request<T>, Map<Request<T>, R>>`

I dunno, just feels off.

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
felixdesouza: I suppose my confusion comes from the fact that the response encodes both `LockLease` and `LockTokenShare` data, yet we want to establish an abstraction/barrier between the two. However, it's already leaked/it's already broken.

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,206 @@
...
+            for (int i = 0; i < numTransactions; i++) {
+                batch.get(i).result().set(startTransactionResponses.get(i));
+            }
+        };
+    }
+
+    private static List<StartIdentifiedAtlasDbTransactionResponse> getStartTransactionResponses(
+            LockLeaseService lockLeaseService, int numberOfTransactions) {
+        List<StartIdentifiedAtlasDbTransactionResponse> result = new ArrayList<>();
+        while (result.size() < numberOfTransactions) {
```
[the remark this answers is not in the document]
gokcanozakdag: yeah that is not guaranteed

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
gokcanozakdag: yep :(

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
gokcanozakdag: yeah, it seems like too much tooling for a test; but wanted to cover as much code as possible without testing `AutoBatcher` - so consumer has exactly the same signature required by the `AutoBatcher`. Used a similar approach with `RequestBatchingTimetampServiceTest`

docs/source/release_notes/release-notes.rst
```diff
@@ -67,6 +67,14 @@ develop
          - `AtlasDbHttpClients`, `FeignOkHttpClients` and `AtlasDbFeignTargetFactory` are refactored to get rid of deprecated methods and overused overloads.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3837>`__)
 
+    *    - |improved|
+         - Concurrent calls to `TimelockService.startIdentifiedAtlasDbTransaction()` now coalesced into a single Timelock rpc to reduce load on Timelock.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3844>`__)
+
+    *    - |devbreak|
```
j-baker: who's using this?

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
j-baker: did you figure this out? We can quickly chat through it if you'd like.

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -60,7 +65,7 @@ private LockTokenShare(ReferenceCounter referenceCounter, LockToken token) {
      * @return referenced shared lock token if all lock token shares are unlocked; that is lock token on server side is
      * good to be unlocked.
      */
-    public synchronized Optional<LockToken> unlock() {
+    synchronized Optional<LockToken> unlock() {
```
j-baker: noted this in the PR that added this class, but I still find this odd. There's levels of synchronization that imply single-threaded-ness; that only one thread will ever be returned the lock token to unlock.

However, there's a race and this could _possibly_ happen twice if the race is hit. And, unlock is not idempotent, which is also weird; it means that the person abusing this class could see different behaviour.

As before, I'd prefer our reference counter return a single share the returned token to unlock; the last person to unlock is handed the token to unlock and the race goes away.

Does this make sense?

lock-api/src/main/java/com/palantir/lock/client/CoalescingTransactionStarter.java
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * A service responsible for coalescing multiple start transaction calls into a single start transactions call. This
+ * service also handles creating {@link LockTokenShare}'s to enable multiple transactions sharing a single immutable
+ * timestamp.
+ *
+ * Callers of this class should use {@link #unlock(Set)} and {@link #refreshLockLeases(Set)} for returned lock tokens,
+ * rather than directly calling delegate lock service.
+ */
+final class CoalescingTransactionStarter implements AutoCloseable {
```
j-baker: just call this TransactionStarter!

j-baker: it's not like we have a non-CoalescingTransactionStarter!

lock-api/src/test/java/com/palantir/lock/client/CoalescingTransactionStarterTest.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final LockImmutableTimestampResponse IMMUTABLE_TS_RESPONSE =
+            LockImmutableTimestampResponse.of(1L, LockToken.of(UUID.randomUUID()));
+
+    private static final Lease LEASE = Lease.of(
+            LeaderTime.of(LeadershipId.random(), NanoTime.createForTests(1L)),
+            Duration.ofSeconds(1L));
+
+    @Before
+    public void before() {
+        coalescingTransactionStarter = CoalescingTransactionStarter.create(lockLeaseService);
```
j-baker: you probably want to close this after each test? otherwise will leak threads

docs/source/release_notes/release-notes.rst
gokcanozakdag: our internal backup service

PR #3847, opened 2019-03-21
register remoteTimelockServiceAdapter closing callback

gokcanozakdag:
**Goals (and why)**:
`RemoteTimelockServiceAdapter` will be using disruptor to batch multiple start transaction calls, and it needs to be closed, that requires registering a closing callback. Rather than marking `TimelockService` closeable as I have tried here: #3841, this pr directly registers the `RemoteTimelockServiceAdapter::close` method, without relying on its wrappers closing it.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -15,7 +15,6 @@
  */
 package com.palantir.atlasdb.factory;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
@@ -303,8 +302,7 @@ public abstract class TransactionManagers {
                 timestampSupplier,
                 managementSupplier,
                 atlasFactory.getTimestampStoreInvalidator(),
-                userAgent(),
-                closeables);
+                userAgent());
         adapter.setTimestampService(lockAndTimestampServices.timestamp());
 
         KvsProfilingLogger.setSlowLogThresholdMillis(config().getKvsSlowLogThresholdMillis());
@@ -407,6 +405,7 @@ public abstract class TransactionManagers {
         TransactionManager instrumentedTransactionManager =
                 AtlasDbMetrics.instrument(metricsManager.getRegistry(), TransactionManager.class, transactionManager);
 
+        instrumentedTransactionManager.registerClosingCallback(lockAndTimestampServices.close());
         instrumentedTransactionManager.registerClosingCallback(transactionService::close);
         instrumentedTransactionManager.registerClosingCallback(schemaInstaller::close);
         instrumentedTransactionManager.registerClosingCallback(targetedSweep::close);
@@ -664,8 +663,7 @@ public abstract class TransactionManagers {
                         time,
                         timeManagement,
                         invalidator,
-                        userAgent,
-                        new ArrayList<>());
+                        userAgent);
         TimeLockClient timeLockClient = TimeLockClient.withSynchronousUnlocker(lockAndTimestampServices.timelock());
         return ImmutableLockAndTimestampServices.builder()
                 .from(lockAndTimestampServices)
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -818,13 +824,15 @@ private static LockAndTimestampServices createRawServicesFromTimeLock(
...
+            String userAgent,
+            List<AutoCloseable> closeables) {
         ServiceCreator creator = ServiceCreator.withPayloadLimiter(metricsManager, userAgent, timelockServerListConfig);
         LockService lockService = creator.createService(LockService.class);
         TimelockRpcClient timelockClient = creator.createService(TimelockRpcClient.class);
-        TimelockService timelockService = RemoteTimelockServiceAdapter.create(timelockClient);
+        RemoteTimelockServiceAdapter timelockService = RemoteTimelockServiceAdapter.create(timelockClient);
         TimestampManagementService timestampManagementService = creator.createService(TimestampManagementService.class);
 
+        closeables.add(timelockService::close);
```
jeremyk-91: nit: I think you can just do `RemoteTimelockServiceAdapter timelockServiceAdapter = initializeCloseable(() -> RemoteTimelockServiceAdapter.create(timelockClient), closeables);`

Also I guess I don't really like naming the `TimelockServiceAdapter` `timelockService`, though that's just a matter of taste

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -405,7 +407,6 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         TransactionManager instrumentedTransactionManager =
                 AtlasDbMetrics.instrument(metricsManager.getRegistry(), TransactionManager.class, transactionManager);
 
-        instrumentedTransactionManager.registerClosingCallback(lockAndTimestampServices.close());
```
jeremyk-91: I don't think this is something we can drop entirely; this ensures that the lock and timestamp services are closed _when the transaction manager is closed_, as compared to `closeables` in this method which seems to be things that are closed _if initialization fails_.

jeremyk-91:
e.g. 
```
    private static final String TIMELOCK_FF_PATH2
            = "/" + CLIENT + "/timestamp-management/fast-forward?currentTimestamp=" + (Long.MIN_VALUE + 1);
    private static final MappingBuilder TIMELOCK_FF_MAPPING2 = post(urlEqualTo(TIMELOCK_FF_PATH2));

    @Test
    public void manyManyManagers() {
        availableServer.stubFor(TIMELOCK_FF_MAPPING2.willReturn(aResponse()
                .withStatus(204)));
        for (int i = 0; i < 10; i++) {
            TransactionManager tm = TransactionManagers.builder()
                    .config(ImmutableAtlasDbConfig.builder().keyValueService(new InMemoryAtlasDbConfig())
                            .timelock(mockClientConfig)
                            .build())
                    .userAgent("test")
                    .globalMetricsRegistry(new MetricRegistry())
                    .globalTaggedMetricRegistry(DefaultTaggedMetricRegistry.getDefault())
                    .registrar(environment)
                    .runtimeConfigSupplier(() -> Optional.of(AtlasDbRuntimeConfig.defaultRuntimeConfig()))
                    .addSchemas(GenericTestSchema.getSchema())
                    .build()
                    .serializable();
            tm.close();
        }

        System.out.println("42"); //breakpoint
    }
```

The thread dump contains 20 auto-batcher, 19 polling refreshables and 21 timelockclient-refresh-0 threads, but I think it shouldn't have any

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -650,19 +650,4 @@ private static TimeLockClientConfig getTimelockConfigForServers(List<String> ser
                         .build())
                 .build();
     }
-
-    private TransactionManagers.LockAndTimestampServices createLockAndTimestampServicesForConfig(
```
gokcanozakdag: this one was not used, so removed that

Issue #3560, opened 2018-10-05
DbKvs check and set is inconsistent with other KVSes


j-baker:
DbKvs check and set throws check and set exceptions. If the CAS is from value exists to value exists, it will throw a CheckAndSetException with appropriate current value set. If it is from value doesn't exist to value exists, this will not be set, which is different to every other KVS.

We should remove this discrepancy. There are two solutions

1. Remove the functionality - it's usually easy to fix it.
2. Fix the SQL. Likely more effort and harder to test fixes, but also plausible.

berler:
Currently internal products use this API specifically to clear locks in the persistent lock table if we think that some process didn't correctly unlock a persistent lock (perhaps it died while holding the lock). We use checkAndSet explicitly to avoid a race condition where two processes attempt to clear out a lock, and one process was still waiting to acquire the lock. The race condition looks like this:

- process 1 and 3 think some lock needs to be cleared (based on its current value etc)
- process 1 clears the lock
- process 2 now acquires the lock (it was waiting to get the lock, and now it can because process 1 just cleared it)
- process 3 clears the lock

Using checkAndSet for the clear operation allows process 3 in this case to *not* clear the lock because the value wasn't equal to the old "stale" one that we were trying to clear. The desired behavior after this order of steps would be that process 2 still has the lock.

How would you handle this without checkAndSet?

PR #3848, opened 2019-03-21
Fix NPE in KeyValueServiceCoordinationStore

gmaretic:
**Goals (and why)**:
On DbKvs, if we fail to CAS, we return a KAEE without the current value present, see #3560. We were assuming that this information was present when updating the coordination value, resulting in an NPE on CAS failure.
 
If we get a KAEE with no details, we instead read the coordination value from the KVS. There is no guarantee of atomicity of course, but that was not part of the contract and is now specifically cleared up in the java doc.

Just some quick unit tests.

Discussed with @jeremyk-91 , and it should be safe to not have atomicity in this sense.

ASAP, it's blocking another product

### DIFF
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationServiceImpl.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationServiceImpl.java
@@ -25,7 +25,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Iterables;
-import com.palantir.atlasdb.keyvalue.api.KeyValueService;
 import com.palantir.atlasdb.keyvalue.impl.CheckAndSetResult;
 import com.palantir.logsafe.SafeArg;
 
@@ -51,8 +50,7 @@ public class CoordinationServiceImpl<T> implements CoordinationService<T> {
 
     /**
      * In case of failure, the returned value and bound are guaranteed to be the ones that were in the KVS at the time
-     * of CAS failure only if {@link KeyValueService#getCheckAndSetCompatibility()} is
-     * {@link com.palantir.atlasdb.keyvalue.api.CheckAndSetCompatibility#SUPPORTED_DETAIL_ON_FAILURE}.
+     * of CAS failure only if the implementation of the CoordinationStore makes such guarantees.
      */
     @Override
     public CheckAndSetResult<ValueAndBound<T>> tryTransformCurrentValue(Function<ValueAndBound<T>, T> transform) {
--- a/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationServiceImpl.java
+++ b/atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationServiceImpl.java
@@ -25,6 +25,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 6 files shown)


atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/CoordinationServiceImpl.java
```diff
@@ -48,6 +49,11 @@ public CoordinationServiceImpl(CoordinationStore<T> store) {
         return Optional.of(cachedReference);
     }
 
+    /**
+     * In case of failure, the returned value and bound are guaranteed to be the ones that were in the KVS at the time
+     * of CAS failure only if {@link KeyValueService#getCheckAndSetCompatibility()} is
+     * {@link com.palantir.atlasdb.keyvalue.api.CheckAndSetCompatibility#SUPPORTED_DETAIL_ON_FAILURE}.
+     */
```
jeremyk-91: I think this one should be along the lines of "in case of failure, the returned value and bound are guaranteed to be the ones that caused the CAS to fail only if the implementation of the CoordinationStore makes such guarantees".

We shouldn't mention `KeyValueService` at this layer, since I could always write a `TimeLockPaxosCoordinationStore`, for example.

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -268,10 +273,14 @@ void putUnlessValueExists(long sequenceNumber, T value) {
...
-                    .map(this::deserializeSequenceAndBound)
-                    .collect(Collectors.toList()));
+            if (e.getActualValues() != null) {
+                return ImmutableCheckAndSetResult.of(false, e.getActualValues()
+                        .stream()
+                        .map(this::deserializeSequenceAndBound)
+                        .collect(Collectors.toList()));
+            } else {
+                return getMostRecentValueAndBound(oldValue, newValue);
+            }
```
jeremyk-91: nit: `else` not required with returns?

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -294,7 +303,7 @@ private T deserializeValue(byte[] contents) {
         }
     }
 
-    private byte[] serializeSequenceAndBound(SequenceAndBound sequenceAndBound) {
+    byte[] serializeSequenceAndBound(SequenceAndBound sequenceAndBound) {
```
jeremyk-91: nit: `@VisibleForTesting`

atlasdb-coordination-impl/src/main/java/com/palantir/atlasdb/coordination/keyvalue/KeyValueServiceCoordinationStore.java
```diff
@@ -325,4 +334,15 @@ private Cell getCellForSequence(long sequenceNumber) {
         return Optional.ofNullable(kvs.get(AtlasDbConstants.COORDINATION_TABLE, ImmutableMap.of(cell, Long.MAX_VALUE))
                 .get(cell));
     }
+
+    private CheckAndSetResult<SequenceAndBound> getMostRecentValueAndBound(Optional<SequenceAndBound> oldValue,
```
jeremyk-91: nit: `getFailedResultWithMostRecentValueAndBound`? This also constructs the `CheckAndSetResult`

PR #3852, opened 2019-03-22
Diagnostic logging for PDS-79173

stefank-palantir:
Log the following:
* All client requests invoked on the lock server, including timings
* isStillLeadingInternal timings (both for the whole method, and the
quorum check)
* getLatestSequencePreparedOrAccepted timings (from the perspective of
the lock leader making the request)
* getLatestSequencePreparedOrAccepted timings (locally on the acceptor
node)

PDS-79173

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorImpl.java
@@ -17,9 +17,12 @@ package com.palantir.paxos;
 
 import java.io.IOException;
 import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Stopwatch;
 
 public class PaxosAcceptorImpl implements PaxosAcceptor {
     private static final Logger logger = LoggerFactory.getLogger(PaxosAcceptorImpl.class);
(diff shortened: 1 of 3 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorImpl.java
```diff
@@ -118,11 +121,20 @@ public BooleanPaxosResponse accept(long seq, PaxosProposal proposal) {
...
+        Thread currentThread = Thread.currentThread();
+        logger.debug("[{} - {}] Entering getLatestSequencePreparedOrAccepted", currentThread.getName(), currentThread.getId());
+
+        long result;
         if (state.isEmpty()) {
-            return greatestInLogAtStartup;
+            result = greatestInLogAtStartup;
         } else {
-            return Math.max(greatestInLogAtStartup, state.lastKey());
+            result = Math.max(greatestInLogAtStartup, state.lastKey());
```
jeremyk-91: I don't imagine this one will be slow at all (local ops), but to be fair it is still useful for evaluating hypotheses.

stefank-palantir: Yep agree. This is just a "make sure I'm not crazy" check really.

stefank-palantir: I think that if the remoting layer is the issue (and it probably is) we'll see that by noting that the start timestamp of the timing call added here is much later than the time the request arrives at the server (by taking packet capture).

PR #3857, opened 2019-03-25
[_transactions2] Part 29: The CellLoader Revolutions

jeremyk-91:
**This PR has a dependency on #3849 - please review that first.**

**Goals (and why)**:
- Implement a new cell-loader that is able to use the new Cassandra multiget-multislice endpoint
- Clean up batching of queries, and leverage batching when we have columns that don't have too many associated rows (because RPC overhead becomes unnecessary in these cases)

- Extract the part of CellLoader that divides up cells by their columns to a separate class `CellLoadingBatcher`
- Implement the CL2+ algorithm, that merges requests for columns with only a few rows.

Overall behaviour of the CellLoader - in particular, whether we accidentally miss any cells, should already be covered by existing Cassandra integration tests. I've added tests for the new batching logic.

- Are the default thresholds sensible? Part 30 is planned to be exposing these in runtime config, but I don't imagine that most deployments want to override this.
- Was anything lost in the refactor?

**Where should we start reviewing?**: Probably CellLoadingBatcherTest is a decent place to start.


jeremyk-91:
re-basing on develop

PR #3863, opened 2019-03-25
add debug/trace logging to aid C* pool exhaustion debugging

clockfort:
for PDS-85097
Specifically, this extra debugging is for the case of bursts of high connection usage that perform very little work / are low duration. These are more difficult to catch with other tooling, e.g. periodic jstacks would be unlikely to trigger during a connection burst, and would just show a pool that has grown to max size, but is mostly full of idle connections.

planning on putting this on an old customer branch for some debugging; if you want, develop can also have it, for future debugging needs

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -15,10 +15,14 @@
  */
 package com.palantir.atlasdb.keyvalue.cassandra;
 
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
 import java.lang.reflect.Field;
 import java.net.InetSocketAddress;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -265,6 +279,27 @@ public String toString() {
         return pool;
     }
 
+    private void logThreadStates() {
+        ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+        for (ThreadInfo info : threadBean.getThreadInfo(threadBean.getAllThreadIds())) {
+            // we're fairly good about annotating our C* pool thread names with the current action
+            if (log.isTraceEnabled()) {
+                log.trace("active thread",
+                        UnsafeArg.of("threadName", info.getThreadName()),
```
jeremyk-91: This is fine for now. I think internal log frameworks usually report this as safe (separately as part of the thread field)

clockfort: That's a bit bad, since the thread name here I think could be something like "connected to cassandra node {unsafe hostname or IP} doing getRows call on table {unsafe tablename}"

PR #3865, opened 2019-03-26
Do not run callback synchronously if initialize async is true

gmaretic:


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManagerTest.java
```diff
@@ -322,15 +322,14 @@ void stopBlocking() {
         public void init(TransactionManager transactionManager) {
             successfullyInvoked =
                     transactionManager.getKeyValueServiceStatus() == KeyValueServiceStatus.HEALTHY_ALL_OPERATIONS;
-            if (block) {
-                throw new RuntimeException();
+            while (block) {
+                try {
+                    Thread.sleep(50);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
```
jeremyk-91: This is a test so probably not an issue, but please re-interrupt (`Thread.currentThread().interrupt()`) or use Awaitility if it's available?

PR #3869, opened 2019-03-27
[_transactions2] Part 31?: DbKVS Support for KeyAlreadyExists/CheckAndSetExceptions

### LINKED ISSUE #3560: DbKvs check and set is inconsistent with other KVSes

jeremyk-91:
**Goals (and why)**:
- Fix #3560 
- This is also needed for eventual transactions2 rollout to DbKVS if that is something we want to do.
- Specifically, PUE exceptions didn't state which keys already exist, and CAS from nothing to something would fail if a value was already in the DB, but such values wouldn't be reported.

- Read from the DB the existing values when an operation fails.
- Remove the no-detail level of CheckAndSetCompatibility, as no key-value services now have that; but distinguish Cassandra (where CAS results are an atomic state) and the DbKVS semantics for the exceptions.

Previous tests in AbstractKVS for the most part only check that a KeyAlreadyExists or CheckAndSetException was thrown, not its details. I've added some tests that details are verified

- This introduces an additional read in the case of PUE failures (I wouldn't see CAS failures as an issue, as users of CAS already can face a second read). Is this an issue?
- Are there more efficient ways to perform the queries I need to do (determine what exists for PUE, determine the values for CAS)?
- Is it possible for a PUE or CAS operation in DbKVS to actually throw a KeyAlreadyExists or CheckAndSetException without details?

**Where should we start reviewing?**: DbKvs.java


atlasdb-api/src/test/java/com/palantir/atlasdb/keyvalue/api/CheckAndSetCompatibilityTest.java
```diff
@@ -33,26 +33,26 @@ public void minThrowsExceptionIfNoCompatibilitiesProvided() {
...
-        assertThat(getMinCompatibility(CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE,
+        assertThat(getMinCompatibility(CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE_ATOMIC,
                         CheckAndSetCompatibility.NOT_SUPPORTED,
-                        CheckAndSetCompatibility.SUPPORTED_NO_DETAIL_ON_FAILURE))
+                        CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE_NOT_ATOMIC))
                 .isEqualTo(CheckAndSetCompatibility.NOT_SUPPORTED);
     }
 
     @Test
     public void minReturnsSupportedNoDetailIfNotSupportedAbsent() {
```
gokcanozakdag: seems like name of the test needs to be changed also

gmaretic:
(assuminf the failing tests are fixed on develop)

gmaretic:
@jeremyk-91 Can you double-check that this still makes sense and we can then merge it?

PR #3875, opened 2019-03-28
[TLS] Timelock Stability: Autobatch block on becoming leader

gmaretic:
**Goals (and why)**:
We really do not need to send so many block on becoming leader requests. This proved to be more efficient than using coalescing supplier

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/BatchingLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/BatchingLeaderElectionService.java
@@ -16,11 +16,13 @@
 
 package com.palantir.leader;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
 import com.google.common.net.HostAndPort;
+import com.palantir.atlasdb.autobatch.BatchElement;
 import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
 
 public class BatchingLeaderElectionService implements LeaderElectionService {
@@ -29,15 +31,7 @@ public class BatchingLeaderElectionService implements LeaderElectionService {
 
     public BatchingLeaderElectionService(LeaderElectionService delegate) {
         this.delegate = delegate;
-        this.batcher = DisruptorAutobatcher.create(batch -> {
-            try {
-                LeaderElectionService.LeadershipToken leadershipToken = delegate.blockOnBecomingLeader();
-                batch.forEach(e -> e.result().set(leadershipToken));
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                throw new RuntimeException(e);
-            }
-        });
+        this.batcher = DisruptorAutobatcher.create(this::processBatch);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


leader-election-impl/src/main/java/com/palantir/leader/BatchingLeaderElectionService.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+
+import com.google.common.net.HostAndPort;
+import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
+
+public class BatchingLeaderElectionService implements LeaderElectionService {
+    private final LeaderElectionService delegate;
+    private final DisruptorAutobatcher<Void, LeadershipToken> batcher;
```
gokcanozakdag: we need to close autobatcher to avoid thread leaks

leader-election-impl/src/main/java/com/palantir/leader/BatchingLeaderElectionService.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public BatchingLeaderElectionService(LeaderElectionService delegate) {
+        this.delegate = delegate;
+        this.batcher = DisruptorAutobatcher.create(batch -> {
+            try {
+                LeaderElectionService.LeadershipToken leadershipToken = delegate.blockOnBecomingLeader();
+                batch.forEach(e -> e.result().set(leadershipToken));
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
```
gokcanozakdag: this will change the behaviour a bit as we are now throwing ExecutionExceptions rather than their original cause. May be a problem depending on the usage of the exception at higher levels.

gokcanozakdag: this comment should have been few lines below :D

gmaretic: Maybe try to unwrap it first?

jeremyk-91:
Some thoughts around interruption:

- when a thread that calls `blockOnBecomingLeader()` is interrupted, its `future.get()` call will throw an `InterruptedException`. This is fine.
- when _the autobatcher thread_ is interrupted, recall that the batch function looked like this:

```
        {
            try {
                LeaderElectionService.LeadershipToken leadershipToken = delegate.blockOnBecomingLeader();
                batch.forEach(e -> e.result().set(leadershipToken));
            }
        }
```
So the batcher thread is reinterrupted, but other threads that are waiting on the futures will be completed exceptionally with a `RuntimeException` with cause `InterruptedException`. This is then rethrown from `blockOnBecomingLeader` meaning that the interruption isn't propagated.

It's unclear what we want to do with interruption here, though one possibility is to have `InterruptedException` thrown on threads that called the autobatcher. We'd want to complete the batch futures exceptionally with a straight `InterruptedException` rather than wrapping that in a `RuntimeException` if that's the case.

PR #3879, opened 2019-03-29
[_transactions2] Part 35: Resubmit Duplicates in Write Batching Transaction Service

jeremyk-91:
**Goals (and why)**:
- We'd previously throw a `SafeIllegalStateException` when making more than one request, which is technically legal but confusing for clients and bad in terms of liveness.

- Enqueue these duplicates in the next batch instead - they might take longer, but will definitely get a sane response.

- Existing tests should verify that we do return legitimate results. The last test in WBTST was changed to verify that we return true for the first element and false for the later ones.

- Re-execution is non-fair. Is that a serious problem?

**Where should we start reviewing?**: WriteBatchingTransactionService


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/WriteBatchingTransactionService.java
```diff
@@ -51,16 +50,16 @@
  */
 public final class WriteBatchingTransactionService implements TransactionService {
     private final EncodingTransactionService delegate;
-    private final DisruptorAutobatcher<TimestampPair, Void> autobatcher;
+    private final DisruptorAutobatcher<TimestampPair, Boolean> autobatcher;
```
j-baker: I think we can do a little better here...

j-baker: gimme 5...

PR #3880, opened 2019-03-29
[TLS] Timelock Stability: Use a single executor for lock reaper and timeout

gmaretic:
**Goals (and why)**:
Reduce number of executors. This is a hack that we used for benchmarking. Exact approach TBD


felixdesouza:
Discussed offline, but a simpler fix would be to just share a cached thread pool between all the clients. Currently, they *all* have a cached thread pool, so there is absolutely no reuse and I don't think the tasks themselves take that long.

This should deal with the provisioning aspect, granted it *is* unbounded, but it has been unbounded in a worse way before, this is just a marginal step in the right direction until it's actually a problem.

felixdesouza:
if possible, it might make sense to add a bit of random jitter so we don't spike with thread usage if they're all trying to do the same thing at the same time => no reuse

I'm assuming it's periodic and not in response to actual requests, happy to correct that assumption.

gmaretic:
Not doing this for now

PR #3882, opened 2019-03-29
[_transactions2] Part 36: Decompose Cassandra putUnlessExists() Return Buffers

jeremyk-91:
**Goals (and why)**:
- Fix an issue where the KeyAlreadyExistsExceptions thrown by Cassandra KVS would be incorrect, as they would have columns expressed in the form `thisIsTheColumnKey || timestamp`. This was causing transactions2 to fail with `NullPointerExceptions` because it looks like it receives timestamps it never asked for, and then attempts to grab the futures for these to mark them as failed, and then explodes.

- Fix PUE to extract only the column name and not the timestamp.

Added a test that the cell included was the one we expected, plus a fuzz test that a large number of highly conflicting writes resolves itself quickly.

- Does this actually fix the issue we saw?

**Where should we start reviewing?**: CassandraKeyValueServiceImpl

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
@@ -15,18 +15,35 @@
  */
 package com.palantir.atlasdb.keyvalue.cassandra;
 
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
 
 import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Rule;
+import org.junit.Test;
 import org.junit.rules.TestRule;
 
 import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.Futures;
 import com.palantir.atlasdb.cassandra.CassandraMutationTimestampProviders;
 import com.palantir.atlasdb.containers.CassandraResource;
+import com.palantir.atlasdb.keyvalue.api.Cell;
+import com.palantir.atlasdb.keyvalue.api.KeyAlreadyExistsException;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.transaction.encoding.TicketsEncodingStrategy;
 import com.palantir.atlasdb.transaction.impl.AbstractTransactionTest;
+import com.palantir.atlasdb.transaction.impl.TransactionConstants;
+import com.palantir.atlasdb.transaction.impl.TransactionTables;
+import com.palantir.atlasdb.transaction.service.SimpleTransactionService;
+import com.palantir.atlasdb.transaction.service.TransactionService;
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
```diff
@@ -60,6 +73,28 @@ protected KeyValueService getKeyValueService() {
...
 
+    @Test
+    public void canUntangleHighlyConflictingPutUnlessExists() {
+        TransactionTables.createTables(keyValueService);
+        TransactionTables.truncateTables(keyValueService);
+        TransactionService transactionService = WriteBatchingTransactionService.create(
+                SimpleTransactionService.createV2(keyValueService));
+        ExecutorService executorService = Executors.newCachedThreadPool();
+        List<Future<?>> futures = LongStream.range(0, 100)
+                .mapToObj(unused -> executorService.submit(() -> {
```
gmaretic: unused 🗡

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
```diff
@@ -60,6 +73,28 @@ protected KeyValueService getKeyValueService() {
...
+                    try {
+                        transactionService.putUnlessExists(unused, unused + 1);
+                        transactionService.putUnlessExists(unused + 1, unused + 3);
+                        transactionService.putUnlessExists(unused - 1, unused + 5);
+                    } catch (KeyAlreadyExistsException ex) {
+                        // Some are expected
```
gmaretic: We could actually collect and verify that we get the expected number of conflicts for each ts pair

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
jeremyk-91: I think the 😅 emoji is appropriate here - the original test had random numbers, but I found this pretty reliably reproduced the issue.

jeremyk-91: also changing the numbers somewhat, as `timestamp - 1` means we might end up with negative timestamps.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTransactionIntegrationTest.java
jeremyk-91: Turns out this is trickier than we expected because of the way batches throw out exceptions (the keys reflect everything in the batch). As discussed offline, we'll make assertions on the number of exceptions, but not at the individual ts-pair level.

jeremyk-91: Actually we can't even guarantee that 😞 because of retry logic

PR #3884, opened 2019-04-01
[TLS] Timelock Stability: Do not retry on the HTTP client level when we just retry anyway

gmaretic:
**Goals (and why)**:
The HTTP client retryer will retry with potentially up to a 500 ms backoff when we will just retry on failure anyway.

Checked that none of these are used in large internal product


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/InterruptHonoringRetryer.java
```diff
@@ -1,45 +0,0 @@
...
- */
-package com.palantir.atlasdb.http;
-
-import java.io.InterruptedIOException;
-
-import feign.RetryableException;
-import feign.Retryer;
-
-/**
- * Simple wrapper around the default Feign retryer that propagates {@link InterruptedIOException}s
```
felixdesouza: do we necessarily want to get rid of the interrupt honouring thing? assuming this was to cover missing okhttp capabilities - it's probably worth checking whether or not it has been fixed upstream, otherwise we *might* want to bring it back.

In my haste to get things working quickly, I didn't wrap things

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/clock/ClockSkewMonitor.java
```diff
@@ -53,7 +54,7 @@ public static ClockSkewMonitor create(MetricsManager metricsManager,
         Map<String, ClockService> clocksByServer = Maps.toMap(
                 remoteServers,
                 (remoteServer) -> AtlasDbHttpClients.createProxy(metricsManager.getRegistry(),
-                        optionalSecurity, remoteServer, ClockService.class));
```
felixdesouza: @gokcanozakdag do we still need this `ClockSkewMonitor`? If not, we should remove most of the skew/leadership lease related code since it's not going to be used + makes refactors easier (don't have to think about whether or not it's safe to remove code etc.)

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -43,17 +43,11 @@ private AtlasDbHttpClients() {
...
      */
-    public static <T> T createProxy(
-            MetricRegistry metricRegistry,
-            Optional<TrustContext> trustContext,
-            String uri,
-            Class<T> type) {
-        return createProxy(metricRegistry, trustContext, uri, type, UserAgents.DEFAULT_USER_AGENT, false);
-    }
-
     public static <T> T createProxy(
```
felixdesouza: propagate the name change here, I believe `AtlasDbFeignTargetFactory` isn't used directly in most places, and it comes through here.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbFeignTargetFactory.java
```diff
@@ -71,7 +72,7 @@ private AtlasDbFeignTargetFactory() {
                 .encoder(encoder)
                 .decoder(decoder)
                 .errorDecoder(errorDecoder)
-                .retryer(new InterruptHonoringRetryer())
+                .retryer(Retryer.NEVER_RETRY)
```
jeremyk-91: Note that there's a semantics change here, as we can now throw Feign exceptions as opposed to general `RuntimeException`. I don't think this is a problem though?

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
jeremyk-91: Seems fair. I'd normally attach the usual warnings of an API break, but _this_ method also seems hardly used outside of Atlas

felixdesouza: I think we should start employing `@RestrictedApi` so we can at least keep track of things that are atlas only.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/clock/ClockSkewMonitor.java
gokcanozakdag: I think it is time to remove `ClockSkewMonitor` 🔥 🔥

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
gmaretic: Updated the method name. As for annotations, shouldn't we just use `@Beta` then?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbFeignTargetFactory.java
gmaretic: That's a good point, but should be fine

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/clock/ClockSkewMonitor.java
gmaretic: Killed with fire

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbFeignTargetFactory.java
felixdesouza: re @jeremyk-91 point, we were already doing that once we'd run out of attempts, or are you specifically talking about the interrupted exception

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
felixdesouza: Nah, `@Beta` implies you can use it and it's subject to change, `@RestrictedApi` can fail error prone checks if it's used outside of where we expect it to be used.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TimeLockMigrationEteTest.java
```diff
@@ -181,11 +182,13 @@ private static void waitForTransactionManagerCreationError() {
 
     private static <T> T createEteClientFor(Class<T> clazz) {
         String uri = String.format("http://%s:%s", ETE_CONTAINER, ETE_PORT);
-        return AtlasDbHttpClients.createProxy(new MetricRegistry(), Optional.empty(), uri, clazz);
+        return AtlasDbHttpClients.createProxyWithoutRetrying(new MetricRegistry(), Optional.empty(), uri, clazz,
```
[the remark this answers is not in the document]
felixdesouza: ping on this

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/IsolatedPaxosTimeLockServerIntegrationTest.java
```diff
@@ -84,15 +85,17 @@ public void canParticipateInPaxosAsLearnerWithoutQuorum() {
     }
 
     private static <T> T createProxyForInternalNamespacedTestService(Class<T> clazz) {
-        return AtlasDbHttpClients.createProxy(
+        return AtlasDbHttpClients.createProxyWithoutRetrying(
```
felixdesouza: same thing re tests

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -125,11 +126,13 @@
 
     @BeforeClass
     public static void waitForClusterToStabilize() {
-        PingableLeader leader = AtlasDbHttpClients.createProxy(
+        PingableLeader leader = AtlasDbHttpClients.createProxyWithoutRetrying(
```
felixdesouza: and here too

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/util/TestProxies.java
```diff
@@ -58,7 +58,7 @@ public TestProxies(String baseUri, List<TestableTimelockServer> servers) {
 
     public <T> T singleNode(Class<T> serviceInterface, String uri) {
         List<Object> key = ImmutableList.of(serviceInterface, uri, "single");
-        return (T) proxies.computeIfAbsent(key, ignored -> AtlasDbHttpClients.createProxy(
+        return (T) proxies.computeIfAbsent(key, ignored -> AtlasDbHttpClients.createProxyWithoutRetrying(
```
felixdesouza: and here

PR #3885, opened 2019-04-01
remove ReadOnlyTransactionManager

gokcanozakdag:
**Goals (and why)**:
Remove unused `ReadOnlyTransactionManager` as a step towards simpler code.

This was not used in Atlas codebase, I also checked other products, especially the large internal product. Any other possible usage that I am not aware of?


jeremyk-91:
There's an argument to be made for moving this into large internal product, though it is unfortunately used (see DACFactory)

PR #3888, opened 2019-04-02
lock immutable timestamp on reads

gokcanozakdag:
**Goals (and why)**:
We believe that it is now safe to move our clients to use thorough sweep. This pr introduces a runtime config flag, that when set to true allows running read only transactions on thorough sweep tables, but also locks immutable timestamp for conservative ones.

By locking immutable ts on conservative tables, we will be able to:
- measure the perf impact
- migrate to thorough sweep with a rolling upgrade.

We go with the option of adding the config flag to client configuration for the sake of easily rolling this out in few canaries and measuring perf. We will be deprecating the config option, and setting it to true by default when we actually decide to migrate tables to thorough sweep.

Added very basic tests checking if we actually make `startIdentifiedAtlasDbTransaction` calls to `TimelockService` when the flag is set to true for read-only transactions.

`ReadOnlyTransactionManager` is not covered here. My investigations show that it is not used outside of large-internal-product, and we plan to roll this out for the rest of the platform first. I think it would make more sense to remove `ReadOnlyTransactionManager` altogether (even though that will require some code change in large-internal-product)

Is it covering all other codepaths for running read-only transactions?


sandorw:
Generally agree about not modifying `ReadOnlyTransactionManager`, though we'll eventually have to figure out a path forward there.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerTest.java
```diff
@@ -175,4 +177,54 @@ public void registersMetrics() throws InterruptedException {
         assertThat(registry.getTimers().get(SETUP_TASK_METRIC_NAME).getCount()).isGreaterThanOrEqualTo(1);
         assertThat(registry.getTimers().get(FINISH_TASK_METRIC_NAME).getCount()).isGreaterThanOrEqualTo(1);
     }
+
```
sandorw: Ideally we'd add a test to confirm lock checks on reads

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManagerTest.java
```diff
@@ -175,4 +177,54 @@ public void registersMetrics() throws InterruptedException {
...
+    @Test
+    public void callsStartTransactionForReadOnlyTransactionsIfFlagIsSet() throws InterruptedException {
+        TimelockService timelockService =
+                spy(new LegacyTimelockService(timestampService, closeableLockService, LockClient.of("lock")));
+        when(closeableLockService.lock(any(), any())).thenReturn(new LockRefreshToken(BigInteger.ONE, Long.MAX_VALUE));
+        SnapshotTransactionManager transactionManager = createSnapshotTransactionManager(timelockService, true);
+
+        transactionManager.runTaskReadOnly(tx -> "ignored");
```
sandorw: Should chat offline, but I believe there are cases where services expect no-op transactions to not have a cost, since some actions are lazy. Should be aware of whether or not we're changing any of those guarantees.

gokcanozakdag: you are right, this change _will_ have an effect there; but I think it makes sense to address that in a separate pr after rolling this one out for few specific services.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -749,11 +749,15 @@ private void validatePreCommitRequirementsOnReadIfNecessary(TableReference table
...
+        return validateLocksOnReads && requiresImmutableTimestampLocking(tableRef);
     }
 
     private boolean isValidationNecessaryOnCommit(TableReference tableRef) {
-        return !validateLocksOnReads && isThoroughlySwept(tableRef);
+        return !validateLocksOnReads && requiresImmutableTimestampLocking(tableRef);
+    }
+
+    private boolean requiresImmutableTimestampLocking(TableReference tableRef) {
+        return isThoroughlySwept(tableRef) || transactionConfig.get().lockImmutableTsOnReadOnlyTransactions();
```
sandorw: This means that the value in the config can change between when we start the transaction and any particular read (or between reads). Since this isn't a correctness issue, I don't think we should worry about it.

Note that this might have caused failures or precondition checks, depending on how our code was written. If the config changes to cause us to check immutable timestamp lock validity but it turns out we never took one (since the flag was false when we opened the transaction), then we might have failed if we had assumed that validation necessary meant that the immutable timestamp lock must exist.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -138,7 +138,7 @@
 
 @SuppressWarnings("checkstyle:all")
 public class SnapshotTransactionTest extends AtlasDbTestCase {
-    private static final TransactionConfig TRANSACTION_CONFIG = ImmutableTransactionConfig.builder().build();
+    private static TransactionConfig transactionConfig = ImmutableTransactionConfig.builder().build();
```
sandorw: Let's not make this static if it's mutable. Just set it to the correct value in each test or in the `@before` method and then update the value for the specific test that needs something different.

As is, the order that tests run in could affect results.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1208,6 +1213,35 @@ public void testIgnoresOrphanedSweepSentinel() {
...
+        SnapshotTransaction transaction = getSnapshotTransactionWith(
+                timelockService,
+                () -> transactionTs,
+                res,
+                PreCommitConditions.NO_OP,
+                true);
+
+        transaction.get(TABLE_SWEPT_CONSERVATIVE, ImmutableSet.of(TEST_CELL));
+        transaction.commit();
+        timelockService.unlock(ImmutableSet.of(res.getLock()));
```
sandorw: `commit` currently checks these locks, so I'd expect two refreshes. Can we confirm the call explicitly after the `get` and `commit` calls?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gokcanozakdag: yeah, I was also concerned about this value changing in the middle of a transaction; but our check logic throws only if immutableTsLock exists and not valid anymore. If it doesn't exist in the first place, than it doesn't complain.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
gokcanozakdag: in this case we will check the lock only once after the get call - but not after the commit call. We either check immutableTsLock after each read (and don't check at commit) or only check at commit - depending on the value of `validateLocksOnReads`
I will move the verify after get() call.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -242,10 +242,18 @@ protected SnapshotTransaction createTransaction(
...
     }
-
     @Override
     public <T, C extends PreCommitCondition, E extends Exception> T runTaskWithConditionReadOnly(
             C condition, ConditionAwareTransactionTask<T, C, E> task) throws E {
+        if (transactionConfig.get().lockImmutableTsOnReadOnlyTransactions()) {
+            return runTaskWithConditionThrowOnConflict(condition, task);
+        } else {
+            return runTaskWithConditionReadOnlyInternal(condition, task);
+        }
```
jeremyk-91: This makes sense.

nit: We don't need `else`; I think I would tend to write this with `? :` notation as well. Any way is fine though!

PR #3890, opened 2019-04-02
Require Rescue 4.4.0

tpetracca:
This is a requirement for rolling _transactions2 out by default. Unclear if we actually want to merge it yet if we're not quite at the "turn on by default" phase yet. As of today (April 2nd) about half of the fleet violates this constraint.


tpetracca:
per offline discussion this is now in a much better state across the fleet and I think we should push this forward.

PR #3892, opened 2019-04-03
Close Cassandra client input and output transports

schlosna:
**Goals (and why)**:

When removing a Cassandra client from the pool, close both the input and
output transports to ensure any underlying sockets are also closed.

`CassandraClient` interface now extends `Closeable` and implementations handle closing their input and output transports, and the Cassandra pool now just closes evicted clients.

none, noticed unclosed sockets & clients while looking through a test heap dump

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
@@ -203,10 +203,12 @@ public class CassandraClientFactory extends BasePooledObjectFactory<CassandraCli
 
     @Override
     public void destroyObject(PooledObject<CassandraClient> client) {
-        client.getObject().getOutputProtocol().getTransport().close();
-        log.debug("Closed transport for client {} of host {}",
-                UnsafeArg.of("client", client),
-                SafeArg.of("cassandraClient", CassandraLogHelper.host(addr)));
+        try (CassandraClient cassandraClient = client.getObject()) {
+            cassandraClient.close();
+            log.debug("Closed transport for client {} of host {}",
+                    UnsafeArg.of("client", client),
+                    SafeArg.of("cassandraClient", CassandraLogHelper.host(addr)));
+        }
     }
 
     static class ClientCreationFailedException extends AtlasDbDependencyException {
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
@@ -203,12 +203,10 @@ public class CassandraClientFactory extends BasePooledObjectFactory<CassandraCli
 
     @Override
     public void destroyObject(PooledObject<CassandraClient> client) {
-        try (CassandraClient cassandraClient = client.getObject()) {
-            cassandraClient.close();
-            log.debug("Closed transport for client {} of host {}",
-                    UnsafeArg.of("client", client),
-                    SafeArg.of("cassandraClient", CassandraLogHelper.host(addr)));
-        }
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
```diff
@@ -203,10 +203,12 @@ public boolean validateObject(PooledObject<CassandraClient> client) {
 
     @Override
     public void destroyObject(PooledObject<CassandraClient> client) {
-        client.getObject().getOutputProtocol().getTransport().close();
-        log.debug("Closed transport for client {} of host {}",
-                UnsafeArg.of("client", client),
-                SafeArg.of("cassandraClient", CassandraLogHelper.host(addr)));
+        try (CassandraClient cassandraClient = client.getObject()) {
+            cassandraClient.close();
```
diogoholanda: Isn't `close()` getting called twice here, once explicitly and once due to the `try` block? Should we remove the `try` block and simply call `client.getObject().close()` directly?

schlosna: fixing, try-with-resources was leftover from before I pushed it into the `CassandraClient `

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientImpl.java
```diff
@@ -277,6 +278,17 @@ public CqlResult execute_cql3_query(CqlQuery cqlQuery,
...
     }
 
+    @Override
+    public void close() {
+        TProtocol inputProtocol = getInputProtocol();
+        TProtocol outputProtocol = getOutputProtocol();
+        try (TTransport inputTransport = inputProtocol.getTransport();
+                TTransport outputTransport = outputProtocol.getTransport()) {
+            inputProtocol.reset();
+            outputProtocol.reset();
```
jeremyk-91: Verified `reset()` shouldn't throw, so it should be safe. Theoretically, if the first `reset()` call throws the second is never invoked and we might want to avoid that, but this is safe here.

schlosna: yeah, the `reset()` isn't actually necessary, mainly just wanted something for inside the try-with-resources, though could just put a `log.trace("Closing CassandraClient");` in the block.

Also debated explicitly clearing the `TFramedTransport` which holds a potentially large (~16MB) `readBuffer`. Don't have strong feelings either way, and this should be unnecessary as these clients with their referenced protocol, sockets, and buffers should be eligible for GC shortly after this point.

```
    @Override
    public void close() {
        TProtocol inputProtocol = getInputProtocol();
        TProtocol outputProtocol = getOutputProtocol();
        try (TTransport inputTransport = inputProtocol.getTransport();
                TTransport outputTransport = outputProtocol.getTransport()) {
            log.trace("Closing CassandraClient");
            tryClear(inputTransport);
            tryClear(outputTransport);
        }
    }

    private static void tryClear(TTransport transport) {
        if (transport instanceof TFramedTransport) {
            ((TFramedTransport) transport).clear();
        }
    }
```

We currently clear that buffer via reflection before returning valid clients to the pool in https://github.com/palantir/atlasdb/blob/94e7c6087896fb071bccd95b5bdaa320908e1403/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java#L173-L194 ) and note that https://github.com/palantir/atlasdb/pull/2971 specifically sets an empty array in the `TMemoryInputTransport`, but that would not be an issue here on eviction.

jeremyk-91: Yep, makes sense - thanks for the additional context!

PR #3894, opened 2019-04-03
Bump min version of timelock

gregakinman:
0.58.0's docker image is busted and I'd like to just in my testing say "use the min version specified by atlasdb"

PR #3898, opened 2019-04-04
Bump min version of TimeLock in product dependencies

jeremyk-91:
To quote @gregakinman on #3894 

> 0.58.0's docker image is busted and I'd like to just in my testing say "use the min version specified by atlasdb"

I think this is fine, though we should probably wait for 0.59.0 to be declared stable before doing this, otherwise it might unnecessarily block upgrades


gregakinman:
makes sense, thanks @jeremyk-91!

pkoenig10:
0.60.0 has been declared stable. Any other blockers to merging this?

gregakinman:
I see there's a conflict but I don't have push perms to this repo so somebody on the team will need to fix it or give me perms

gokcanozakdag:
I'll fix it now

PR #3903, opened 2019-04-05
Bump timelock-server product dependency minimum version

pkoenig10:
**Goals (and why)**:
Bump the timelock-server product dependency minimum version. This is because timelock-server 0.58.0 was a bad release and using it as the minimum version prevent projects from using automatically using dependency recommendations as integration test versions. See https://github.com/palantir/sls-packaging/pull/484 for more details.


felixdesouza:
Thanks, an open PR exists for this already, waiting on STABLE tag for 0.59.0, see #3898.

PR #3911, opened 2019-04-08
Single source of timestamp

felixdesouza:
**Goals (and why)**:
Due to #3524 which memoizes the `TimestampService` supplier, we can get timestamps going back in time. This is due to the fact that when leadership changes (without the service going down), the internal state is never reset, and the invariants regarding the upper bounds are effectively violated. Instead of restarting handing out timestamps from the agreed upon upper bound, we just resume from where we left off when we were last leader, which could mean we don't ever check the bound store for quite some time.

See internal ticket PDS-87236.

If you are in embedded mode and you've used `TransactionManagers` on 0.106.0 and 0.107.0 and there were leadership changes without restarts with a node becoming a leader more than once, it is likely there is corruption. You're likely safe past 0.108.0 as a result of `TimestampCorroboratingTimelockService`, although that assumes you've seen fresher timestamps as a result of servicing `getFreshTimestamp` calls. 

* Stop memoizing the supplier.
* Unify interface between `TimestampService` and `TimestampManagementService`.
* Wire all the way through.

* This case is somewhat difficult to test for (in an automated fashion), but roughly speaking, the test looks like this:

1. Spin up an embedded service e.g. `AtlasDbEteServer` in docker.
1. Figure out who the leader is and record the timestamp
1. Force a leader election
  a. With docker you can do:
  b. `docker network disconnect <network> <leader-container>`
  c. this stops node from communicating outward and also receiving requests => forces a leader election but retains all in memory state.
1. Figure out who the new leader is and record that timestamp
should start from the next multiple of `1_000_000`.
1. Kill that node (to make it easier to force the leader in 2. to be reelected again)
1. Disconnect and reconnect the remaining node until leader in 2. is reelected leader.
1. Calling `getFreshTimestamp` on the original leader after it has gained leadership for the second time will then reuse the bound store from a previous leadership term and then we go back in "time".

As a follow up PR, we probably want to make it clearer that it doesn't behave like a normal `Supplier` and that it's never okay to memoize since it affects correctness. It also makes a bit difficult to test without a full on ETE. I envisage a class along the lines of `Resettable<T>` which upon leadership election, will reset the instance returned. NB: This is just a `Supplier` but in a way that can't be screwed up I believe. Better still, have `AtlasDbFactory` return `Resettable<ManagedTimestampService>` and `AwaitingLeadershipProxy` take in `Resettable<ManagedTimestampService>` instead of `Supplier<...>`.

`TransactionManagers`, `AtlasDbFactory`, `ServiceDiscoveringAtlasSupplier`

ASAP


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplierTest.java
```diff
@@ -50,23 +51,42 @@
...
-        ServiceDiscoveringAtlasSupplier atlasSupplier = new ServiceDiscoveringAtlasSupplier(
-                metrics, kvsConfig, leaderConfig);
+        ServiceDiscoveringAtlasSupplier atlasSupplier = new ServiceDiscoveringAtlasSupplier(metrics,
+                kvsConfig,
+                Optional::empty,
+                leaderConfig,
+                Optional.empty(),
+                Optional.empty(),
+                AtlasDbConstants.DEFAULT_INITIALIZE_ASYNC,
+                AtlasDbFactory.THROWING_FRESH_TIMESTAMP_SOURCE);
```
jeremyk-91: In general for these creators it seems

```java
new ServiceDiscoveringAtlasSupplier(
                metrics,
                kvsConfig,
                leaderConfig,
                Optional.empty(),
                Optional.empty())
```

will do, or possibly extracting a static method in here that just takes the 3 original params

docs/source/release_notes/release-notes.rst
```diff
@@ -119,6 +119,12 @@ develop
          - Cassandra client input and output transports are now properly closed.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3892>`__)
 
+    *    - |fixed| |devbreak|
+         - Stop memoizing the `Supplier` of `TimestampService`, as we **must** get a fresh instance on each `Supplier.get()` call to ensure correctness after leadership elections. Without it, there is a possibility of data corruption.
+           Dev break to force `AtlasDbFactory` and `ServiceDiscoveringAtlasSupplier` to return unified interface of `TimestampService` and `TimestampManagementService`.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3911>`__)
```
jeremyk-91: :+1: Looks good. Couple of suggestions here:

1. We should probably state that there is only a risk of corruption if you're using services with a leader block in a multinode configuration. Services using External TimeLock, Embedded or Leader with 1 Node won't be affected. 
2. You probably intend ` ``X`` ` for monospacing, rather than `` `X` `` which usually is used for references to other parts of the document.
3. Might be worth including the name of the unified interface (`ManagedTimestampService`).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
```diff
@@ -85,23 +74,6 @@ public ServiceDiscoveringAtlasSupplier(
...
-        this(metricsManager,
-                config,
-                runtimeConfig,
-                leaderConfig,
-                namespace,
-                Optional.empty(),
-                initializeAsync,
-                AtlasDbFactory.THROWING_FRESH_TIMESTAMP_SOURCE);
-    }
-
```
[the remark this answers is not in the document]
felixdesouza: the todo about duplication?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -864,17 +855,22 @@ private static LockAndTimestampServices createRawLeaderServices(
...
+                managedTimestampProxy,
                 TimestampService.class);
+
         TimestampManagementService localManagement = ServiceCreator.createInstrumentedService(
                 metricsManager.getRegistry(),
-                AwaitingLeadershipProxy.newProxyInstance(TimestampManagementService.class,
-                        timeManagement,
-                        leader),
+                managedTimestampProxy,
                 TimestampManagementService.class);
```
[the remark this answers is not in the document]
felixdesouza: I couldn't help myself 🙈

docs/source/release_notes/release-notes.rst
felixdesouza: good catch re monospacing, will make these changes

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplierTest.java
felixdesouza: I think I actually wanted to get rid of that constructor, I can't remember why I didn't though. I'll make a helper method inside the test however.

PR #3917, opened 2019-04-10
Timelock Auth Part 1: Introducing Authorizer

gokcanozakdag:
**Goals (and why)**:
This pr introduces authn/authz api - and implements a simple authorizer.

Simple interfaces for `Authenticator` and `Authorizer` are added.

Authorizer is implemented such that:
- Admin users have access to all namespaces
- A many-to-many relationship can be established between authorized users and namespaces. That is done by using a map of `User` to `NamespaceMatcher`, where a `NamespaceMatcher` can match multiple namespaces.

Also an `AuthRequirer` is introduced to enable switching between multiple behaviours of:
- Not requiring auth at all
- Requiring auth only for namespaces when privileges are defined for that namespace
- Requiring auth for all namespaces

New unit tests are added for `Authorizer` - checking if users are authorized as expected.

Currently this api lives in an isolated state, it is not used anywhere else. I wanted to keep pr's small, and thought this an acceptable trade-off.

`Authenticator` and `Authorizer` interfaces.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/User.java
```diff
@@ -0,0 +1,41 @@
...
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface User {
+    User ANONYMOUS = User.of("anonymous");
+
+    String getUserName();
+    boolean isAdmin();
+
+    static User of(String userName) {
```
gmaretic: Something more descriptive like `regular`, or `user`?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/User.java
```diff
@@ -0,0 +1,41 @@
...
+        return ImmutableUser.builder()
+                .userName(userName)
+                .isAdmin(false)
+                .build();
+    }
+
+    static User createAdmin(String userName) {
```
gmaretic: I'd just name it `admin`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/User.java
```diff
@@ -0,0 +1,41 @@
...
+package com.palantir.atlasdb.timelock.auth.api;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface User {
+    User ANONYMOUS = User.of("anonymous");
+
+    String getUserName();
+    boolean isAdmin();
```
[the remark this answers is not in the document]
gmaretic: Admin is a role, and we have only two roles, so we can do it with a boolean. Otherwise this would return an enum type describing the role.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -0,0 +1,52 @@
...
+        this.authRequirer = authRequirer;
+    }
+
+    public static Authorizer of(Map<User, NamespaceMatcher> privileges, AuthRequirer authRequirer) {
+        return new SimpleAuthorizer(new HashMap<>(privileges), authRequirer);
+    }
+
+    @Override
+    public boolean isAuthorized(User user, Namespace namespace) {
+        if (!isAuthorizationRequired(namespace) || user.isAdmin()) {
```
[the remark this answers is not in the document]
gmaretic:
```
return user.isAdmin() || privileges.getOrDefault(user, NamespaceMatcher.ALWAYS_DENY).matches(namespace)
                || !isAuthorizationRequired(namespace);
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -0,0 +1,52 @@
...
+import com.palantir.atlasdb.timelock.auth.api.Authorizer;
+import com.palantir.atlasdb.timelock.auth.api.NamespaceMatcher;
+import com.palantir.atlasdb.timelock.auth.api.User;
+
+public class SimpleAuthorizer implements Authorizer {
+    private final Map<User, NamespaceMatcher> privileges;
+    private final AuthRequirer authRequirer;
+
+    SimpleAuthorizer(Map<User, NamespaceMatcher> privileges, AuthRequirer authRequirer) {
+        this.privileges = privileges;
```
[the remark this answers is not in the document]
gokcanozakdag: we won't have concrete `Namespace`s for some of the clients - that is we will be using regexes. Not sure how would you implement it in other way around for these clients.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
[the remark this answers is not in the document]
gokcanozakdag: I think having a explicit definition of admin user is clearer than using `.*` matcher with a regular user (which would be equal to implicitly stating admin user. Also this would make it easier to admin privileges in future I believe.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authenticator.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.auth.api;
+
+import com.palantir.tokens.auth.BearerToken;
+
+public interface Authenticator {
+    User authenticate(BearerToken token);
```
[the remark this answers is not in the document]
gokcanozakdag: I see your point, will think about the alternatives - removing Authenticator from this pr just to unblock (Authenticator is kind of irrelevant right now anyways)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Client.java
```diff
@@ -0,0 +1,41 @@
...
+package com.palantir.atlasdb.timelock.auth.api;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface Client {
+    Client ANONYMOUS = Client.regular("anonymous");
+
+    String getUserName();
+    boolean isAdmin();
```
jeremyk-91: I think we should probably standardise the naming here and in the factories below, the `userName` field can probably just be `name` (or `clientName`, if you prefer)

Also, nit: we typically name the components in immutable classes as just the components (i.e. `userName` and `admin`); these getters should be autogenerated if I remember correctly.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authorizer.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.auth.api;
+
+import com.palantir.atlasdb.keyvalue.api.Namespace;
```
jeremyk-91: Hmm, I think we're not using the right construct here. This `Namespace` object is at a layer below the atlas keyspace/db-name/sid, which products can use to create multiple table-references with the same table name (which can be useful if you're using libraries that want to create tables and you want to avoid clashes), so it doesn't seem right.

There can exist namespace objects that aren't valid timelock namespaces, for example `@!@#(*$#@%&*` is a valid `UNCHECKED_NAME` `Namespace` object.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/NamespaceMatcher.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.auth.api;
+
+import com.palantir.atlasdb.keyvalue.api.Namespace;
+
+public interface NamespaceMatcher {
+    NamespaceMatcher NEVER_MATCH = ignored -> false;
+    boolean matches(Namespace namespace);
```
jeremyk-91: As mentioned earlier, I really don't think we should use the `keyvalue.api.Namespace` here. A separate string-wrapper class makes sense though.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedAuthRequirer.java
```diff
@@ -0,0 +1,38 @@
...
+public class PrivilegeBasedAuthRequirer implements AuthRequirer {
+    private List<NamespaceMatcher> authorizedNamespaces;
+
+    PrivilegeBasedAuthRequirer(Map<Client, NamespaceMatcher> privileges) {
+        this.authorizedNamespaces = new ArrayList<>(privileges.values());
+    }
+
+    @Override
+    public boolean requiresAuth(Namespace namespace) {
+        return authorizedNamespaces.stream().anyMatch(namespaceMatcher -> namespaceMatcher.matches(namespace));
```
jeremyk-91: Maybe premature, but at some point this feels like `LoadingCache` material. It's a little unfortunate, but I don't see a good way to make the computation faster

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedAuthRequirer.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.palantir.atlasdb.keyvalue.api.Namespace;
+import com.palantir.atlasdb.timelock.auth.api.Client;
+import com.palantir.atlasdb.timelock.auth.api.NamespaceMatcher;
+
+public class PrivilegeBasedAuthRequirer implements AuthRequirer {
+    private List<NamespaceMatcher> authorizedNamespaces;
+
+    PrivilegeBasedAuthRequirer(Map<Client, NamespaceMatcher> privileges) {
+        this.authorizedNamespaces = new ArrayList<>(privileges.values());
```
jeremyk-91: nit: `ImmutableList.copyOf()`? Unless we plan to have some clever runtime live reloading or something

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -0,0 +1,49 @@
...
+    private final Map<Client, NamespaceMatcher> privileges;
+    private final AuthRequirer authRequirer;
+
+    SimpleAuthorizer(Map<Client, NamespaceMatcher> privileges, AuthRequirer authRequirer) {
+        this.privileges = privileges;
+        this.authRequirer = authRequirer;
+    }
+
+    public static Authorizer of(Map<Client, NamespaceMatcher> privileges, AuthRequirer authRequirer) {
+        return new SimpleAuthorizer(new HashMap<>(privileges), authRequirer);
```
jeremyk-91: nit: again, I'm thinking `ImmutableMap.copyOf()` unless we want this to specifically be mutable

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedAuthRequirer.java
gokcanozakdag: will add a cache here, but want to do it in a separate pr, I first need to clarify the live-reloading part.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedAuthRequirer.java
gokcanozakdag: makes sense, I think using a mutable parameter for live-reloading purposes would be hacky anyways.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/User.java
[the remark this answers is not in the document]
gokcanozakdag: thinking again, I feel like it makes more sense to move admin knowledge to authorizer, to keep it in one place. Otherwise both authenticator and authorizer need to be aware of admin users.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authenticator.java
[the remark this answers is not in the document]
felixdesouza: based on what we discussed yesterday, it's not necessarily the case that `user == client`, so I suppose it's okay? unless I've confused myself again

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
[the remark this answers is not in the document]
felixdesouza: fair enough

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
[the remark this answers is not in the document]
felixdesouza: is this before or after your previous comment 😛

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authenticator.java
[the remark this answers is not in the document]
gokcanozakdag: `User == Client`, (we can have multiple namespaces per client). My main concern here was creating the `BearerToken`. I believe it makes more sense to have a client id and a password to authenticate clients - addressed that here: https://github.com/palantir/atlasdb/pull/4036

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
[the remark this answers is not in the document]
gokcanozakdag: it was before haha :D Still explicitly defining `admin`s but doing it in `Authorizer`. (Rather than `Authenticator`)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -0,0 +1,61 @@
...
+import com.palantir.atlasdb.timelock.auth.api.Authorizer;
+import com.palantir.atlasdb.timelock.auth.api.NamespaceMatcher;
+import com.palantir.atlasdb.timelock.auth.api.Client;
+
+public class SimpleAuthorizer implements Authorizer {
+    private final Map<Client, NamespaceMatcher> privileges;
+    private final Set<Client> admins;
+    private final AuthRequirer authRequirer;
+
+    SimpleAuthorizer(Map<Client, NamespaceMatcher> privileges, Set<Client> admins, AuthRequirer authRequirer) {
```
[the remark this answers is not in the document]
felixdesouza: If I've understood correctly, we have:
* `Client` which consists of a username and password.
* We have a `PrivilegeConfig` block, which looks roughly like this:
```
privileges:
  - id: "my-first-timelock-client"
    namespaces:
      - namespace_1
      - namespace_2
    namespace-regexes/prefixes:
      - *.*
  - id: "my-admin-client"
    type: admin
```

So effectively from config you get a `List<PrivilegeConfig>` -> `Map<ClientId, Privileges>` where `Privileges` is either `StandardPrivileges` or `AdminPrivileges`. So it's somewhat similar to `NamespaceMatcher` but slightly different. It does mean we can get rid of `Set<Client>` as admins, since you can just call `Privilege.hasPrivilege(Namespace)`.

`AuthRequirer` seems a bit odd, it might make sense to have a class/interface called `Namespaces` and it's that which you ask `isNamespaceLocked`? It takes a `List<StandardPrivileges>`. I imagine this is 0.x only for namespace locking and not anything more general.

With above, it then feels weird to have a `Namespaces`/`AuthRequirer` when `List<PrivilegeConfig>` is the canonical source of truth. So I would just not inject it as a dependency. 🤷‍♂ 

It's unclear to me how if given a config that looks like the above, whether or not it maps to this authoriser thing cleanly imo.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/AuthRequirer.java
```diff
@@ -0,0 +1,34 @@
...
+package com.palantir.atlasdb.timelock.auth.impl;
+
+import java.util.Map;
+
+import com.palantir.lock.TimelockNamespace;
+import com.palantir.atlasdb.timelock.auth.api.Client;
+import com.palantir.atlasdb.timelock.auth.api.NamespaceMatcher;
+
+interface AuthRequirer {
+    AuthRequirer ALWAYS_REQUIRE = ignored -> true;
```
felixdesouza: when do you use these? outside of this PR of course

felixdesouza: that is both `ALWAYS` and `NEVER`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/impl/AuthRequirerTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public void shouldRequireAuthIfPrivilegeIsSpecified() {
+        Map<Client, NamespaceMatcher> privilege = ImmutableMap.of(CLIENT, createMatcherFor(NAMESPACE_1));
+        AuthRequirer authRequirer = AuthRequirer.deriveFromPrivileges(privilege);
+
+        assertThat(authRequirer.requiresAuth(NAMESPACE_1)).isTrue();
+        assertThat(authRequirer.requiresAuth(NAMESPACE_2)).isFalse();
+    }
+
+    @Test
+    public void ALWAYS_REQUIRE_shouldAlwaysRequireAuth() {
```
felixdesouza: I don't believe `ALWAYS_REQUIRE` and `NEVER_REQUIRE` tests are valuable. If anything these should be called `PrivilegeBasedAuthRequirerTests`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizerTest.java
```diff
@@ -0,0 +1,108 @@
...
+        withPrivilege(CLIENT_1, NAMESPACE_1);
+
+        assertAuthorized(CLIENT_1, NAMESPACE_1);
+        assertUnauthorized(CLIENT_2, NAMESPACE_1);
+        assertUnauthorized(Client.ANONYMOUS, NAMESPACE_1);
+    }
+
+    @Test
+    public void usersOnlyHaveAccessToTheirNamespaces() {
+        withLockedNamespace(NAMESPACE_1);
```
felixdesouza: Given how `SimpleAuthorizer` is setup, it is possible that you can have an unlocked namespace (that shouldn't be unlocked), but that namespace showing up in the privilege block. I would be better imo to derive that state from the state objects i.e. `Map<Client, Privileges>` block I was talking about before.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/AuthRequirer.java
gokcanozakdag: this will be used to configure authorizer for timelock 0.x and timelock 1.x; that is 0.x will be privilege based, 1.x will be always_require - and there will be a break-glass flag for both for a while that will set it to never_require. I am changing the structure here; with better documentation.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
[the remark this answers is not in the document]
gokcanozakdag: I agree; the possibility of inconsistency also bothered me. Though, thinking again; we need to be able to configure Authorizer for different auth requirement cases. So the new implementation takes `Map<Client, Privileges> privileges, AuthRequirement authRequirement`; where `AuthRequirement` is just a simple enum class without any implementation; and that prevents inconsistency I believe.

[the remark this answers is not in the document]
felixdesouza: what are the different cases?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/AuthRequirement.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.auth.impl;
+
+public enum AuthRequirement {
+    /**
+     * All namespaces are locked, only privileged clients have access.
+     */
+    ALWAYS_REQUIRE,
```
felixdesouza: this will only ever be true in timelock 1.x, at this point `AuthRequirement` would no longer be needed, and you'd just delete no?

`NEVER_REQUIRE` is the current state of the world, and I'm not sure how `PRIVILEGE_BASED` is disjoint from `NEVER_REQUIRE`.

if this is the break glass, it's a rather large break glass

gokcanozakdag: We wanted to keep the difference between 1.x and 0.x small, preferably just a single value rather than different implementations - as we will be supporting 0.x and 1.x for a while.
`PRIVILEGE_BASED` will be the current state of the world for 0.x; that is auth will be functional if a client publishes privileges block.
`NEVER_REQUIRE` will really be a break glass, it is actually a large break glass - but given we don't have any auth for timelock right now; it is not a regression not to use auth at all. I would like to be as safe as possible - if a client won't be able to talk to timelock for a while because of auth; it should be really easy to fix it to prevent stack wide outages.

lock-api/src/main/java/com/palantir/lock/TimelockNamespace.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.lock;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface TimelockNamespace {
+    String value();
```
felixdesouza: In #4042 you should add `@JsonSerialize/@JsonDeserialize` and `@JsonValue` so you don't have to have extra methods to convert `Set<String>` to `Set<TimelockNamespace>`.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedNamespaceLocker.java
```diff
@@ -0,0 +1,40 @@
...
+
+    PrivilegeBasedNamespaceLocker(Map<Client, Privileges> privileges) {
+        this.nonAdminPrivileges = privileges.values().stream()
+                .filter(privilege -> privilege != Privileges.ADMIN)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public boolean isLocked(TimelockNamespace namespace) {
+        return nonAdminPrivileges.stream().anyMatch(privilege -> privilege.hasPrivilege(namespace));
```
felixdesouza: don't think this is what you want, if this is *just* checking whether or not the namespace is locked, it shouldn't check whether or not then namespace has privilege

felixdesouza: ignore me

felixdesouza: however, this will suck, since we'd have like 80 or so clients on current deployments

felixdesouza: basically I think you should merge your config PR #4042 in first, and work with what you've got, otherwise you're providing impls that are somewhat pointless and will change immediately.

I would in the constructor or whatever, take in the privilege map, visit each privilege, and extract the static namespaces into a set, and then additionally extract the regexes (and that'd be the only thing you'd have to loop through, but you can slap on a cache there).

gokcanozakdag: yeah, I know this one is not optimal for perf, I was thinking of having a CachedAuthenticator to prevent going through this path all the time. Wanted to do it in a separate pr though.

lock-api/src/main/java/com/palantir/lock/TimelockNamespace.java
gokcanozakdag: Will be adding it to following pr.

gokcanozakdag:
Shall we merge this @felixdesouza ?

PR #3918, opened 2019-04-10
Avoid Iterables.partition when Lists.partition is available

diogoholanda:
**Goals (and why)**:
Reduce the number of unnecessary memory allocations by Atlas

Iterables.partition pre-allocates an Array of the desired batch size for every batch. This is generally a good perf optimization, but can lead to bad situations in extreme scenarios.
For instance, LockRefreshingLockService refreshes the locks in batches of 500K every 5 seconds. For a service containing just a few locks, this leads to allocation of arrays of size 500K every 5 seconds (2Mb every 5 seconds).
If the Iterable being partitioned is a List, Lists.partition is much more efficient and simply relies on the .sublist method.

Existing tests passing should be enough. This is a simple refactor.

I couldn't convert all usages of Iterables.partition because Atlas uses Iterables as parameters extensively. I think that the main cases were converted. Refactoring the code to be able to switch the other cases might not be worth it. I will leave this up to the dev team.

Anywhere

Not super high priority, but a nice improvement

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetEmptyLatestValues.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetEmptyLatestValues.java
@@ -20,7 +20,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.palantir.atlasdb.keyvalue.api.Cell;
@@ -53,7 +53,7 @@ public class GetEmptyLatestValues {
      */
     public Set<Cell> execute() {
         Set<Cell> result = Sets.newHashSet();
-        for (List<CellWithTimestamps> batch : Iterables.partition(cellTimestamps, batchSize)) {
+        for (List<CellWithTimestamps> batch : Lists.partition(cellTimestamps, batchSize)) {
             result.addAll(getSingleBatch(batch));
         }
         return result;
--- a/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
@@ -24,8 +24,9 @@ import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.palantir.common.concurrent.PTExecutors;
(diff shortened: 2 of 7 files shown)


lock-api/src/test/java/com/palantir/lock/client/AsyncTimeLockUnlockerTest.java
```diff
@@ -84,7 +83,7 @@ public void enqueueingTokensEnqueuedSinglyAllEventuallyCaptured() {
     public void enqueueingTokensEnqueuedInGroupsAllEventuallyCaptured() {
         setupTokenCollectingTimeLock();
 
-        Iterables.partition(tokenList, 7).forEach(tokens -> unlocker.enqueue(ImmutableSet.copyOf(tokens)));
```
jeremyk-91: Hmm, 7 isn't too bad...

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -124,13 +125,13 @@ public boolean unlock(HeldLocksToken token) {
...
     private void refreshLocks() {
-        ImmutableSet<LockRefreshToken> refreshCopy = ImmutableSet.copyOf(toRefresh);
+        ImmutableList<LockRefreshToken> refreshCopy = ImmutableList.copyOf(toRefresh);
         if (refreshCopy.isEmpty()) {
             return;
         }
         Set<LockRefreshToken> refreshedTokens = new HashSet<>();
         // We batch refreshes to avoid sending payload of excessive size
-        for (List<LockRefreshToken> tokenBatch : Iterables.partition(refreshCopy, REFRESH_BATCH_SIZE)) {
+        for (List<LockRefreshToken> tokenBatch : Lists.partition(refreshCopy, REFRESH_BATCH_SIZE)) {
```
jeremyk-91: this one is 500k so a big win

schlosna:
I [really wish Guava would add the `instanceof List` check to `Iterables.partition` and `Iterables#transform` to get the optimized versions without copies](https://github.com/google/guava/issues/1413#issuecomment-61353163). We do have our own `IterableView#partition` that does the check here that we should probably audit and use throughout AtlasDB: 

https://github.com/palantir/atlasdb/blob/e972ae5624435bfe75c4e887cd5f280d193d4554/atlasdb-commons/src/main/java/com/palantir/common/collect/IterableView.java#L118

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/sweep/GetEmptyLatestValues.java
```diff
@@ -53,7 +53,7 @@ public GetEmptyLatestValues(
      */
     public Set<Cell> execute() {
         Set<Cell> result = Sets.newHashSet();
-        for (List<CellWithTimestamps> batch : Iterables.partition(cellTimestamps, batchSize)) {
+        for (List<CellWithTimestamps> batch : Lists.partition(cellTimestamps, batchSize)) {
```
iamdanfox: I've filed https://github.com/palantir/gradle-baseline/issues/621 so we can build this into an error-prone rule and hopefully catch other occurrences!

PR #3924, opened 2019-04-12
[_transactions2] Part 41: Autobatcher Profiling

jeremyk-91:
**Goals (and why)**:
- Get some idea as to what batch sizes for the write batching transaction service.

- Add profiling logging every 10 seconds.

Not much, I'll admit. Existing tests should catch that nothing has gone terribly wrong.

- I would have preferred to go for a histogram of sizes as well as the delay between submission and processing. However, workarounds already exist, and it seems annoyingly nontrivial to implement this nicely (e.g. MetricsManager is in atlasdb-api, which depends on lock-api, which depends on atlasdb-autobatch... and moving MetricsManager to atlasdb-autobatch seems wrong)
- ~Should this be extended to the other autobatchers?~ Extended it to the other autobatchers.
- The batch size logging function is not thread safe. Am I correct in implementing it to not be thread safe?

**Where should we start reviewing?**: ProfilingAutobatchers

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/ProfilingAutobatchers.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/ProfilingAutobatchers.java
@@ -1,0 +1,54 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.autobatch;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.function.Consumer;
+
+import org.slf4j.Logger;
+
+import com.google.common.util.concurrent.RateLimiter;
+import com.palantir.logsafe.SafeArg;
+
+public final class ProfilingAutobatchers {
+
+    private static final Duration AUTOBATCHER_LOGGING_INTERVAL = Duration.ofSeconds(10);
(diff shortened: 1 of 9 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/ProfilingAutobatchers.java
```diff
@@ -0,0 +1,54 @@
...
+        RateLimiter profilingLimiter = RateLimiter.create(AUTOBATCHER_LOGGING_PERMITS_PER_SECOND); // every 10 seconds
+        return DisruptorAutobatcher.create(elements -> {
+            batchFunction.accept(elements);
+
+            if (profilingLimiter.tryAcquire()) {
+                logger.info("Autobatcher with ID {} just processed a batch of size {}."
+                                + " This message is rate limited to once every 10 seconds per autobatcher, so there"
+                                + " may be more batches being processed.",
+                        SafeArg.of("autobatcherIdentifier", safeIdentifier),
+                        SafeArg.of("batchSize", elements.size()));
```
gokcanozakdag: So we are logging number of elements batched for that single call, every 10 seconds. If calls are homogeneous enough, that should give decent signal - but if it is not, this may not be that helpful. Can we keep a running average and log that, instead of logging only the current value?

docs/source/release_notes/release-notes.rst
```diff
@@ -130,6 +130,10 @@ develop
          - Removed unnecessary memory allocations in the lock refresher, and in several other classes, by using Lists.partition(...) instead of Iterables.partition(...).
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3918>`__)
 
+    *    - |improved|
+         - AtlasDB now logs diagnostic information about uses of the request batching timestamp service.
```
gokcanozakdag: I guess this should have been write batching transaction service instead of request batching timestamp service, right?

jeremyk-91: Yeah, made an error there. I've expanded the scope of this to hit all the batchers, so this has been reworded.

felixdesouza:
Can `DisruptorAutobatcher.create` not give us a `ProfilingDisruptor` or whatever? I can't think of a case where I'd want an autobatcher without any profiling.

felixdesouza:
I suppose for later, but why is `MetricsManager` in api? it's a utility class.. :S Agree moving it to autobatch is wrong, but surely there is a better place for it, like atlasdb-common?

felixdesouza:
api depends on atlasdb-common

felixdesouza:
re thread safety, it doesn't matter if it is subsumed by the `DisruptorAutobatcher` imo, then make the class and its static method package private and you should be good to go.

PR #3929, opened 2019-04-15
Improve CassandraClientPoolingContainer metrics and settings

diogoholanda:
**Goals (and why)**:
I am investigating some potential resource not being closed by the Cassandra integration.
I am interested in how many clients get created/evicted and when does that happen.
Those questions are hard to answer with the current metrics.

Produce the desired metrics.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -257,7 +257,6 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
 
         // immediately throw when we try and borrow from a full pool; dealt with at higher level
         poolConfig.setBlockWhenExhausted(false);
-        poolConfig.setMaxWaitMillis(config.socketTimeoutMillis());
 
         // this test is free/just checks a boolean and does not block; borrow is still fast
         poolConfig.setTestOnBorrow(true);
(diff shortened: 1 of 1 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -257,7 +257,6 @@ public String toString() {
 
         // immediately throw when we try and borrow from a full pool; dealt with at higher level
         poolConfig.setBlockWhenExhausted(false);
-        poolConfig.setMaxWaitMillis(config.socketTimeoutMillis());
```
diogoholanda: `MaxWaitMillis` is ignored when `BlockWhenExhausted` is false.
This configuration means how long to block if the pool is exhausted.
It is also weird that this was getting set to the socket timeout millis. This seemed like an oversight

diogoholanda: Note that this was originally added in a PR when `BlockWhenExhausted` was true: https://github.com/palantir/atlasdb/commit/99de2f76d2ea9b47ba16b3eb0c3ffc52970d3278#diff-b88d7ebe9dabc333cf7e4dff942e1830R149

jeremyk-91: Yep, confirming that this is correct - it is indeed ignored (from `GenericObjectPool::borrowObject`)

jeremyk-91:
will add no-changelog tag

diogoholanda:
Thanks! Can you merge this @jeremyk-91 since I don't have the permissions needed?
I would also appreciate a release if that's possible, so that I can continue my investigation

PR #3913, opened 2019-04-09
[TLS] Timelock Stability: Single CachedThreadPool for LockServiceImpls

gmaretic:
**Goals (and why)**:
Achieve better thread reuse and reduce overhead of creating an executor service per client on gaining leadership

**Implementation Description (bullets)**:
Pretty straightforward

**Testing (What was existing testing like?  What have you done to improve it?)**:
Smoke tested

**Concerns (what feedback would you like?)**:
Did not address the single threaded executors for async lock, as that is more tricky to do and seems less beneficial.

PR #3933, opened 2019-04-16
[TLS] TImelockServiceImpl #2

gmaretic:
**Goals (and why)**:
Merge in #3913 without breaking develop


atlasdb-commons/src/main/java/com/palantir/util/Owned.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.util;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface Owned<R> {
```
jeremyk-91: hmm, `Owned.notOwned(stuff)` seems like an awkward construction to me

PR #3935, opened 2019-04-16
[TLS] Do not retry for pingable leaders

gmaretic:
**Goals (and why)**:
#3884 without breaking develop

Same as before, but separate the change from removal of clock skew monitor (will be done in a separate PR), and use the non-retrying client only for the pingable leader to not break tests.


felixdesouza:
This only affects the timelock to timelock retries? I feel like we did something with client to timelock retries no?

felixdesouza:
good shout on separating into two prs!

gmaretic:
We were playing around with reducing the backoff in FFT, which might make sense. I'll look into it

gmaretic:
@felixdesouza do we still think this makes sense?

felixdesouza:
As opposed to?

PR #3937, opened 2019-04-16
Use SlidingTimeWindowArrayReservoir for metrics

gmaretic:
**Goals (and why)**:
Replace the exponentially decaying reservoir with SlidingTimeWindowArrayReservoir in our timers.

Replace tritium's invocation handler with ours that wires in the change.

Nothing so far

This replaces all metrics that are created with calls to `AtlasDbMetrics.instrument`. This may be too broad, and maybe we want to be more selective.
Also, the approach is more hacky than I'd like, but also seems like the most reasonable way to do it.


atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowMetricsInvocationHandler.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        markGlobalFailure();
+        String failuresMetricName = MetricRegistry.name(getBaseMetricName(context), FAILURES_METRIC_NAME);
+        metricRegistry.meter(failuresMetricName).mark();
+        metricRegistry.meter(MetricRegistry.name(failuresMetricName, cause.getClass().getName())).mark();
+    }
+
+    private static Timer createNewTimer() {
+        return new Timer(new SlidingTimeWindowArrayReservoir(35, TimeUnit.SECONDS));
```
schlosna: curious why 35 second sliding window?

schlosna:
Related to https://github.com/palantir/tritium/pull/238

atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowMetricsInvocationHandler.java
felixdesouza: 30 sec is our collection rate I believe, and didn't want to have anything missed out due to weird timing things, so put it at 35. Ideally we'd want it at 30 - to be fair, I'm not sure I tried with 30 seconds whilst testing.

felixdesouza: also re window size, if we keep it small, we have to worry less about memory footprint since we're only keeping last 30 sec of metrics in memory as opposed to last 5 mins for example plus dd stops lying to me

atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsInvocationEventHandlerUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.SlidingTimeWindowArrayReservoir;
+import com.codahale.metrics.Timer;
+import com.palantir.tritium.event.InstrumentationProperties;
+import com.palantir.tritium.event.InvocationContext;
+
+public final class MetricsInvocationEventHandlerUtils {
```
felixdesouza: package private?

felixdesouza: can we call this something easier to consume or just static import where we use this?

gmaretic:
@jeremyk-91 are we OK with this as is?

schlosna:
Might be worth waiting for https://github.com/palantir/tritium/pull/249

jeremyk-91:
Yep, I'm fine with this as is, but it looks like tritium is incorporating this change

schlosna:
Tritium 0.12.0 has the default sliding window tagged registry you all can hook into.

felixdesouza:
LGTM!

PR #3943, opened 2019-04-17
[TLS] Change Default Paxos Configs

gmaretic:
**Goals (and why)**:
Change the default paxos configs to values that make leader elections happen faster when the leader goes down, while not introducing instability.

Tested out many configurations on our test cluster. The end configuration was:
```
        leader-ping-response-wait-in-ms: 2000
        maximum-wait-before-proposal-in-ms: 200
        ping-rate-in-ms: 50
```
Decided to go for a more conservative 3000, 300, 500 here just in case.

Are we too conservative, or should we be more conservative?


felixdesouza:
I feel like 2s is sufficient for `leader-ping-wait-time`. Together with the lack of retrying, it feels odd that a request will respond meaningfully after 2.8 seconds.

felixdesouza:
for random-wait, you've set it to 300ms in code, as opposed to 500ms in the pr description, I imagine you want 300.

Is 300ms sufficient gap such that 3 timelock nodes (we have a single node trying at most once with our autobatcher fix) won't collide?

felixdesouza:
I'm not entirely sure I understand why ping rate has gone back up to 500ms? Granted most of the testing I did before you tweaked numbers was with 500ms.

I feel like we can lower it to around 50ms or maybe 200ms if we're being conservative, reason being:
* request through a leadership proxy comes in and realises it's not the leader
* immediately marks as not leading
* assuming a constant rate of requests (not great, but walk with me), it's likely that only one client is the one blocking on leadership, it may be us, it may be a different client, but either way, assuming the proxy knew we lost leadership the moment one of the background thread (i.e. leader has died, client has failed over to current node), then we will 503 *until* some number between `(0, ping-rate)` has elapsed + `random-wait-time` no?

My hypothesis is that this goes away if `LeaderElectionService` didn't do any sleeping and was more of a `ScheduledExecutorService`. Then you could schedule a leader election immediately vs the schedule at the leader ping rate etc. 

I'm not sure what the harm is in having a lower ping, which is around 20 calls per second.

gmaretic:
Sure, we can definitely lower the two other configs. I've seen instability on our test cluster when leader-ping-wait-time was set to 1 second, but 2 is probably reasonable.

Reducing ping to 50 is probably reasonable, just seemed a bit excessive to reduce it by a factor of 100.

As for the max wait, 200 has been stable on our test cluster, and we do want to have it as low as possible.

Should we do (3000, 300, 500) -> (2000, 300, 50) ?

felixdesouza:
I'm happy to do (2000, 300, 50), happy to do (2000, 300, 150-200) if we're being conservative.

Would be good to get other eyes on this too.

docs/source/release_notes/release-notes.rst
```diff
@@ -66,6 +66,14 @@ v0.133.0
     *    - Type
          - Change
 
+    *    - |changed|
+         - Changed the default values in ``PaxosConfiguration``.
+           ``leader-ping-response-wait-in-ms`` was reduced to 2000 ms from 5000 ms.
+           ``maximum-wait-before-proposal-in-ms`` was reduced to 300 ms from 1000 ms.
+           ``ping-rate-in-ms`` was reduced to 50 ms from 5000 ms.
+           These settings have empirically improved the performance of timelock when the leader node goes down without negatively affecting stability.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/???>`__)
```
jeremyk-91: please fill this in

docs/source/release_notes/release-notes.rst
```diff
@@ -66,6 +66,14 @@ v0.133.0
     *    - Type
          - Change
 
+    *    - |changed|
```
jeremyk-91: nit: The last line of this makes this sound like an `|improved|` is reasonable

PR #3944, opened 2019-04-17
Close CassandraClients constructed via CassandraClientFactory::getClientInternal

diogoholanda:
**Goals (and why)**:
Close CassandraClients rather than relying on Java finalizers to close them.
CassandraClients that are not closed keep references to open connections/sockets that will only be cleaned up at Finalizer time, which delays garbage collection of large objects.
The Finalizer runs in a single thread with low priority. For some reason this problem is exarcebated in java 11.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
@@ -205,8 +205,7 @@ public final class CassandraVerifier {
     // swallows the expected TException subtype NotFoundException, throws connection problem related ones
     private static boolean keyspaceAlreadyExists(InetSocketAddress host, CassandraKeyValueServiceConfig config)
             throws TException {
-        try {
-            CassandraClient client = CassandraClientFactory.getClientInternal(host, config);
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraVerifier.java
```diff
@@ -205,8 +205,7 @@ private static boolean attemptToCreateIfNotExists(CassandraKeyValueServiceConfig
     // swallows the expected TException subtype NotFoundException, throws connection problem related ones
     private static boolean keyspaceAlreadyExists(InetSocketAddress host, CassandraKeyValueServiceConfig config)
             throws TException {
-        try {
-            CassandraClient client = CassandraClientFactory.getClientInternal(host, config);
+        try (CassandraClient client = CassandraClientFactory.getClientInternal(host, config)) {
```
jeremyk-91: Given we do similar checks in e.g. `sanityCheckRingConsistency()`, and closing the client involves resetting the input and output protocols, this seems right.

jeremyk-91:
I'll add a changelog entry

diogoholanda:
@jeremyk-91 I don't have merge rights. Can you merge this one, and potentially release?

PR #3945, opened 2019-04-17
[TLS] Reduce Backoffs in FailoverFeignTarget

gmaretic:
**Goals (and why)**:
Reduce the backoffs in FFT, because they can cause huge delays when leadership election occurs.

Two different backoffs -- reduced the max backoff in the normal case from 3 to 1 second, reduced the backoff before round robin retry from 250-750 to 0-250 millis. 

Is this too aggressive? Did not see a compelling reason for such conservative backoffs, especially in the case of round robin, because that is always what you will hit when the leader goes down.


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
```diff
@@ -113,8 +114,7 @@ public void continueOrPropagate(RetryableException ex) {
                 // We implement some randomness around the expected value of BACKOFF_BEFORE_ROUND_ROBIN_RETRY_MILLIS.
                 // Even though this is not exponential backoff, should be enough to avoid a thundering herd problem.
                 long pauseTimeWithJitter = ThreadLocalRandom.current()
-                        .nextLong(BACKOFF_BEFORE_ROUND_ROBIN_RETRY_MILLIS / 2,
-                                (BACKOFF_BEFORE_ROUND_ROBIN_RETRY_MILLIS * 3) / 2);
+                        .nextLong(MAX_BACKOFF_BEFORE_ROUND_ROBIN_RETRY.toMillis());
```
jeremyk-91: OK. I think this is fine as the point is only to avoid a thundering herd problem as mentioned.

nit: The comment should be updated

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/FailoverFeignTarget.java
```diff
@@ -46,8 +47,8 @@
 public class FailoverFeignTarget<T> implements Target<T>, Retryer {
     private static final Logger log = LoggerFactory.getLogger(FailoverFeignTarget.class);
 
-    public static final int DEFAULT_MAX_BACKOFF_MILLIS = 3000;
-    public static final long BACKOFF_BEFORE_ROUND_ROBIN_RETRY_MILLIS = 500L;
+    public static final Duration DEFAULT_MAX_BACKOFF = Duration.ofMillis(1000);
```
felixdesouza: ❤️

PR #3946, opened 2019-04-17
[TLS] Better getSuspectedLeaderOverNetwork

gmaretic:
**Goals (and why)**:
If none of the nodes give us the desired uuid as a response, we just end up waiting until timeout when we know there are no more requests. Instead just fail immediately. Also, remove the pointless attempt to get remaining responses after.

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
@@ -304,9 +304,8 @@ public class PaxosLeaderElectionService implements PingableLeader, LeaderElectio
         // collect responses
         boolean interrupted = false;
         try {
-            long deadline = System.nanoTime()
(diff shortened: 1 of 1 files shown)


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -334,22 +331,6 @@ boolean getAndRecordLeaderPingResult(@Nullable Future<Boolean> pingFuture) throw
                     log.warn("unable to get uuid from server", e);
                 }
             }
-
-            // poll for extra completed futures
-            Future<Entry<String, PingableLeader>> future;
-            while ((future = pingService.poll()) != null) {
```
jeremyk-91: In theory, if the ping was interrupted you can still cache the results of any computations that finished in between the interruption and its handler finishing, which looks like what this does - so strictly speaking this does have a point, though I agree it's not worth the complexity.

PR #3949, opened 2019-04-19
remove db username from hikari pool name

berler:
The Hikari connection pool name gets used as the name for metrics, so it
should not include the DB username credentials.

**Goals (and why)**: When trying to create default metrics dashboards for internal products, I noticed that all of the hikari pool metrics contain the DB username as part of the metric name. The DB username is something that is usually going to be customized on each install, so this would make it impossible to have a set of standard metrics dashboards for various internal products (each dashboard would need to be customized to query a different metric name depending on the configured DB username).

There is still plenty of room to control the name of metrics by setting the pool identifier and the connection id (which don't reveal any actual DB credentials). These config values are usually set by products (and not changed on individual installs).

**Implementation Description (bullets)**: Removed the username from the connection pool name

### DIFF
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
@@ -114,10 +114,9 @@ public abstract class ConnectionConfig {
     @Value.Derived
     public String getConnectionPoolName() {
         return String.format(
-                "%s-%s-%s",
+                "%s-%s",
                 getConnectionPoolIdentifier(),
-                getConnId(),
-                getDbLogin());
+                getConnId());
     }
 
     /**
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -145,6 +145,10 @@ develop
          - ``AtlasDbHttpClients.createProxyWithFailover()`` now requires ``UserAgent`` parameter.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/3996>`__)
 
+    *    - |fixed|
+         - Removed the DB username from the Hikari connection pool name.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/3949>`__)
+
 ========
 v0.133.0
 ========


atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
```diff
@@ -114,10 +114,9 @@ public String getConnectionPoolIdentifier() {
     @Value.Derived
     public String getConnectionPoolName() {
         return String.format(
-                "%s-%s-%s",
+                "%s-%s",
                 getConnectionPoolIdentifier(),
-                getConnId(),
-                getDbLogin());
+                getConnId());
```
carterkozak: Could we use a tritium tagged metric registry + tagged MetricName for hikari pools? the connId and pool identifier would be great tag values on a single `database.pool` metric name (example name was the first thing I though of, I'm sure there's a better name for it)

berler: I'm not sure if this is possible, because I think the metrics are just coming from an MBean from Hikari (the name we set here I think becomes the name of the MBean)

berler: Actually, it looks like we can make an improvement here, Hikari has a PrometheusMetricsTracker that should be able to register prometheus metrics directly (the pool name becomes a label on the metrics instead of being part of the metric name). If we want some custom tagged metrics registry we could probably implement our own MetricsTrackerFactory

carterkozak: I'd prefer to wire it directly to the tritium registry, which can be connected to prometheus as well as sls-logging metrics pipelines.

berler: To do that we'd need to make a custom MetricsTrackerFactory. I think it would probably be better to do that in a separate change.

clockfort:
I'm pretty sure we did this because otherwise it gets Real Confusing<sup>TM</sup> when you're looking at the multiple pools, since we used have separate legacy/atlas ones in our old product. Even if those were combined into one now, things can still get a little confusing when you're doing KVS migrations etc where you'd still have multiple pools in the same JVM.

As long as at the end of the day the pools look at least slightly different in name and don't stomp on the same metric endpoint registrations, this change is fine, though.

berler:
> As long as at the end of the day the pools look at least slightly different in name and don't stomp on the same metric endpoint registrations, this change is fine, though.

In our internal product now, we set the pool id to be different for each pool (and descriptive of what that pool is used for), so the username is not needed to distinguish them.

PR #3953, opened 2019-04-23
[PDS-88514 / _transactions2 Part 44] Instrument CoordinationStore in addition to CoordinationService

jeremyk-91:
**Goals (and why)**:
- Have metrics tracking coordination _store_ operations, in addition to coordination service operations which were already tracked. We initially tracked service operations as these manifest as direct latency to the user, but store operations, especially rates, are interesting as they may cause KVS load.

- Instrument the coordination store.
- Push instrumentation of CoordinationService (and the now-added CoordinationStore) in to the CoordinationServices factory.

Not much. Existing tests should verify the store still functions properly.

**Concerns (what feedback would you like?)**: Nothing in particular.

**Where should we start reviewing?**: CoordinationServices

**Priority (whenever / two weeks / yesterday)**: today 🔥 

@tpetracca for SA

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -456,14 +456,11 @@ public abstract class TransactionManagers {
             MetricsManager metricsManager,
             LockAndTimestampServices lockAndTimestampServices,
             KeyValueService keyValueService) {
-        @SuppressWarnings("unchecked") // Coordination service clearly has this type.
-        CoordinationService<InternalSchemaMetadata> metadataCoordinationService = AtlasDbMetrics.instrument(
-                metricsManager.getRegistry(),
-                CoordinationService.class,
-                CoordinationServices.createDefault(
-                        keyValueService,
-                        lockAndTimestampServices.timestamp(),
-                        config().initializeAsync()));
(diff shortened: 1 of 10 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -456,14 +456,11 @@ private static TransactionSchemaInstaller initializeTransactionSchemaInstaller(
...
             KeyValueService keyValueService) {
-        @SuppressWarnings("unchecked") // Coordination service clearly has this type.
-        CoordinationService<InternalSchemaMetadata> metadataCoordinationService = AtlasDbMetrics.instrument(
-                metricsManager.getRegistry(),
-                CoordinationService.class,
-                CoordinationServices.createDefault(
-                        keyValueService,
-                        lockAndTimestampServices.timestamp(),
-                        config().initializeAsync()));
+        CoordinationService<InternalSchemaMetadata> metadataCoordinationService = CoordinationServices.createDefault(
```
felixdesouza: ❤️

jeremyk-91:
`CoordinationStore` and a bunch of the other internal pieces can now be package private, actually. Will update with that

tpetracca:
I kindly remind you that I'm at least a consumer of `CoordinationServices.createDefault` so that method needs to remain public :)

PR #3954, opened 2019-04-24
[PDS-88767 / _transactions2 Part 45] Semantic Equivalence for VersionedInternalSchemaMetadata

jeremyk-91:
**Goals (and why)**:
- See PDS-88767

- Check VISMs by deserializing them and checking if the payloads are equal (assuming we know how to), as opposed to checking for byte equality on the payloads.

**Testing (What was existing testing like?  What have you done to improve it?)**: Added some tests for these cases.

- Is rejecting different-versions (even if they deserialize to equal objects) correct? I'd argue they are, since they become clearly different on a version that only understands how to deserialize one of them.
- Is rejecting anything we cannot deserialize reasonable?

**Where should we start reviewing?**: VersionedInternalSchemaMetadata then KeyValueServiceCoordinationStore

**Priority (whenever / two weeks / yesterday)**: yesterday :fire:

@tpetracca for SA


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadata.java
```diff
@@ -28,11 +31,38 @@
...
 public interface VersionedInternalSchemaMetadata {
     int version();
     byte[] payload();
+
+    /**
+     * Returns the underlying {@link InternalSchemaMetadata} for this {@link VersionedInternalSchemaMetadata},
+     * provided we are able to deserialize it. Otherwise returns empty.
+     */
+    @Value.Lazy
+    default Optional<InternalSchemaMetadata> underlyingData() {
```
felixdesouza: I think you want a `@JsonIgnore` on here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadata.java
```diff
@@ -28,11 +31,38 @@
...
+        return InternalSchemaMetadataPayloadCodec.tryDecode(this);
+    }
+
+    /**
+     * Returns true iff this {@link VersionedInternalSchemaMetadata} object and the other object are knowably
+     * semantically equivalent - that is, they have the same version, which this product version is able to understand,
+     * and the {@link InternalSchemaMetadata} that underlies the payloads are equal.
+     */
+    @Value.Lazy
+    default boolean knowablySemanticallyEquivalent(VersionedInternalSchemaMetadata other) {
```
felixdesouza: similarly here

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadata.java
jeremyk-91: Sure, will add this.

felixdesouza:
As mentioned offline, Immutables + Jackson allow you to have generic params + deserializing etc.

I feel like if we have to do *this* to get the right behaviour, we might as well use jackson + immutables properly as this is more their territory than ours.

What do I mean by this?

VersionedInternalSchemaMetadata.java:

```java
public interface VersionedInternalSchemaMetadata<T> {
  int version();
  T payload();
}
```

When you want to deserialize:
```java
objectMapper.readValue(object, new TypeReference<VersionedInternalSchemaMetadata<InternalSchemaMetadata>> {});
```

We should now have something that can deserialise into an object with expected equality checks with a lot less effort. I also believe its compatible with what we have now since `byte[]` is very general.

If we need to change `InternalSchemaMetadata`, we can always do polymorphic deserialization on it.

Unless I'm missing context/extra constraints, I feel like this would be a better forward step.

jeremyk-91:
> Unless I'm missing context/extra constraints, I feel like this would be a better forward step.

The main constraints we have here would be time I think, since this would be a somewhat larger change and stuff is on fire. Also I think you'd need the polymorphic deserialization from the start, since version 1 is already going to have a byte array for the payload - you'd start with version 2 (unless I'm missing something)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadata.java
```diff
@@ -28,11 +32,40 @@
...
+
+    /**
+     * Returns true iff this {@link VersionedInternalSchemaMetadata} object and the other object are knowably
+     * semantically equivalent - that is, they have the same version, which this product version is able to understand,
+     * and the {@link InternalSchemaMetadata} that underlies the payloads are equal.
+     */
+    @Value.Lazy
+    @JsonIgnore
+    default boolean knowablySemanticallyEquivalent(VersionedInternalSchemaMetadata other) {
+        if (version() != other.version()) {
```
gokcanozakdag: do we need a null check here?

jeremyk-91: Hmm, `other` shouldn't be null - though I guess this isn't as easy to enforce. I've added one.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadataTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+public class VersionedInternalSchemaMetadataTest {
+    private static final Path BASE_METADATA_PATH = getResourcePath("metadata-v1-base.json");
+    private static final byte[] BASE_METADATA_BYTES = readAllBytesUnchecked(BASE_METADATA_PATH);
+    private static final VersionedInternalSchemaMetadata V1_BASE = createVersionedMetadata(1, BASE_METADATA_BYTES);
+
+    @Test
+    public void metadataSemanticallyEquivalentToItselfIfCanBeDeserializedAndSemanticsAgree() {
+        Assume.assumeTrue(V1_BASE.underlyingData().isPresent());
+        assertThat(V1_BASE.knowablySemanticallyEquivalent(V1_BASE)).isTrue();
```
gokcanozakdag: why do we want to use `Assume` here? Can it cause silent broken tests at some point? I feel like it is mostly meant to be used for `Theories`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadata.java
```diff
@@ -28,11 +32,40 @@
...
+     */
+    @Value.Lazy
+    @JsonIgnore
+    default Optional<InternalSchemaMetadata> underlyingData() {
+        return InternalSchemaMetadataPayloadCodec.tryDecode(this);
+    }
+
+    /**
+     * Returns true iff this {@link VersionedInternalSchemaMetadata} object and the other object are knowably
+     * semantically equivalent - that is, they have the same version, which this product version is able to understand,
```
gokcanozakdag: I think it will return true even if you have different versions, but still be able decode it into same object - so if only difference is the way you encrypt it. Is that correct?

jeremyk-91: (Discussed offline - there is actually a version check)

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/internalschema/persistence/VersionedInternalSchemaMetadataTest.java
jeremyk-91: I originally wrote this this way because I didn't want this test to fail if we suddenly only became to deserialize v2 onwards, but we wouldn't have a test for this case if we did that which is bad. I've removed the Assume

PR #3957, opened 2019-04-25
Use soft evict limits not evict limits with the pool

j-baker:
At the moment, we evict every connection that has been idle for 10
minutes every 10 minutes. If there is a min pool size (our default is
30) this is then immediately replaced with another SSL connection;
essentially just churning every 10 minutes.

In practice, we've seen this cause issues internally around GC; SSL
connections filling up the old gen and then finalizer issues leading to
an ultimate GC overhead exceeded OOM.

I guess my concern would be that the Thrift connections actually break
after a certain period and so we are essentially evicting anyway (but
later, taking a performance hit); in this case we will see logging
occurring in such services indicating this.

@clockfort would love your feedback here as to whether we'd need to make other changes here (e.g. changing the healthcheck functionality?).

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -261,8 +261,10 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
         // this test is free/just checks a boolean and does not block; borrow is still fast
         poolConfig.setTestOnBorrow(true);
 
-        poolConfig.setMinEvictableIdleTimeMillis(
+        poolConfig.setSoftMinEvictableIdleTimeMillis(
                 TimeUnit.MILLISECONDS.convert(config.idleConnectionTimeoutSeconds(), TimeUnit.SECONDS));
+        poolConfig.setMinEvictableIdleTimeMillis(Long.MAX_VALUE);
+
         // the randomness here is to prevent all of the pools for all of the hosts
         // evicting all at at once, which isn't great for C*.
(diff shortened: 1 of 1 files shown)


clockfort:
Are we talking about GC issues on atlas side or C* server side?




This fix seems fine, we might also want to lower the default **CassandraConfig.proportionConnectionsToCheckPerEvictionRun** if we feel that the GC pressure from mass connection close/reopen is too great, which will not only fix the GC issue you're seeing in idle time but also during pool spin-down times (with a side effect being that it might take longer for the pool to shrink to idle after a burst). Though, this might not greatly improve things if the problem is only with connection opens and not connection closes.



I know that well-enforcing a max lifetime / recycling idle connections is more necessary for postgres/oracle, where we've seen better behaviour using it in weird situations like CNAME changes in RDS and having updated oracle settings start to apply to sessions; but none of the issues I can think of should apply to C*.

I mirror your worry about long-open idle connections going sour at some point, though afaik there's no explicit Cassandra-side max thrift connection lifetime and it relies only on host OS TCP keepalive to keep long-idle thrift connections reliably open. We may or may not have keepalive enabled everywhere.

clockfort:
I guess what I'm saying is, you can do this if you want, but maybe we want to just make the evictor run with a much smaller batch size and still reap idle connections.

j-baker:
It's on the Atlas side, clocks. There are some servers that use Atlas but barely do anything. In this case, their old gen fills up with SSLSockets, which have a finalize method. Finalization delays GC and causes an OOM due to GC overhead limits. It's pretty miserable.

diogoholanda:
How can I check what is wrong with this build? I am interested on helping push this through.

jeremyk-91:
Missing changelog - I'll drive this through today.

jeremyk-91:
(Diogo and I spoke offline, but the 'issue' was actually a flake)

PR #3974, opened 2019-05-01
Upgrade gradle-baseline to 0.57.0

schlosna:
**Goals (and why)**:
Update gradle-baseline to latest version.

Bumped versions and structure to conform to new error-prone argument structure.

only automated CI tests

curious why atlasdb's gradle-baseline was so far behind previously, is there something I'm missing for why we're not keeping it up-to-date?

`build.gradle`, `gradle/baseline.gradle`, and `versions.props`

whenever


schlosna:
Thanks @jeremyk-91 , I am actually going to split this up into separate PRs, this one just to bump baseline, then a follow on to actually apply the error-prone checks & fixes

felixdesouza:
hmm, so something to do with errorprone and circle :( would be nice to get to latest, so we can unbreak the build

felixdesouza:
To close this out, 0.32.0 of baseline introduced a different errorprone gradle plugin. For some reason, this *consistently* caused OOMs whenever we tried to run errorprone.  

Turns out for whatever reason, we were spinning up quite a few Gradle Workers and since out build is running in docker, 8 cpus (our resource allocation) wasn't being respected it seemed.

Bumping the circle JDK to 212 and setting `ActiveProcessorCount=8` has gotten things all the way through. There is still a high memory usage, but it seems to be handling it just find. We can either tune the processor count or the memory given to gradle if it becomes a problem in the future.

PR #3977, opened 2019-05-02
more extensive marking of empty value meter

gokcanozakdag:
**Goals (and why)**:
Goal is to instrument all the places that we do empty value filtering; so that we can pinpoint services that would benefit from moving to thorough sweep.

Marking the already existing `CellFilterMetrics.EMPTY_VALUE` meter whenever we remove empty values from raw results.

I haven't added any more testing, trusting the existing ones.

- Is there any other place that filtering takes place that I haven't covered

`SnapshotTransaction`


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -402,8 +402,21 @@ protected void checkGetPreconditions(TableReference tableRef) {
...
-        // Filter empty columns.
-        return Iterators.filter(mergedIterator, entry -> entry.getValue().length > 0);
+
+        return filterDeletedValues(mergedIterator, tableRef);
+    }
+
+    private Iterator<Map.Entry<Cell, byte[]>> filterDeletedValues(
+            Iterator<Map.Entry<Cell, byte[]>> unfiltered,
+            TableReference tableReference) {
+        Meter emptyValueMeter = getMeter(AtlasDbMetricNames.CellFilterMetrics.EMPTY_VALUE, tableReference);
```
felixdesouza: do you want to track global empty value filtering? or per transaction?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -852,16 +871,30 @@ protected static int getRequestHintToKvStore(int userRequestedSize) {
...
                 isReverse ? ordering.reverse() : ordering,
                 from -> RowResults.merge(from.lhSide, from.rhSide)); // prefer local writes
-        return RowResults.filterDeletedColumnsAndEmptyRows(mergeIterators);
+
+        Iterator<RowResult<byte[]>> purgeDeleted = Iterators.transform(mergeIterators,
+                createFilterColumnValues(tableReference));
+        return Iterators.filter(purgeDeleted, Predicates.not(RowResults.<byte[]>createIsEmptyPredicate()));
+    }
+
+    private Function<RowResult<byte[]>, RowResult<byte[]>> createFilterColumnValues(TableReference tableReference) {
```
felixdesouza: nit: kinda weird, can we just have a method? i.e. 

```java
private RowResult<byte[]> createFilterColumnValues(
    TableReference tableReference, 
    RowResult<byte[]> unfiltered) {
  ...
}

File: atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java (hunk 1 above)
Comment (gokcanozakdag) [reply atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java]: per table should be fine (at least for now)

File: atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java [hunk 3]
```diff
@@ -852,16 +871,30 @@ protected static int getRequestHintToKvStore(int userRequestedSize) {
...
+        return unfilteredRow -> {
+            SortedMap<byte[], byte[]> filteredColumns =
+                    Maps.filterValues(unfilteredRow.getColumns(), Predicates.not(Value.IS_EMPTY));
```
felixdesouza: out of curiosity, perhaps for later, is it worth unifying `Value.IS_EMPTY` and `Value::isTombstone`? i.e. picking one over the other, feels like the former is pre-java8 style

also, why do we always deal with `byte[]` here, as opposed to some useful safe wrapper like `Value`? This isn't really relevant to this PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -510,7 +523,10 @@ private void finishConsumingPreviousRow(PeekingIterator<Map.Entry<Cell, Value>>
                                                                   Map<Cell, Value> rawResults,
                                                                   ImmutableMap.Builder<Cell, byte[]> result) {
         getWithPostFiltering(tableRef, rawResults, result, Value.GET_VALUE);
+        Map<Cell, byte[]> withDeletedValues = result.build();
```
felixdesouza: code reuse between here and https://github.com/palantir/atlasdb/pull/3977/files#diff-bf062bd4c13b8c0022080ca8ef49c4d2R586

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: 1. I'd prefer the latter as long as it isn't used elsewhere outside of our control.
2. Probably because `Value` has a timestamp which a good chunk of the code here isn't directly interested in?

felixdesouza: 1) I had a quick look in PG and it's not.
2) fair enough, I suppose it's the raw byte arrays that make my eyes crawl aha, but it's okay

gokcanozakdag: moved everything to use `Value::isTombstone`, and removed `Value.IS_EMPTY`

PR #3987, opened 2019-05-07
code configurable grab immutable ts lock

gokcanozakdag:
**Goals (and why)**:
While migrating to thorough sweep, we need to have an intermediate step where we sweep conservatively but also grab immutable timestamp lock on read only transactions. Locking immutable ts option was introduced as a runtime config before to be able to see the perf effects. We would like to make it code configurable also, so that we can rely on that when we actually move to thorough (relying on a runtime configurable option doesn't seem safe).

Adding a builder option to TransactionManagers that overwrites the behaviour of runtime flag (if set to true)

Haven't introduced any testing. Probably would be better to find some place to add them though.

I have used a cache to avoid building a new `TransactionConfig` object every time `Supplier.get()` is called. Is it an overkill?


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -373,6 +382,15 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                 targetedSweep.singleAttemptCallback(),
                 asyncInitializationCallback());
 
+        LoadingCache<TransactionConfig, TransactionConfig> transactionConfigLoadingCache = Caffeine.newBuilder()
+                .maximumSize(1)
+                .build(this::withOverwrittenGrabImmutableTsLockFlag);
+
+        com.google.common.base.Supplier<TransactionConfig> transactionConfigSupplier = () -> {
```
gmaretic: There is an open PR to get rid of guava suppliers

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -373,6 +382,15 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                 targetedSweep.singleAttemptCallback(),
                 asyncInitializationCallback());
 
+        LoadingCache<TransactionConfig, TransactionConfig> transactionConfigLoadingCache = Caffeine.newBuilder()
```
gmaretic: This does seem like an overkill. You can just have a simple static inner class that maintains a reference to the last observed config, and do caching yourself

jeremyk-91: Hmm, unfortunately yeah, this stuff has been requested in several places - https://github.com/google/guava/issues/872 - but there doesn't seem to be nice library support

PR #3988, opened 2019-05-07
Resilient health check

felixdesouza:
**Goals (and why)**:
If for some reason a timelock node is unresponsive for longer than 20s (which is the deployment infrastructures timeout for health checks. We would fail the health check despite being able to find a leader.

We fire off the requests in parallel, and have a strict (and arbitrary) deadline of 7 seconds.

Relied on previous tests. If we really wanted to be fancy, we could do some stuff with deterministic scheduler, but unclear whether it's worth it/feels like we're just testing whether `PaxosQuorumChecker` works or not.

Have I bastardised `PaxosQuorumChecker` lol, how do we feel about the fixed thread pool for the health check?

Everywhere

ASAP

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/timelock/paxos/LeaderPingHealthCheck.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/paxos/LeaderPingHealthCheck.java
@@ -15,38 +15,49 @@
  */
 package com.palantir.timelock.paxos;
 
-import java.util.Map;
+import java.time.Duration;
 import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
+import java.util.concurrent.ExecutorService;
 
+import com.google.common.collect.ImmutableList;
+import com.palantir.common.concurrent.NamedThreadFactory;
+import com.palantir.common.concurrent.PTExecutors;
 import com.palantir.leader.PingableLeader;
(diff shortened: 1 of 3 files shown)


timelock-agent/src/main/java/com/palantir/timelock/paxos/LeaderPingHealthCheck.java
```diff
@@ -15,38 +15,49 @@
...
 
 public class LeaderPingHealthCheck {
-    private enum PingResult {
-        SUCCESS,
-        FAILURE,
-        EXCEPTION
-    }
 
-    private Set<PingableLeader> leaders;
+    private static final Duration HEALTH_CHECK_TIME_LIMIT = Duration.ofSeconds(7);
```
gokcanozakdag: out of curiosity, why 7 seconds :)

sandorw:
One concern with this change is it seems we won't have any alerting around hosts being down. Right now it's not guaranteed (seems to depend on ordering of host pings), but there was instance this morning where 1/3 Timelock hosts were down, and this is the only check that caused any alerting to go through.

gmaretic:
Can we instead collect "quorum" where quorum is all nodes, and then report a special status when we have the leader, but not all nodes responded?

PR #3992, opened 2019-05-09
disable background sweep if lock immutable ts is set to true

gokcanozakdag:
**Goals (and why)**:
If we decide to move a service to use thorough sweep; we need to make sure that background sweep won't cause any trouble by deleting large number of empty values at once - causing Cassandra OOMs.

Don't run background sweep process if `lockImmutableTsOnReadOnTransactions` is set to true. Sweep endpoint will still be functional.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -439,7 +439,7 @@ public abstract class TransactionManagers {
                         follower,
                         instrumentedTransactionManager,
                         persistentLockManager,
-                        runBackgroundSweep()),
+                        runBackgroundSweepProcess()),
                 closeables);
         initializeCloseable(
                 initializeCompactBackgroundProcess(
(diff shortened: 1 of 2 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -438,7 +438,8 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                         sweepStrategyManager,
                         follower,
                         instrumentedTransactionManager,
-                        persistentLockManager),
+                        persistentLockManager,
+                        runBackgroundSweep()),
```
jeremyk-91: nit: `runBackgroundSweepProcess()`, to avoid confusion with sweep being enabled?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -452,6 +453,10 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         return instrumentedTransactionManager;
     }
 
+    private boolean runBackgroundSweep() {
+        return !lockImmutableTsOnReadOnlyTransactions();
```
jeremyk-91: nit: could we have a comment here explaining why we want it to be this way?

PR #3658, opened 2018-11-26
 Use correct bounds for InMemoryKeyValueService column range

pkoenig10:
**Goals (and why)**:

When calling `InMemoryKeyValueService#getRowsColumnRange` with a column selection in which `startCol` equals `endCol` the call would throw the following exception:
```
java.lang.IllegalArgumentException: inconsistent range
    at java.util.concurrent.ConcurrentSkipListMap$SubMap.<init>(ConcurrentSkipListMap.java:2603)
    at java.util.concurrent.ConcurrentSkipListMap.subMap(ConcurrentSkipListMap.java:2061)
    at java.util.concurrent.ConcurrentSkipListMap.subMap(ConcurrentSkipListMap.java:2097)
    at com.palantir.atlasdb.keyvalue.impl.InMemoryKeyValueService.getColumnRangeForSingleRow(InMemoryKeyValueService.java:344)
    at com.palantir.atlasdb.keyvalue.impl.InMemoryKeyValueService.lambda$getRowsColumnRange$2(InMemoryKeyValueService.java:323)
    at com.google.common.collect.Iterators$6.transform(Iterators.java:788)
    at com.google.common.collect.TransformedIterator.next(TransformedIterator.java:47)
    at com.google.common.collect.Iterators$ConcatenatedIterator.hasNext(Iterators.java:1340)
    at com.palantir.atlasdb.keyvalue.impl.LocalRowColumnRangeIterator.hasNext(LocalRowColumnRangeIterator.java:34)
    at com.palantir.atlasdb.transaction.impl.SnapshotTransaction.getRowsColumnRange(SnapshotTransaction.java:380)
    at com.palantir.atlasdb.transaction.impl.SerializableTransaction.getRowsColumnRange(SerializableTransaction.java:207)
    ...
``` 

This is because calling `RangeRequests#previousLexicographicName` will result in `rowEnd` being less than `rowBegin`.

**Implementation Description (bullets)**:

Using an inclusive `toKey` is incorrect when calling this variant of `NavigableMap#subMap`. This PR changes the implementation to use the correct exclusive key, which resolves the above issue.

**Testing (What was existing testing like?  What have you done to improve it?)**:

Updated key value services tests to test a column range request when `startCol` equals `endCol`.

PR #3993, opened 2019-05-09
[PDS-87779] Throw early on finding a backwards column range

jeremyk-91:
**Goals (and why)**:
- See PDS-87779
- Generally, there are use cases where if you make a `ColumnRangeSelection` that is backwards (e.g. `[tom, jeremy)`) you get a cryptic error message from the underlying KVS, e.g. `InvalidRequestException: null`. We should fail more loudly and more clearly.

- Add a preconditions check that the start to end forms a valid range.

Existing tests were very limited. I have added quite a few cases.

Nothing in particular.

**Where should we start reviewing?**: ColumnRangeSelectionTest probably

**Priority (whenever / two weeks / yesterday)**: today or tomorrow please

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
@@ -90,7 +90,8 @@ public class ColumnRangeSelection implements Serializable {
         return Joiner.on(',').join(ImmutableList.of(start, end));
     }
 
-    private static boolean isValidRange(byte[] startCol, byte[] endCol) {
-        return RangeRequests.isValidRange(false, startCol, endCol);
+    public static boolean isValidRange(byte[] startCol, byte[] endCol) {
+        return RangeRequests.isContiguousRange(false, startCol, endCol) &&
+                !RangeRequests.isExactlyEmptyRange(startCol, endCol);
     }
 }
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
@@ -90,7 +90,7 @@ public class ColumnRangeSelection implements Serializable {
         return Joiner.on(',').join(ImmutableList.of(start, end));
     }
 
-    private boolean isValidRange(byte[] startCol, byte[] endCol) {
+    private static boolean isValidRange(byte[] startCol, byte[] endCol) {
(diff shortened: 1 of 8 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
```diff
@@ -84,4 +89,8 @@ public String toString() {
         String end = PtBytes.encodeBase64String(endCol);
         return Joiner.on(',').join(ImmutableList.of(start, end));
     }
+
+    private static boolean isValidRange(byte[] startCol, byte[] endCol) {
```
jkozlowski: Can you expose this as a method, same as in the RangeRequest? Or delete it from both and then I'll happily use RangeRequests.isValidRange directly.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/ColumnRangeSelectionTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void canCreateUnboundedOneEnd() {
+        assertThatCode(() -> new ColumnRangeSelection(BYTES_1, PtBytes.EMPTY_BYTE_ARRAY)).doesNotThrowAnyException();
+        assertThatCode(() -> new ColumnRangeSelection(PtBytes.EMPTY_BYTE_ARRAY, BYTES_2)).doesNotThrowAnyException();
+    }
+
+    @Test
+    public void canCreateEmptyRange() {
+        assertThatCode(() -> new ColumnRangeSelection(BYTES_1, BYTES_1)).doesNotThrowAnyException();
```
gmaretic: This should be an invalid column range, most notably since it would throw on Cassandra; see #3658

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/ColumnRangeSelection.java
jeremyk-91: Will do. It turns out the logic here is not equivalent (if you see @gmaretic's comment below)

jeremyk-91: so I'll expose this, and we should probably stick with that.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/keyvalue/impl/ColumnRangeSelectionTest.java
jeremyk-91: good catch!

PR #3996, opened 2019-05-10
add user agent parameter to proxy creator

gokcanozakdag:
**Goals (and why)**:
In recent refactors, we removed number of redundant methods from ``AtlasDbHttpClients``. Our internal backup services was using one of them - and now there isn't a way to specify userAgent for `createProxyWithFailover`. Given that we are reducing number of methods, I think it would make more sense to keep more capable ones, rather than restricted.

Add userAgent parameter to `createProxyWithFailover` method, and fix internal usage of the method.


felixdesouza:
looks good!

PR #3997, opened 2019-05-10
Increase default number of targeted sweep shards

gmaretic:
**Goals (and why)**:
This enables us to increase parallelism if necessary to process existing entries faster, as opposed to only being able to do so for future writes.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/config/TargetedSweepRuntimeConfig.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/config/TargetedSweepRuntimeConfig.java
@@ -42,7 +42,7 @@ public abstract class TargetedSweepRuntimeConfig {
      */
     @Value.Default
     public int shards() {
-        return AtlasDbConstants.DEFAULT_SWEEP_QUEUE_SHARDS;
+        return 8;
     }
 
     @Value.Check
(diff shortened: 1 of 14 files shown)


gmaretic:
@tpetracca for SA: in the future we will be able to speed up targeted sweep

atlasdb-config/src/main/java/com/palantir/atlasdb/config/TargetedSweepRuntimeConfig.java
```diff
@@ -42,7 +42,7 @@ public boolean enabled() {
      */
     @Value.Default
     public int shards() {
-        return AtlasDbConstants.DEFAULT_SWEEP_QUEUE_SHARDS;
```
tpetracca: does this constant have any other references? should we delete it from that class?

gmaretic: It is used in tests and, more importantly, as the default value when reading the persisted number of shards and no value is present -- we take the min of this and the runtime config.

gmaretic:
Intentionally did not increase the number of threads, so we have wiggle room to increase speed when necessary. Otherwise, by the next time we realise we are falling behind it will again be too late. (Note that by increasing the number of shards, we are actually increasing parallelism already as we have one thread per host doing the sweeping)

atlasdb-config/src/main/java/com/palantir/atlasdb/config/TargetedSweepRuntimeConfig.java
jeremyk-91: maybe leave a comment as to why it's not the default? Logic makes sense, but it's still a bit surprising

PR #4031, opened 2019-05-13
[_transactions2, Part 48] Remove NPEs in MetadataCoordinationServiceMetrics

jeremyk-91:
**Goals (and why)**:
- Fix NPEs that were showing up in MetadataCoordinationServiceMetrics

- Actually handle the null case (which can happen if you haven't read from the store yet) nicely

... not much.

- Is there a nicer way to do this?

**Where should we start reviewing?**: `MetadataCoordinationServiceMetrics`

**Priority (whenever / two weeks / yesterday)**: this week. Internal reference 132053 on large internal product.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/metrics/MetadataCoordinationServiceMetrics.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/metrics/MetadataCoordinationServiceMetrics.java
@@ -42,9 +42,9 @@ public final class MetadataCoordinationServiceMetrics {
 
     /**
      * Registers metrics that may be useful for diagnosing the status of a {@link CoordinationService}.
-     *
-     * Only one coordination service should be registered to a single metrics manager - if multiple are registered,
-     * then it is non-deterministic as to which coordination service the metrics being reported are referring to.
+     * <p>
+     * Only one coordination service should be registered to a single metrics manager - if multiple are registered, then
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/metrics/MetadataCoordinationServiceMetrics.java
```diff
@@ -129,10 +129,10 @@ private static void registerMetricForTransactionsSchemaVersionAtTimestamp(Metric
...
                         Clock.defaultClock(),
                         metricName,
-                        () -> {
-                            ValueAndBound<TimestampPartitioningMap<Integer>> mapAndBound = timestampMapSupplier.get();
-                            return getVersionAtTimestamp(mapAndBound.value(), timestampQuery.apply(mapAndBound));
-                        }));
+                        () -> timestampMapSupplier.get()
+                                .map(mapAndBound -> getVersionAtTimestamp(
+                                        mapAndBound.value(), timestampQuery.apply(mapAndBound)))
+                                .orElse(null)));
```
gokcanozakdag: I was concerned about the null value here also, but seems like returning null here is safe.

jeremyk-91: Yep. For metrics null basically means don't plot a data point / don't return a value

PR #4034, opened 2019-05-15
Unrevert supplier change revert

gmaretic:
Reverts palantir/atlasdb#3995


felixdesouza:
Technically, isn't this a revert?

PR #4036, opened 2019-05-15
Timelock Auth Part 2: Introducing Authenticator

gokcanozakdag:
**Goals (and why)**:
To add authentication capability to timelock.

Timelock clients will be passing a client id and a password to be authenticated. Passwords are stored as `BCryptedSecret` on server side (timelock server). Authenticator is responsible for checking the client id and password and returning a `Client` object that would be used by `Authorizer`.

A caching layer is added in between checking the password and returning the respective `Client` object, as password checks using `BCrypt` are expensive (and this will be living in a very hot code path)

A few tests are added to test basic functionality.

`Authenticator` returns `Client.ANONYMOUS` if it is not aware of the client yet; rather then throwing `ForbiddenException`. This is to prevent problems if clients start to send auth headers before timelock being aware of them (if runtime configuration of timelock is not updated for a while for example).

`Authenticator`

This week please


timelock-impl/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     }
 
     compile group: 'com.palantir.safe-logging', name: 'safe-logging'
+    compile group: 'org.mindrot', name: 'jbcrypt'
```
felixdesouza: lol `mindrot`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authenticator.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.auth.api;
+
+public interface Authenticator {
+    Client authenticate(String id, Password password);
```
felixdesouza: from our earlier discussion: `AuthenticatedClient`, `ClientId`, `Password`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/BCryptedSecret.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @JsonCreator
+    public static BCryptedSecret of(String hashedSecret) {
+        return ImmutableBCryptedSecret.builder()
+                .hashedSecret(hashedSecret)
+                .build();
+    }
+
+    public static BCryptedSecret forPassword(Password password) {
```
felixdesouza: when would we use this? it feels like this is just for tests, somewhat skeptical

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/CachingAuthenticator.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.palantir.atlasdb.timelock.auth.api.Authenticator;
+import com.palantir.atlasdb.timelock.auth.api.BCryptedSecret;
+import com.palantir.atlasdb.timelock.auth.api.Client;
+import com.palantir.atlasdb.timelock.auth.api.Password;
+
+public class CachingAuthenticator implements Authenticator {
+    private final Map<String, BCryptedSecret> credentials;
```
felixdesouza: let's make this an `ImmutableMap`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/CachingAuthenticator.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private final LoadingCache<ClientCredentials, Optional<Client>> cache;
+
+    CachingAuthenticator(Map<String, BCryptedSecret> credentials) {
```
felixdesouza: I can never decide, but do we want to separate out the caching layer and the actual auth layer? I'm not fussed either way, just wanted to bring it up.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/Authenticator.java
felixdesouza: Although the anonymous client seems to break your `AuthenticatedClient`, but this is non blocking

felixdesouza: Still think we should probably do `Optional<Client>` and handle all the web exception mapper stuff in the filter/proxy/resource code we're adding.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/impl/CachingAuthenticatorTest.java
```diff
@@ -0,0 +1,74 @@
...
+        Authenticator cachingAuthenticator = CachingAuthenticator.create(ImmutableMap.of(
+                CLIENT_1, bcrypted(PASSWORD_1),
+                CLIENT_2, bcrypted(PASSWORD_2)));
+
+        assertThatThrownBy(() -> cachingAuthenticator.authenticate(CLIENT_1, PASSWORD_2))
+                .isInstanceOf(ForbiddenException.class);
+    }
+
+    private static BCryptedSecret bcrypted(Password password) {
+        return BCryptedSecret.forPassword(password);
```
felixdesouza: let's not add `forPassword` since it's just for tests. Inline it here, it's super trivial to do so.

felixdesouza:
What happens regarding live reload? both for the `Authorizer` _and_ the `Authenticator`

gokcanozakdag:
I am thinking of something like an `AuthManager` that encapsulates everything related to auth, to be used in TimelockResource. That class will handle live reloading - we will be creating new `Authorizer` and `Authenticator` when we detect a change in runtime config - here is a draft pr for that: https://github.com/palantir/atlasdb/pull/4043

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/NamespaceLocker.java
```diff
@@ -18,17 +18,17 @@
...
-import com.palantir.atlasdb.timelock.auth.api.Client;
 
 interface NamespaceLocker {
     NamespaceLocker ALL_LOCKED = ignored -> true;
     NamespaceLocker NONE_LOCKED = ignored -> false;
 
     boolean isLocked(TimelockNamespace namespace);
 
-    static NamespaceLocker deriveFromPrivileges(Map<Client, Privileges> privileges) {
+    static NamespaceLocker deriveFromPrivileges(Map<AuthenticatedClient, Privileges> privileges) {
```
felixdesouza: since this comes from config, you probably want this to be `ClientId`.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -23,34 +23,34 @@
...
+    public static Authorizer of(Map<AuthenticatedClient, Privileges> privileges, AuthRequirement authRequirement) {
         NamespaceLocker namespaceLocker = createNamespaceLocker(authRequirement, privileges);
         return new SimpleAuthorizer(
                 ImmutableMap.copyOf(privileges),
                 namespaceLocker);
     }
 
     @Override
-    public boolean isAuthorized(Client client, TimelockNamespace namespace) {
+    public boolean isAuthorized(AuthenticatedClient authenticatedClient, TimelockNamespace namespace) {
```
felixdesouza: let `AuthenticatedClient` contain `ClientId`, and then you can do `authenticatedClient.id()`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/SimpleAuthorizer.java
```diff
@@ -23,34 +23,34 @@
 import com.palantir.atlasdb.timelock.auth.api.Privileges;
 import com.palantir.lock.TimelockNamespace;
 import com.palantir.atlasdb.timelock.auth.api.Authorizer;
-import com.palantir.atlasdb.timelock.auth.api.Client;
+import com.palantir.atlasdb.timelock.auth.api.AuthenticatedClient;
 
 public class SimpleAuthorizer implements Authorizer {
-    private final Map<Client, Privileges> privileges;
+    private final Map<AuthenticatedClient, Privileges> privileges;
```
felixdesouza: `AuthenticatedClient` -> `ClientId`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/api/BCryptedSecret.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+@Value.Immutable
+public abstract class BCryptedSecret {
+    private static final Logger log = LoggerFactory.getLogger(BCryptedSecret.class);
+
+    abstract String getHashedSecret();
```
jeremyk-91: nit: just `hashedSecret()`? `getHashedSecret()` should be auto-generated from that for you, if you want this specific wording

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/CachingAuthenticator.java
```diff
@@ -0,0 +1,75 @@
...
+    private final ImmutableMap<String, BCryptedSecret> credentials;
+
+    private final LoadingCache<ClientCredentials, Optional<AuthenticatedClient>> cache;
+
+    CachingAuthenticator(Map<String, BCryptedSecret> credentials) {
+        this.credentials = ImmutableMap.copyOf(credentials);
+        this.cache = Caffeine.newBuilder()
+                .maximumSize(1000)
+                .build(this::authenticateInternal);
+    }
```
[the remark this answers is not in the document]
gokcanozakdag: wanted to handle live reloading outside of this class. Part 4 handles that by introducing an AuthManager, which creates a fresh Authenticator and Authorizer when it detects a change in runtime config.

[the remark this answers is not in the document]
jeremyk-91: OK. I guess changes are not common, so this is fine (possibly simpler than computing the diff and invalidating diffed entries).

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/impl/PrivilegeBasedNamespaceLocker.java
```diff
@@ -0,0 +1,40 @@
...
+
+    PrivilegeBasedNamespaceLocker(Map<ClientId, Privileges> privileges) {
+        this.nonAdminPrivileges = privileges.values().stream()
+                .filter(privilege -> privilege != Privileges.ADMIN)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public boolean isLocked(TimelockNamespace namespace) {
+        return nonAdminPrivileges.stream().anyMatch(privilege -> privilege.hasPrivilege(namespace));
```
jeremyk-91: I'm probably forgetting something, but I think we had some resolution somewhere as to how we planned on ensuring this isn't a hot code-path?

gokcanozakdag: Yes, I will be adding a `CachingAuthorizer` (and removing the simple one) but wanted to do this in a separate pr.

PR #4037, opened 2019-05-15
[PDS-89716] Finite Work Queues for Delete Executors

jeremyk-91:
**Goals (and why)**: see PDS-89716

- Include a bounded length work queue for delete executors

- Added one test that the queue doesn't get too long

- Should this just silently reject the elements (`DiscardPolicy`)?

**Where should we start reviewing?**: DefaultTaskExecutors


docs/source/release_notes/release-notes.rst
```diff
@@ -60,6 +60,13 @@ develop
          - Coordination service metrics no longer throw ``NullPointerException`` when attempting to read the metric value before reading anything from the coordination store.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4031>`__)
 
+    *    - |fixed|
+         - AtlasDB now maintains a finite length for the delete executor's work queue, to avoid OOMs on services with high conflict rates for transactions.
+           In the event the queue length is reached, we will not proactively schedule cleanup of values written by a transaction that was rolled back.
+           Note that it is not essential that these deletes are carried out immediately, as targeted sweep will eventually clear them out.
+           Previously, this queue was unbounded, meaning that service nodes could end up using lots of memory.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/4QQQ>`__)
```
felixdesouza: don't forget to update this

felixdesouza:
It could reject them, assuming that another node that runs a transaction could potentially clean them up if need be - assuming that that cell gets hit again of course. Nothing actually went wrong with the transaction here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1736,7 +1737,15 @@ private boolean rollbackFailedTransactions(
             }
         }
 
-        deleteExecutor.submit(() -> deleteCells(tableRef, keysToDelete));
+        try {
+            deleteExecutor.submit(() -> deleteCells(tableRef, keysToDelete));
```
gokcanozakdag: It may help to batch these delete calls. We can union "keys to delete maps" into a single batch before sending these to C* - would be helpful if we submit these calls with a very high rate. We can address that in a separate pr maybe?

felixdesouza: you wouldn't necessarily want to put them all in the same batch would you? a la our lock request too big etc., but yeah batching seems reasonable

jeremyk-91: Yes, autobatcher would make sense (though I guess not for this PR, as we want to get this out quickly)

PR #4043, opened 2019-05-20
Timelock Auth Part 4: Introducing AuthManager

gokcanozakdag:
**Goals (and why)**:
Thoughts on live reloading auth configuration.

Using `PollingRefreshable` to  check updates on configuration.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/AuthManager.java
```diff
@@ -0,0 +1,125 @@
...
+        AuthServices authServices = getAuthServices(timelockAuthConfigurationSupplier.get());
+
+        return new AuthManager(
+                configPollingRefreshable.getRefreshable(),
+                configPollingRefreshable::close,
+                authServices);
+    }
+
+    public void checkAuthorized(String clientId, Password password, TimelockNamespace timelockNamespace) {
+        update();
```
felixdesouza: `update` needs to return `authServices` otherwise it's not thread-safe.

felixdesouza: / consistent

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/AuthManager.java
```diff
@@ -0,0 +1,125 @@
...
+    public void close() {
+        closingCallback.run();
+    }
+
+    private void update() {
+        authConfigurationRefreshable.getAndClear().ifPresent(this::updateInternal);
+    }
+
+    private synchronized void updateInternal(TimelockAuthConfiguration authConfiguration) {
+        authServices = getAuthServices(authConfiguration);
```
felixdesouza: doesn't need to be synchronized anymore, also needs to set and return `authServices`

felixdesouza: might be convenient to be use `AtomicReference` here. `getAndSet`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/AuthManager.java
```diff
@@ -0,0 +1,125 @@
...
+                .authenticator(newAuthenticator)
+                .authorizer(newAuthorizer)
+                .useAuth(authConfiguration.useAuth())
+                .build();
+    }
+
+    private static Authorizer getAuthorizer(TimelockAuthConfiguration authConfiguration) {
+        Map<AuthenticatedClient, Privileges> privilegesMap = authConfiguration.privileges().stream()
+                .collect(Collectors.toMap(p -> AuthenticatedClient.create(p.id()), PrivilegesConfiguration::privileges));
+        return SimpleAuthorizer.of(privilegesMap, AuthRequirement.PRIVILEGE_BASED);
```
felixdesouza: extract out `AuthRequirement.PRIVLEGE_BASED` into a constant explaining that this is what makes timelock 0.x, 1.x etc.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/AuthManager.java
```diff
@@ -0,0 +1,125 @@
...
+    AuthManager(Refreshable<TimelockAuthConfiguration> authConfigurationRefreshable,
+            Runnable closingCallback,
+            AuthServices authServices) {
+        this.authConfigurationRefreshable = authConfigurationRefreshable;
+        this.closingCallback = closingCallback;
+        this.authServices = authServices;
+    }
+
+    public static AuthManager of(Supplier<TimelockAuthConfiguration> timelockAuthConfigurationSupplier) {
+        PollingRefreshable<TimelockAuthConfiguration> configPollingRefreshable = PollingRefreshable
```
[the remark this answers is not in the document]
gokcanozakdag: we will need to have a way of updating `Authenticator` and `Authorizer` only if supplied config has changed. That can be either done with something like `PollingRefreshable` (which enable non-blocking updates especially after https://github.com/palantir/atlasdb/pull/4057) or something like a `CachingComposedSupplier`.

PR #4046, opened 2019-05-20
Add live reloadable config for TS pause duration, reduce the default value

gmaretic:
**Goals (and why)**:
Currently, TS processes batches that cover 50k timestamps, and then the thread pauses for 5 seconds. If we have one thread per shard, that means that the maximum throughput (provided there are entries in every 50k timestamp batch) is 10k timestamps per second.

We have already encountered a case where a service is churning through 30k timestamps per second, and are falling behind as a result. This can be remediated by throwing more threads at the problem, but the real solutions are to 1) wait less, 2) enable processing batches of larger size.

This implements 1) from above.

Mostly plumbing

ASAP, as we want to get this deployed to the product that's falling behind.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
```diff
@@ -51,6 +51,11 @@ void checkShardSize() {
                 "Shard number must be between 1 and 256 inclusive, but it is %s.", shards());
     }
 
+    @Value.Default
+    public long pauseMillis() {
+        return 500L;
```
felixdesouza: do you want to keep the default the same, adjust on particular deployment and change default if it's not too aggressive?

felixdesouza: not too fussed either way

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepInstallConfig.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
```
felixdesouza: hahaha

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
gmaretic: I think the default is too high for no real reason

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -211,7 +229,8 @@ private void scheduleBackgroundThreads() {
                 executorService = PTExecutors
                         .newScheduledThreadPoolExecutor(numThreads, new NamedThreadFactory("Targeted Sweep", true));
                 for (int i = 0; i < numThreads; i++) {
-                    executorService.scheduleWithFixedDelay(this::runOneIteration, 1, 5, TimeUnit.SECONDS);
+                    executorService.scheduleWithFixedDelay(this::runOneIteration, 1000,
+                            Math.max(runtime.get().pauseMillis(), 1L), TimeUnit.MILLISECONDS);
```
jeremyk-91: nit: this does enforce reasonability, though you might want to include a `Value.Check` in the immutable class itself (that `pauseMillis() > 0` or `>= 0` depending on taste)

gmaretic: Yeah, I just don't want to error out if someone wants to disable pausing and puts in 0, which is quite a reasonable thing to do IMO.

PR #4049, opened 2019-05-20
Do not delegate.

felixdesouza:
**Goals (and why)**:
Not every method should be delegated i.e. deprecated default methods that "delegate" to the non-deprecated method on the same interface. Doing so, can result in weird correctness issues since implementors of the generated `AutoDelegate_*` interfaces are not forced to implement everything (problematic).

Add a `@DoNotDelegate` annotation that tells the processor to skip generating code to delegate that method.

Somewhat difficult to test since we do all this as part of our build, as opposed to a test, so compilation proves it to be okay, I suppose.

### DIFF
--- a/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
+++ b/atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
@@ -15,14 +15,13 @@
  */
 package com.palantir.processors;
 
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.empty;
-import static org.hamcrest.Matchers.hasItems;
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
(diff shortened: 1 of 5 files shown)


atlasdb-processors-tests/src/test/java/com/palantir/processors/AutoDelegateInterfaceTests.java
```diff
@@ -32,80 +31,109 @@
...
         TestInterfaceImpl mockImpl = mock(TestInterfaceImpl.class);
-        AutoDelegate_TestInterface instanceOfInterface = () -> mockImpl;
+        AutoDelegate_TestInterface instanceOfInterface = new AnImpl(mockImpl);
 
         instanceOfInterface.methodWithReturnType();
         verify(mockImpl, times(1)).methodWithReturnType();
     }
+
+    private static Set<String> doNotDelegateMethods(Set<String> originalMethods) {
+        return Sets.filter(originalMethods, m -> m.contains("methodThatMustBeImplemented"));
```
gmaretic: kind of sketchy, but ok I guess

PR #4052, opened 2019-05-21
Close response body in the event of exceptions

gokcanozakdag:
**Goals (and why)**:
Stop leaking connections, see PDS-88834


felixdesouza:
changelog?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/ExceptionCountingRefreshingClient.java
```diff
@@ -55,8 +55,7 @@ public static Client createRefreshingClient(Supplier<Client> baseClientSupplier)
 
     @Override
     public Response execute(Request request, Request.Options options) throws IOException {
-        try {
-            Response response = delegate().execute(request, options);
+        try (Response response = delegate().execute(request, options)) {
```
gokcanozakdag: haha this doesn't work of course, as we are returning a closed response in any case

PR #4057, opened 2019-05-24
Add option to compose PollingRefreshable

gmaretic:
**Goals (and why)**:
Generally useful to have, will help wiring some CJR stuff more easily, although may end up not needed later.

### DIFF
--- a/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
+++ b/atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
@@ -18,6 +18,7 @@ package com.palantir.atlasdb.http;
 import java.time.Duration;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.slf4j.Logger;
@@ -32,11 +33,13 @@ import com.palantir.logsafe.UnsafeArg;
 
 /**
  * A PollingRefreshable serves as a bridge between a {@link Supplier} and {@link Refreshable}, polling for changes
- * in the value of the Supplier and, if detecting a change, pushing it to the linked Refreshable.
+ * in the value of the Supplier and, if detecting a change, pushing it to the linked Refreshable, after possibly
+ * applying a transform to the value.
  *
- * @param <T> type of the value supplied / pushed to the Refreshable
+ * @param <T> type of the value supplied
+ * @param <U> type of the Refreshable
  */
-public final class PollingRefreshable<T> implements AutoCloseable {
+public final class PollingRefreshable<T, U> implements AutoCloseable {
     @VisibleForTesting
     static final Duration DEFAULT_REFRESH_INTERVAL = Duration.ofSeconds(5L);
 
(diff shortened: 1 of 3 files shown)


atlasdb-feign/src/test/java/com/palantir/atlasdb/http/PollingRefreshableTest.java
```diff
@@ -181,7 +236,15 @@ public void cannotSetNegativeRefreshInterval() {
                     return newValue;
                 },
                 REFRESH_INTERVAL,
+                Function.identity(),
                 scheduler);
         return pollingRefreshable.getRefreshable();
     }
+
+    private long doubleOrThrowOnOne(long input) {
```
jeremyk-91: nit: `private static`, surely?

Also this is fine, but I think actually having the behaviour be more explicit e.g. via a mock might be more readable for people picking this up later

atlasdb-feign/src/test/java/com/palantir/atlasdb/http/PollingRefreshableTest.java
```diff
@@ -181,7 +236,15 @@ public void cannotSetNegativeRefreshInterval() {
...
+        if (input == 1L) {
+            throw new RuntimeException("ONE!!11!!2");
```
jeremyk-91: 💯

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
```diff
@@ -46,50 +49,58 @@
...
         this.refreshInterval = refreshInterval;
+        this.transform = transform;
         this.poller = poller;
 
         try {
             lastSeenValue = supplier.get();
-            refreshable.set(lastSeenValue);
+            refreshable.set(transform.apply(lastSeenValue));
         } catch (Exception e) {
             log.info("Exception occurred in supplier when trying to populate the initial value.");
```
jeremyk-91: nit: This is no longer correct (we might get an exception that was in the transform, not in the supplier). Also, probably best to include what the exception was.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
```diff
@@ -32,11 +33,13 @@
 
 /**
  * A PollingRefreshable serves as a bridge between a {@link Supplier} and {@link Refreshable}, polling for changes
- * in the value of the Supplier and, if detecting a change, pushing it to the linked Refreshable.
+ * in the value of the Supplier and, if detecting a change, pushing it to the linked Refreshable, after possibly
+ * applying a transform to the value.
```
jeremyk-91: I'd suggest adding a note here on what the semantics of transforms are. Specifically, the transform is computed once and its result is cached - so even if the transform somehow changes such that its output on the same input value from the supplier changes, new values will _not_ be propagated in that case.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
```diff
@@ -46,50 +49,58 @@
     private final Duration refreshInterval;
     private final ScheduledExecutorService poller;
 
-    private final Refreshable<T> refreshable = Refreshable.empty();
+    private final Function<T, U> transform;
+    private final Refreshable<U> refreshable = Refreshable.empty();
 
-    private T lastSeenValue;
+    private T lastSeenValue = null;
```
jeremyk-91: nit: OK, but not sure why this is needed.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
felixdesouza: i.e. it needs to be a pure fuction? 😁

jeremyk-91: Yeah, it should be a pure function - or if it's not, then the user should understand that it may not be recalculated!

jeremyk-91: Some non-pure functions would be ok if their state only evolves as a result of changes in the input value, though I can't think of any significant examples beyond contrived ones e.g. a function that ignores the input value, other than counting the number of occasions on which we saw that it changed.

gmaretic:
Effectively merged

PR #4058, opened 2019-05-29
[TimeLock Partitioning] Abdication Mechanism Part 1

jeremyk-91:
**Goals (and why)**:
- Provide a mechanism for a leader node to declare it is itself not the leader. If we're relying on randomness for timelock to balance itself, we want to have a mechanism for ensuring that we can recover (without manual action to bounce leader nodes)

- Implement a `stepDown()` method in `LeaderElectionService` that makes a leading node step down, by proposing a value at a randomized UUID which isn't actually a member of the cluster.
- Only nodes that believe they are the leader can attempt to step down.

- Added tests at the `LeaderElectionService` level (see `PaxosConsensusFastTests`).

- Is this mechanism safe?
- Should this be exposed?
- No integration tests. Plan is to do this in a Part 2; this isn't used anywhere yet.

**Where should we start reviewing?**: `LeaderElectionService`

### DIFF
--- a/leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
+++ b/leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.paxos;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
@@ -91,9 +92,7 @@ public class PaxosConsensusFastTest {
     @Test
     public void loseQuorum() {
         LeadershipToken token = state.gainLeadership(0);
-        for (int i = 1; i < NUM_POTENTIAL_LEADERS - QUORUM_SIZE + 2; i++) {
-            state.goDown(i);
-        }
+        knockOutQuorumNotIncludingZero();
         assertNotSame("leader cannot maintain leadership without quorum",
                 state.leader(0).isStillLeading(token), StillLeadingStatus.LEADING);
         state.comeUp(1);
@@ -143,9 +142,7 @@ public class PaxosConsensusFastTest {
         future.cancel(true);
         exec.shutdown();
         exec.awaitTermination(10, TimeUnit.SECONDS);
-        for (int i = 0; i < NUM_POTENTIAL_LEADERS; i++) {
-            state.comeUp(i);
-        }
+        restoreAllNodes();
     }
 
     @SuppressWarnings("EmptyCatchBlock")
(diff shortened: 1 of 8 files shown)


leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
```diff
@@ -189,5 +186,67 @@ public void learnerRecovery() {
...
+    }
+
+    @Test
+    public void otherNodeCanBecomeLeaderAfterSteppingDown() {
+        LeadershipToken token = state.gainLeadership(0);
+        assertThat(state.leader(0).isStillLeading(token)).isEqualTo(StillLeadingStatus.LEADING);
+        assertThat(state.leader(0).stepDown()).isTrue();
+
+        LeadershipToken token2 = state.gainLeadership(1);
+        assertThat(state.leader(1).isStillLeading(token2)).isEqualTo(StillLeadingStatus.LEADING);
```
felixdesouza: is it worth double checking that `leader(0)` is *not* leading?

leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
```diff
@@ -189,5 +186,67 @@ public void learnerRecovery() {
...
+    }
+
+    @Test
+    public void failToStepDownIfNoQuorum() {
+        LeadershipToken token = state.gainLeadership(0);
+        assertThat(state.leader(0).isStillLeading(token)).isEqualTo(StillLeadingStatus.LEADING);
+
+        knockOutQuorumNotIncludingZero();
+        assertThat(state.leader(0).stepDown()).isFalse();
+
```
felixdesouza: wouldn't this fail at the proposer step? why do you need to have a special condition inside PLES?

leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
jeremyk-91: Yep, can do - I thought this would be covered by the loseLeadershipAfterSteppingDown test, but fair enough - maybe something that makes 1 gain leadership also restores it for everyone

leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
jeremyk-91: This fails actually because you realise you're not the leader. Failing at the proposer bit would throw an exception - this was unfortunately a bit trickier to test, because it relies on a node coming down after you've confirmed you're the leader but before you initiate the Paxos instance.

As for the leadership check (I'm assuming that's the condition we're referring to) - that's because I didn't want a call to `stepDown()` on a non-leader to cause a leadership change in general.

PR #4061, opened 2019-05-30
[Timelock Partitioning] Part 2: Batch Paxos API classes.

felixdesouza:
**Goals (and why)**:
For the timelock partitioning, we're going to be running a Paxos instance per client. Whilst this would work out of the box, it does multiply the number of calls. So I'm opening up batch apis for paxos. These are just the endpoint definitions.

The hot path currently is `getLatestSequencePreparedOrAccepted`. We want to do something a bit smarter than just pass in every client that you're servicing to this call.

I've included two endpoints:
```
AcceptorCacheDigest latestSequencesPreparedOrAccepted(
            @HeaderParam(HttpHeaders.IF_MATCH) Optional<AcceptorCacheKey> cacheKey,
            Set<Client> clients);
``` 
and
```
Optional<AcceptorCacheDigest> latestSequencesPreparedOrAcceptedCached(
            @HeaderParam(HttpHeaders.IF_MATCH) AcceptorCacheKey cacheKey);
```

With the potential caching algorithm as follows:
```

currentCacheKey := null
latestSequences := emptyMap<client, seq>

// autobatcher consumer
void processBatch(set<client> clients) {
  if (currentCacheKey == null) {
    newCacheKey, updates := latestSequencesPreparedOrAccepted(Optional.empty(), clients);
    currentCacheKey = newCacheKey;
    seenClients += updates;
    resolveFutures(updates);
    return;
  }
  
  newClients := clients - latestSequences.keySet();
  if (newClients.isEmpty()) {
    try {
      newCacheKey, updates := latestSequencesPreparedOrAcceptedCached(currentCacheKey);
      // update cache key and sequences
    } catch (PreconditionFailedException e) {
      newCacheKey, updates := latestSequencesPreparedOrAccepted(Optional.empty(), latestSequences.keySet());
      // update cache key and sequences, resolve futures
    }
  }

  // we have new clients we haven't received back from server, force latest to 
  // include these clients
  try {
	newCacheKey, updates := latestSequencesPreparedOrAccepted(Optional.of(currentCacheKey, newClients);
  } catch (PreconditionFailedException e) {
      allClients := newClients + latestSequences.keySet();
      newCacheKey, updates := latestSequencesPreparedOrAccepted(Optional.empty(), allClients);
      // update cache key and sequences, resolve futures
  }
}
```
This should give us the following benefits:
* if you've just become the leader, you won't have a cache key, you'll fetch all clients you've gotten requests for
* if there have been no updates, you'll just use your cache key and get back an empty optional (fast path)
* if there have been updates, you'll get back a new cache key, along with the updates from the provided cache key
* if you're the leader, and the server you're talking to restarts, it wouldn't have seen your cache key (stored in memory), so you retry your request for all clients you're currently servicing

None

Do the caching apis provide full coverage? i.e. is there a case where we'll miss an update?

`BatchPaxosAcceptor`, `BatchPaxosLearner`

ASAP


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
```diff
@@ -0,0 +1,77 @@
...
+     * the learner learns the given {@link PaxosValue}.
+     * <p>
+     * @param paxosValuesByClientAndSeq {@link ClientAndSeq} identifies the client and the instance of paxos that is
+     * being taught; {@link PaxosValue} - for the above {@link ClientAndSeq} - is value being taught.
+     */
+    @POST
+    @Path("learn")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    void learn(Map<ClientAndSeq, PaxosValue> paxosValuesByClientAndSeq);
```
jeremyk-91: Not actionable: reading this was initially a bit strange to me (a `PaxosValue` already contains the `seq`), but it looks like we already do this in one-client `PaxosLearner` anyway so it seems fair to maintain consistency.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/Client.java
```diff
@@ -0,0 +1,76 @@
...
+        int cursor = 0;
+
+        for (; cursor < length; cursor++) {
+            if (!allowedCharacters.get(value().charAt(cursor))) {
+                break;
+            }
+        }
+
+        // Need at least one valid character
+        return cursor != 0;
```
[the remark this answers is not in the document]
jeremyk-91: Benchmarks:

```
Benchmark                                   Mode     Cnt      Score     Error  Units
RegexBenchmarks.bitSetHappyOne            sample  138907      0.058 ±   0.006  us/op
RegexBenchmarks.bitSetHappyTwo            sample  103660      0.136 ±   0.004  us/op
RegexBenchmarks.bitSetSuperSad            sample    2277   2194.203 ±  12.292  us/op
RegexBenchmarks.precompiledRegexHappyOne  sample  190927      0.152 ±   0.005  us/op
RegexBenchmarks.precompiledRegexHappyTwo  sample  152872      1.078 ±   0.019  us/op
RegexBenchmarks.precompiledRegexSuperSad  sample     191  26208.910 ± 525.809  us/op
```

Test strings: happy1 was `tom`, happy2 was `qwertyuiopasdfghjjklzxcvbnm1234567890_-_-_-_-_-eineineeineneinemeiner`, super-sad was `String.join("_", Collections.nCopies(1_000_000, "a")) + "."`

Yep, worth doing this.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AcceptorCacheDigest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableAcceptorCacheDigest.class)
+@JsonSerialize(as = ImmutableAcceptorCacheDigest.class)
+public interface AcceptorCacheDigest {
+    UUID newCacheKey();
+    Map<Client, Long> updates();
```
jeremyk-91: nit: might be good to define what happens if an acceptor receives new proposals at multiple sequence numbers for a given client? (Specifically, either in doc or parameter name, that it returns the latest in that case?)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptor.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+
+import com.palantir.paxos.BooleanPaxosResponse;
+import com.palantir.paxos.PaxosAcceptor;
+import com.palantir.paxos.PaxosPromise;
+import com.palantir.paxos.PaxosProposal;
+import com.palantir.paxos.PaxosProposalId;
+
+@Path("/batch/acceptor")
```
jeremyk-91: Do we want this to fall under part of the `INTERNAL_NAMESPACE`?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptor.java
```diff
@@ -0,0 +1,117 @@
...
+     * <p>
+     * If a valid {@code cacheKey} is provided, it will return all unseen sequences from when that {@code cacheKey} was
+     * issued. If the server has not prepared or accepted any sequences past that point, it will return
+     * {@code 204 No Content}.
+     * <p>
+     * In addition to the updates, a new {@code cacheKey} is provided to use on the next invocation of this method to
+     * minimise on payload size as this method is on the hot path.
+     *
+     * @param cacheKey
+     * @return {@code 204 No Content} if there is no update, a digest containing
```
jeremyk-91: nit: a digest containing updates plus a new cache key, or a 412 if the cache key is not valid

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Batch counterpart to {@link PaxosLearner#getLearnedValue}. For a given {@link Client} on paxos instance
+     * {@code seq}, it returns the learnt value. Values where nothing has been learnt are included as a separate field.
+     * <p>
+     * @param clientAndSeqs the {@link ClientAndSeq}'s - which identifies the client and the instance of paxos - to
+     * retrieve the learnt values for.
+     * @return for each {@link ClientAndSeq}, the {@link PaxosValue} learnt including {@link ClientAndSeq}'s where
+     * nothing was learnt.
```
jeremyk-91: This seems fair, though is there any reason we wanted to do it this way as opposed to just not including the keys where nothing was learnt? (a la LockService or KVS getRows())

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param seqLowerBoundsByClient for each {@link Client}, the lower bound for the seq-th paxos round to fetch all
+     * learnt paxos values since that paxos round.
+     * @return for each {@link Client}, all learnt {@link PaxosValue}'s past the given lower bound
+     */
+    @POST
+    @Path("learned-values-since")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    SetMultimap<Client, PaxosValue> getLearnedValuesSince(Map<Client, Long> seqLowerBoundsByClient);
+
```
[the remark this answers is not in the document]
felixdesouza: yep, isn't called over network, didn't want to include impl for it

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
felixdesouza: Hmm, just realised that I can't serialise a map of `ClientAndSeq` since it'd make it a complex key *weeps. Will have to change it to `SetMultimap<Client, SeqAndPaxosValue>` or as you said, just `SetMultimap<Client, PaxosValue>` since it already contains the sequence.

I'm going to make a class as follows:
```java
@Value.Immutable
interface Seq<T> {
  long seq();
  T value();
}
```

since we use it in a few places.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptor.java
felixdesouza: leaving it up to the `LeadershipResource`, I'm thinking we might be able to use this for the `clientPaxos` root at a later time.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
felixdesouza: yeah will do, I felt like I had a compelling reason, but can't remember it, plus we can always change back if it surfaces and is important

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
felixdesouza: With the above change where I can't have complex map keys, it might make sense to just have a `SetMultimap<Client, PaxosValue>`. I suppose it makes a bit more sense anyway: not sure if there is validation that `seq` and `paxosValue.seq` is the same.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/Client.java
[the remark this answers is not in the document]
felixdesouza: nice!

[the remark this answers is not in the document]
felixdesouza: alternatively, I suppose we could remove the check method, and just have a validator cache or something, means we only pay it "once" per unique client. With a Caffeine cache, where valid clients are weighed lower/higher (whatever biases away from eviction) than invalid clients. Not fussed either way.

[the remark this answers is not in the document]
felixdesouza: what do you propose?

PR #4062, opened 2019-05-30
Timelock Auth Part 6: Using suffixes to grant privileges

gokcanozakdag:
**Goals (and why)**:
Some services do need to have access to multiple namespaces. Declaring each namespace one by one would not work in these cases.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/config/ClientPrivilegesConfiguration.java
```diff
@@ -41,9 +43,12 @@
 
     abstract Set<TimelockNamespace> namespaces();
 
+    @JsonProperty("namespace-suffixes")
+    abstract Optional<Set<String>> namespaceSuffixes();
```
jeremyk-91: I guess unless you want to be able to know explicitly when this hasn't been set, it seems like specifying a default of `ImmutableSet.of()` rather than making the return type be an `Optional` would work a bit more cleanly.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/config/NamespacePrivilegesTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Test
+    public void privilegesAreGrantedCorrectly() {
+        Privileges namespacePrivileges = NamespacePrivileges.of(
+                ImmutableSet.of(NAMESPACE_1),
+                ImmutableSet.of(SUFFIX));
+
+        assertThat(namespacePrivileges.hasPrivilege(NAMESPACE_1)).isTrue();
+        assertThat(namespacePrivileges.hasPrivilege(TimelockNamespace.of("foo" + SUFFIX))).isTrue();
+
+        assertThat(namespacePrivileges.hasPrivilege(NAMESPACE_2)).isFalse();
```
jeremyk-91: maybe one more assertion: `namespacePrivileges.hasPrivilege(TimelockNamespace.of(SUFFIX + "-")).isFalse()`?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/config/ClientPrivilegesConfiguration.java
jeremyk-91: Also, should we ban the empty string from being configured as a suffix?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/config/NamespacePrivilegesTest.java
```diff
@@ -0,0 +1,44 @@
...
+    private static final String SUFFIX = "bar";
+
+    @Test
+    public void privilegesAreGrantedCorrectly() {
+        Privileges namespacePrivileges = NamespacePrivileges.of(
+                ImmutableSet.of(NAMESPACE_1),
+                ImmutableSet.of(SUFFIX));
+
+        assertThat(namespacePrivileges.hasPrivilege(NAMESPACE_1)).isTrue();
+        assertThat(namespacePrivileges.hasPrivilege(TimelockNamespace.of("foo" + SUFFIX))).isTrue();
```
jeremyk-91: We probably want one that the suffix itself is OK (or, if we want to explicitly ban this, a small change in the impl + one that the suffix itself is not OK)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/config/ClientPrivilegesConfiguration.java
```diff
@@ -35,15 +39,29 @@
...
     @Value.Derived
     @Override
     public Privileges privileges() {
-        return NamespacePrivileges.of(namespaces());
+        return NamespacePrivileges.of(namespaces(), namespaceSuffixes());
+    }
+
+    @Value.Check
+    protected void check() {
+        BANNED_SUFFIXES.forEach(bannedSuffix -> Preconditions.checkArgument(!namespaceSuffixes().contains(bannedSuffix),
```
felixdesouza: I don't believe this check is sufficient. You want to `trim()` and then check it's non-empty + you also want to validate that the suffix would be an acceptable timelock namespace. Something along the lines of this:

Make `TimelockSuffix` an immutable type or call it something better. See [here](https://immutables.github.io/immutable.html#normalization) for immutables normalisation usage.

```java
@Value.Check
protected TimelockSuffix check() {
  if (!value().trim().equals(value()) {
    return TimelockSuffix.of(value().trim());
  }

  Preconditions.checkArgument(isValidClient(value(), "suffix is not a valid client");
  return this;
}
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/config/NamespacePrivileges.java
```diff
@@ -26,15 +26,21 @@
...
     }
 
     @Override
     public boolean hasPrivilege(TimelockNamespace namespace) {
-        return namespaces().contains(namespace);
+        return namespaces().contains(namespace) || hasMatchingSuffix(namespace);
+    }
+
+    private boolean hasMatchingSuffix(TimelockNamespace namespace) {
+        return namespaceSuffixes().stream().anyMatch(suffix -> namespace.value().endsWith(suffix));
```
felixdesouza: should it be an error for the suffix to be the namespace i.e. if suffix is "_alta", someone coming through with "_alta" as a namespace?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/config/NamespacePrivilegesTest.java
felixdesouza: yeah I was just commenting on this

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/auth/config/PrivilegesConfigurationTest.java
```diff
@@ -68,6 +75,20 @@ public void canDeserializeClientConfig() throws IOException {
         assertThat(deserializedConfiguration).isEqualTo(expectedConfiguration);
     }
 
+    @Test
+    public void canDeserializeClientConfigWithSuffixes() throws IOException {
+        PrivilegesConfiguration deserializedConfiguration =
```
felixdesouza: just bundle this test with the test above, they're only two tests because you implemented them in separate prs

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/auth/config/ClientPrivilegesConfiguration.java
felixdesouza: alternatively, it'd be sufficient to just check that the `value()`  is a valid namespace, instead of trying to normalise it, this is the same as a normal check.

PR #4067, opened 2019-06-04
Reduce TS default pause

gmaretic:
**Goals (and why)**:
To deal with very sparse rows in sweepable cells

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
@@ -53,7 +53,7 @@ public abstract class TargetedSweepRuntimeConfig {
 
     @Value.Default
     public long pauseMillis() {
-        return 500L;
+        return 5L;
     }
 
     public static TargetedSweepRuntimeConfig defaultTargetedSweepRuntimeConfig() {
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
```diff
@@ -53,7 +53,7 @@ void checkShardSize() {
 
     @Value.Default
     public long pauseMillis() {
-        return 500L;
+        return 5L;
```
sandorw: Let's do 50 until we're confident we need to go lower. I'm wary of the load increase from always being this fast. If we need to lower the pause this much, we may need to invest in smarter backoffs instead (continuing if the last iteration did work, otherwise pausing for longer).

felixdesouza:
We can't do this. This increases load on timelock for every single . We've gone from 2 per second per targeted sweep thread per node per service to 20 per second per node per service. This includes a lock and unlock at the very least.

If we didn't have atlas as a service, and with a rough estimate of around 80 different services with 2 nodes each. 18 * 2 * 80 ~ 3k extra requests per second, this is with only 1 targeted sweep thread! (although autobatcher might help here across multiple threads) So things will eventually begin to crawl to halt as more products upgrade to this.

pkoenig10:
I also noticed that this change caused a significant increase in unit test time. If we make this change again in the future, we should not modify the default configuration for the in-memory transaction manager used in tests.

sandorw:
Did you investigate or find anything specific @pkoenig10 ? I'm surprised that it had any impact on your unit tests unless we're not cleaning up or stopping the process properly.

PR #4070, opened 2019-06-05
[_transactions2] Part 50: Clarity for DbKVS and JDBC users

jeremyk-91:
**Goals (and why)**:
There is an irregularity in the past which could have happened if users of KVSes other than memory and Cassandra attempted to install transaction schema versions V > 1. Specifically, the TransactionSchemaManager would still attempt to install V, and this could happen fine. Thus you could get a log message along the lines of `We attempted to install the transactions schema version 2, and this was successful. This version will take effect no later than timestamp 31415926535.`, which is incredibly misleading as the actual TransactionService used always only references transactions1. 

Also the old behaviour can lead to **SEVERE DATA CORRUPTION**™️ if we end up subsequently releasing a version of DbKVS that _does_ support transactions2, as in #3869, since everything has actually been written at version 1 despite whatever the coordination service says is correct.

- Don't start the background thread that attempts to install new schema versions if we're using a KVS that doesn't support.
- Log at WARN if, on startup, the runtime config has a non-1 value and you're using a non-"has details of failures" KVS, saying that this is a weird configuration and is not allowed.

This was pretty awkward to test. ETEs should verify that we're still able to operate normally.

**Concerns (what feedback would you like?)**: Nothing in particular.

**Where should we start reviewing?**: TransactionManagers.java

**Priority (whenever / two weeks / yesterday)**: ASAP - will be good before @h-korus's team's work on enabling transactions2 kicks off

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -488,15 +488,29 @@ public abstract class TransactionManagers {
                 TransactionService.class,
                 TransactionServices.createTransactionService(keyValueService, transactionSchemaManager)),
                 closeables);
-        Optional<TransactionSchemaInstaller> schemaInstaller =
-                keyValueService.getCheckAndSetCompatibility() == CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE
-                        ? Optional.of(initializeTransactionSchemaInstaller(
-                                closeables, runtimeConfigSupplier, transactionSchemaManager))
-                        : Optional.empty();
+        Optional<TransactionSchemaInstaller> schemaInstaller = getTransactionSchemaInstaller(
+                closeables, keyValueService, runtimeConfigSupplier, transactionSchemaManager);
         return ImmutableTransactionComponents.builder()
                 .transactionService(transactionService)
                 .schemaInstaller(schemaInstaller)
                 .build();
+    }
+
+    private Optional<TransactionSchemaInstaller> getTransactionSchemaInstaller(
+            @Output List<AutoCloseable> closeables,
+            KeyValueService keyValueService,
+            Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier,
+            TransactionSchemaManager transactionSchemaManager) {
+        if (keyValueService.getCheckAndSetCompatibility() == CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE) {
+            return Optional.of(initializeTransactionSchemaInstaller(
+                    closeables, runtimeConfigSupplier, transactionSchemaManager));
+        }
+        runtimeConfigSupplier.get().internalSchema().targetTransactionsSchemaVersion().ifPresent(version ->
+                log.warn("This service seems like it has been configured to use transaction schema version {},"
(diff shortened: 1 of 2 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -475,6 +474,48 @@ private boolean targetedSweepIsFullyEnabled() {
...
+                closeables);
+        Optional<TransactionSchemaInstaller> schemaInstaller = getTransactionSchemaInstaller(
+                closeables, keyValueService, runtimeConfigSupplier, transactionSchemaManager);
+        return ImmutableTransactionComponents.builder()
+                .transactionService(transactionService)
+                .schemaInstaller(schemaInstaller)
+                .build();
+    }
+
+    private Optional<TransactionSchemaInstaller> getTransactionSchemaInstaller(
```
gmaretic: ``getTransactionSchemaInstallerIfSupported``

docs/source/release_notes/release-notes.rst
```diff
@@ -58,6 +58,12 @@ develop
          - Fixed a bug causing connection leaks to timelock.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4052>`__)
 
+    *    - |improved|
+         - If using DbKVS or JDBC KVS, we no longer spin up a background thread attempting to install new transaction schema versions.
+           Furthermore, we now log at WARN if configuration indicates that a new schema version should be installed in these cases where it won't be respected.
+           Previously, we would always read to and write from transactions1, even though the installer might actually have installed a non-1 schema version (and logged that this happened!).
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/qqqq>`__)
```
gmaretic: 4070

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -475,6 +474,48 @@ private boolean targetedSweepIsFullyEnabled() {
...
+            Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier,
+            TransactionSchemaManager transactionSchemaManager) {
+        if (keyValueService.getCheckAndSetCompatibility() == CheckAndSetCompatibility.SUPPORTED_DETAIL_ON_FAILURE) {
+            return Optional.of(initializeTransactionSchemaInstaller(
+                    closeables, runtimeConfigSupplier, transactionSchemaManager));
+        }
+        runtimeConfigSupplier.get().internalSchema().targetTransactionsSchemaVersion()
+                .filter(version -> version != TransactionConstants.DIRECT_ENCODING_TRANSACTIONS_SCHEMA_VERSION)
+                .ifPresent(version ->
+                        log.warn("This service seems like it has been configured to use transaction schema version {},"
```
gmaretic: Makes sense

docs/source/release_notes/release-notes.rst
jeremyk-91: ⭐️ will fix!

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: Will do

PR #4073, opened 2019-06-05
[Timelock Partitioning] Part 6: Batch Pingable Leader API + Resource

felixdesouza:
**Goals (and why)**:
If we are to run `n` instances of paxos, then we need a way to be able to coalesce all the pings for each instance without necessarily increasing our request load by `n`.

This is in a similar vein to #4061 combined with #4065.

Fairly trivial, just call through to the local knowledge (can get this through `PaxosComponents`, and then we're done!

* Duplicated the code from `PaxosLeaderElectionService` for ping. Wiring through a `PingableLeader` is somewhat complicated.

ASAP 💃

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
@@ -1,0 +1,59 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.paxos;
+
+import java.util.Set;
+import java.util.UUID;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.palantir.leader.PingableLeader;
+
+@Path("/batch/leader")
+public interface BatchPingableLeader {
+
+    /**
+     * Batch counterpart to {@link PingableLeader#ping}. If this call returns, then the server is reachable.
+     * <p>
+     * For the given set of {@code clients}, the remote server returns the clients for which it thinks it is the leader
+     * for. Clients that the remote server is not the leader for will be excluded from the results.
+     *
+     * @param clients set of clients to check remote servers leadership on
+     * @return clients which the remote server believes to be the leader
+     */
+    @GET
+    @Path("ping")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    Set<Client> ping(Set<Client> clients);
+
+    /**
+     * Re-exported version of {@link PingableLeader#getUUID}. Returns the remote servers unique leadership identifier.
+     * This will be the same across all clients that the remote server it is currently aware of. It is therefore safe to
+     * coalesce ping calls to the same remote server for different clients.
+     *
+     * @return the remote servers unique leadership identifier
+     */
+    @GET
+    @Path("uuid")
+    @Produces(MediaType.APPLICATION_JSON)
+    UUID uuid();
+}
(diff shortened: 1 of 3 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
```diff
@@ -0,0 +1,59 @@
...
+public interface BatchPingableLeader {
+
+    /**
+     * Batch counterpart to {@link PingableLeader#ping}. If this call returns, then the server is reachable.
+     * <p>
+     * For the given set of {@code clients}, the remote server returns the clients for which it thinks it is the leader
+     * for. Clients that the remote server is not the leader for will be excluded from the results.
+     *
+     * @param clients set of clients to check remote servers leadership on
+     * @return clients which the remote server believes to be the leader
```
jeremyk-91: nit: param - set of clients to check the remote server's leadership on
return - clients for which the remote server believes that it is the leader for?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
```diff
@@ -0,0 +1,59 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    Set<Client> ping(Set<Client> clients);
+
+    /**
+     * Re-exported version of {@link PingableLeader#getUUID}. Returns the remote servers unique leadership identifier.
+     * This will be the same across all clients that the remote server it is currently aware of. It is therefore safe to
+     * coalesce ping calls to the same remote server for different clients.
+     *
+     * @return the remote servers unique leadership identifier
```
jeremyk-91: nit: apostrophes please

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeaderResource.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Set<Client> ping(Set<Client> clients) {
+        return KeyedStream.of(clients)
+                .map(components::learner)
+                .map(PaxosLearner::getGreatestLearnedValue)
+                .filter(Objects::nonNull)
+                .filter(this::isThisNodeTheLeaderFor)
+                .keys()
+                .collect(Collectors.toSet());
```
jeremyk-91: Yep, this makes sense. We might want to be a bit cautious about perf at some point since this seems like it might be a semi-hot code path, but this should be fine for now

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.palantir.leader.PingableLeader;
+
+@Path("/batch/leader")
```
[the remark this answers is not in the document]
felixdesouza: I would like to name space it, probably under `.internal/namespacedClient/batch/*` or something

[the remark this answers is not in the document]
felixdesouza: the purpose of `PaxosResource`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeaderResource.java
felixdesouza: we could do something with the acceptor cache? where the cache also contains uuid, and it could just be a thing of cache keys?

felixdesouza: I think that would be next steps if anything.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeader.java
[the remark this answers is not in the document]
felixdesouza: apologies, I mean `LeadershipResource` but the meaning still applies.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/BatchPingableLeaderResourceTests.java
```diff
@@ -0,0 +1,92 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class BatchPingableLeaderResourceTests {
+
+    private static final Client CLIENT_1 = Client.of("client1");
+    private static final Client CLIENT_2 = Client.of("client2");
+
+    private static final UUID LEADER_UUID = UUID.randomUUID();
+
+    @Mock(answer = Answers.RETURNS_DEEP_STUBS)
```
gmaretic: <strong>every time a mock returns a mock a fairy dies</strong>

felixdesouza: deep stubs is a wonderful exception

felixdesouza: the alternative is to make this into an interface, make an impl of it, that then returns me the mocks I want, which is horrible

PR #4076, opened 2019-06-05
Require Cassandra Credentials, and bump max version to 4.0

leonz:
**Goals (and why)**:

Starting in sls-cassandra 4.0, we will no longer allow anonymous users.  The easiest way to enforce this is by requiring clients to supply credentials to Atlas.

Changes `CassandraCredentialsConfig` to be non-optional in the KVS config, and opens up an upgrade path to `sls-cassandra 4.0`.

This is a breaking change for clients that may not necessarily be discovered at test-time.  However, at this point we expect that all modern clients are already supplying credentials.


sandorw:
Has auth rollout extended to the large internal product and their use of Cassandra?

leonz:
Yes, at this point all products should be using authentication.

PR #4081, opened 2019-06-06
The TransactionManager used on callbacks is instrumented

diogoholanda:
**Goals (and why)**:
Callbacks currently runs with an uninstrumented TransactionManager.
Some internal libraries do most of the work as a callback and we don't have visibility on those.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
@@ -196,6 +196,53 @@ public class SerializableTransactionManager extends SnapshotTransactionManager {
         }
     }
 
+    public static TransactionManager createInstrumented(
+            MetricsManager metricsManager,
+            KeyValueService keyValueService,
+            TimelockService timelockService,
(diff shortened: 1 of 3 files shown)


felixdesouza:
Might want to have a separate factory method that includes instrumentation. Big internal product is the only one that uses this, and they're likely to do their own instrumentation => double metrics.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -261,6 +311,54 @@ public static TransactionManager create(MetricsManager metricsManager,
...
+                timestampCache,
+                sweepQueueWriter,
+                callback,
+                initializer,
+                validateLocksOnReads,
+                transactionConfig,
+                false);
+    }
+
+    public static TransactionManager create(
```
jeremyk-91: this one could probably be private? no strong opinions either way

diogoholanda:
@jeremyk-91 should be good now. Could we merge and release? I am fairly keen on having those metrics back

sandorw:
Just tagged 0.146.0 @diogoholanda

PR #4084, opened 2019-06-11
Change ts pause millis back to 500ms.

felixdesouza:
**Goals (and why)**:
Revert #4067. See https://github.com/palantir/atlasdb/pull/4067#issuecomment-500870622

Need a better approach to this. Maybe holding the lock for longer/smarter backoffs etc.

ASAP


sandorw:
I think we'll need to follow-up soon with a correct batching fix, since we were tuning this time down to help targeted sweep keep up with recent changes in timestamp usage.

PR #4087, opened 2019-06-17
[_transactions2] Part 51: ADR Chapter 3 / Live Migrations and Coordination

jeremyk-91:
**Goals (and why)**:
- Document reasoning for decisions made as part of the transactions2 project. This chapter focuses more on live migrations and the coordination service.

- Write the docs.

- This is a docs change. The docs build passes. 

- Nothing in particular.

**Priority (whenever / two weeks / yesterday)**: whenever.


docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+  transactions2.
+- **(A3) I know about the coordination service and about transactions2.** This is fairly self-explanatory.
+
+Client services come with a version of AtlasDB, and at some point they typically enable transactions2 by default.
+We can thus similarly partition them:
+
+- C1 and C2 are defined as client service versions using versions of AtlasDB that would be classified as A1 and A2
+  respectively.
+- C3 refers to client service versions that use an A3 version of AtlasDB, but don't enable transactions2.
+- C4 refers to client service versions that use an A3 version of AtlasDB and enable transactions2. (Transactions2
```
felixdesouza: if C3 and C4 run concurrently, does C3 behave like C2 i.e. throw because transactions2 is not enabled, or will it just transparently work?

I suppose what I'm getting at is, it being weird that C3 has it disabled but can still work. It feels like it's exposing/relying on internal config server semantics no? as opposed to being baked into the binary.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+current value. We then read the value again, and keep trying until we know what version to use. If we read a version
+we don't support yet, we throw (presumably, this is part of a rolling upgrade and should resolve itself soon).
+
+Installing transactions2 itself is done by a background thread that submits a transform that installs version 2 (or
+whatever specific version is included in config).
+
+### Live Migrations
+
+Transactions2 was written with some of the heaviest users of AtlasDB in mind. These are core Palantir services where
+shutdown upgrades are costly or even verboten, and we thus implemented a mechanism for performing upgrades without
```
felixdesouza: > or even *verboten*

cute 😏

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+ranges [1, +∞) and be connected; note that this does not mean future behaviour is fixed, since in practice this map
+is read together with a validity bound. Thus, even though the map will contain as a key some range [C, +∞) for a
+constant C, it should be valid up to some point V > C - and behaviour at timestamps after V may subsequently be changed.
+
+We attempt to read the latest version of the range map, and if our timestamp falls within the validity bound, we
+retrieve the version. If it does not, we submit an identity transformation, which extends the validity bound of the
+current value. We then read the value again, and keep trying until we know what version to use. If we read a version
+we don't support yet, we throw (presumably, this is part of a rolling upgrade and should resolve itself soon).
+
+Installing transactions2 itself is done by a background thread that submits a transform that installs version 2 (or
```
felixdesouza: continuously? I suppose I think of "installation" as a one time thing, as opposed to something that's continuously running, not sure how else to describe.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+time) depending on the specific timestamps involved in calls.
+
+The mechanism for determining which ``TransactionService`` to use goes through the aforementioned
+``CoordinationService``. We agree on a ``TimestampPartitioningMap``, which is a mapping of timestamps to transaction
+schema versions. This is a range map, as we want to support rollbacks easily. The map also is guaranteed to span the
+ranges [1, +∞) and be connected; note that this does not mean future behaviour is fixed, since in practice this map
+is read together with a validity bound. Thus, even though the map will contain as a key some range [C, +∞) for a
+constant C, it should be valid up to some point V > C - and behaviour at timestamps after V may subsequently be changed.
+
+We attempt to read the latest version of the range map, and if our timestamp falls within the validity bound, we
```
felixdesouza: capital L?

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+#### Physical Implementation
+
+The coordination service needs to persist state; we've implemented this as a non-transactional table in the AtlasDB
+key-value service called ``_coordination``. Within this table, each sequence of values being agreed on takes up one row.
+The values themselves are stored as dynamic columns with a var-long dynamic column key.
+
+- The value stored at column 0 is a ``SequenceAndBound`` which indicates the sequence number of the currently valid
+  value and the validity bound.
+- Values are stored at column IDs equal to a fresh timestamp at the time they're written.
```
felixdesouza: not entirely necessary, but it would be nice to see a tabular example, in the same vein as previous sections of the ADR

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+#### Coordination Service
+
+We introduce a notion of a coordination service, which agrees on values being relevant or correct at a given timestamp.
+The sequence of values being agreed needs to evolve in a *backwards consistent* manner - that is, one should avoid
+changes that may affect behaviour for decisions made at timestamps before the point where one is making one's update.
+More formally, if we read a value for some timestamp TS, then all future values written must then ensure that decisions
+made at TS would be done in a way consistent with our initial read.
+
+Values provided are valid up to a specific timestamp, called the validity bound. When a transform is applied, a fresh
+timestamp F is taken, and the new value will be written with a validity bound F + C for some constant C. Going forward,
```
felixdesouza: nit: wrap `TS`, `F`, `C`, `F + C` etc. in back ticks

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
 one requesting thread, and although CellLoader2 also spins up many more requesting threads on the Cassandra side,
 the Cassandra cluster was unable to actually have these therads all do work concurrently.
 
-### Live Migrations and the Coordination Service
+### Determining Which Transaction Service To Use
+
+Effectively, we want to define  an internal schema version for AtlasDB, and implement a mechanism for performing schema
```
felixdesouza: more of a question than a comment, what happens when you have a transaction that occurs over a schema version switch? i.e. start timestamp in the transactions 1 range, but commit timestamp in the tx2 range? I imagine we do things just in terms of the start timestamp?

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: C3 will transparently work. Unfortunately it exists very much for the purposes of no-downtime upgrades, in that rolling from C1 or C2 to C4 causes problems, while rolling from C1/C2 to C3 is okay, and rolling from C3 and C4 is okay.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: The thread runs continuously, though there's some logic to not apply the transform if the current version is already 2. I can rephrase this to something like

```
There is a background thread that continuously runs, polling the AtlasDB runtime configuration. If the configuration is configured to a specific version (which at time of writing can be either 1 or 2), it will run a transform setting the transactions schema version going forward to that version. This is optional; if not configured, we will not attempt to run any transforms.
```

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: Yeah, that's correct. I will explicitly address this case in the doc.

- [ ] clarify starting before and finishing after a change-point case

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
felixdesouza: sounds good.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: I've added one of these after this part.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+}
+```
+
+#### Physical Implementation
+
+The coordination service needs to persist state; we've implemented this as a non-transactional table in the AtlasDB
+key-value service called ``_coordination``. Within this table, each sequence of values being agreed on takes up one row.
+The values themselves are stored as dynamic columns with a var-long dynamic column key.
+
+- The value stored at column 0 is a ``SequenceAndBound`` which indicates the sequence number of the currently valid
```
felixdesouza: somewhat ambiguous, you have `column1` and `column2`, there is no `column 0` and `value` is itself a column, so do you mean the value inside `column1` or do you mean the value inside the `value` column where `column1 == 0x00`?

It may make sense to talk about these in terms of triples i.e.

> consider the triple `(sequence_id, sequence_number, value)` to map to a physical table where:
> ```
> key = sequence_id
> column1 = sequence_number
> column2 = unused
> value = value
> ```

Then write everything in terms of the triples.

Either that or you have to be pretty explicit about what you mean when you say "value". 

NB: Not entirely clear whether those map to the "columns" but that's what I've inferred as I've found it quite confusing.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+For example, the Cassandra representation of the coordination table may look as follows.
+
+|  key | column1 | column2 |                                                                  value |
+|-----:|--------:|--------:|-----------------------------------------------------------------------:|
+| 0x6d |    0x00 |      -1 | 0x7b2273657175656e6365223a31333030372c22626f756e64223a353031333030377d |
+| 0x6d |  0xa711 |      -1 |                                     0x7b2276657...3334227d (216 bytes) |
+| 0x6d |  0xb2cf |      -1 |                                     0x7b2276657...3930227d (380 bytes) |
+
+The single ``key`` reflects that these values are all part of the same sequence of values. The column IDs are
```
felixdesouza: I think it's also useful to concretise the example/tie it back to transactions2.

* what is special about `0x6d` here?, what would it mean for transactions2?
  * 0x6d is the sequence identifier?

It can also do with the above wrt triples perhaps, as I can't make sense of this example whatsoever.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+``VAR_LONG`` encoded; ``0x00`` maps to zero, and the associated value decodes to a ``SequenceAndBound``, in this
+example ``{"sequence":13007,"bound":5013007}``. This means that the value at the column associated with sequence
+``13007`` (which encodes to ``b2cf``) is the currently agreed value; the value in column ``a711`` is not current.
+
+The main reason for this scheme is that while the validity bound itself changes frequently (approximately every
+``ADVANCEMENT_QUANTUM`` timestamps, which is 5,000,000 at time of writing), the value itself changes much less often.
```
felixdesouza: nit: `5,000,000`

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+For example, with transactions2, in the absence of rollbacks, the value is expected to change just once in the
```
felixdesouza: decide on whether to backtick `transactions2` or not and then stick to it

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+The mechanism for determining which ``TransactionService`` to use goes through the aforementioned
+``CoordinationService``. We agree on a ``TimestampPartitioningMap``, which is a mapping of timestamps to transaction
+schema versions. This is a range map, as we want to support rollbacks easily. The map also is guaranteed to span the
+ranges ``[1, +∞)`` and be connected; note that this does not mean future behaviour is fixed, since in practice this map
+is read together with a validity bound. Thus, even though the map will contain as a key some range ``[C, +∞)`` for a
+constant ``C``, it should be valid up to some point ``V > C`` - and behaviour at timestamps after ``V`` may subsequently
+be changed.
+
+Note that the ranges are based on the start timestamp. To give a concrete example, if our range map is
```
felixdesouza: Think this is a bit of a leap. We've got to the point where we've cut up timestamp space into difference ranges. But the connecting bit (in my mind) is:
> Since we accessed the original `TransactionService` via the `startTimestamp`, we will use the `startTimestamp` against the range map to find out the behaviour at that point.

Or something along those lines.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+we don't support yet, we throw (presumably, this is part of a rolling upgrade and should resolve itself soon).
+
+#### Changing Transaction Schema Versions
+
+AtlasDB clients continuously run a background thread that polls the runtime configuration. If that sets the
+target schema version to a specific version (which at time of writing can be 1 or 2), a transform will be run
+setting the transactions schema version going forward to that version.
+
+This is optional; if not configured, we will not attempt to run any transforms so the transaction schema version will
+remain what it is.
```
felixdesouza: Does that mean it has a default? What is the default behaviour?

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,159 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+### Live Migrations
+
+Transactions2 was written with some of the heaviest users of AtlasDB in mind. These are core Palantir services where
+shutdown upgrades are costly or even verboten, and we thus implemented a mechanism for performing upgrades without
+downtime. The key problem here involves reasoning about service nodes running different binary versions that are
+concurrently operating on an AtlasDB deployment.
```
felixdesouza: I believe the important distinction is 
> different binary versions that may be running different atlasdb versions.

This paragraph feels somewhat incomplete, maybe the bottom sections are sufficient

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: At this point we haven't introduced the Cassandra terms yet (`column1` and `column2` are Cassandra-schema specific constants that are used across Atlas), so here I refer to the column 0 in a dynamic column sense (which physically in Cassandra would be the value in `value` where `column1 = 0x00`). The real physical part probably starts after this.

Your inference is correct - though note that we can't use `column1` and `column2` in general to talk about physical tables, because these are Cassandra-specific. I've added a section above the Cassandra table example recalling the correspondence between the Atlas and Cassandra layers for dynamic columns.

jeremyk-91: I've separated this paragraph and some of the discussion later on (which isn't really about the physical layer) into a persistence section, and left the physical section with the table.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: `0x6d` is a not-inherently-meaningful sequence identifier which we've picked for AtlasDB usage. Conceptually, if we decided we wanted AtlasDB clients to agree on something else and unrelated to transactions2, they could use a different row in the table.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: In my usage, transactions2 shouldn't have backticks unless it is the table specifically. So earlier in the doc we have 

> As far as transactions2 is concerned, we observe that the number of partitions is very small, and thus we can go with a very low setting. We thus set ``bloom_filter_fp_chance`` to 0.0001. Empirically, we observed that for 1,000,000,000 timestamps, the bloom filter we had for the ``_transactions2`` table even with this setting was about 15 KB, while that
for ``_transactions`` was about 580 MB.

As for the phrase "read(ing) from transactions2", that would depend if it means reading values written by the scheme in general (no backticks), or reading values from the table itself (should have backticks). So I'd consider this correct:

> To determine if we should read from transactions2, we check when usage of transactions2 began, and then read from `_transactions2` if appropriate.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: Yep, added pretty much verbatim (apart from not using start timestamp as a symbol)

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: Yep, added

> If this was never configured at all, we default to using transactions1, to be consistent with
previous behaviour.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: Yeah, I will add that. The purpose of this paragraph is to argue the case for why live migrations for transactions are important, so I think it (or something to that effect) should stay - the bottom sections explain how you would do it but not why we want it.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,186 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+Conceptually, the state of the table may look as follows:
+
+| Sequence Id | Sequence Number |                              Value |
+|------------:|----------------:|-----------------------------------:|
+|           m |               0 | {"sequence":13007,"bound":5013007} |
+|           m |           10001 |                  <metadata object> |
+|           m |           13007 |        <different metadata object> |
+
+The sequence ID is the same, to reflect that these values are a part of the same sequence; ``m`` is chosen for
+AtlasDB's internal schema metadata in general, and isn't particularly meaningful in and of itself.
```
felixdesouza: might be worth mentioning either here or in the `Determining Transaction Schema Versions` section that this isn't inherently timestamp aware, and it's up to the implementation for a sequence id `m` to partition the timestamp space etc. Possibly also introducing something along the lines of, adding an abstraction layer (which already exists) such that this is handled for you automatically i.e. you never deal with the coordination service directly etc.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,186 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+#### Determining Transaction Schema Versions
+
+We still need to retain the ability to read from the old transactions table, when checking if/when values written
+before the transactions2 migration occurred. Thus, we replace the old ``TransactionService`` (that only reads/writes
+from transactions1) with one that delegates between multiple TransactionServices (concretely, just V1 and V2 at this
+time) depending on the specific timestamps involved in calls.
+
+The mechanism for determining which ``TransactionService`` to use goes through the aforementioned
+``CoordinationService``. We agree on a ``TimestampPartitioningMap``, which is a mapping of timestamps to transaction
```
felixdesouza: to be explicit, 

> to agree, is to have the value in the coordination store be the range map

something better worded obviously

felixdesouza:
lol, you have quite the PR here 😛

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+- C3 refers to client service versions that use an A3 version of AtlasDB, but don't enable transactions2.
+- C4 refers to client service versions that use an A3 version of AtlasDB and enable transactions2. (Transactions2
+  cannot be enabled with A1 or A2 versions of AtlasDB, as it doesn't exist in those versions!)
+
+Now, we can consider how multiple client service versions interact:
+
+- Any sets of versions that don't include C4s are safely interoperable, because in these versions transactions2 is never
+  installed anyway, so we always use transactions1. Even C1s and C3s can safely interoperate, since although C1s never
+  read the coordination table, their effective assumption that it contains everything mapping to 1 is correct.
+- C1s and C4s running concurrently can lead to **SEVERE DATA CORRUPTION**, since the C4s might install transactions2
```
gmaretic: Having C4 at and point in the past makes it potentially unsafe to roll back to C1 even if not running them concurrently.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,131 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+bound less than their timestamp of interest.
+
+```java
+public interface CoordinationService<T> {
+    Optional<ValueAndBound<T>> getValueForTimestamp(long timestamp);
+
+    CheckAndSetResult<ValueAndBound<T>> tryTransformCurrentValue(Function<ValueAndBound<T>, T> transform);
+
+    // ONLY FOR METRICS AND MONITORING - NOT FOR PRODUCTION DECISIONS.
+    Optional<ValueAndBound<T>> getLastKnownLocalValue();
```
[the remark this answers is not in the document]
jeremyk-91: Makes sense, removed.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
 
-### Live Migrations and the Coordination Service
+### Determining Which Transaction Service To Use
+
+Effectively, we want to define an internal schema version for AtlasDB, and implement a mechanism for performing schema
+transitions while ensuring nodes that want to commit transactions agree on the state of the world - or at least,
+agree sufficiently that they won't contradict one another.
+
+For a given start timestamp, we want to be able to decide what mechanism to use to find the commit timestamp (or,
+alternatively, if the transaction was aborted).
```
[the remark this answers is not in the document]
jeremyk-91: Yep, changed to mention this.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+Conceptually, the state of the table may look as follows:
+
+| Sequence Id | Sequence Number |                              Value |
+|------------:|----------------:|-----------------------------------:|
+|           m |               0 | {"sequence":13007,"bound":5013007} |
+|           m |           10001 |                  <metadata object> |
+|           m |           13007 |        <different metadata object> |
+
+The sequence ID is the same, to reflect that these values are a part of the same sequence; ``m`` is chosen for
```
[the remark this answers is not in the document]
jeremyk-91: This part is still referring to logical concepts though (so these different entries would be different rows in DbKVS). I will add a mention that this is one row in the Physical Implementation section later.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+time) depending on the specific timestamps involved in calls.
+
+The mechanism for determining which ``TransactionService`` to use goes through the aforementioned
+``CoordinationService``. Nodes agree on a ``TimestampPartitioningMap``, a mapping of timestamps to transaction schema
+versions, using the coordination service; this map is always the current value of the coordination store.
+This is a range map, as we want to support rollbacks easily. The map also is guaranteed to span the
+ranges ``[1, +∞)`` and be connected; note that this does not mean future behaviour is fixed, since in practice this map
+is read together with a validity bound. Thus, even though the map will contain as a key some range ``[C, +∞)`` for a
+constant ``C``, it should be valid up to some point ``V > C`` - and behaviour at timestamps after ``V`` may subsequently
+be changed.
```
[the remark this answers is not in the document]
jeremyk-91: Yep, that's right.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+Note that the ranges are based on the start timestamp; since we accessed the original ``TransactionService`` via the
+start timestamp, we will use the start timestamp against the range map to find out the behaviour at that point.
+To give a concrete example, if our range map is ``{[1, 5000) = 1, [5000, +∞) = 2}`` and we want to commit a transaction
+that started at timestamp 4990 and finished at 5010, it is still stored under schema version 1 (even if other
+transactions that started later may have already written to schema version 2 - if they started at 5000 and finished at
+5001, for instance).
+
+We attempt to read the latest version of the range map, and if our timestamp falls within the validity bound, we
+retrieve the version. If it does not, we submit an identity transformation, which extends the validity bound of the
+current value. We then read the value again, and keep trying until we know what version to use. If we read a version
```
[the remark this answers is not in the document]
jeremyk-91: Added a line about this:

>  Our
transform may be rejected if another transform has taken place and changed the value from underneath us; however, because all transforms push the bound forward, we will eventually have a value that is valid for our timestamp.

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+- C1 and C2 are defined as client service versions using versions of AtlasDB that would be classified as A1 and A2
+  respectively.
+- C3 refers to client service versions that use an A3 version of AtlasDB, but don't enable transactions2.
+- C4 refers to client service versions that use an A3 version of AtlasDB and enable transactions2. (Transactions2
+  cannot be enabled with A1 or A2 versions of AtlasDB, as it doesn't exist in those versions!)
+
+Now, we can consider how multiple client service versions interact:
+
+- Any sets of versions that don't include C4s are safely interoperable, because in these versions transactions2 is never
```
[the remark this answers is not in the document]
jeremyk-91: Yep, that's a very good point 💯

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: Added mention of how _if_ transactions2 has been used at some point in the past _then_ C1s can result in **SEVERE DATA CORRUPTION**™️

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
```diff
@@ -499,7 +499,193 @@ the worker pool has a finite size and even with the full batching algorithm in t
...
+
+#### Implementing Safe Upgrades
+
+To do rolling or blue/green upgrades from C1s or C2s to C4s, we thus recommend a checkpointing process that ensures that
+C1s (and C2s) never run concurrently with C4s.
+
+1. Upgrade the version of AtlasDB to an A3 version.
+2. Release your product (this is a C3), making this a *checkpoint release* - that is, all upgrades to subsequent
+   versions must go through this version.
+3. Enable transactions2, and release your product (this is a C4).
```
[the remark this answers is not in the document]
jeremyk-91: I'll add a section on downgrading from transactions2 in general (including not rolling back to a C1) - this comment made me realise that we hadn't talked about that at all.

jeremyk-91:
Tasks remaining:

- [ ] Render tables differently
- [ ] Implementation details for changing schema versions / validity bound perpetuates
- [ ] Downgrading from transactions2

docs/source/miscellaneous/doc/adr/0016-use-tickets-encoding-for-transactions.md
jeremyk-91: There is a strange technical argument where if you ran with a C4 but it never did a transaction... 😅

gmaretic:
@jeremyk-91 it would be a shame if we had to close this

PR #4093, opened 2019-06-21
[WIPlash / POCketbook] Basic Lock Watching, Draft 1

jeremyk-91:
**Goals (and why)**:
- Provide a mechanism for being able to determine if locks were possibly taken out. The expected use case is a type of cache, where if one wishes to perform a read, one can execute a TimeLock RPC, and if determining that no lock was taken out since the last time the value was read with things unlocked, one knows then that the value has not changed.

- Provide an API for registering lock watches and retrieving their state, which is a pair of integers indicating when things were locked or unlocked.

- Added a couple of new tests for the lock watch service (both 

- Is it possible for an unlock to fail in such a way that we never register that a lock got unlocked?

- LockWatchResource probably


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.MultimapBuilder;
+import com.google.common.collect.Multimaps;
+import com.palantir.lock.LockDescriptor;
+
+@Path("/lock-watch")
```
sandorw: Do we need these annotations on the impl as well as the interface?

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/ExplicitLockPredicate.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.lock.LockDescriptor;
+
+@JsonSerialize(as = ImmutableExplicitLockPredicate.class)
+@JsonDeserialize(as = ImmutableExplicitLockPredicate.class)
+@Value.Immutable
+@JsonTypeName("explicit")
+public interface ExplicitLockPredicate extends LockPredicate {
+    Set<LockDescriptor> descriptors();
```
sandorw: Will be interesting to think about how broadly we end up applying this, as it scales Timelock from only retaining the collection of locks for ongoing writes to retaining some extra amount of objects that's theoretically unrelated to the write load.

The user also presumably has to pick the right one of row/cell locking corresponding to the table in question.

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchService.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * A service that allows users of a {@link com.palantir.lock.v2.TimelockService} to track what is happening with their
+ * locks. Specifically, users can register a watch over some lock IDs. Whenever a watched lock is locked or unlocked,
+ * an event is published to this service.
+ */
+@Path("/lock-watch")
+public interface LockWatchService {
```
sandorw: To summarize my rough understanding of usage:

- I must register a watch before I can be alerted of any changes, so should generally register a watch before populating my cache. If I must load my cache first to know what the keys are, then I should reread the values after setting the watches.
- To use the cache in a transaction, I must fetch my read timestamp before checking my watch states. In a serializable transaction, I must fetch my commit timestamp before checking watch states for read-write conflicts. 
- My `LockWatchState` response is the same size scale as my set of lock descriptors.

The last point is interesting because we might instead have considered a response with less granularity, at the expense of less granularity about which cache entries are stale. I'm not sure of any use cases off hand for batched lock watches, though, which is the only place this would particularly matter. 

Could we still avoid returning the lock descriptor in the response, given that the client presumably already knows the descriptor -> UUID mapping? I guess that only works when each watch is for a single lock/key.

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/ExplicitLockPredicate.java
jeremyk-91: Yeah, that's why the impl has a limit on the number of watches you can register. I should probably limit the number of descriptors too.

Correct - users who want to cache things need to know the type of locking involved (but I would imagine they would as they have the schema).

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
jeremyk-91: we... don't. I think I added the interface later when I found that our proxying infrastructure mandates interfaces

jeremyk-91: will remove

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchService.java
jeremyk-91: Correct: I only know that the value `V` I'm reading for some cell `C` hasn't changed if I knew that the watch for `C` was in some state `S` before I read `V`, and that the watch for `C` is still in state `S`. Thus point 1 is correct.

The intention was for this to replace a KVS read that a user might do in their transaction task, and yes the start/read timestamp _must_ have been retrieved by this point. Otherwise, we could read the cache, think our value is still current, and then have another transaction lock and commit before we get our read timestamp. This logic similarly holds for commit timestamps in serializable transactions, as you mention correctly.

The third point is correct, and yep there is some freedom with regard to granularity. I guess the idea was that if you were reading a cache that had say three values and one of them changed, you could register one watch, and then only need to re-read the one that changed. That said, this might be better achieved by allowing some form of a batched endpoint for retrieving lock state (rather than passing the descriptors around) - happy to change this if we decide we prefer that.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        };
+    }
+
+    @PUT
+    @Path("watch")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public UUID registerWatch(LockPredicate predicate) {
+        if (activeWatches.size() > WATCH_LIMIT) {
```
felixdesouza: I don't believe calling `activeWatches` will give you a consistent hard limit under concurrency. We have a similar problem with timelock and clients i.e. it is possible to register more than the limit.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+@Path("/lock-watch")
+public class LockWatchResourceImpl implements LockWatchResource {
+    private static final int WATCH_LIMIT = 50;
+
+    private final Multimap<LockDescriptor, LockWatch> explicitDescriptorsToWatches;
+    private final Map<UUID, LockWatch> activeWatches;
+    private final LockEventProcessor eventProcessor;
+
+    public LockWatchResourceImpl() {
+        this.explicitDescriptorsToWatches = Multimaps.synchronizedListMultimap(MultimapBuilder.hashKeys()
```
felixdesouza: you should tease apart the actual logic and the http request processing

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.MultimapBuilder;
+import com.google.common.collect.Multimaps;
+import com.palantir.lock.LockDescriptor;
+
+@Path("/lock-watch")
+public class LockWatchResourceImpl implements LockWatchResource {
+    private static final int WATCH_LIMIT = 50;
```
felixdesouza: is there a way to share between nodes of the same client?

if I've understood correctly, wouldn't this just mean that you have on average `WATCH_LIMIT / num-of-nodes`?

felixdesouza: continuing based on that assumption, it feels like you would need some sort of additional identifier, similar to reentrancy but for the lock watches, so that different nodes of the same service will get the same watch/id for the same set of things they want to watch

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatch.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.google.common.collect.Maps;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.lock.LockDescriptor;
+
+class LockWatch {
+    private Map<LockDescriptor, LockIndexState> indexStates;
+    private Map<LockDescriptor, AtomicLong> counters;
```
felixdesouza: it feels as if you actually want to have one global `Map<LockDescriptor, LockIndexState>` and `Map<LockDescriptor, AtomicLong>`, and then query based on watch ids, where you reconstruct based on the global map. I suppose I'm not super clear how this will be consumed just yet.

`computeIfPresent` also runs atomically, so there doesn't seem to be a need to keep separate maps here, since you only ever update one when updating the other.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
felixdesouza: we'd definitely need this for client node restarts no? i.e. timelock doesn't go down but the client restarts? or maybe some sort of timeout where the client is also refreshing (you could do this via `getWatchState` I suppose?

felixdesouza: how do you claim back watches that aren't being observed?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
```diff
@@ -83,6 +88,11 @@ private void acquireLocks() {
             try {
                 AsyncResult<Void> lockResult = AsyncResult.completedResult();
                 for (AsyncLock lock : locks.get()) {
+                    // This must happen BEFORE the lock is locked, it's safe to have a spurious lock.
+                    lockResult = lockResult.concatWith(() -> {
+                        lockEventProcessor.registerLock(lock.getDescriptor());
```
felixdesouza: for the caching perspective client side, is registering a lock despite not necessarily acquiring it, fine as long as the last unlock remains the same?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/HeldLocks.java
```diff
@@ -79,6 +85,7 @@ public synchronized boolean unlock() {
 
         for (AsyncLock lock : acquiredLocks) {
             lock.unlock(token.getRequestId());
+            lockEventProcessor.registerUnlock(lock.getDescriptor());
```
felixdesouza: I think this is a bit weird with the targeted sweep lock (unlocks are async) + immutable timestamp lock (why would you want to watch this lock)? which further reinforces my belief that this should exist inside `ExclusiveLock`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResource.java
```diff
@@ -0,0 +1,50 @@
...
+    UUID registerWatch(LockPredicate predicate);
+
+    @DELETE
+    @Path("watch/{id}")
+    void unregisterWatch(@PathParam("id") UUID watchIdentifier) throws NotFoundException;
+
+    @POST
+    @Path("watch/{id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    LockWatchState getWatchState(@PathParam("id") UUID watchIdentifier) throws NotFoundException;
```
felixdesouza: I think to minimise payload, we should probably do something similar to what I did on the `AcceptorCache` where you pass in some sort of cache key, and timelock can tell you whether or not something has changed since x, a diff of sorts.

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchService.java
```diff
@@ -0,0 +1,68 @@
...
+    @DELETE
+    @Path("watch/{id}")
+    void unregisterWatch(@PathParam("id") UUID watchIdentifier) throws NotFoundException;
+
+    /**
+     * Gets the state a given watch is in.
+     *
+     * @param watchIdentifier watch identifier
+     * @return true if and only if some lock guarded by this watch was locked
+     * @throws NotFoundException if we do not recognise the watch identifier the user provides
```
felixdesouza: Don't think this is accurate if this is supposed to be client side. It will always be something else. Prefer to just say it will translate to a 404.

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchService.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    @DELETE
+    @Path("watch/{id}")
+    void unregisterWatch(@PathParam("id") UUID watchIdentifier) throws NotFoundException;
+
+    /**
+     * Gets the state a given watch is in.
+     *
+     * @param watchIdentifier watch identifier
+     * @return true if and only if some lock guarded by this watch was locked
```
gmaretic: This seems to be out of date. Also we probably want to document (somewhere?) that no change to locked sequence guarantees that there was no lock in between, and a change to unlocked state guarantees there was an unlock, but the converses are not guaranteed

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
```diff
@@ -83,6 +88,11 @@ private void acquireLocks() {
             try {
                 AsyncResult<Void> lockResult = AsyncResult.completedResult();
                 for (AsyncLock lock : locks.get()) {
+                    // This must happen BEFORE the lock is locked, it's safe to have a spurious lock.
```
gmaretic: This comment is useful, but needs rephrasing. Maybe

> This must happen BEFORE the lock is locked, it's safe to register a spurious lock event?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockEventProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.watch;
+
+import com.palantir.lock.LockDescriptor;
+
+public interface LockEventProcessor {
+    static LockEventProcessor createNoOpForTests() {
```
gmaretic: Not the hugest fan of this design. I'd make it a class that takes two Consumers of LockDescriptor instead.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public LockWatchResourceImpl() {
+        this.explicitDescriptorsToWatches = Multimaps.synchronizedListMultimap(MultimapBuilder.hashKeys()
+                .arrayListValues()
+                .build());
+        this.activeWatches = Maps.newConcurrentMap();
+        this.eventProcessor = new LockEventProcessor() {
+            @Override
+            public void registerLock(LockDescriptor descriptor) {
+                System.out.println(activeWatches);
```
gmaretic: 🎖

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    private final Map<UUID, LockWatch> activeWatches;
+    private final LockEventProcessor eventProcessor;
+
+    public LockWatchResourceImpl() {
+        this.explicitDescriptorsToWatches = Multimaps.synchronizedListMultimap(MultimapBuilder.hashKeys()
+                .arrayListValues()
+                .build());
+        this.activeWatches = Maps.newConcurrentMap();
+        this.eventProcessor = new LockEventProcessor() {
+            @Override
```
gmaretic: If you have a constructor taking two Consumers, you can extract the two methods and just do `new LockEventProcessor(this::lock, this::unlock)`

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -448,6 +457,88 @@ public void instrumentationSmokeTest() throws IOException {
...
+        LockWatchState watchState = lockWatchService.getWatchState(watch);
+        assertThat(watchState).isEqualTo(ImmutableLockWatchState.builder()
+                .addLockStates(LockIndexState.createDefaultForLockDescriptor(LOCK_1))
+                .build());
+        lockWatchService.unregisterWatch(watch);
+    }
+
+    @Test
+    public void requestingWatchThatDoesNotExistThrows404() {
+        // Sigh. Dropwizard and remoting2 errors.
```
gmaretic: This is hilarious

timelock-api/src/main/java/com/palantir/atlasdb/timelock/watch/LockIndexState.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.palantir.lock.LockDescriptor;
+
+@JsonSerialize(as = ImmutableLockIndexState.class)
+@JsonDeserialize(as = ImmutableLockIndexState.class)
+@Value.Immutable
+public interface LockIndexState {
+    LockDescriptor lockDescriptor();
```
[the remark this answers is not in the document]
jeremyk-91: Yeah, descriptor hash should generally be sufficient (alternatives: mandate the user assigns IDs, or as discussed above, enforce a 1 lock/1 watch requirement)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
```diff
@@ -83,6 +88,11 @@ private void acquireLocks() {
...
+                    lockResult = lockResult.concatWith(() -> {
+                        lockEventProcessor.registerLock(lock.getDescriptor());
+                        return AsyncResult.completedResult();
+                    });
                     lockResult = lockResult.concatWith(() -> lockFunction.apply(lock));
```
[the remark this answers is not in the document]
jeremyk-91: Yeah, putting it in the lambda is fine. Applying the lock function only _submits the lock request_ (which can fail). I could modify the exclusive lock to do this precisely, though wasn't initially keen on touching that code

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public UUID registerWatch(LockPredicate predicate) {
+        if (activeWatches.size() > WATCH_LIMIT) {
+            throw new IllegalStateException("Cannot register watch, too many watches registered.");
+        }
+
+        UUID idToAssign = UUID.randomUUID();
+
+        if (!(predicate instanceof ExplicitLockPredicate)) {
```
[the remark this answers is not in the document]
jeremyk-91: This is future-proofing, the intention was to have regex-style locks which would need to be tested separately

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
jeremyk-91: That'll invalidate the old value. As discussed elsewhere it's probably worth actually putting it into `ExclusiveLock` though

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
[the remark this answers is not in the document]
felixdesouza: sure, but you're not testing regex-style locks inside this impl, you only care about the interface, so it's still fine to defer to deserialisation

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/HeldLocks.java
jeremyk-91: Conceptually that makes sense, and I think it's fine to put it there. I was initially trying to avoid touching `ExclusiveLock` completely

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatch.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    void registerUnlock(LockDescriptor descriptor) {
+        indexStates.computeIfPresent(
+                descriptor,
+                (unused, oldState) -> oldState.withUnlockSequence(counters.get(descriptor).incrementAndGet()));
+    }
+
+    LockWatchState getState() {
+        return ImmutableLockWatchState.builder()
```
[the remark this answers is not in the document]
jeremyk-91: The guarantees we get are enough, I think:

> The view's returned iterator is a "weakly consistent" iterator that will never throw ConcurrentModificationException, and guarantees to traverse elements as they existed upon construction of the iterator, and may (but is not guaranteed to) reflect any modifications subsequent to construction.

It's ok to read old values here - the dangerous case is if you miss a lock, since an unlock makes behaviour more permissive on clients. Users acquire locks before they can get their commit timestamp. However, by this point the calling transaction will have gotten their start, so it is not wrong to read the cached value - since any associated updates would not be visible to the calling transaction.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResource.java
jeremyk-91: Yep, that makes sense. I haven't thought much about serialization yet

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
jeremyk-91: Yep, it won't. We can synchronize this if needed

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/watch/LockWatchResourceImpl.java
jeremyk-91: 1/3 - Not in this version. Yeah, it's possible if you calculate some kind of hash and share IDs along the lines of stream store blocks, or have users explicitly supply an ID.

2 - Yes.

4 - Yes. Having a timeout on watch states seems fine.

5 - Right now there is an explicit un-register endpoint, though it won't help with servers going down. The solution to 4 could call this (there may be more efficient ways of doing it, too)

felixdesouza: if by version you mean PR that's not consumed by anything, sure, otherwise, I would say it's pretty much required imo as it has api ramifications.

jeremyk-91:
Think the core structure is in this PR. I'll need to build a few pieces, possibly separately:

- We seem to need some kind of expiry mechanism to deal with node failures. The simplest thing to do could be to evict a watch after it is not read for 10 minutes / some user specified parameter when registering the watch. Clients can then touch watches every 5 minutes or so (along the lines of `LockRefreshingLockService`).
- We probably want a mechanism for sharing locks between client nodes (otherwise if you say define the lock limit as 20, each client only gets . One way to do this could be based on equivalence of lock predicates; it is up to the client to independently track state to ensure that they don't skip reads when they shouldn't.
- Some smart (cough) mechanism for keeping payloads small (in particular, we probably don't want scaling with the length of lock descriptors).

PR #4094, opened 2019-06-25
Cross-Partitioning TS, Part 1: KeyValueService#deleteRows()

jeremyk-91:
**Goals (and why)**:
- Provide a mechanism to delete entire specified rows from a key-value-service in a single call. The current APIs don't support this: `delete(TableReference, Multimap)` requires column names to be provided explicitly, and `delete(TableReference, RangeRequest)` requires the deletion to take place over a contiguous range. 
- This is needed for efficient batch cleanup of multiple rows from the targeted sweep queues, since the operation we want to do is "for a given shard and strategy, delete the rows corresponding to some partitions"; by the time we're in this operation we may not have all of the columns to hand (e.g. if we're clearing out only half of a batch), and the queues have hashed row components so they are non-contiguous.

- Added a new `deleteRows()` method to the `KeyValueService`. This has a default implementation that creates single-row range requests and pushes them through sequentially to `deleteRange`.
- Implemented `deleteRows()` in `CassandraKeyValueService` to use `batch_mutate` so that only one RPC is needed regardless of the number of rows involved.
- Extended other implementations of `KeyValueService` to implement `deleteRows()` accordingly.

- Added several tests to the Abstract KVS integration test that exercise the new functionality.

- Is the abstract implementation workable? I'm pretty sure it does give correct results. My fear is more that users aren't expecting this to take one RPC per row. It's been documented.
- Do we need a better implementation for DbKVS? This is only planned to be used in TS so as far as I'm aware it's not urgently needed at this time.

**Where should we start reviewing?**: The API docs and test probably give a good feel of what the method is supposed to do.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -27,6 +27,7 @@ import com.palantir.common.annotation.NonIdempotent;
 import com.palantir.common.base.ClosableIterator;
 import com.palantir.common.exception.AtlasDbDependencyException;
 import com.palantir.processors.AutoDelegate;
+import com.palantir.processors.DoNotDelegate;
 import com.palantir.util.paging.BasicResultsPage;
 import com.palantir.util.paging.TokenBackedBasicResultsPage;
 
@@ -347,6 +348,7 @@ public interface KeyValueService extends AutoCloseable {
      * @param rows rows to delete
      */
     @Idempotent
+    @DoNotDelegate
     default void deleteRows(TableReference tableRef, Iterable<byte[]> rows) {
         rows.forEach(row -> deleteRange(tableRef, RangeRequests.ofSingleRow(row)));
     }
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -330,6 +330,24 @@ public interface KeyValueService extends AutoCloseable {
     void deleteRange(TableReference tableRef, RangeRequest range);
 
     /**
+     * Deletes multiple rows from the key-value store.
+     *
+     * Does not guarantee atomicity in any way (deletes may be partial within *any* of the rows provided, and
+     * there is no guarantee of any correlation or lack thereof between success of the deletes for each of the rows
+     * provided).
+     *
+     * Some systems may require more nodes to be up to ensure that a delete is successful. If this
+     * is the case then this method may throw if the delete can't be completed on all nodes.
+     *
+     * @param tableRef the name of the table to delete values from.
+     * @param rows rows to delete
+     */
+    @Idempotent
+    default void deleteRows(TableReference tableRef, Iterable<byte[]> rows) {
+        rows.forEach(row -> deleteRange(tableRef, RangeRequests.ofSingleRow(row)));
+    }
+
+    /**
      * For each cell, deletes all timestamps prior to the associated maximum timestamp. If this
      * operation fails, it's acceptable for this method to leave an inconsistent state, however
      * implementations of this method <b>must</b> guarantee that, for each cell, if a value at the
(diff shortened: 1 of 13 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -329,6 +329,26 @@ default boolean supportsCheckAndSet() {
...
     void deleteRange(TableReference tableRef, RangeRequest range);
 
+    /**
+     * Deletes multiple complete rows from the key-value store.
+     *
+     * Does not guarantee atomicity in any way (deletes may be partial within *any* of the rows provided, and
+     * there is no guarantee of any correlation or lack thereof between success of the deletes for each of the rows
+     * provided).
+     *
+     * Some systems may require more nodes to be up to ensure that a delete is successful. If this is the case then
```
gmaretic: Disclaimer level over 9000!

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1574,6 +1577,37 @@ private static boolean isForSingleRow(byte[] startInclusive, byte[] endExclusive
...
+                .map(deletion -> new Mutation().setDeletion(deletion))
+                .map(mutation -> keyMutationMapByColumnFamily(tableRef, mutation))
+                .collectToMap();
+
+        try {
+            clientPool.runWithRetry(client -> {
+                client.batch_mutate("deleteRows", mutationMap, DELETE_CONSISTENCY);
+                return null;
+            });
+        } catch (UnavailableException e) {
```
gmaretic: This is currently broken, but one day we might fix it

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/ProfilingKeyValueService.java
```diff
@@ -132,6 +133,19 @@ private ProfilingKeyValueService(KeyValueService delegate) {
...
+    private static BiConsumer<LoggingFunction, Stopwatch> logTimeAndTableRows(String method,
+            TableReference tableRef,
+            Iterable<byte[]> rows) {
+        long startTime = System.currentTimeMillis();
+        return (logger, stopwatch) ->
+                logger.log("Call to KVS.{} at time {}, on table {} for (estimated) {} rows took {} ms.",
+                        LoggingArgs.method(method),
+                        LoggingArgs.startTimeMillis(startTime),
+                        LoggingArgs.tableRef(tableRef),
+                        LoggingArgs.rowCount(Ints.saturatedCast(rows.spliterator().estimateSize())),
```
gmaretic: Glad we are coding defensively to not mess up our logging if deleting more than `Integer.MAX_VALUE` takes longer than a second 🗡

docs/source/release_notes/release-notes.rst
```diff
@@ -89,6 +89,11 @@ develop
            should improve targeted sweep throughput without adding additional load on Timelock for places where the downtime between targeted sweep iterations is a bottleneck.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4086>`__)
 
+    *    - |devbreak|
+         - ``KeyValueService`` implementations now have a new endpoint ``deleteRows()`` that allows row-level deletes to be performed.
+           Users who extend ``KeyValueService`` will need to implement this method.
```
gmaretic: but can refer to AbstractKVS for a functional implementation?

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/ProfilingKeyValueService.java
jeremyk-91: Yeah... I think it's a defensive habit, but I end up thinking twice when I downcast things

docs/source/release_notes/release-notes.rst
jeremyk-91: yeah, will add this.

PR #4095, opened 2019-06-26
Cross-Partitioning TS, Part 2: SweepQueueReader Ingests Multiple Batches

jeremyk-91:
**Goals (and why)**:
- Make targeted sweep faster as it's been struggling with internal shopping product at a few stacks.

- Allow the `SweepQueueReader` to read multiple fine partitions before proceeding to delete the relevant write-infos. This should save some RPCs.
- Extend the `SweepQueueCleaner` to deal with cases where a batch that was read spans multiple fine partitions (these may extend across a coarse partition safely).

Verified existing tests work.
Also not visible here, but @gmaretic and I ran the tests with batching up to 8 partitions, and verified that behaviour was sane / all failures in `TargetedSweeperTest` were expected and accounted for.

- Are there some edge cases we've missed out which could lead to things never getting cleaned?

**Where should we start reviewing?**: SweepBatchAccumulator probably

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepBatchAccumulator.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepBatchAccumulator.java
@@ -1,0 +1,96 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.sweep.queue;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
+import com.palantir.atlasdb.keyvalue.api.CellReference;
+import com.palantir.atlasdb.keyvalue.api.ImmutableCellReference;
+import com.palantir.atlasdb.schema.generated.SweepableCellsTable;
+import com.palantir.common.streams.KeyedStream;
+
+class SweepBatchAccumulator {
+    private final List<WriteInfo> accumulatedWrites = Lists.newArrayList();
+    private final List<SweepableCellsTable.SweepableCellsRow> accumulatedDedicatedRows = Lists.newArrayList();
+    private final long sweepTimestamp;
+
+    private long progressTimestamp;
+    private boolean anyBatchesPresent = false;
+
+    SweepBatchAccumulator(long sweepTimestamp, long progressTimestamp) {
+        this.sweepTimestamp = sweepTimestamp;
+        this.progressTimestamp = progressTimestamp;
+    }
+
+    void accumulateBatch(SweepBatch sweepBatch) {
+        Preconditions.checkState(sweepBatch.lastSweptTimestamp() < sweepTimestamp,
+                "Tried to accumulate a batch %s at timestamp %s that went beyond the sweep timestamp %s!"
+                        + " This is unexpected, and suggests a bug in the way we read in targeted sweep."
+                        + " This by itself does not mean that AtlasDB service is compromised, but targeted sweep"
+                        + " may not be working.",
+                sweepBatch,
+                sweepBatch.lastSweptTimestamp(),
+                sweepTimestamp);
+
+        accumulatedWrites.addAll(sweepBatch.writes());
+        accumulatedDedicatedRows.addAll(sweepBatch.dedicatedRows().getDedicatedRows());
+        progressTimestamp = Math.max(progressTimestamp, sweepBatch.lastSweptTimestamp());
+        anyBatchesPresent = true;
+    }
+
+    long getProgressTimestamp() {
+        return progressTimestamp;
+    }
+
+    SweepBatch toSweepBatch() {
+        return SweepBatch.of(
+                getLatestWritesByCellReference(),
+                DedicatedRows.of(accumulatedDedicatedRows),
+                getLastSweptTimestamp());
+    }
+
+    boolean shouldAcceptAdditionalBatch() {
+        return accumulatedWrites.size() < SweepQueueUtils.SWEEP_BATCH_SIZE
+                && progressTimestamp < (sweepTimestamp - 1);
+    }
+
+    private List<WriteInfo> getLatestWritesByCellReference() {
+        Map<CellReference, List<WriteInfo>> writes = accumulatedWrites.stream()
+                .collect(Collectors.groupingBy(writeInfo -> ImmutableCellReference.builder()
+                        .cell(writeInfo.cell())
+                        .tableRef(writeInfo.tableRef())
+                        .build()));
+        return KeyedStream.stream(writes)
+                .map(list -> list.stream().max(Comparator.comparing(WriteInfo::timestamp)))
+                .map(Optional::get) // groupingBy() won't return empty lists
+                .values()
+                .collect(Collectors.toList());
+    }
+
+    private long getLastSweptTimestamp() {
+        if (anyBatchesPresent) {
+            return progressTimestamp;
+        }
+        return sweepTimestamp - 1;
+    }
+}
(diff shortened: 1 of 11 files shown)


jeremyk-91:
Note: Build is broken, because we need something that is able to handle large ranges without actually attempting to create and/or delete all intermediate rows.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
```diff
@@ -161,56 +168,67 @@ public int getNumShards() {
...
         }
 
         private SweepQueueWriter createWriter() {
             return new SweepQueueWriter(timestamps, cells);
         }
 
         private SweepQueueReader createReader() {
-            return new SweepQueueReader(timestamps, cells);
+            return new SweepQueueReader(
+                    timestamps, cells, () -> runtimeConfigSupplier.get().maximumPartitionsToBatchInSingleRead());
```
gmaretic: I'm not sure if pushing the whole RuntimeConfig all the way down just for this one value makes sense opposed to just having a Supplier<Integer> and leaking less abstraction. On the other hand, not a fan of injecting magic numbers out of thin air either, so feel like there is no great option here.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -1218,7 +1220,7 @@ private void createAndInitializeSweepersAndWaitForOneBackgroundIteration(int swe
     }
 
     private void waitUntilSweepRunsOneIteration() throws InterruptedException {
-        Thread.sleep(3000L);
+        Thread.sleep(4000L);
```
gmaretic: "fine tuning"

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -523,17 +524,19 @@ public void sweepableCellsGetsScrubbedWheneverLastSweptInNewPartition() {
         enqueueWriteCommitted(TABLE_CONS, tsSecondPartitionFine);
         enqueueWriteCommitted(TABLE_CONS, getSweepTsCons());
 
-        // last swept timestamp: TS_FINE_GRANULARITY - 1
+        // last swept timestamp: TS_FINE_GRANULARITY - 1: fine partition 0 is completely swept
```
gmaretic: 🎖

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
```diff
@@ -15,21 +15,40 @@
...
 package com.palantir.atlasdb.sweep.queue;
 
-import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+import java.util.function.IntSupplier;
 
 class SweepQueueReader {
     private final SweepableTimestamps sweepableTimestamps;
     private final SweepableCells sweepableCells;
+    private final IntSupplier maximumPartitionsToBatchInSingleRead;
```
gmaretic: Ah didn't even realise you do in the end use an IntSupplier, so would definitely replace it by IntSupplier up the chain

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
```diff
@@ -15,21 +15,40 @@
...
+                currentBatch++) {
+            Optional<Long> nextFinePartition = sweepableTimestamps.nextSweepableTimestampPartition(
+                    shardStrategy, accumulator.getProgressTimestamp(), sweepTs);
+            if (!nextFinePartition.isPresent()) {
+                return accumulator.toSweepBatch();
+            }
+            SweepBatch batch = sweepableCells.getBatchForPartition(
+                    shardStrategy, nextFinePartition.get(), accumulator.getProgressTimestamp(), sweepTs);
+            accumulator.accumulateBatch(batch);
+            if (batch.isEmpty()) {
```
gmaretic: What does this do?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueCleaner.java
```diff
@@ -78,12 +86,15 @@ private void cleanNonDedicatedRow(ShardAndStrategy shardStrategy, long partition
     }
 
     private void cleanSweepableTimestamps(ShardAndStrategy shardStrategy, long oldProgress, long newProgress) {
-        if (firstIterationOfSweep(oldProgress)) {
+        if (firstIterationOfSweep(oldProgress) && SweepQueueUtils.tsPartitionCoarse(newProgress + 1) == 0) {
```
gmaretic: extract method

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueCleaner.java
```diff
@@ -50,21 +50,29 @@ public void clean(ShardAndStrategy shardStrategy, long oldProgress, long newProg
     }
 
     private void cleanSweepableCells(ShardAndStrategy shardStrategy, long oldProgress, long newProgress) {
```
gmaretic: This will now need to take in the SweepBatch, and extract the exact rows to clean up from there

jeremyk-91:
We'll merge Part 3 directly, as this commit by itself will cause OOMs in testing and so the build will not be green. However, I'll make sure that comments on this PR are addressed in the Part 3 PR, whether directly during our pairing or subsequently.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepBatchAccumulator.java
```diff
@@ -0,0 +1,95 @@
...
+                getLastSweptTimestamp());
+    }
+
+    boolean shouldAcceptAdditionalBatch() {
+        return accumulatedWrites.size() < SweepQueueUtils.SWEEP_BATCH_SIZE
+                && progressTimestamp < (sweepTimestamp - 1);
+    }
+
+    private List<WriteInfo> getLatestWritesByCellReference() {
+        Map<CellReference, List<WriteInfo>> writes = accumulatedWrites.stream()
```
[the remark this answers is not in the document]
jeremyk-91: yeah, we wrote this in a much nicer way via a merging `toMap()` collector

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueue.java
jeremyk-91: Yep. The alternative is supplying the number, but I didn't think it was too bad since at least the pieces engaged here are clearly relating to Targeted sweep.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
jeremyk-91: Yeah, the "regression" this PR introduces is a bit more than I expected, and it's not safe to give numbers greater than 5000 or else in-memory KVS will fail because there's enough time to run more than one iteration.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
jeremyk-91: alright, will do. I really didn't want to leak the abstraction all the way here, though I thought e.g. in sweep-queue-factory it might be acceptable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
jeremyk-91: discussed offline: it's probably not actually needed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueCleaner.java
jeremyk-91: Yeah, this nasty condition doesn't exist after we refactored the cleaners to use the partition-sets

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueCleaner.java
jeremyk-91: This is done on the part 3 branch

jeremyk-91:
closing out with merge of #4097

PR #4097, opened 2019-06-26
Cross-Partitioning TS, Part 3: Clean SweepableCells, SweepableTimestamps Via DeleteRows()

jeremyk-91:
**Goals (and why)**:
- Improve targeted sweep performance, specifically by reducing per-batch overhead.

- Perform cleanup of dedicated rows and non-dedicated rows as singular RPCs if going across batches, as opposed to separate deletes.

- Existing tests still work. I added an assertion to SweepableCellsTest to confirm that the dedicated rows are taken out in one call.

- I didn't bother with merging cleanup of dedicated and non dedicated rows as this seems cleaner as far as the code is concerned. Should I have done that?
- Are there any other pitfalls I've missed?

**Where should we start reviewing?**: SweepQueueCleaner


jeremyk-91:
Note: Build is broken, because we need something that is able to handle large ranges without actually attempting to create and/or delete all intermediate rows.

Issue #4089, opened 2019-06-19
C* KVS should throw InsufficientConsistencyException


gmaretic:
When a C* node is down, and a KVS operation that requires all nodes is called, we eventually throw a `RetryLimitReachedException`. We should identify when the cause for this failure is insufficient consistency by checking the suppressed exceptions and throw an `InsufficientConsistencyException` instead, since there are places in our codebase where we are relying on catching ICE for proper behaviour.

gmaretic:
One way this is manifesting is TS failing with generic error rather than not enough nodes online

PR #4100, opened 2019-06-28
Determine insufficient consistency when talking to Cassandra with retries

gmaretic:
**Goals (and why)**:
We want to determine if the request failed with insufficient consistency when retrying. This was previously done by checking if the last thrown exception was UnavailableException, but since we changed the thrown exception to RetryLimitReachedException, this logic was not working. See #4089 

Note that previous behaviour was working only sometimes (no guarantee the last node we try to talk to was up), but this one should always work as long as we have successfully talked to a node.
 
Catch the correct exception and check whether we encountered an UnavailableException in any of the attempts. Still catch UnavailableException when not running a retryable request.

Modified one node down delete ete test to test that now we get the correct exception

Did I miss anything?
 
Small

Would be nice to get it in...


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1556,9 +1556,8 @@ public void deleteRange(final TableReference tableRef, final RangeRequest range)
                     client.remove("deleteRange", tableRef, row, timestamp, DELETE_CONSISTENCY);
                     return null;
                 });
-            } catch (UnavailableException e) {
-                throw new InsufficientConsistencyException(
-                        "Deleting requires all Cassandra nodes to be up and available.", e);
+            } catch (RetryLimitReachedException e) {
+                throw CassandraUtils.wrapInIceForDeleteOrRethrow(e);
```
jeremyk-91: cool, `Ice` is a little ambiguous without context but won't object

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellDeleter.java
```diff
@@ -115,9 +114,8 @@ public String toString() {
                             + numVersions + " total versions of " + cellVersionsMap.size() + " keys)";
                 }
             });
-        } catch (UnavailableException e) {
-            throw new InsufficientConsistencyException("Deleting requires all Cassandra nodes to be up and available.",
-                    e);
+        } catch (RetryLimitReachedException e) {
```
jeremyk-91: is the original ever thrown / do we still want to preserve that behaviour? I think it's not and so it's fine to replace the entire block, but would like a sanity check

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/RetryLimitReachedException.java
```diff
@@ -27,4 +28,8 @@ public RetryLimitReachedException(List<Exception> exceptions) {
         super(String.format(MESSAGE, exceptions.size()));
         exceptions.forEach(this::addSuppressed);
     }
+
+    public <E extends Exception> boolean suppressed(Class<E> type) {
+        return Arrays.stream(getSuppressed()).map(type::isInstance).anyMatch(x -> x);
```
jeremyk-91: sure, I'd probably just write `Arrays.stream(getSuppressed()).anyMatch(type::isInstance)`?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraUtils.java
```diff
@@ -37,4 +41,11 @@ private CassandraUtils() {}
             CassandraKeyValueServiceConfig config) {
         return client -> client.describe_ring(config.getKeyspaceOrThrow());
     }
+
+    public static AtlasDbDependencyException wrapInIceForDeleteOrRethrow(RetryLimitReachedException ex) {
+        if (ex.suppressed(UnavailableException.class) || ex.suppressed(InsufficientConsistencyException.class)) {
+            throw new InsufficientConsistencyException("Deleting requires all Cassandra nodes to be available.", ex);
+        }
```
jeremyk-91: This happens if at any point we got one of these exceptions. So we might not be basing this on the latest state, but I think this is fine

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
jeremyk-91: I won't put this PR on ice because of this `Ice` (sorry)

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/RetryLimitReachedException.java
gmaretic: 😞

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellDeleter.java
gmaretic: Calls to runWithRetryOnHost will generally throw RetryLimitReachedException, we won't be able to get Unavailable at this level

PR #4104, opened 2019-07-02
Cross-Partitioning TS, Part 5: Parameterized Integration Tests

jeremyk-91:
**Goals (and why)**:
- Extend the known-good and known-rigorous `TargetedSweeperTest` to the multinode case. @gmaretic and I already ran these tests manually with batch sizes of 8 and confirmed that they work, but it's good to automate these.

- Re-write setup and assertions in the relevant tests to work parametrically in the batch size
- Fix one bug: don't keep requesting batches if you aren't making progress.

This is a test-only PR.

- The tests aren't fully parameterizable, in particular things get funky because some of them assume that a small constant number of fine partitions is less than a coarse partition. I tested with 1, 8 and 99 and these seem to pass; I tried 199 which causes a few failures, and 12345 which caused a few more.

**Where should we start reviewing?**: TargetedSweeperTest

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
@@ -38,6 +38,7 @@ import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.TS_COARSE_GRANULA
 import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.TS_FINE_GRANULARITY;
 import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.maxTsForFinePartition;
 import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.minTsForCoarsePartition;
+import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.minTsForFinePartition;
 import static com.palantir.atlasdb.sweep.queue.SweepQueueUtils.tsPartitionFine;
 
 import java.util.ArrayList;
@@ -52,7 +53,6 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.eclipse.jetty.util.ConcurrentHashSet;
-import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -787,23 +827,27 @@ public void stopReadingEarlyWhenEncounteringEntryKnownToBeCommittedAfterSweepTs(
...
+        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(
+                4 + writesInDedicated + 3 + (readBatchSize > 1 ? 2 : 0));
 
         immutableTs = 250;
 
         // we now read all to the end
         sweepNextBatch(ShardAndStrategy.conservative(CONS_SHARD));
-        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(4 + writesInDedicated + 3 + writesInDedicated + 2);
+        assertThat(metricsManager).hasEntriesReadConservativeEqualTo(
+                4 + writesInDedicated + 3 + writesInDedicated + 2 + + (readBatchSize > 1 ? 2 : 0));
```
gmaretic: `+ +`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
```diff
@@ -90,11 +95,21 @@
 import com.palantir.lock.v2.TimelockService;
 
 // TODO (jkong): Assert stricter properties on the types of ranged deletes that are created.
```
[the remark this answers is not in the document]
jeremyk-91: I think at the time I wasn't too happy that we test absence by just saying "we can't read the value", since theoretically a legacy sweeper that inserted single-cell deletes could pass.

Is this realistically going to happen? Probably not... I'll remove it on the basis that we don't plan on doing it.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/TargetedSweeperTest.java
jeremyk-91: doh! Amazingly it works because of unary operators

gmaretic:
Looks good, but tests going infinite again?

PR #4105, opened 2019-07-02
[TLS] TimeLock Slow Logger

jeremyk-91:
**Goals (and why)**:
- Have clearer signalling on the client side when `TimelockService` operations are slow. This is partially available using the metrics on internal logging systems, but this allows for better visibility (e.g. errors, stacktraces).

- Implement `ProfilingTimelockService` - see javadocs for detail on what it does.

- This is additional code, and tests have been added for what was added. The logic around picking the rate-limiter's limit isn't really covered by the tests, unfortunately.

- Are stack trace elements safe for logging?
- I chose to keep track of the longest operation satisfying the criteria; is this reasonable? Should I have kept more? My worry is taking up too much memory or logging very large operations. 
- I did think as a compromise logging the _number_ of slow operations and the _details_ of the slowest, but didn't as I didn't see a use for it immediately; would that be useful?
- A design decision here was that a calling thread would flush the logs; this concerned me a little as it is on a semi-hot path, though given you can only do that if you successfully acquired the rate limiter permit I think it's fine. 
- This also means that if you had a single slow operation complete immediately after a flush and no further operations, you wouldn't see anything. This seems very much like an edge case to me.

**Where should we start reviewing?**: ProfilingTimelockService

**Priority (whenever / two weeks / yesterday)**: two weeks?

### DIFF
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -103,6 +103,11 @@ develop
          - Relaxed concurrency model of ``MetricsManager`` allowing for more concurrency.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4098>`__)
 
+    *    - |improved|
+         - Clients now log at most once every 10 seconds when a timelock-level request is slow.
+           Please see the JavaDocs of ``ProfilingTimelockService`` for more information.
+           (`Pull Request <https://github.com/palantir/atlasdb/pull/4QQQ>`__)
+
 ========
 v0.144.0
 ========
--- a/docs/source/release_notes/release-notes.rst
+++ b/docs/source/release_notes/release-notes.rst
@@ -104,7 +104,7 @@ develop
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4098>`__)
 
     *    - |improved|
-         - Clients now log at most once every 10 seconds when a timelock-level request is slow.
+         - Clients now log at most once every 10 seconds when a ``TimelockService`` level request is slow.
            Please see the JavaDocs of ``ProfilingTimelockService`` for more information.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4QQQ>`__)
 
--- a/lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
@@ -39,6 +39,7 @@ import com.palantir.lock.v2.TimelockService;
 import com.palantir.lock.v2.WaitForLocksRequest;
 import com.palantir.lock.v2.WaitForLocksResponse;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 import com.palantir.timestamp.TimestampRange;
 
 /**
@@ -59,9 +60,11 @@ import com.palantir.timestamp.TimestampRange;
 public class ProfilingTimelockService implements AutoCloseable, TimelockService {
     private static final Logger log = LoggerFactory.getLogger(ProfilingTimelockService.class);
 
-    private static final Duration SLOW_THRESHOLD = Duration.ofSeconds(1);
+    @VisibleForTesting
+    static final Duration SLOW_THRESHOLD = Duration.ofSeconds(1);
     private static final Duration LOGGING_TIME_WINDOW = Duration.ofSeconds(10);
 
+    private final Logger logger;
     private final TimelockService delegate;
     private final Supplier<Stopwatch> stopwatchSupplier;
 
@@ -71,9 +74,11 @@ public class ProfilingTimelockService implements AutoCloseable, TimelockService 
 
     @VisibleForTesting
     ProfilingTimelockService(
+            Logger logger,
             TimelockService delegate,
             Supplier<Stopwatch> stopwatchSupplier,
             BooleanSupplier loggingPermissionSupplier) {
+        this.logger = logger;
         this.delegate = delegate;
         this.stopwatchSupplier = stopwatchSupplier;
         this.loggingPermissionSupplier = loggingPermissionSupplier;
@@ -81,7 +86,7 @@ public class ProfilingTimelockService implements AutoCloseable, TimelockService 
 
     public static ProfilingTimelockService create(TimelockService delegate) {
         RateLimiter rateLimiter = RateLimiter.create(1. / LOGGING_TIME_WINDOW.getSeconds());
-        return new ProfilingTimelockService(delegate, Stopwatch::createStarted, rateLimiter::tryAcquire);
+        return new ProfilingTimelockService(log, delegate, Stopwatch::createStarted, rateLimiter::tryAcquire);
     }
 
     @Override
@@ -159,14 +164,14 @@ public class ProfilingTimelockService implements AutoCloseable, TimelockService 
 
     private void trackActionAndMaybeLog(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
         stopwatch.stop();
-        accumulateSlowOperationTracking(actionName, stopwatch);
+        accumulateSlowOperationTracking(actionName, stopwatch, failure);
         logIfCanAcquirePermit();
     }
 
-    private void accumulateSlowOperationTracking(String actionName, Stopwatch stopwatch) {
+    private void accumulateSlowOperationTracking(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
         if (stopwatchDescribesSlowOperation(stopwatch)) {
             slowestOperation.accumulateAndGet(
-                    Optional.of(ImmutableActionProfile.of(actionName, stopwatch.elapsed())),
+                    Optional.of(ImmutableActionProfile.of(actionName, stopwatch.elapsed(), failure)),
                     (current, update) -> !update.isPresent()
                             || update.get().duration().compareTo(stopwatch.elapsed()) < 0
                             ? current
@@ -176,14 +181,17 @@ public class ProfilingTimelockService implements AutoCloseable, TimelockService 
 
     private void logIfCanAcquirePermit() {
         if (loggingPermissionSupplier.getAsBoolean()) {
-            Optional<ActionProfile> actionNameAndDuration = slowestOperation.getAndSet(Optional.empty());
-            if (actionNameAndDuration.isPresent()) {
-                ActionProfile presentActionNameAndDuration = actionNameAndDuration.get();
-                log.info("Call to TimeLockService#{} took {}. This was the slowest operation that completed"
-                                + " in the last {}.",
-                        SafeArg.of("actionName", presentActionNameAndDuration.actionName()),
-                        SafeArg.of("duration", presentActionNameAndDuration.duration()),
-                        SafeArg.of("timeWindow", LOGGING_TIME_WINDOW));
+            Optional<ActionProfile> actionProfile = slowestOperation.getAndSet(Optional.empty());
+            if (actionProfile.isPresent()) {
+                ActionProfile presentActionProfile = actionProfile.get();
+                logger.info("Call to TimeLockService#{} took {}. This was the slowest operation that completed"
+                                + " in the last {}. The operation completed with outcome {} - if it failed, the error"
+                                + " was {}.",
+                        SafeArg.of("actionName", presentActionProfile.actionName()),
+                        SafeArg.of("duration", presentActionProfile.duration()),
+                        SafeArg.of("timeWindow", LOGGING_TIME_WINDOW),
+                        SafeArg.of("outcome", presentActionProfile.failure().isPresent() ? "fail" : "success"),
+                        UnsafeArg.of("failure", presentActionProfile.failure()));
             }
         }
     }
--- a/lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
@@ -1,0 +1,188 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import java.io.Closeable;
+import java.time.Duration;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+import org.immutables.value.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Stopwatch;
+import com.google.common.util.concurrent.RateLimiter;
+import com.palantir.lock.v2.LockImmutableTimestampResponse;
+import com.palantir.lock.v2.LockRequest;
+import com.palantir.lock.v2.LockResponse;
+import com.palantir.lock.v2.LockToken;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.lock.v2.WaitForLocksRequest;
+import com.palantir.lock.v2.WaitForLocksResponse;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.timestamp.TimestampRange;
+
+/**
+ * Logs operations on an underlying {@link TimelockService} that take a long time (defined as longer than
+ * {@link ProfilingTimelockService#SLOW_THRESHOLD}). This logger is configured not to record more than one
+ * operation every {@link ProfilingTimelockService#LOGGING_TIME_WINDOW}. It will log what the longest operation that
+ * completed in the past {@link ProfilingTimelockService#LOGGING_TIME_WINDOW} window was, and how long it took. If no
+ * operation took longer than {@link ProfilingTimelockService#SLOW_THRESHOLD} it will not log.
+ *
+ * The {@link ProfilingTimelockService} does not cover specific operations which are at risk for taking long
+ * times owing to contention - in particular, this includes {@link #lock(LockRequest)} and
+ * {@link #waitForLocks(WaitForLocksRequest)}.
+ *
+ * Profiling is done explicitly at this level (and not at {@link com.palantir.lock.v2.TimelockRpcClient} level)
+ * to reflect the impact of timelock operations and cluster state changes (e.g. leader elections, rolling bounces) on
+ * clients.
+ */
+public class ProfilingTimelockService implements AutoCloseable, TimelockService {
+    private static final Logger log = LoggerFactory.getLogger(ProfilingTimelockService.class);
+
+    private static final Duration SLOW_THRESHOLD = Duration.ofSeconds(1);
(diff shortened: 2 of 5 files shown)


lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,228 @@
...
+import com.palantir.lock.StringLockDescriptor;
+import com.palantir.lock.v2.LockRequest;
+import com.palantir.lock.v2.TimelockService;
+import com.palantir.lock.v2.WaitForLocksRequest;
+import com.palantir.logsafe.Arg;
+
+public class ProfilingTimelockServiceTest {
+    private static final Duration SHORT_DURATION = ProfilingTimelockService.SLOW_THRESHOLD.dividedBy(5);
+    private static final Duration LONG_DURATION = ProfilingTimelockService.SLOW_THRESHOLD.multipliedBy(5);
+    private static final Duration TWO_CENTURIES = Duration.ofDays(365 * 100 * 2 + 50 - 2); // most of the time
```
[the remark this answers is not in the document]
jeremyk-91: Changed. Yeah, I vaguely remember this existing when trying to type `Duration.ofCenturies`, but that didn't work.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,228 @@
...
+    private final Ticker ticker = new Ticker() {
+        @Override
+        public long read() {
+            clock += operationTiming;
+            return clock;
+        }
+    };
+
+    private final AtomicBoolean allowLogging = new AtomicBoolean();
+    private final BooleanSupplier loggingPermissionSupplier = () -> allowLogging.getAndSet(false);
```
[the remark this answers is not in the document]
jeremyk-91: Yes, that's what it means. I don't have a strong opinion on this - it's just existing for this test. I will rename the method to be more explicit

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    @SuppressWarnings("unchecked") // Captors of generics
+    private void verifyLoggerInvokedWithSpecificProfile(String operation, Duration duration) {
+        // XXX Maybe there's a better way? The approaches I tried didn't work because of conflict with other methods
+        ArgumentCaptor<Arg<String>> messageCaptor = ArgumentCaptor.forClass(Arg.class);
+        ArgumentCaptor<Arg<Duration>> durationCaptor = ArgumentCaptor.forClass(Arg.class);
+        verify(logger).info(
+                any(String.class), messageCaptor.capture(), durationCaptor.capture(), any(), any(), any(), any());
+
```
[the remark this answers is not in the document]
jeremyk-91: Ah, yeah it can be. I wrote this first when I was thinking of verifying more complicated stuff on the operations

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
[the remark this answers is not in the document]
felixdesouza: I think that's weird, if you're explicit in turning logging on, you should be explicit in turning it off.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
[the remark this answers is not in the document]
jeremyk-91: Ah wait. It can't be that, because the params are Arg<String> while I'm testing that the value matches.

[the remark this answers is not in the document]
felixdesouza: I'm still not following, can you not just pass in an `Arg` to `eq(durationArg)`?

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -0,0 +1,227 @@
...
+                        SafeArg.of("duration", presentActionProfile.duration()),
+                        SafeArg.of("timeWindow", LOGGING_TIME_WINDOW),
+                        SafeArg.of("outcome", presentActionProfile.failure().isPresent() ? "fail" : "success"),
+                        UnsafeArg.of("failure", presentActionProfile.failure()),
+                        SafeArg.of("stacktrace", new RuntimeException().getStackTrace()));
+            }
+        }
+    }
+
+    private boolean stopwatchDescribesSlowOperation(Stopwatch stopwatch) {
```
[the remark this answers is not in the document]
jeremyk-91: Added.

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -0,0 +1,227 @@
...
+        stopwatch.stop();
+        accumulateSlowOperationTracking(actionName, stopwatch, failure);
+        tryFlushLogs();
+    }
+
+    private void accumulateSlowOperationTracking(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
+        if (stopwatchDescribesSlowOperation(stopwatch)) {
+            slowestOperation.accumulateAndGet(
+                    Optional.of(ImmutableActionProfile.of(actionName, stopwatch.elapsed(), failure)),
+                    (existing, update) -> !existing.isPresent()
```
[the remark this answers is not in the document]
jeremyk-91: I've split the absence condition to be its own thing:

```
                        if (!existing.isPresent()) {
                            return update;
                        }
                        return existing.get().duration().compareTo(stopwatch.elapsed()) < 0 ? update : existing;
```

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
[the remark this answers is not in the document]
jeremyk-91: Yeah, I wouldn't keep it in its current state. `allowLogging()` here probably should have been named `allowOneInstanceOfLogging()` (it's basically the analogue of releasing a rate limiter permit). I've switched to explicitly disabling, as only one test depends on this behaviour

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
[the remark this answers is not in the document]
jeremyk-91: You can, but `Arg` equality requires the names and log-safety of the arguments to match what's specified so it's not the same check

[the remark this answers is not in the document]
jeremyk-91: (If you feel strongly about this we can change it, though to me neither solution is clearly preferable over the other)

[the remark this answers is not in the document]
felixdesouza: I see, I suppose it's fine, I was tempted to say add an interface, but I don't think it's worth the effort.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+        allowLogging();
+        profilingTimelockService.getFreshTimestamp();
+
+        verify(delegate).getFreshTimestamp();
+        verifyLoggerInvokedWithSpecificProfile("getFreshTimestamp", LONG_DURATION);
+    }
+
+    @Test
+    public void doesNotLogIfLoggingNotAllowedEvenIfOperationsAreSlow() {
+        makeOperationsTakeSpecifiedDuration(LONG_DURATION);
```
felixdesouza: nit: explicitly add `disallowLogging`

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Test
+    public void logsIfFailedOperationsAreSlow() {
+        RuntimeException exception = new RuntimeException("kaboom");
+        when(profilingTimelockService.getFreshTimestamp()).thenThrow(exception);
+
+        makeOperationsTakeSpecifiedDuration(LONG_DURATION);
+        allowLogging();
+        assertThatThrownBy(profilingTimelockService::getFreshTimestamp).isEqualTo(exception);
```
felixdesouza: want to check that exception is logged

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -0,0 +1,229 @@
...
+        accumulateSlowOperationTracking(actionName, stopwatch, failure);
+        tryFlushLogs();
+    }
+
+    private void accumulateSlowOperationTracking(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
+        if (stopwatchDescribesSlowOperation(stopwatch)) {
+            slowestOperation.accumulateAndGet(
+                    Optional.of(ImmutableActionProfile.of(actionName, stopwatch.elapsed(), failure)),
+                    (existing, update) -> {
+                        if (!existing.isPresent()) {
```
felixdesouza: alternatively, feel free to ignore, not super convinced myself:

```java
existing.filter(e -> e.duration.compareTo(stopwatch.elapsed()) >= 0)
  .orElse(update)
```

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+    private final AtomicBoolean allowLogging = new AtomicBoolean();
+    private final BooleanSupplier loggingPermissionSupplier = allowLogging::get;
+
+    private final ProfilingTimelockService profilingTimelockService = new ProfilingTimelockService(
+            logger, delegate, () -> Stopwatch.createStarted(ticker), loggingPermissionSupplier);
+
+    @Test
+    public void doesNotLogIfOperationsAreFast() {
+        makeOperationsTakeSpecifiedDuration(SHORT_DURATION);
+        allowLogging();
```
felixdesouza: this means, "flush on each call" effectively?

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+        makeOperationsTakeSpecifiedDuration(LONG_DURATION);
+        for (int i = 0; i < 100; i++) {
+            profilingTimelockService.lockImmutableTimestamp();
+        }
+
+        makeOperationsTakeSpecifiedDuration(SHORT_DURATION);
+        allowLogging();
+        profilingTimelockService.startIdentifiedAtlasDbTransaction();
+
+        verifyLoggerInvokedWithSpecificProfile("lockImmutableTimestamp", LONG_DURATION);
```
felixdesouza: have to verify that it was only done once?, I can't remember if calling `verify` by is saying at least once or exactly once.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
jeremyk-91: yeah, makes sense

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+    private long operationTiming = 0;
+    private final Ticker ticker = new Ticker() {
+        @Override
+        public long read() {
+            clock += operationTiming;
+            return clock;
+        }
+    };
+
+    private final AtomicBoolean allowLogging = new AtomicBoolean();
```
felixdesouza: I'm somewhat struggling to make sense of the tests. I'm not sure whether having methods like:

```java
flushLogsWithCall(LONG_DURATION, profilingTimelockService::getFreshTimestamp);
accumulateLogsWithCall(SHORT_DURATION, profilingTimelockService::getFreshTimestamp);
```

they'll still do the same things, I just think it's closer to the behaviour you're trying to achieve,

and assertions will be somewhat clearer i.e.

```java
accumulateLogsWithCall(...) // call 1
...
accumulateLogsWithCall(...) // call n-1
flushLogsWithCall(...) // call n
assert based on the slowest call between 1 and n
```

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+        profilingTimelockService.lock(LockRequest.of(ImmutableSet.of(StringLockDescriptor.of("mortice")), 2366));
+
+        verifyLoggerInvokedWithSpecificProfile("getFreshTimestamp", LONG_DURATION);
+    }
+
+    @Test
+    public void logsOnlyOnceEvenIfManyRequestsAreSlow() {
+        makeOperationsTakeSpecifiedDuration(LONG_DURATION);
+        for (int i = 0; i < 100; i++) {
+            profilingTimelockService.lockImmutableTimestamp();
```
felixdesouza: what about different calls, but same duration? do we care about that?

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+        verifyLoggerInvokedWithSpecificProfile("lockImmutableTimestamp", LONG_DURATION);
+        verifyNoMoreInteractions(logger);
+    }
+
+    @Test
+    public void logsMultipleTimesIfAllowedToMultipleTimes() {
+        makeOperationsTakeSpecifiedDuration(LONG_DURATION);
+        for (int i = 0; i < 100; i++) {
+            allowLogging();
```
felixdesouza: I think this is where the `accumulateLogsWithCall` and `flushLogsWithCall` makes things a *lot* clearer

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
jeremyk-91: Hmm, we need this to return `Optional<MaximumThing>` so we can't quite use an `orElse`. We could do something like

```
Stream.of(existing, update)
                            .filter(Optional::isPresent)
                            .map(Optional::get)
                            .max(Comparator.comparing(ActionProfile::duration))
```

but I think I'll stick with what we have

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
jeremyk-91: Yep - `verify(foo).bar()` unqualified means exactly once.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
jeremyk-91: sure, we can change that

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
jeremyk-91: We could do an 'any one of' check. I didn't intend there to be any guarantees as to which one would be returned in that case

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        verifyLoggerNeverInvoked();
+    }
+
+    @Test
+    public void logsSlowestOperationIfMultipleOperationsExceedTheSlowThresholdIfFirst() {
+        accumulateLogsWithCall(TWO_CENTURIES, profilingTimelockService::getFreshTimestamp);
+        flushLogsWithCall(TWO_CENTURIES, profilingTimelockService::startIdentifiedAtlasDbTransaction);
+
+        verifyLoggerInvokedOnceWithSpecificProfile("getFreshTimestamp", TWO_CENTURIES);
```
felixdesouza: still confused as to how you select `getFreshTimestamp`? is it just the first one?

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -0,0 +1,229 @@
...
+                ActionProfile presentActionProfile = actionProfile.get();
+                logger.info("Call to TimeLockService#{} took {}. This was the slowest operation that completed"
+                                + " in the last {}. The operation completed with outcome {} - if it failed, the error"
+                                + " was {}. We include a stack trace of the call: {}",
+                        SafeArg.of("actionName", presentActionProfile.actionName()),
+                        SafeArg.of("duration", presentActionProfile.duration()),
+                        SafeArg.of("timeWindow", LOGGING_TIME_WINDOW),
+                        SafeArg.of("outcome", presentActionProfile.failure().isPresent() ? "fail" : "success"),
+                        UnsafeArg.of("failure", presentActionProfile.failure()),
+                        SafeArg.of("stacktrace", new RuntimeException().getStackTrace()));
```
felixdesouza: somewhat worried that a fast operation may be slowed down by a previous window with slow calls when they have to dump a stack trace, unclear how expensive it is.

docs/source/release_notes/release-notes.rst
```diff
@@ -103,6 +103,11 @@ develop
          - Relaxed concurrency model of ``MetricsManager`` allowing for more concurrency.
            (`Pull Request <https://github.com/palantir/atlasdb/pull/4098>`__)
 
+    *    - |improved|
+         - Clients now log at most once every 10 seconds when a ``TimelockService`` level request is slow.
```
felixdesouza: nit: if you could make a changelog in the new form, would be grand

jeremyk-91:
I think currently the tie breaker is the first to be registered, though I wasn't working under the assumption that one can pick which one is returned (just that it's _one of_ them).

I'll investigate and manually add the changelogs.

lock-api/src/test/java/com/palantir/lock/client/ProfilingTimelockServiceTest.java
jeremyk-91: ugh, this test was badly copied. The second call is supposed to be `LONG_DURATION`.

For this case it'll always be the first (because of implementation details), but the intention was never to assert that

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
jeremyk-91: Hmm, yeah on second thoughts I'll remove this. If we really need to investigate just get someone on the box

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private void trackActionAndMaybeLog(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
+        stopwatch.stop();
+        accumulateSlowOperationTracking(actionName, stopwatch, failure);
+        tryFlushLogs();
+    }
+
+    private void accumulateSlowOperationTracking(String actionName, Stopwatch stopwatch, Optional<Exception> failure) {
+        if (stopwatchDescribesSlowOperation(stopwatch)) {
+            slowestOperation.accumulateAndGet(
```
felixdesouza: should have mentioned this earlier, do we want to add a meter as well? or something that we can get a dashboard of more easily?

jeremyk-91: I think in general you have `TimeLockService.<op>` for individual operations (though the granularity there is not as fine, and you also can't check failures). I'd leave this for a separate PR if we wanted it?

jeremyk-91:
As per offline discussion - adding meter for number of slow ops, I think there was a bit of confusion earlier

PR #4111, opened 2019-07-05
Require Users to Explicitly Specify <3 Node TimeLock

jeremyk-91:
**Goals (and why)**:
- We've run into cases internally where users have been burned by the failures of single node TimeLocks and they didn't know they were doing something dangerous. This should stop that.

- If a flag is not specified, fail on startup in 1- and 2-node configurations on the timelock _server_ side.

- Added tests on the interaction between the flag and the number of servers.

- User break for users where the cluster size is <3, so the relevant release of TimeLock needs to be signposted clearly, or at the very least the AtlasDB support POC will need to be aware that this is going out so that they know what to do should users have questions.

**Where should we start reviewing?**: ClusterConfiguration

**Priority (whenever / two weeks / yesterday)**: today, or early next week

@j-baker for SA


jeremyk-91:
Note: Before this goes out to internal deployments we should send comms around. The process has been kicked off, but we should make sure people are aware that this is happening and have had a bit of time to react to it.

sandorw:
Be careful of the implications of a break for versioning if this gets automatically released.

timelock-server/src/test/java/com/palantir/atlasdb/timelock/config/TimeLockServerConfigurationTest.java
```diff
@@ -41,6 +41,7 @@
     private static final ClusterConfiguration CLUSTER = ImmutableClusterConfiguration.builder()
             .localServer(ADDRESS)
             .addServers(ADDRESS)
+            .enableNonstandardAndPossiblyDangerousTopology(true) // Tests
```
felixdesouza: just remove the comment

felixdesouza:
wdym @sandorw ?

felixdesouza:
oh re autorelease, I believe it should be fine considering we're still 0.x and I believe the bot respects that.

jeremyk-91:
I think Sandor is worried that the autorelease bot may bump the major version. Since it's 0 it shouldn't though

PR #4117, opened 2019-07-09
Fix batching logic for TS

gmaretic:
**Goals (and why)**:
When encountering an entry with commit TS after sweep TS, we would thing there is more to sweep, and aggressively re-read it. Along the way made it so that the last swept timestamp is set to just before the entry that cannot be processed, and used this information to inform the multi-batching.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepBatch.java
```diff
@@ -29,15 +29,22 @@
     List<WriteInfo> writes();
     DedicatedRows dedicatedRows();
     long lastSweptTimestamp();
+    boolean hasNext();
```
jeremyk-91: This is true if for the last batch shoved into this accumulator, we weren't able to process everything at the given sweep timestamp. Technically speaking it should probably be `couldHaveNext()` or a negative that is `noMoreSweepableAtThisTimestamp()`, but this is fine.

PR #4118, opened 2019-07-10
Deprecate AtlasSchema

pkoenig10:
**Goals (and why)**:
Schema classes do not need to extend `AtlasSchema`. It's essentially unused (it would be trivial to remove it in the few places it is used). We should mark it as deprecated to discourage continued usage of it.

- Is this class actually useful?
- I originally planned to stop using it internally to follow best practices, but this is technically a dev break since the static instances were public. I'm happy to remove these uses if we feel the minor dev break is worth it.


clockfort:
pg does use this a decent bit, and does have stuff like List<AtlasSchema> that hold all of our schemas to iterate through them and initialize them etc, but would also be fine with owning this class if you want to get rid of it upstream.

gmaretic:
I don't think this is worth the potential breaks due to errors on deprecation

PR #4119, opened 2019-07-10
Replace java plugin with java-library plugin

pkoenig10:
**Goals (and why)**:
AtlasDB leaks a lot of dependencies to the compile classpath of users. Users then incorrectly use these transitive dependencies without declaring the dependency themselves.

This PR is a first step towards cleaning that up. In follow up PRs we can move internal dependencies to the `implementation` configuration.

- This PR also removes some redundant `apply plugin` statements (the removed plugins are applied by `shared.gradle`)

### DIFF
--- a/atlasdb-autobatch/build.gradle
+++ b/atlasdb-autobatch/build.gradle
@@ -3,9 +3,6 @@ apply plugin: 'org.inferred.processors'
 apply from: "../gradle/publish-jars.gradle"
 apply from: "../gradle/shared.gradle"
 
-apply plugin: 'java'
-apply plugin: 'eclipse'
-
 repositories {
     mavenCentral()
 }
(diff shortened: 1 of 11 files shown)


felixdesouza:
Is this safe/does this introduce a break?

Particularly the part about separation of compilation classpath and runtime classpath. It's a good step forward, but if it's a break, I'd recommend we flag it as such/probably send out comms to internal products with how to resolve it, since it'd be a pretty annoying+large break.

pkoenig10:
This shouldn't have any effect until we actually move a dependencies to the `implementation` configuration. Furthermore, consumers have to explicitly enable compile/runtime separation when consuming POMs (either by enabling `IMPROVED_POM_SUPPORT` or using Gradle 5.0+).

@iamdanfox can keep me honest here.

felixdesouza:
Fair enough, I ran into the latter when doing gradle 5 upgrades and wasn't sure whether it was tied to gradle 5 or java-library. Thanks for investigating!

PR #4120, opened 2019-07-10
getRowsColumnRangeIterator

sandorw:
**Goals (and why)**:
Adds a new method to tables `getRowsColumnRangeIterator` that is similar to `getRowsColumnRange`, except it returns an `Iterator` instead of a `BatchingVisitable`.

Mostly modifications to generated tables that are checked in.

Copied existing tests using `getRowsColumnRange`. I didn't want to combine the testing in case any implementation changes in the future, and the additional tests should be cheap. The contract should be the same, so existing tests should cover the functionality.

Skip all the generated files, everything else should be relevant and fairly quick.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -16,11 +16,13 @@
 package com.palantir.atlasdb.transaction.impl;
 
 import java.nio.ByteBuffer;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NavigableMap;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
@@ -202,6 +204,44 @@ public class SerializableTransaction extends SnapshotTransaction {
             throw new UnsupportedOperationException("This method does not support serializable conflict handling");
         }
         return super.getRowsColumnRange(tableRef, rows, columnRangeSelection, batchHint);
+    }
+
+    @Override
+    public Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> getRowsColumnRangeIterator(
+            TableReference tableRef,
+            Iterable<byte[]> rows,
+            BatchColumnRangeSelection columnRangeSelection) {
+        Map<byte[], Iterator<Entry<Cell, byte[]>>> ret =
+                super.getRowsColumnRangeIterator(tableRef, rows, columnRangeSelection);
+        return Maps.transformEntries(ret, (row, iterator) -> new Iterator<Entry<Cell, byte[]>>() {
+            Entry<Cell, byte[]> next = null;
+
+            @Override
+            public boolean hasNext() {
+                if (next != null) {
+                    return true;
+                }
+
+                if (iterator.hasNext()) {
+                    next = iterator.next();
+                    markRowColumnRangeRead(tableRef, row, columnRangeSelection, Collections.singletonList(next));
+                    return true;
+                }
+
+                reachedEndOfColumnRange(tableRef, row, columnRangeSelection);
+                return false;
+            }
+
+            @Override
+            public Entry<Cell, byte[]> next() {
+                if (!hasNext()) {
+                    throw new NoSuchElementException();
+                }
+                Entry<Cell, byte[]> result = next;
+                next = null;
+                return result;
+            }
+        });
     }
 
     @Override
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
@@ -38,6 +38,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Ordering;
 import com.google.common.primitives.UnsignedBytes;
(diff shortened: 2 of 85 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -670,8 +672,19 @@ public void readsFromThoroughlySweptTableShouldFailWhenLocksAreInvalid() {
 
         tasks.add(Pair.of("getRowsColumnRange(TableReference, Iterable<byte[]>, ColumnRangeSelection, int)",
                 (t, heldLocks) -> {
-                    t.getRowsColumnRange(TABLE_SWEPT_THOROUGH, Collections.singleton(PtBytes.toBytes("row1")),
-                            new ColumnRangeSelection(null, null), batchHint);
+                    Iterators.getLast(
```
felixdesouza: why `getLast`?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
```diff
@@ -745,7 +823,38 @@ public void testReadMyWritesColumnRangePagingTransaction() {
     }
 
     @Test
-    public void testReadMyDeletesColumnRangePagingTransaction() {
+    public void testReadMyWritesColumnRangePagingTransaction_iterator() {
+        Transaction t = startTransaction();
```
felixdesouza: a better/easier/clearer test would be doing the setup as normal, and then assert in the batching visitable view, and then do the iterator view, and then assert that the contents are the same? at the same time, not too fussed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -204,6 +206,44 @@ public SerializableTransaction(MetricsManager metricsManager,
         return super.getRowsColumnRange(tableRef, rows, columnRangeSelection, batchHint);
     }
 
+    @Override
+    public Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> getRowsColumnRangeIterator(
+            TableReference tableRef,
+            Iterable<byte[]> rows,
+            BatchColumnRangeSelection columnRangeSelection) {
+        Map<byte[], Iterator<Entry<Cell, byte[]>>> ret =
```
felixdesouza: if this is identical to `getRowsColumnRange` in terms of functionality, should this also be unsupported if serializable?

sandorw: good point - let me revisit it to understand why we had to do this in the first place

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
sandorw: yeah, got stuck in this pattern after touching a few tests where it was not appropriate to combine the two

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
sandorw: I want to drain the iterator to verify that we've actually gone through with the operation, otherwise we might have done nothing if the iterator was sufficiently lazy - this mimics what the other `getRowsColumnRange` test here does

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
sandorw: I consolidated in the few cases where it was safe to do so without making the test harder to understand or where the it was important to the test itself

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
sandorw: So it's technically doable, with some caveats, and since no one was using the method with serializable transactions, we chose not to implement it.

The other implementations allow us to track which range you've read as you progress through the BatchingVisitable or Iterator, including marking that you've read the end of the row past the final key if `hasNext()` is false or you visit the last element.

For this method, since we've already combined the results of all the iterators into a single iterator, we can't really reconstruct that data (is this the last key in the row?) without eventually draining the entire iterator anyway.

sandorw: Above comment was about the other method, summary is that we handle it correctly in the new method.

sandorw:
Added a few more tests to explicitly cover the case where we do read a result from the row, but only conflict with another writer if we attempt to read past the first result.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -708,7 +936,26 @@ public void testMultipleReadsToSameColumnRange() {
...
+        Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> columnRange =
+                t1.getRowsColumnRangeIterator(TEST_TABLE, ImmutableList.of(row),
+                        BatchColumnRangeSelection.create(PtBytes.toBytes("col"), PtBytes.toBytes("col0"), 1));
+        assertThat(Iterables.getOnlyElement(columnRange.values()).hasNext()).isFalse();
+        Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> columnRangeAgain =
+                t1.getRowsColumnRangeIterator(TEST_TABLE, ImmutableList.of(rowDifferentReference),
+                        BatchColumnRangeSelection.create(PtBytes.toBytes("col"), PtBytes.toBytes("col0"), 1));
+        assertThat(Iterables.getOnlyElement(columnRangeAgain.values()).hasNext()).isFalse();
+        put(t1, "mutation to ensure", "conflict", "handling");
+        t1.commit();
```
felixdesouza: is this test incomplete?

felixdesouza: or is it that you expect this to pass without issue?

PR #4123, opened 2019-07-11
Move some dependencies to the implementation configuration

pkoenig10:
**Goals (and why)**:
Follow up to #4119. Goal is to remove some dependencies that should not be transitively exposed to consumers. 

The changes in this PR are pretty conservative and the modified dependencies are almost certainly not used in any API surface. The modified dependencies are:
- `com.google.code.findbugs:findbugs-annotations`
- `com.google.code.findbugs:jsr305`
- `com.github.rholder:guava-retrying`
- `com.palantir.common:streams`
- `com.palantir.safe-logging:preconditions`
- `com.palantir.safe-logging:safe-logging`
- `org.squareup:javapoet`

- Are any of these dependencies exposed in a project's API?

### DIFF
--- a/atlasdb-autobatch/build.gradle
+++ b/atlasdb-autobatch/build.gradle
@@ -11,8 +11,9 @@ libsDirName = file('build/artifacts')
 dependencies {
     compile project(":atlasdb-commons")
     compile group: 'com.lmax', name: 'disruptor'
-    compile group: 'com.palantir.safe-logging', name: 'safe-logging'
     compile group: 'com.palantir.tritium', name: 'tritium-registry'
+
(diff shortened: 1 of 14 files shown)


j-baker:
not immediately thrilled with this idea because Atlas relies on people being able to continually take new releases without making changes, and so if we do things that start causing dev breaks, that stops our ability to get automerged and thus our ability to stay up-to-date.

gmaretic:
@jeremyk-91 what's the verdict on this? Do we want to potentially break consumers in order to reduce our footprint?

pkoenig10:
@jeremyk-91 any update here?

Issue #4112, opened 2019-07-05
Timelock Partitioning refactors after merges


felixdesouza:
* swap arguments of `WithSeq<T>#of(long seq, T value)` to the other way around.
* get rid of `defaultValue`/document current behaviour

gmaretic:
👍

felixdesouza:
encapsulate `SetMultimap<Client, WithSeq<T>>` into some object that also checks constraints of seq uniqueness for a given client etc.

PR #4126, opened 2019-07-15
[Timelock Partitioning] Part 12: Unify autobatcher exception handling + Close autobatchers

felixdesouza:
**Goals (and why)**:
Due to splitting out the two impls i.e. `Learner` and `Acceptor`, exception handling wasn't handled correctly, and effectively diverged as part of the review iteration. Now I'm just closing the loop.

Similarly, even though it shouldn't matter *too* much, we want the ability to close autobatchers to claim resources back.

Ticks some of the checkboxes in #4112.

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatcherExecutionExceptions.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatcherExecutionExceptions.java
@@ -18,6 +18,8 @@ package com.palantir.atlasdb.timelock.paxos;
 
 import java.util.concurrent.ExecutionException;
 
+import com.palantir.logsafe.Preconditions;
+
 class AutobatcherExecutionExceptions {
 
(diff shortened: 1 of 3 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatcherExecutionExceptions.java
```diff
@@ -30,7 +30,7 @@ static RuntimeException handleAutobatcherExceptions(Exception e) {
             }
             throw new RuntimeException(cause);
         } else {
-            // handle interrupted
```
jeremyk-91: Yep, this is correct. Would it be worth having some preconditions check on the exception (that it is an `InterruptedException`)? - in case someone later on tries to re-use this and passes in some random other exception type

felixdesouza: sure

PR #4127, opened 2019-07-15
[Timelock Partitioning] Part 13: Autobatching `PingableLeader`

felixdesouza:
**Goals (and why)**:
With number of leaders going from 1:n clients to 1:1, the number of requests that go to the same node increase by the same amount, hampering the batching effort everywhere else.

Implemented similar to `PaxosAcceptor` and `PaxosLearner` *before* the `PaxosQuorumChecker` modifications. Have a factory that spawns client specific implementations of `PingableLeader`.

* Should I just use a `CoalescingSupplier` for some of them?

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/CoalescingRequestSupplier.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/CoalescingRequestSupplier.java
@@ -1,0 +1,43 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.autobatch;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import com.google.common.base.Preconditions;
+
+public final class CoalescingRequestSupplier<T> implements CoalescingRequestFunction<Autobatchers.SupplierKey, T> {
+
+    private final Supplier<T> supplier;
+
+    public CoalescingRequestSupplier(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    @Override
+    public T defaultValue() {
+        throw new AssertionError();
+    }
+
+    @Override
+    public Map<Autobatchers.SupplierKey, T> apply(Set<Autobatchers.SupplierKey> request) {
+        Preconditions.checkArgument(request.size() == 1, "invalid request");
+        return Autobatchers.SupplierKey.wrap(supplier.get());
+    }
+}
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -19,9 +19,11 @@ package com.palantir.atlasdb.autobatch;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.io.Closeable;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.function.Function;
 
+import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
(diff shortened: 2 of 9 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -36,10 +38,19 @@
...
+     */
+    private static final ConcurrentMap<String, ThreadFactory> threadFactories = Maps.newConcurrentMap();
 
     private static ThreadFactory threadFactory(String safeLoggablePurpose) {
-        String namePrefix = String.format("autobatcher-%s-", safeLoggablePurpose);
+        return threadFactories.computeIfAbsent(safeLoggablePurpose, DisruptorAutobatcher::createThreadFactory);
+    }
+
+    private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
+        String namePrefix = String.format("autobatcher.%s-", safeLoggablePurpose);
```
jeremyk-91: nit: I think this change from `-` to `.` is intentional, given what you're doing in `AutobatchingPaxosPingableLeaderFactory`. This seems good to allow clear delimiting between multi-word phrases. In that case, `BatchingPaxosAcceptorFactory` should probably be updated to use the new format as well (as in `batch-paxos-acceptor.latest-sequence-cache`, for instance)

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -36,10 +38,19 @@
  */
 public final class DisruptorAutobatcher<T, R>
         implements AsyncFunction<T, R>, Function<T, ListenableFuture<R>>, Closeable {
-    private static final int BUFFER_SIZE = 1024;
+
+    /*
+        By memoizing thread factories per loggable purpose, the names are correct for multiple instances of the same
+        autobatcher function.
```
jeremyk-91: I agree. Would be worth saying what `correct` means, though (i.e. threads are numbered uniquely across different instances)

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/CoalescingRequestSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.palantir.atlasdb.autobatch;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import com.google.common.base.Preconditions;
+
+public final class CoalescingRequestSupplier<T> implements CoalescingRequestFunction<Autobatchers.SupplierKey, T> {
```
jeremyk-91: OK. So the main difference I can see between this and `CoalescingSupplier` is that in `CoalescingSupplier` one of the caller threads runs the operation, while with this one we delegate to the specific autobatcher thread. There are pros and cons - this may give better visibility into what work is being done (e.g. if one jstacks), though I'd guess that the other way has less overhead if the operation isn't called very often (since the caller thread will just run the operation without any JVM context switches).

I'm not sure we really need this over `CoalescingSupplier` (as you question in the comments), though if we think the visibility is useful / there's something I'm missing, would be good to clarify that.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPaxosPingableLeaderFactory.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.Closeable;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
+import com.palantir.atlasdb.autobatch.Autobatchers;
+import com.palantir.atlasdb.autobatch.Autobatchers.SupplierKey;
+import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
+import com.palantir.leader.PingableLeader;
+
+public class AutobatchingPaxosPingableLeaderFactory implements Closeable {
```
jeremyk-91: nit: I guess it's named this way because the package has `paxos` in it, but it doesn't look like the class is specific to Paxos leaders

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/CoalescingRequestSupplier.java
felixdesouza: jstacks would be equivalent in either case. Work would be done in the autobatcher thread, and the caller thread would be blocked waiting on the future from autobatcher. Whereas in `CoalescingSupplier` it would wait on the result of the internal future.

My only concern that comes to mind from when we were doing the timelock stability stuff is the `fairLock`, which seems to be essential to get the correctness guarantees of `CoalescingSupplier` but seems too coarse. Initially we had a coalescing supplier that did the work of `blockOnBecomingLeader`. When I moved it to autobatchers, it was markedly better. The only reasoning that made sense was that threads were trying to wake up to respond to the fact that a result had been published, but because they weren't in order, they'd go back to sleep again waiting to woken up in the correct order. This then sucks because we have one result that needs to be shared across n callers, but they're effectively sorting amongst themselves. More than happy to be wrong here, but I couldn't explain the difference any other way unfortunately.

Initially indifferent, however the above concern shifts it in favour of using autobatcher.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPaxosPingableLeaderFactory.java
felixdesouza: fair, this was just muscle memory tbh aha

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/CoalescingRequestSupplier.java
jeremyk-91: Ah, I didn't see the *fair* lock - that makes sense then, it usually imposes quite a bit of a performance overhead. The disruptor-based concurrency model seems like it would work better then.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
felixdesouza: done, eagle eyed! 🦅 👁

felixdesouza:
This is not in its ideal state, since there is still some thread explosion due to a PQC-style request happening inside `PaxosLeaderElectionService`. That however is a bigger change and probably easier to iterate/review/get other PRs in when this is in for the mean time.

PR #4128, opened 2019-07-15
TS batch size metrics

gmaretic:
**Goals (and why)**:
Metrics for entries read in iterations of targeted sweep

Added some bookkeeping for number of entries before any filtering is performed to be consistent.

Added some unit tests and a couple to TargetedSweeperTest

We are registering a histogram somewhat unnecessarily, this also results in locking twice per metrics collection cycle. Should be fine, but pointing out.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -58,7 +58,6 @@ public final class AtlasDbMetricNames {
     public static final String SWEEP_TS = "sweepTimestamp";
     public static final String LAST_SWEPT_TS = "lastSweptTimestamp";
     public static final String LAG_MILLIS = "millisSinceLastSweptTs";
-    public static final String BATCH_SIZE = "batchSizeHistogram";
     public static final String BATCH_SIZE_MEAN = "batchSizeMean";
 
     public static final String SWEEP_OUTCOME = "outcome";
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -58,6 +58,8 @@ public final class AtlasDbMetricNames {
     public static final String SWEEP_TS = "sweepTimestamp";
     public static final String LAST_SWEPT_TS = "lastSweptTimestamp";
     public static final String LAG_MILLIS = "millisSinceLastSweptTs";
+    public static final String BATCH_SIZE = "batchSizeHistogram";
+    public static final String BATCH_SIZE_MEAN = "batchSizeMean";
 
     public static final String SWEEP_OUTCOME = "outcome";
     public static final String TAG_OUTCOME = "status";
--- a/atlasdb-commons/src/main/java/com/palantir/util/AggregatingVersionedSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/AggregatingVersionedSupplier.java
@@ -50,10 +50,10 @@ public class AggregatingVersionedSupplier<T> implements Supplier<VersionedType<T
     }
 
     public static <C extends Comparable<C>> AggregatingVersionedSupplier<C> min(long expirationMillis) {
-        return new AggregatingVersionedSupplier<>(AggregatingVersionedSupplier::minimum, expirationMillis);
+        return new AggregatingVersionedSupplier<>(AggregatingVersionedSupplier::min, expirationMillis);
     }
 
-    private static <C extends Comparable<C>> C minimum(Collection<C> currentValues) {
+    private static <C extends Comparable<C>> C min(Collection<C> currentValues) {
         return currentValues.stream().min(Comparator.naturalOrder()).orElse(null);
     }
 
--- a/atlasdb-commons/src/main/java/com/palantir/util/AggregatingVersionedSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/AggregatingVersionedSupplier.java
@@ -16,6 +16,7 @@
 package com.palantir.util;
 
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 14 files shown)


changelog/@unreleased/pr-4128.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: improvement
+improvement:
+  description: Targeted sweep metrics now expose a histogram `batchSizeHistogram`
```
felixdesouza: I think this will mess up changelog, although could be wrong. I believe you want:

```
description: |
  Targeted sweep metrics...
```

IIRC the `|` is necessary for multiline strings.

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
```diff
@@ -58,6 +58,8 @@ private CellFilterMetrics() {
     public static final String SWEEP_TS = "sweepTimestamp";
     public static final String LAST_SWEPT_TS = "lastSweptTimestamp";
     public static final String LAG_MILLIS = "millisSinceLastSweptTs";
+    public static final String BATCH_SIZE = "batchSizeHistogram";
```
jeremyk-91: We probably don't need to say it's a histogram; the metrics will have names like `blah.batchSize.p99` that seem to imply histograms

atlasdb-commons/src/main/java/com/palantir/util/AggregatingVersionedSupplier.java
```diff
@@ -48,6 +49,14 @@ public AggregatingVersionedSupplier(Function<Collection<T>, T> aggregator, long
                 .memoizeWithExpiration(this::recalculate, expirationMillis, TimeUnit.MILLISECONDS);
     }
 
+    public static <C extends Comparable<C>> AggregatingVersionedSupplier<C> min(long expirationMillis) {
+        return new AggregatingVersionedSupplier<>(AggregatingVersionedSupplier::minimum, expirationMillis);
+    }
+
+    private static <C extends Comparable<C>> C minimum(Collection<C> currentValues) {
+        return currentValues.stream().min(Comparator.naturalOrder()).orElse(null);
+    }
```
jeremyk-91: nit: probably just pick one of `min` or `minimum`?

We can implement this in log-time by maintaining a sorted multi-set, but probably fine as is :)

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsAssert.java
```diff
@@ -115,15 +141,15 @@ public void hasNotRegisteredTargetedOutcome(SweepStrategy strategy, SweepOutcome
         objects.assertNull(info, getGaugeForTargetedOutcome(strategy, outcome));
     }
 
-    private Gauge<Long> getGaugeConservative(String name) {
+    private <N> Gauge<N> getGaugeConservative(String name) {
```
jeremyk-91: I think this should be `private Gauge<?> getGaugeConservative(String name)`, it doesn't look like the type param is particularly helpful here

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsAssert.java
```diff
@@ -139,18 +165,37 @@ public void hasNotRegisteredTargetedOutcome(SweepStrategy strategy, SweepOutcome
                         AtlasDbMetricNames.TAG_STRATEGY, getTagForStrategy(strategy)));
     }
 
+    private Double getEntriesReadInBatchMeanConservative() {
+        return (Double) getGaugeConservative(AtlasDbMetricNames.BATCH_SIZE_MEAN).getValue();
+    }
+
+    private Double getEntriesReadInBatchMeanThorough() {
+        return (Double) getGaugeThorough(AtlasDbMetricNames.BATCH_SIZE_MEAN).getValue();
```
jeremyk-91: yeah, seems we still need to cast so the params above don't seem to be helpful, unless I'm missing something

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsAssert.java
```diff
@@ -103,6 +109,26 @@ public void hasMillisSinceLastSweptThoroughEqualTo(long value) {
         objects.assertEqual(info, getGaugeThorough(AtlasDbMetricNames.LAG_MILLIS).getValue(), value);
     }
 
+    public void hasEntriesReadInBatchConservativeEqualTo(long... outcomes) {
+        arrays.assertContainsExactlyInAnyOrder(info,
+                getBatchSizeHistogram(SweepStrategy.CONSERVATIVE).getSnapshot().getValues(),
+                outcomes);
+    }
```
jeremyk-91: nit: I'd probably name this just `hasEntriesReadInBatchConservative(...)` or even `containsEntriesReadInBatchConservative(...)` for similarity with the existing assertions. Might just be me, but `EqualTo` makes me expect a single following value (or that the thing is equal to all of the vararg parameters)

changelog/@unreleased/pr-4128.v2.yml
```diff
@@ -1,7 +1,6 @@
 type: improvement
 improvement:
-  description: Targeted sweep metrics now expose a histogram `batchSizeHistogram`
-    of the number of entries read in each iteration of targeted sweep, as well as
-    a gauge `batchSizeMean` that samples the mean of that histogram.
+  description: |
+    Targeted sweep metrics now expose a gauge `batchSizeMean` with the mean of the number of entries read in iterations of targeted sweep durinng a sliding time window of one hour.
```
jeremyk-91: nit: `durinng`

Also, the original format is actually allowed - see https://yaml.org/spec/1.2/spec.html#id2788859. Basically

```
key: value 123123123123
ohnoohnoohno
```

is broken, but the additional indent actually makes

```
key: value 123123123123
 ohnoohnoohno
```

acceptable.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsAssert.java
gmaretic: 👍

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/metrics/SweepMetricsAssert.java
gmaretic: Good call

PR #4129, opened 2019-07-15
[Timelock Partitioning] Part 14: Get rid of default value

felixdesouza:
**Goals (and why)**:
Semantics of the `CoalescingFunction` were too confusing. Removing it to simplify code. 

Came about as part of #4112.


felixdesouza:
blocked on #4127

PR #4132, opened 2019-07-16
[Timelock Partitioning] Part 16: Reuse paxos quorum checker for pingable leader get uuid

felixdesouza:
**Goals (and why)**:
Ran into the same problem with `PaxosAcceptor` and `PaxosLearner`. Have to defer creating threads to communicate to other nodes till *after* we've accumulated a batch of requests.

`PaxosQuorumChecker` is close, but turns out we have different exit constraints. Instead we modify `PaxosQuorumChecker` to take in a custom exit condition, with everything else defined in terms of new exit condition.

Alternatively, I could *copy* the entire impl of the `getUuid` code from `PaxosLeaderElectionService`, but it does mean that any bugs we have fixed in `PaxosQuorumChecker` never get fixed in this similar-but-different piece of code. So it makes sense to unify it.

Trying to avoid more generics, as it's quite hairy so far. 

This is a refactor, should be able to rely on existing tests for this.

* `PaxosResponseAccumulator`
* `PaxosLeaderElectionService`
* `PaxosQuorumChecker`

### DIFF
--- a/leader-election-impl/build.gradle
+++ b/leader-election-impl/build.gradle
@@ -12,6 +12,8 @@ dependencies {
   compile group: "commons-io", name: "commons-io"
   compile group: 'com.palantir.safe-logging', name: 'safe-logging'
 
+  implementation group: 'com.palantir.common', name: 'streams'
+
   processor group: 'org.immutables', name: 'value'
 
   testCompile(group: "org.jmock", name: "jmock", version: libVersions.jmock) {
--- a/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
@@ -15,17 +15,16 @@
  */
 package com.palantir.leader;
 
+import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -40,7 +39,6 @@ import com.google.common.base.Function;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.net.HostAndPort;
@@ -55,7 +53,10 @@ import com.palantir.paxos.PaxosLatestRoundVerifier;
 import com.palantir.paxos.PaxosLearner;
 import com.palantir.paxos.PaxosLearnerNetworkClient;
 import com.palantir.paxos.PaxosProposer;
+import com.palantir.paxos.PaxosQuorumChecker;
+import com.palantir.paxos.PaxosResponse;
 import com.palantir.paxos.PaxosResponses;
+import com.palantir.paxos.PaxosResponsesWithRemote;
 import com.palantir.paxos.PaxosRoundFailureException;
 import com.palantir.paxos.PaxosUpdate;
 import com.palantir.paxos.PaxosValue;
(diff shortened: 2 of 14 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -160,24 +171,26 @@ private PaxosQuorumChecker() {
      * @param request the request to make on each of the remote endpoints
      * @param quorumSize number of acknowledge requests after termination
      * @param executors run the requests
+     * @param shouldSkipNextRequest whether or not the next request should be skipped
```
gmaretic: You renamed the parameter, but not the javadoc above

leader-election-impl/build.gradle
```diff
@@ -12,6 +12,8 @@ dependencies {
   compile group: "commons-io", name: "commons-io"
   compile group: 'com.palantir.safe-logging', name: 'safe-logging'
 
+  implementation group: 'com.palantir.common', name: 'streams'
```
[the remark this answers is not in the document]
felixdesouza: related to #4123, it just means we don't leak as many dependencies on the compile classpath. Obviously, this doesn't even make a dent, but it's good practice. See [here](https://docs.gradle.org/current/userguide/java_library_plugin.html)

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -153,31 +164,33 @@ private PaxosQuorumChecker() {
 
     /**
      * Collects a list of responses from remote services.
-     * This method may short-circuit if a quorum can no longer be obtained (depending on the
-     * shortcircuitIfQuorumImpossible parameter) and cancels pending requests once a quorum has been obtained.
+     * This method may short-circuit depending on the {@code shouldSkipNextRequest} predicate parameter and cancels
```
jeremyk-91: nit: I'd name the param `shouldSkipRemainingRequests` or even just `shouldShortcircuit` or something, since it does a lot more than skip the next request...

leader-election-impl/src/main/java/com/palantir/paxos/PaxosResponsesWithRemote.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+import org.immutables.value.Value;
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.common.streams.KeyedStream;
+
+@Value.Immutable
+public abstract class PaxosResponsesWithRemote<S, T extends PaxosResponse> {
```
jeremyk-91: Hmm, I'd prefer we use `REMOTE` or something rather than `S`, also possibly have the params the other way around (the thing extending PaxosResponse first).

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -257,63 +260,40 @@ boolean getAndRecordLeaderPingResult(@Nullable Future<Boolean> pingFuture) throw
...
+                ImmutableList.copyOf(otherPotentialLeadersToHosts.keySet()),
+                pingableLeader -> new PaxosString(pingableLeader.getUUID()),
+                leaderPingExecutors,
+                Duration.ofMillis(leaderPingResponseWaitMs),
+                state -> state.responses().values().stream().map(PaxosString::get).anyMatch(uuid::equals));
+
+        for (Map.Entry<PingableLeader, PaxosString> cacheEntry : responses.responses().entrySet()) {
+            String uuidFromRequest = cacheEntry.getValue().get();
+            PingableLeader service = uuidToServiceCache.putIfAbsent(uuidFromRequest, cacheEntry.getKey());
+            throwIfInvalidSetup(service, cacheEntry.getKey(), uuidFromRequest);
```
jeremyk-91: Yeah, this is fine as a refactor. It seems like a strange place to do this in general, but this change is already large enough as is.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
felixdesouza: will do, added to #4112

leader-election-impl/src/main/java/com/palantir/paxos/PaxosResponsesWithRemote.java
felixdesouza: `S` initially stood for service, but checkstyle was complaining, not sure why I'm able to do it elsewhere.

Disagree about the second point, i.e. you get a response `T` by executing a command against `S(ERVICE)` similar to `method(self, args)` in python, `method.invoke(object, args...)` in java reflection etc.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
felixdesouza: what feels like a strange place? is it the `throwIfInvalidSetup`?

if so, the old code is also weird, re what happens in the presence of failures etc.

PR #4134, opened 2019-07-17
[Timelock Partitioning] Part 17: Remove `getSuspectedLeaderInMemory`

felixdesouza:
**Goals (and why)**:
Simplify `LeaderElectionService` and make it easier to abstract out a `LeaderPinger` interface.

The value (`HostAndPort`) isn't used anywhere, not even in big internal product. This made extracting out an interface a lot more difficult.

This however is a break, but I'm confident that it doesn't actually matter (plus it means we get a higher version number 💃 

Anywhere

ASAP unblocks further work re extracting a `Ping` interface that we can autobatch!


leader-election-impl/src/test/java/com/palantir/leader/proxy/AwaitingLeadershipProxyTest.java
```diff
@@ -103,42 +96,6 @@ public void shouldAllowObjectMethodsWhenNotLeading() {
         assertThat(proxy.toString()).startsWith("com.palantir.leader.proxy.AwaitingLeadershipProxy@");
     }
 
-    @Test
-    public void shouldNotRedirectToItself() {
```
felixdesouza: these tests make no sense, if anything, this should be tested at the `PaxosLeaderElectionService` level. 

`LeaderElectionService` provides the `currentToken`, and you should be able to tell whether or not you're the current leader or not. Either way, I'm struggling to find the value in these tests or the `getSuspectedLeaderInMemory` method whatsoever. So they're gone. 🤷‍♂

timelock-server/src/test/java/com/palantir/atlasdb/timelock/paxos/PaxosLeaderElectionServiceSerializationTest.java
```diff
@@ -1,63 +0,0 @@
...
-
-    static {
-        MAPPER.registerModule(new GuavaModule());
-        MAPPER.registerModule(new Jdk8Module());
-    }
-
-    @Test
-    public void canSerializeNoSuspectedLeader() throws JsonProcessingException {
-        LeaderElectionService leaderElectionService = mock(LeaderElectionService.class);
-        when(leaderElectionService.getSuspectedLeaderInMemory()).thenReturn(Optional.empty());
```
felixdesouza: these tests also make no sense, it's just testing the serialisation of `HostAndPort`, and if we've bumped guava or in the case of leader block atlas, a service has bumped guava which has changed the serde, then we're screwed anyway.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -160,7 +154,7 @@ public LeadershipToken blockOnBecomingLeader() throws InterruptedException {
 
     private void proposeLeadershipOrWaitForBackoff(LeadershipState currentState)
             throws InterruptedException {
-        if (pingLeader()) {
+        if (pingLeader(currentState.greatestLearnedValue())) {
```
jeremyk-91: Do we need this piece? I think it's safe, but at least right now `knowledge.getGreatestLearnedValue()` is cheap, and I'd prefer not to keep Paxos constructs in scope where not strictly needed

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -280,7 +260,7 @@ public String get() {
 
     private Optional<PingableLeader> getSuspectedLeaderOverNetwork(String uuid) {
         PaxosResponsesWithRemote<PingableLeader, PaxosString> responses = PaxosQuorumChecker.collectUntil(
-                ImmutableList.copyOf(otherPotentialLeadersToHosts.keySet()),
+                ImmutableList.copyOf(leaderPingExecutors.keySet()),
```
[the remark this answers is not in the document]
felixdesouza: they're not equal! to my surprise and dismay. `otherPingables` contains only remotes. Which I suppose makes sense, if you're the leader, you don't need to ping yourself etc. This is only ever reached when you've determined you're not leading etc.

after a second refactor, `otherPingables` *is* equal to `otherPotentialLeadersToHosts.keySet`. It's pretty messy, fortunately there was a test that was failing saying that a remote was claiming to be a local, which was local was inside `leaderPingExecutors`.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
felixdesouza: snippet of next refactor came through, basically, I want the construct `PingableLeader.pingLeaderWithUuid`. If I have an external pinger, it just needs to know about all uuids, and can coalesce the uuid resolution calls. Otherwise there'd need to be an intermediate layer that can retrieve knowledge (client-leader specific) and then pass that onto the autobatching pinger. I want most of the logic to live inside `PaxosLeaderElectionService` so I don't have to duplicate in both autobatch and non-autobatch case non trivial elements.

Plus it's weird that after we get the leadership state, `pingLeader` is the only place that rerequests latest knowledge, we should use the state we've just learnt.

timelock-server/src/test/java/com/palantir/atlasdb/timelock/paxos/PaxosLeaderElectionServiceSerializationTest.java
felixdesouza: Potentially, but if we do bump guava, what then? and if we don't bump guava but a consumer does, also what then?

felixdesouza: In addition I suppose it's just a roundabout way of testing it, it could have just been `HostAndPortSerDeTests` 🤷‍♂

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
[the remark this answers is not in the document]
jeremyk-91: Ah! Okay, my bad. Yeah, I didn't know we stuffed the local into `leaderPingExecutors`. Good that we did catch this in any case

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
jeremyk-91: Agree it's inconsistent with the others (but I did entertain the hypothesis that it was inconsistent for a specific reason - even though I could find no such reason).

Yep, this seems reasonable if we need it.

timelock-server/src/test/java/com/palantir/atlasdb/timelock/paxos/PaxosLeaderElectionServiceSerializationTest.java
jeremyk-91: For 1, the tests fail the version bump, we see it, and then decide either to take it and eat the break, or not to take it (more realistically not to take it immediately, for whatever reasons). This doesn't catch 2, of course - but I'd argue that 1 would be enough.

Yeah, it's indirect.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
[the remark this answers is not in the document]
jeremyk-91: current version uses `otherPingables` which would be what I would expect here

PR #4136, opened 2019-07-18
Do not log byte arrays

gmaretic:
**Goals (and why)**:
Noticed we were logging byte arrays while testing something else


leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderLearnerNetworkClient.java
```diff
@@ -64,7 +65,9 @@ public void learn(long seq, PaxosValue value) {
                     learner.learn(seq, value);
                 } catch (Throwable e) {
                     log.warn("Failed to teach learner the value {} at sequence {}",
-                            UnsafeArg.of("value", value.data),
+                            UnsafeArg.of("value", Optional.ofNullable(value.data)
+                                    .map(bytes -> BaseEncoding.base16().encode(bytes))
+                                    .orElse(null)),
```
jeremyk-91: I'm guessing the old behaviour is that it spits out a ref based on where in the memory you are

PR #4138, opened 2019-07-18
Fix changelog.

felixdesouza:
**Goals (and why)**:
Changelog doesn't compile, so can't release. Fixing it up, can't start with backticks, needs to be inside a string or a verbatim block.

ASAP, blocks release.


felixdesouza:
It was *not* auto-generated.

PR #4140, opened 2019-07-19
[Timelock Partitioning] Part 19: `AutobatchingLeaderPinger`

felixdesouza:
**Goals (and why)**:
We want to be able to ping nodes about leadership and also discover their uuids in a manner than doesn't mean we spin up so many more threads. In #4139 we introduced a `LeaderPinger` interface, where you attempt to locate the node which has a given `leaderUuid`. 

We now introduce the autobatching variant. 

This should also be the final main piece when it comes to being able to wire things up. Effectively we'll have:

```java
PaxosAcceptorNetworkClient acceptor = acceptorFactory.paxosAcceptorForClient(client);
PaxosLearnerNetworkClient learner = learnerFactory.paxosLearnerForClient(client);
LeaderPinger leaderPinger = leaderPingerFactory.leaderPingerFor(client);

PaxosProposer proposer = PaxosProposerImpl.newProposer(acceptor, learner, quorumSize, leadershipUuid);

return new PaxosLeaderElectionService(
        proposer,
        components.learner(client),
        leaderPinger,
        null, // only used for health check, somehow throw here
        ImmutableList.of(), // health check doesn't make a whole lot of sense in multi leader world, throw here
        ImmutableList.of(), // we won't run with this cli in timelock, also throw here
        new PaxosLatestRoundVerifierImpl(acceptor),
        learner,
        updatePollingWait.toMillis(),
        randomWaitBeforeProposingLeadership.toMillis(),
        PaxosLeaderElectionEventRecorder.NO_OP); // replace with client scoped event recorder
```

After this, 

A bit more convoluted than the autobatching variants of `PaxosAcceptor` and `PaxosLearner`. We don't want to introduce head of line blocking since all nodes will be hit (assuming uncached) to discover the uuid. So instead of using the coalescing autobatcher, we use the independent autobatcher where we can complete individual requests.

We want to batch:
* calls to `ping` or "are you the leader for this client". 
* what is your leader uuid?

Each follower will have a suspected leader, which it can garner from its knowledge. It then needs to ping that leader to see whether it is still leading.

Under the hood, each `PingableLeader` is client scoped. So when `ping` is called, and *after* we've determined which remote is the suspected leader, it will actually autobatch passing client as a parameter.

* First smoke test for the autobatcher code since it's somewhat non-trivial.

* Likely should add a `LeaderUuidCache` that can be shared across both `SingleLeaderPinger` and `AutobatchingLeaderPinger`
* We could probably get rid of the leaderUuid api endpoint, since the node is client aware.
  So instead of asking are you the leader for this uuid, we can just ask are you the leader for this client. Locally the node can then look at its knowledge for each of its clients and then decide then and there. Although I'm not sure whether that breaks the protocol/algorithm by any change. Doesn't block for now, but would be interesting to get rid of if plausible.
* Event recording, unimplemented for now since it's just for debuggability, and isn't super blocking, but would also need to be client scoped to make things clearer.

* `AutobatchingPingableLeaderFactory`
* `GetSuspectedLeaderWithUuid`

ASAP 💃

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPingableLeaderFactory.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPingableLeaderFactory.java
@@ -17,75 +17,168 @@
 package com.palantir.atlasdb.timelock.paxos;
 
 import java.io.Closeable;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
 import com.palantir.atlasdb.autobatch.Autobatchers;
-import com.palantir.atlasdb.autobatch.Autobatchers.SupplierKey;
 import com.palantir.atlasdb.autobatch.DisruptorAutobatcher;
-import com.palantir.leader.PingableLeader;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.paxos.LeaderPinger;
 
 public class AutobatchingPingableLeaderFactory implements Closeable {
 
-    private final DisruptorAutobatcher<Client, Boolean> pingAutobatcher;
-    private final DisruptorAutobatcher<SupplierKey, UUID> uuidAutobatcher;
+    private static final Logger log = LoggerFactory.getLogger(AutobatchingPingableLeaderFactory.class);
 
-    public AutobatchingPingableLeaderFactory(
-            DisruptorAutobatcher<Client, Boolean> pingAutobatcher,
-            DisruptorAutobatcher<SupplierKey, UUID> uuidAutobatcher) {
-        this.pingAutobatcher = pingAutobatcher;
-        this.uuidAutobatcher = uuidAutobatcher;
+    private final Collection<? extends Closeable> closeables;
+    private final DisruptorAutobatcher<UUID, Optional<ClientAwarePingableLeader>> uuidToRemoteAutobatcher;
+    private final Map<ClientAwarePingableLeader, ExecutorService> executors;
+    private final Duration leaderPingResponseWait;
+
+    private AutobatchingPingableLeaderFactory(
+            Collection<? extends Closeable> closeables,
+            DisruptorAutobatcher<UUID, Optional<ClientAwarePingableLeader>> uuidAutobatcher,
+            Map<ClientAwarePingableLeader, ExecutorService> executors,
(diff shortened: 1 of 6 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/GetSuspectedLeaderWithUuid.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        Set<UUID> missingUuids = Sets.difference(
+                uncachedUuids,
+                results.withoutRemotes().stream().map(PaxosContainer::get).collect(toSet()));
+
+        missingUuids.forEach(missingUuid -> completeRequest(uuidsToRequests, missingUuid, Optional.empty()));
+    }
+
+    private static void completeRequest(
```
gmaretic: This should just take a `Collection<SettableFuture<Optional<ClientAwarePingableLeader>>>`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPingableLeaderFactoryTests.java
```diff
@@ -0,0 +1,142 @@
...
+
+import org.junit.After;
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.palantir.paxos.LeaderPinger;
+
+public class AutobatchingPingableLeaderFactoryTests {
```
gmaretic: Nice tests!

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/GetSuspectedLeaderWithUuidTests.java
```diff
@@ -0,0 +1,165 @@
...
+import org.immutables.value.Value;
+import org.junit.After;
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.SettableFuture;
+import com.palantir.atlasdb.autobatch.BatchElement;
+
+public class GetSuspectedLeaderWithUuidTests {
```
gmaretic: 🎖

felixdesouza:
Completely forgot to merge this, will address comments in a separate PR because this will be super out of date otherwise.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPingableLeaderFactory.java
```diff
@@ -17,75 +17,168 @@
...
+            Map<BatchPingableLeader, ExecutorService> executors,
+            Duration leaderPingResponseWait,
+            UUID localUuid) {
+        BiMap<BatchPingableLeader, ClientAwarePingableLeaderImpl> correspondence = KeyedStream.of(
+                executors.keySet())
+                .map(ClientAwarePingableLeaderImpl::create)
+                .collectTo(HashBiMap::create);
+
+        Map<ClientAwarePingableLeader, ExecutorService> clientAwarePingableExecutors =
+                KeyedStream.stream(correspondence.inverse())
```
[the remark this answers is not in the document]
felixdesouza: looking back, I had no idea what I wrote haha, this cleared this up a treat

PR #4145, opened 2019-07-23
CloseableTimelockService interface (tiny refactor)

leonz:
**Goals (and why)**:

Because `RemoteTimelockServiceAdapter` is a final class that inherits from two different interfaces, it's hard to interact with in our unit tests.  By adding an intermediary interface we can more easily mock the class.

Adds an intermediary `CloseableTimelockService` interface which unifies the `TimelockService` and `AutoCloseable` interfaces.

The test failures on this branch don't seem related to this PR.

Whenever.


felixdesouza:
final classes *are* __finally__ (sorry) mockable in [mockito 2](https://www.baeldung.com/mockito-final) if this is the only reason you want this change I'm gonna close it out in recommendation of the above link. If something else, lmk!

PR #4154, opened 2019-07-29
[Sunsetting Background Sweep] Don't run BS if TS is fully enabled, unless the user says so explicitly

jeremyk-91:
**Goals (and why)**:
- Reduce the use of background sweep in places where targeted sweep is up and running.

- Make `SweepConfig#enabled()` optional.
- If it's not explicitly specified AND targeted sweep is fully enabled, treat it as disabled.

- Added tests for the new `isSweepEnabled` supplier to the background sweeper.
- Deserialization test is still relevant, and verifies that you can specify nothing.

- Double check that wire format of configs is preserved - I think so, and the `AtlasDbRuntimeConfigDeserializationTest` passes, but this is the main concern.
- This does mean that in some cases, we won't get round to sweeping cells that are never written to again. However, given how long Targeted Sweep has been in production, my guess would be that these are unlikely to be actively causing problems; also, if we want to sweep something explicitly, we can. More generally, @gmaretic and @jcvitanovic's work on sweep by migration should take care of these stragglers.
- Does the little decision tree at the top of `ShouldRunBackgroundSweepSupplier` make sense?

**Where should we start reviewing?**: ShouldRunBackgroundSweepSupplier


atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
```diff
@@ -178,12 +179,15 @@ public TargetedSweepInstallConfig targetedSweep() {
...
-     * live-reloadable.
+     * If true, a background thread will periodically delete cells that have been overwritten or deleted. This differs
+     * from scrubbing because it is an untargeted cleaning process that scans all data looking for cells to delete.
+     *
+     * Note that this relates to Background or Legacy Sweep, and is separate from Targeted Sweep (which is enabled or
+     * disabled depending on {@link TargetedSweepInstallConfig#enableSweepQueueWrites()} and
+     * {@link TargetedSweepRuntimeConfig#enabled()}).
+     *
+     * @deprecated Use {@link AtlasDbRuntimeConfig#sweep#enableSweep} to make this value live-reloadable. This value
+     * may be disregarded if {@link AtlasDbRuntimeConfig#sweep#enableSweep} is specified.
```
felixdesouza: This value is ignored entirely I believe. I wonder whether there's a way to say "recognise this value, but don't error on it" in jackson

atlasdb-config/src/test/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplierTest.java
```diff
@@ -0,0 +1,119 @@
...
+        Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier = mock(Supplier.class);
+        when(runtimeConfigSupplier.get())
+                .thenReturn(createRuntimeConfig(TARGETED_SWEEP_ENABLED, BACKGROUND_SWEEP_UNSET))
+                .thenReturn(createRuntimeConfig(TARGETED_SWEEP_DISABLED, BACKGROUND_SWEEP_UNSET))
+                .thenReturn(createRuntimeConfig(TARGETED_SWEEP_ENABLED, BACKGROUND_SWEEP_ENABLED));
+
+        ShouldRunBackgroundSweepSupplier supplier = new ShouldRunBackgroundSweepSupplier(
+                SWEEP_QUEUE_WRITES_ENABLED, runtimeConfigSupplier);
+
+        assertThat(supplier.get()).isFalse();
```
felixdesouza: nit, add `as` for each assert please, can be as simple as 👍 

```java
assertThat(supplier.get())
  .as("TARGETED_SWEEP_ENABLED, BACKGROUND_SWEEP_UNSET")
  .isFalse();
```

etc.

changelog/@unreleased/pr-4154.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: We no longer run Background Sweep by default if Targeted Sweep is fully
```
felixdesouza: try opening autorelease but on your branch to see whether or not it renders correctly

atlasdb-config/src/main/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplier.java
```diff
@@ -0,0 +1,65 @@
...
+import com.palantir.atlasdb.sweep.queue.config.TargetedSweepRuntimeConfig;
+
+/**
+ * Synthesises provided install and runtime configurations, and uses this to decide whether the background sweeper
+ * should run. Specifically:
+ *
+ * - if the background sweeper has been explicitly enabled or disabled, use that setting;
+ * - otherwise if Targeted Sweep is fully enabled (both writing to the sweep queue and actually sweeping), disable
+ *   background sweep
+ * - otherwise, follow {@link AtlasDbConstants#DEFAULT_ENABLE_SWEEP}.
```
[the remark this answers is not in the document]
jeremyk-91: I think we want to be defensive here, mainly because for products in the ecosystem of large internal product they're likely to not use TS and I can't be certain that all of them explicitly set BS to enabled as opposed to relying on it being the default.

This can be changed later on once TS has broader adoption internally, but I think is a separate change that I wouldn't want to make yet

atlasdb-config/src/test/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplierTest.java
jeremyk-91: Ah, okay. Sure

changelog/@unreleased/pr-4154.v2.yml
jeremyk-91: Yep, confirmed it works on branch `0000000aaaaaaa`

atlasdb-config/src/test/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplierTest.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.function.Supplier;
+
+import org.junit.Test;
+
+import com.palantir.atlasdb.sweep.queue.config.ImmutableTargetedSweepInstallConfig;
+import com.palantir.atlasdb.sweep.queue.config.ImmutableTargetedSweepRuntimeConfig;
+import com.palantir.atlasdb.sweep.queue.config.TargetedSweepInstallConfig;
+import com.palantir.atlasdb.sweep.queue.config.TargetedSweepRuntimeConfig;
+
+public class ShouldRunBackgroundSweepSupplierTest {
```
j-baker: Think you get to use at most 60 lines of code for these tests!

atlasdb-config/src/main/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplier.java
```diff
@@ -0,0 +1,67 @@
...
+ * should run. Specifically:
+ *
+ * <ul>
+ *     <li>if the background sweeper has been explicitly enabled or disabled, use that setting;</li>
+ *     <li>otherwise if Targeted Sweep is fully enabled (both writing to the sweep queue and actually sweeping), disable
+ *     background sweep</li>
+ *     <li>otherwise, follow {@link AtlasDbConstants#DEFAULT_ENABLE_SWEEP}.</li>
+ * </ul>
+ */
+public class ShouldRunBackgroundSweepSupplier implements Supplier<Boolean> {
```
felixdesouza: nit: `BooleanSupplier`

jeremyk-91: Sure, will change this one. Note that pushing it down into `BackgroundSweeperImpl` onwards would be a break for large internal product, so I'd prefer not to do that now

atlasdb-config/src/test/java/com/palantir/atlasdb/config/ShouldRunBackgroundSweepSupplierTest.java
jeremyk-91: Yep, I wrote these for (developer) speed - will condense them

atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
felixdesouza: and here re `@JsonIgnoreProperties`?

jeremyk-91: Confirmed that's correct, and added

felixdesouza: have you pushed this?

PR #4155, opened 2019-07-30
Fix precommit condition check order

sandorw:
**Goals (and why)**:
Fix the ordering of the pre-commit condition check in `SnapshotTransaction`. No one is currently affected AFAIK, but it should generally run before we check locks for the same reason that we check for read-write conflicts before checking locks. Given in-progress work that uses a pre-commit condition to do work analogous to a read-write conflict check, this will be important soon.

### DIFF
--- a/docs/source/transactions/life_of_a_transaction.rst
+++ b/docs/source/transactions/life_of_a_transaction.rst
@@ -92,8 +92,8 @@ The ordering of these steps is important:
    could involve transactional reads to the database, they require the same protections as the read-write conflict
    checks in the prior step. An alternative implementation of those checks could be done as a pre-commit condition, for
    example, so it is important that we handle this before the final lock check.
-9. The lock check may be run in parallel with pre-commit condition checks, though it must strictly be run before writing
-   to the transactions table, as we cannot finish our commit if we can't be certain we still have locks.
+9. The lock check must be run strictly before writing to the transactions table, as we cannot finish our commit if
+   we can't be certain we still have locks.
 
 Read-Only Variant
 ~~~~~~~~~~~~~~~~~
--- a/docs/source/transactions/life_of_a_transaction.rst
+++ b/docs/source/transactions/life_of_a_transaction.rst
@@ -55,8 +55,8 @@ Committing a write transaction is a multi-stage process:
 6. Get the commit timestamp.
 7. For serializable transactions, check that the state of the world at commit time is same as that at start time.
    Please see :ref:`Isolation Levels<isolation-levels>` for more details.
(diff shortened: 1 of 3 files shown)


sandorw:
The only interesting edge case that I can think of here is that advisory locks are handled by the pre-commit condition rather than the lock checks, which handle the transaction locks that are part of the protocol.

Previous behaviour: check transaction locks, then advisory locks
New behaviour: check advisory locks, then transaction locks

jeremyk-91:
I think the interesting case that you mention is fine. We need both to pass before we commit, so the cases we need to be careful of are the ones where we lose the locks for the first check before executing the second. Since you could lose your locks anyway the moment the response packet leaves timelock, I'd expect this to be a problem only if the pre-commit condition and the transaction lock checks directly influence each other, which isn't the case for advisory and transaction locks as those should be distinct. 

For example, this is obviously contrived, but a pre-commit condition that this change _would_ 'break' would be one that attempts to test if a lock used by the transaction can be acquired. This could under lucky circumstances pass in the old impl, but will always fail in the new impl.

Other related thoughts (non-blocking, but probably should be tracked somewhere):

1. Might want to say that the behaviour is expected to not be dependent on transaction locks.
2. The Life of a Transaction docs probably need to be updated where we flip the ordering of these steps.

docs/source/transactions/life_of_a_transaction.rst
```diff
@@ -88,9 +88,12 @@ The ordering of these steps is important:
    in between our start and commit, and someone else deleted the value after our commit. If we pass our lock check, but
    then lose our locks and have a very long GC, Thorough Sweep might clear all evidence of the conflict, meaning that
    we miss a read-write conflict. This would be safe for conservative sweep because of the deletion sentinel.
-8. The lock check may be run in parallel with pre-commit condition checks, though it must strictly be run before writing
+8. We need to check that the pre-commit conditions, if any, still hold. Given that these checks are user defined and
+   could involve transactional reads to the database, they require the same protections as the read-write conflict
+   checks in the prior step. An alternative implementation of those checks could be done as a pre-commit condition, for
+   example, so it is important that we handle this before the final lock check.
+9. The lock check may be run in parallel with pre-commit condition checks, though it must strictly be run before writing
```
jeremyk-91: I guess this was moved down directly, but based on the logic in 8 I don't think we can say that it can be run in parallel any more. It's probably fine for this to just say "The lock check must strictly be run before writing ..."

sandorw: Yes, good catch, I just changed the order without reading :(

PR #4157, opened 2019-07-30
Use a simple Caffeine cache around MetricName construction.

carterkozak:
**Goals (and why)**:

We look up the same metrics a lot, and string concatenation is
relatively expensive. MetricName values are already stored in a
map in the TaggedMetricRegistry, so there's little risk of causing
excessive heap pressure due to this cache.

Uses an unbounded Caffeine cache which roughly mirrors the TaggedMetricRegistry.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
@@ -52,6 +52,7 @@ public class MetricsManager {
     private final Set<MetricName> registeredTaggedMetrics;
     private final Predicate<TableReference> isSafeToLog;
     private final ReadWriteLock lock;
+    private final MetricNameCache metricNameCache;
 
     public MetricsManager(MetricRegistry metricRegistry,
             TaggedMetricRegistry taggedMetricRegistry,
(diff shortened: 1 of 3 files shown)


carterkozak:
Do we have docs on locating circleci internal builds?

atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
```diff
@@ -230,6 +231,9 @@ public void deregisterTaggedMetrics(Predicate<MetricName> predicate) {
                     .collect(Collectors.toList());
 
             metricsToRemove.forEach(taggedMetricRegistry::remove);
+
+            // Don't be clever, invalidate everything. This should be relatively uncommon.
+            metricNameCache.invalidate();
```
felixdesouza: seems fine, but I am curious, do we really ever need to invalidate this cache?

felixdesouza:
There are other instances of `MetricRegistry#name` but they aren't in the hot path, so seems fine to leave.

atlasdb-api/src/main/java/com/palantir/atlasdb/util/MetricsManager.java
carterkozak: Need to? Probably not, but there's a risk of a memory leak if we remove data from the TaggedMetricRegistry without removing it from the cache.

Based on callers to this method, we could remove cache invalidation, but I don't think it buys us enough to be worth sacrificing safety.

felixdesouza: Yeah, that's fair.

PR #4158, opened 2019-07-31
AssertUtils.getDebuggingException returns a SafeLoggable Exception

carterkozak:


jeremyk-91:
There _may_ be consequences for people who rely on the class of the exception, but I think that's a bit too hacky so this should probably be okay.

PR #4162, opened 2019-08-05
[PDS-94229] TimeLock - Thread Leak in Async Lock Service

jeremyk-91:
**Goals (and why)**:
- See PDS-94229. #4121 was the first part which addressed a thread leak in the legacy synchronous lock service, but there is another leak, this time in the async lock service.

- An async lock service takes two executors, `reaperExecutor` (to reap lock grants when the client has not refreshed them for too long) and `timeoutExecutor` (to mark lock requests as failed when they block for too long). 
- The former was shut down when the `AsyncLockService` is closed, but the latter was not. This PR shuts down the timeout executor on close.

This isn't tested. I've added a test.

- I did not switch to `Ownable<T>` and the infra used to distinguish things in `LockServiceImpl`, as this currently as far as I know is only called in one place, where the executors are created and thus need to be shut down.

**Where should we start reviewing?**: `LockAcquirer`

**Priority (whenever / two weeks / yesterday)**: tomorrow? P1 is about right for this one

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
@@ -26,7 +26,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.base.Throwables;
 import com.palantir.logsafe.SafeArg;
 
-public class LockAcquirer {
+public class LockAcquirer implements AutoCloseable {
 
     private static final Logger log = LoggerFactory.getLogger(LockAcquirer.class);
 
@@ -50,6 +50,11 @@ public class LockAcquirer {
     public AsyncResult<Void> waitForLocks(UUID requestId, OrderedLocks locks, TimeLimit timeout) {
         return new Acquisition(requestId, locks, timeout, lock -> lock.waitUntilAvailable(requestId))
                 .execute();
+    }
+
+    @Override
+    public void close() {
+        timeoutExecutor.shutdown();
     }
 
     private class Acquisition {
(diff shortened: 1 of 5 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockAcquirer.java
```diff
@@ -26,7 +26,7 @@
 import com.google.common.base.Throwables;
 import com.palantir.logsafe.SafeArg;
 
-public class LockAcquirer {
+public class LockAcquirer implements AutoCloseable {
```
felixdesouza: kinda weird to make `LockAcquirer` closeable here. Is it not sufficient to just close the executor like `reaperExecutor` already is? Since `AsyncLockService` "owns" the executor so to speak.

jeremyk-91: No strong preference - reason I went with this is that `AsyncLockService` currently doesn't maintain any references to `timeoutExecutor`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/lock/AsyncLockServiceTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.function.Supplier;
+
+import org.junit.Test;
+
+import com.palantir.atlasdb.timelock.config.ImmutableRateLimitConfig;
+import com.palantir.atlasdb.timelock.config.TargetedSweepLockControlConfig;
+import com.palantir.atlasdb.util.MetricsManagers;
+import com.palantir.common.concurrent.PTExecutors;
+
+public class AsyncLockServiceTest {
```
felixdesouza: this seems quite redundant imo, not sure if it's worth all the awkward wiring etc.

jeremyk-91: I'm not aware of any other tests that cover this behaviour. Most of the wiring can be removed once we revert #4096

PR #4169, opened 2019-08-13
Update old ADR status

jeremyk-91:
**Goals (and why)**:
- Ensure ADR readers aren't confused by things that were true at the time but have since changed.

- Update the status of several older ADRs. Most of these pertain to the schema mutation lock or to DisruptorAutobatchers.

Docs change. Docs build passing should do

- Are there any other documents which should have such notes attached?

**Where should we start reviewing?**: no particular order

### DIFF
--- a/docs/source/miscellaneous/doc/adr/0015-batch-asynchronous-post-transaction-unlock-calls.md
+++ b/docs/source/miscellaneous/doc/adr/0015-batch-asynchronous-post-transaction-unlock-calls.md
@@ -4,7 +4,9 @@ Date: 28/06/2018
 
 ## Status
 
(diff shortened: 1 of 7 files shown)


docs/source/miscellaneous/doc/adr/0015-batch-asynchronous-post-transaction-unlock-calls.md
```diff
@@ -4,7 +4,8 @@ Date: 28/06/2018
 
 ## Status
 
-Accepted
```
gmaretic: For consistency with the rest, leave the `Accepted`?

PR #4170, opened 2019-08-13
[Timelock Partitioning] Part 20: Proper usage of get learned values

felixdesouza:
**Goals (and why)**:
Whilst wiring up the impl for trying to batch timestamp paxos, I ended up having passing tests for `PaxosTimestampBoundStoreTests` for both the new and old clients. However I was receiving `NullPointerException`.

Some of the tests *expected* an `NPE`, which somewhat masked the issue. This should fix it.

The problem lies in that when we haven't learned any values instead of `PaxosLearner` returning `Optional<PaxosValue>` it returns null. This also means we have no idea what sequence number the "no learned values" corresponds to.

Change the type (this is fine, none of this is released) to have a `WithSeq<Optional<PaxosValue>>` which correctly encodes the type that we want to use.

Added extra test cases.

ASAP 💃

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
@@ -50,10 +50,12 @@ public interface BatchPaxosLearner {
      * ({@link WithSeq}), it returns the learnt value. Values where nothing has been learnt are excluded. If for a given
      * {@link Client} nothing has been learnt, the {@link Client} is also excluded.
      * <p>
+     * In addition to the above, if a {@link PaxosValue} <em>is</em> returned, it is guaranteed to be non-null.
+     * <p>
      * @param clientAndSeqs each {@link Client} with a given paxos instance ({@link WithSeq}) to retrieve the learnt
      * values for
      * @return for each {@link Client} the different {@link PaxosValue}s learnt for different
-     * {@link PaxosValue#getRound}s
+     * {@link PaxosValue#getRound}s. Any {@link PaxosValue}s are guaranteed to be non-null
      */
     @POST
     @Path("learned-values")
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearnerResource.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearnerResource.java
@@ -18,6 +18,7 @@ package com.palantir.atlasdb.timelock.paxos;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import com.google.common.collect.SetMultimap;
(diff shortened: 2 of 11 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
```diff
@@ -59,7 +61,7 @@
     @Path("learned-values")
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)
-    SetMultimap<Client, PaxosValue> getLearnedValues(Set<WithSeq<Client>> clientAndSeqs);
+    SetMultimap<Client, WithSeq<Optional<PaxosValue>>> getLearnedValues(Set<WithSeq<Client>> clientAndSeqs);
```
jeremyk-91: I don't think this is consistent with the javadoc we wrote though:

> Values where nothing has been learnt are excluded. If for a given Client nothing has been learned, the Client is also excluded.

Probably need to either change that to reflect this (in which case I assume that the client/WithSeq values returned form `clientAndSeqs`), or change the implementation so it filters correctly

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearnerResource.java
```diff
@@ -39,14 +40,17 @@ public void learn(SetMultimap<Client, PaxosValue> paxosValuesByClient) {
...
 
     @Override
-    public SetMultimap<Client, PaxosValue> getLearnedValues(Set<WithSeq<Client>> clientAndSeqs) {
+    public SetMultimap<Client, WithSeq<Optional<PaxosValue>>> getLearnedValues(Set<WithSeq<Client>> clientAndSeqs) {
         return KeyedStream.of(clientAndSeqs)
                 .mapKeys(WithSeq::value)
-                .map(WithSeq::seq)
-                .map((client, seq) -> paxosComponents.learner(client).getLearnedValue(seq))
+                .map(clientAndSeq -> clientAndSeq.map(this::getLearnedValue))
                 .collectToSetMultimap();
```
jeremyk-91: I think if you want the old behaviour at this point you can filter out things which are paired with an empty

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
felixdesouza: that's fair, I think I'm in favour of filtering it out, it's just cruft otherwise

felixdesouza: good shout

felixdesouza:
oop, a bit too much went in to this PR, fixing...

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosLearner.java
jeremyk-91: Yeah, also in favour of filtering

PR #4173, opened 2019-08-15
File comparison test does not check __CLASS_HASH

### LINKED ISSUE #2548: We should ignore `__CLASS_HASH` changes on Schema tests
On `SchemaTest`, we check if the generated classes and the old classes are the same, to detect api breaks. We compare the files and require all the lines to be the same. As the `__CLASS_HASH = <hash>` line is not part of the api, it should be ignored when comparing the two files.

OStevan:
**Goals (and why)**:
Fixes #2548, when comparing two files in SchemaTest ignore __CLASS_HASH. Comparing two generated class source files to detect API breaks checks also the __CLASS_HASH which does not contribute to the API.

- check if the files have the same size
- stream the files contents per line to detect differences ignoring the lines with __CLASS_HASH

This PR improves tests.

Source code readability

n/a(single file changed)

whenever


atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -209,10 +214,31 @@ private void checkIfFilesAreTheSame(List<String> generatedTestTables) {
...
             File expectedFile = new File(EXPECTED_FILES_FOLDER_PATH, generatedFilePath);
             File actualFile = new File(testFolder.getRoot(), generatedFilePath);
-            assertThat(actualFile).hasSameContentAs(expectedFile);
+
+            assertThat(expectedFile.length()).isEqualTo(actualFile.length());
+
+            try (Stream<String> expectedFileStream = java.nio.file.Files
+                        .lines(Paths.get(EXPECTED_FILES_FOLDER_PATH, generatedFilePath));
+                Stream<String> actualFileStream = java.nio.file.Files
+                        .lines(Paths.get(testFolder.getRoot().getPath(), generatedFilePath));
```
jeremyk-91: There's a bit of duplication here with getting the paths here and in getting the Files above. I think you can just call `File.toPath()` on the files created from lines 217/218

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -209,10 +214,31 @@ private void checkIfFilesAreTheSame(List<String> generatedTestTables) {
...
+                Stream<Boolean> zipped = Streams.zip(
+                        expectedFileStream,
+                        actualFileStream,
+                        (first, second) -> {
+                            if (first.equals(second))
+                                return true;
+                            return first.contains("__CLASS_HASH") && second.contains("__CLASS_HASH");
```
jeremyk-91: nit: The `__CLASS_HASH` string should probably be a constant. Also I normally would just write this check one line, though that's more of a matter of style preference

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -209,10 +214,31 @@ private void checkIfFilesAreTheSame(List<String> generatedTestTables) {
...
             File actualFile = new File(testFolder.getRoot(), generatedFilePath);
-            assertThat(actualFile).hasSameContentAs(expectedFile);
+
+            assertThat(expectedFile.length()).isEqualTo(actualFile.length());
+
+            try (Stream<String> expectedFileStream = java.nio.file.Files
+                        .lines(Paths.get(EXPECTED_FILES_FOLDER_PATH, generatedFilePath));
+                Stream<String> actualFileStream = java.nio.file.Files
+                        .lines(Paths.get(testFolder.getRoot().getPath(), generatedFilePath));
+                Stream<Boolean> zipped = Streams.zip(
```
jeremyk-91: nit: probably best to say what the booleans mean (maybe something like `correspondingLinesMatchOrAreHashes`)?

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -209,10 +214,31 @@ private void checkIfFilesAreTheSame(List<String> generatedTestTables) {
...
+                                return true;
+                            return first.contains("__CLASS_HASH") && second.contains("__CLASS_HASH");
+                        })) {
+
+                assertThat(zipped
+                        .anyMatch(elem -> !elem))
+                        .isFalse();
+            } catch (IOException e) {
+                Assertions.fail("Exception on stream creation", e);
+            }
```
jeremyk-91: Usually we just add the exception to the method signature (junit will treat a test as a failure if an exception gets thrown out of the test method)

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -209,10 +214,31 @@ private void checkIfFilesAreTheSame(List<String> generatedTestTables) {
...
+                        actualFileStream,
+                        (first, second) -> {
+                            if (first.equals(second))
+                                return true;
+                            return first.contains("__CLASS_HASH") && second.contains("__CLASS_HASH");
+                        })) {
+
+                assertThat(zipped
+                        .anyMatch(elem -> !elem))
+                        .isFalse();
```
jeremyk-91: AssertJ does have pretty nice support for streams, so you can do something like `assertThat(zipped).allMatch(elem -> elem);` directly

Also note that we're asserting `not exists not elem`, which is equivalent to `for-all elem` which I think is clearer

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
OStevan: Exception is raised inside a lambda so removing a try is not an option.

PR #4175, opened 2019-08-15
[Timelock Partitioning] Part 21: Integration pain points

felixdesouza:

**Goals (and why)**:

A few pain points that can only really be detected when running integration tests:
  * (de)serialisation for path/query/header params
  * empty optional handling for dropwizard (unaffected by internal server)
  * (de)serialisation for string aliases in maps

  * individual tests are not worth it, integration tests that are coming later are sufficient.

ASAP 💃 🔥

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AcceptorCacheKey.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AcceptorCacheKey.java
@@ -27,14 +27,22 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 @Value.Immutable
 @JsonDeserialize(as = ImmutableAcceptorCacheKey.class)
 @JsonSerialize(as = ImmutableAcceptorCacheKey.class)
-public interface AcceptorCacheKey {
+public abstract class AcceptorCacheKey {
 
     @JsonValue
     @Value.Parameter
-    UUID value();
+    public abstract UUID value();
 
-    static AcceptorCacheKey newCacheKey() {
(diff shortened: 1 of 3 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AcceptorCacheKey.java
```diff
@@ -27,14 +27,22 @@
...
+
+    public static AcceptorCacheKey newCacheKey() {
         return ImmutableAcceptorCacheKey.of(UUID.randomUUID());
     }
 
+    @Override
+    public String toString() {
+        return value().toString();
+    }
 }
```
jeremyk-91: Should be fine. This looks like it breaks wire format, but is unavoidable if we want to use this as a map key, and shouldn't be used anywhere yet

timelock-server/src/main/java/com/palantir/atlasdb/timelock/TimeLockServerLauncher.java
```diff
@@ -70,4 +76,12 @@ public void run(CombinedTimeLockServerConfiguration configuration, Environment e
                 CombinedTimeLockServerConfiguration.blockingTimeoutMs(),
                 registrar);
     }
+
+    @Provider
+    private static final class EmptyOptionalTo204ExceptionMapper implements ExceptionMapper<EmptyOptionalException> {
+        @Override
+        public Response toResponse(EmptyOptionalException exception) {
+            return Response.noContent().build();
```
jeremyk-91: seems reasonable, I guess we never had issues here because optionals were not part of the API till now

PR #4176, opened 2019-08-15
[Timelock Partitioning] Part 22: Fix up `BatcingPaxosLatestSequenceCache`.

felixdesouza:

**Goals (and why)**:
During integration, the error for cache miss was not the same between the server and client meaning we were never handling it.

In addition to that, our custom feign has its own custom error decoding which is incompatible with conjure-java-runtime errors. This means our error handling for cache miss never gets invoked. Obviously we don't want to block on conjure-java-runtime. 

I evaluated the following options:
  * (picked) search for the error name in the `RuntimeException` that gets thrown 
    from being unable to deserialise the conjure-java-runtime error.
  * go to cjr (infeasible)
  * use retrofit instead
    * very self contained and doable, but we are subject to limits which the 
      conjure-java-runtime workstream will eventually resolve, and doesn't seem 
      useful to duplicate this work

I went with the ugly hack which is unfortunate but works in testing. I've also added it to #4112 of things to do at a later date, perhaps a separate issue?

I also pulled out the error into `BatchPaxosAcceptor` which is where it should live really.

At a later date when the RPC layer changes have merged, it might make more sense to have the rpc layer do this, and then throw an application exception such as `AcceptorCacheKeyMiss`, which means we don't need to worry about the remoting layer in a mainly non-remoting aware class.

Covered by integration tests. Also seems pointless to test the error is same unless we're not sharing the same constant.

ASAP 🔥 💃 🚒

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
@@ -63,7 +63,7 @@ final class BatchingPaxosLatestSequenceCache implements CoalescingRequestFunctio
                 return handleCacheMiss(clients);
             }
 
-            log.error("received remote exception that is not a cache key miss", e);
+            log.warn("received remote exception that is not a cache key miss", e);
             throw e;
         } catch (RuntimeException e) {
             // TODO(fdesouza): Remove this once we've moved to CJR properly and the above works.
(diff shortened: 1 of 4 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
```diff
@@ -62,21 +58,35 @@
...
+            log.error("received remote exception that is not a cache key miss", e);
+            throw e;
+        } catch (RuntimeException e) {
+            // TODO(fdesouza): Remove this once we've moved to CJR properly and the above works.
+            if (e.getMessage().contains(BatchPaxosAcceptor.CACHE_KEY_NOT_FOUND.name())) {
+                log.info("Cache key is invalid, invalidating cache - using deprecated detection method");
+                return handleCacheMiss(clients);
             }
 
+            log.error("received unexpected runtime exception", e);
```
jeremyk-91: should these be logged at `error`? Specifically I'm thinking about exceptions where one node is unreachable so the delegate could throw

PR #4178, opened 2019-08-16
Stop publishing with Jdk7

j-baker:
It causes issues internally and everyone uses a new enough version of Jackson that they don't need it.


felixdesouza:
are there any blockers to actually just removing `Jdk7Module` from `AtlasDbConfigs`?

j-baker:
It's a dep of the dropwizard crap we deal with

felixdesouza:
Thanks!

PR #4180, opened 2019-08-16
Seed timestamp on node start

### LINKED ISSUE #4174: False Positives on Conservative TS Queue Age on Quiet Stacks
This actually happened on an internal stack (event reference 378322):

- Suppose a transaction is committed at timestamp T on one node of a service. This writes entries to the sweep queue.
- This node is bounced before the next time the async puncher punches to the store.
- The system then lies unused for a day. During this time, the puncher on the bounced node has in memory `-1`, so no actual punching takes place. Other nodes continue punching at the last transaction they did, but since the service is quiet, this is at a timestamp Ki less than T.
- In this way, the unreadable timestamp remains at the maximum of the Kis, but there _is_ a transaction committed after the maximum of the Kis.
- Thus the head of the sweep queue gets older and older, and eventually we get an alert.

Not sure if we should necessarily fix this, but if we get too many false alarms could be worth implementing.

OStevan:
**Goals (and why)**:
Fixes #4174, initialize AsyncPuncher with a seed timestamp if quorum is present. Front of the sweep queue can become stale if the service is silent, detailed situation described in the linked issue.
  
- on DefaultCleaner build try to get a seed timestamp
- initialize the AsyncPuncher with the optional seed

coding style

n/a
whenever

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
@@ -36,7 +36,7 @@ import com.palantir.logsafe.SafeArg;
  */
 public final class AsyncPuncher implements Puncher {
     private static final Logger log = LoggerFactory.getLogger(AsyncPuncher.class);
-    private static final long INVALID_TIMESTAMP = -1L;
+    public static final long INVALID_TIMESTAMP = -1L;
 
     public static AsyncPuncher create(Puncher delegate, long interval, long creationTimestamp) {
         AsyncPuncher asyncPuncher = new AsyncPuncher(delegate, interval, creationTimestamp);
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
@@ -38,8 +38,8 @@ public final class AsyncPuncher implements Puncher {
     private static final Logger log = LoggerFactory.getLogger(AsyncPuncher.class);
     private static final long INVALID_TIMESTAMP = -1L;
 
-    public static AsyncPuncher create(Puncher delegate, long interval) {
-        AsyncPuncher asyncPuncher = new AsyncPuncher(delegate, interval);
+    public static AsyncPuncher create(Puncher delegate, long interval, long creationTimestamp) {
+        AsyncPuncher asyncPuncher = new AsyncPuncher(delegate, interval, creationTimestamp);
         asyncPuncher.start();
         return asyncPuncher;
     }
@@ -49,11 +49,12 @@ public final class AsyncPuncher implements Puncher {
 
     private final Puncher delegate;
     private final long interval;
-    private final AtomicLong lastTimestamp = new AtomicLong(INVALID_TIMESTAMP);
+    private final AtomicLong lastTimestamp;
 
-    private AsyncPuncher(Puncher delegate, long interval) {
+    private AsyncPuncher(Puncher delegate, long interval, long creationTimestamp) {
         this.delegate = delegate;
         this.interval = interval;
+        this.lastTimestamp = new AtomicLong(creationTimestamp);
     }
 
     private void start() {
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
@@ -37,7 +37,7 @@ import com.palantir.logsafe.SafeArg;
  */
 public final class AsyncPuncher implements Puncher {
     private static final Logger log = LoggerFactory.getLogger(AsyncPuncher.class);
-    public static final long INVALID_TIMESTAMP = -1L;
+    private static final long INVALID_TIMESTAMP = -1L;
 
     public static AsyncPuncher create(Puncher delegate, long interval, Optional<Long> creationTimestamp) {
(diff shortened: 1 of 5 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/cleaner/AsyncPuncherTest.java
```diff
@@ -74,27 +93,43 @@ public void delegatesInitializationCheck() {
...
+    }
+
+    @Test
+    public void testSeededPuncherUpdate() throws Exception {
+        asyncPuncherSeeded.punch(PUNCHER_UPDATE_TIMESTAMP);
+        checkExpectedValue(asyncPuncherSeeded, PUNCHER_UPDATE_TIMESTAMP);
+    }
+
+    private void checkExpectedValue(Puncher puncher, Long expected) throws InterruptedException {
+        Long retrieved = Long.MIN_VALUE;
```
jeremyk-91: Yeah, I guess this is consistent with the style of the other tests. I'd probably write it with the primitive (`long`) form, but this should be fine

changelog/@unreleased/pr-4180.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: AsnycPuncher is now intilized with a seed timestamp if a quorum is present. Public factory method for
```
jeremyk-91: nit: `initialized`

Two points here:
- I think this needs to be a `break` as well, because of the second part you mention.
- It's worth mentioning why the seed timestamp is useful (here, it provides more accurate metrics for the unreadable timestamp on stacks that have a low volume of transactions)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/DefaultCleanerBuilder.java
```diff
@@ -147,4 +148,12 @@ public Cleaner buildCleaner() {
                 puncher,
                 Suppliers.ofInstance(transactionReadTimeout));
     }
+
+    private static Optional<Long> timestampSeed(TimelockService timelockService) {
+        try {
+            return Optional.of(timelockService.getFreshTimestamp());
+        } catch (Exception e) {
+            return Optional.empty();
```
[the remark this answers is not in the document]
OStevan: Logging added.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
```diff
@@ -38,8 +39,8 @@
     private static final Logger log = LoggerFactory.getLogger(AsyncPuncher.class);
     private static final long INVALID_TIMESTAMP = -1L;
 
-    public static AsyncPuncher create(Puncher delegate, long interval) {
-        AsyncPuncher asyncPuncher = new AsyncPuncher(delegate, interval);
+    public static AsyncPuncher create(Puncher delegate, long interval, Optional<Long> creationTimestamp) {
```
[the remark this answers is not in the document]
OStevan: Changelog added.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/cleaner/AsyncPuncherTest.java
OStevan: Updated to long.

changelog/@unreleased/pr-4180.v2.yml
OStevan: Updated changelog.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/cleaner/AsyncPuncherTest.java
```diff
@@ -120,8 +120,8 @@ public void testSeededPuncherUpdate() throws Exception {
         checkExpectedValue(asyncPuncherSeeded, PUNCHER_UPDATE_TIMESTAMP);
     }
 
-    private void checkExpectedValue(Puncher puncher, Long expected) throws InterruptedException {
-        Long retrieved = Long.MIN_VALUE;
+    private void checkExpectedValue(Puncher puncher, long expected) throws InterruptedException {
+        long retrieved = Long.MIN_VALUE;
```
jeremyk-91: Ah, didn't really intend this to be actionable! But thanks for improving it :)

changelog/@unreleased/pr-4180.v2.yml
```diff
@@ -1,6 +1,7 @@
-type: improvement
-improvement:
-  description: AsnycPuncher is now intilized with a seed timestamp if a quorum is present. Public factory method for
-  AsyncPuncher is changed, projects depending on it have to be updated.
+type: break
+break:
+  description: AsnycPuncher is now intilized with a seed timestamp if a quorum is present. As a result metrics related
```
jeremyk-91: nit: `AsyncPuncher`, `initialized`

changelog/@unreleased/pr-4180.v2.yml
```diff
@@ -1,6 +1,7 @@
...
+  to unreadable timestamps on individual with low volume become more accurate. Public factory method for AsyncPuncher
```
jeremyk-91: probably `on individual stacks with low volume`?

changelog/@unreleased/pr-4180.v2.yml
```diff
@@ -1,6 +1,7 @@
...
+  is changed, projects depending on it have to be updated.
   links:
   - https://github.com/palantir/atlasdb/pull/4180
```
jeremyk-91: I actually meant that we should have two changelog entries, one for `improved` (that has the accuracy part) and one for `break` (which has the factory change)

PR #4181, opened 2019-08-16
add logging and reaping of leaked atlas locks

clockfort:
**Goals (and why)**:
There's at least three different cases I've found in prod where locks are leaked and constantly refreshed by the LockRefreshingLockService ad infinitum despite their original threads being long-dead.
This is bad, but is critically bad when the locks are from old transactions and are suppressing the immutable timestamp.

I'm proposing this as a band-aid fix while we track down every case of locks being leaked. This is essentially automating the action that an FDE is currently doing manually to get his stack unblocked, using a dashboard that I wrote that tracks the immutable timestamp getting stuck along with a CLI that dumps lock state and another CLI that allows manual unlocking.

PDS-94791
PG-127836
PG-128400

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
@@ -38,8 +38,7 @@ import com.palantir.lock.LockService;
 import com.palantir.lock.SimpleHeldLocksToken;
 import com.palantir.lock.SimplifyingLockService;
 
-@SuppressWarnings("checkstyle:FinalClass") // Avoid breaking API in case someone extended this
-public class LockRefreshingLockService extends SimplifyingLockService {
+public final class LockRefreshingLockService extends SimplifyingLockService {
     public static final int REFRESH_BATCH_SIZE = 500_000;
     private static final Logger log = LoggerFactory.getLogger(LockRefreshingLockService.class);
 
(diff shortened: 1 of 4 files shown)


clockfort:
also why do we have our own TimeDuration class 😑

lock-impl/src/main/java/com/palantir/lock/impl/LockServiceImpl.java
```diff
@@ -777,13 +779,25 @@ private void logIfAbnormallyOld(ExpiringToken token, long now, Supplier<String>
...
+                            SafeArg.of("ageMillis", age),
+                            UnsafeArg.of("description", description.get()));
+                }
             }
         }
     }
 
+    private boolean isFromAtlasTransactionWithLockedImmutable(ExpiringToken token) {
+        return token.getClient() != null
+                && token.getClient().getClientId().startsWith(ATLAS_LOCK_PREFIX)
```
jeremyk-91: Note that this _only_ works for large internal product, because of how the lock clients are injected. We might want to add support for the LockClient `atlas instance` if we want this to apply more broadly to other atlas users

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -38,8 +38,7 @@
 import com.palantir.lock.SimpleHeldLocksToken;
 import com.palantir.lock.SimplifyingLockService;
 
-@SuppressWarnings("checkstyle:FinalClass") // Avoid breaking API in case someone extended this
```
clockfort: just so we're clear I did look and the only project who could have done this, didn't.
It won't cause a break and make @mswintermeyer fret

PR #4182, opened 2019-08-20
[Timelock Partitioning] Part 24: Further separation of rpc layer and logical layer.

felixdesouza:

**Goals (and why)**:
Artifacts of the rpc layer are leaking into the logical layer and making it somewhat annoying to work with. We extend the work done in #4177 to encapsulate all RPC related code into its own class.

* RPC layer deals with `RemoteException` and it also deals with the CJR bug.
* `BatchPaxosAcceptor` only deals in terms of `InvalidCacheKeyException` and `Optional`'s no mention of HTTP status codes or `RemoteException`s anywhere, that's all been pushed into `BatchPaxosAcceptorRpcClient`.
* As a result of this, the `BatchPaxosAcceptorResource` also deals with `Optional<..>` and any related remoting bits and then delegates to a `LocalBatchPaxosAcceptor` which as in the interface suggests deals with *just* the logical layer. 
  * turns out not doing this meant that different exceptions would be thrown depending on when you'd use a local acceptor or a remote one (the remote would do the right thing with the exceptions, the local (which would just be a resource) would be throwing the server side remote exceptions, meaning cache misses never caught. So this separation is pretty useful.

Mainly mechanical refactor, testing done in future branch should still hold, fortunately this is not out so I can amend at a later date if need to.

* `BatchPaxosAcceptorResource` & `LocalBatchPaxosAcceptor`
* `UseCaseBatchPaxosAcceptorAdapter` for localising rpc weirdness, converts rpc client into a logical client.
* `BatchingPaxosLatestSequenceCache`

ASAP

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptorRpcClient.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptorRpcClient.java
@@ -30,6 +30,7 @@ import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 
 import com.google.common.collect.SetMultimap;
+import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.paxos.BooleanPaxosResponse;
 import com.palantir.paxos.PaxosAcceptor;
 import com.palantir.paxos.PaxosPromise;
@@ -41,6 +42,9 @@ import com.palantir.paxos.PaxosProposalId;
         + "/" + PaxosTimeLockConstants.BATCH_INTERNAL_NAMESPACE
         + "/acceptor")
 public interface BatchPaxosAcceptorRpcClient {
+
+    ErrorType CACHE_KEY_NOT_FOUND =
+            ErrorType.create(ErrorType.Code.NOT_FOUND, "TimelockBatchPaxosAcceptor:CacheKeyNotFound");
 
     /**
      * Batch counterpart to {@link PaxosAcceptor#prepare}. For a given {@link Client} on paxos instance {@code seq},
@@ -89,8 +93,9 @@ public interface BatchPaxosAcceptorRpcClient {
      * acceptor has received multiple proposals at multiple sequence numbers for the same client past {@code cacheKey},
      * it will return the sequence number for the latest proposal.
      * <p>
-     * If a provided {@code cacheKey} has expired/is invalid, a {@code 404 Not Found} is thrown and this request
-     * should be retried without a {@code cacheKey} and also with a full set of clients to ensure a correct response.
+     * If a provided {@code cacheKey} has expired/is invalid, a {@code 404 Not Found} is returned with the above conjure
+     * error {@link BatchPaxosAcceptorRpcClient#CACHE_KEY_NOT_FOUND}. This request should be retried without a
+     * {@code cacheKey} and also with a full set of clients to ensure a correct response.
      * <p>
      * If an acceptor has received multiple proposals at multiple sequence numbers for a given client, only the latest
      * sequence number is returned for that client.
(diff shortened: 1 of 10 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
```diff
@@ -84,10 +69,15 @@
                 .addAll(cachedEntries.keySet())
                 .build();
         cachedEntries.clear();
-        return unsafeGetLatest(allClients);
+        try {
+            return unsafeGetLatest(allClients);
+        } catch (InvalidAcceptorCacheKeyException e) {
+            log.warn("");
```
jeremyk-91: nit: Throwing an exception isn't a criminal offence - we don't need to read the cache its Miranda rights

I think this is a bug if we get here?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
```diff
@@ -57,23 +56,9 @@
...
-            // TODO(fdesouza): Remove this once we've moved to CJR properly and the above works.
-            if (e.getMessage().contains(BatchPaxosAcceptor.CACHE_KEY_NOT_FOUND.name())) {
-                log.info("Cache key is invalid, invalidating cache - using deprecated detection method");
-                return handleCacheMiss(clients);
-            }
-
-            log.warn("received unexpected runtime exception", e);
-            throw e;
+        } catch (InvalidAcceptorCacheKeyException e) {
+            log.info("Cache key is invalid, invalidting cache - using deprecated detection method");
```
jeremyk-91: nit: `invalidating`

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalBatchPaxosAcceptorTests.java
```diff
@@ -128,16 +125,16 @@ public void weProxyAcceptRequests() {
...
+//    @Test
+//    public void throwsConjureRuntime404WhenCacheKeyIsNotFound() throws InvalidAcceptorCacheKeyException {
+//        AcceptorCacheKey cacheKey = AcceptorCacheKey.newCacheKey();
+//        when(cache.updatesSinceCacheKey(cacheKey))
+//                .thenThrow(new InvalidAcceptorCacheKeyException(cacheKey));
+//
+//        assertThatServiceExceptionThrownBy(() -> resource.latestSequencesPreparedOrAcceptedCached(cacheKey))
+//                .hasType(BatchPaxosAcceptorResource.CACHE_KEY_NOT_FOUND)
+//                .hasArgs(SafeArg.of("cacheKey", cacheKey));
+//    }
```
jeremyk-91: We probably do want a version of this test (except that it throws the `InvalidAcceptorCacheKeyException` rather than `ServiceException`) I think?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptorRpcClient.java
```diff
@@ -114,19 +119,19 @@ AcceptorCacheDigest latestSequencesPreparedOrAccepted(
...
-     * {@code 204 No Content}.
+     * {@code 204 No Content} which translates to {@link Optional#empty}.
      * <p>
      * In addition to the updates, a new {@code cacheKey} is provided to use on the next invocation of this method to
      * minimise on payload size as this method is on the hot path.
      *
      * @param paxosUseCase whether this is a timestamp paxos or leader paxos batch call
      * @return {@code 204 No Content} if there is no update, a digest containing updates plus a new cache key, or a
-     * {@code 412 Precondition Failed} if the cache key is not valid.
+     * {@code 404 Precondition Failed} if the cache key is not valid.
```
jeremyk-91: good catch. The message should be `Not Found` I think (but yeah, it is a 404, not a 412)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalBatchPaxosAcceptorTests.java
felixdesouza: oop my bad, I thought I had made an equivalent test for the resource, will fix up.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
felixdesouza: it is indeed a bug, will add exception message

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalBatchPaxosAcceptorTests.java
felixdesouza: I changed one of the tests that exercises the cache key not found logic to use the logical exception instead of the remote exception.

I'm going to make a `BatchPaxosAcceptorResourceTest` class, which tests the conversion.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchPaxosAcceptorRpcClient.java
jeremyk-91: (ping on this)

felixdesouza: oop, sorry I misinterpreted, fixed now.

PR #4183, opened 2019-08-20
[Conjure Java Runtime] AtlasDB HTTP Schema Version Headers

jeremyk-91:
**Goals (and why)**:
- Allow TimeLock Server (and leader-block users) to determine whether their clients plan on using CJR or not, and use this to decide how to map `NotCurrentLeaderExceptions` to HTTP responses. The eventual goal is that if the client specifies version 2, then if the leader is known we would return a `308`, and if not a `429` with a `Retry-After` of zero (or possibly a small value simulated probabilistically, since unfortunately this has to be specified in seconds).

- Add a new HTTP header that specifies this. This is prefixed `AtlasDb-` so unlikely to conflict with other applications.
- On the client-side, register an interceptor that adds the schema version header.
- On the server-side, grab the `HttpHeaders` from Jersey context and use them to decide how to map the `NotCurrentLeaderException` to a response.

Existing tests verify that current workflows still work properly. I manually tested that not specifying the interceptor works normally, and @gmaretic's previous testing suggests that CJR + legacy behaviour works fine.

- Is the header name too long? This of course adds overhead on requests. An alternative was to parse the user-agent string, but I felt that might be more hacky than I would like.
- Sanity check on the thread-safety of `NotCurrentLeaderExceptionMapper` would be good; research I did suggests that it is fine, but not 100% sure.

**Where should we start reviewing?**: AtlasDbHttpProtocolVersion


atlasdb-config/src/main/java/com/palantir/atlasdb/http/ExceptionMappers.java
```diff
@@ -48,6 +51,19 @@ public static Response encode503ResponseWithRetryAfter(Exception exception) {
...
 
+    /**
+     * Attempts to parse the {@link AtlasDbHttpProtocolVersion} indicated by {@link HttpHeaders}.
+     *
+     * Does not handle multiple values.
+     */
+    static Optional<AtlasDbHttpProtocolVersion> tryParseProtocolVersion(HttpHeaders headers) {
+        List<String> httpProtocolVersions = headers.getRequestHeader(AtlasDbHttpProtocolVersion.VERSION_HEADER);
+        String httpProtocolVersion = Iterables.getOnlyElement(httpProtocolVersions);
+        return Optional.ofNullable(httpProtocolVersion)
```
jeremyk-91: doh, not needed in new design

felixdesouza:
Somewhat confused, we can't add interceptors in CJR land, so how can we specify that we're in CJR land?

I would recommend we go via the `UserAgent` approach. We end up getting all of this for free basically. 

See [here](https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java#L252), `conjure-java-runtime` is added as an auxiliary user agent. All you'd have to do is call `UserAgents.parse` in the code where you're parsing the header, but instead parse the `UserAgent` header and you're good to go.

We don't have to worry about the indeterminate state, because up until now, we've not been producing the conjure-java-runtime user agent.

jeremyk-91:
Ah, didn't know interceptors aren't allowed with CJR. 

I did consider UserAgents, but my concern with that would be that if we want to move from CJR to some third protocol version, then we would need to maintain something on timelock that understands the CJR UserAgent format while protocol-2 versions exist, as opposed to only depending on our internal header. I'd also be hesitant about CJR behaviour breaks (though I guess in that case having the schema-version header doesn't really save you)

In any case doesn't look like there's a better option, so will change to that

atlasdb-config/src/test/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolVersionTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.palantir.conjure.java.api.config.service.UserAgent;
+import com.palantir.conjure.java.api.config.service.UserAgents;
+
+public class AtlasDbHttpProtocolVersionTest {
+    private static final UserAgent USER_AGENT_1 = createBaseUserAgent("Bond", "0.0.7");
+    private static final UserAgent USER_AGENT_2 = createBaseUserAgent("Smith", "1.2.3");
+
+    private static final UserAgent.Agent AGENT_1 = UserAgent.Agent.of("Shield", "24.7.0");
```
felixdesouza: hahaha this is brill

atlasdb-config/src/main/java/com/palantir/atlasdb/http/HttpProtocolAwareExceptionTranslator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public Response translate(HttpHeaders httpHeaders, E exception) {
+        AtlasDbHttpProtocolVersion protocolVersion = AtlasDbHttpProtocolVersion.inferFromHttpHeaders(httpHeaders);
+
+        switch (protocolVersion) {
+            case LEGACY_OR_UNKNOWN:
+                return httpProtocolHandler.handleLegacyOrUnknownVersion(exception);
+            case CONJURE_JAVA_RUNTIME:
+                QosException qosException = httpProtocolHandler.handleConjureJavaRuntime(exception);
+                return exceptionMappersProvider.get().findMapping(qosException).toResponse(qosException);
```
felixdesouza: nit: we should probably fail with something useful if we can't find the `QosException` mapper.

atlasdb-config/src/main/java/com/palantir/atlasdb/http/HttpProtocolAwareExceptionTranslator.java
```diff
@@ -0,0 +1,60 @@
...
+
+public class HttpProtocolAwareExceptionTranslator<E extends Exception> {
+    private static final Logger log = LoggerFactory.getLogger(HttpProtocolAwareExceptionTranslator.class);
+
+    private final AtlasDbHttpProtocolHandler<E> httpProtocolHandler;
+
+    @Inject
+    private Provider<ExceptionMappers> exceptionMappersProvider;
+
+    public HttpProtocolAwareExceptionTranslator(AtlasDbHttpProtocolHandler<E> httpProtocolHandler) {
```
felixdesouza: not sure how you get this when you don't instantiate via hk2 but manually

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotCurrentLeaderExceptionMapper.java
```diff
@@ -20,43 +20,35 @@
...
 public class NotCurrentLeaderExceptionMapper implements ExceptionMapper<NotCurrentLeaderException> {
-    private static final Logger log = LoggerFactory.getLogger(NotCurrentLeaderExceptionMapper.class);
-
     @Context
     private HttpHeaders httpHeaders;
 
-    /**
-     * Returns a 503 response, with body corresponding to the serialized exception.
-     */
+    private static final HttpProtocolAwareExceptionTranslator<NotCurrentLeaderException> translator = new
```
felixdesouza: is this for tests?

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotCurrentLeaderExceptionMapper.java
```diff
@@ -20,43 +20,35 @@
...
+            HttpProtocolAwareExceptionTranslator<>(new AtlasDbHttpProtocolHandler<NotCurrentLeaderException>() {
```
felixdesouza: a small nit I suppose, but it would be nice to see a `LambdaAtlasDbHttpProtocolHandler`

so you could do stuff like this:
```java
new HttpProtocolAwareExceptionTranslator<>(
  ProtocolHandler.of(
    underlyingException -> ExceptionMappers.encode503ResponseWithRetryAfter(underlyingException),
    $ -> QosException.unavailable()
  ));
```

Basically, I like @derive4j 😢

jeremyk-91:
Offline discussion: We should probably publish our own auxiliary agent in AtlasDB, so that needs to be added to this PR

jeremyk-91:
oops, messed up this PR

atlasdb-config/src/main/java/com/palantir/atlasdb/http/HttpProtocolAwareExceptionTranslator.java
jeremyk-91: This probably shouldn't be instantiated other than by the ExceptionMappers. I'll make some of these classes package visibility

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotCurrentLeaderExceptionMapper.java
jeremyk-91: No, this is where the logic for deciding what the things actually map to lives

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotCurrentLeaderExceptionMapper.java
jeremyk-91: Yeah, can add this. Makes sense

jeremyk-91:
Plan on doing the auxiliary agent in a later PR in the chain, once we're more decided on how the client side logic should look like. This version doesn't support the new clients just yet, and as far as I know there aren't issues with assuming the agent is there

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolHandler.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.function.Function;
+
+import javax.ws.rs.core.Response;
+
+import org.immutables.value.Value;
+
+import com.palantir.conjure.java.api.errors.QosException;
+
+interface AtlasDbHttpProtocolHandler<E extends Exception> {
```
felixdesouza: this is effectively the visitor interface/it's equivalent

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolVersion.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.ws.rs.core.HttpHeaders;
+
+import com.palantir.conjure.java.api.config.service.UserAgent;
+import com.palantir.conjure.java.api.config.service.UserAgents;
+
+public enum AtlasDbHttpProtocolVersion {
+    LEGACY_OR_UNKNOWN,
+    CONJURE_JAVA_RUNTIME;
+
+    public static AtlasDbHttpProtocolVersion inferFromHttpHeaders(HttpHeaders headers) {
```
felixdesouza: talked about this in #4192

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotCurrentLeaderExceptionMapper.java
felixdesouza: talked about this in #4192, abstract class means we can just pass in the lambda handler most of this away, if anything we can still have the translator but as a packpage private static class (if you're testing it, can be private if not and just rely on integration tests)

atlasdb-config/src/main/java/com/palantir/atlasdb/http/HttpProtocolAwareExceptionTranslator.java
```diff
@@ -29,27 +31,31 @@
...
         AtlasDbHttpProtocolVersion protocolVersion = AtlasDbHttpProtocolVersion.inferFromHttpHeaders(httpHeaders);
 
         switch (protocolVersion) {
             case LEGACY_OR_UNKNOWN:
                 return httpProtocolHandler.handleLegacyOrUnknownVersion(exception);
             case CONJURE_JAVA_RUNTIME:
                 QosException qosException = httpProtocolHandler.handleConjureJavaRuntime(exception);
-                return exceptionMappersProvider.get().findMapping(qosException).toResponse(qosException);
+                return Optional.ofNullable(exceptionMappersProvider.get())
+                        .flatMap(provider -> Optional.ofNullable(provider.findMapping(qosException)))
```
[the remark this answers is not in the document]
jeremyk-91: The null on the first line should probably only happen by product bug, if we inject it into a class that isn't wired up to receive the injections. 

That's cool, didn't know `null` in a `map()` makes it empty

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolVersion.java
jeremyk-91: Yeah, the abstract class model would probably work better - as discussed on that PR. I'll do that refactoring in 4192.

PR #4185, opened 2019-08-22
[Timelock Partitioning] Part 25: Wire up batch timestamp paxos

felixdesouza:

**Goals (and why)**:
At startup time, have the ability to decide to use batch paxos for the timestamps or not. The resources are registered either way. In addition to this config, also have the ability to use batch timestamp paxos.

So much wiring, could really do with dagger here, but alas, whatever.
* Change `PaxosTimestampBoundStore` to use the `Paxos(Acceptor|Learner)NetworkClient`s. 
* Change the setup of the `PaxosTimestampBoundStoreTest` to incorporate this, however we keep the tests the same so we have a comparison to the old impl.
* `PaxosTimestampCreator` now takes in (more or less) `Paxos(A|L)NetworkClientFactory` and creates clients on the fly, backed by either an autobatcher or the single approach.

I have a branch where it runs both of these tests in a suite, however the setup for this is rather large and detracts from the crux of this PR. The tests here are sufficient to show that this refactor/change is still valid for the single client case.

* Perhaps we move the config to runtime so we can have a quicker kill switch if something goes awry. Should we block endpoints if batch paxos is disabled, or have the two be independent i.e. enable batch paxos resource vs enable batch paxos client.

* For `PaxosTimestampBoundStore`, tests can pass with invalid behaviour which is somewhat worrying, not fixing it here as I believe they're covered by the bigger integration tests.

* The rollout of this feature. It's still generally safe because you can opt to use the batch version vs not, it's all going to the same paxos instances. However, we can have availability concerns. We don't have schema versions so we can't detect when we've reached certain features. So we want some sort of proxy where in the presence of 404s where the batch endpoints don't exist, we delegate to the single version or something along those lines, and then try again in 5 mins from the last 404 received or something.

* They're all using the same shared executor, that said, I don't believe it's doing anything different from current behaviour, although happy to be wrong here. It *is* something we probably should look at, although I don't believe `MultiplexingCompletionService` in its current form is what we want (we don't get thread reuse).

* Resource usage in general, carefully managed to not fail in a weird way such that we run out of threads, but this is incredibly fragile and we need a better way of managing resources, upon reset etc. I don't think I've made it significantly worse, but it's not better.

* `PaxosTimestampBoundStore`
* `PaxosTimestampBoundStoreTest`
* `PaxosTimestampCreator`
* `ClientPaxosResourceFactory`

🔥


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosTimestampBoundStore.java
```diff
@@ -207,15 +198,20 @@ SequenceAndBound forceAgreedState(long seq, @Nullable Long oldState) {
...
-                QUORUM_OF_ONE,
-                executor,
-                PaxosQuorumChecker.DEFAULT_REMOTE_REQUESTS_TIMEOUT).withoutRemotes();
-        return Optional.ofNullable(Iterables.getFirst(responses.get(), null))
+
+        PaxosResponses<PaxosContainer<Optional<PaxosLong>>> responses =
+                learnerClient.getLearnedValue(seq, maybeValue -> PaxosContainer.of(maybeValue
+                        .map(PaxosValue::getData)
+                        .map(PtBytes::toLong)
+                        .map(ImmutablePaxosLong::of)));
```
jeremyk-91: Note that this (I think) runs with the standard quorum, so we'd now wait for (generally) one more other node. In general looking back on it I have some concerns about the way we now use quorums in learners (right now we seek majority quorums, which seems unnecessary), though that's probably a separate discussion.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    @Value.Immutable
+    public interface ClientResources {
+        int quorumSize();
+        PaxosComponents components();
+        PaxosResource nonBatchedResource();
+        UseCaseAwareBatchPaxosResource batchedResource();
+        NetworkClientFactories networkClientFactories();
+        List<Closeable> closeables();
```
jeremyk-91: nit: May be worth formalising if we guarantee anything about the order in which these are closed

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -0,0 +1,264 @@
...
+
+        AutobatchingPaxosLearnerNetworkClientFactory learnerFactory =
+                AutobatchingPaxosLearnerNetworkClientFactory.create(allBatchLearners, sharedExecutor, quorumSize);
+
+        return ImmutableNetworkClientFactories.builder()
+                .acceptor(acceptorFactory::paxosAcceptorForClient)
+                .learner(learnerFactory::paxosLearnerForClient)
+                .addCloseables(acceptorFactory, learnerFactory)
+                .build();
+    }
```
jeremyk-91: nit: This is getting pretty long in general - I think I'd prefer if we split the `NetworkClientFactoriesFactory` (sigh) piece of this out to its own class

jeremyk-91:
First pass, before I return to interviews:

> Perhaps we move the config to runtime so we can have a quicker kill switch if something goes awry. Should we block endpoints if batch paxos is disabled, or have the two be independent i.e. enable batch paxos resource vs enable batch paxos client.

The only callers should be timelock, so in theory I would think that stopping clients would be sufficient (though no opposition to a blocking filter or something if we want to do that)

> For PaxosTimestampBoundStore, tests can pass with invalid behaviour which is somewhat worrying, not fixing it here as I believe they're covered by the bigger integration tests.

What specific behaviour are you thinking of?

> The rollout of this feature. It's still generally safe because you can opt to use the batch version vs not, it's all going to the same paxos instances. However, we can have availability concerns. We don't have schema versions so we can't detect when we've reached certain features. So we want some sort of proxy where in the presence of 404s where the batch endpoints don't exist, we delegate to the single version or something along those lines, and then try again in 5 mins from the last 404 received or something.

Yes, we've done that in the past with a bunch of the timelock endpoints - that strategy is fine. The underlying instances are the same, as you say. May be good to have a short integration test for different nodes using batched and unbatched and the sequences being linked (I haven't looked at the tests PR yet so it's possible you've covered that there).

> They're all using the same shared executor, that said, I don't believe it's doing anything different from current behaviour, although happy to be wrong here. It is something we probably should look at, although I don't believe MultiplexingCompletionService in its current form is what we want (we don't get thread reuse).

TODO - I don't think this is a massive problem (as you say this is, I think no worse than current behaviour)

> Resource usage in general, carefully managed to not fail in a weird way such that we run out of threads, but this is incredibly fragile and we need a better way of managing resources, upon reset etc. I don't think I've made it significantly worse, but it's not better.

TODO - I will have a second look at the closing code later on, but it looked reasonable (if not obviously so)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosTimestampBoundStore.java
felixdesouza: Yeah, I went back and forth on this one, I didn't really want to include a quorum size parameter because that would make the batching logic really awkward and we'd be lying somewhat re passing in a quorum size.

I suppose we could special this method and say `getLearnedValueQuorumOfOne` or something. If anything will change in a later PR.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
felixdesouza: The order is super important, otherwise we end up blowing up timelock. I've currently opted to not do anything (akin to what we're already doing) but it's super sketchy.

I think for the most part, we need to walk along the creation path per client and note down anything that is creating executors, and either memoize them per client or pass them down via constructor otherwise we'll get screwed every time.

It's also something to consider for when we roll out the multi-leadership stuff, we should in theory spin down resources used by clients that we aren't going to be leaders of. But I suppose it's not so different from todays world.

felixdesouza:
I will mix and match single vs batch after the test PR has gone in as I can just add it as another test entry. Not entirely sure what sort of integration test you'd want expect the `MultiNodePaxosTimeLockIntegrationTest`

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
felixdesouza: this has made the PR size go up, but the additions are mainly imports and whitespace lol from having many separate files, really wish I had dagger 😣

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosTimestampBoundStore.java
jeremyk-91: Yeah. I think in general the Learner at least for `getLearnedValue` and probably in general should probably be created with a quorum of 1 for reads; quorum doesn't really give you any additional correctness guarantees. Not a super-high priority, in any case.

PR #4186, opened 2019-08-22
[Timelock Partitioning] Part 26: Test suite to test both batch and non-batch timestamp paxos.

felixdesouza:

**Goals (and why)**:
In #4185, we introduced batch timestamp paxos, but didn't have any integration tests for it. This introduces parameterized suites i.e. I want to run a test with one large external resource, and then I want to run that same test with a different external resource.

* Introduced a new test set `suiteTest`, and when this runs test, it only runs tests in the `suite` package. This means we don't rerun tests. 
* Added some scaffolding to ensure that we can still run the tests in a non-suite configuration in the IDE for quick debugging.
* Changed the logging formats that made debugging a bit easier i.e. which node is receiving requests/logging specific logs.

It's all tests!

* `PaxosSuite`
* `MultiNodePaxosTimeLockServerIntegrationTest`

everything else is just config really

🚒


atlasdb-client/build.gradle
```diff
@@ -41,7 +41,7 @@ configurations {
 }
 
 artifacts {
-    testArtifacts myTestJar
+    testArtifacts clientTestJar
```
jeremyk-91: 💯

gradle/publish-jars.gradle
```diff
@@ -13,7 +13,7 @@ task sourceJar(type: Jar) {
     classifier 'sources'
 }
 
-task testJar(type: Jar) {
+task myTestJar(type: Jar) {
```
jeremyk-91: I'm... not sure about the logic for task naming, but sure

timelock-server/src/testCommon/resources/paxosMultiServer1.yml
```diff
@@ -21,7 +21,17 @@ runtime:
   paxos:
     leader-ping-response-wait-in-ms: 1000
 
+logging:
+  appenders:
+    - type: console
+      logFormat: "server-1 %-5p [%d{ISO8601,UTC}] %c: %m%n%rEx"
```
jeremyk-91: Ah, this is good. Thanks

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -144,55 +145,55 @@ public void nonLeadersReturn503() {
...
 
     @Test
     public void canPerformRollingRestart() {
         bringAllNodesOnline();
-        for (TestableTimelockServer server : CLUSTER.servers()) {
+        for (TestableTimelockServer server : cluster.servers()) {
             server.kill();
-            waitForClusterToStabilize();
-            CLUSTER.getFreshTimestamp();
+//            waitForClusterToStabilize();
```
jeremyk-91: Is this intentional? I'm not sure how long start/kill take, but this seems like it could be dangerous if we kill the leader and then try and get a fresh timestamp, especially if one of the other nodes is just starting up. I haven't looked at how `start` works, so it's possible it's not needed

(If definitely not needed, then the comment should probably be removed - and possibly one added on why this is safe)

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -55,16 +57,23 @@
...
     private static final List<String> ADDITIONAL_CLIENTS = ImmutableList.of(CLIENT_2, CLIENT_3);
 
-    private static final TestableTimelockCluster CLUSTER = new TestableTimelockCluster(
-            "https://localhost",
-            "paxosMultiServer0.yml",
-            "paxosMultiServer1.yml",
-            "paxosMultiServer2.yml");
+    @ClassRule
+    public static ParameterInjector<TestableTimelockCluster> injector =
+            new ParameterInjector<>(() -> PaxosSuite.NON_BATCHED_PAXOS);
```
jeremyk-91: I think it would be good to have this go through a static-factory for purposes of naming, it wasn't obvious without looking through `ParameterInjector` that this is supposed to be the configuration provided if running outside a suite

timelock-server/build.gradle
```diff
@@ -52,5 +67,13 @@ dependencies {
...
+    testCommonImplementation group: 'com.github.peterwippermann.junit4', name: 'parameterized-suite'
+    testCommonImplementation group: 'com.ea.agentloader', name: 'ea-agent-loader'
+    testCommonImplementation group: 'io.dropwizard', name: 'dropwizard-testing'
+    testCommonImplementation (project(":atlasdb-tests-shared")) {
+        exclude group: 'com.fasterxml.jackson.jaxrs'
+    }
+
+    testCommonApi group: 'io.dropwizard', name: 'dropwizard-testing'
+    testCommonApi group: 'com.github.peterwippermann.junit4', name: 'parameterized-suite'
+}
```
jeremyk-91: nit: line

gradle/publish-jars.gradle
felixdesouza: so somehow they're conflicting and it's just a massive PITA, so I just called it something different lol not great but 🤷‍♂

felixdesouza: renamed to `testJarsForPublishing`

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
felixdesouza: good shout

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
felixdesouza: technically not needed, tests passed just fine, but I'd rather not reason about it, so I just inlined what the previous thing did into this method. The only thing I removed was the ability to do it as a `@ClassRule` which can be a bit awkward but I think it's fine. I'm not sure what the value is (in the current impl) of waiting for "leadership" of multiple clients at the beginning of the test (suite), and doesn't seem to affect results 🤷‍♂

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -144,55 +145,55 @@ public void nonLeadersReturn503() {
...
 
     @Test
     public void canPerformRollingRestart() {
         bringAllNodesOnline();
-        for (TestableTimelockServer server : CLUSTER.servers()) {
+        for (TestableTimelockServer server : cluster.servers()) {
             server.kill();
-            waitForClusterToStabilize();
-            CLUSTER.getFreshTimestamp();
+            cluster.waitUntilAllServersOnlineAndReadyToServeClients(ADDITIONAL_CLIENTS);
```
jeremyk-91: I think this should be `cluster.waitUntilLeaderIsElected()`. This one actually starts the dead servers again (before we get the fresh timestamp), which is probably not what the test wants to do

PR #4192, opened 2019-08-28
[Conjure Java Runtime] Part 2: Server Infrastructure - New ExceptionMappers, Custom Auxiliary Agents

jeremyk-91:
**Goals (and why)**:
- Capture the semantics of how we want clients to interpret exceptions.
- Implement the server side infrastructure required for timelock to work with new ExceptionMappers.

- Make the other exception mappers (Blocking Timeout and Not Initialized) support CJR semantics, with justifications
- Define our own auxiliary agent; eventually we will add this to a user-provided UserAgent before creating a `JaxRsClient`

Nothing much. Existing tests, in particular MNPTLSIT would verify that things still work and this doesn't break things too aggressively.

- Does the reasoning behind the decisions for the various exception mappers make sense?
- Is the overhead of using our custom user-agent and searching for its presence too high?

**Where should we start reviewing?**: AtlasDbHttpProtocolVersion


atlasdb-config/src/main/java/com/palantir/atlasdb/http/BlockingTimeoutExceptionMapper.java
```diff
@@ -15,14 +15,37 @@
...
+/**
+ * Converts {@link BlockingTimeoutException}s into appropriate status responses, depending on the user's
+ * {@link AtlasDbHttpProtocolVersion}. The intention is that clients should retry on the same node (as in the absence
+ * of exceptional circumstances, it would still be the leader), and they may do so immediately (as an individual lock
+ * being locked does not imply that the server is struggling).
+ *
+ * This is a 503 without a Retry-After header and with a message body corresponding to {@link BlockingTimeoutException}
+ * in {@link AtlasDbHttpProtocolVersion#LEGACY_OR_UNKNOWN}.
+ */
 public class BlockingTimeoutExceptionMapper implements ExceptionMapper<BlockingTimeoutException> {
```
felixdesouza: As much as I hate to say it, let's have a abstract class:

```java
abstract class ProtocolAwareExceptionMapper<T> implements ExceptionMapper<T> {

  private final HttpProtocolAwareExceptionTranslator<T> translator;
  @Context private HttpHeaders httpHeaders;
  
  public ProtocolAwareExceptionMapper(AtlasDbHttpProtocolHandler<T> handler) {
    ...
  }

  @Override
  public Response toResponse(T exception) {
    return translator.translate(httpHeaders, exception);
  }
} 
```

with impls:

```java
public class BlockingTimeoutExceptionMapper extends ProtocolAwareExceptionMapper<BlockingTimeoutException> {
  BlockingTimeoutExceptonMapper() {
     super(LambdaHandler.of(...));
  }
}
```

In an ideal world, we'd just create the exception mapper and pass in the protocol handler ourselves, but I'm not sure how that plays with the `@Context` injection. If it *does* work, I'd prefer it this way - that is without the abstract class.

felixdesouza: I think that means we can get rid of the `HttpProtocolAwareExceptionTranslator` and just inline its implementation, what do you think?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbAgents.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.http;
+
+import com.palantir.conjure.java.api.config.service.UserAgent;
+
+public final class AtlasDbAgents {
+    public static final String ATLASDB_HTTP_CLIENT = "atlasdb-http-client";
+    public static final AtlasDbHttpProtocolVersion CURRENT_CLIENT_PROTOCOL_VERSION
```
felixdesouza: nit: not an agent

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolVersion.java
```diff
@@ -16,16 +16,34 @@
...
+
+    AtlasDbHttpProtocolVersion(String protocolVersionString) {
+        this.protocolVersionString = protocolVersionString;
+    }
+
+    public String getProtocolVersionString() {
+        return protocolVersionString;
+    }
 
     public static AtlasDbHttpProtocolVersion inferFromHttpHeaders(HttpHeaders headers) {
```
felixdesouza: not a fan of `HttpHeaders` here, do all the request processing inside the filter/exception mapper bla bla, and this should just concern itself with mapping string to `AtlasDbHttpProtocolVersion`

probably should go in the other PR, don't care where

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbAgents.java
```diff
@@ -0,0 +1,31 @@
...
+            = AtlasDbHttpProtocolVersion.LEGACY_OR_UNKNOWN;
+    public static final UserAgent.Agent ATLASDB_CONJURE_JAVA_RUNTIME_HTTP_AGENT
```
felixdesouza: name is misleading if the current client points to legacy

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotInitializedExceptionMapper.java
```diff
@@ -15,18 +15,34 @@
...
+ * This is a 503 without a Retry-After header and with a message body corresponding to {@link NotInitializedException}
+ * in {@link AtlasDbHttpProtocolVersion#LEGACY_OR_UNKNOWN}.
  */
 public class NotInitializedExceptionMapper implements ExceptionMapper<NotInitializedException> {
+    @Context
+    private HttpHeaders httpHeaders;
+
+    private static final HttpProtocolAwareExceptionTranslator<NotInitializedException> translator = new
+            HttpProtocolAwareExceptionTranslator<>(
+            AtlasDbHttpProtocolHandler.LambdaHandler.of(
```
felixdesouza: nit: torn between using the builder of `LambdaHandler` to make it a bit more obvious as to what each parameter means e.g.:

```java
AtlasDbHttpProtocolHandler.LambdaHandler.builder()
  .legacyBehaviour(...)
  .version2Behaviour(...)
  .build();
```

We basically have a visitor right?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbAgents.java
jeremyk-91: Discussed offline - yeah the constants here other than the first were a bit premature, only the first is really needed now. The other ones are used in part 4.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbAgents.java
jeremyk-91: Discussed offline - yep, the previous one _should_ be CJR

jeremyk-91:
Note: Original design doesn't fully work because the injected things need to be in the exceptionmapper classes which are registered. The abstract class approach should dodge these

atlasdb-config/src/main/java/com/palantir/atlasdb/http/BlockingTimeoutExceptionMapper.java
jeremyk-91: Yep, will test this out. As discussed this design would be good (if it works)

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/AtlasDbAgents.java
jeremyk-91: (Hindsight: I think here you meant that these constants shouldn't be in a class called `Agents`. I've renamed this to `AtlasDbRemotingConstants`)

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotInitializedExceptionMapper.java
jeremyk-91: This makes sense - though the lambda handler shouldn't exist any more as we move to an abstract class.

atlasdb-config/src/main/java/com/palantir/atlasdb/http/NotInitializedExceptionMapper.java
```diff
@@ -16,17 +16,26 @@
...
+import com.palantir.conjure.java.api.errors.QosException;
 import com.palantir.exception.NotInitializedException;
 
 /**
- * Maps a {@link NotInitializedException} to a 503 status code.
+ * Converts {@link NotInitializedException}s into appropriate status responses, depending on the user's
+ * {@link AtlasDbHttpProtocolVersion}. The intention is that clients may retry on an arbitrary node of the service,
+ * and they should backoff, as that node of the cluster is still starting up.
+ *
+ * This is a 503 without a Retry-After header and with a message body corresponding to {@link NotInitializedException}
```
felixdesouza: nit: put `Retry-After` in `{@code Retry-After}`

atlasdb-config/src/test/java/com/palantir/atlasdb/http/AtlasDbHttpProtocolVersionTest.java
```diff
@@ -17,83 +17,33 @@
...
-
     @Test
-    public void detectsConjureJavaRuntimeAgent() {
-        when(httpHeaders.getRequestHeader(HttpHeaders.USER_AGENT)).thenReturn(
-                ImmutableList.of(UserAgents.format(USER_AGENT_1.addAgent(CONJURE_AGENT))));
-
-        assertThat(AtlasDbHttpProtocolVersion.inferFromHttpHeaders(httpHeaders))
-                .isEqualTo(AtlasDbHttpProtocolVersion.CONJURE_JAVA_RUNTIME);
+    public void infersFromKnownVersionStrings() {
+        Arrays.stream(AtlasDbHttpProtocolVersion.values())
```
felixdesouza: would prefer if this was written out explicitly for each `AtlasDbHttpProtocolVersion`

jeremyk-91: Sure, makes sense. Will fix

PR #4198, opened 2019-09-02
Upgrade to Gradle 5

j-baker:


j-baker:
Migrating annotation processors to native isn't done yet - there's an excavator that will finish it off after this merges.

PR #4199, opened 2019-09-02
Factor out timestamp cache

j-baker:
Internally, we have many AtlasDBs, each with their own timestamp cache.
We'd like to have one big one, with a single expiry policy - basically
because we think that it'll lead us to a more stable situation with less
GC-ing and OOMing.


atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultTimestampCache.java
```diff
@@ -0,0 +1,84 @@
...
+                .recordStats()
+                .build();
+    }
+
+    public DefaultTimestampCache(MetricRegistry metricRegistry, Supplier<Long> size) {
+        this.size = size;
+        startToCommitTimestampCache = createCache(size.get());
+        evictionPolicy = startToCommitTimestampCache.policy().eviction().get();
+        AtlasDbMetrics.registerCache(metricRegistry, startToCommitTimestampCache,
+                MetricRegistry.name(DefaultTimestampCache.class, "startToCommitTimestamp"));
```
jeremyk-91: This might be a break, but probably fine

felixdesouza:
Metric name seemed to come up as a test failure, the other test failure is a flake :(

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/TimestampCache.java
```diff
@@ -13,85 +13,30 @@
...
      * effectively using the timestamp table as existing concurrency control for who wins a commit.
      *
      * @param startTimestamp transaction start timestamp
      * @param commitTimestamp transaction commit timestamp
      */
-    public void putAlreadyCommittedTransaction(Long startTimestamp, Long commitTimestamp) {
-        startToCommitTimestampCache.put(startTimestamp, commitTimestamp);
-    }
-
+    void putAlreadyCommittedTransaction(Long startTimestamp, Long commitTimestamp);
```
jeremyk-91: This name feels kind of awkward (IIRC we cache -1s) but we should preserve this

PR #4201, opened 2019-09-02
Stream Store Metadata Cleanup Task Recovery

### LINKED ISSUE #4190: Stream Store Deaths At Specific Times Leak Streams
Suppose that we write to the stream store nontransactionally. In doing this, we write metadata that says we're STORING the stream, write the blocks, mark it STORED, and then need to manually mark it as used. Each of these is done in distinct transactions.

If someone finishes writing the stream, but never marks it as used, then the data from that stream lingers permanently:
- The metadata cleanup task, which is supposed to fix streams that were never fully uploaded, won't delete it because it *is* stored. 
- The index cleanup task, which is supposed to deal with streams that had their reference counts hit zero, would never have been written to for that stream ID because it was never marked, and so it won't delete it either.

The most intuitive fix (to me) here is probably to make the metadata cleanup task also eliminate things that are stored but unmarked. Also possible is to transactionally mark and unmark a stream before storeStream returns. However, I'd like a proof of correctness before proceeding.

Proof to follow in comments (I'm looking at this today).

@diogoholanda for SA

jeremyk-91:
**Goals (and why)**:
- Fix #4190 - this is an implementation of the method described in the first comment there

- When running the metadata cleanup task, if a stream ID is stored but has zero entries in the index table, delete it.

- Existing ETE tests should catch egregious breaks.
- I ~will see if I can~ added one ETE test that requires this behaviour - write streams without marking them, perform an iteration of targeted sweep, and then verify that in this case the stream data is cleared.

- Correctness. Are there any cases where this deletes things that shouldn't be deleted? Consider the proof/argument described on the issue.
- Performance. This is not a hot codepath so I'm less concerned here, but we are basically trying to find rows that are not-stored OR stored AND have nothing in the index table. There may be some clever way of reducing the size of the queries that I haven't thought about.

**Where should we start reviewing?**: `StreamStoreRenderer`

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
@@ -794,13 +794,10 @@ public class StreamStoreRenderer {
                     line("        .map(", StreamIndexRow, "::getId)");
                     line("        .map(", StreamMetadataRow, "::of)");
                     line("        .collect(Collectors.toSet());");
-                    line("Map<", StreamMetadataRow, ", StreamMetadata> currentMetadata = metaTable.getMetadatas(");
-                    line("        Sets.difference(rows, rowsWithNoIndexEntries));");
-                    line("Set<", StreamId, "> toDelete = Sets.newHashSet(rowsWithNoIndexEntries.stream()");
-                    line("        .map(", StreamMetadataRow, "::getId)");
-                    line("        .collect(Collectors.toSet()));");
+                    line("Map<", StreamMetadataRow, ", StreamMetadata> currentMetadata = metaTable.getMetadatas(rows);");
+                    line("Set<", StreamId, "> toDelete = Sets.newHashSet();");
                     line("for (Map.Entry<", StreamMetadataRow, ", StreamMetadata> e : currentMetadata.entrySet()) {"); {
-                        line("if (e.getValue().getStatus() != Status.STORED) {"); {
+                        line("if (e.getValue().getStatus() != Status.STORED || rowsWithNoIndexEntries.contains(e.getKey())) {"); {
                             line("toDelete.add(e.getKey().getId());");
                         } line("}");
                     } line("}");
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
@@ -721,6 +721,10 @@ public class StreamStoreRenderer {
         final String StreamMetadataTable = name + "StreamMetadataTable";
         final String StreamMetadataRow = StreamMetadataTable + "." + name + "StreamMetadataRow";
 
+        final String StreamIndexTable = name + "StreamIdxTable";
+        final String StreamIndexRow = StreamIndexTable + "." + name + "StreamIdxRow";
+        final String StreamIndexColumnValue = StreamIndexTable + "." + name + "StreamIdxColumnValue";
+
         final String TableFactory = schemaName + "TableFactory";
         final String MetadataCleanupTask = getMetadataCleanupTaskClassName();
         final String StreamId = streamIdType.getJavaObjectClassName();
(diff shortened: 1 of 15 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -768,14 +775,29 @@ private void cellsCleanedUp() {
                 line("@Override");
                 line("public boolean cellsCleanedUp(Transaction t, Set<Cell> cells) {"); {
                     line(StreamMetadataTable, " metaTable = tables.get", StreamMetadataTable, "(t);");
-                    line("Collection<", StreamMetadataRow, "> rows = Lists.newArrayListWithCapacity(cells.size());");
+                    line("Set<", StreamMetadataRow, "> rows = Sets.newHashSetWithExpectedSize(cells.size());");
```
felixdesouza: is it possible to add a comment in this file? not necessarily so that it renders as a comment in the generated code, but in this file itself?

wrangling with atlas types directly isn't inherently readable

atlasdb-ete-test-utils/src/main/java/com/palantir/atlasdb/todo/generated/SnapshotsMetadataCleanupTask.java
```diff
@@ -25,14 +28,29 @@ public SnapshotsMetadataCleanupTask(Namespace namespace) {
...
         }
+        SnapshotsStreamIdxTable indexTable = tables.getSnapshotsStreamIdxTable(t);
+        Set<SnapshotsStreamIdxTable.SnapshotsStreamIdxRow> indexRows = rows.stream()
+                .map(SnapshotsStreamMetadataTable.SnapshotsStreamMetadataRow::getId)
+                .map(SnapshotsStreamIdxTable.SnapshotsStreamIdxRow::of)
+                .collect(Collectors.toSet());
+        Map<SnapshotsStreamIdxTable.SnapshotsStreamIdxRow, Iterator<SnapshotsStreamIdxTable.SnapshotsStreamIdxColumnValue>> indexIterator
+                = indexTable.getRowsColumnRangeIterator(indexRows,
+                        BatchColumnRangeSelection.create(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 1));
+        Set<SnapshotsStreamMetadataTable.SnapshotsStreamMetadataRow> rowsWithNoIndexEntries
```
felixdesouza: commenting here because it's easier to read,

if I've understood correctly, index table keeps track of references in dynamic columns
and you're just checking that for a given stream, its index row has no columns (references)?

perhaps rename `indexIterator` to `referenceIteratorByStream` and

`rowsWithNoIndexEntries` to `streamsWithNoReferences`

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TargetedSweepEteTest.java
```diff
@@ -93,6 +93,15 @@ public void targetedSweepLargeStreamsTest() {
         assertDeletedAndSwept(4, 4, 4, 4 * 4);
     }
 
+    @Test
+    public void targetedSweepCleanupUnmarkedStreamsTest() {
+        todoClient.storeUnmarkedSnapshot("snap");
+        todoClient.storeUnmarkedSnapshot("crackle");
+        todoClient.storeUnmarkedSnapshot("pop");
+        todoClient.runIterationOfTargetedSweep();
+        assertDeleted(0, 3, 3, 3);
```
felixdesouza: lol wat????

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
jeremyk-91: (As mentioned in overall review, renaming the variables would suffice)

atlasdb-ete-test-utils/src/main/java/com/palantir/atlasdb/todo/generated/SnapshotsMetadataCleanupTask.java
jeremyk-91: Correct, that's what the code is doing (and what you describe would be more descriptive at the application level of what is going on - will change to those)

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TargetedSweepEteTest.java
jeremyk-91:
```
    private void assertDeleted(long idx, long hash, long meta, long val) {
        Assert.assertThat(todoClient.numberOfCellsDeleted(INDEX_TABLE), equalTo(idx));
        Assert.assertThat(todoClient.numberOfCellsDeleted(HASH_TABLE), equalTo(hash));
        Assert.assertThat(todoClient.numberOfCellsDeleted(METADATA_TABLE), equalTo(meta));
        Assert.assertThat(todoClient.numberOfCellsDeleted(VALUES_TABLE), equalTo(val));
    }
```

It's kind of horrible but I didn't really fancy inlining everything in to the existing tests. I'll leave a comment why the numbers are as they are

PR #4203, opened 2019-09-03
[WiP] Bump container setup timeout to 10 minutes, rather than 2

jeremyk-91:
**Goals (and why)**: Have less flaky builds

**Implementation Description (bullets)**: Bump container setup timeout to 10m

**Testing (What was existing testing like?  What have you done to improve it?)**: We'll see if builds are happier. In progress.

**Concerns (what feedback would you like?)**: We need to root cause the underlying

**Where should we start reviewing?**: +1/-1

**Priority (whenever / two weeks / yesterday)**: if we're happy with this, yesterday


jeremyk-91:
Still fails. So there's a probability that the cluster goes into a bad state, and when this happens it doesn't recover (or at least doesn't recover in 10 mins, which makes me think it won't recover at all)

jeremyk-91:
Closing out. #4213 is the actual fix, and this did nothing.

PR #4205, opened 2019-09-03
[Conjure Java Runtime] Part 4: Server-Side 308 Construction

jeremyk-91:
**Goals (and why)**:
- Support not-current-leader style workflows with 308s in CJR land.

- Extend NCLEM to be able to fire not-current-leader-exceptions that are 308 with a location header for Conjure Java Runtime.
- Have a RedirectRetryTargeter that figures out where to point at.

- Added a couple of tests for the RedirectRetryTargeter, plus a quick one for PaxosRemotingUtils.

- Technically the `RedirectRetryTargeter` does two things: it picks a new base URI, and then it does prefix matching, so if you have `Scheme1Host1:Port1Path1` and `Scheme2Host2:Port2Path2` as base URIs and a request made to `Scheme1Host1:Port1Path1Path3`, then this needs to return `Scheme2Host2:Port2Path2Path3`. I thought about splitting out the first piece to a separate class/interface (possible future work to re-enable leader hinting, for example), but it felt like there were a lot of small classes.
- I made `RedirectRetryTargeter` an `Optional` in `NotCurrentLeaderExceptionMapper` to avoid breaking the API, and to play nicely with consumers that use a leader block. These don't have information about hosts and just fall back to 503. Is this a reasonable thing to do? 
- If retry targeting blows up we generally just fall back to 503s. Does this make sense? I think it does as remoting will recover and find the right node, just perhaps with some maybe less-than-desirable exponential backoff in between.
- The manipulating of URL components feels a bit fragile to me, though I don't really see a better way to do this, and lifting components directly off the config and just returning that doesn't work, because remoting requires that the hint comes from the same set of base URLs that the client knows about, which is crucially *not* the same as that on the server.

**Where should we start reviewing?**: `RedirectRetryTargeterTest`, probably.

### DIFF
--- a/atlasdb-config/src/test/java/com/palantir/atlasdb/http/RedirectRetryTargeterTest.java
+++ b/atlasdb-config/src/test/java/com/palantir/atlasdb/http/RedirectRetryTargeterTest.java
@@ -35,7 +35,7 @@ public class RedirectRetryTargeterTest {
     private static final int PORT_2 = 4222;
 
     private static final String PATH_1 = "/api/masyu/";
-    private static final String PATH_2 = "/api/snake/";
+    private static final String PATH_2 = "/kpi/snake/";
     private static final String REQUEST_PATH_1 = "request/jewelled-branch-of-hourai";
     private static final String REQUEST_PATH_2 = "/request/swallows-cowrie-shell";
 
--- a/atlasdb-config/src/test/java/com/palantir/atlasdb/http/RedirectRetryTargeterTest.java
+++ b/atlasdb-config/src/test/java/com/palantir/atlasdb/http/RedirectRetryTargeterTest.java
@@ -1,0 +1,97 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.http;
+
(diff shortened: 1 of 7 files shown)


felixdesouza:
> The manipulating of URL components feels a bit fragile to me, though I don't really see a better way to do this, and lifting components directly off the config and just returning that doesn't work, because remoting requires that the hint comes from the same set of base URLs that the client knows about, which is crucially not the same as that on the server.

I agree. So, this could be an opportunity to undo some string mangling that we've done in a previous previous life. Namely [here](https://github.com/palantir/atlasdb/blob/develop/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L903).

If instead each method of `TimelockRpcClient` - and analogously `LockService` and `TimestampManagementService` I think? - took in a `client` as a path param, (this requires no changes in the server), then we can create a `RemoteTimelockServiceAdapter` that takes in a `client/namespace` that we pass to the `TimelockRpcClient` delegate. Then resulting code is just returning the next host.

> I made `RedirectRetryTargeter` an `Optional` in `NotCurrentLeaderExceptionMapper` to avoid breaking the API, and to play nicely with consumers that use a leader block. These don't have information about hosts and just fall back to 503. Is this a reasonable thing to do?

It would be good to get rid of this method at a later date, once we've fixed up internal product to use the new `RedirectRetryTargeter`.

atlasdb-config/src/test/java/com/palantir/atlasdb/http/RedirectRetryTargeterTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableList;
+
+public class RedirectRetryTargeterTest {
+    private static final URL URL_1;
+    private static final URL URL_2;
+    private static final URL URL_3;
+
+    static {
```
felixdesouza: *sniff 😢

OStevan: It might be cleaner to have a helper method `createURL` which would deal with the exceptions. That way we could remove the static block.

jeremyk-91: :+1: Yep, will use @OStevan's suggestion

PR #4209, opened 2019-09-05
Do not log CassandraClients.

jkozlowski:
**Goals (and why)**:

Fix logging errors.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -143,8 +143,7 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
             return fn.apply(resource);
         } catch (Exception e) {
             if (isInvalidClientConnection(resource)) {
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -143,8 +143,7 @@ public int getPoolSize() {
             return fn.apply(resource);
         } catch (Exception e) {
             if (isInvalidClientConnection(resource)) {
-                log.warn("Not reusing resource {} due to {} of host {}",
-                        UnsafeArg.of("resource", resource),
```
jkozlowski: Logging framework is complaining of not being able to serialize this QosCassandraClient as JSON. We can investigate why it is failing that, but either way we should likely not rely on that behavior being correct.

If we want to maintain the logging here for debugging purposes, I suggest we invent a synthetic id (incrementing counter every time QosCassandraClient is instantiated) or force the jackson serialization to just return #toString(). But that class has fields and jackson is likely trying to serialize those and failing.

jkozlowski: Happy to followup with something, but would like to fix this first, because it is failing upgrades.

PR #4210, opened 2019-09-05
Mark node as not eligible for leadership.

jkozlowski:
**Goals (and why)**:

Improve graceful shutdown logic, FLUP from https://github.com/palantir/atlasdb/pull/4208.


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -156,7 +170,12 @@ private boolean pingLeader(Optional<PaxosValue> maybeGreatestLearnedValue) {
...
     @Override
     public Set<PingableLeader> getPotentialLeaders() {
-        return Sets.union(ImmutableSet.of(localPingable), ImmutableSet.copyOf(otherPingables));
+        // TODO(jakubk) Not sure about the implications here
+        if (leaderEligible.get()) {
+            return Sets.union(ImmutableSet.of(localPingable), ImmutableSet.copyOf(otherPingables));
+        } else {
+            return ImmutableSet.copyOf(otherPingables);
+        }
     }
```
jeremyk-91: I think this one should return the original `Sets.union(...)` - this is used by health checks in TimeLock and in LargeInternalProduct that check that a quorum of nodes is reachable, among other things - so removing ourselves from the picture _could_ make LargeInternalProduct healthchecks report quorum when there is no quorum, as they do sometimes run their lock services in even numbers.

jkozlowski: Yep, I looked at the code around it too and wasn't convinced I want to touch this.

PR #4213, opened 2019-09-05
[Build Flakes] Resolve Race Condition in ETE Test Container Setup

jeremyk-91:
**Goals (and why)**:
- Actually have builds that work under most normal circumstances, rather than having to retry five or six times before a release gets to prod.

- When seeding the async puncher, run that with a one second timeout, and should that fail just don't seed.
- The builds were flaking because successfully bringing up the ETE Docker containers was dependent on the docker proxy rule having already been set up by the time the ETE server containers talked to one another to elect a leader.
- I hypothesise that the reason this manifested as DbKvs failures (but interestingly much, much less on Cassandra Multinode) was because historically the Cassandra containers take longer than Postgres before they're fully ready, and we need to create tables and indices before we actually get to grabbing the seed timestamp for the puncher.

See internal builds on the branch `jkong/flaky-2`; that is this branch jury-rigged to run seven instances of the DbKvsIntegrationTest on containers 1-7. We have a 14/14 success rate.

- We may have some false positives in the targeted sweep monitor, but given that that pages at P1 I'm not too worried.
- Are there some funky conditions where the executor blows up and is sad?

AsyncPuncher

**Priority (whenever / two weeks / yesterday)**: yesterday. Actually the day we made the first PR.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/AsyncPuncher.java
@@ -16,6 +16,7 @@
 package com.palantir.atlasdb.cleaner;
 
 import java.time.Duration;
+import java.util.Optional;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


felixdesouza:
Might it be worth just spinning up a background thread which runs every 30 mins to punch?

PR #4215, opened 2019-09-06
Improve Throwables documentation

jeremyk-91:
**Goals (and why)**:
- See internal issue securityproduct#7650
- ~The idea is that internal clients are interrupting requests that may have Atlas on the path and we don't always respect that~ We do, but I got confused by the docs.

- ~Propagate the interrupt flag before throwing an exception out.~
- Update the docs of Throwables to highlight the difference from the Guava implementation.

N/A, existing tests should account for this

- ~Is this actually the right thing to do in each of these cases? I think so, since otherwise we'll lose the interrupt flag within Atlas (whether in the Legacy Timelock Service or on DBKVS code paths).~
- Are the docs readable?

**Where should we start reviewing?**: Throwables.java

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/base/Throwables.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/base/Throwables.java
@@ -33,6 +33,13 @@ import com.palantir.common.exception.PalantirRuntimeException;
 import com.palantir.exception.PalantirInterruptedException;
 import com.palantir.logsafe.Preconditions;
 
+/**
+ * Utilities for creating and propagating exceptions.
+ *
+ * Note: Differently from Guava, the methods in this class handle interruption; that is, they will re-set the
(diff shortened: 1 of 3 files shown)


OStevan:
After taking a look at `Throwables.throwUncheckedException` and `Throwables.rewrapAndThrowUncheckedException`which call [`Throwables.createPalantirRuntimeException`](https://github.com/palantir/atlasdb/blob/4392b270383f92078e61d478091621341d0972db/atlasdb-commons/src/main/java/com/palantir/common/base/Throwables.java#L113) seems like the lines added are already there when processing raised exception with `Throwables`. In my opinion the added lines are redundant.

jeremyk-91:
Good spot - didn't see that it was `atlasdb-commons/.../Throwables` (vs. `guava/.../Throwables`). 

The added lines are behaviourally redundant as you say. I'm not particularly satisfied with the readability of what we have (while the code is correct, to me at least it reads like it is wrong), but don't have a good solution.

PR #4217, opened 2019-09-06
Fix broken changelog

jeremyk-91:
**Goals (and why)**:
- Be able to release

- Backticks seem to have some semantic significance in YAML and thus need to be escaped.

- I stuck the changelog in a YAML parser.

**Concerns (what feedback would you like?)**: none

**Where should we start reviewing?**: +1-1


felixdesouza:
Yeah, I've generally found using yaml's multiline strings to be pretty useful i.e. 

```yaml
description: >
  `AsyncPuncher` ...
```

turns out there are [9 different ways](https://stackoverflow.com/a/21699210) of specifying multiline strings in yaml...

PR #4220, opened 2019-09-09
adding gzip compression support to the schema

mmigdiso:
**Goals (and why)**: Adding Gzip support to the schema. Currently atlasdb supports Lz4 algorithm and in some cases, gzip might be more useful as it is more than two times more space effective (specially in json streams)

**Implementation Description (bullets)**: Lz4 and Gzip compressions now derives from the same abstract class. 

**Testing (What was existing testing like?  What have you done to improve it?)**: The existing unit test class for lz4 has been generalized now to support Gzip compression 

**Priority (whenever / two weeks / yesterday)**: two weeks


palantirtech:
Thanks for your interest in [palantir/atlasdb](https://github.com/palantir/atlasdb), @mmigdiso! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -577,15 +581,27 @@ private void storeBlocksAndGetFinalMetadata() {
             private void loadStreamWithCompression() {
                 line("@Override");
                 line("public InputStream loadStream(Transaction t, final ", StreamId, " id) {"); {
-                    line("return new LZ4BlockInputStream(super.loadStream(t, id));");
+                    line("try {"); {
+                        line("return new " + clientSideCompression.inputClass + "(super.loadStream(t, id));");
+                    } line("} catch (IOException e) {"); {
```
schlosna: If an exception is thrown here, this will leak the stream returned by super.loadStream

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -577,15 +581,27 @@ private void storeBlocksAndGetFinalMetadata() {
...
             private void loadSingleStreamWithCompression() {
                 line("@Override");
                 line("public Optional<InputStream> loadSingleStream(Transaction t, final ", StreamId, " id) {"); {
                     line("Optional<InputStream> inputStream = super.loadSingleStream(t, id);");
-                    line("return inputStream.map(LZ4BlockInputStream::new);");
+
+                    line("return inputStream.map( s-> {"); {
+                        line("try {"); {
+                            line("return new " + clientSideCompression.inputClass + "(s);");
+                        } line("} catch (IOException e) {"); {
```
schlosna: If an exception is thrown here, this will leak the stream

atlasdb-commons/src/main/java/com/palantir/common/compression/AbstractCompressingInputStream.java
```diff
@@ -0,0 +1,140 @@
...
+import com.palantir.logsafe.Preconditions;
+
+/**
+ * {@link InputStream} that wraps a delegate InputStream, compressing its
+ * contents as they are read. Compression is managed via a compressing
+ * {@link OutputStream}. Reads pull data as necessary from the delegate
+ * InputStream, write the uncompressed data through the OutputStream, then
+ * serve the read from the resulting compressed buffer.
+ */
+public abstract class AbstractCompressingInputStream extends BufferedDelegateInputStream {
```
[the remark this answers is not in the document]
mmigdiso: Hello @j-baker , I've changed the implementation for the Gzip. And for your second comment on mark/reset, it is possible to do it for LZ4 as LZ4CompressingInputStream already using an internal buffer for keeping the compressed buffers and I can mark/reset on that buffer. 
But when it comes to Gzip implementation, DeflaterInputStream is not supporting the mark/reset. Thus if ones passes a readlimit to mark(readlimit) that is bigger than the gzip header, I'll have to rewind the deflaterinputstream as when reset is called. 

I see only one solution here: We can enforce all compression implementations to implement an interface with a getHeader() method.

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
```diff
@@ -0,0 +1,117 @@
...
+public class GzipCompressingInputStream extends SequenceInputStream {
+    public GzipCompressingInputStream(InputStream in) throws IOException {
+        this(in, 512);
+    }
+
+    public GzipCompressingInputStream(InputStream in, int bufferSize) throws IOException {
+        super(new GzipStreamEnumeration(in, bufferSize));
+    }
+
+    protected static class GzipStreamEnumeration implements Enumeration<InputStream> {
```
j-baker: Implementing `Enumeration` for 3 elements is the Java 1.0 equivalent of implementing an `Iterable` for the same thing.

What you do here is create a list, turn it into an iterator, and then implement an enumeration around the iterator. Instead, just do `Collections.enumeration(theListYouHad)` in order to convert the list into an enumeration.

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
```diff
@@ -0,0 +1,117 @@
...
+
+        public GzipStreamEnumeration(InputStream in) {
+            this(in, DEFAULT_DEFLATER_BUFFER_SIZE);
+        }
+
+        public GzipStreamEnumeration(InputStream in, int bufferSize) {
+            this.in = in;
+            this.bufferSize = bufferSize;
+            streamSupplierIt = Arrays.<Supplier<InputStream>>asList(() -> createHeaderStream(),
+                    () -> createContentStream(),
```
j-baker: this is a bit sketchy because the supplier implies that the content stream can be reused, but it cannot. Why do we have Suppliers here?

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
```diff
@@ -0,0 +1,117 @@
...
+        private final Iterator<Supplier<InputStream>> streamSupplierIt;
+        private CheckedInputStream checkedInputStream;
+        private DeflaterInputStream contentStream;
+        private CountingInputStream counting;
+
+        public GzipStreamEnumeration(InputStream in) {
+            this(in, DEFAULT_DEFLATER_BUFFER_SIZE);
+        }
+
+        public GzipStreamEnumeration(InputStream in, int bufferSize) {
```
j-baker: what is the purpose of allowing the consumer to configure the buffer size?

atlasdb-commons/src/test/java/com/palantir/common/compression/AbstractCompressionTests.java
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    public void testMultiBlock_readPastEnd() throws IOException {
+        byte[] uncompressedData = new byte[16 * BLOCK_SIZE];
+        fillWithCompressibleData(uncompressedData);
+        initializeStreams(uncompressedData);
+
+        byte[] decompressedData = new byte[17 * BLOCK_SIZE];
+        int bytesRead = ByteStreams.read(decompressingStream, decompressedData, 0, decompressedData.length);
+        assertEquals(uncompressedData.length, bytesRead);
+        assertArrayEquals(uncompressedData, Arrays.copyOf(decompressedData, bytesRead));
```
j-baker: instead of arrayEquals, assertArrayEquals and the other JUnit asserts, please can we use AssertJ's `assertThat` methods? Leads to better perf. `assertThat(decompressedData).startsWith(uncompressedData)` might well be possible.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -558,52 +561,86 @@ private void unmarkStreamsAsUsed() {
             }
 
             private void storeBlocksAndGetFinalMetadata() {
+                String compression = clientSideCompression.compressionType;
```
j-baker: I would propose here that you instead add a method to your compression enum which is something like `InputStream compress(InputStream input)` which does the wrapping for you. In this way you can do any extra exception handling that you need there.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -558,52 +561,86 @@ private void unmarkStreamsAsUsed() {
...
-                    line("        InputStream compressingStream = new LZ4CompressingInputStream(hashingStream)) {"); {
+                    line("        InputStream compressingStream = new " + compression + "(hashingStream)) {"); {
                         line("StreamMetadata metadata = storeBlocksAndGetHashlessMetadata(t, id, compressingStream);");
                         line("return StreamMetadata.newBuilder(metadata)");
                         line("        .setHash(ByteString.copyFrom(digest.digest()))");
                         line("        .build();");
-                    } line("} catch (IOException e) {"); {
-                        line("throw new RuntimeException(e);");
+                    } line("} catch (Exception e) {"); {
+                        line("throw Throwables.rewrapAndThrowUncheckedException(\"Error in creating compression stream\", e);");
```
j-baker: I don't think this change needed to be made - more directly, you shouldn't really be catching Exception.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -558,52 +561,86 @@ private void unmarkStreamsAsUsed() {
...
                     } line("}");
                 } line("}");
             }
 
+            private void closeStreamWithExceptionLogging(String streamVarName, String streamId) {
```
j-baker: likewise, please can you put this code in your enum, rather than code generating it? Anything that _can_ be not-code-generated, should be :)

j-baker: that would allow you to remove almost all of your changes to this class.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -849,6 +886,8 @@ private void cellsCleanedUp() {
         Generated.class,
         LZ4CompressingInputStream.class,
         LZ4BlockInputStream.class,
+        GzipCompressingInputStream.class,
+        GZIPInputStream.class,
```
j-baker: this.. probably does not contain enough of the inputs you need. I think you should be able to remove the imports for the LZ4 and GZIP streams, and instead just have a single import for your enum.

atlasdb-commons/src/main/java/com/palantir/common/compression/EnumClientCompressor.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.common.compression;
+
+import java.util.zip.GZIPInputStream;
+
+import net.jpountz.lz4.LZ4BlockInputStream;
+
+public enum EnumClientCompressor {
```
j-baker: you don't need to preface all enum types with `Enum` - recommend just calling this type `ClientCompressor`

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
mmigdiso: it is because the deflater allows it, and we allow the internal buffer configuration in LZ4 as well.

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
mmigdiso: We use suppliers here because the trailer needs to be created after the header and content deflater. Otherwise the trailer gets corrupted with wrong values. the nextElement() implementation gives a better view on why we use suppliers. I could implement a state machine as well but suppliers look cleaner to me in this case.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.zip.GZIPInputStream;
+
+import com.palantir.common.base.Throwables;
+
+import net.jpountz.lz4.LZ4BlockInputStream;
+
+public enum ClientCompressor {
+    GZIP(GzipCompressingInputStream.class, GZIPInputStream.class, GzipCompressingInputStream.GZIP_HEADER),
+    LZ4(LZ4CompressingInputStream.class, LZ4BlockInputStream.class,
+            new byte[] {'L', 'Z', '4', 'B', 'l', 'o', 'c', 'k'}),
```
j-baker: `"LZ4Block".getBytes(StandardCharsets.UTF_8)`

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+    NONE(null, null, new byte[] {});
+
+    public final Class<InputStream> compressorClass;
```
j-baker: instead of `Class` objects, have `UnaryOperator<InputStream>`; you're wrapping with an InputStream. Don't use reflection for this stuff.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+    ClientCompressor(Class compressorClass, Class decompressorClass, byte[] magic) {
+        this.compressorClass = compressorClass;
+        this.decompressorClass = decompressorClass;
+        this.magic = magic;
+    }
+
+    private static boolean matchMagic(ClientCompressor compressorType, byte[] buffer, int bufferLen) {
+        byte[] signature = compressorType.magic;
+        int i = 0;
+        while (i < signature.length && i < bufferLen && signature[i] == buffer[i++]);
```
j-baker: this should be an instance method on ClientCompressor, rather than a static method which takes one. As in, instead of writing matchMagic(compressorType, buffer, bufferLen), write compressorType.matchMagic(buffer, bufferLen)

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+        return i >= signature.length && signature.length > 0;
+    }
+
+    static InputStream getDecompressorStream(InputStream stream) throws IOException {
+        BufferedInputStream buff = new BufferedInputStream(stream);
+        int maxLen = Arrays.stream(ClientCompressor.values()).max(
+                Comparator.comparingInt(x -> x.magic.length)).get().magic.length;
+        buff.mark(maxLen);
+        byte[] headerBuffer = new byte[maxLen];
+        int len = buff.read(headerBuffer);
```
j-baker: this actually doesn't do the right thing, because buff.read can return any number of bytes, it doesn't have to actually fill the header buffer

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+        buff.reset();
+        Optional<ClientCompressor> compressor = Arrays.stream(ClientCompressor.values()).filter(t -> matchMagic(t, headerBuffer, len)).findFirst();
+        return compressor.map(t -> {
+            try {
+                return t.decompressorClass.getConstructor(InputStream.class).newInstance(buff);
+            } catch (Exception exc) {
+                throw new RuntimeException(exc);
+            }
+        }).orElseGet(() -> buff);
```
j-baker: this isn't right - if it doesn't match you shouldn't just return the compressed information - you should throw an exception.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+
+    static InputStream getDecompressorStream(InputStream stream) throws IOException {
+        BufferedInputStream buff = new BufferedInputStream(stream);
+        int maxLen = Arrays.stream(ClientCompressor.values()).max(
+                Comparator.comparingInt(x -> x.magic.length)).get().magic.length;
+        buff.mark(maxLen);
+        byte[] headerBuffer = new byte[maxLen];
+        int len = buff.read(headerBuffer);
+        buff.reset();
+        Optional<ClientCompressor> compressor = Arrays.stream(ClientCompressor.values()).filter(t -> matchMagic(t, headerBuffer, len)).findFirst();
```
j-baker: this is actually really quite broken, because it doesn't flag the case where _two_ compressors somehow share the same prefix. Instead, use .collect(MoreCollectors.onlyElement()), which will catch that bug.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.palantir.common.base.Throwables;
+
+import net.jpountz.lz4.LZ4BlockInputStream;
+
+public enum ClientCompressor {
+    GZIP(GzipCompressingInputStream.class, GZIPInputStream.class, GzipCompressingInputStream.GZIP_HEADER),
+    LZ4(LZ4CompressingInputStream.class, LZ4BlockInputStream.class,
+            new byte[] {'L', 'Z', '4', 'B', 'l', 'o', 'c', 'k'}),
+    NONE(null, null, new byte[] {});
```
j-baker: this code is quite scary, because the prefix you provide actually matches _all_ magic byte sequences

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
j-baker: I would be tempted to write a method like:

```java
private static boolean startsWith(InputStream stream, byte[] maybePrefix) throws IOException {
     stream.mark();
     try {
         for (int i = 0; i < maybePrefix.length; i++) {
             if (stream.read() != maybePrefix[i]) {
                 return false;
             }
         }
         return true;
     } finally {
          stream.reset();
     }
}
```

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
mmigdiso: this looks nicer but different than the original implementation. Unfortunately the access modifier does not allow to use it. 
https://github.com/lz4/lz4-java/blob/d43546e24388533eebd40fccb4be5468f0411788/src/java/net/jpountz/lz4/LZ4BlockOutputStream.java#L37

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
mmigdiso: But the code does not expect the headerBuffer to be filled. that is why we have the len parameter which is passed to matchMagic method later on. 
on the other hand, as we are abstracting inputstream, read() and read(byte[]) has different performances when it comes to fileinputstream(at least in the sense of # of system calls even if we consider file system cache)

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
mmigdiso: by returning the buff, It handles the case where the given stream is not a compression stream at all. So it is safe to pass a non compressed stream and read from it as it will just wrap it with buffered stream. We are not exposing the decompressor class to the user anyways.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
mmigdiso: Two things that avoids this situation. (if the comment above was for the NONE (byte[]{}) part. 
1- there is  ```magic.length > 0;``` in the return statement of the matchMagic method. 
2- values() method returns the enum's in the order they are defined. And the NONE is defined last.

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
j-baker: guess I'm saying you have a correctness bug right now. Also, you're only calling the single byte read method like 8 times, so the performance difference is zero.

atlasdb-commons/src/main/java/com/palantir/common/compression/CompressorForwardingInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+    public int read() throws IOException {
+        if (delegate == null) {
+            delegate = ClientCompressor.getDecompressorStream(compressedStream);
+        }
+        return delegate.read();
+    }
+
+    @Override
+    public void close() throws IOException {
+        super.close();
```
j-baker: there's no need to call super.close here

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
j-baker: You still don't need to implement enumeration for that, though. That said, using suppliers makes sense now.

atlasdb-commons/src/test/java/com/palantir/common/compression/NotCompressedStreamTests.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.util.NoSuchElementException;
+
+import org.junit.Test;
+
+public class NotCompressedStreamTests {
+
+    @Test(expected = NoSuchElementException.class)
```
j-baker: please don't ever use `@Test(expected` - it's super error prone (you don't get to specify _where_ in the test the exception should come out). Instead, please use `assertThatExceptionOfType(NoSucheElementException.class).isThrownBy(() -> decompressingStream.read())`

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,83 @@
...
+            throw new RuntimeException(exc);
+        }
+    }, GzipCompressingInputStream.GZIP_HEADER),
+    LZ4(LZ4CompressingInputStream.class, LZ4BlockInputStream::new,
+            new byte[] {'L', 'Z', '4', 'B', 'l', 'o', 'c', 'k'}),
+    NONE(null, UnaryOperator.identity(), new byte[] {});
+
+    public final Class<InputStream> compressorClass;
+    public UnaryOperator<InputStream> decompressorCreator;
+    public final byte[] magic;
```
j-baker: these should all be private - especially magic, since other classes should not be able to overwrite it.

atlasdb-commons/src/main/java/com/palantir/common/compression/CompressorForwardingInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.common.compression;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class CompressorForwardingInputStream extends InputStream {
```
j-baker: this isn't the way to implement InputStream - basically you only implemented single byte read and so performance will end up an issue. Make sure you implemented the `read(buf, off, len)` method as well.

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
j-baker: So explicitly, I'm proposing you write something like:

```java
List<Supplier<InputStream>> streams = ImmutableList.of(this::createHeaderStream, () -> in, () -> countingStream, () -> createTrailerStream(crc, counting));
Enumeration<InputStream> inputStream = Collections.enumeration(Collections2.transform(streams, Supplier::get)); 
```

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
```diff
@@ -0,0 +1,83 @@
...
+     * IllegalArgumentException} if more than one decompressor is detected,  a {@code NoSuchElementException} if no
+     * compressor detected.
+     */
+    static InputStream getDecompressorStream(InputStream stream) throws IOException {
+        BufferedInputStream buff = new BufferedInputStream(stream);
+        List<ClientCompressor> compressors = Arrays.stream(ClientCompressor.values()).sorted(
+                Comparator.comparingInt((ClientCompressor t) -> t.magic.length).reversed()
+        ).collect(
+                Collectors.toList());
+        int maxLen = compressors.get(0).magic.length;
```
j-baker: `int maxLen = Arrays.stream(ClientCompressor.values()).mapToInt(c -> c.magic.length).max()`

j-baker: can avoid all of the sorting and reversing stuff :)

mmigdiso: but i need to start from the longest prefix, otherwise a shorter magic char which is a substring of another magic char would cause a bug. the purpose of this code was not only to find the maxlen.

atlasdb-commons/src/main/java/com/palantir/common/compression/GzipCompressingInputStream.java
mmigdiso: But as the GzipCompressingInputStream is extending from the SequenceInputStream, the first call of the constructor needs to be a super(). So I cannot initialize the inputstream enum above in the  GzipCompressingInputStream constructor.

atlasdb-commons/src/main/java/com/palantir/common/compression/CompressorForwardingInputStream.java
mmigdiso: done

atlasdb-commons/src/main/java/com/palantir/common/compression/ClientCompressor.java
mmigdiso: If I understood you correctly,  this is more an optimization comment than the correctness. But implementing the method above will cause (number of decompressor) * (len of magic char) IO calls. I still believe that reading a block of bytes with a single I/O and running maybe 3 extra comparison makes sense.

tpetracca:
This was merged via #4311

PR #4221, opened 2019-09-09
Explicitly authorize KeyedStream.keys() in stream stores

jeremyk-91:
**Goals (and why)**:
- Large internal product has its own static analysis rules (different from baseline). As part of these, they have a check banning `Multimap.keys()` (recommending `keySet()` instead). The check isn't sophisticated enough to distinguish that from `KeyedStream.keys()`, which we use in the stream store renderer for the cleanup tasks now.
- See internal alternative VCS 328401.

- Add an explicit authorization to this call of the `keys()` method.

**Testing (What was existing testing like?  What have you done to improve it?)**: not much

**Concerns (what feedback would you like?)**: Did I miss any of the stream stores?

**Where should we start reviewing?**: StreamStoreRenderer.java


felixdesouza:
lol oh wow!

PR #4222, opened 2019-09-09
Publish correct scm information within POM

ferozco:
Update the scm information published in POMs so that it is consistent with all other published POMS.

### DIFF
--- a/gradle/publish-jars.gradle
+++ b/gradle/publish-jars.gradle
@@ -37,7 +37,7 @@ publishing {
             }
             pom.withXml {
                 def scm = asNode().appendNode('scm')
-                scm.appendNode('url', 'https://github.com/palantir/atlasdb')
+                scm.appendNode('url', 'git@github.com:palantir/atlasdb.git')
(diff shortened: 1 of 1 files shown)


ferozco:
I'm not sure how its used externally, but we've recently started using this within excavator to give you release notes in the upgrade PR. 

There isn't a plugin which does this for everyone else, but we do have a managed gradle file which we to have consistent publishing in all our oss repos: https://github.com/palantir/conjure-java/blob/develop/gradle/publish-jar.gradle

It should be possible to get you guys to use it too!

felixdesouza:
hmm, it's not building because this is from a fork, you should be able to just make a PR directly on atlas I believe.

I see! I think we have some custom stuff still, if it's possible to have it all work, I'd be up for integrating it!

ferozco:
unfortunately I don't have write access on this repo so I'm not able to push a branch. Happy to recreate if you can grant me perms

ferozco:
Closing in favour of #4224

PR #4223, opened 2019-09-09
Improve AbstractPoolingContainer log levels to reduce noise

carterkozak:
See:
https://github.com/palantir/gradle-baseline/tree/develop/docs/best-practices/java-coding-guidelines#log-levels

**Goals (and why)**:

Reduce noise in systems with info logging by default.


carterkozak:
That large internal product has an open ticket to reduce logging noise from this class ;-)

PR #4225, opened 2019-09-09
[Conjure Java Runtime] Part 6: Namespaced TimeLock RPC Client

jeremyk-91:
**Goals (and why)**:
- Eventually remove the need to tack on `/namespace` to the base URI list when creating proxies. This is useful for the migration to CJR; see discussion on #4205.

- Implement a namespaced version of `TimelockRpcClient`

- Not much; existing tests passing (especially integration tests) should be evidence that this works.
- Some changes to the behaviour of `parseInstallAndRuntimeConfigs` are tested explicitly.

- I opted to rename the RPC interface most Atlas code goes through to `NamespaceAwareTimelockRpcClient`, and keep `TimelockRpcClient` as the base interface. This is for (1) continuity of metrics, and (2) because I feel awkward having a raw RPC client not be the layer that actually talks to timelock (at least, I prefer this to having `NamespaceAgnosticTimelockRpcClient`). Do you agree with having it this way?

**Where should we start reviewing?**: `NamespaceAwareTimelockRpcClient`


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -219,24 +220,24 @@ TimelockService timelockServiceForClient(String name) {
...
 
-    <T> CompletableFuture<T> runWithRpcClientAsync(Function<TimelockRpcClient, T> function) {
-        return CompletableFuture.supplyAsync(() -> function.apply(timelockRpcClient()));
+    <T> CompletableFuture<T> runWithRpcClientAsync(Function<NamespaceAwareTimelockRpcClient, T> function) {
+        return CompletableFuture.supplyAsync(() -> function.apply(namespaceAwareClient()));
     }
 
-    TimelockRpcClient timelockRpcClient() {
-        return timelockRpcClient(client);
+    NamespaceAwareTimelockRpcClient namespaceAwareClient() {
```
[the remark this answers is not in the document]
felixdesouza: fair enough

lock-api/src/main/java/com/palantir/lock/v2/NamespacedTimelockRpcClient.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Set;
+
+import com.palantir.lock.client.IdentifiedLockRequest;
+import com.palantir.timestamp.TimestampRange;
+
+/**
+ * Adapter that mirrors {@code TimelockRpcClient}, but for convenience automatically provides the namespace as a
+ * parameter.
+ */
+
```
felixdesouza: nit: extra space.

felixdesouza: I'm very conflicted, the naming scheme in part 8 seems/feels more intuitive if inverted here. It feels natural to be like "I'm going to pass around a `TimelockRpcClient`" and the actual thing that has the namespace is `NamespaceAwareTimelockRpcClient`. Why I think the other one works is because we never actually use the intermediate types/adapters bar the top level `RemoteLockServiceAdaptor` one which implements `LockService`. But here, we specifically *do* want to use the middle layer i.e. this class. Then when using it, it's like "what's the namespace? I haven't had to pass one in when using any of the methods on it."

So in a sense, it makes a bit of sense for the base layer to be `NamespaceAwareTimelockRpcClient` and for this class to be `TimelockRpcClient`. Happy for either to go in, just feeling somewhat uneasy.

jeremyk-91: discussed offline: this is named `Namespaced` because the namespace is already provided (references to `NamespaceAware` were accidental left-overs, which I will clean up) - as discussed will go through with that. I think it would be nice to keep the thing that actually makes the RPC calls to be `TimelockRpcClient`.

PR #4232, opened 2019-09-11
Lock refresh

jkozlowski:
**Goals (and why)**:

Seeing firstFailures logged as:

```
firstFailures
"[com.palantir.lock.client.LockTokenShare@7c45d29a]"
```

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
@@ -68,7 +68,8 @@ public class LockRefresher implements AutoCloseable {
             Set<LockToken> refreshFailures = Sets.difference(toRefresh, successfullyRefreshedTokens);
             tokensToRefresh.removeAll(refreshFailures);
             if (!refreshFailures.isEmpty()) {
-                log.info("Failed to refresh {} lock tokens, most likely because they were lost on the server."
+                log.info("Successfully refreshed {}, but failed to refresh {} lock tokens, "
+                                + "most likely because they were lost on the server."
                                 + " The first (up to) 20 of these were {}.",
                         SafeArg.of("successfullyRefreshed", successfullyRefreshedTokens.size()),
                         SafeArg.of("numLockTokens", refreshFailures.size()),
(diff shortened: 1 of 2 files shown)


lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
```diff
@@ -70,8 +70,10 @@ private void refreshLocks() {
             if (!refreshFailures.isEmpty()) {
```
jkozlowski: On a related note, does this logging ever get racey? TimelockClient#unlock calls this first and then unlock on timelock, so if the refresher is currently refreshing, it won't see the fact that we just unlocked it.

jkozlowski: Basically this logline suggests either a leader election or something going weird with timelock, whereas it seems to me that it could fail to refresh if the lock was unlocked in between.

jkozlowski: I guess it might depend on how timelock unlocks this stuff

jeremyk-91: Yep, it does. I had a brief look at the log spam last week, though didn't really get further than "async unlocker and refresher probably have a race condition"

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
```diff
@@ -70,8 +70,10 @@ private void refreshLocks() {
...
                 log.info("Failed to refresh {} lock tokens, most likely because they were lost on the server."
                                 + " The first (up to) 20 of these were {}.",
+                        SafeArg.of("successfullyRefreshed", successfullyRefreshedTokens.size()),
```
jeremyk-91: LargeInternalProduct will probably be sad on its internal checkstyle, since the number of `{}` won't line up - please add one to the beginning of the message i.e. `Successfully refreshed {} lock tokens, but failed to refresh {} ...

jkozlowski: Fixed

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
jkozlowski: cool, so it does sound like this logline might be incorrect and spammy. I would vote to remove it, instead the more useful logline would probably be something along the lines of "Change of leader detected", because that's roughtly what you'd like to track.

jkozlowski: We will also hopefully now see if these are normal transaction locks that are being lost or immutable timestamp locks.

PR #4219, opened 2019-09-06
Remove some AtlasDB allocations

j-baker:
Looking at a profile for an internal service, I noticed that there were
many allocations near the top of the profile.

Over 5 minutes, AtlasDB methods allocated about 25GB of StringBuilders,
for example :) - inside the metrics thing.

PR #4234, opened 2019-09-11
Tagged Metrics Event Invocation Handler Memory Optimisations

jeremyk-91:
**Goals (and why)**:
- See #4219, specifically @carterkozak's comment there

- Have a loading cache mapping Methods to Timers, rather than reconstructing the MetricName object every time.
- Extract a single global failure meter.

- Existing tests not breaking should suffice.

- There is an API break, though all functions passed in never actually used the context.

**Where should we start reviewing?**: TaggedMetricsEventInvocationHandler.java

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
@@ -17,8 +17,9 @@ package com.palantir.atlasdb.util;
 
 import java.lang.reflect.Method;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -26,7 +27,9 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.tritium.event.AbstractInvocationEventHandler;
@@ -37,7 +40,7 @@ import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 
 /**
  * A simplified and yet extended version of Tritium's MetricsInvocationEventHandler. This class supports augmenting
- * metric data with tags, possibly as a function of the method's invocation context.
+ * metric data with a constant set of tags.
  *
  * Note that this class does not yet offer support for processing of specific metric groups, unlike
  * MetricsInvocationEventHandler.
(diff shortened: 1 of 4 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -47,17 +52,36 @@
...
+                    @Nonnull
+                    @Override
+                    public Timer load(@Nonnull Method key) {
+                        MetricName finalMetricName = MetricName.builder()
+                                .safeName(MetricRegistry.name(serviceName, key.getName()))
+                                .putAllSafeTags(tagFunction.apply(key))
+                                .build();
+                        return taggedMetricRegistry.timer(finalMetricName);
+                    }
+                });
```
carterkozak: I don't think it makes sense to bound this because every MetricName will also be held in the TaggedMetricRegistry map. The methods are bounded by the proxy implemented interfaces.
Based on published benchmarks, caffeine has overhead beyond the concurrent hash map: https://github.com/ben-manes/caffeine/wiki/Benchmarks#read-100-1

jeremyk-91: In general this approach doesn't work, because we have metrics functions that are not pure functions :(

carterkozak: ack 👍

jeremyk-91: Yeah, the argument for not bounding makes sense - will switch to CHM if we continue pursuing this cache approach

carterkozak: actually, I'm not sure I follow, can you provide an example of a case that doesn't work to help me understand?

jeremyk-91: So currently say for `TimestampService`, we have a tag `isCurrentSuspectedLeader` that indicates whether the node thinks it is the timelock leader or not. If I was initially not the leader, the metric I build will say in the name that I am not the leader - but if I become leader after that, even though it's the same method I need to generate one that says I am now the leader, and report that.

carterkozak: I see, in that case the current cache won't quite work either because nothing triggers eviction.

jeremyk-91: Right - yeah that's why I applied the do not merge label. The cache keys would need to have both the `Method` and `Map<String, String>` for the tags. We still pay an allocation to create this object, but I anticipate it will be quite a bit less than the `StringBuilder` issues we would have previously with the `MetricName` objects.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -73,16 +87,21 @@ public void onSuccess(@Nullable InvocationContext context, @Nullable Object resu
...
+
+    private Timer registerTimer(TimerCacheKey timerCacheKey) {
         MetricName finalMetricName = MetricName.builder()
-                .safeName(MetricRegistry.name(serviceName, context.getMethod().getName()))
-                .putAllSafeTags(tagFunction.apply(context))
+                .safeName(MetricRegistry.name(serviceName, timerCacheKey.method().getName()))
+                .putAllSafeTags(timerCacheKey.tags())
                 .build();
-        taggedMetricRegistry.timer(finalMetricName).update(nanos, TimeUnit.NANOSECONDS);
+        return taggedMetricRegistry.timer(finalMetricName);
```
carterkozak: This is a lot of complexity, how confident are we that it performs better than the string concatenation approach?

What if we replace
```java
.safeName(MetricRegistry.name(serviceName, context.getMethod().getName()))
```
with
```java
.safeName(serviceName + '.' + context.getMethod().getName())
```

I imagine that will perform better, especially on java 9+ with [JEP 280](https://openjdk.java.net/jeps/280).

gmaretic:
What's the state of this work?

PR #4235, opened 2019-09-12
upgrade libraries

clockfort:
I know, this is going to obviously stink for a ton of downstream projects, but the stink I believe is wanted.

These are library bumps past versions of external libraries that have known vulnerabilities.

Some of the vulns aren't applicable to how we're using the libraries, but we should still upgrade them because customer security scans will still ding us, etc.


clockfort:
gonna work on this off-ticket

Issue #4233, opened 2019-09-11
Targeted Sweep can get into an infinite retry loop if a table is dropped improperly


jeremyk-91:
Internal reference 137731

Let `T` be a table that is written to on a deployment where targeted sweep is enabled. Now, suppose `T` is dropped. Dropping a table in most implementations non-atomically drops the table from the underlying KVS and then deletes the relevant cell in the metadata table. So suppose the table `T` is deleted, but its metadata cell remains (because the service went down, or something like that). When processing the targeted sweep queue, attempting to delete from that table will usually result in some kind of exception as the table doesn't exist - but the table does exist as far as targeted sweep is concerned (see `SweepQueueDeleter.java:77`). We thus have an infinite retry loop.

clockfort:
Found another infinite retry loop, again specific to Oracle with dropped tables.

Specifically, targeted sweep retries infinitely to delete from the (possibly existing for this cell) section of an overflow table. However, the corresponding table (and overflow table) have been previously dropped.

Image:
![image](https://user-images.githubusercontent.com/57031/71853584-bba81480-3090-11ea-87e1-5f67dc7d9006.png)

PR #4236, opened 2019-09-12
fix aborted transaction deletes retrying forever on dead tables

clockfort:
This is causing high logspam, because this condition never gets cleared and it keeps trying (and logging) over and over.

related to #4233


clockfort:
Oh hey, I'm back from vacation. Added the test.

I had to move the exception's package because unlike every other KVS, the in-memory KVS is actually part of atlasdb-client (gross), and also previously didn't throw quite the right kind of exceptions that the standard chain of KVSs would throw.

clockfort:
@gmaretic thanks for changes.
Yeah, I only looked at how the exceptions were thrown from oracle and inmemory.

your much cleaner way of checking for table deletion probably has some edge-cases-of-this-edge-case where its wrong (e.g. table drop and putMeta(EmptyMetadata) aren't atomic), but I'm going to merge in anyways, this should solve the tens of thousands of errors per day on our internal testing stack.

PR #4238, opened 2019-09-12
[Timelock Partioning] Part 28: Move timestamp paxos to runtime config.

felixdesouza:

**Goals (and why)**:
Being able to turn this off via config instead of having to restart an entire cluster seems to be a better fit here.

Abuse usage of `DynamicDecoratingProxy`, seeing as the batch is not a decoration of the single, just a means of switching between the two implementations.

Existing test suite that tests both batch and non batch should cover this.

* `ClientPaxosResourceFactory`

Relatively soon please.


felixdesouza:
Yeah, it's probably due a rename, but should be done separately.

PR #4240, opened 2019-09-12
[Conjure Java Runtime] Part 8: Namespaced Lock and Nullable/Optional Bridging

jeremyk-91:
**Goals (and why)**:
- Namespace the lock service, to unblock #4205 
- Also switch Nullables to Optionals, since remoting framework does not support them.

- `BareLockRpcClient`, a version of the LockService that uses `Optional`s, but does not know about namespaces. This is intended to be used by embedded timelock users, and also is used internally (not for remoting).
- `LockRpcClient`, a version of the LockService that uses `Optional`s and namespace arguments.
- `NamespacedLockRpcClient`, a `BareLockRpcClient` implemented using a `LockRpcClient` and namespaces. See also `NamespacedTimelockRpcClient` from #4225 
- `RemoteLockServiceAdapter`, a LockService implementation that delegates to a `BareLockRpcClient` (that may be created from a `LockRpcClient`

Updated integration tests to use the new `LockRpcClient` and surrounding interfaces.

- Is there a more elegant way to sort out the interfaces, while still supporting leader blocks?

**Where should we start reviewing?**: LockRpcClient.java

**Priority (whenever / two weeks / yesterday)**: 2 weeks

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
+++ b/lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
@@ -33,7 +33,7 @@ import com.palantir.logsafe.Safe;
 /**
  * See also {@link LockRpcClient}. This interface exists to support legacy users not using an external TimeLock
  * server (as in these cases, lock services will not be namespaced) but also using remoting frameworks where
- * null values are not permitted.
+ * endpoints producing null values are not permitted.
  */
 @Path("/lock")
 @Produces(MediaType.APPLICATION_JSON)
--- a/lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
+++ b/lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
@@ -1,0 +1,129 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock;
+
+import java.math.BigInteger;
+import java.util.Optional;
+import java.util.Set;
+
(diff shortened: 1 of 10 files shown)


lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * See also {@link LockRpcClient}. This interface exists to support legacy users not using an external TimeLock
+ * server (as in these cases, lock services will not be namespaced) but also using remoting frameworks where
+ * endpoints producing null values are not permitted.
+ */
+@Path("/lock")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface BareLockRpcClient {
```
felixdesouza: from previous discussion, perhaps `NamespaceAgnosticLockRpcClient`, which captures the intent of leader lock rpc client not needing a namespace and the timelock namespace agnostic adaptor.

lock-api/src/main/java/com/palantir/lock/client/NamespacedLockRpcClient.java
```diff
@@ -0,0 +1,149 @@
...
+import com.palantir.lock.HeldLocksToken;
+import com.palantir.lock.LockClient;
+import com.palantir.lock.LockRefreshToken;
+import com.palantir.lock.LockRequest;
+import com.palantir.lock.LockResponse;
+import com.palantir.lock.LockRpcClient;
+import com.palantir.lock.LockServerOptions;
+import com.palantir.lock.SimpleHeldLocksToken;
+
+public class NamespacedLockRpcClient implements BareLockRpcClient {
```
felixdesouza: Perhaps something like `NamespaceAgnosticLockClientAdaptor` which implements `NamespaceAgnosticLockRpcClient`?

lock-api/src/main/java/com/palantir/lock/BareLockRpcClient.java
jeremyk-91: Yep, will change to that

lock-api/src/main/java/com/palantir/lock/client/RemoteLockServiceAdapter.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public long currentTimeMillis() {
+        return bareLockRpcClient.currentTimeMillis();
+    }
+
+    @Override
+    public void logCurrentState() {
+        bareLockRpcClient.currentTimeMillis();
```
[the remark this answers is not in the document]
jeremyk-91: Yeah, it should. Will fix

PR #4241, opened 2019-09-13
Revert "A quick hack to see the viability of table level conflict han…

jeremyk-91:
…dling (#4228)"

This reverts commit 5013cc7f5ea15e7d3e0c4fccfb4aff2947273c04.

Please do not merge without James's approval.

@j-baker This commit was in the LargeInternalProduct autobump where the regression was detected and the AtlasInterfaceService.

```
jkong$ git log --pretty=format:"%h%x09%x09%ad%x09%s" --date=short ba884b7..0.154.2
dc86078         2019-09-11      Autorelease 0.154.2
288e999         2019-09-11      Remove some AtlasDB allocations (#4219)
169b0c1         2019-09-11      Autorelease 0.154.1
5013cc7         2019-09-11      A quick hack to see the viability of table level conflict handling (#4228)
07597f7         2019-09-11      PTExecutors provides reasonable operation names (#4231)
bf34425         2019-09-11      Fix LockRefresher log message (#4232)
509de29         2019-09-10      Improve AbstractPoolingContainer log levels to reduce noise (#4223)
12ca8b4         2019-09-10      Excavator: Migrate annotation processors to native gradle imple… (#4195)
47d26af         2019-09-09      Update publish-jars.gradle (#4224)
```

I'm not entirely sure that this is the root cause, but think we should see if this resolves the perf issues. It could be #4219 as well, or one of the others if they went grossly wrong, but to me this seems like the highest risk. I'm happy to unrevert this quickly if this does not show an improvement.


j-baker:
Yeah so this is definitely it. My error is that after e.g. 1000 writes to a table ahead of the immutable timestamp, checking conflicts is O(1000). As in, by being more coarse grained, this causes issues.

j-baker:
For the future, a fix here is that we only need to keep the _first_ write with given properties set after any individual transaction, which should keep it at a lot less state. but right now we do a lot of unnecessary work because of this. Apologies.

That said, we got the data we wanted!

jkozlowski:
@j-baker yeah that was my comment on that PR this morning, it's not so much the synchronized but the fact it goes through the whole state, so it holds the lock a bunch. We could maybe also batch the processing at commit time, so it's not every transaction that needs to go through the whole state, but that said I didn't read the whole PR to see if that's possible.

PR #4245, opened 2019-09-17
Apply PreferAssertJ suggested fixes

carterkozak:
Applied suggested fixes, then made a few manual fixes where generics went wonky.

### DIFF
--- a/atlasdb-api/src/test/java/com/palantir/atlasdb/keyvalue/api/SweepResultsTest.java
+++ b/atlasdb-api/src/test/java/com/palantir/atlasdb/keyvalue/api/SweepResultsTest.java
@@ -15,10 +15,10 @@
  */
 package com.palantir.atlasdb.keyvalue.api;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.util.Optional;
 
-import org.hamcrest.Matchers;
-import org.junit.Assert;
 import org.junit.Test;
 
 import com.palantir.atlasdb.encoding.PtBytes;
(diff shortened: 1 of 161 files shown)


atlasdb-api/src/test/java/com/palantir/atlasdb/keyvalue/api/SweepResultsTest.java
```diff
@@ -64,52 +66,49 @@
 
     @Test
     public void accumulateWithNextRowInOrder() {
-        Assert.assertThat(RESULTS.accumulateWith(OTHER_RESULTS), Matchers.equalTo(COMBINED_RESULTS_AND_OTHER));
+        assertThat(RESULTS.accumulateWith(OTHER_RESULTS)).is(new HamcrestCondition<>(Matchers.equalTo(COMBINED_RESULTS_AND_OTHER)));
```
schlosna: Should we update the refactor tolling to transform `Matchers.equalTo` to `isEqualTo`?

carterkozak: Yep, that's reasonable. I don't think we'll be able to support migration of all hamcrest matchers, but the most common ones should be doable.

carterkozak: Added support for migrating common hamcrest matchers here: https://github.com/palantir/gradle-baseline/pull/850

jeremyk-91:
@carterkozak Would it be plausible to separate the uncontentious changes from the ones along the lines of the Hamcrest matchers which are arguably negative? Thanks!

PR #4252, opened 2019-09-23
[Timelock Partitioning] Part 30: Ensure everything is instrumented

felixdesouza:
**Goals (and why)**:
During testing, several metrics were missing. 

* Use tagged metrics everywhere, happy to own changing the boards and what not. Should be easier to navigate imo.
  * Toying with the idea of hierarchical metric registries. In theory, we could use this to get what we want regarding sliding window metrics everywhere (although we would still have to deal with the non tagged metrics :( )
* Pulling out some helper classes, I really wish I was allowed to use dagger here. These will help with further wiring changes coming down the line. Some of them may disappear after everything is done, but we'll see.

This is mainly a refactor + adding metrics, so nothing changed on that front.

* I would like to at some point collapse `timelock-agent` and `timelock-impl`.

`TimelockMetrics` perhaps? and `ClientPaxosResourceFactory` to start. So many hoops to jump through when there are circular dependency conflicts.

Relatively soon.

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockProxyFactories.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockProxyFactories.java
@@ -23,7 +23,7 @@ import java.util.stream.Collectors;
 
 import org.immutables.value.Value;
 
-import com.palantir.atlasdb.http.AtlasDbFeignTargetFactory;
+import com.palantir.atlasdb.http.AtlasDbHttpClients;
 import com.palantir.conjure.java.config.ssl.SslSocketFactories;
 import com.palantir.conjure.java.config.ssl.TrustContext;
 import com.palantir.timelock.config.TimeLockInstallConfiguration;
@@ -33,7 +33,7 @@ import com.palantir.timelock.paxos.PaxosRemotingUtils;
 public abstract class TimelockProxyFactories {
 
     abstract TimeLockInstallConfiguration install();
-    abstract TimelockMetrics metrics();
+    abstract TimelockPaxosMetrics metrics();
 
     <T> List<T> createInstrumentedRemoteProxies(Class<T> clazz, String name) {
         Set<String> remoteUris = PaxosRemotingUtils.getRemoteServerPaths(install());
@@ -41,7 +41,7 @@ public abstract class TimelockProxyFactories {
                 .getSslConfigurationOptional(install())
                 .map(SslSocketFactories::createTrustContext);
         return remoteUris.stream()
-                .map(uri -> AtlasDbFeignTargetFactory.DEFAULT.createProxy(
+                .map(uri -> AtlasDbHttpClients.DEFAULT_TARGET_FACTORY.createProxy(
                         trustContext,
                         uri,
                         clazz,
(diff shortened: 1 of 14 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockMetrics.java
```diff
@@ -0,0 +1,59 @@
...
+        return new SlidingWindowTaggedMetricRegistry(35, TimeUnit.SECONDS);
+    }
+
+    public static TimelockMetrics of(PaxosUseCase paxosUseCase, TaggedMetricRegistry parentRegistry) {
+        TimelockMetrics metrics = ImmutableTimelockMetrics.builder().paxosUseCase(paxosUseCase).build();
+        metrics.attachToParentMetricRegistry(parentRegistry);
+        return metrics;
+    }
+
+    private void attachToParentMetricRegistry(TaggedMetricRegistry parent) {
```
felixdesouza: this is the hierarchical stuff I was talking about

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockMetrics.java
```diff
@@ -0,0 +1,59 @@
...
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+@Value.Immutable
+public abstract class TimelockMetrics {
+
+    abstract PaxosUseCase paxosUseCase();
+
+    @Value.Derived
+    TaggedMetricRegistry metrics() {
+        return new SlidingWindowTaggedMetricRegistry(35, TimeUnit.SECONDS);
```
felixdesouza: a bit bummed I have to do this here, but I do need a fresh tagged metric registry instead of the singleton thing.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockMetrics.java
jeremyk-91: I had a look at the interface/implementation Tritium provides, this looks fine. Only concern would be possible perf/allocation consequences if there are large numbers of metrics (now getMetrics needs to create new MetricName objects on every tagged thing), though we should be able to catch that and revert it if necessary.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockProxyFactories.java
```diff
@@ -35,26 +33,37 @@
...
+    <T> List<T> createInstrumentedRemoteProxies(Class<T> clazz, String name) {
         Set<String> remoteUris = PaxosRemotingUtils.getRemoteServerPaths(install());
-        Optional<TrustContext> trustContext = PaxosRemotingUtils.getSslConfigurationOptional(install())
+        Optional<TrustContext> trustContext = PaxosRemotingUtils
+                .getSslConfigurationOptional(install())
                 .map(SslSocketFactories::createTrustContext);
         return remoteUris.stream()
-                .map(uri -> AtlasDbHttpClients.createProxy(
-                        metrics(),
+                .map(uri -> AtlasDbFeignTargetFactory.DEFAULT.createProxy(
```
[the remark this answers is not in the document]
felixdesouza: it was somewhat all over the place, if we can get the new metrics to use the `TaggedMetricRegistry`, would be pretty neat, but easier to start in timelock I suppose.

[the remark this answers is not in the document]
felixdesouza: I'll make `AtlasDbHttpClients#DEFAULT_TARGET_FACTORY` public then for the mean time.

PR #4255, opened 2019-09-23
dropwizard 1.2.x dw metrics 3.x

### LINKED ISSUE #4254: Downgrade dropwizard-core to 1.2.9
dropwizard 1.3.x (what we currently use) has a dependency on dropwizard-metrics 4.x. So much of our downstream consumers are hadoop/spark clients that cannot take the jump from dropwizard-metrics 3.x to 4.x. Would be just a bit more convenient if we used dropwizard 1.2.x (latest is 1.2.9) here as that depends on dropwizard-metrics 3.x.

tpetracca:
Hopefully resolves #4254

### DIFF
--- a/changelog/@unreleased/pr-4255.v2.yml
+++ b/changelog/@unreleased/pr-4255.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: Lower test projects' dropwizard version to ``1.2.9`` and dropwizard.metrics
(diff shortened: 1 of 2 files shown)


jeremyk-91:
This one shouldn't affect production paths (no locks change, even) so I think should be fine

felixdesouza:
Ps atlas doesn't have any lock files :P

PR #4260, opened 2019-09-24
[Conjure Java Runtime] Part 9: Version Selecting Clients

jeremyk-91:
**Goals (and why)**:
- Prepare for a smoother rollout of CJR clients, by implementing a version-selecting layer.
- Allow easy comparison of performance metrics for users on legacy and new code paths (along with aggregate metrics to assess the overall health of a client service).

- Move the existing `DynamicDecoratingProxy` to `atlasdb-commons` and rename it `PredicateSwitchingProxy`. This fits better with how we are using it going forward (it's used in batched Paxos, and now in remoting layer selection).
- Implement `VersionSelectingClients`, which handles instrumentation of clients with a tag indicating whether the client is a new or legacy client, along with deciding which way to go. This is a factory for `PredicateSwitchingProxy`.

- Added a simple test for `VersionSelectingClients`.
- I did change a method in `AtlasDbHttpClients` to route through this; for that, existing integration tests should catch any gross errors.

- There is going to be some duplication in `AtlasDbHttpClients` creation, for engaging the old and new `TargetFactory` implementations. Since there are only 2 I'm not sure eliminating the duplication would make things very much more readable.
- If people wrote stateful proxies and shoved them into `VersionSelectingClients` there might be unexpected consequences. I think keeping this in the `http` package plus a comment there should be enough of a warning, do I need to be louder about this?

**Where should we start reviewing?**: Bottom up. Start with PredicateSwitchedProxy then probably go to VersionSelectingClients.

**Priority (whenever / two weeks / yesterday)**: this week, please


atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
```diff
@@ -0,0 +1,74 @@
...
+        return PredicateSwitchedProxy.newProxyInstance(
+                instrumentedNewClient,
+                instrumentedLegacyClient,
+                () -> ThreadLocalRandom.current().nextDouble() < newClientProbabilitySupplier.getAsDouble(),
+                clazz);
+    }
+
+    private static <T> T instrumentWithClientVersionTag(
+            TaggedMetricRegistry taggedMetricRegistry,
+            T uninstrumentedLegacyClient,
```
felixdesouza: nit: `uninstrumentedLegacyClient` -> `client` or `instance`?

atlasdb-config/src/test/java/com/palantir/atlasdb/http/VersionSelectingClientsTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private final TimestampService oldService = mock(TimestampService.class);
+    private final TimestampService selectingService = VersionSelectingClients.createVersionSelectingClient(
+            SharedTaggedMetricRegistries.getSingleton(),
+            newService,
+            oldService,
+            probability::get,
+            TimestampService.class);
+
+    @Test
+    public void probabilityZeroAlwaysGoesToOldService() {
```
felixdesouza: nit: set the probability at the start of the test so we don't have to assume default/look up at the initialised version

atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Factory for randomly selecting from a new and a legacy version of a client, based on a live-reloading probability.
+ *
+ * Please note that your clients will run independently; if there are stateful invariants that need to be enforced
+ * across individual clients, you may need to share state appropriately.
+ */
+final class VersionSelectingClients {
+    private static final String CLIENT_VERSION = "clientVersion";
+    private static final Map<String, String> NEW_CLIENT_TAG = ImmutableMap.of(CLIENT_VERSION, "new");
```
felixdesouza: offline discussion about using versions here, probably also want to distinguish them from server versions maybe? unclear

atlasdb-commons/src/test/java/com/palantir/common/proxy/PredicateSwitchedProxyTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private final AtomicBoolean atomicBoolean = new AtomicBoolean(false);
+
+    private final TestInterface testInterface = PredicateSwitchedProxy.newProxyInstance(
+            decorated,
+            delegate,
+            atomicBoolean::get,
+            TestInterface.class);
+
+    @Test
+    public void dynamicallySwitchesCorrectlyForNoArgMethods() throws Exception {
```
felixdesouza: can this be nuked? in favour of the test below

atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
OStevan: I'm not familiar, but doesn't this naming scheme introduce problems if there are more than two versions?

atlasdb-commons/src/test/java/com/palantir/common/proxy/PredicateSwitchedProxyTest.java
jeremyk-91: yeah, this is a leftover from when this was used to test TimestampService I think

atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
jeremyk-91: Correct, this naming scheme would not work if you have >2 versions. I think the idea was that _clients_ would rarely have to work with >2 versions at a time, though our discussion was that it might still be useful to support more precise names, since if you had a service using `V1` or `V2`, and then upgraded to one with `V2` or `V3` (having the old one is useful in case you need to quickly stop the new one from running), then `V2` is new and then suddenly becomes old, and `AtlasVersion` isn't a tag that our metrics normally have so it's not very easy to unpick that immediately.

atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
jeremyk-91: I guess we still need to differentiate the `new` and the `legacy` ones? I will remove `uninstrumented` (as you could in theory have your own instrumentation below this layer if you wanted it for some reason)

PR #4262, opened 2019-09-25
[Timelock Partitioning] Part 31: Refactor to make leadership deps easy

felixdesouza:
**Goals (and why)**:
The factory that creates resources and clients and what not was tied to the the timestamp paxos and made a ton of assumptions. We "generalise" this a bit more, and it's a bit more modular.

Names are a bit awful, and happy to take suggestions. Sadly most of the wiring/logic would go away if I could use dagger 😢, well not go away completely, but it'd be a lot more declarative and easier to grok imo.

* `PaxosComponents` can give you a `Paxos(Acceptor|Learner)` provided you give it a client. Now it can also give you a `BatchPaxos(Acceptor|Learner)`. This simplifies a few places where we're extracting an acceptor/learner out of a resource which is all sorts of roundabout.
* `TimelockProxyFactories` no longer exists, it did too much.
* Introduce some context object that's common to both use cases i.e. has a network client, and has a proposer, the local components 
* `PaxosTimestampCreator` takes a context object instead
* `TimelockPaxosMetrics`
* `PaxosInstallationContext` common things we want from `TimelockInstallConfiguration` but we have to pass a load of statics around, super annoying otherwise.

Existing integration tests should catch any bad behaviour.

* Names are all over the place, have factories of factories and what not
* Am I doing repeat work where I don't need to where that will consumer resources? e.g. create new remoting clients?

`ClientPaxosResourceFactory`.

ASAP


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -117,19 +124,154 @@ private static NetworkClientFactories factories(
...
+        @Value.Derived
+        public List<BatchPaxosLearnerRpcClient> batchLearner() {
+            return createInstrumentedRemoteProxies(
+                    BatchPaxosLearnerRpcClient.class,
+                    "batch-paxos-learner-rpc-client");
+        }
+
+        private <T> List<T> createInstrumentedRemoteProxies(Class<T> clazz, String name) {
+            return context().remoteUris().stream()
+                    .map(uri -> AtlasDbHttpClients.DEFAULT_TARGET_FACTORY.createProxy(
```
jeremyk-91: could we leave a `TODO` here to wire up the configurable cutover to CJR?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -117,19 +124,154 @@ private static NetworkClientFactories factories(
...
+                            uri,
+                            clazz,
+                            name,
+                            false))
+                    .map(proxy -> AtlasDbMetrics.instrumentWithTaggedMetrics(
+                            metrics(),
+                            clazz,
+                            proxy,
+                            name,
+                            _unused -> ImmutableMap.of()))
```
jeremyk-91: I think normally if we don't provide tags we just use `AtlasDbMetrics.instrument(metrics(), clazz, proxy, name)` (unless there's some reason we need this to be the tagged metrics registry? I've somewhat lost track of how the instrumentation goes, honestly)

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -117,19 +124,154 @@ private static NetworkClientFactories factories(
...
+                    BatchPaxosAcceptorRpcClient.class,
+                    "batch-paxos-acceptor-rpc-client");
+        }
+
+        @Value.Derived
+        public List<BatchPaxosLearnerRpcClient> batchLearner() {
+            return createInstrumentedRemoteProxies(
+                    BatchPaxosLearnerRpcClient.class,
+                    "batch-paxos-learner-rpc-client");
+        }
```
jeremyk-91: Following your concern on avoiding unnecessary creation: we _technically_ could have these 4 be `@Value.Lazy` and only initialize the proxies on first usage, the idea being that if `useBatchPaxos` is always true then you don't care about the unbatched ones. 

However, we already create all of them (line 113 onward). It's possible to change DynamicDecoratingProxy to lazily initialize stuff, but it seems like a large change for what I suspect is questionable gain.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -65,7 +65,7 @@
 
     @ClassRule
     public static ParameterInjector<TestableTimelockCluster> injector =
-            ParameterInjector.withFallBackConfiguration(() -> PaxosSuite.NON_BATCHED_PAXOS);
+            ParameterInjector.withFallBackConfiguration(() -> PaxosSuite.BATCHED_PAXOS);
```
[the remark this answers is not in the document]
felixdesouza: oop, partial testing, shouldn't really matter either way unless you have strong preferences?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -117,19 +124,154 @@ private static NetworkClientFactories factories(
...
+        TimelockPaxosMetrics metrics();
         PaxosComponents components();
-        PaxosResource nonBatchedResource();
-        UseCaseAwareBatchPaxosResource batchedResource();
         NetworkClientFactories networkClientFactories();
+        Factory<PaxosProposer> proposerFactory();
+    }
+
+    @Value.Immutable
+    public abstract static class Resources {
```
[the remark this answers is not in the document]
felixdesouza: haha, so many resources classes and what not, but yeah fair shout.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
```diff
@@ -16,94 +16,101 @@
...
+import com.palantir.paxos.PaxosProposer;
+import com.palantir.paxos.PaxosProposerImpl;
 import com.palantir.timelock.config.PaxosRuntimeConfiguration;
 import com.palantir.timelock.config.TimeLockInstallConfiguration;
 import com.palantir.timelock.paxos.PaxosRemotingUtils;
+import com.palantir.timelock.paxos.TimelockPaxosAcceptorRpcClient;
+import com.palantir.timelock.paxos.TimelockPaxosLearnerRpcClient;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 
 public final class ClientPaxosResourceFactory {
```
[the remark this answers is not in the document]
felixdesouza: oop, good shout, I have no idea what to call this besides `ResourceFactory` lol

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
felixdesouza: I think at some point I had a tag that was in here, but it might have gotten lost. For the timelock metrics, the naming scheme is just plain text, whereas  if I use the other stuff I believe I'll end up using the class name which is overly verbose.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
felixdesouza: yeah, it was more of a "am I instantiating the world each time to create something per client?" I feel like I've pinned down most of them.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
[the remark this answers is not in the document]
jeremyk-91: Nope, no strong preference - it just seemed a bit random. We can leave as is

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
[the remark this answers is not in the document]
jeremyk-91: `PaxosResourceFactory`? Yeah, don't really have a good suggestion

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/ClientPaxosResourceFactory.java
jeremyk-91: `AtlasDbMetrics#instrument` is overloaded - there is a `String name` that is optional, but if provided will just use the plain text (if not provided uses the class name). Don't feel strongly, in any case

PR #4263, opened 2019-09-25
[Timelock Partitioning] Part 32: Client aware event recorders

felixdesouza:
**Goals (and why)**:
As part of leadership paxos, there is the `LeadershipEventRecorder` which tracks metrics on leadership events and notifies any observers. Such metrics should be tagged with the client.

* Rely on hierarchical `TaggedMetricRegistry` which has client populated in it already
* Downside is that we don't get this for logs, toyed with the idea of using a `Proxy`, that would append a client for the logger calls, but that feels fragile and easy to forget, so just ended up passing in a list of safe args that will be additionally logged.

Removed smoke tests that are no longer applicable, not particularly valuable + we can't get tagged metrics out of dropwizard.

* This is a break for large internal product, but should be a fairly minor one that I can fix up.

🚒

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
@@ -15,14 +15,18 @@
  */
 package com.palantir.leader;
 
+import java.util.List;
+
+import org.apache.commons.lang3.ArrayUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.Meter;
-import com.codahale.metrics.MetricRegistry;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.paxos.PaxosRoundFailureException;
 import com.palantir.paxos.PaxosValue;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 
 import net.jcip.annotations.ThreadSafe;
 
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -40,61 +44,75 @@
...
-    LeadershipEvents(MetricRegistry metrics) {
-        gainedLeadership = metrics.meter("leadership.gained");
-        lostLeadership = metrics.meter("leadership.lost");
-        noQuorum = metrics.meter("leadership.no-quorum");
-        proposedLeadership = metrics.meter("leadership.proposed");
-        proposalFailure = metrics.meter("leadership.proposed.failure");
-        leaderPingFailure = metrics.meter("leadership.ping-leader.failure");
-        leaderPingTimeout = metrics.meter("leadership.ping-leader.timeout");
-        leaderPingReturnedFalse = metrics.meter("leadership.ping-leader.returned-false");
+    LeadershipEvents(TaggedMetricRegistry metrics, List<SafeArg<Object>> safeArgs) {
```
jeremyk-91: If I'm not mistaken we don't actually pass in any args yet?

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -40,61 +44,75 @@
     private final Meter leaderPingFailure;
     private final Meter leaderPingTimeout;
     private final Meter leaderPingReturnedFalse;
+    private final Object[] safeArgs;
```
jeremyk-91: I'd prefer a more descriptive name (the idea is along the line of `globallyAddedArgs` though that may be too verbose). Maybe `contextArgs` or something like that?

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -40,61 +44,75 @@
...
         proposalFailure.mark();
     }
+
+    private Object[] withArgs(Object arg) {
+        if (safeArgs.length == 0) {
+            return new Object[] { arg };
+        } else {
+            return ArrayUtils.add(safeArgs, arg);
+        }
+    }
```
jeremyk-91: if renaming the base, probably something like `withContextArgs`?

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
jeremyk-91: discussed offline: this is planned for later, and is intentional for now

PR #4265, opened 2019-09-27
[Conjure Java Runtime] Part 11: Transformable Polling Refreshables

jeremyk-91:
**Goals (and why)**:
- Have a means for running an expensive transform if and only if needed; in particular, we want to cache the transformed value, but only rerun the transform if the underlying value changes.
- The eventual use-case is for live-reloadable configuration in CJR.

- Grab impl of `CachedTransformingSupplier` from internal security product
- I named it differently here (internally it's `CachedComposingSupplier`), because we already have a `CachedComposedSupplier` that is built to handle cases where `Objects.equals()` on the input type is expensive.

**Testing (What was existing testing like?  What have you done to improve it?)**: The new class is tested

- Not much. Please have a sanity check on the thread safety of the model.

- `PollingRefreshable.java`

**Priority (whenever / two weeks / yesterday)**: ideally by next week Tuesday

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.util;
+
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Transforms a Supplier by an output function, possibly to a different type. This Supplier caches results of
+ * applying the output function to the last seen input. This is thus usually appropriate where the value of the input
+ * does not change excessively frequently, and where the computation of the output function is expensive.
+ *
+ * See also {@link CachedComposedSupplier}. However, that differs in that that Supplier implements periodic cache
+ * invalidation, and also supports explicit versioning (in particular, for use cases where equals() comparison on
+ * the input objects is also expensive).
+ */
+public class CachedTransformingSupplier<T, U> implements Supplier<U> {
+    private final Supplier<T> inputSupplier;
+    private final Function<T, U> outputFunction;
+
+    private T cachedInput = null;
+    private U cachedOutput = null;
+
+    public CachedTransformingSupplier(Supplier<T> inputSupplier, Function<T, U> outputFunction) {
+        this.inputSupplier = inputSupplier;
+        this.outputFunction = outputFunction;
+    }
+
+    @Override
+    public synchronized U get() {
+        T currentInput = inputSupplier.get();
+
+        if (Objects.equals(cachedInput, currentInput)) {
+            return cachedOutput;
+        }
+
+        cachedInput = currentInput;
+        cachedOutput = outputFunction.apply(cachedInput);
+        return cachedOutput;
+    }
+}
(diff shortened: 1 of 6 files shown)


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
```diff
@@ -46,50 +49,58 @@
     private final Duration refreshInterval;
     private final ScheduledExecutorService poller;
 
-    private final Refreshable<T> refreshable = Refreshable.empty();
+    private final Function<T, U> transform;
+    private final Refreshable<U> refreshable = Refreshable.empty();
 
-    private T lastSeenValue;
+    private T lastSeenValue = null;
```
jkozlowski: Is this stuff threadsafe?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
```diff
@@ -46,50 +49,58 @@
...
         Preconditions.checkArgument(!refreshInterval.isNegative() && !refreshInterval.isZero(),
                 "Refresh interval must be positive, but found %s", refreshInterval);
 
         this.supplier = supplier;
         this.refreshInterval = refreshInterval;
+        this.transform = transform;
         this.poller = poller;
 
         try {
             lastSeenValue = supplier.get();
```
jkozlowski: Also, I think refreshable#get shouldn't really be called on construction.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
OStevan: In my opinion, it is since this value is used by only the refreshing thread. Also the value transformed is stored in `Refreshable` which is threadsafe and can be considered as a synchronization point between different threads.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
jeremyk-91: Hmm, yeah that seems reasonable, though not that keen on changing existing behaviour (especially since the new approach following @OStevan's suggestion doesn't really touch this class)

jeremyk-91:
Changed to a model closer to @OStevan's approach (and that @jkozlowski discussed offline with me as well). There is a small semantic difference on non-injective transforms, but I think the new behaviour is more intuitive actually.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
jkozlowski: But you also set it in the constructor, that's what I referred to.

atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Transforms a Supplier by an output function, possibly to a different type. This Supplier caches results of
+ * applying the output function to the last seen input. This is thus usually appropriate where the value of the input
+ * does not change excessively frequently, and where the computation of the output function is expensive.
+ *
+ * See also {@link CachedComposedSupplier}. However, that differs in that that Supplier implements periodic cache
+ * invalidation, and also supports explicit versioning (in particular, for use cases where equals() comparison on
+ * the input objects is also expensive).
+ */
+public class CachedTransformingSupplier<T, U> implements Supplier<U> {
```
jkozlowski: +1

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
jkozlowski: @OStevan to explain here the problem of safe publication.

OStevan: After the last comment I had a discussion with @jkozlowski about thread safety and here are the notes.

- access to `lastSeenValue` inside the updating thread and using the `Refreshable` for reads/writes is thread safe (as previously discussed)
- however based on what we read, `PoolingRefreshable` in this implementation can not be considered thread safe as the `lastSeenValue` is not final or volatile which means that we do not have safe publication of the constructed object

atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+
+        if (Objects.equals(cachedInput, currentInput)) {
+            return cachedOutput;
+        }
+
+        cachedInput = currentInput;
+        cachedOutput = outputFunction.apply(cachedInput);
+        return cachedOutput;
+    }
+}
```
jeremyk-91: argh

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/PollingRefreshable.java
jeremyk-91: Ah, good catch! Yeah, forgot about the constructor

PR #4266, opened 2019-09-27
[Conjure Java Runtime] Part 12: Conjure Java Runtime Target Factory

jeremyk-91:
🔨 Both components of the diamond dependency have merged, so this PR should be reviewed as normal.

~**💎WARNING! This PR introduces a diamond dependency. Specifically, this PR depends on both #4264 AND #4265. #4265 is much smaller and IMO less contentious, so this PR is written atop the branch for #4264. Comments concerning the implementation or tests of `PollingRefreshable` should be made on #4265. Thanks!**~

---

**Goals (and why)**:
- Actually wire up `JaxRsClient` creation.
- Note that even if CJR is broken this should not make develop unreleasable since the probability of delegating to the CJR target should be zero.

- Write `ClientOptions` and the actual `ConjureJavaRuntimeTargetFactory`
- Wire up `AtlasDbHttpClients` to have the version selecting client for live reloadable endpoints point to the CJR target factory.

- Not much yet, throwing this out for review first.

- Is there any risk that something could happen on develop that we aren't expecting?

**Where should we start reviewing?**: `ClientOptions`


atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public <T> T createProxyWithFailover(
+            Optional<TrustContext> trustContext,
+            Optional<ProxySelector> proxySelector,
+            Collection<String> endpointUris,
+            Class<T> type,
+            AuxiliaryRemotingParameters parameters) {
+        ClientConfiguration clientConfiguration = ClientOptions.FAST_RETRYING_FOR_TEST.create(
```
jeremyk-91: Please change this to the `DEFAULT` target!

jeremyk-91:
Failures here seem to be because we have failures that CJR must be configured with SSL. I think the structure of the PR shouldn't be changed much by that, so this should still be reviewable.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public <T> InstanceAndVersion<T> createLiveReloadingProxyWithFailover(
+            Supplier<ServerListConfig> serverListConfigSupplier,
+            Class<T> type,
+            AuxiliaryRemotingParameters parameters) {
+        Supplier<ServerListConfig> nonEmptyServerList = () -> injectDummyServer(serverListConfigSupplier);
+
+        // TODO (jkong): Thread leak for days. Though no regression over existing implementation in Feign.
```
gmaretic: 😢

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -0,0 +1,115 @@
...
+                        ClientOptions.DEFAULT_RETRYING::serverListToClient)
+                .getRefreshable();
+        return wrapWithVersion(JaxRsClient.create(
+                type,
+                addAtlasDbRemotingAgent(parameters.userAgent()),
+                HOST_METRICS_REGISTRY,
+                refreshableConfig));
+    }
+
+    // TODO (gmaretic): This is a hack because CJR doesn't like configurations with 0 servers, yet we claim
```
felixdesouza: if CJR doesn't like configurations with 0 servers, then we have to reassess whether or not we're realistically getting these configurations in k8s, would rather ask and then delete rather than keep since we'll never actually look at this again tbh.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public static final ConjureJavaRuntimeTargetFactory DEFAULT = new ConjureJavaRuntimeTargetFactory();
+    public static final String CLIENT_VERSION_STRING = "Conjure-Java-Runtime";
+
+    private ConjureJavaRuntimeTargetFactory() {
+        // Use the instances.
+    }
+
+    @Override
+    public <T> InstanceAndVersion<T> createProxy(Optional<TrustContext> trustContext, String uri, Class<T> type,
```
felixdesouza: we're not really using the parameters portion i.e. we're not passing in `shouldRetry`.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
felixdesouza: why are we leaking threads? plus what happened to the `CachedComposedSupplier`?

jkozlowski: Yeah, what happened to that?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
jeremyk-91: @ashrayjain I recall the rationale for doing this in the past was to allow timelock to bootstrap itself in k8s - could we check if this is still correct / needed?

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
jeremyk-91: Why: The `PollingRefreshable` has a single threaded executor that needs to be shutdown when the service is closed. This normally won't happen, but can if you close and reopen transaction managers (internal security product used to do this; some use-cases in large internal product also does this). Most services will be fine.

`CachedComposedSupplier`: Yeah, I botched the rebase - will fix.

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
jeremyk-91: good call! Fixed.

jeremyk-91: Also added a comment on the TargetFactory interface explaining that `shouldRetry` is (correctly) ignored for the failover methods

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
jkozlowski: Yeah, I was wondering what happened here. Switching to the CachedComposedSupplier fixes all those issues.

jeremyk-91: The CachedComposedSupplier is still passed to the PollingRefreshable (the change via CCS is that the PollingRefreshable does not do the transform), so I'm confident the thread leak still remains actually.

This is present for legacy reasons; switching Atlas to push-based configuration would be a break of relatively unknown extent (unclear if this is even possible) but remoting seems to demand it - so we have this workaround as a bit of a hack. It is possible to fix this with some proxy tricks (there's possibly a better way), but I would not immediately prioritise that as there is no regression relative to today.

atlasdb-config/src/test/java/com/palantir/atlasdb/http/AtlasDbHttpClientsTest.java
```diff
@@ -214,7 +227,7 @@ public void httpProxyIsConfigurableOnClientRequests() {
         clientWithHttpProxy.getTestNumber();
 
         proxyServer.verify(getRequestedFor(urlMatching(GET_ENDPOINT))
-                .withHeader(AtlasDbInterceptors.USER_AGENT_HEADER, WireMock.equalTo(DEFAULT_USER_AGENT)));
+                .withHeader(AtlasDbInterceptors.USER_AGENT_HEADER, WireMock.equalTo(DEFAULT_USER_AGENT_STRING)));
```
felixdesouza: nit, default should go to legacy

jeremyk-91: (Clarified offline - this means naming; the behaviour is already in that direction.)

atlasdb-feign/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -0,0 +1,130 @@
...
+@Value.Immutable
+public abstract class ClientOptions {
+    // TODO (jkong): Re-enable client QoS after response body leaks are handled correctly.
+    public static final ClientOptions DEFAULT_RETRYING = ImmutableClientOptions.builder()
+            .connectTimeout(Duration.ofSeconds(10))
+            .readTimeout(Duration.ofSeconds(65))
+            .backoffSlotSize(Duration.ofMillis(100))
+            .failedUrlCooldown(Duration.ofMillis(100))
+            .maxNumRetries(5)
+            .clientQoS(ClientConfiguration.ClientQoS.DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS)
```
jkozlowski: Timelock doesn't really do QoS?

Issue #3315, opened 2018-06-25
cross-AZ traffic reduction


clockfort:
We can optimistically reduce some amount of cross-AZ traffic by 
- allowing atlas service nodes to discover their own AZ
   - with some fail-through logic for zone discovery via use of AWS/gcloud/azure service URI + cassandra rack info, local hostname and cassandra node hostname parsing, and just giving up and defaulting to old behaviour
- altering the cassandra pool logic to, after establishing the set of 3 nodes with the requested data range's ownership, then have a small weighting/stickiness towards making the initial connection to the data owner of the same AZ

I did run the numbers on cost benefit and it seems worth impl'ing, even though I'm miffed that it adds complication to a codepath in atlas that is already fairly complex / state-machine-y. 

(The data I looked at was traffic costs in USD per impl at our biggest cloud impls; since I couldn't get numbers for just atlas traffic, I fudged it a bit by also looking our cassandra traffic metrics and being logical about how much of the traffic would be to and from services instead of intra-cassandra-node traffic)

tboam:
Going to pull some numbers on this this week to measure the benefit

PR #4268, opened 2019-09-30
Implement weighting by local rack when selecting cassandra hosts

Jolyon-S:
**Goals (and why)**:
* Decrease the amount of cross-rack routing required when requesting a host by increasing the likelihood of selecting a host in the current rack.

* A config option has been added to control the chance of selecting a local host over a random one.


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraServiceTest.java
```diff
@@ -46,6 +50,22 @@
     private CassandraKeyValueServiceConfig config;
     private Blacklist blacklist;
 
+    @Test
+    public void shouldReturnOnlyLocalHost() {
+        ImmutableSet<InetSocketAddress> hosts = ImmutableSet.of(HOST_1, HOST_2);
```
felixdesouza: based on the above, I would split this into two tests, one where it will use the default, and one where it will use the new logic. Where the threshold/weighting is 0.0 or 1.0 respectively.

felixdesouza: It would however be interesting to see our ete docker set up with different dcs and racks, we have a `MultiCassandraTestSuite` but you could easily add to that

and maybe having a test class similar in structure to `CassandraClientPoolIntegrationTest`, where token ranges are exposed and you could actually see pretty quickly how it interacted with different cassandra hosts.

I am also curious, is there a way to tell if a request was serviced "sub-optimally"? i.e. not using the logic you've implemented?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -84,8 +105,10 @@ public void close() {
             ImmutableRangeMap.Builder<LightweightOppToken, List<InetSocketAddress>> newTokenRing =
                     ImmutableRangeMap.builder();
 
-            // grab latest token ring view from a random node in the cluster
+            // grab latest token ring view from a random node in the cluster and update local hosts
             List<TokenRange> tokenRanges = getTokenRanges();
+            myLocation = new HostLocationSupplier(getSnitch(), myLocation).get();
```
felixdesouza: does this mean the default location changes every time you refresh token ranges? if so, then it's a bit strange to call it a default, it's more of an initial location?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public final class EC2HostLocationSupplier implements Supplier<Optional<HostLocation>> {
+
+    @Override
+    public Optional<HostLocation> get() {
+        OkHttpClient client = new OkHttpClient.Builder()
```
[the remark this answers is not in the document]
felixdesouza: also from the javadoc you want to store an instance of this in a field and reuse it. This also means that you shouldn't recreate this class inside `HostLocationSupplier` and also recreate `HostLocationSupplier` each time inside `CassandraService`.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraServiceTest.java
```diff
@@ -46,6 +50,22 @@
...
+        ImmutableSet<InetSocketAddress> localHosts = ImmutableSet.of(HOST_1);
```
felixdesouza: So looking at `filterHostsByDatacentre` there's a piece where if there is no intersection you return the original hosts, is there a relevant test you can add there? does it make sense to add this test? is the set of hosts passed in always a subset of `hosts`?

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraServiceTest.java
jkozlowski: I would be interested in a metric of local vs remote calls

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -130,6 +151,44 @@ public void close() {
...
+                if (isHostLocal(details, myLocation.get())) {
+                    newLocalHosts.add(getAddressForHostThrowUnchecked(details.getHost()));
+                }
+            }
+        }
+
+        return newLocalHosts;
+    }
+
+    private boolean isHostLocal(EndpointDetails details, HostLocation myLocation) {
```
jeremyk-91: nit: This should be `static`. 

Also, would like to check that `details.isSetHost()` is intentionally included (since it looks like it's not involved in the equals comparison).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -128,6 +130,24 @@ default int gcGraceSeconds() {
...
     }
 
+    /**
+     * This increases the likelihood of selecting an instance that is hosted in the same data centre as the process.
+     * Weighting is a ratio from 0 to 1, where 0 disables the feature and 1 forces the same data centre if possible.
+     */
+    @Value.Default
+    default double localHostWeighting() {
+        return 0.0;
+    }
```
jeremyk-91: nit: I think a `Value.Check` would be useful for the range.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -128,6 +130,24 @@ default int gcGraceSeconds() {
...
+
+    /**
+     * Controls the behaviour of the host location supplier when the snitch description does not match any known
+     * snitches.
+     */
+    @Value.Default
+    default Supplier<Optional<HostLocation>> defaultHostLocationSupplier() {
+        return () -> Optional.empty();
```
jeremyk-91: nit: `Optional::empty`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,57 @@
...
+            com.palantir.logsafe.Preconditions.checkState(response.isSuccessful(),
+                    "Getting AWS host metadata was not successful");
+
+            String responseString = response.body().string();
+            String datacentre = responseString.substring(0, responseString.length() - 2);
+            String rack = responseString.substring(responseString.length() - 1);
+
+            return Optional.of(HostLocation.of(datacentre, rack));
+        } catch (IOException e) {
+            return Optional.empty();
```
jeremyk-91: Please log a message with `e` - otherwise debugging is likely to be tricky

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
Jolyon-S: this check is intentional as I use details.getHost() in the line where it is added to newLocalHosts (so it's not involved in the equals comparison, but is rather an additional check to make sure the endpoint details contains the relevant host).

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolIntegrationTest.java
```diff
@@ -16,6 +16,7 @@
 package com.palantir.atlasdb.keyvalue.cassandra;
 
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
```
j-baker: please use AssertJ's assertions rather than junit ones, even where there is precedence in a class (and maybe convert classes you touch). It produces better error messages and is easier to use.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -128,6 +130,24 @@ default int gcGraceSeconds() {
...
+    default double localHostWeighting() {
+        return 0.0;
+    }
+
+    /**
+     * Controls the behaviour of the host location supplier when the snitch description does not match any known
+     * snitches.
+     */
+    @Value.Default
+    default Supplier<Optional<HostLocation>> defaultHostLocationSupplier() {
```
j-baker: Don't have this be a `Supplier`, just have it be the HostLocation. When you need a supplier, write `config::hostLocation` or similar.

j-baker: if you are writing methods which return suppliers but take no arguments and don't try to do anything clever, assume you're making a mistake, because the method reference itself can be thought of as the supplier.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +75,26 @@
...
+    private Set<InetSocketAddress> localHosts;
+    private HostLocationSupplier myLocationSupplier;
+
+    private final Counter randomHostsSelected;
+    private final Counter localHostsSelected;
+
     public CassandraService(MetricsManager metricsManager, CassandraKeyValueServiceConfig config, Blacklist blacklist) {
         this.metricsManager = metricsManager;
+        this.randomHostsSelected = metricsManager.getTaggedRegistry().counter(
+                MetricName.builder().safeName("RandomHostsSelected").build());
```
j-baker: for metric names in AtlasDB, generally camelCase, like classes.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Returns the client's datacentre and rack derived from Cassandra's EC2Snitch.
+ * <p>
+ * AWS has an endpoint that returns the datacentre and rack (in Cassandra terms) - this request will fail if not on AWS.
+ * The reply comes in the form "datacentre"+"rack", e.g. "us-east-1a", where datacentre is "us-east-1" and rack is "a".
+ */
+public final class EC2HostLocationSupplier implements Supplier<Optional<HostLocation>> {
+
+    OkHttpClient client = new OkHttpClient.Builder().build();
```
j-baker: please make this be private static final.

- Fields should always be final unless they need to be changed
- They should almost always be private, to avoid others using them
- They should be static, unless they need to be provided by someone else, in which case they should be passed in througha  constructor

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public Optional<HostLocation> get() {
+        try {
+            Response response = client.newCall(new Request.Builder()
+                    .get()
+                    .url("http://169.254.169.254/latest/meta-data/placement/availability-zone")
+                    .build())
+                    .execute();
+            com.palantir.logsafe.Preconditions.checkState(response.isSuccessful(),
```
j-baker: this should really be a static import of `checkState`. View fully qualified calls (e.g. `com.palantir.SomeClass.doSomething()` rather than `SomeClass.doSomething()` as 'I should only have this if I really really need it'.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocation.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import org.immutables.value.Value;
+
+import com.palantir.conjure.java.client.config.ImmutablesStyle;
+
+@Value.Immutable
+@ImmutablesStyle
+public interface HostLocation {
```
j-baker: you placed one of these in config, which means that it needs to be JSON serializable. Check out the Immutables Jackson docs for how to do this.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public final class HostLocationSupplier implements Supplier<Optional<HostLocation>> {
+
+    private String snitch;
```
j-baker: In general, you should try to inject things rather than using setters. The advantage is that your objects are always in a fully constructed state, rather than there being a middle phase where it somehow has to be initialized. In this, case, I'd be tempted to construct this object with a supplier of snitch, rather than having a field which causes the behaviour to switch out.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/WeightedHosts.java
```diff
@@ -26,8 +26,8 @@
 import com.palantir.logsafe.Preconditions;
 
 /**
- * Weights hosts inversely by the number of active connections. {@link #getRandomHost()} should then be used to
- * pick a random host
+ * Weights hosts inversely by the number of active connections. {@link #getRandomHost()} should then be used to pick a
```
j-baker: probably just don't change this class at all, given you've made no changes now?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -209,9 +268,26 @@ private String getRingViewDescription() {
         return currentPools;
     }
 
+    @VisibleForTesting
+    Set<InetSocketAddress> filterLocalHosts(Set<InetSocketAddress> hosts) {
+        if (ThreadLocalRandom.current().nextDouble() < config.localHostWeighting()) {
```
j-baker: as commented before - don't use ThreadLocalRandom. Instead, use Random.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +75,26 @@
 
     private List<InetSocketAddress> cassandraHosts;
 
+    private Set<InetSocketAddress> localHosts;
+    private HostLocationSupplier myLocationSupplier;
```
j-baker: this should be final

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +75,26 @@
...
+
     public CassandraService(MetricsManager metricsManager, CassandraKeyValueServiceConfig config, Blacklist blacklist) {
         this.metricsManager = metricsManager;
+        this.randomHostsSelected = metricsManager.getTaggedRegistry().counter(
+                MetricName.builder().safeName("RandomHostsSelected").build());
+        this.localHostsSelected = metricsManager.getTaggedRegistry().counter(
+                MetricName.builder().safeName("LocalHostsSelected").build());
         this.config = config;
         this.blacklist = blacklist;
+        this.myLocationSupplier = new HostLocationSupplier(config.defaultHostLocationSupplier());
```
j-baker: ideally you'd pass `this::getSnitch` into the constructor here

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -209,9 +268,26 @@ private String getRingViewDescription() {
         return currentPools;
     }
 
+    @VisibleForTesting
+    Set<InetSocketAddress> filterLocalHosts(Set<InetSocketAddress> hosts) {
+        if (ThreadLocalRandom.current().nextDouble() < config.localHostWeighting()) {
+            Set<InetSocketAddress> localFilteredHosts = Sets.intersection(localHosts, hosts);
```
j-baker: this will error if the localHosts thing is not set when you get here.

j-baker: because localHosts can be null. Instead, initialize it to ImmutableSet.of()

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -209,9 +268,26 @@ private String getRingViewDescription() {
         return currentPools;
     }
 
+    @VisibleForTesting
+    Set<InetSocketAddress> filterLocalHosts(Set<InetSocketAddress> hosts) {
```
j-baker: call this `maybeFilterLocalHosts`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -130,6 +151,44 @@ public void close() {
...
+
+    private Set<InetSocketAddress> refreshLocalHosts(List<TokenRange> tokenRanges) {
+        Optional<HostLocation> myLocation = myLocationSupplier.get();
+
+        if (!myLocation.isPresent()) {
+            return ImmutableSet.of();
+        }
+
+        Set<InetSocketAddress> newLocalHosts = Sets.newHashSet();
+        for (TokenRange tokenRange : tokenRanges) {
```
j-baker: you should prefer declarative programming to imperative programming in general - easier to read. So you could write this as:

```java
return tokenRanges.stream()
    .map(TokenRange::getEndpoint_details)
    .flatMap(Collection::stream)
    .filter(details -> isHostLocal(details, myLocation.get())
    .map(EndpointDetails::getHost)
    .map(this::getAddressForHostThrowUnchecked)
    .collect(toList());
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocation.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Value.Parameter
+    String datacentre();
```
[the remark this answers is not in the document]
j-baker: yep - we write code in American English.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/TestHostLocationSupplier.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public final class TestHostLocationSupplier implements Supplier<Optional<HostLocation>> {
```
j-baker: you shouldn't need to write a class for this

j-baker: should be enough to use Java lambdas

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +75,26 @@
 
     private List<InetSocketAddress> cassandraHosts;
 
+    private Set<InetSocketAddress> localHosts;
```
j-baker: for thread safety, this really needs to be `volatile`. Let's briefly catch up about what can happen if you _don't_ do this (bluf: it's bad).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public void setSnitch(String snitch) {
+        this.snitch = snitch;
+    }
+
+    @Override
+    public Optional<HostLocation> get() {
+
+        switch (snitch) {
```
j-baker: I'd suggest here that the default supplier be used in preference of anything automatic if it's present. Basically - if the user configured something, use their thing. This would take into account things like 'the server is in AWS but we are not'.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolIntegrationTest.java
```diff
@@ -74,14 +75,34 @@ public void testTokenMapping() {
...
         }
     }
 
+    @Test
+    public void testRefreshLocalHosts() {
+        HostLocation localLocation = HostLocation.of("dc1", "rack1");
+        HostLocation remoteLocation = HostLocation.of("dc1", "rack4");
+
+        assertThat(getLocalHostsUsingLocation(localLocation)).hasSize(1);
+        assertThat(getLocalHostsUsingLocation(remoteLocation)).hasSize(0);
```
j-baker: Your asserts should describe the exact condition you want. So here, you want `.isEmpty()` and `.containsExactly(someLocalHost)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +78,27 @@
...
+
+    private final Counter randomHostsSelected;
+    private final Counter localHostsSelected;
+
     public CassandraService(MetricsManager metricsManager, CassandraKeyValueServiceConfig config, Blacklist blacklist) {
         this.metricsManager = metricsManager;
+        this.randomHostsSelected = metricsManager.getTaggedRegistry().counter(
+                MetricName.builder().safeName("randomHostsSelected").build());
+        this.localHostsSelected = metricsManager.getTaggedRegistry().counter(
+                MetricName.builder().safeName("localHostsSelected").build());
```
j-baker: ah - yes, noticed you missed something. It's important that we namespace our metrics, since otherwise in metrics systems it's unclear who's producing what. We typically do this by writing something like `MetricRegistry.name(CassandraService.class, "localHostsSelected)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -209,9 +274,26 @@ private String getRingViewDescription() {
         return currentPools;
     }
 
+    @VisibleForTesting
+    Set<InetSocketAddress> maybeFilterLocalHosts(Set<InetSocketAddress> hosts) {
+        if (new Random().nextDouble() < config.localHostWeighting()) {
```
j-baker: make this Random be a field - no need to create a new one for each usage

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+            checkState(response.isSuccessful(),
+                    "Getting AWS host metadata was not successful");
+
+            String responseString = response.body().string();
+            String datacenter = responseString.substring(0, responseString.length() - 2);
+            String rack = responseString.substring(responseString.length() - 1);
+
+            return Optional.of(HostLocation.of(datacenter, rack));
+        } catch (IOException | SafeIllegalStateException e) {
+            log.info("Could not retrieve location information from AWS", e);
```
j-baker: This should really be a log.warn - this won't be terminal, but is concerning and should be looked at as to why this isn't working.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+                    .execute();
+            checkState(response.isSuccessful(),
+                    "Getting AWS host metadata was not successful");
+
+            String responseString = response.body().string();
+            String datacenter = responseString.substring(0, responseString.length() - 2);
+            String rack = responseString.substring(responseString.length() - 1);
+
+            return Optional.of(HostLocation.of(datacenter, rack));
+        } catch (IOException | SafeIllegalStateException e) {
```
j-baker: There are two things here.

1. You should prefer to not use exceptional control flow where you can, since it's hard to parse as a reader - hard to figure out where something will be caught. In some cases it's the best way, but generally try to avoid it.
2. This is especially the case where you're catching a very generic exception class such as SafeIllegalStateException. If you're catching stuff like that, you're in trouble.

j-baker: in this case, would write code like

```java
if (!response.isSuccessful()) {
    log.warn("Host metadata endpoint call was not successful", SafeArg.of("statusCode", response.code());
    return Optional.empty();
}
```
and don't catch the exception.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import javax.swing.text.html.Option;
```
j-baker: think you're probably failing checkstyle somewhere here

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+            Supplier<Optional<HostLocation>> defaultSupplier) {
+        this.snitchSupplier = snitchSupplier;
+        this.ec2Supplier = new EC2HostLocationSupplier();
+        this.defaultSupplier = defaultSupplier;
+    }
+
+    @Override
+    public Optional<HostLocation> get() {
+        Optional<HostLocation> defaultHostLocation = defaultSupplier.get();
+        if(defaultHostLocation.isPresent()) {
```
j-baker: Principle:

1. With method calls, the bracket immediately succeeds the name of the method, with no whitespace.
2. With Java keywords (if/else/switch/while/for/etc) there is a single character of whitespace between the keyword and the open bracket.

```java
someOptional.get() // good
someOptional.get () // bad
if (someCond) { // good
if(someCond) // bad
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/WeightedHosts.java
```diff
@@ -61,6 +61,7 @@ public static WeightedHosts create(Map<InetSocketAddress, CassandraClientPooling
             totalOpenRequests += openRequests;
         }
 
+
```
j-baker: probably just delete this line?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -274,13 +274,21 @@ static int getMaxTriesTotal() {
...
 
+    @VisibleForTesting
+    Set<InetSocketAddress> getLocalHosts() {
+        return cassandra.getLocalHosts();
+    }
+
     private synchronized void refreshPool() {
         blacklist.checkAndUpdate(cassandra.getPools());
 
+        Set<InetSocketAddress> resolvedConfigAddresses = config.servers()
```
j-baker: is this an error?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -68,14 +78,30 @@
...
         this.metricsManager = metricsManager;
+        this.randomHostsSelected = metricsManager.getTaggedRegistry().counter(MetricName.builder()
+                .safeName(MetricRegistry.name(CassandraService.class, "randomHostsSelected")).build());
+        this.localHostsSelected = metricsManager.getTaggedRegistry().counter(MetricName.builder()
+                .safeName(MetricRegistry.name(CassandraService.class, "localHostsSelected")).build());
         this.config = config;
         this.blacklist = blacklist;
+        this.localHosts = ImmutableSet.of();
+        this.myLocationSupplier = new HostLocationSupplier(this::getSnitch, config.overrideHostLocation());
+        this.random = new Random();
```
j-baker: as before - you can initialize this inline - doesn't have to happen in the constructor. Same with the localHosts bit.

j-baker: in general, prefer to initialize things in the constructor _that need to be_ initialized there

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -121,15 +148,68 @@ public void close() {
                     SafeArg.of("poolRefreshIntervalSeconds", config.poolRefreshIntervalSeconds()),
                     e);
 
-            // return the set of servers we knew about last time we successfully constructed the tokenMap
-            return tokenMap.asMapOfRanges().values().stream().flatMap(Collection::stream).collect(Collectors.toSet());
+            // Attempt to re-resolve addresses from the configuration; this is important owing to certain race
+            // conditions where the entire pool becomes invalid between refreshes.
```
j-baker: you've somehow managed to make other changes than your PR contained. Are you sure you merged changes in correctly?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -121,15 +148,68 @@ public void close() {
...
     private List<TokenRange> getTokenRanges() throws Exception {
         return getRandomGoodHost().runWithPooledResource(CassandraUtils.getDescribeRing(config));
     }
 
+    private String getSnitch() {
+        try {
+            return getRandomGoodHost().runWithPooledResource(
+                    (FunctionCheckedException<CassandraClient, String, Exception>) CassandraClient::describe_snitch);
+        } catch (Exception e) {
+            log.warn("Failed to retrieve snitch description", e);
```
j-baker: here you can avoid the log.warn, because you're throwing an exception

clockfort:
Fixes #3315

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/EC2HostLocationSupplier.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Returns the client's datacenter and rack derived from Cassandra's EC2Snitch.
+ * <p>
+ * AWS has an endpoint that returns the datacenter and rack (in Cassandra terms) - this request will fail if not on AWS.
+ * The reply comes in the form "datacenter"+"rack", e.g. "us-east-1a", where datacenter is "us-east-1" and rack is "a".
+ */
+public final class EC2HostLocationSupplier implements Supplier<HostLocation> {
+
+    private static final OkHttpClient client = new OkHttpClient.Builder().build();
```
carterkozak: Can we use a memoized supplier here?

Jolyon-S: Good point. This class is only ever used in conjunction with HostLocationSupplier, and that file memoizes EC2HostLocationSupplier, so it effectively is memoized, but if this class was used elsewhere then this could be a worthwhile change.

https://github.com/palantir/atlasdb/blob/617bb3e4e2452b4781eef227e109b7d616795163/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java#L40

felixdesouza: I don't see why not, I'm gonna try it.

PR #4269, opened 2019-09-30
Do not log a warning if attempting a try-lock and it fails

mswintermeyer:
**Goals (and why)**:

Get rid of extraneous logging. It should not be too surprising if we fail to grab the lock in this scenario.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AdvisoryLockConditionSuppliers.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AdvisoryLockConditionSuppliers.java
@@ -76,7 +76,16 @@ public final class AdvisoryLockConditionSuppliers {
             if (response == null) {
                 RuntimeException ex = new LockAcquisitionException(
                         "Failed to lock using the provided lock request: " + lockRequest);
-                log.warn("Could not lock successfully", ex);
+                switch (lockRequest.getBlockingMode()) {
+                    case DO_NOT_BLOCK:
+                        log.debug("Could not lock successfully", ex);
+                        break;
+                    case BLOCK_UNTIL_TIMEOUT:
+                    case BLOCK_INDEFINITELY:
+                    case BLOCK_INDEFINITELY_THEN_RELEASE:
+                        log.warn("Could not lock successfully", ex);
+                        break;
+                }
                 ++failureCount;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AdvisoryLockConditionSuppliers.java
```diff
@@ -76,7 +76,16 @@ private static HeldLocksToken acquireLock(LockService lockService, LockRequest l
             if (response == null) {
                 RuntimeException ex = new LockAcquisitionException(
                         "Failed to lock using the provided lock request: " + lockRequest);
-                log.warn("Could not lock successfully", ex);
+                switch (lockRequest.getBlockingMode()) {
```
jkozlowski: nit: Put in a separate method called "handleCouldNotLock"

mswintermeyer: Now that I also handle failureCount (and logging when we surpass the limit), it's harder to extract. Thoughts?

jkozlowski: Well now it looks like you changed semantics? Previously you'd always retry, now you won't retry on DO_NOT_BLOCK (it feels correct, but I'm surprised there isn't a test that's failing now?)

mswintermeyer: Agreed that this changes semantics.. do you agree that these semantics are indeed what we want or were the old semantics somehow preferable?

jeremyk-91: The new semantics are what I would expect and prefer overall. The concern is that a user using `DO_NOT_BLOCK` might end up failing more often, but I don't think that is actively a problem as previously these attempts would be clustered, and in the worst case the transaction will just need to try again.

jeremyk-91:
Overriding status check for changelog, as it clearly exists.

jkozlowski:
Can we followup with a test for this?

PR #4270, opened 2019-10-01
Tunable CQL client creation

OStevan:
**Goals (and why)**:
Give an interface which can be reimplemented by other internal products for better resource management.

- `CqlClientFactory` interface to be used for constructing clients
- `CqlClientFactoryImpl` default implementation which just creates cluster and executors without any custom logic behind it

- existing integration tests should cover all the cases

- is the way the decision during construction of the CqlClient at the right place?

`CqlClientFactory`
would be nice today, also paring can help if we want to roll it out quickly


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientFactoryImpl.java
```diff
@@ -0,0 +1,170 @@
...
+                    executorService.shutdownNow();
+                }
+            } catch (InterruptedException e) {
+                log.warn("Thread was interrupted while waiting for executor service to terminate.", e);
+            } catch (Exception e) {
+                log.warn("Exception on executor service termination", e);
+            }
+            Cluster cluster = session.getCluster();
+            session.close();
+            cluster.close();
```
felixdesouza: this isn't really what you want, because if you close this `cluster`, then every other session that is created from this cluster then gets shutdown.

OStevan: My idea was to make this factory "stupid" so this was intentional. However your comments do make sense.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AbstractCqlClientFactory.java
```diff
@@ -0,0 +1,159 @@
...
+
+        private final SocketAddress proxyAddress;
+
+        SocksProxyNettyOptions(SocketAddress proxyAddress) {
+            this.proxyAddress = proxyAddress;
+        }
+
+        @Override
+        public void afterChannelInitialized(SocketChannel channel) {
+            LoggerFactory.getLogger(SocksProxyNettyOptions.class).info("Proxy set up");
```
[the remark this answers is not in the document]
OStevan: Missed during self review. Removed

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AbstractCqlClientFactory.java
```diff
@@ -0,0 +1,159 @@
...
+import com.palantir.atlasdb.cassandra.CassandraServersConfigs;
+import com.palantir.conjure.java.config.ssl.SslSocketFactories;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.proxy.Socks5ProxyHandler;
+
+public abstract class AbstractCqlClientFactory implements CqlClientFactory {
+
+    AbstractCqlClientFactory() {
```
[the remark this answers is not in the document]
OStevan: Removed.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -163,6 +165,11 @@ default String getKeyspaceOrThrow() {
      */
     Optional<SslConfiguration> sslConfiguration();
 
+    @Value.Default
+    default CqlClientFactory cqlClientFactory() {
```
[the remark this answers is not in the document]
OStevan: This was one of my concerns, does it make sense to have it here or put everything in the `CqlCapableConfig`. However it can so I'll move it, it will require some changes.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientFactoryImpl.java
OStevan: Implemented caching resource manager.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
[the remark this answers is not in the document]
OStevan: Decided to leave it in the top level, the code later is simplified.

OStevan:
The PR is now simplified to reflect the discussion with @jkozlowski .

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -182,6 +184,10 @@ default String getKeyspaceOrThrow() {
      */
     Optional<SslConfiguration> sslConfiguration();
 
+    default CqlClientFactory cqlClientFactory() {
```
felixdesouza: you want to mark this as `@Value.Default` otherwise consumers will have to override it

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientFactoryImpl.java
```diff
@@ -0,0 +1,180 @@
...
+
+public final class CqlClientFactoryImpl implements CqlClientFactory {
+
+    public static final CqlClientFactory DEFAULT = new CqlClientFactoryImpl();
+
+    private CqlClientFactoryImpl() {
+        // Use instance
+    }
+
+    public CqlClient constructClient(
```
felixdesouza: I will defer to @jkozlowski for he wants internally, but it doesn't feel like this will compose  well for what he wants.

jkozlowski: Can we try to use it internally?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
OStevan: Added.

PR #4271, opened 2019-10-01
Fix caching of config hostnames in cassandra client pool

ashrayjain:
**Goals (and why)**:
Fix caching of ip <-> hostnames for servers specified in configuration. This is important for some internal use cases where cassandra ips churn pretty frequently.

Force resolve config hostnames when auto refreshing pool.

unit test.


gmaretic:
Doesn't this make it so that we can never remove decommissioned hosts that were specified in the config?

PR #4272, opened 2019-10-01
[Timelock Partitioning] Part 33: Clearer `LeadershipObserver`

felixdesouza:
**Goals (and why)**:
Previously, we would create a leadership observer once. We would then for each *client* add a task that would deregister metrics when leadership changed. This involves going through all the metrics and removing any that match. This seems rather redundant as you will likely process all of them when the first client has run the action. All of these events are tied to leader election and effectively the `LeaderElectionService` not the client itself.

Note, this somewhat becomes redundant since each client will likely have their own `LeaderElectionService`. But it removes the logic/complexity from `AsyncTimelockServicesCreator` which shouldn't really concern itself with all of that.

* Get rid of some unused methods, they're not used in big internal product or anywhere else for that matter.
* `ServiceCreator` doesn't create instrumented services, it instruments already created services. So I named it correctly.
* Move most of the logic to `PaxosLeadershipCreator` since it's tied to events emitted by the `LeaderElectionService` as opposed to client creation.

A refactor, with some potential semantics changes, wasn't really tested before, and don't think the test here is meaningful without having the old code checked in.

* `PaxosLeadershipCreator`
* `AsyncTimeLockServicesCreator`

ASAP

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
@@ -121,11 +121,10 @@ public final class Leaders {
             Supplier<LeadershipObserver> leadershipObserverFactory) {
         UUID leaderUuid = UUID.randomUUID();
 
-        LeadershipObserver leadershipObserver = leadershipObserverFactory.get();
         PaxosLeadershipEventRecorder leadershipEventRecorder = PaxosLeadershipEventRecorder.create(
                 metricsManager.getTaggedRegistry(),
                 leaderUuid.toString(),
-                leadershipObserver,
+                leadershipObserverFactory.get(),
                 ImmutableList.of());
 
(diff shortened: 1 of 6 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -114,22 +103,28 @@ public static LocalPaxosServices createInstrumentedLocalServices(
...
+                runtime,
+                remotePaxosServerSpec,
+                userAgent,
+                AsyncLeadershipObserver::create);
     }
 
     public static LocalPaxosServices createInstrumentedLocalServices(
             MetricsManager metricsManager,
             LeaderConfig config,
             Supplier<LeaderRuntimeConfig> runtime,
```
gmaretic: Since we are already refactoring, this is never used...

PR #4275, opened 2019-10-03
Additional config to allow allSafeForLogging on transaction manager level

### LINKED ISSUE #3834: Logging doesn't work as expected if multiple TransactionManagers are used
`LoggingArgs` keeps a static reference to a `KeyValueServiceLogArbitrator logArbitrator`, which is "hydrated" at TransactionManager initialization. This hydration process overrides any previous hydration from previously initialized TransactionManager.

Ideally, `LoggingArgs` would not be static. Another alternative, faster to implement but in my mind less ideal, would be to make the hydration process be cumulative, rather than override any previous run of it.

JiahuiJiang:
**Goals (and why)**:
Allow setting a flag to mark all information to be logged to be considered safe.
This is useful when there are multiple transaction managers, right now the hydrate code will make the allowed tableReference override the earlier initialized transaction manager's safe table names.

Fixes #3834 

Initially everything is unsafe.
The first time a transaction manager is initialized, if it has "allSafeForLogging" to true, everything is safe for logging / if it has "allSafeForLogging" to false, only the tableReferences/columnNames/RowNames marked as safe are safe. Then for the following transaction managers, if there is one where the allSafeForLogging is false, we stop considering all the transaction managers to be safe, and this latest transaction manager will override the whole safe data information.


atlasdb-client/src/main/java/com/palantir/atlasdb/logging/LoggingArgs.java
```diff
@@ -58,6 +59,7 @@
     }
 
     private static volatile KeyValueServiceLogArbitrator logArbitrator = KeyValueServiceLogArbitrator.ALL_UNSAFE;
+    private static volatile Optional<Boolean> allSafeForLogging = Optional.empty();
```
jeremyk-91: If I'm not mistaken `allSafeForLogging` is only written here by the static initializer and thereafter read to and written from in `setAllSafeForLogging`, so all accesses across threads should already have the happens-before relationship.

atlasdb-client/src/main/java/com/palantir/atlasdb/logging/LoggingArgs.java
```diff
@@ -67,11 +69,33 @@ public static synchronized void hydrate(Map<TableReference, byte[]> tableRefToMe
         logArbitrator = SafeLoggableDataUtils.fromTableMetadata(tableRefToMetadata);
     }
 
+    public static synchronized void setAllSafeForLogging(boolean isNewKeyValueServiceAllSafeForLogging) {
```
jeremyk-91: nit: can we name this `accumulate...` or something along those lines? (since setting to `true` won't always lead to behaviour we'd expect with `true` if there was a `false` passed earlier)

atlasdb-client/src/test/java/com/palantir/atlasdb/logging/KeyValueServiceLogArbitratorTest.java
```diff
@@ -43,4 +44,19 @@ public void allUnsafeArbitratorMarksRowNamesAsUnsafe() {
...
+
+    @Test
+    public void allSafeArbitratorMarksRowNamesAsUnsafe() {
+        assertThat(ALL_SAFE.isRowComponentNameSafe(TABLE_REFERENCE, ROW_NAME)).isTrue();
+    }
+
+    @Test
+    public void allSafeArbitratorMarksColumnNamesAsUnsafe() {
+        assertThat(ALL_SAFE.isColumnNameSafe(TABLE_REFERENCE, COLUMN_NAME)).isTrue();
+    }
```
jeremyk-91: I think these test names should end in `Safe`

JiahuiJiang: oops!

JiahuiJiang:
updated :D

jeremyk-91:
Merging: The only failing check is Changelog, but that *is* present, and I suspect related to #4272 in some way.

PR #4276, opened 2019-10-03
apply plugin: 'com.palantir.revapi'

iamdanfox:
**Goals (and why)**:

https://github.com/palantir/gradle-revapi helps people _spot_ API/ABI breaks, so you can decide whether you really want to ship them.

Example output:
```
java.method.removed: Method was removed.

  old: method <T> T com.palantir.atlasdb.factory.ServiceCreator::createInstrumentedService(com.codahale.metrics.MetricRegistry, T, java.lang.Class<T>)
  new: <none>

  BINARY: BREAKING, SOURCE: BREAKING

  From old archive: atlasdb-config-0.156.1.jar
  From new archive: <none>

  If this is an acceptable break that will not harm your users, you can ignore it in future runs like so for:

    * Just this break:
        ./gradlew :atlasdb-config:revapiAcceptBreak --justification "{why this break is ok}" \
          --code "java.method.removed" \
          --old "method <T> T com.palantir.atlasdb.factory.ServiceCreator::createInstrumentedService(com.codahale.metrics.MetricRegistry, T, java.lang.Class<T>)"
    * All breaks in this project:
        ./gradlew :atlasdb-config:revapiAcceptAllBreaks --justification "{why this break is ok}"
    * All breaks in all projects:
        ./gradlew revapiAcceptAllBreaks --justification "{why this break is ok}"

```

- I've only applied this to _one_ project so far, which doesn't really give you much assurance.
- We've seen revapi OOM when run on lots and lots of projects.
- Maybe this is actually unusable because there are so many _public_ classes which are actually considered _private_ API of atlasdb.

cc @felixdesouza


iamdanfox:
@j-baker can you clarify the `do not merge` here pls?

felixdesouza:
he's ooto I believe, I'm happy to have it even if it's not enabled on any particular gradle project (is what I imagine the 'do not merge' is for), it can be a lightweight tool to check whether or not a change you're about to do that's potentially breaking will actually be breaking, we're pretty judicious about this generally

gmaretic:
Had troubles getting this to pass as it was comparing to the wrong artifact, then botched a rebase so opened a fresh PR #4290

PR #4281, opened 2019-10-07
Rename EC2 to Ec2 as per the docs

Jolyon-S:
Changed the matched string from EC2Snitch to Ec2Snitch, as the docs imply that that is what will be returned. Also added logging to confirm the exact type of snitch supplied.


felixdesouza:
I think you still have the old version of `EC2HostLocationSupplier` in develop? unless they're not useful anymore. 

Also please fill in the template, it's useful for posterity sake.

PR #4283, opened 2019-10-07
Fix Ec2HostLocationSupplier using a different format to Cassandra

Jolyon-S:
**Goals (and why)**:
Modify Ec2HostLocationSupplier to emulate Cassandra's Ec2Snitch logic (as it is later comparing against parsed info from that class).

The response string from the AWS endpoint is now parsed using code copied from Cassandra's implementation (or as close as possible)

Added a test to make sure that the parsing gives the expected result

Today

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
@@ -67,8 +67,9 @@ public final class Ec2HostLocationSupplier implements Supplier<HostLocation> {
 
         // hack for CASSANDRA-4026
         ec2region = az.substring(0, az.length() - 1);
-        if (ec2region.endsWith("1"))
+        if (ec2region.endsWith("1")) {
             ec2region = az.substring(0, az.length() - 3);
+        }
 
         return HostLocation.of(ec2region, ec2zone);
 
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
@@ -19,6 +19,7 @@ package com.palantir.atlasdb.keyvalue.cassandra.pool;
 import java.io.IOException;
 import java.util.function.Supplier;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.palantir.logsafe.Preconditions;
 
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
```diff
@@ -46,13 +47,32 @@ public HostLocation get() {
...
+
+        // hack for CASSANDRA-4026
+        ec2region = az.substring(0, az.length() - 1);
+        if (ec2region.endsWith("1")) {
+            ec2region = az.substring(0, az.length() - 3);
+        }
+
+        return HostLocation.of(ec2region, ec2zone);
+
+
```
jeremyk-91: nit: We don't normally leave whitespace after the return statement

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
```diff
@@ -46,13 +47,32 @@ public HostLocation get() {
...
         }
     }
+
+    @VisibleForTesting
+    static HostLocation parseHostLocation(String az) {
+        // Code is copied from Cassandra's Ec2Snitch. The result of this parsing must match Cassandra's as closely as
+        // possible, as the strings are later matched exactly.
+
+        String ec2region;
+        String ec2zone;
```
jeremyk-91: Non-actionable: This is fine and I guess copied from original, though we normally prefer to declare the variables as they occur

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2HostLocationSupplier.java
```diff
@@ -46,13 +47,32 @@ public HostLocation get() {
...
-            return HostLocation.of(datacenter, rack);
+            return parseHostLocation(response.body().string());
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
     }
+
+    @VisibleForTesting
+    static HostLocation parseHostLocation(String az) {
+        // Code is copied from Cassandra's Ec2Snitch. The result of this parsing must match Cassandra's as closely as
```
jeremyk-91: We should reimplement this method, as otherwise we'll need to go through the documentation for licensing (which is really too much work for this).

changelog/@unreleased/pr-4283.v2.yml
```diff
@@ -0,0 +1,18 @@
...
+    Modify Ec2HostLocationSupplier to emulate Cassandra's Ec2Snitch logic (as it is later comparing against parsed info from that class).
+
+    **Implementation Description (bullets)**:
+    The response string from the AWS endpoint is now parsed using code copied from Cassandra's implementation (or as close as possible)
+
+    **Testing (What was existing testing like?  What have you done to improve it?)**:
+    Added a test to make sure that the parsing gives the expected result
+
+    **Priority (whenever / two weeks / yesterday)**:
+    Today
```
jeremyk-91: Probably don't want to include all of this in the changelog. Most of the time, something along the lines of the initial message (i.e. "Ec2HostLocationSupplier now users a format consistent with Cassandra's Ec2Snitch") is sufficient.

Agree this is a fix - though for fixes it's usually good to say what the broken behaviour was, and what impact that might have had on users on the version without the fix (i.e. "`Ec2HostLocationSupplier` may have produced results inconsistent with Cassandra's `Ec2Snitch`, meaning that our feature to send more traffic to local nodes wouldn't work properly" or something like that).

Issue #4196, opened 2019-08-30
Deal with concurrent C* schema mutations


jkozlowski:
We are experiencing issues with C* gossip getting out of sync when multiple concurrent schema mutations (create keyspace/create table) are occurring on large stacks.

Raising here to track possible improvements in AtlasDB to help alleviate these issues. 

cc @tpetracca @j-baker

gmaretic:
This should be made better if we merge https://github.com/palantir/atlasdb/pull/3754 ?

tpetracca:
Another instance today: PDS-101155

clockfort:
(necro bump to put in my two-penny worth)
we used to have this problem with %excessively_old_cassandra_version% , so we fully serialized all schema mutations with client(atlas) side locks. but that proved quite slow for some use cases

jkozlowski:
I think we fixed this in Cass, so closing

PR #4285, opened 2019-10-07
Better checks on schema changes

gmaretic:
**Goals (and why)**:
Reduce the likelihood of hitting  #4196, and revert erroneous quorum calculation for schema agreement. A copy of #3754 with fixed merge conflicts, since the old PR was too old to pass checks.

Schema mutations are now always preceded, and followed, by a check that schema versions of all nodes agree.
We revert to allowing a single node to be unreachable here to have some measure of HA. The old check that allowed for half the nodes (rounded down) to be unreachable was done in error and makes little sense.

Tests that verify that we fail in the cases where we have more than one unreachable node, or schema disagreement.

Technically, this can make us less HA when we have more than 3 nodes, but if we have more than 1 unreachable node, we should probably refrain from making schema changes and instead fix our cluster.

small

Not urgent, but maybe try to merge it in less than 8 months


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -71,15 +72,41 @@ private CassandraKeyValueServices() {
     }
 
     /**
-     * Attempt to wait until nodes' schema versions match.
+     * Wraps the execution of the specified task with checks verifying that schema versions on Cassandra nodes are
```
jeremyk-91: nit: I get that `wraps` implies enclosing, but would be prefer to be more explicit and say something like "performs checks both before and after ..." as I think it's reasonable in a computing context to have wrappers only intercept things on one side.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -126,29 +150,10 @@ static void waitForSchemaVersions(
...
-            String unsafeSchemaChangeDescription)
-            throws TException {
-        waitForSchemaVersions(config, client, "before " + unsafeSchemaChangeDescription);
-        task.run();
-        waitForSchemaVersions(config, client, "after " + unsafeSchemaChangeDescription);
-    }
-
-    static boolean uniqueSchemaWithQuorumAgreementAndOtherNodesUnreachable(
-            CassandraKeyValueServiceConfig config,
+    private static boolean uniqueSchemaOnAllNodesAllowingForOneUnreachable(
```
jeremyk-91: I'd prefer we left a comment explaining why we chose this, and why it's not a function of topology (following the discussion on the previous version of this PR)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -88,11 +115,8 @@ static void waitForSchemaVersions(
...
         do {
-            // This only returns the schema versions of nodes that the client knows exist. In particular, if a node we
-            // shook hands with goes down, it will have schema version UNREACHABLE; however, if we never shook hands
-            // with a node, there will simply be no entry for it in the map. Hence the check for the number of nodes.
             versions = client.describe_schema_versions();
-            if (uniqueSchemaWithQuorumAgreementAndOtherNodesUnreachable(config, versions)) {
+            if (uniqueSchemaOnAllNodesAllowingForOneUnreachable(versions)) {
                 return;
             }
             sleepTime = sleepAndGetNextBackoffTime(sleepTime);
```
jkozlowski: Would it make sense to add some jitter to the backoff time? That would hopefully further reduce any contention we may have if multiple changes are happening at the same time?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -157,13 +162,6 @@ static boolean uniqueSchemaWithQuorumAgreementAndOtherNodesUnreachable(
...
-    private static int getNumberOfReachableNodes(Map<String, List<String>> versions) {
-        return versions.entrySet().stream().filter(entry -> !entry.getKey().equals(VERSION_UNREACHABLE))
-                .map(Entry::getValue)
-                .mapToInt(List::size)
-                .sum();
-    }
-
     private static long sleepAndGetNextBackoffTime(long sleepTime) {
         try {
             Thread.sleep(sleepTime);
```
jkozlowski: Maybe could use some logging here? We don't expect any to happen if there's only a single mutation happening at a time, but when there are many, it would probably help with figuring out what's going on.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
gmaretic: Added to the above java doc

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
gmaretic: Added in 10% jitter

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
gmaretic: Added an info log line, it doesn't provide details, but that's probably fine as we get details if we fail to converge on the schema version

gmaretic:
@amandachow @tpetracca can one of you ack that this makes sense?

gmaretic:
I don't think this is ever happening

PR #4289, opened 2019-10-08
More flexible config

OStevan:
**Goals (and why)**:
The previous config was too restrictive compared to what we need. The change makes it more flexible for further usage. We want to keep the implementation less prone to errors in internal deployment set up PDS-101162.

- the config is more verbose but more flexible depending on what is needed
- there is check if CQL and Thrift connect to the same set of hosts

- tests were updated to reflect the changes

- this is in general a breaking change but nobody should be using this currently

- CassandraServersConfigs

- by Wednesday


changelog/@unreleased/pr-4289.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: CQL capable config format change.
```
jeremyk-91: Probably need a bit more detail here :)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraServersConfigs.java
```diff
@@ -131,30 +131,25 @@ public CqlCapableConfigTuning tuning() {
...
-        }
+        @Value.Check
+        final void check() {
+            checkPortNumbers(thriftHosts(), "'thriftPort'");
 
-        private Set<InetSocketAddress> constructHosts(int port) {
-            return hosts().stream()
-                    .map(host -> new InetSocketAddress(host, port))
-                    .collect(Collectors.toSet());
+            checkPortNumbers(cqlHosts(), "'cqlPort'");
```
jeremyk-91: In general for these, it looks fine - but do we need the single quotes around the names?

OStevan: Kept it because it was like that previously.

PR #4290, opened 2019-10-08
apply plugin: 'com.palantir.revapi'

gmaretic:
**Goals (and why)**:
Fix the bothced rebase in #4276 and get it to pass circle


gmaretic:
@j-baker what were the concerns with merging this in? Seems reasonable to me

PR #4291, opened 2019-10-08
[Conjure Java Runtime] Part 14: Experiment Running Proxy and Data Collection

jeremyk-91:
**Goals (and why)**:
- Allow users to enable conjure java runtime while still having a safeguard against failures/gross misconfiguration. 

- Users can set the conjure java runtime probability above 0; in the event of a failure, we knock the probability down to 0, so we only use Feign.
- We can attempt 10 minutes later to make a request via CJR again (probability reverts to its above-0 value).
- Note that we probably *don't* want people to actually run experiments just yet - we need to declare a product dependency on suitable versions of timelock-server before we can be sure that the values are meaningful. It won't explode in that CJR clients will behave correctly (if inefficiently) with legacy servers, but the perf numbers won't be representative.

- Added tests for the new class.

- Is setting the default maximum CJR probability to 1% right off the bat too aggressive? If this causes problems we can quickly mitigate the damage by setting the config explicitly to zero.
- Are there possible unexpected race conditions inside the proxy?
- Should I have implemented a slow ramp-up to the user's specified max probability? I think that is more work than is necessary.

**Where should we start reviewing?**: ExperimentRunningProxy.java

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
@@ -74,11 +74,7 @@ public final class ExperimentRunningProxy<T> extends AbstractInvocationHandler {
             return method.invoke(target, args);
         } catch (InvocationTargetException e) {
             if (runExperiment) {
-                log.info("Experiment failed; we will revert to the fallback service. We will allow attempting to"
-                                + " use the experimental service again in {} milliseconds.",
-                        SafeArg.of("retryDurationMillis", REFRESH_INTERVAL.toMillis()),
-                        e);
-                markExperimentFailure();
+                markExperimentFailure(e);
             }
             if (e.getTargetException() instanceof NotInitializedException) {
                 log.warn("Resource is not initialized yet!");
@@ -91,8 +87,10 @@ public final class ExperimentRunningProxy<T> extends AbstractInvocationHandler {
         return useExperimental.getAsBoolean() && Instant.now(clock).compareTo(nextPermittedExperiment.get()) > 0;
     }
 
-    private void markExperimentFailure() {
+    private void markExperimentFailure(Exception exception) {
         nextPermittedExperiment.accumulateAndGet(Instant.now(clock).plus(REFRESH_INTERVAL),
                 (existing, current) -> existing.compareTo(current) > 0 ? existing : current);
+        log.info("Experiment failed; we will revert to the fallback service. We will allow attempting to use"
+                + " the experimental service again in {}.", SafeArg.of("retryDuration", REFRESH_INTERVAL), exception);
     }
 }
--- a/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
@@ -1,0 +1,91 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.common.proxy;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BooleanSupplier;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
```diff
@@ -0,0 +1,98 @@
...
+    protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
+        boolean runExperiment = useExperimental();
+        Object target = runExperiment ? experimentalService : fallbackService;
+        try {
+            return method.invoke(target, args);
+        } catch (InvocationTargetException e) {
+            if (runExperiment) {
+                log.info("Experiment failed; we will revert to the fallback service. We will allow attempting to"
+                                + " use the experimental service again in {} milliseconds.",
+                        SafeArg.of("retryDurationMillis", REFRESH_INTERVAL.toMillis()),
```
gmaretic: Why not just log the duration?

atlasdb-commons/src/test/java/com/palantir/common/proxy/ExperimentRunningProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+        throwOnFirstExperiment();
+
+        assertThatThrownBy(experimentSupplier::getAsInt).isEqualTo(RUNTIME_EXCEPTION);
+        assertThat(experimentSupplier.getAsInt()).isEqualTo(FALLBACK_INT);
+    }
+
+    @Test
+    public void attemptsExperimentAgainAfterEnoughTimeHasElapsed() {
+        enableExperiment();
+        throwOnFirstExperiment();
```
felixdesouza: separate out the two, i.e. don't have it be throw then succeed, just have it be throw, and then  when you need it to succeed, say it should succeed

atlasdb-commons/src/test/java/com/palantir/common/proxy/ExperimentRunningProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    public void attemptsExperimentIfRequested() {
+        enableExperiment();
+        assertThat(experimentSupplier.getAsInt()).isEqualTo(EXPERIMENTAL_INT);
+    }
+
+    @Test
+    public void canFallBackIfExperimentFails() {
+        enableExperiment();
+        throwOnFirstExperiment();
```
felixdesouza: same thing here, it doesn't need to suddenly start working after the first experiment, plus is this test strictly necessary, considering it's subsumed by `attemptsExperimentAgainAfterEnoughTimeHasElapsed`

atlasdb-commons/src/test/java/com/palantir/common/proxy/ExperimentRunningProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Test
+    public void attemptsExperimentAgainAfterEnoughTimeHasElapsed() {
+        enableExperiment();
+        throwOnFirstExperiment();
+
+        assertThatThrownBy(experimentSupplier::getAsInt).isEqualTo(RUNTIME_EXCEPTION);
+        assertThat(experimentSupplier.getAsInt()).isEqualTo(FALLBACK_INT);
+
+        when(clock.instant()).thenReturn(Instant.ofEpochSecond(1_000));
```
felixdesouza: why does this work (well I know why because I read the source)? I think you want to say something along the lines of:

```java
when(clock.instant()).thenReturn(Instant.ofEpochSecond(REFRESH_INTERVAL.minus(Duration.ofSecond(1))

// assert that it doesn't work

when(clock.instant()).thenReturn(Instant.ofEpochSecond(REFRESH_INTERVAL.plus(Duration.ofSecond(1))
```

if you don't want to refer to the constant, then just add it as a constructor parameter.

It's not clear from the test why adding a 1000 seconds in particular does anything, nor where any sort of refresh/backoff period comes into play.

atlasdb-commons/src/test/java/com/palantir/common/proxy/ExperimentRunningProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final RuntimeException RUNTIME_EXCEPTION = new RuntimeException("foo");
+
+    private final IntSupplier experimentalIntSupplier = mock(IntSupplier.class);
+    private final IntSupplier fallbackIntSupplier = mock(IntSupplier.class);
+    private final BooleanSupplier useExperimental = mock(BooleanSupplier.class);
+    private final Clock clock = mock(Clock.class);
+    private final ExperimentRunningProxy<IntSupplier> proxy = new ExperimentRunningProxy<>(
+            experimentalIntSupplier, fallbackIntSupplier, useExperimental, clock);
+
+    private final IntSupplier experimentSupplier = intSupplierForProxy(proxy);
```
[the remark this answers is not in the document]
gmaretic: Better?

atlasdb-commons/src/test/java/com/palantir/common/proxy/ExperimentRunningProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+        enableExperiment();
+        throwOnFirstExperiment();
+
+        assertThatThrownBy(experimentSupplier::getAsInt).isEqualTo(RUNTIME_EXCEPTION);
+        assertThat(experimentSupplier.getAsInt()).isEqualTo(FALLBACK_INT);
+    }
+
+    @Test
+    public void attemptsExperimentAgainAfterEnoughTimeHasElapsed() {
+        enableExperiment();
```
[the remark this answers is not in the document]
gmaretic: Better?

PR #4294, opened 2019-10-09
Remove local host routing metrics

Jolyon-S:
Remove local host and remote host counters from CassandraService to save money as they are not necessary in the long term (only useful for verifying that the change worked initially).


Jolyon-S:
decided this was not necessary for the foreseeable future.

PR #4300, opened 2019-10-11
Reap by immutable timestamp

mswintermeyer:
**Goals (and why)**:


j-baker:
Please can we get a test for the behaviour that was clearly broken before this PR but that works after this PR?

clockfort:
> Placing the token on the queue says that you don't want to unlock it now but want to consider it again later, basically, so #4181 didn't change the reaping directly. After this change, it definitely looks like we do execute reaping.

That's why #4181 doesn't place it back on the queue and elects to just log and not place it back on the queue

clockfort:
I looked into writing a test for this; the tests for the old old lock server whose behaviour is being modified here seem to be almost non existent and just smoke and logging tests 😞

j-baker:
Ok, but because of that your change no-opped and the customer still had a broken stack

PR #4304, opened 2019-10-15
Listenable future refactoring for transaction get

OStevan:
**Goals (and why)**:
Preparations for getAsync on transactions. Make it easier to share code in the current and future call path.

- wrapped the results in `Futures.immediateFuture`

- current tests should cover everything

- `SnapshotTransaction`

today

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
@@ -122,7 +122,7 @@ public class CachingTransaction extends ForwardingTransaction {
             return get(
                     tableRef,
                     cells,
-                    ((tableReference, toRead) -> Futures.immediateFuture(super.get(tableReference, toRead)))).get();
+                    (tableReference, toRead) -> Futures.immediateFuture(super.get(tableReference, toRead))).get();
         } catch (InterruptedException | ExecutionException e) {
             throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
         }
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -596,16 +618,24 @@ private void extractLocalWritesForRow(
             }
         }
 
-        // We don't need to read any cells that were written locally.
-        result.putAll(getFromKeyValueService(tableRef, Sets.difference(cells, result.keySet())));
+        return Futures.transform(getFromKeyValueService(tableRef, Sets.difference(cells, result.keySet()), cellLoader),
+                fromKeyValueService -> {
+                    // We don't need to read any cells that were written locally.
```
felixdesouza: nit: comment is in the wrong place

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
```diff
@@ -113,8 +118,19 @@ public Transaction delegate() {
 
     @Override
     public Map<Cell, byte[]> get(TableReference tableRef, Set<Cell> cells) {
+        try {
+            return get(
+                    tableRef,
+                    cells,
+                    ((tableReference, toRead) -> Futures.immediateFuture(super.get(tableReference, toRead)))).get();
```
felixdesouza: nit: you have an extra parenthesis

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
OStevan: Fixed.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/CachingTransaction.java
OStevan: Fixed.

jkozlowski:
I don't think this change is complete: it makes the initial load async, but those are always LIMIT 1 right? So you could read a value that another transaction is currently writing, but is not visible to you. That's why as part of this codepath we eventually end up in #getWithPostFilteringInternal where we look at timestamps (those require talking to the KVS) and reload them at the appropriate timestamp.

PR #4305, opened 2019-10-15
[Timelock Partitioning] Part 36: Introduce a new LeaderElectionServiceBuilder

felixdesouza:
`PaxosLeaderElectionService` has had quite the overhaul in terms of its parameters, the builder was introduced as a way to construct things safely.

It's rather annoying to try and plumb through changes to the existing builder when some consumers want a `PaxosLearner` vs others want a `PaxosLearnerNetworkClient` for example. If you're not in the process of making these directly, it's just annoying as you'd have to create a `SingleLeader(Acceptor|Network)Client` which is composed of `Paxos(Acceptor|Learner)s` and construction just becomes that bit longer. Although I do not mind making the change to internal product if necessary.

**Goals (and why)**:
Trim down the api surface of `PaxosLeaderElectionService` so that we can untangle some of its dependencies and make it easier to instantiate.

* No longer implements `PingableLeader` if you want a local one you can get a `LocalPingableLeader`, if you want a remote, then just make one.
* `PaxosProposer` is now made for you, you have all the constituent parts anyway.
* Remove `getPotentialLeaders` from `LeaderElectionService`, untangles the dependencies, and the health check can be passed in the leaders via the composing method
* Introduce `LeaderElectionServiceBuilder`, the "canonical way" to build `LeaderElectionService`. The `PaxosLeaderElectionServiceBuilder` delegates to this class.
* `PaxosLeaderElectionServiceBuilder` breaks, so that we can construct the proposer for you.
* Delete some unused/unnecessary `PaxosProposerImpl` static factory methods, no longer used + you don't need to make one anymore.
* Timelock healthcheck was weirdly a supplier, and I don't really know why. Simplified that too.

Mainly refactors, anything here should be caught by existing tests.

Should I just bite the bullet and get internal product + non-timelock atlas to use the new builder?

* `PaxosLeaderElectionService`
* `LeaderElectionServiceBuilder`
* `PaxosLeaderElectionServiceBuilder`

ASAP 💃


leader-election-impl/src/main/java/com/palantir/leader/LeaderElectionServiceBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+                randomWaitBeforeProposingLeadership().toMillis(),
+                eventRecorder());
+    }
+
+    private PaxosProposer buildProposer() {
+        return PaxosProposerImpl.newProposer(acceptorClient(), learnerClient(), quorumSize(), leaderUuid());
+    }
+
+    private int quorumSize() {
+        Preconditions.checkArgument(quorumSize > 0, "quorumSize not set");
```
jeremyk-91: Is there a way to verify that the quorum size isn't greater than the number of nodes? I think we hide this detail now so it's not easy, in which case it's fine.

---

Also, non-actionable: There's nothing stopping someone from trying to set the quorum size in the builder to zero or a negative number.

leader-election-impl/src/main/java/com/palantir/leader/LeaderElectionServiceBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private Duration pingRate() {
+        return Preconditions.checkNotNull(pingRate, "pingRate not set");
+    }
+
+    private Duration randomWaitBeforeProposingLeadership() {
+        return Preconditions.checkNotNull(randomWaitBeforeProposingLeadership,
+                "randomWaitBeforeProposingLeadership not set");
+    }
```
jeremyk-91: nit / maybe non-actionable: Do we want to check, along the lines of the quorum thing, that durations are not negative?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -265,11 +238,6 @@ private boolean isThisNodeTheLeaderFor(PaxosValue value) {
         return value.getLeaderUUID().equals(proposer.getUuid());
     }
 
-    // This is used by an internal product CLI.
-    public ImmutableList<PaxosAcceptor> getAcceptors() {
-        return acceptors;
-    }
-
```
jeremyk-91: To confirm: We've since changed the internal product CLI in question, I'm guessing?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionServiceBuilder.java
```diff
@@ -114,34 +107,51 @@ public PaxosLeaderElectionServiceBuilder eventRecorder(PaxosLeaderElectionEventR
...
+    }
+
+    private int quorumSize() {
+        Preconditions.checkArgument(quorumSize > 0, "quorum size not set");
+        return quorumSize;
+    }
+
+    private UUID leaderUuid() {
+        return Preconditions.checkNotNull(leaderUuid, "leader uuid not set");
+    }
```
jeremyk-91: Hmm, I think it's okay to not check here and do all the checks in `LeaderElectionServiceBuilder` - it seems a bit strange that we do some but not all of them here, and then do a full set of them in LESB. Might have read something incorrectly.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -200,7 +195,7 @@ public long getLatestSchemaVersion() {
 
     // No runtime configuration at the moment.
     private void registerPaxosResource() {
-        paxosResources.resourcesForRegistration().forEach(registrar::accept);
+        paxosResources.resourcesForRegistration().forEach(registrar);
```
jeremyk-91: non-actionable: This is correct. I prefer how the old one reads, but don't feel strongly.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -155,9 +156,6 @@ public static LocalPaxosServices createInstrumentedLocalServices(
                 PTExecutors.newCachedThreadPool(daemonThreadFactory("atlas-proposer")),
                 metricsManager.getRegistry(),
                 MetricRegistry.name(PaxosProposer.class, "executor"));
-        PaxosProposer proposer = AtlasDbMetrics.instrument(metricsManager.getRegistry(), PaxosProposer.class,
```
jeremyk-91: It looks like this change drops the `PaxosProposer` metrics, which I'm not sure we want to do.

leader-election-impl/src/main/java/com/palantir/leader/LeaderElectionServiceBuilder.java
felixdesouza: hmm, so it turns out, we don't actually need quorum size, because it's only there to be exposed through the `PaxosProposer`, can remove, then this comment is moot

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
felixdesouza: yeah, I've made most of the migrations, it's just the builder parts that will break.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionServiceBuilder.java
felixdesouza: sure, I think I started off here, and then moved into a new builder

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
felixdesouza: not fussed, can revert

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
felixdesouza: makes things a bit awkward maybe, as now I have to pass in a metrics registry, and you might want to instrument it with certain tags/metrics etc. Another thought I had was having some sort of decorator function which is just a `UnaryOperator` on the builder which would allow you to instrument the metrics as you see fit, but seemed a bit overkill, but can do

.palantir/revapi.yml
```diff
@@ -1,8 +1,98 @@
...
+      new: null
+      justification: "this is all internal code"
+    - code: "java.method.removed"
+      old: "method com.palantir.atlasdb.factory.ImmutableLocalPaxosServices com.palantir.atlasdb.factory.ImmutableLocalPaxosServices::withIsCurrentSuspectedLeader(java.util.function.Supplier<java.lang.Boolean>)"
+      new: null
+      justification: "this is all internal code"
+    - code: "java.method.addedToInterface"
+      old: null
+      new: "method com.palantir.leader.PingableLeader com.palantir.atlasdb.factory.Leaders.LocalPaxosServices::localPingableLeader()"
+      justification: "this is all internal code"
```
iamdanfox: @CRogers kinda interesting feedback here - felix ended up running `acceptAllBreaks` twice because develop was released while this PR was still in flight. What do you think about getting rid of the version number and just hashing the content of the break somehow?

PR #4309, opened 2019-10-16
Added table creation tests.

OStevan:
**Goals (and why)**:
Add tests for code paths which use Datastax Cassandra driver even before async additions. Bumping the driver should flag and query string construction changes so they can be addressed.

- extracted a function for query string creation `CassandraTableCreator`

- previously there were no tests for this

- should there be more tests or is just the default case enough

whenever


felixdesouza:
As spoken offline, I'm not sure this is useful to be checked in. We know the query string will change the moment we upgrade to the newer version. Then we will have to change the test to make it work with the new stuff. Which just feels like the test is useless to be checked in at this point. 

It's useful whilst doing the upgrade to ensure you're doing the right thing, but I don't get its value to be checked in. @jkozlowski for SA.

jkozlowski:
@felixdesouza I suggested this specifically so it lets us verify HOW does the query string change and if the changes are acceptable. Without this the actual diff is kinda hidden, all you see is that a bunch of methods on the builder have been renamed.

felixdesouza:
Sure, I suppose it felt more like we could do this offline as part of doing the change and just put the query strings before and after as part of the PR description.

jkozlowski:
Ok, that works for me.

jeremyk-91:
Is this PR still needed, in the light of #4376?

OStevan:
This PR is no longer relevant as pointed out by @jeremyk-91 therefor it is closed.

PR #4312, opened 2019-10-17
[WIP] Cleanup metrics.

jkozlowski:
**Goals (and why)**:

Cleanup metrics.

The below will be explaining how those metrics scale per TransactionManager (so if an app uses multiple TransactionManagers, the number of series should be multipled by that):

Timer: 4 series (it's actually 6, but we usually filter out 2)
Histogram: dunno if we filter any, but we log 5 series
Meter: 2 series, again not sure about filtering

### Per table metrics
* com.palantir.atlasdb.transaction.impl.SnapshotTransaction.transactionsReadFromDB: histogram
* com.palantir.atlasdb.transaction.impl.SnapshotTransaction.emptyValuesCellFilterCount: meter
* com.palantir.atlasdb.transaction.impl.SnapshotTransaction.numCellsRead: meter

### Not interesting metrics
* ~CoordinationStore: not a useful set of metrics: 3 methods x 4 series each = 12 series~

### Expensive metrics

* com.palantir.atlasdb.sweep.metrics.TargetedSweepMetrics.outcome: this one explodes quite badly, because it registers about 10 metrics per Transaction manager: 2 Strategies ("thorough" and "conservative") x 5 SweepOutcome values = 10 series
* com.palantir.atlasdb.sweep.BackgroundSweeperImpl.outcome: one per SweepOutcomes value = 9 series

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/CoordinationServices.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/CoordinationServices.java
@@ -50,7 +50,7 @@ public final class CoordinationServices {
                 MetricsManager metricsManager,
                 boolean initializeAsync) {
         CoordinationService<VersionedInternalSchemaMetadata> versionedService = new CoordinationServiceImpl<>(
-                createCoordinationStore(keyValueService, timestampSupplier, metricsManager, initializeAsync));
+                createCoordinationStore(keyValueService, timestampSupplier, initializeAsync));
 
         @SuppressWarnings("unchecked") // The service has the same type as the version-hiding service.
         CoordinationService<InternalSchemaMetadata> instrumentedService = AtlasDbMetrics.instrument(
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/internalschema/persistence/CoordinationServices.java
```diff
@@ -63,17 +63,8 @@ private CoordinationServices() {
...
-        CoordinationStore<VersionedInternalSchemaMetadata> rawCoordinationStore
-                = createRawCoordinationStore(keyValueService, timestampSupplier, initializeAsync);
-
-        @SuppressWarnings("unchecked") // The store has the same type as the rawCoordinationStore.
-        CoordinationStore<VersionedInternalSchemaMetadata> store = AtlasDbMetrics.instrument(
-                metricsManager.getRegistry(),
-                CoordinationStore.class,
-                rawCoordinationStore);
-        return store;
+        return createRawCoordinationStore(keyValueService, timestampSupplier, initializeAsync);
```
felixdesouza: can you just inline this method? looks the same

PR #4313, opened 2019-10-17
[WIP]Async get complete call path async

OStevan:
**Goals (and why)**:
In the previous PR certain paths were still calling the synchronous call paths. This PR address that fact.

- existing tests should cover this

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -744,7 +744,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 MoreExecutors.directExecutor());
     }
 
-    private interface CellLoader {
+    protected interface CellLoader {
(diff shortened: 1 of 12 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -854,7 +856,7 @@ private boolean isThoroughlySwept(TableReference tableRef) {
     }
 
     private List<Map.Entry<Cell, byte[]>> getPostFilteredWithLocalWrites(final TableReference tableRef,
-                                                                     final SortedMap<Cell, byte[]> postFiltered,
```
jeremyk-91: I believe it is essential that this is a `SortedMap`. We should discuss offline

jeremyk-91: Or, at least, that entries provided by the iterator here are in sorted order. I get that we haven't changed the concrete implementation, but would prefer to codify this guarantee in the type system.

PR #4316, opened 2019-10-17
Refactored CQL query specs.

OStevan:
**Goals (and why)**:
Make the `QuerySpec` less bloated with simple refactoring. It should also ease adding new specs.

- class is split into three more manageable classes/interfaces

- tests were updated to reflect this change

- `CqlQueryContext`

whenever

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -100,6 +100,8 @@ import com.palantir.atlasdb.keyvalue.api.Value;
 import com.palantir.atlasdb.keyvalue.cassandra.CassandraKeyValueServices.StartTsResultsCollector;
 import com.palantir.atlasdb.keyvalue.cassandra.async.AsyncCellLoader;
 import com.palantir.atlasdb.keyvalue.cassandra.async.CqlClient;
+import com.palantir.atlasdb.keyvalue.cassandra.async.futures.CqlFuturesCombiner;
+import com.palantir.atlasdb.keyvalue.cassandra.async.futures.DefaultCqlFuturesCombiner;
 import com.palantir.atlasdb.keyvalue.cassandra.cas.CheckAndSetRunner;
 import com.palantir.atlasdb.keyvalue.cassandra.paging.RowGetter;
 import com.palantir.atlasdb.keyvalue.cassandra.sweep.CandidateRowForSweeping;
@@ -221,6 +223,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     private final AsyncCellLoader asyncCellLoader;
     private final RangeLoader rangeLoader;
     private final TaskRunner taskRunner;
+    private final CqlFuturesCombiner cqlFuturesCombiner;
     private final CellValuePutter cellValuePutter;
     private final CassandraTableMetadata tableMetadata;
     private final CassandraTableCreator cassandraTableCreator;
@@ -428,15 +431,13 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
         this.wrappingQueryRunner = new WrappingQueryRunner(queryRunner);
         this.cassandraTables = new CassandraTables(clientPool, config);
         this.taskRunner = new TaskRunner(executor);
+        this.cqlFuturesCombiner = new DefaultCqlFuturesCombiner(createInstrumentedFixedThreadPool(
+                config,
+                metricsManager.getRegistry(),
+                "Atlas Cassandra KVS async",
+                "asyncExecutorService"));
         this.cellLoader = CellLoader.create(clientPool, wrappingQueryRunner, taskRunner, runtimeConfigSupplier);
-        this.asyncCellLoader = AsyncCellLoader.create(
-                cqlClient,
-                createInstrumentedFixedThreadPool(
-                        config,
-                        metricsManager.getRegistry(),
-                        "Atlas Cassandra KVS async",
-                        "asyncExecutorService"),
-                config.getKeyspaceOrThrow());
+        this.asyncCellLoader = AsyncCellLoader.create(cqlClient, cqlFuturesCombiner, config.getKeyspaceOrThrow());
         this.rangeLoader = new RangeLoader(clientPool, queryRunner, metricsManager, readConsistency);
         this.cellValuePutter = new CellValuePutter(
                 config,
@@ -1684,7 +1685,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     @Override
     public void close() {
         clientPool.shutdown();
-        asyncCellLoader.close();
+        cqlFuturesCombiner.close();
         try {
             log.info("Trying to close a CQL client");
             cqlClient.close();
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -413,7 +413,11 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             Supplier<CassandraKeyValueServiceRuntimeConfig> runtimeConfigSupplier,
             CassandraClientPool clientPool,
             CassandraMutationTimestampProvider mutationTimestampProvider) {
-        super(createInstrumentedFixedThreadPool(config, metricsManager.getRegistry()));
+        super(createInstrumentedFixedThreadPool(
+                config,
+                metricsManager.getRegistry(),
+                "Atlas Cassandra KVS",
+                "executorService"));
         this.log = log;
         this.metricsManager = metricsManager;
         this.config = config;
@@ -425,7 +429,14 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
         this.cassandraTables = new CassandraTables(clientPool, config);
         this.taskRunner = new TaskRunner(executor);
         this.cellLoader = CellLoader.create(clientPool, wrappingQueryRunner, taskRunner, runtimeConfigSupplier);
-        this.asyncCellLoader = AsyncCellLoader.create(cqlClient, executor, config.getKeyspaceOrThrow());
+        this.asyncCellLoader = AsyncCellLoader.create(
+                cqlClient,
+                createInstrumentedFixedThreadPool(
+                        config,
+                        metricsManager.getRegistry(),
+                        "Atlas Cassandra KVS async",
+                        "asyncExecutorService"),
+                config.getKeyspaceOrThrow());
         this.rangeLoader = new RangeLoader(clientPool, queryRunner, metricsManager, readConsistency);
         this.cellValuePutter = new CellValuePutter(
                 config,
@@ -444,12 +455,14 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
 
     private static ExecutorService createInstrumentedFixedThreadPool(
             CassandraKeyValueServiceConfig config,
-            MetricRegistry registry) {
+            MetricRegistry registry,
+            String threadNamePrefix,
+            String executorServiceMetricsPrefix) {
         return Tracers.wrap(new InstrumentedExecutorService(
-                createFixedThreadPool("Atlas Cassandra KVS",
+                createFixedThreadPool(threadNamePrefix,
                         config.poolSize() * config.servers().numberOfThriftHosts()),
                 registry,
-                MetricRegistry.name(CassandraKeyValueService.class, "executorService")));
+                MetricRegistry.name(CassandraKeyValueService.class, executorServiceMetricsPrefix)));
     }
 
     @Override
@@ -1671,6 +1684,7 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     @Override
     public void close() {
         clientPool.shutdown();
+        asyncCellLoader.close();
         try {
             log.info("Trying to close a CQL client");
             cqlClient.close();
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -298,7 +298,8 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             CassandraKeyValueServiceConfig config,
             CassandraMutationTimestampProvider mutationTimestampProvider,
             Logger log) {
-        return create(metricsManager,
+        return create(
+                metricsManager,
                 config,
                 CassandraKeyValueServiceRuntimeConfig::getDefault,
                 mutationTimestampProvider,
@@ -313,7 +314,8 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             CassandraMutationTimestampProvider mutationTimestampProvider,
             Logger log,
             boolean initializeAsync) {
-        CassandraClientPool clientPool = CassandraClientPoolImpl.create(metricsManager,
+        CassandraClientPool clientPool = CassandraClientPoolImpl.create(
+                metricsManager,
                 config,
                 runtimeConfigSupplier,
                 initializeAsync);
@@ -417,10 +419,9 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
             CassandraClientPool clientPool,
             CassandraMutationTimestampProvider mutationTimestampProvider) {
         super(createInstrumentedFixedThreadPool(
-                config,
-                metricsManager.getRegistry(),
-                "Atlas Cassandra KVS",
-                "executorService"));
+                config.poolSize() * config.servers().numberOfThriftHosts(),
+                metricsManager.getRegistry()
+        ));
         this.log = log;
         this.metricsManager = metricsManager;
         this.config = config;
@@ -431,8 +432,9 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
         this.wrappingQueryRunner = new WrappingQueryRunner(queryRunner);
         this.cassandraTables = new CassandraTables(clientPool, config);
         this.taskRunner = new TaskRunner(executor);
-        this.cqlFuturesCombiner = new DefaultCqlFuturesCombiner(createInstrumentedFixedThreadPool(
-                config,
+        this.cqlFuturesCombiner = new DefaultCqlFuturesCombiner(createInstrumentedDynamicThreadPool(
+                0,
+                config.poolSize() * config.servers().numberOfThriftHosts(),
                 metricsManager.getRegistry(),
                 "Atlas Cassandra KVS async",
                 "asyncExecutorService"));
@@ -455,15 +457,27 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
     }
 
     private static ExecutorService createInstrumentedFixedThreadPool(
-            CassandraKeyValueServiceConfig config,
+            int corePoolSize,
+            MetricRegistry registry) {
+        return createInstrumentedDynamicThreadPool(
+                corePoolSize,
+                corePoolSize,
+                registry,
+                "Atlas Cassandra KVS",
+                "executorService");
+    }
+
+    private static ExecutorService createInstrumentedDynamicThreadPool(
+            int corePoolSize,
+            int maxPoolSize,
             MetricRegistry registry,
             String threadNamePrefix,
             String executorServiceMetricsPrefix) {
-        return Tracers.wrap(new InstrumentedExecutorService(
-                createFixedThreadPool(threadNamePrefix,
-                        config.poolSize() * config.servers().numberOfThriftHosts()),
-                registry,
-                MetricRegistry.name(CassandraKeyValueService.class, executorServiceMetricsPrefix)));
+        return Tracers.wrap(
+                new InstrumentedExecutorService(
+                        createThreadPool(threadNamePrefix, corePoolSize, maxPoolSize),
+                        registry,
+                        MetricRegistry.name(CassandraKeyValueService.class, executorServiceMetricsPrefix)));
     }
 
     @Override
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
@@ -28,10 +28,10 @@ import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.keyvalue.api.Value;
 import com.palantir.atlasdb.keyvalue.cassandra.async.futures.CqlFuturesCombiner;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.CqlQueryContext;
+import com.palantir.atlasdb.keyvalue.cassandra.async.queries.GetQuerySpec;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.GetQuerySpec.GetQueryParameters;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.ImmutableCqlQueryContext;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.ImmutableGetQueryParameters;
-import com.palantir.atlasdb.keyvalue.cassandra.async.queries.ImmutableGetQuerySpec;
 import com.palantir.atlasdb.logging.LoggingArgs;
 import com.palantir.common.streams.KeyedStream;
 import com.palantir.logsafe.SafeArg;
@@ -67,7 +67,7 @@ public final class AsyncCellLoader {
                 .map((cell, timestamp) -> loadCellWithTimestamp(tableReference, cell, timestamp))
                 .collectToMap();
 
-        return cqlFuturesCombiner.combineToMap(cellListenableFutureMap);
+        return cqlFuturesCombiner.allAsMap(cellListenableFutureMap);
     }
 
     private ListenableFuture<Optional<Value>> loadCellWithTimestamp(
@@ -83,10 +83,6 @@ public final class AsyncCellLoader {
                 .humanReadableTimestamp(timestamp)
                 .build();
 
-        return cqlClient.executeQuery(
-                ImmutableGetQuerySpec.builder()
-                        .cqlQueryContext(queryContext)
-                        .queryParameters(getQueryParameters)
-                        .build());
+        return cqlClient.executeQuery(new GetQuerySpec(queryContext, getQueryParameters));
     }
 }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
@@ -18,18 +18,15 @@ package com.palantir.atlasdb.keyvalue.cassandra.async;
 
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.keyvalue.api.Value;
+import com.palantir.atlasdb.keyvalue.cassandra.async.futures.CqlFuturesCombiner;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.CqlQueryContext;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.GetQuerySpec.GetQueryParameters;
 import com.palantir.atlasdb.keyvalue.cassandra.async.queries.ImmutableCqlQueryContext;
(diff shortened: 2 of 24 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -47,40 +48,46 @@
...
-    @Override
-    public Statement makeExecutableStatement(PreparedStatement preparedStatement) {
-        return preparedStatement.bind()
-                .setBytes("row", row())
-                .setBytes("column", column())
-                .setLong("timestamp", queryTimestamp());
-    }
+    @org.immutables.value.Value.Immutable
+    public interface GetQueryParameters {
+        ByteBuffer row();
```
jkozlowski: How about com.palantir.conjure.java.lib.Bytes? It has a zero-cost converter to a read-only ByteBuffer

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlQueryContext.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.concurrent.Executor;
+
+import org.immutables.value.Value;
+
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+
+@Value.Immutable
+public interface CqlQueryContext {
+    String keySpace();
```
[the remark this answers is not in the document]
jkozlowski: I guess you'd need to do one extra copy to create it.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
```diff
@@ -82,12 +84,15 @@ private AsyncCellLoader(CqlClient cqlClient, Executor executor, String keyspace)
...
-                        .row(ByteBuffer.wrap(cell.getRowName()))
-                        .column(ByteBuffer.wrap(cell.getColumnName()))
-                        .humanReadableTimestamp(timestamp)
-                        .executor(executor)
+                        .cqlQueryContext(ImmutableCqlQueryContext.builder()
+                                .tableReference(tableRef)
+                                .keySpace(keyspace)
+                                .executor(executor).build())
+                        .queryParameters(ImmutableGetQueryParameters.builder()
+                                .row(ByteBuffer.wrap(cell.getRowName()))
```
jkozlowski: .asReadOnlyBuffer?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlQueryContext.java
```diff
@@ -0,0 +1,33 @@
...
+
+    TableReference tableReference();
+
+    @Value.Auxiliary
+    Executor executor();
```
jkozlowski: Why does the query context know about the executor?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -47,40 +48,46 @@
             + "ORDER BY column1, column2 ASC "
             + "LIMIT 1;";
 
-    @org.immutables.value.Value.Auxiliary
-    public abstract ByteBuffer row();
-
-    @org.immutables.value.Value.Auxiliary
-    public abstract ByteBuffer column();
+    @Override
+    public String formatQueryString() {
```
jkozlowski: In the format string, do you need the ORDER BY? it's a limit 1 anyway, and also it's ordering by things you're not even selecting.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -47,40 +48,46 @@
...
-    private long queryTimestamp() {
-        return ~humanReadableTimestamp();
+        long humanReadableTimestamp();
+
+        default long queryTimestamp() {
+            return ~humanReadableTimestamp();
+        }
     }
 
     @ThreadSafe
```
jkozlowski: In #accumulateRowStream, you are verifying that the resultValue is not set, and then mapping anyway. You should just assign it. Better still, I would have written the query specs for single row things first, and if we ever decide to move more queries, we can refactor and introduce that capability.

We could then remove all the paging and page iteration code there, or keep it but specialize it for a single page.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientImpl.java
```diff
@@ -124,7 +125,10 @@ public void close() {
...
                 .setConsistencyLevel(querySpec.queryConsistency());
 
-        return execute(executableStatement, querySpec.executor(), querySpec.rowStreamAccumulatorFactory().get());
+        return execute(
+                executableStatement,
+                querySpec.cqlQueryContext().executor(),
+                querySpec.rowStreamAccumulatorFactory().get());
     }
 
     private <V> ListenableFuture<V> execute(
```
jkozlowski: So how about this for a threading model: 

1. We do the iteration and accumulation on the Cassandra threads - this code should be super simple and super fast, expecially if we specialize to LIMIT 1 queries for now.
2. (maybe, but might not be worth it to start with) Remove all the synchronization and threadsafety stuff from RowStreamAccumulator implementations: for this to work we simply need to double check that the driver only ever fetches pages serially and only ever returns control serially. Then just make sure that memory visibility is kept.
3. Then the only thing that happens on the special threadpool is AsyncCellLoader's Futures.whenAllSucceed: and memory visibility should be there nicely, because all the futures implementations we use are threadsafe so everything should be propagated correctly.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
```diff
@@ -82,12 +84,15 @@ private AsyncCellLoader(CqlClient cqlClient, Executor executor, String keyspace)
             long timestamp) {
         return cqlClient.executeQuery(
```
jkozlowski: Can we add some unit tests for this class in general?

jkozlowski: In fact I see tests only for QueryCache, so we very likely should have more.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
```diff
@@ -82,12 +84,15 @@ private AsyncCellLoader(CqlClient cqlClient, Executor executor, String keyspace)
...
-                        .executor(executor)
+                        .cqlQueryContext(ImmutableCqlQueryContext.builder()
+                                .tableReference(tableRef)
+                                .keySpace(keyspace)
+                                .executor(executor).build())
+                        .queryParameters(ImmutableGetQueryParameters.builder()
+                                .row(ByteBuffer.wrap(cell.getRowName()))
+                                .column(ByteBuffer.wrap(cell.getColumnName()))
+                                .humanReadableTimestamp(timestamp)
+                                .build())
```
jkozlowski: AsyncCellLoader#getDone is a bit suspicious, I wouldn't think Cass would throw any checked exceptions in this codepath.

jkozlowski:
Also, CassandraKeyValueServiceImpl: when we're instantiating AsyncCellLoader, we pass a fixed size pool that is also used for running blocking Cass queries. I'd rather we pass a separate one there.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: The `ORDER BY` was added more for documenting purposes, the data is already stored in C* by the order specified there so this ORDER BY to actually be a NO OP during execution of the query. I will remove it.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: This library isn't used anywhere else and is not a project dependency. Can I add it? @felixdesouza @jeremyk-91

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlQueryContext.java
OStevan: The way I imagined it is that the `CqlClientImpl` will be responsible just for executing the prepared queries on correct session. While the `AsyncCellLoader` (or some other class for future queries) will know what to do with results (how to combine the results and on which pool, etc.).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: The `RowStreamAccumulator` does not force you to process pages, multiple rows etc. If you know that you'll get all the data in the first call of `accumulateRowStream` you can do whatever you want in the impl. That is why I intended `GetQueryAccumulator` to be a private internal class of GetQuerySpec, it is an implementation detail based on the query we wrote so we can make it as specific to the query as we want. We can discuss in person if this is not a right way to organize code.

I agree we can specialize the code more without paging etc. I think it makes more sense to do it after I switch to the new driver version since it has a built in support for async.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
OStevan: This is a recommended way from Guava docs to work with `Futures#whenAllSucceed.call(...)`.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientImpl.java
OStevan: 1. Iteration and accumulation have been resolved by a commit from @jkozlowski 

2. Removed synchronization and thread safety. Each page of a query is submitted for processing only after the current page is processed and since an accumulator is on a per query basis there is not possibility for concurrent access of two threads to the same accumulator. 

3. However we have no guarantees about the visibility of the results in the `AsyncCellLoder` collection phase so internal fields of `GetQueryAccumulator` are `volatile`.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
OStevan: Added unit tests for AsyncCell loader. They are based on around weather visibility is maintained. Integration tests for CKVS should cover if the `getAsync` behaves correctly.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlQueryContext.java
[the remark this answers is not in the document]
OStevan: Validation is added to `CqlQueryContext`, for me it makes sense to validate that the context is in fact formed correctly.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
felixdesouza: why not?

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/async/QueryCacheTest.java
```diff
@@ -49,53 +49,37 @@ public void setUp() {
...
                 .isEqualTo(expectedPreparedStatement);
     }
 
     @Test
-    public void testGetQuerySpecRowIgnored() {
-        PreparedStatement expectedPreparedStatement = prepareStatement(KEYSPACE, TABLE_REFERENCE, 10, 10, 3);
-        assertThat(prepareStatement(KEYSPACE, TABLE_REFERENCE, 4, 10, 3))
+    public void testGetQuerySpecParametersIgnored() {
+        PreparedStatement expectedPreparedStatement = prepareStatement(KEYSPACE, TABLE_REFERENCE, GET_QUERY_PARAMETERS);
+        assertThat(prepareStatement(KEYSPACE, TABLE_REFERENCE, mock(GetQueryParameters.class)))
```
felixdesouza: please do not mock value objects! just make them!

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -38,67 +36,74 @@
...
+            // the query can only ever return one page with one row
+            assigned = true;
+            resultValue = rowStream.findFirst().map(GetQueryAccumulator::parseValue).orElse(null);
         }
 
         @Override
-        public synchronized Optional<Value> result() {
-            return resultValue;
+        public Optional<Value> result() {
+            return Optional.ofNullable(resultValue);
```
felixdesouza: do we want to throw here if we get without it being assigned?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -38,67 +36,74 @@
 
     /**
      * Since each query is constructed for one cell we are using an optimisation that we can ask the CQL to do most of
-     * the work internally. First of all we are using the fact that timestamps in {@code column2} are ordered in
-     * {@code ASC} order and since we are interested in the most recent timestamp we use {@code LIMIT 1} to get the
-     * latest value. This should help with both cassandra workload and amount of transferred data.
+     * the work internally. First of all we are using the fact that cells are ordered by {@code column} and
+     * {@code timestamp} value in {@code ASC} order and since we are interested in the value with the most recent
```
felixdesouza: might the mention of us storing timestamps as negative timestamps make this make a bit more sense? @jeremyk-91

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlQueryContext.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Value.Immutable
+public interface CqlQueryContext {
+    String keyspace();
+
+    TableReference tableReference();
+
+    @Value.Check
+    default void check() {
+        Preconditions.checkState(!keyspace().contains("\""), "Keyspace should not be quoted");
```
felixdesouza: seems like a fairly odd place to put this check, surely this would happen a lot further up? closer to the config?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
felixdesouza: unless the builder can be done in one line, please refrain from nesting builders.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
felixdesouza: + if you have it like this, you force consumers to know how to serialise things into a form you should know about.

Either make this contain `Cell` and you convert when you are making the executable statement or  you add a builder method here e.g.

```java
public interface GetQueryParameters {
  ...
  public static class Builder extends ImmutableGetQueryParameters.Builder {
    public Builder cell(Cell cell) {
      return this.row(...);
          .column(...);
    }
  } 
}

File: atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java [hunk 14]
```diff
@@ -38,67 +36,74 @@
...
+    public Supplier<RowStreamAccumulator<Optional<Value>>> rowStreamAccumulatorFactory() {
         return GetQueryAccumulator::new;
     }
 
-    @Override
-    public String formatQueryString() {
-        return String.format(QUERY_FORMAT, keySpace(), AbstractKeyValueService.internalTableName(tableReference()));
-    }
+    @org.immutables.value.Value.Auxiliary
+    public abstract GetQueryParameters getQueryParameters();
```
felixdesouza: I don't think these need to be auxiliary anymore? since you have a context key or whatever?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -38,67 +36,74 @@
...
+    }
 
     @Override
     public ConsistencyLevel queryConsistency() {
         return ConsistencyLevel.LOCAL_QUORUM;
     }
 
     @Override
-    public final Supplier<RowStreamAccumulator<Optional<Value>>> rowStreamAccumulatorFactory() {
+    public Supplier<RowStreamAccumulator<Optional<Value>>> rowStreamAccumulatorFactory() {
```
felixdesouza: as an aside, why is this a `Supplier`? `GetQuerySpec` is not long-lived, you make one for every cell, so you might as well just return a `RowStreamAccumulator` directly.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoaderTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+        Map<Cell, Long> request = ImmutableMap.of(NON_VISIBLE_CELL, TIMESTAMP);
+        AsyncCellLoader asyncCellLoader = AsyncCellLoader.create(cqlClient, testExecutor, DEFAULT_KEYSPACE);
+        Map<Cell, Value> result = asyncCellLoader.loadAllWithTimestamp(DEFAULT_TABLE, request).get();
+        assertThat(result.isEmpty()).isTrue();
+    }
+
+    @Test
+    public void testFilteringNonVisible() throws Exception {
+        when(cqlClient.executeQuery(any()))
```
felixdesouza: I don't really understand why we're relying on a particular ordering of execution?

It should be as easy as just making a get query no?

i.e. I want to be able to see/say

if I get a query of this form, return this etc.

with this, I'm relying on the implementation firing off requests in a particular order

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
felixdesouza: +1 to what @OStevan said, if we didn't, we'd also have to then deal with interrupted exception etc. less code 💃 , although part of me is tempted to do a `getDoneUnchecked` but it's probably a bad idea

jeremyk-91: nit: I'm not sure where in the docs you're linking to - [whenAllSucceed](https://guava.dev/releases/20.0/api/docs/com/google/common/util/concurrent/Futures.html#whenAllSucceed-java.lang.Iterable-) or [FutureCombiner](https://guava.dev/releases/20.0/api/docs/com/google/common/util/concurrent/Futures.FutureCombiner.html#call-java.util.concurrent.Callable-), or something else.

I guess this is a refactor so fine, but the original seems a bit questionable in that it would wrap unchecked exceptions in another `RuntimeException` as well

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
jeremyk-91: Yep. I'm not sure why the `column` (presumably `column1`) ordering matters here since it's a single SELECT query? The important bit is that `column2` in the clustering key is ASC, and it is the timestamps stored in reverse bit order, so we can get the value with the largest real timestamp with a LIMIT 1.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
jeremyk-91: For completeness, data is returned in order of clustering key given a partition key is provided, so yeah `ORDER BY` isn't needed. (Note that this query if interpreted as SQL for another databases would likely require the ORDER BY.)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
```diff
@@ -38,67 +36,74 @@
...
+        }
     }
 
-    @ThreadSafe
-    public static class GetQueryAccumulator implements RowStreamAccumulator<Optional<Value>> {
+    private static class GetQueryAccumulator implements RowStreamAccumulator<Optional<Value>> {
 
-        private Optional<Value> resultValue = Optional.empty();
+        private volatile Value resultValue = null;
+        private volatile boolean assigned = false;
```
jeremyk-91: Non-actionable: Depending on usage this may not need to be volatile (if it's guaranteed that calls to `accumulateRowStream` aren't concurrent, even if it may accidentally be called more than once), which could lead to probably a very very minor perf improvement. Don't do this unless we see that it somehow becomes a problem

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: I'm using `Cell` as a parameter and I prepare everything internally.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoaderTests.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final Cell NON_VISIBLE_CELL = Cell.create(PtBytes.toBytes(100), PtBytes.toBytes(100));
+    private static final Cell VISIBLE_CELL_1 = Cell.create(PtBytes.toBytes(100), PtBytes.toBytes(200));
+    private static final Cell VISIBLE_CELL_2 = Cell.create(PtBytes.toBytes(100), PtBytes.toBytes(300));
+
+    private CqlClient cqlClient;
+    private final ExecutorService testExecutor = MoreExecutors.newDirectExecutorService();
+
+    @Before
+    public void setUp() {
+        cqlClient = mock(CqlClient.class);
```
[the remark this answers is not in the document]
OStevan: Shut down.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoaderTests.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Test
+    public void testNoDataVisible() throws Exception {
+        when(cqlClient.executeQuery(any()))
+                .thenReturn(Futures.immediateFuture(Optional.empty()));
+
+        Map<Cell, Long> request = ImmutableMap.of(NON_VISIBLE_CELL, TIMESTAMP);
+        AsyncCellLoader asyncCellLoader = AsyncCellLoader.create(cqlClient, testExecutor, DEFAULT_KEYSPACE);
```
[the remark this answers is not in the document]
OStevan: Moved.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: Added a clarifying comment about how timestamps are stored.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: Removed.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/query/GetQuerySpec.java
OStevan: Agreed.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoaderTests.java
OStevan: Tests are fixed so we don't rely on the order.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -425,7 +429,14 @@ private CassandraKeyValueServiceImpl(
         this.cassandraTables = new CassandraTables(clientPool, config);
         this.taskRunner = new TaskRunner(executor);
         this.cellLoader = CellLoader.create(clientPool, wrappingQueryRunner, taskRunner, runtimeConfigSupplier);
-        this.asyncCellLoader = AsyncCellLoader.create(cqlClient, executor, config.getKeyspaceOrThrow());
+        this.asyncCellLoader = AsyncCellLoader.create(
+                cqlClient,
+                createInstrumentedFixedThreadPool(
```
jkozlowski: I feel like you should inject this threadpool, when an internal product instantiates a lot of these it will probably want to reuse the threadpool

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
```diff
@@ -73,29 +76,39 @@ private AsyncCellLoader(CqlClient cqlClient, Executor executor, String keyspace)
...
         try {
             return Futures.getDone(listenableFuture);
         } catch (ExecutionException e) {
             throw new RuntimeException(e.getCause());
         }
     }
+
+    @Override
+    public void close() {
+        executorService.shutdown();
```
jkozlowski: Something feels weird here, why is it closing the ExecutorService and not the client?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientImpl.java
jkozlowski: 3. There should be guarantee, because futures are synchronized and thread-safe by default. So there should be a happens-before for any stuff that happens.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
```diff
@@ -31,25 +30,29 @@
...
     private final String keyspace;
 
-    public static AsyncCellLoader create(CqlClient cqlClient, Executor executor, String keyspace) {
-        return new AsyncCellLoader(cqlClient, executor, keyspace);
+    public static AsyncCellLoader create(CqlClient cqlClient, ExecutorService executorService, String keyspace) {
+        return new AsyncCellLoader(cqlClient, executorService, keyspace);
     }
 
-    private AsyncCellLoader(CqlClient cqlClient, Executor executor, String keyspace) {
+    private AsyncCellLoader(CqlClient cqlClient, ExecutorService executorService, String keyspace) {
```
jkozlowski: idea: what if you don't pass ExecutorService here, but something that only exposes whatever transformation logic you need for futures. Then make CqlContext that contains CqlClient and whatever you call this new interface, and have CqlContext be AutoCloseable. Then just have one #close call.

OStevan: I now have a bit of a different interface which should in fact abstract  away how futures are combined and on which thread.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/AsyncCellLoader.java
OStevan: Closing is now more centralized.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
OStevan: Discussed for now and this solution is fine for now.

jkozlowski: It's fine *IF* we make sure we don't create the threadpool the same way the other one is created: with corePoolSize == maxPoolSize. Let's create with corePoolSize == 0

OStevan: Resolved.

PR #4319, opened 2019-10-18
Return and use `ManagedTimestampService` everywhere.

felixdesouza:
**Goals (and why)**:
We introduced `ManagedTimestampService` which unified `TimestampService` and `TimestampManagementService`. At some point in `TransactionManagers` we returned that type wherever we wanted a `TimestampService` etc.

At some point we changed the internal hierarchy for timelock and what interfaces it exposed, as a result the underlying type of `TimestampService` is not `ManagedTimestampService` so cannot be cast. This causes migration CLIs to fail (not caught since we don't run these often).

Only expose `ManagedTimestampService` so no matter how we wrap individual services, we will still combine them and return that.

A mechanical refactor, don't think there's existing testing.

ASAP


atlasdb-cli/src/test/java/com/palantir/atlasdb/cli/command/KeyValueServiceMigratorsTest.java
```diff
@@ -312,22 +313,22 @@ public void atomicTablesDelegatedToSourceAreNotDropped() {
         assertThat(toSplittingServices.getTransactionService().get(100_000)).isEqualTo(100_001L);
     }
 
-    private AtlasDbServices createMock(KeyValueService kvs) {
-        TimestampService timestampService = new InMemoryTimestampService();
+    private static AtlasDbServices createMock(KeyValueService kvs) {
```
jeremyk-91: 💯

PR #4322, opened 2019-10-21
Add Timed annotation and remove instrumentation from some TransactionManager methods

jkozlowski:
**Goals (and why)**:

AtlasDBMetrics times all the methods on interfaces, which a lot of the time includes things that are not useful being timed. This PR adds another codepath there where only annotated methods will be timed. This PR just moves TransactionManager to that codepath and I will send PRs for other interfaces later, to keep this PR small.

TransactionManager methods:
* isInitialized: not timed
* runTaskWithRetry/runTaskThrowOnConflict/runTaskReadOnly/runTaskWithLocksWithRetry/runTaskWithLocksWithRetry/runTaskWithLocksWithRetry/runTaskWithLocksWithRetry/runTaskWithLocksThrowOnConflict/runTaskWithConditionWithRetry/runTaskWithConditionWithRetry/runTaskWithConditionThrowOnConflict/runTaskWithConditionReadOnly/: timed, but I would like to get rid of those.
* getImmutableTimestamp: not timed, it just delegates and we should have metrics for that on TimestampService.
* getLockService/getTimelockService/getTimestampService/getTimestampManagementService/getTransactionService/getCleaner/getKeyValueService: not timed, in memory instance variable access
* getKeyValueServiceStatus: not timed, we have metrics at KVS layer for individual calls.
* getTimelockServiceStatus: just looks at timelock metrics and returns a value, so not useful.
* getUnreadableTimestamp: timed until we have a more specific metric for this in the Puncher store
* clearTimestampCache: not timed, this is probably just timing guava/caffeine
* registerClosingCallback: not timed
* setupRunTaskWithConditionThrowOnConflict/finishRunTaskWithLockThrowOnConflict: internal product relies on that, but might be able to move off those metrics.
* close: not timed

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/Timed.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/Timed.java
@@ -22,7 +22,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ElementType.METHOD})
+@Target(ElementType.METHOD)
 public @interface Timed {
 }
 
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/Timed.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/Timed.java
@@ -1,0 +1,28 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.metrics;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface Timed {
+}
+
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
@@ -366,7 +366,6 @@ public interface TransactionManager extends AutoCloseable {
      * <p>
      * This call must be implemented so that it completes synchronously.
      */
-    // There are metrics for the individual call
     KeyValueServiceStatus getKeyValueServiceStatus();
 
     /**
@@ -377,7 +376,6 @@ public interface TransactionManager extends AutoCloseable {
      *
      * @return status of the timelock service
      */
-    // Just looks at in-memory metrics
     TimelockServiceStatus getTimelockServiceStatus();
 
     /**
@@ -394,14 +392,12 @@ public interface TransactionManager extends AutoCloseable {
      *
      * @throws IllegalStateException if the transaction manager has been closed.
      */
-    // I would say we don't really care about this? If we want this metric it should be in the puncher store.
     long getUnreadableTimestamp();
 
     /**
      * Clear the timestamp cache. This is mostly useful for tests that perform operations that would invalidate
      * the cache, although this can also be used to free up some memory.
      */
-    //
     void clearTimestampCache();
 
     /**
@@ -424,6 +420,7 @@ public interface TransactionManager extends AutoCloseable {
      * @return the transaction and associated immutable timestamp lock for the task
      */
     @Deprecated
+    @Timed
     TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
 
     /**
@@ -436,6 +433,7 @@ public interface TransactionManager extends AutoCloseable {
      * @return value returned by the task
      */
     @Deprecated
+    @Timed
     <T, E extends Exception> T finishRunTaskWithLockThrowOnConflict(TransactionAndImmutableTsLock tx,
             TransactionTask<T, E> task)
             throws E, TransactionFailedRetriableException;
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
@@ -392,6 +392,7 @@ public interface TransactionManager extends AutoCloseable {
      *
      * @throws IllegalStateException if the transaction manager has been closed.
      */
+    @Timed
     long getUnreadableTimestamp();
 
     /**
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
@@ -19,6 +19,7 @@ import java.util.function.Supplier;
 
 import com.palantir.atlasdb.cleaner.api.Cleaner;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.metrics.Timed;
 import com.palantir.atlasdb.transaction.service.TransactionService;
 import com.palantir.exception.NotInitializedException;
 import com.palantir.lock.HeldLocksToken;
@@ -77,6 +78,7 @@ public interface TransactionManager extends AutoCloseable {
      *
      * @throws IllegalStateException if the transaction manager has been closed.
      */
+    @Timed
     <T, E extends Exception> T runTaskWithRetry(TransactionTask<T, E> task) throws E;
 
     /**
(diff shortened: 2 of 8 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -379,12 +394,14 @@ default boolean isInitialized() {
      *
      * @throws IllegalStateException if the transaction manager has been closed.
      */
+    // I would say we don't really care about this? If we want this metric it should be in the puncher store.
```
jeremyk-91: I think there was an issue a while back where this actually turned out to be helpful (technically this could end up doing arbitrarily large range scans on the `_punch` table), so we may want to keep this in some form - though the internal database dashboard should be enough to identify this based on range scan metrics. @gmaretic what do you think?

felixdesouza:
It would be interesting to know how the annotation plays with @palantir/tritium metric proxying. If anything, I'd be interested to know if tritium supports something along these lines.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
jkozlowski: My point was that this just delegates to the puncher, so probably should have it there.

jkozlowski: re: whitelist vs blacklist: if someone doesn't think about it from the start, probably we don't need this metric. I will do a pass over the other ones and eventually deprecate the one without filtering, so you will always have @Timed on some method in the interface, I think that's enough of a ping.

jkozlowski:
@felixdesouza agreed, we can probably push something into Tritium, but I would like to start with trying it out here. Plus as you can see there's already some useful hooks to implement this.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
```diff
@@ -79,4 +84,28 @@ public static void registerCache(MetricRegistry metricRegistry, Cache<?, ?> cach
...
     }
+
+    private static <T, U extends T> T instrument(
+            MetricRegistry metricRegistry,
+            Class<T> serviceInterface,
+            U service,
+            String name,
+            InstrumentationFilter instrumentationFilter) {
+        return Instrumentation.builder(serviceInterface, service)
+                .withFilter(instrumentationFilter)
```
jkozlowski: This should be the only change here, the diff is weird because I'm making this into a private method.

atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/Timed.java
```diff
@@ -0,0 +1,28 @@
...
+package com.palantir.atlasdb.metrics;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Timed {
```
jkozlowski: This is probably a tentative place for this, but as good as anything? or do we have strong guarantees around the atlasdb-api project?

felixdesouza: I don't think you can put this anywhere else if you want to annotate `atlasdb-api` classes with them.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
```diff
@@ -37,20 +39,23 @@
...
 
+    public static <T, U extends T> T instrumentTimed(
+            MetricRegistry metricRegistry, Class<T> serviceInterface, U service) {
+        return instrument(metricRegistry, serviceInterface, service, serviceInterface.getName(), instrumentTimedOnly());
+    }
+
+    /**
+     * @deprecated use {@link #instrumentTimed(MetricRegistry, Class, Object)}
+     */
+    @Deprecated
```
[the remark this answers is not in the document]
jeremyk-91: We don't run with `-Werror` and people using this directly should be able to make the change, so I see little harm declaring this early.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -521,7 +521,7 @@ private TransactionComponents createTransactionComponents(
                 metricsManager, lockAndTimestampServices, keyValueService);
         TransactionSchemaManager transactionSchemaManager = new TransactionSchemaManager(coordinationService);
 
-        TransactionService transactionService = initializeCloseable(() -> AtlasDbMetrics.instrument(
+        TransactionService transactionService = initializeCloseable(() -> AtlasDbMetrics.instrumentTimed(
```
jeremyk-91: Hmm, we probably want this call at `SerializableTransactionManager:385`, not here. `TransactionService` is a separate class from `TransactionManager` (the former is the thing that manages start->commit timestamp mappings, the latter is the interface users have to AtlasDB).

atlasdb-client/src/test/java/com/palantir/atlasdb/util/TimedOnlyInstrumentationFilterTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+import com.palantir.atlasdb.metrics.Timed;
+
+@RunWith(MockitoJUnitRunner.class)
+public final class TimedOnlyInstrumentationFilterTest {
+
+    @Mock
+    private TestService testService;
+    private final TimedOnlyInstrumentationFilter filter = new TimedOnlyInstrumentationFilter();
```
felixdesouza: nit: move the final field above the non-final

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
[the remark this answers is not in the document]
felixdesouza: no one uses this internally

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Ah yes, what was I thinking.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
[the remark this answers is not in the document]
jkozlowski: Cool, in that case I will leave the deprecation and move classes through separate PRs. once last one is moved, I will remove the deprecated method.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Fixed

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
```diff
@@ -37,20 +39,23 @@
...
-                .withHandler(new SlidingWindowMetricsInvocationHandler(metricRegistry, name))
-                .withLogging(
-                        LoggerFactory.getLogger("performance." + name),
-                        LoggingLevel.TRACE,
-                        LoggingInvocationEventHandler.LOG_DURATIONS_GREATER_THAN_1_MICROSECOND)
-                .build();
+        return instrument(metricRegistry, serviceInterface, service, name, instrumentAllMethods());
     }
 
     public static <T, U extends T> T instrumentWithTaggedMetrics(
```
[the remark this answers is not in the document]
jkozlowski: Yes, I initially implemented that but wanted to do it separately. I will move all the classes that use the non-tagged registry first, and then move onto this.

atlasdb-client/src/test/java/com/palantir/atlasdb/util/TimedOnlyInstrumentationFilterTest.java
jkozlowski: Done

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
```diff
@@ -15,54 +15,94 @@
...
+            String methodTimerName,
+            Supplier<String> invocation,
+            boolean shouldBeInvoked) {
+        assertTimerNotRegistered(methodTimerName);
 
-        assertThat(service.ping(), is(equalTo(PING_RESPONSE)));
+        assertThat(invocation.get()).isEqualTo(PING_RESPONSE);
 
-        assertThat(metrics.timer(methodTimerName).getCount(), is(equalTo(1L)));
+        if (shouldBeInvoked) {
```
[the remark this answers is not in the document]
jkozlowski: I think the problem is that the timer should not even be registered, so I kinda treated this test as an integration test. But happy to remove that.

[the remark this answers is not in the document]
jkozlowski: How about if I just rename the assertion method to something like assertThatInstrumented vs assertThatNotInstrumented.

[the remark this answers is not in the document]
felixdesouza: having two methods, or having `assertThatInstrumented` return a `BooleanAssertion` would also work

jkozlowski:
@jeremyk-91 kept the #getUnreadableTimestamp timer for now.

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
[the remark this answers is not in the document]
jkozlowski: Went for 2 methods, should now be pretty readable.

jkozlowski:
@jeremyk-91 I won't be adding comments as to why certain ones are whitelisted, since the code is pretty easy to follow.

PR #4323, opened 2019-10-21
[Timelock Partitioning] Part 39: Have separate paxos leadership remote clients.

felixdesouza:
**Goals (and why)**:
I thought I could use `TimelockPaxos(Acceptor|Learner)Adapter` for both timestamp and non-partitioned acceptor/learner remote clients. This was not true.

The former produces URLs in the form:

    /.internal/leaderPaxos/leaderPaxos/learner/learned-values

where `leaderPaxos` is both the `useCase` and the `client`. Which doesn't have a matching resource, server side, where the server side expects

   /.internal/leaderPaxos/learner/learned-values

I thought I had a hack where I would produce a string client with value `""`. But it turns out that

    /.internal/leaderPaxos//learner/learned-values

is not recognised by the server.

I previously added new endpoints to `LeadershipResource` but that's not backwards compatible even when not using partitioned timelock. So we have to make the current clients still speak to the old endpoints.

Have a new interface that extends `PaxosLearner` and `PaxosAcceptor`, but just with a different base.

Tested in a future branch, this code is unused for now, trying to merge some changes in incrementally.

ASAP

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
@@ -19,6 +19,8 @@ package com.palantir.atlasdb.timelock.paxos;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import javax.ws.rs.Path;
+
 import org.immutables.value.Value;
 
 import com.google.common.collect.ImmutableMap;
@@ -26,6 +28,8 @@ import com.palantir.atlasdb.config.AuxiliaryRemotingParameters;
 import com.palantir.atlasdb.http.AtlasDbHttpClients;
 import com.palantir.atlasdb.util.AtlasDbMetrics;
 import com.palantir.conjure.java.api.config.service.UserAgents;
+import com.palantir.paxos.PaxosAcceptor;
+import com.palantir.paxos.PaxosLearner;
 import com.palantir.timelock.paxos.TimelockPaxosAcceptorRpcClient;
 import com.palantir.timelock.paxos.TimelockPaxosLearnerRpcClient;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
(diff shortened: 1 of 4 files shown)


timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosLearnerAdapter.java
```diff
@@ -72,9 +74,23 @@ public PaxosValue getGreatestLearnedValue() {
...
+                return client -> remoteClients.nonBatchTimestampLearner().stream()
+                        .map(learner -> new TimelockPaxosLearnerAdapter(
+                                paxosUseCase,
+                                client.value(),
+                                learner)).collect(Collectors.toList());
+            default:
+                throw new IllegalStateException("Unexpected value: " + paxosUseCase);
+        }
     }
+
```
gmaretic: 😨

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosAcceptorAdapter.java
```diff
@@ -63,9 +65,23 @@ public long getLatestSequencePreparedOrAccepted() {
...
     public static Function<Client, List<PaxosAcceptor>> wrap(
             PaxosUseCase paxosUseCase,
-            List<TimelockPaxosAcceptorRpcClient> acceptors) {
-        return client -> acceptors.stream()
-                .map(acceptor -> new TimelockPaxosAcceptorAdapter(paxosUseCase, client, acceptor))
-                .collect(Collectors.toList());
+            PaxosRemoteClients remoteClients) {
+        switch (paxosUseCase) {
+            case LEADER_FOR_ALL_CLIENTS:
+                return _client -> remoteClients.singleLeaderAcceptor().stream()
```
gmaretic: is underscore now a fancy way to denote it's unused? I prever just using `unused`

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -40,15 +44,29 @@
...
+        return createInstrumentedRemoteProxies(
+                TimelockSingleLeaderPaxosLearnerRpcClient.class,
+                "paxos-leader-learner-rpc-client");
+    }
+
+    @Value.Derived
+    public List<TimelockSingleLeaderPaxosAcceptorRpcClient> singleLeaderAcceptor() {
+        return createInstrumentedRemoteProxies(
+                TimelockSingleLeaderPaxosAcceptorRpcClient.class,
+                "paxos-leader-acceptor  -rpc-client");
```
gmaretic: spaces?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -83,4 +101,16 @@
                         _unused -> ImmutableMap.of()))
                 .collect(Collectors.toList());
     }
+
+    @Path("/" + PaxosTimeLockConstants.INTERNAL_NAMESPACE
```
gmaretic: nit: extract constant?

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosAcceptorAdapter.java
```diff
@@ -63,9 +65,23 @@ public long getLatestSequencePreparedOrAccepted() {
...
+                        .<PaxosAcceptor>map(Function.identity())
+                        .collect(Collectors.toList());
+            case LEADER_FOR_EACH_CLIENT:
```
gmaretic: These paxos use cases sounds surreal, but I trust it makes sense

gmaretic:
Some nits and a typo

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosAcceptorAdapter.java
felixdesouza: I would have put this logic inside the `PaxosUseCase`, but dependencies :'(

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
felixdesouza: meh, don't think it's worth it, I find it easier to parse the path here in terms of how things are grouped together

i.e. is it clientPaxos vs leaderPaxos, given the use case is it batch or just normal etc.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosAcceptorAdapter.java
felixdesouza: baseline will convert everyone to it anyway, I prefer $ but alas, I'm poor

PR #4326, opened 2019-10-21
More efficient batch size increasing iterator

j-baker:
By remembering whether we've ran out of underlying iterator, we can omit
a KVS call each time and easily call ourselves done.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/BatchSizeIncreasingIterator.java
```diff
@@ -49,7 +50,7 @@ public BatchSizeIncreasingIterator(BatchProvider<T> batchProvider,
         Preconditions.checkArgument(originalBatchSize > 0);
         this.batchProvider = batchProvider;
         this.originalBatchSize = originalBatchSize;
-        this.currentResults = currentResults;
+        this.currentResults = new ExhaustibleClosableIterator<>(currentResults);
```
jeremyk-91: I'm pretty sure we only want to set this if `currentResults` is not null (i.e. this line should be moved to inside the `if currentResults != null` block). If `currentResults` is null, we probably want it to stay null so that the first block of `updateResultsIfNeeded()` can fire.

j-baker: you are right! I'll write a test for this and let you know when i'm done.

PR #4327, opened 2019-10-22
[Timelock Partitioning] Part 40: `LeaderPinger` returns `PingResult`

felixdesouza:
**Goals (and why)**:
`LeaderPinger` now returns a `LeaderPingResult`, meaning each implementation no longer needs to know about a `PaxosLeadershipEventRecorder` (which is client scoped, meaning annoying dependency injection), 

Recording the event will all be handled by the `PaxosLeaderElectionService` which it partially does already.

* Introduce a new type (using derive4j) representing the different states a leader ping will return.

Amend existing tests.

* `LeaderPingResult`
* `SingleLeaderPinger`
* `AutobatchingLeaderPinger`

ASAP 💃

### DIFF
--- a/leader-election-impl/build.gradle
+++ b/leader-election-impl/build.gradle
@@ -14,6 +14,8 @@ dependencies {
 
   annotationProcessor group: 'org.immutables', name: 'value'
   compileOnly 'org.immutables:value::annotations'
+  annotationProcessor 'org.derive4j:derive4j'
+  compileOnly 'org.derive4j:derive4j-annotation'
 
   testCompile(group: "org.jmock", name: "jmock", version: libVersions.jmock) {
     exclude group: 'org.hamcrest'
--- a/leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingResult.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingResult.java
@@ -1,0 +1,84 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.paxos;
+
+import java.util.function.Consumer;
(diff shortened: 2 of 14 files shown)


.palantir/revapi.yml
```diff
@@ -80,6 +80,18 @@ acceptedBreaks:
         \ java.util.function.Supplier<com.palantir.atlasdb.config.ServerListConfig>,\
         \ java.lang.Class<T>, com.palantir.atlasdb.config.AuxiliaryRemotingParameters)"
       justification: "Just let me merge"
```
jeremyk-91: [non-blocking] !

leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingResult.java
```diff
@@ -0,0 +1,84 @@
...
+                .pingTimedOut(wrap(eventRecorder::recordLeaderPingTimeout))
+                .pingReturnedFalse(wrap(eventRecorder::recordLeaderPingReturnedFalse))
+                .pingCallFailure(wrap(eventRecorder::recordLeaderPingFailure))
+                .otherwise_(null);
+    }
+
+    public boolean isSuccessful() {
+        return LeaderPingResults.caseOf(this)
+                .pingReturnedTrue_(true)
+                .otherwise_(false);
```
jeremyk-91: Possibly just a matter of taste, but I really don't like the trailing underscores. I'd prefer the `otherwise(() -> false);` that this actually is under the hood

leader-election-impl/build.gradle
```diff
@@ -14,6 +14,8 @@ dependencies {
 
   annotationProcessor group: 'org.immutables', name: 'value'
   compileOnly 'org.immutables:value::annotations'
+  annotationProcessor 'org.derive4j:derive4j'
+  compile 'org.derive4j:derive4j-annotation'
```
jeremyk-91: Pretty sure this needs to be `compileOnly` - `compile` would be bad because of licensing reasons

versions.props
```diff
@@ -79,6 +79,7 @@ org.objenesis:objenesis = 3.0.1
 org.ow2.asm:asm = 7.0
 org.slf4j:* = 1.7.25
 org.yaml:snakeyaml = 1.23
+org.derive4j:* = 1.1.1
```
jeremyk-91: nit: Please put this in the right place

.palantir/revapi.yml
felixdesouza: I'm undoing this commit.

leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingResult.java
felixdesouza: you get used to it pretty quickly, having to write suppliers/functions for constants gets pretty annoying/visual cruft. 

Agree underscores aren't ideal, but I think pointless suppliers are worse.

PR #4329, opened 2019-10-22
Lazily create timers.

jkozlowski:
* Should help with instrumented methods that are never called.

**Goals (and why)**:


changelog/@unreleased/pr-4329.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Lazily create timers.
```
felixdesouza: if we're going to have changelog, can it be a bit more descriptive please

PR #4330, opened 2019-10-22
[CJR] Runtime config for disabling experiment fallbacks

gmaretic:
**Goals (and why)**:
If we really want to test out the CJR client, we need to disable fallbacks. This allows us to do so, while still allowing us to flip the switch if things start going south.

Not happy about effectively leaking the fallback logic into `VersionSelectingClients`, but there really isn't a great way around it.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
@@ -101,11 +101,15 @@ public final class ExperimentRunningProxy<T> extends AbstractInvocationHandler {
     }
 
     private boolean useExperimental() {
-        return useExperimental.getAsBoolean() && !fallback();
-    }
+        if (!useExperimental.getAsBoolean()) {
+            return false;
+        }
 
-    private boolean fallback() {
-        return enableFallback.getAsBoolean() && !sufficientTimeSinceFailure();
+        if (!enableFallback.getAsBoolean()) {
+            return true;
+        }
+
+        return sufficientTimeSinceFailure();
     }
 
     private boolean sufficientTimeSinceFailure() {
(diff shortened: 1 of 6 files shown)


atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
```diff
@@ -50,19 +51,22 @@
...
         this.experimentalService = experimentalService;
         this.fallbackService = fallbackService;
         this.useExperimental = useExperimental;
+        this.enableFallback = enableFallback;
         this.clock = clock;
         this.errorTask = errorTask;
     }
 
+    @SuppressWarnings("unchecked")
     public static <T> T newProxyInstance(T experimentalService, T fallbackService, BooleanSupplier useExperimental,
```
felixdesouza: nit: new lines please

atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
```diff
@@ -87,14 +101,25 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
...
+        return useExperimental.getAsBoolean() && !fallback();
+    }
+
+    private boolean fallback() {
+        return enableFallback.getAsBoolean() && !sufficientTimeSinceFailure();
+    }
+
+    private boolean sufficientTimeSinceFailure() {
+        return Instant.now(clock).compareTo(nextPermittedExperiment.get()) > 0;
     }
```
jeremyk-91: This is fine. My thought process would be closer to something like this:

```java
    private boolean useExperimental() {
        if (!useExperimental.getAsBoolean()) {
            return false;
        }

        if (!enableFallback.getAsBoolean()) {
            return true;
        }

        return sufficientTimeSinceFailure();
    }
```

PR #4331, opened 2019-10-22
Only whitelist certain CassandraClient methods for metrics.

jkozlowski:
**Goals (and why)**:

CassandraClient methods and their metrics status:

### Not timed:
* isValid: not timed
* describe_snitch: not timed, just a background thing. If there are issues with this, likely there are issues with the cluster?
* getOutputProtocol/getInputProtocol: not timed, seems to be only used for pool management?
* describe_ring: not timed, pool management
* describe_version: not timed, startup checks?
* describe_schema_versions: not timed, startup
* describe_partitioner: not timed, pool management
* describe_keyspace: not timed, seems like mixture of pool management, and startup
* describe_keyspaces: not timed, seems like only used by CassandraVerifier
* system_add_keyspace: not timed, seems like only used by CassandraVerifier; **might be used by internal product?**
* system_update_keyspace: not timed, seems like only used by CassandraVerifier; **might be used by internal product?**
* system_update_column_family: not timed, startup or create tables; **might be used by internal product?**
* system_drop_column_family: not timed, dropping tables.; **might be used by internal product?**

### Timed:
* multiget_slice: timed
* multiget_multislice: timed
* get_range_slices: timed
* batch_mutate: timed, there's additionally some metrics in InstrumentedCassandraClient for this.
* get: timed, although the only usage I can see is in PersistentTimestampServiceImpl? Which seems to query this stuff on startup.
* cas: timed, same as above, what is this stuff?
* put_unless_exists: timed
* execute_cql3_query: timed. It's used in CassandraTableCreator (so only at startup), but also *I think* in other places. We might be able to remove this, but I'm not that fussed.
* remove: timed. used by sweep?

### Other:
* **system_add_column_family: removed, doesn't look like it's used?**

Would like a reviewer to go through the list of methods that I chose to stop timing, to double check that we're not missing anything. I have done cursory looking by classifying things quickly into "probably only used at startup, so don't really care about metrics", but could have possibly missed things. If internal products rely on the table/keyspace management methods being timed, we need to check if they have accurate custom metrics for monitoring that.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
@@ -83,7 +83,7 @@ public interface CassandraClient extends Closeable {
             ConsistencyLevel consistency_level)
             throws InvalidRequestException, UnavailableException, TimedOutException, org.apache.thrift.TException;
 
-    // Instrumented manually
+    @Timed
     void batch_mutate(String kvsMethodName,
             Map<ByteBuffer, Map<String, List<Mutation>>> mutation_map,
             ConsistencyLevel consistency_level)
@@ -150,9 +150,6 @@ public interface CassandraClient extends Closeable {
             throws InvalidRequestException, SchemaDisagreementException, TException;
 
     String system_update_keyspace(KsDef ks_def)
-            throws InvalidRequestException, SchemaDisagreementException, TException;
-
-    String system_add_column_family(CfDef cf_def)
             throws InvalidRequestException, SchemaDisagreementException, TException;
 
     String system_update_column_family(CfDef cf_def)
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
@@ -114,7 +114,6 @@ public interface CassandraClient extends Closeable {
             ConsistencyLevel commit_consistency_level)
             throws InvalidRequestException, UnavailableException, TimedOutException, org.apache.thrift.TException;
 
-    @Timed
     CqlResult execute_cql3_query(CqlQuery cqlQuery,
             Compression compression,
             ConsistencyLevel consistency)
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
@@ -44,6 +44,7 @@ import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TProtocol;
 
 import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.metrics.Timed;
 import com.palantir.processors.AutoDelegate;
 
 @SuppressWarnings({"all"}) // thrift variable names.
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
```diff
@@ -142,23 +152,24 @@ String system_add_keyspace(KsDef ks_def)
     String system_update_keyspace(KsDef ks_def)
             throws InvalidRequestException, SchemaDisagreementException, TException;
 
-    String system_add_column_family(CfDef cf_def)
```
felixdesouza: unused?

felixdesouza:
> system_update_keyspace: not timed, CassandraVerifier

is this just saying that `CassandraVerifier` is what instruments this?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
jkozlowski: yeah, everything builds without it? See PR comment

jkozlowski: But could be that internal products use this? @felixdesouza how can we check that?

felixdesouza: I usually just open it up, it's unused inside big internal product.

jkozlowski: Sweet, so should be fine to nuke it.

jeremyk-91:
The specific internal product I was worried about does rely on keyspace/table management being reasonably fast, but does not at this time have instrumentation there. However, they don't frequently use Cassandra. I'll see if there's a better place to re-instate those metrics, but these can definitely go - the worst case is I put them back on `KeyValueService` probably.

Now looking at the other metrics.

jeremyk-91:
Some thoughts, going through what we have:

> getOutputProtocol/getInputProtocol: not timed, seems to be only used for pool management?

Actually, getOutputProtocol is called synchronously every time we return connections to the pool (which I think is after every Cassandra operation we run). However, it should still be OK for this to be untimed as it is an in-memory read.

> get: timed, although the only usage I can see is in PersistentTimestampServiceImpl? Which seems to query this stuff on startup.
> cas: timed, same as above, what is this stuff?

These are used by embedded timelock / leader block services. Might be used by large internal product (the codepath is still used by them). I think they may also be used by the persistent lock.

> execute_cql3_query: timed. It's used in CassandraTableCreator (so only at startup), but also I think in other places. We might be able to remove this, but I'm not that fussed.

Also used in some timelock migration paths (similarly startup). Most other use cases should be covered by CAS, so this is actually probably removable.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClient.java
```diff
@@ -99,19 +106,22 @@ CASResult cas(TableReference tableReference,
...
 
+    @Timed
     public CASResult put_unless_exists(TableReference tableReference,
             ByteBuffer key,
             List<Column> updates,
             ConsistencyLevel serial_consistency_level,
             ConsistencyLevel commit_consistency_level)
             throws InvalidRequestException, UnavailableException, TimedOutException, org.apache.thrift.TException;
 
+    @Timed
```
jeremyk-91: I _think_ this can be removed (see earlier comment)

jkozlowski: Removed. If we need it we can bring it back

PR #4333, opened 2019-10-22
[Timelock Partitioning] Part 42: `LeadershipMetrics` propagates its client.

felixdesouza:
**Goals (and why)**:
Due to #4327, we can now create a `TimelockLeadershipMetrics` for each client and don't have to worry about maintaining a "global" leadership metrics and then a client aware leadership metrics.

* If we pass `contextArgs` to `LeadershipEvents` we should then make sure any metrics are tagged with those args so it's not just the logging etc.
* Add the ability to create an `AutobatchingLeadershipObserverFactory` that will deregister the right set of things depending on the use case.
  * with single leader, we want to deregister everything, since we only have one `LeaderElectionService`
  * with partitioned leadership, we want to deregister everything that matches the `isCurrentSuspectedLeader` predicate *and* is the client that had a leadership event change.

ASAP

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockLeadershipMetrics.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockLeadershipMetrics.java
@@ -16,58 +16,120 @@
 
 package com.palantir.atlasdb.timelock.paxos;
 
+import static com.palantir.atlasdb.AtlasDbMetricNames.TAG_CLIENT;
+import static com.palantir.atlasdb.AtlasDbMetricNames.TAG_CURRENT_SUSPECTED_LEADER;
+import static com.palantir.atlasdb.AtlasDbMetricNames.TAG_PAXOS_USE_CASE;
+
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 
 import org.immutables.value.Value;
 
-import com.codahale.metrics.MetricRegistry;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.palantir.atlasdb.AtlasDbMetricNames;
+import com.google.common.collect.Multimaps;
+import com.google.common.collect.SetMultimap;
+import com.palantir.atlasdb.timelock.paxos.AutobatchingLeadershipObserverFactory.LeadershipEvent;
 import com.palantir.atlasdb.util.AtlasDbMetrics;
-import com.palantir.atlasdb.util.MetricsManager;
-import com.palantir.atlasdb.util.MetricsManagers;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.leader.LeadershipObserver;
+import com.palantir.leader.PaxosLeadershipEventRecorder;
 import com.palantir.leader.PingableLeader;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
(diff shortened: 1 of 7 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockLeadershipMetrics.java
```diff
@@ -16,58 +16,120 @@
...
-        return ImmutableList.of(SafeArg.of(AtlasDbMetricNames.TAG_PAXOS_USE_CASE, metrics().paxosUseCase().toString()));
+    public static AutobatchingLeadershipObserverFactory createFactory(TimelockPaxosMetrics metrics) {
+        return AutobatchingLeadershipObserverFactory.create(getMetricDeregistrator(metrics));
+    }
+
+    private static Consumer<SetMultimap<LeadershipEvent, Client>> getMetricDeregistrator(TimelockPaxosMetrics metrics) {
+        PaxosUseCase paxosUseCase = metrics.paxosUseCase();
+        switch (paxosUseCase) {
+            case LEADER_FOR_ALL_CLIENTS:
+                return deregisterMetricsForSingleLeader(metrics);
```
felixdesouza: this would be a perfect use for derive4j, but alas that is a refactoring for another day

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockLeadershipMetrics.java
```diff
@@ -16,58 +16,120 @@
...
+
+    private static Consumer<SetMultimap<LeadershipEvent, Client>> deregisterMetricsForPartitionedLeader(
+            TimelockPaxosMetrics metrics) {
+        return eventsToDeregister -> {
+            Map<LeadershipEvent, Set<String>> eventsToClients = Multimaps.asMap(KeyedStream.stream(eventsToDeregister)
+                    .map(Client::value)
+                    .collectToSetMultimap());
+            eventsToClients.keySet().forEach(event -> metrics.asMetricsManager().deregisterTaggedMetrics(
+                    withTagIsCurrentSuspectedLeader(event.isCurrentSuspectedLeader())
+                            .and(withClientTag(eventsToClients.get(event)))));
```
jeremyk-91: maybe use `entrySet`? Not superhot codepath so I'm not too strongly concerned

felixdesouza: with entry set, run time becomes `#metrics * #clients`, since deregister will enumerate each metric and run the predicate, can't get rid of the `#metrics`

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockPaxosMetrics.java
```diff
@@ -37,6 +40,12 @@ TaggedMetricRegistry metrics() {
         return new SlidingWindowTaggedMetricRegistry(35, TimeUnit.SECONDS);
     }
 
+    @Value.Derived
+    MetricsManager asMetricsManager() {
+        // we don't use the normal metric registry so we don't care about this
+        return MetricsManagers.of(new MetricRegistry(), metrics());
```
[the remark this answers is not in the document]
felixdesouza: this tagged metric registry is brand new and then is attached to the parent tagged metric registry, so everything here is contained to just this paxos use case

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/TimelockLeadershipMetrics.java
felixdesouza: spoke offline, entry set of `Map<Bla, Collection<Bla2>>`

PR #4336, opened 2019-10-22
[Timelock Partitioning] Part 43: `LocalPaxosComponents` knows about `LocalPingableLeader`

felixdesouza:
**Goals (and why)**:
`LocalPingableLeader` will be queried indirectly for the tagged metrics. All it does is wrap `PaxosLearner`, as such it makes sense for it to live inside `LocalPaxosComponents`.

ASAP


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -81,7 +82,16 @@
                 "batch-paxos-learner-rpc-client");
     }
 
+    @Value.Derived
+    public List<PingableLeader> nonBatchPingableLeaders() {
+        return createInstrumentedRemoteProxies(PingableLeader.class, "paxos-pingable-leader-rpc-client", false);
```
jeremyk-91: consistent with TimeLock Stability work, so `false` is OK here

PR #4340, opened 2019-10-23
[Timelock Partitioning] Part 44: `LeadershipContextFactory`

felixdesouza:
**Goals (and why)**:
This is the part we've been waiting for 💃. Given a use case, and some seed materials, it will construct the right things, and give a `LeadershipContextFactory`. Given a `LeadershipContextFactory` and a few other things, we can create a `LeadershipComponents` introduced in #4314, which can then be used to wrap arbitrary interfaces in an `AwaitingLeadershipProxy` backed by a (potentially client scoped) `LeaderElectionService`.

* `LeadershipContextFactory` initialises a few commonly used "dependencies", and seeds creation of each `ClientAwareComponents` which in turn gives us a `LeaderElectionService`.
* `LeadershipContextFactory.ClientAwareComponents` takes dependencies from `LeadershipContextFactory` and creates client scoped components, eventually getting a `LeaderElectionService` and a `LeadershipContext` 💃 
* `PaxosResourcesFactory` is doing a lot of the work where depending on what mode has been set, (well it's only one currently), it will initialise a `LeadershipContextFactory` but with single leader dependencies, and also register the resources accordingly.
  * The idea moving forward is that when we want to do the partitioned piece, we will then initialise a different `LeadershipContextFactory` but with different network client factories, leader pinger, different resources etc. This change should a be a lot simpler as all the wiring is already done. Less than 50 lines, but tests will take up the majority of that.

This is changing the wiring path, so current tests should still pass.

* I'm currently using a shared executor, undoing the executor stuff introduced a while back. This is mainly for simplicity.
* * `PaxosUseCase` is becoming a mishmash of an enum, and I think would benefit from derive4j, and then for the weird methods which are used in many different places, we can use the sweet visitor form with some good docs and it works out quite nicely. Obviously not in this PR.

* commit by commit, disregard the first commit since it is based on #4336 
* `LeadershipContextFactory`
  * `Dependencies`, this defines a set of interfaces that immutables implements which allows us to use the `from` method on the builder without having to individually instantiate each thing.
* `PaxosUseCase`
* `PaxosResourcesFactory`
* `AsyncTimelockServicesCreator` to see how we're using `LeadershipComponents`.

Other things are mainly wiring bits and aren't super interesting.

ASAP 💃


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosUseCase.java
```diff
@@ -27,22 +27,39 @@
...
-            CLIENT_PAXOS_NAMESPACE,
-            Paths.get("")); // <data-directory>
+            Paths.get(LEADER_PAXOS_NAMESPACE, MULTI_LEADER_PAXOS_NAMESPACE)) {
+        @Override
+        public Client resolveClient(Client client) {
+            return client;
+        }
+    },
+
+    // <data-directory>
```
jeremyk-91: `<data-directory>/<client>/{acceptor/learner}`?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
```diff
@@ -53,10 +58,52 @@ public static PaxosResources create(
...
+                    .leaderPingHealthCheckFactory(healthCheckFactory)
+                    .build();
+
+            resourcesBuilder.leadershipContextFactory(factory);
+            resourcesBuilder.addAdhocResources(
+                    new LeadershipResource(
+                            factory.components().acceptor(PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT),
+                            factory.components().learner(PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT)),
+                    factory.components().pingableLeader(PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT));
+        }
```
jeremyk-91: This part looks right, but we should see if it's possible to split this up into functions etc. so it's easier to read

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/LeadershipContextFactory.java
```diff
@@ -0,0 +1,165 @@
...
+import com.palantir.paxos.PaxosProposer;
+import com.palantir.timelock.paxos.LeaderPingHealthCheck;
+
+@Value.Immutable
+public abstract class LeadershipContextFactory implements
+        Factory<LeadershipContext>,
+        Dependencies.NetworkClientFactories,
+        Dependencies.LeaderPinger,
+        Dependencies.ClientAwareComponents,
+        Dependencies.LeaderPingHealthCheck {
```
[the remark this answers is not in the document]
felixdesouza: 🗡 is what we need 🔥 💃 but until then, this is the poor approximation to it 😅

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/LeadershipContextFactory.java
```diff
@@ -0,0 +1,165 @@
...
+        @Value.Derived
+        LeadershipContext leadershipContext() {
+            return ImmutableLeadershipContext.builder()
+                    .leadershipMetrics(leadershipMetrics())
+                    .leaderElectionService(leaderElectionService())
+                    .build();
+        }
+
+        private PaxosProposer instrumentProposer(PaxosProposer uninstrumentedPaxosProposer) {
+            return metrics().instrument(
```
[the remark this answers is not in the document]
felixdesouza: This doesn't work because this work hasn't been done for `TaggedMetricRegistry`s. There isn't a single class that's been instrumented with `MetricRegistry`.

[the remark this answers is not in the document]
felixdesouza: I can add the annotation to "future"-proof it, but considering it doesn't do anything and it's not related to this PR, I can do that in a later PR.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
felixdesouza: I've extracted logic for single leader into a `configureLeaderForAllClients` method. It's quite self contained there and I don't think it makes sense to break it even further.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
```diff
@@ -65,45 +65,61 @@ public static PaxosResources create(
...
-
-            Factories.LeaderPingHealthCheckFactory healthCheckFactory = dependencies -> {
-                Set<PingableLeader> allPingableLeaders = ImmutableSet.<PingableLeader>builder()
-                        .add(dependencies.components().pingableLeader(PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT))
-                        .addAll(dependencies.remoteClients().nonBatchPingableLeaders())
-                        .build();
-                return new LeaderPingHealthCheck(allPingableLeaders);
-            };
+            configureLeaderForAllClients(
+                    resourcesBuilder, install,
```
jeremyk-91: nit: separate lines

felixdesouza:
Force pushed to resolve merge conflicts.

felixdesouza:
Manually verified that when timelock switches to this we still get timestamps that are monotonically increasing. This is just verifying we haven't done anything funky with directories etc.

PR #4344, opened 2019-10-24
[Transaction Async get] Part 1: Return type for getWithPostFiltering

OStevan:
**Goals (and why)**:
Add a return type for `getWithPostFiltering` for easier migration to `ListenableFuture` chaining.
- added a return type with usage of generics

- existing tests should cover this

- `SnapshotTransaction`

- this week

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -453,25 +453,27 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 batchProvider, columnRangeSelection.getBatchHint(), ClosableIterators.wrap(rawIterator));
         Iterator<Iterator<Map.Entry<Cell, byte[]>>> postFilteredBatches =
                 new AbstractIterator<Iterator<Map.Entry<Cell, byte[]>>>() {
-            @Override
-            protected Iterator<Map.Entry<Cell, byte[]>> computeNext() {
-                ImmutableMap.Builder<Cell, Value> rawBuilder = ImmutableMap.builder();
-                List<Map.Entry<Cell, Value>> batch = batchIterator.getBatch();
-                for (Map.Entry<Cell, Value> result : batch) {
-                    rawBuilder.put(result);
-                }
-                Map<Cell, Value> raw = rawBuilder.build();
-                validatePreCommitRequirementsOnReadIfNecessary(tableRef, getStartTimestamp());
-                if (raw.isEmpty()) {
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1119,10 +1119,11 @@ private int estimateSize(List<RowResult<Value>> rangeRows) {
...
     }
 
-    private <T> void getWithPostFiltering(TableReference tableRef,
-                                          Map<Cell, Value> rawResults,
-                                          @Output ImmutableMap.Builder<Cell, T> results,
-                                          Function<Value, T> transformer) {
+    private <T, R extends ImmutableMap.Builder<Cell, T>> R getWithPostFiltering(
+            TableReference tableRef,
+            Map<Cell, Value> rawResults,
+            R results,
```
jeremyk-91: Hmm, `@Output` isn't exactly right but I'd still suggest having some form of documentation here (possibly in variable names or otherwise) that this will be modified and used to return the output

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -453,25 +453,27 @@ public void disableReadWriteConflictChecking(TableReference tableRef) {
...
-                    return endOfData();
-                }
-                ImmutableSortedMap.Builder<Cell, byte[]> post = ImmutableSortedMap.naturalOrder();
-                getWithPostFiltering(tableRef, raw, post, Value.GET_VALUE);
-                SortedMap<Cell, byte[]> postFiltered = post.build();
-                batchIterator.markNumResultsNotDeleted(postFiltered.size());
-                return postFiltered.entrySet().iterator();
-            }
-        };
+                    @Override
```
jeremyk-91: This is why we generally recommend against automatic reformatting, the actual diff is much smaller than this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
OStevan: Renamed to `resultsAccumulator`.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
OStevan: Revered and just applied the necessary changes.

PR #4356, opened 2019-10-29
Stop using a hardcoded password

safreiberg:


atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/OracleConnectionConfig.java
```diff
@@ -135,7 +135,7 @@ public Properties getHikariProperties() {
             // Create the truststore
             File clientTrustore = new File(getTruststorePath().get());
             props.setProperty("javax.net.ssl.trustStore", clientTrustore.getAbsolutePath());
-            props.setProperty("javax.net.ssl.trustStorePassword", "ptclient");
+            props.setProperty("javax.net.ssl.trustStorePassword", getTruststorePassword().get());
```
carterkozak: Perhaps set the property if the optional is present? The Truststore password should only be used to verify contents, we should be able to use the truststore without a password.

safreiberg: I think that's confusing because the `@Value.Check` method validates that the truststore and password are both set, it's not _really_ optional here

carterkozak: Ya, it looks a bit wonky since a truststore password isn't generally required to use a truststore, and we're not validating presence of an optional in the same place we're consuming it. I don't have a strong preference here, it just looked odd at a glance.

felixdesouza: does this mean that either this mode is never used, or everyone has hardcoded their trust store password to be `ptclient`? If so, is this not a break? i.e. if I provide the wrong password to a truststore will it not scream at me as opposed to not providing one because it's not required?

PR #4358, opened 2019-10-29
Fail loudly when a closed LockRefreshingLockService is used

carterkozak:
Previously a LockRefreshingLockService could be closed, but
still used despite no longer refreshing locks. It's better to
fail quickly and deterministically when the service is in an
unexpected state.


carterkozak:
Thanks for the quick review!

PR #4360, opened 2019-10-29
[Conjure Java Runtime] Part 16: Revising Project Paths, Fix FastFailoverProxy

jeremyk-91:
**IMPORTANT NOTE TO REVIEWERS: Unlike in many other PRs, in this PR the movement of files is a considerable part of the PR. Please read carefully and check that those movements make sense. Thanks!**

a.k.a. Fix FastFailoverProxy from fogging fanciness and fragment Feign folder for favorable future functioning.

**Goals (and why)**:
- Structure dependencies in a clear, unambiguous way. This also avoids some conceptual problems later on, where `atlasdb-conjure` code expects a shaded `feign.RetryableException` but receives an unshaded one.

- Implement two new projects `atlasdb-remoting-api` and `atlasdb-conjure`.

- The tests still work in the new projects, and integration tests should check the visibility of everything.

- I shoved `PollingRefreshable` into `atlasdb-remoting-api`, which is a bit sad because it requires a dependency on `commons-executors`. Not a perfect fit but I felt making another `atlasdb-remoting-impl-shared` project was a bit overkill.

- `build.gradle`s of the two projects.

**Priority (whenever / two weeks / yesterday)**: today or tomorrow?


versions.props
```diff
@@ -73,6 +73,7 @@ org.hdrhistogram:HdrHistogram = 2.1.10
 org.hibernate:hibernate-validator = 5.1.3.Final
 org.immutables:* = 2.7.3
 org.javassist:javassist = 3.18.2-GA
+org.jmock:jmock = 2.8.2
```
gmaretic: Remove from gradle/versions.gradle and update timestamp-impl?

jeremyk-91: Will do

PR #4361, opened 2019-10-30
[Conjure Java Runtime] Part 17: 🚨🚚🚨 Optionals for Paxos Learners

jeremyk-91:
**Goals (and why)**:
- Fulfil a prerequisite for internode communication via Conjure Java Runtime

- Switch `PaxosLearner` to return `Optional<PaxosValue>` and fix attendant breaks.
- Also switched `TimelockPaxosLearnerRpcClient` to use this.

- Updated tests that used to use nullable values to now use Optionals. These passing should be sufficient.

- This is obviously an API break, is it unforgivably large?
- There is no wire break (a wire break would obviously be disastrous) - please confirm this.

**Where should we start reviewing?**: `PaxosLearner`

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
@@ -107,9 +107,9 @@ public final class PaxosLearnerImpl implements PaxosLearner {
 
     @Override
     public Optional<PaxosValue> getGreatestLearnedValue() {
-        if (!state.isEmpty()) {
-            return Optional.ofNullable(state.get(state.lastKey()));
+        if (state.isEmpty()) {
+            return Optional.empty();
         }
-        return Optional.empty();
+        return Optional.ofNullable(state.get(state.lastKey()));
     }
 }
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
@@ -16,14 +16,17 @@
 package com.palantir.paxos;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Optional;
 import java.util.SortedMap;
 import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableList;
 import com.palantir.leader.PaxosKnowledgeEventRecorder;
 import com.palantir.logsafe.SafeArg;
 
@@ -69,7 +72,7 @@ public final class PaxosLearnerImpl implements PaxosLearner {
     }
 
     @Override
-    public PaxosValue getLearnedValue(long seq) {
+    public Optional<PaxosValue> getLearnedValue(long seq) {
         try {
             if (!state.containsKey(seq)) {
                 byte[] bytes = log.readRound(seq);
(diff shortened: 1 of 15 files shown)


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponentsTest.java
```diff
@@ -65,9 +65,13 @@ public void setUp() throws IOException {
     public void newClientCanBeCreated() {
         PaxosLearner learner = paxosComponents.learner(CLIENT);
         learner.learn(PAXOS_ROUND_ONE, PAXOS_VALUE);
-        assertThat(learner.getGreatestLearnedValue()).isNotNull();
-        assertThat(learner.getGreatestLearnedValue().getLeaderUUID()).isEqualTo(PAXOS_UUID);
-        assertThat(learner.getGreatestLearnedValue().getData()).isEqualTo(PAXOS_DATA);
+
+        assertThat(learner.getGreatestLearnedValue())
+                .isPresent()
```
OStevan: Call to `isPresent` can be removed, [`hasValueSatisfying`](https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AbstractOptionalAssert.html#hasValueSatisfying-org.assertj.core.api.Condition-) will also assert that the value is present.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
```diff
@@ -78,38 +81,35 @@ public PaxosValue getLearnedValue(long seq) {
...
+                .map(Optional::get)
+                .collect(Collectors.toList());
     }
 
     @Override
-    public PaxosValue getGreatestLearnedValue() {
+    public Optional<PaxosValue> getGreatestLearnedValue() {
         if (!state.isEmpty()) {
-            return state.get(state.lastKey());
+            return Optional.ofNullable(state.get(state.lastKey()));
```
OStevan: If I understood correctly that empty result is a "non-standard" path I would invert the if.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponentsTest.java
jeremyk-91: you're right! Will change this

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
felixdesouza: unrelated to above: this shouldn't be `null` right? since the map is not empty, unless you're thinking it has null values, but non-null keys

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderLearnerNetworkClient.java
```diff
@@ -84,7 +84,7 @@ public void learn(long seq, PaxosValue value) {
             Function<Optional<PaxosValue>, T> mapper) {
         return PaxosQuorumChecker.collectQuorumResponses(
                 allLearners,
-                learner -> mapper.apply(Optional.ofNullable(learner.getLearnedValue(seq))),
+                learner -> mapper.apply(learner.getLearnedValue(seq)),
```
felixdesouza: 💃

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
jeremyk-91: Agree on inverting if (didn't as I just translated what was there, but yeah it does make sense).

I wrote `Optional.of()` first, but there were some tests that failed. Possibly a bad test, I didn't investigate in detail

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponentsTest.java
felixdesouza: can we go further:

```java
Optional<PaxosValue> value = learner.getGreatestLearnedValue();
assertThat(value).map(PaxosValue::getLeaderUUID).contains(PAXOS_UUID);
assertThat(value).map(PaxosValue::getData).contains(PAXOS_DATA);
```

felixdesouza: note, `map` is a method on the `OptionalAssert`

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosLearnerRpcClient.java
```diff
@@ -54,20 +54,20 @@ void learn(
...
     @Path("learned-value/{seq}")
     @Produces(MediaType.APPLICATION_JSON)
-    PaxosValue getLearnedValue(
+    Optional<PaxosValue> getLearnedValue(
             @PathParam("useCase") PaxosUseCase paxosUseCase,
             @PathParam("client") String client,
             @PathParam("seq") long seq);
 
-    @Nullable
+    @Nonnull
```
felixdesouza: Don't think we need the `@Nonnull` annotations, since that should be implied with code wide `Nonnull` by default and also the fact that it's a huge code smell to pass around a `null` `Optional`

leader-election-impl/src/main/java/com/palantir/paxos/PaxosLearnerImpl.java
felixdesouza: Sadness is sad

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosLearnerRpcClient.java
jeremyk-91: This was done for consistency with the other methods in the client. I guess now all of them are nonnull it's probably fine to remove them.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponentsTest.java
jeremyk-91: (yep, changed to `OptionalAssert`'s map)

jeremyk-91:
Thanks!

PR #4366, opened 2019-10-30
[Conjure Java Runtime] Part 18: Internode Support

jeremyk-91:
**NOTE TO REVIEWERS: CI may fail until #4361 merges, since CJR will scream at the Nullable things it sees. This PR should nonetheless be reviewable.**

**Goals (and why)**:
- Clear another blocker to TimeLock nodes using CJR to talk to one another.
- Run tests with CJR.

- Switch single node clients to use CJR with provided probability.
- Run tests with CJR (that use `TestProxyUtils`). This includes integration and ETE tests with TimeLock. This does _not_ include running internode communication with CJR just yet, that will be in a follow-up PR.

Updated existing tests, all integration tests passing should be sufficient

- Removed two old tests for ThreadPooledLockService because exceptions are now handled internally in the CJR client as opposed to being bubbled up - please check this is fine.
- Some tests for payload limiting were removed because CJR seems to work fine with them!

**Where should we start reviewing?**: AtlasDbHttpClients

### DIFF
--- a/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
@@ -21,19 +21,11 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.SortedMap;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 
-import org.assertj.core.util.Lists;
 import org.awaitility.Awaitility;
 import org.eclipse.jetty.http.HttpStatus;
 import org.junit.BeforeClass;
@@ -44,7 +36,6 @@ import org.junit.rules.TemporaryFolder;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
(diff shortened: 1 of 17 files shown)


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -166,28 +166,6 @@ public void multipleClientsCanUseSharedThreads() throws Exception {
...
-    }
-
-    @Test
-    public void throwsOnTwoClientsRequestingSameLockTooManyTimes() throws Exception {
-        List<LockService> lockServiceList = ImmutableList.of(
-                getLockService(CLIENT_1), getLockService(CLIENT_2));
-        int exceedingRequests = 10;
-        int requestsPerClient = (MAX_CONCURRENT_LOCK_REQUESTS + exceedingRequests) / 2;
-
-        assertThat(lockAndUnlockAndCountExceptions(lockServiceList, requestsPerClient))
```
felixdesouza: is this still used after you've gotten rid of these invocations?

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/util/ExceptionMatchers.java
```diff
@@ -17,16 +17,15 @@
...
 
     public static void isRetryableExceptionWhereLeaderCannotBeFound(Throwable throwable) {
         assertThat(throwable)
-                .hasMessageContaining("method invoked on a non-leader");
-
-        // We shade Feign, so we can't rely on our client's RetryableException exactly matching ours.
-        assertThat(throwable.getClass().getName())
-                .contains("RetryableException");
+                .hasMessageContaining("receiving QosException.RetryOther")
+                .isInstanceOf(RetryableException.class);
```
felixdesouza: do we want to look at the cause and map that to a `QosException.RetryOther`?

jeremyk-91: I'll see what I can do about this

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
jeremyk-91: It was still used by the happy case test, though I've removed this and removed those as well as they don't really make sense after CJR.

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/util/ExceptionMatchers.java
jeremyk-91: (yep, the root cause can be attributed to one of those)

PR #4368, opened 2019-10-31
[Transaction Async get] Part 6: Catch unintentional switches between blocking and non-blocking calls during tests.

OStevan:
**Goals (and why)**:
Adds tests to prevent unintentional switching from blocking to non-blocking paths. Previously some paths were missed.

- `Transaction` and `KeyValueService` are wrapped such that we keep track of the path we are on. 

- these tests should make it harder for further additions to break the intended behaviour
- should give a template how to add more async paths

- have I covered everything
- are the changes to `SnapshotTransactionManager` allowed
- should I use something 

- `SnapshotTransactionTest`

- this week

### DIFF
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
@@ -18,19 +18,25 @@ package com.palantir.atlasdb.transaction.impl;
 
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
 
+import com.palantir.atlasdb.futures.AtlasFutures;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.transaction.api.Transaction;
-import com.palantir.common.base.Throwables;
 
 public class GetAsyncDelegate extends ForwardingTransaction {
 
     private final Transaction delegate;
+    private final SynchronousTracker tracker;
 
     public GetAsyncDelegate(Transaction transaction) {
         this.delegate = transaction;
+        this.tracker = SynchronousTracker.NO_OP;
+    }
+
+    public GetAsyncDelegate(Transaction transaction, SynchronousTracker tracker) {
+        this.delegate = transaction;
+        this.tracker = tracker;
     }
 
     @Override
@@ -40,12 +46,8 @@ public class GetAsyncDelegate extends ForwardingTransaction {
 
     @Override
     public Map<Cell, byte[]> get(TableReference tableRef, Set<Cell> cells) {
-        try {
-            return super.getAsync(tableRef, cells).get();
-        } catch (InterruptedException e) {
-            throw Throwables.rewrapAndThrowUncheckedException(e);
-        } catch (ExecutionException e) {
-            throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
+        try (SynchronousTracker resource = tracker.enterAsyncCall()) {
+            return AtlasFutures.getUnchecked(delegate().getAsync(tableRef, cells));
         }
     }
 }
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
@@ -27,14 +27,14 @@ import com.palantir.atlasdb.transaction.api.Transaction;
 public class GetAsyncDelegate extends ForwardingTransaction {
 
     private final Transaction delegate;
-    private final SynchronousTracker tracker;
+    private final PathTypeTracker tracker;
 
     public GetAsyncDelegate(Transaction transaction) {
         this.delegate = transaction;
-        this.tracker = SynchronousTracker.NO_OP;
+        this.tracker = PathTypeTracker.NO_OP;
     }
 
-    public GetAsyncDelegate(Transaction transaction, SynchronousTracker tracker) {
+    public GetAsyncDelegate(Transaction transaction, PathTypeTracker tracker) {
         this.delegate = transaction;
         this.tracker = tracker;
     }
@@ -46,7 +46,7 @@ public class GetAsyncDelegate extends ForwardingTransaction {
 
     @Override
     public Map<Cell, byte[]> get(TableReference tableRef, Set<Cell> cells) {
-        try (SynchronousTracker resource = tracker.enterAsyncCall()) {
+        try (PathTypeTracker resource = tracker.enterAsyncPath()) {
             return AtlasFutures.getUnchecked(delegate().getAsync(tableRef, cells));
         }
     }
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
@@ -25,13 +25,12 @@ import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.transaction.api.Transaction;
 
 public class GetAsyncDelegate extends ForwardingTransaction {
-
     private final Transaction delegate;
     private final PathTypeTracker tracker;
 
     public GetAsyncDelegate(Transaction transaction) {
         this.delegate = transaction;
-        this.tracker = PathTypeTracker.NO_OP;
+        this.tracker = PathTypeTrackers.NO_OP;
     }
 
     public GetAsyncDelegate(Transaction transaction, PathTypeTracker tracker) {
--- a/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
+++ b/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
@@ -94,8 +94,8 @@ public class AtlasDbTestCase {
         txManager = new CachingTestTransactionManager(serializableTxManager);
     }
 
-    private TrackingKeyValueService trackingKeyValueService(KeyValueService keyValueService) {
-        return spy(new TrackingKeyValueService(new StatsTrackingKeyValueService(keyValueService)));
+    private TrackingKeyValueService trackingKeyValueService(KeyValueService originalKeyValueService) {
+        return spy(new TrackingKeyValueService(new StatsTrackingKeyValueService(originalKeyValueService)));
     }
 
     protected TestTransactionManager constructTestTransactionManager() {
--- a/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
+++ b/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
@@ -59,7 +59,6 @@ public class AtlasDbTestCase {
     protected LockService lockService;
 
     protected final MetricsManager metricsManager = MetricsManagers.createForTests();
-    protected StatsTrackingKeyValueService keyValueServiceWithStats;
     protected TrackingKeyValueService keyValueService;
     protected InMemoryTimestampService timestampService;
     protected ConflictDetectionManager conflictDetectionManager;
@@ -76,10 +75,8 @@ public class AtlasDbTestCase {
         lockClient = LockClient.of("fake lock client");
         lockService = LockServiceImpl.create(LockServerOptions.builder().isStandaloneServer(false).build());
         timestampService = new InMemoryTimestampService();
-        KeyValueService kvs = getBaseKeyValueService();
-        keyValueServiceWithStats = new StatsTrackingKeyValueService(kvs);
-        keyValueService = spy(new TrackingKeyValueService(keyValueServiceWithStats));
-        TransactionTables.createTables(kvs);
+        keyValueService = trackingKeyValueService(getBaseKeyValueService());
+        TransactionTables.createTables(keyValueService);
         transactionService = TransactionServices.createRaw(keyValueService, timestampService, false);
         conflictDetectionManager = ConflictDetectionManagers.createWithoutWarmingCache(keyValueService);
         sweepStrategyManager = SweepStrategyManagers.createDefault(keyValueService);
@@ -92,7 +89,17 @@ public class AtlasDbTestCase {
     }
 
     private void setUpTransactionManagers() {
-        serializableTxManager = wrapTestTransactionManager(new TestTransactionManagerImpl(
+        serializableTxManager = constructTestTransactionManager();
+
+        txManager = new CachingTestTransactionManager(serializableTxManager);
+    }
+
+    private TrackingKeyValueService trackingKeyValueService(KeyValueService keyValueService) {
+        return spy(new TrackingKeyValueService(new StatsTrackingKeyValueService(keyValueService)));
+    }
+
+    protected TestTransactionManager constructTestTransactionManager() {
+        return new TestTransactionManagerImpl(
                 metricsManager,
                 keyValueService,
                 timestampService,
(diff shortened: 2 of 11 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
```diff
@@ -92,7 +89,24 @@ public void setUp() throws Exception {
...
-        serializableTxManager = wrapTestTransactionManager(new TestTransactionManagerImpl(
+        serializableTxManager = constructTestTransactionManager();
+
+        txManager = new CachingTestTransactionManager(serializableTxManager);
+    }
+
+    protected TrackingKeyValueService setUpKeyValueService() {
+        KeyValueService kvs = wrapKeyValueService(getBaseKeyValueService());
+        keyValueServiceWithStats = new StatsTrackingKeyValueService(kvs);
+        keyValueService = spy(new TrackingKeyValueService(keyValueServiceWithStats));
```
felixdesouza: this is a bad refactoring/at the very least it's confusing.

Have this method take in the `baseKeyValueService`, and then *don't* set the field in this method, and return the wrapped kvs

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
```diff
@@ -92,7 +89,24 @@ public void setUp() throws Exception {
     }
 
     private void setUpTransactionManagers() {
-        serializableTxManager = wrapTestTransactionManager(new TestTransactionManagerImpl(
+        serializableTxManager = constructTestTransactionManager();
+
+        txManager = new CachingTestTransactionManager(serializableTxManager);
+    }
+
+    protected TrackingKeyValueService setUpKeyValueService() {
```
felixdesouza:
```suggestion
    protected KeyValueService setUpKeyValueService() {
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
```diff
@@ -92,7 +89,24 @@ public void setUp() throws Exception {
...
+        KeyValueService kvs = wrapKeyValueService(getBaseKeyValueService());
+        keyValueServiceWithStats = new StatsTrackingKeyValueService(kvs);
+        keyValueService = spy(new TrackingKeyValueService(keyValueServiceWithStats));
+        return keyValueService;
+    }
+
+    protected KeyValueService wrapKeyValueService(KeyValueService trackingKeyValueService) {
```
felixdesouza:
```suggestion
    protected KeyValueService wrapKeyValueService(KeyValueService keyValueService) {
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
```diff
@@ -92,7 +89,24 @@ public void setUp() throws Exception {
...
+        return trackingKeyValueService;
```
felixdesouza:
```suggestion
        return keyValueService;
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -154,14 +162,20 @@
     @Parameterized.Parameters(name = "{0}")
     public static Collection<Object[]> data() {
         Object[][] data = new Object[][] {
-                {SYNC, UnaryOperator.identity()},
-                {ASYNC, (UnaryOperator<Transaction>) GetAsyncDelegate::new}
+                {SYNC,
+                 (BiFunction<Transaction, AtomicInteger, Transaction>) (transaction, counter) -> transaction,
```
felixdesouza: Let's introduce an interface, the type parameters are getting in the way:

```java
interface WrapperWithCounter<T> {
  T apply(T toWrap, AtomicInteger counter);
}
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -295,8 +313,20 @@ public void setUp() throws Exception {
     }
 
     @Override
-    protected TestTransactionManager wrapTestTransactionManager(TestTransactionManager testTransactionManager) {
-        return new WrappingTestTransactionManagerImpl(testTransactionManager, transactionWrapper);
+    protected TestTransactionManager constructTestTransactionManager() {
+        return new WrappingTestTransactionManagerImpl(
```
felixdesouza: why do we need to have a `constructTetsTransactionManager`, is it not sufficient to wrap as before? somewhat confused

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -422,28 +456,32 @@ public void testPutCleanup() throws Exception {
...
-                TransactionReadSentinelBehavior.THROW_EXCEPTION,
-                false,
-                timestampCache,
-                getRangesExecutor,
-                defaultGetRangesConcurrency,
-                MultiTableSweepQueueWriter.NO_OP,
-                MoreExecutors.newDirectExecutorService(),
-                true,
-                () -> transactionConfig));
+        AtomicInteger counter = new AtomicInteger();
```
felixdesouza: from what I can see, you don't really do anything with the counter besides pass it in, is it not sufficient to just include it in the delegate?

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1451,19 +1492,132 @@ private static SnapshotTransaction unwrapSnapshotTransaction(Transaction transac
...
+                            defaultGetRangesConcurrency,
+                            sweepQueueWriter,
+                            deleteExecutor,
+                            validateLocksOnReads,
+                            transactionConfig),
+                    counter);
+        }
+    }
+
+    private static class VerifyingDelegate implements AutoDelegate_KeyValueService {
```
felixdesouza:
```suggestion
    private static class VerifyingKeyValueServiceDelegate implements AutoDelegate_KeyValueService {
```

felixdesouza: change the constructors etc.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
```diff
@@ -41,11 +51,10 @@ public Transaction delegate() {
...
         try {
-            return super.getAsync(tableRef, cells).get();
-        } catch (InterruptedException e) {
-            throw Throwables.rewrapAndThrowUncheckedException(e);
-        } catch (ExecutionException e) {
-            throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
+            enterFunction.run();
+            return AtlasFutures.getUnchecked(delegate().getAsync(tableRef, cells));
+        } finally {
+            exitFunction.run();
```
jeremyk-91: This is possibly correct, but is a bit of a non-standard pattern: if `enterFunction.run()` throws, do we still want `exitFunction` to run? This depends on the semantics you want. Since it's tests it probably doesn't matter too much.

If we adopt Felix's suggestion (which I agree with), then this is a non-issue.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
```diff
@@ -18,19 +18,29 @@
...
+import com.palantir.atlasdb.futures.AtlasFutures;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.transaction.api.Transaction;
-import com.palantir.common.base.Throwables;
 
 public class GetAsyncDelegate extends ForwardingTransaction {
 
     private final Transaction delegate;
+    private final Runnable enterFunction;
```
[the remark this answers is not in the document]
felixdesouza: so I suppose I'm also confused, we have a counter to make sure that we're tracking synchronous gets, are we running things in parallel such that we expect this to be > 1? I don't believe `get` is at any point recursive?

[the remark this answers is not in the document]
jeremyk-91: I wouldn't be opposed to even having an `AtomicInteger` that is unused in the cases where you don't care about, since this lives in tests and I wouldn't be too concerned about perf.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1451,19 +1492,132 @@ private static SnapshotTransaction unwrapSnapshotTransaction(Transaction transac
...
+        private final KeyValueService delegate;
+        private final AtomicInteger atomicInteger;
```
jeremyk-91: This should ideally be better named, it's a bit hard to track what's going on / how this relates to things being synchronous or asynchronous (e.g. reading this class out of context would be super confusing).

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1451,19 +1492,132 @@ private static SnapshotTransaction unwrapSnapshotTransaction(Transaction transac
...
+                            defaultGetRangesConcurrency,
+                            sweepQueueWriter,
+                            deleteExecutor,
+                            validateLocksOnReads,
+                            () -> TRANSACTION_CONFIG),
+                    counter);
+        }
+
+        @Override
+        protected Transaction createTransaction(
```
felixdesouza: how come there's no wrapping going on here? what's the point of the counter then even?

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -295,8 +313,20 @@ public void setUp() throws Exception {
...
+                timestampService,
+                timestampService,
+                lockClient,
+                lockService,
+                transactionService,
+                conflictDetectionManager,
+                sweepStrategyManager,
+                sweepQueue,
+                MoreExecutors.newDirectExecutorService(),
+                transactionWrapper, keyValueServiceWrapper);
```
jeremyk-91: nit
```suggestion
                transactionWrapper, 
                keyValueServiceWrapper);
```

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1451,19 +1492,132 @@ private static SnapshotTransaction unwrapSnapshotTransaction(Transaction transac
         return (SnapshotTransaction) transaction;
     }
 
-    private static class WrappingTestTransactionManagerImpl extends WrappingTestTransactionManager {
-        private final Function<Transaction, Transaction> transactionWrapper;
+    private static class WrappingTestTransactionManagerImpl extends TestTransactionManagerImpl {
+        private final BiFunction<Transaction, AtomicInteger, Transaction> transactionWrapper;
```
jeremyk-91: At some point I'm starting to think that we should just have a special `UnaryOperator<Transaction>` that is a bit more descriptive. To me it's hard to follow what the `AtomicInteger`s are trying to do. Might be worth discussing offline

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91:
```suggestion
    private static class VerifyingDelegate implements ForwardingKeyValueService {
```

otherwise this could break workflows reliant on `getDelegates`

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1451,19 +1492,132 @@ private static SnapshotTransaction unwrapSnapshotTransaction(Transaction transac
...
+                            transactionConfig),
+                    counter);
+        }
+    }
+
+    private static class VerifyingDelegate implements AutoDelegate_KeyValueService {
+        private final KeyValueService delegate;
+        private final AtomicInteger atomicInteger;
+
+        VerifyingDelegate(KeyValueService keyValueService, AtomicInteger atomicInteger) {
```
felixdesouza: I think we should extract a class that wraps the `AtomicInteger`, and the methods are self describing,

e.g.

```java
interface SynchronousGetTracker {
  void checkInAsyncCall();
  void checkInSyncCall();
}
```

something along those lines, having the `counter` spread over a 1000 lines is not great

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Wrapper.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Wrapping doesn't work since it when we switch to delegate we lose control how transactions are created. Refactored `TestTransactionManagerImpl` to solve this.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: This is removed and we have an interface now for this.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Mistake, fixed.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Changed.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Changed to interface and now has a better name.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: Switched to an interface.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
OStevan: In general we don't expect this `enterFunction.run()` to throw, now it should be clearer with the refactoring also curious about what you think about this specific part.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/GetAsyncDelegate.java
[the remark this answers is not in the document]
OStevan: This is now resolved with an interface.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/AtlasDbTestCase.java
OStevan: Wrapping is now also obsolete since it is done differently. Removed `keyValueServiceWithStats`.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
OStevan: In general it has to be shared, even now with an interface.

PR #4370, opened 2019-10-31
[Conjure Java Runtime] Part 19: Retrying is Futile... Sometimes.

jeremyk-91:
**Goals (and why)**:
- Fix a bug in #4366: a `FastFailoverProxy` should *not* be created when creating a single node proxy, as that's futile.

- Remove `FastFailoverProxy` creation from the `createProxy` (no failover) codepath.
- Rename semantics of configuration in `TestProxyUtils` to make it obvious that the default params are with retrying enabled.

- Existing tests should work, and should be faster.

- In production things shouldn't be affected, as the single node clients created e.g. timelock's PaxosLearner etc. aren't gated behind an AwaitingLeadershipProxy. Does this make sense / do we need to perform more drastic action?

**Where should we start reviewing?**: ConjureJavaRuntimeTargetFactory

**Priority (whenever / two weeks / yesterday)**: yesterday 🔥Until this merges, all builds will probably be ten minutes slower than they should be.

### DIFF
--- a/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/http/TestProxyUtils.java
+++ b/atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/http/TestProxyUtils.java
@@ -29,11 +29,11 @@ public final class TestProxyUtils {
 
     public static final AuxiliaryRemotingParameters AUXILIARY_REMOTING_PARAMETERS_RETRYING
             = AuxiliaryRemotingParameters.builder()
-            .shouldLimitPayload(false)
-            .userAgent(UserAgent.of(UserAgent.Agent.of("bla", "0.1.2")))
-            .remotingClientConfig(() -> REMOTING_CLIENT_CONFIG)
-            .shouldRetry(true)
-            .build();
+                    .shouldLimitPayload(false)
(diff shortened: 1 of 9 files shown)


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/http/TestProxyUtils.java
```diff
@@ -27,13 +27,13 @@
...
 
-    public static final AuxiliaryRemotingParameters AUXILIARY_REMOTING_PARAMETERS
+    public static final AuxiliaryRemotingParameters AUXILIARY_REMOTING_PARAMETERS_RETRYING
             = AuxiliaryRemotingParameters.builder()
-                    .shouldLimitPayload(false)
-                    .userAgent(UserAgent.of(UserAgent.Agent.of("bla", "0.1.2")))
-                    .remotingClientConfig(() -> REMOTING_CLIENT_CONFIG)
-                    .shouldRetry(true)
-                    .build();
+            .shouldLimitPayload(false)
```
felixdesouza: strange indent

jeremyk-91: Fixed. Not sure what happened here

PR #4372, opened 2019-10-31
[Conjure Java Runtime] Part 20: The Beginning of the End / Force CJR for Inter-Node Communications

jeremyk-91:
**Goals (and why)**:
- Roll out our work on CJR!

- Switch `PaxosRemoteClients` to pass through configuration that always enables Conjure with probability 1.0, and disables the fallback mode.
- Include just enough wiring so that leader blocks still use Feign mode.

- Existing tests should suffice. There are currently failing tests at time of writing, but I would like to seek opinion on whether these tests should be retained / if retaining these tests should block rollout. There was an opinion that was changed from AtlasDB-Feign to CJR that we may not have realised initially.

- Am I taking too much risk? I believe not: we have tested both the correctness and performance of AtlasDB clients to TimeLock. Also, this change can be rolled back/forward **much** more cheaply than a similar failure on AtlasDB clients.
- Four existing tests fail - I have root caused this to a change in behaviour from AtlasDB-Feign to CJR. Specifically, AtlasDB-Feign would throw out socket exceptions while CJR attempts to retry them. This causes the tests in question to take much longer and eventually time out. I'm not sure how best to proceed; I'd claim that CJR's choice is reasonable and probably better, and fixing the tests will require plumbing configurable logic through TimeLock Server that will only be used for the tests, so I'm not so keen on that.

**Where should we start reviewing?**: PaxosRemoteClients

**Priority (whenever / two weeks / yesterday)**: this week would be great

Specific notes for each reviewer:
@felixdesouza Regarding leadership creation, is there any path I missed out? 
@gmaretic In general you know CJR the best, so I'm putting you generally on this!

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
@@ -42,6 +42,7 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.palantir.atlasdb.config.AuxiliaryRemotingParameters;
 import com.palantir.atlasdb.config.LeaderConfig;
 import com.palantir.atlasdb.config.LeaderRuntimeConfig;
+import com.palantir.atlasdb.config.RemotingClientConfig;
 import com.palantir.atlasdb.http.AtlasDbHttpClients;
 import com.palantir.atlasdb.http.NotCurrentLeaderExceptionMapper;
 import com.palantir.atlasdb.util.AtlasDbMetrics;
@@ -85,7 +86,7 @@ public final class Leaders {
             Supplier<LeaderRuntimeConfig> runtime,
             UserAgent userAgent) {
         LocalPaxosServices localPaxosServices = createInstrumentedLocalServices(
-                metricsManager, config, runtime, userAgent);
+                metricsManager, config, userAgent);
 
         env.accept(localPaxosServices.ourAcceptor());
         env.accept(localPaxosServices.ourLearner());
@@ -97,7 +98,6 @@ public final class Leaders {
     public static LocalPaxosServices createInstrumentedLocalServices(
             MetricsManager metricsManager,
             LeaderConfig config,
-            Supplier<LeaderRuntimeConfig> runtime,
             UserAgent userAgent) {
         Set<String> remoteLeaderUris = Sets.newHashSet(config.leaders());
         remoteLeaderUris.remove(config.localServer());
@@ -110,8 +110,8 @@ public final class Leaders {
         return createInstrumentedLocalServices(
                 metricsManager,
                 config,
-                runtime,
                 remotePaxosServerSpec,
+                () -> RemotingClientConfig.DEFAULT, // TODO (jkong): Wire this up or change it to Conjure
                 userAgent,
                 LeadershipObserver.NO_OP);
     }
@@ -119,8 +119,8 @@ public final class Leaders {
     public static LocalPaxosServices createInstrumentedLocalServices(
             MetricsManager metricsManager,
             LeaderConfig config,
-            Supplier<LeaderRuntimeConfig> runtime,
             RemotePaxosServerSpec remotePaxosServerSpec,
+            Supplier<RemotingClientConfig> remotingClientConfig,
             UserAgent userAgent,
             LeadershipObserver leadershipObserver) {
         UUID leaderUuid = UUID.randomUUID();
@@ -145,6 +145,7 @@ public final class Leaders {
                 metricsManager,
                 ourLearner,
                 remotePaxosServerSpec.remoteLearnerUris(),
+                remotingClientConfig,
                 trustContext,
                 PaxosLearner.class,
                 userAgent);
@@ -161,6 +162,7 @@ public final class Leaders {
                 metricsManager,
                 ourAcceptor,
                 remotePaxosServerSpec.remoteAcceptorUris(),
+                remotingClientConfig,
                 trustContext,
                 PaxosAcceptor.class,
                 userAgent);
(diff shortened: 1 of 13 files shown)


atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/RemotingClientConfig.java
```diff
@@ -28,6 +28,10 @@
 @Value.Immutable
 public interface RemotingClientConfig {
     RemotingClientConfig DEFAULT = ImmutableRemotingClientConfig.builder().build();
```
felixdesouza: default is poorly named, we should be explicit about what that means, somewhat unrelated to this PR but was confused when I was seeing it being used

jeremyk-91: I'll rename default to `ALWAYS_USE_LEGACY` (which is the current behaviour).

jeremyk-91: (and, goes without saying, but add explicit specifications in case the real default changes!)

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/LeadersTest.java
```diff
@@ -41,7 +43,8 @@
 
 public class LeadersTest {
 
-    public static final Set<String> REMOTE_SERVICE_ADDRESSES = ImmutableSet.of("foo:1234", "bar:5678");
+    private static final Set<String> REMOTE_SERVICE_ADDRESSES = ImmutableSet.of("foo:1234", "bar:5678");
+    private static final Supplier<RemotingClientConfig> REMOTING_CLIENT_CONFIG = () -> RemotingClientConfig.ALWAYS_USE_LEGACY;
```
gmaretic: These tests are not the best, given that it passes with the other RCC...

gmaretic: I don't even know what this is supposed to test, it's all broken mocks

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1008,7 +1008,7 @@ private static LockAndTimestampServices createRawLeaderServices(
                 .sslConfiguration(leaderConfig.sslConfiguration())
                 .build();
         ServiceCreator creator = ServiceCreator.noPayloadLimiter(
-                metricsManager, () -> serverListConfig, userAgent, () -> RemotingClientConfig.DEFAULT);
+                metricsManager, () -> serverListConfig, userAgent, () -> RemotingClientConfig.ALWAYS_USE_LEGACY);
```
gmaretic: Don't we want conjure here?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -110,17 +110,17 @@ public static LocalPaxosServices createInstrumentedLocalServices(
         return createInstrumentedLocalServices(
                 metricsManager,
                 config,
-                runtime,
                 remotePaxosServerSpec,
+                () -> RemotingClientConfig.ALWAYS_USE_LEGACY, // TODO (jkong): Wire this up or change it to Conjure
```
gmaretic: Why legacy?

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/RemotingClientConfig.java
```diff
@@ -27,7 +27,14 @@
 @JsonDeserialize(as = ImmutableRemotingClientConfig.class)
 @Value.Immutable
 public interface RemotingClientConfig {
-    RemotingClientConfig DEFAULT = ImmutableRemotingClientConfig.builder().build();
+    RemotingClientConfig ALWAYS_USE_LEGACY = ImmutableRemotingClientConfig.builder()
```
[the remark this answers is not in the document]
jeremyk-91: I'm not sure we came to a precise conclusion whether this was dangerous, but will change things to not do this.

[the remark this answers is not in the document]
jeremyk-91: Specifically, this can be done by instantiating a separate constant class.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/LeadersTest.java
jeremyk-91: This test just validates certain configurations (e.g. that a list of Paxos learners will include the local learner). The test doesn't actually test any CJR or Feign functionality, just that a remoting configuration is necessary.

I might as well make this CJR then.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: Not quite. This would be used by the embedded leader mode, which is still used in certain installations of internal auth product. I'm not as confident we have good test coverage for the way CJR works in the context of such situations - we'll need to do this soon, but didn't see this as in scope for this specific PR.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
jeremyk-91: As before, this is used by the embedded leader mode, which I fear we haven't tested as well under CJR yet, so would prefer to do this in another PR.

PR #4380, opened 2019-11-05
revert #4371

clockfort:
running into issues; it's probably this questionable design decision postgres jdbc made ( https://github.com/pgjdbc/pgjdbc/issues/1307 ) to just kinda ignore how SSL normally works in java


clockfort:
Yes, I'm also curious why none of the other SSL-enabled postgres testing hit this, like the downstream atlas product I tested this in previously.

PR #4382, opened 2019-11-06
Paging test

OStevan:
A fairly simple test to confirm paging logic works for multiple pages.


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientImpl.java
```diff
@@ -97,7 +97,7 @@ public static CqlClient create(
                 tuningConfig.preparedStatementCacheSize());
     }
 
-    private static CqlClient create(
+    static CqlClient create(
```
jeremyk-91: nit: There's an annotation `@VisibleForTesting` you can use for this kind of broadening that's only for tests

OStevan: This is wrong refactoring but I added it to the constructor.

PR #4383, opened 2019-11-06
Do not use RevAPI for Atlas

jeremyk-91:
**Goals (and why)**:
- Avoid confusing and time-consuming semantic merge conflicts, this caused publish to get borked a few times.
- RevAPI isn't particularly appropriate for AtlasDB, because we have a lot of internal APIs that for legacy reasons are public.

- Remove plugin
- Remove record of breaks

- Does anyone still want the record of breaks?

**Where should we start reviewing?**: It's short


felixdesouza:
I still think it's useful, and re semantic merge conflicts, I think that's been fixed, we just need to upgrade our version of gradle-revapi

j-baker:
argh sorry felix, didn't see your comment

j-baker:
please can we discuss?

PR #4384, opened 2019-11-06
Writes digest

felixdesouza:
**Goals (and why)**:
This is a function that can be called during some sort of post-mortem on the key value service for transactions that have seemingly violated the protocol. 

Given a row and a column, we will list all given writes to that in the kvs, and any unswept entries and their corresponding start timestamps.

We will then attempt to get the commit timestamps for all transactions/start timestamps seen. This also includes aborted transactions

If there are any transactions that are ongoing, we will list those start timestamps too.

We also list the written values at that time.

ASAP.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
@@ -1,0 +1,90 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.debug;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
+import com.palantir.atlasdb.keyvalue.api.Cell;
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.api.Value;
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
```diff
@@ -0,0 +1,90 @@
...
+
+        Multimap<Cell, Long> allWrittenCells =
+                keyValueService.getAllTimestamps(tableReference, ImmutableSet.of(asCell), Long.MAX_VALUE);
+
+        Collection<Long> allWrittenTimestamps = allWrittenCells.get(asCell);
+
+        Map<Long, Long> transactionCommitStatuses = transactionService.get(allWrittenTimestamps);
+        Map<Long, Long> completedOrAbortedTransactions = KeyedStream.stream(transactionCommitStatuses)
+                .filter(Objects::nonNull)
+                .collectToMap();
```
jeremyk-91: This shouldn't be needed - if I'm not mistaken, `transactionService#get()` will not return start timestamp keys of transactions that had neither committed nor aborted.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
```diff
@@ -0,0 +1,90 @@
...
+
+        Set<Long> inProgressTransactions = Sets.difference(
+                transactionCommitStatuses.keySet(),
+                completedOrAbortedTransactions.keySet());
```
jeremyk-91: Because of the above, the first argument here needs to be `allWrittenTimestamps`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
OStevan: Not the best solution, but `AsyncTransactionService#getAsync` should have the logic explained just in terms of futures.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
```diff
@@ -0,0 +1,90 @@
...
+import com.palantir.common.persist.Persistable;
+import com.palantir.common.streams.KeyedStream;
+
+public class WritesDigestEmitter {
+
+    private final TransactionService transactionService;
+    private final TableReference tableReference;
+    private final KeyValueService keyValueService;
+
+    public WritesDigestEmitter(TransactionManager transactionManager, TableReference tableReference) {
```
OStevan: How do we feel about a three argument package-private or private constructor and a static factory method? It seems like passing in the whole `TransactionManager` is more than we actually need.

felixdesouza:
Going to close this out and work on the changes in #4388

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/debug/WritesDigestEmitter.java
felixdesouza: The thing that constructs this (see #4388) has already 5 params (including the transaction manager), don't particularly want to make that thing 6, and if so, then changing the constructor/factory in `WritesDigestEmitter` is not particularly beneficial.

PR #4386, opened 2019-11-06
Bump gradle-rev-api to deal with semantic merge conflicts.

felixdesouza:
**Goals (and why)**:
An alternative proposal to #4383, the disruption of not being able to quickly release due to breaks, is mainly solved by 1.0.8 of gradle-rev-api. Acknowledging breaks is still useful despite the noise when you're breaking internal apis. We should iterate with them for things that don't work well for us.

Whenever.


felixdesouza:
Closing this since the other thing merged and this was never looked at.

PR #4391, opened 2019-11-07
Remove a bunch of deprecated and unused code

gmaretic:
So we don't have to keep maintaining it. Checked for usage in large internal product


felixdesouza:
💃

PR #4392, opened 2019-11-07
[LW] Part 2: Transaction Related LW APIs

gmaretic:
Only did a tiny bit of wiring, to give an idea of how it would fit together


lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -78,4 +81,14 @@ default void tryUnlock(Set<LockToken> tokens) {
...
 
+    // todo(gmaretic): implement
+    default StartTransactionWithWatchesResponse startTransactionWithWatches() {
+        return StartTransactionWithWatchesResponse.of(
+                startIdentifiedAtlasDbTransaction(), LockWatchState.of(ImmutableMap.of()));
+    }
+
+    // todo(gmaretic): implement
+    default TimestampedLockResponse acquireLocksForWrites(LockRequest lockRequest) {
+        return TimestampedLockResponse.of(getFreshTimestamp(), lock(lockRequest));
```
jeremyk-91: This is 1.5 RPCs, which is more than on the normal path

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionLockWatchingCacheView.java
```diff
@@ -0,0 +1,57 @@
...
+
+package com.palantir.atlasdb.transaction.api;
+
+import java.util.Map;
+import java.util.Set;
+
+import com.palantir.atlasdb.keyvalue.api.Cell;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+
+// todo (gmaretic): decide if the logic for what can be cached should live here or in the cache itself
```
jeremyk-91: I guess you mean the implementations of `tryCache(...)`? To me it feels like the cache should arbitrate such things - I'd perceive the role of a view as just letting you do `readCached` correctly - though there's room for discussion here.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionLockWatchingCacheView.java
```diff
@@ -0,0 +1,57 @@
...
+public interface TransactionLockWatchingCacheView {
+    /**
+     * Attempts to read the cached versions of cells. If there is no suitable cached value for the cell, it will not be
+     * present in the returned map.
+     *
+     * @param tableRef table to read from
+     * @param cells cells to read
+     * @return a mapping with all cells that were cached and guaranteed to be equal to what the transaction would have
+     * read from the kvs.
```
jeremyk-91: Based on this I figure that a view is created at some point after the transaction is initialised (so you have the start timestamp) and that there is one per transaction - is this right?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionLockWatchingCacheView.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    Map<Cell, byte[]> readCached(TableReference tableRef, Set<Cell> cells);
+
+    /**
+     * Try to cache values read from the kvs in this transaction. It is the responsibility of the caller to ensure the
+     * values can be cached (the corresponding locks were not open)
```
jeremyk-91: nit: `.`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1967,19 +1968,19 @@ private boolean rollbackOtherTransaction(long startTs, TransactionService transa
     /**
      * This method should acquire any locks needed to do proper concurrency control at commit time.
      */
-    protected LockToken acquireLocksForCommit() {
+    protected TimestampedLockResponse acquireLocksForCommit() {
         Set<LockDescriptor> lockDescriptors = getLocksForWrites();
 
         LockRequest request = LockRequest.of(lockDescriptors, transactionConfig.get().getLockAcquireTimeoutMillis());
-        LockResponse lockResponse = timelockService.lock(request);
+        TimestampedLockResponse lockResponse = timelockService.acquireLocksForWrites(request);
```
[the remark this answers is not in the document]
gmaretic: We can have it as an optional, where empty() will denote basically unknown timestamp, thus invalidating cached values

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
gmaretic: Fair enough, fixed

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionLockWatchingCacheView.java
gmaretic: Yeah, makes sense -- made the change

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionLockWatchingCacheView.java
gmaretic: yup

gmaretic: Also, this sentence made no sense so elaborated

PR #4396, opened 2019-11-08
Remove references to populate_io_cache_on_flush

leonz:
**Goals (and why)**:

The property "populate_io_cache_on_flush" had no functionality started in Cassandra 2.1 (see https://issues.apache.org/jira/browse/CASSANDRA-7493), and Cassandra 3 now throws a SyntaxException if a query has this property.

Removes any use of "populate_io_cache_on_flush".

Should `cachePriority` be removed also, or left as a no-op field (maybe deprecated)?  It doesn't seem to be used anywhere anymore, but it might be a breaking change to remove it.

### DIFF
--- a/changelog/@unreleased/pr-4396.v2.yml
+++ b/changelog/@unreleased/pr-4396.v2.yml
@@ -1,0 +1,11 @@
+type: improvement
+improvement:
+  description: |-
+    Remove all references to the Cassandra table property `populate_io_cache_on_flush`, which is supported or used in Cassandra.
+    <!---
+    Please remember to:
+    - Add any necessary release notes (including breaking changes)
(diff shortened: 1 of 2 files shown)


changelog/@unreleased/pr-4396.v2.yml
```diff
@@ -0,0 +1,11 @@
+type: improvement
+improvement:
+  description: |-
+    Remove all references to the Cassandra table property `populate_io_cache_on_flush`, which is supported or used in Cassandra.
+    <!---
+    Please remember to:
+    - Add any necessary release notes (including breaking changes)
+    - Make sure the documentation is up to date for your change
+    --->
```
jeremyk-91: This should not be part of the changelog

PR #4398, opened 2019-11-08
add metrics to scrubber

clockfort:
Add metrics to scrubbing, for the purposes of debugging a stack that may not be keeping up with its own scrub queue.

(Aside) I believe the eventual goal is to get rid of 90% of scrubbing and trust in targeted sweep, but in the interim this needs metrics.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/Scrubber.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/Scrubber.java
@@ -36,6 +36,7 @@ import javax.annotation.concurrent.GuardedBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.codahale.metrics.Meter;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.HashMultimap;
@@ -52,6 +53,7 @@ import com.google.common.collect.Sets;
 import com.google.common.primitives.UnsignedBytes;
 import com.google.common.util.concurrent.Futures;
 import com.palantir.atlasdb.AtlasDbConstants;
+import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.encoding.PtBytes;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.KeyAlreadyExistsException;
@@ -66,6 +68,7 @@ import com.palantir.atlasdb.transaction.api.TransactionFailedRetriableException;
 import com.palantir.atlasdb.transaction.api.TransactionManager;
 import com.palantir.atlasdb.transaction.impl.TransactionConstants;
 import com.palantir.atlasdb.transaction.service.TransactionService;
+import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.common.base.BatchingVisitable;
 import com.palantir.common.base.Throwables;
 import com.palantir.common.collect.Maps2;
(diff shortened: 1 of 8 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/Scrubber.java
```diff
@@ -172,6 +184,13 @@ private Scrubber(KeyValueService keyValueService,
         this.threadCount = threadCount;
         this.readThreadCount = readThreadCount;
         this.followers = followers;
+        this.metricsManager = metricsManager;
+
+        this.enqueuedCells = metricsManager.registerOrGetMeter(Scrubber.class, AtlasDbMetricNames.ENQUEUED_CELLS);
```
jkozlowski: Can we make those metrics be lazily registered? I am only mildly familiar with the Scrubber, but it looks like it's only used for AGGRESSIVE_HARD_DELETE?

```
 private void scrubForAggressiveHardDelete(SnapshotTransaction tx) {
        if ((tx.getTransactionType() == TransactionType.AGGRESSIVE_HARD_DELETE) && !tx.isAborted()) {
            // t.getCellsToScrubImmediately() checks that t has been committed
            cleaner.scrubImmediately(this,
                    tx.getCellsToScrubImmediately(),
                    tx.getTimestamp(),
                    tx.getCommitTimestamp());
        }
    }
```

```
if (getTransactionType() == TransactionType.AGGRESSIVE_HARD_DELETE
                || getTransactionType() == TransactionType.HARD_DELETE) {
            cleaner.queueCellsForScrubbing(getCellsToQueueForScrubbing(), getStartTimestamp());
        }

```

So we wouldn't want to add all of those always.

Have a look at SweepOutcomeMetrics or LegacySweepMetrics.

felixdesouza: we should add logic to this in `MetricsManager`/`TaggedMetricRegistry` or port it upstream to `tritium`. Some sort of proxy which wraps around the metric and only instantiates/registers it when it's first used. If any additional metrics have to do this themselves, it becomes unwieldy .

jeremyk-91: This is a good call, as Atlas-level metrics cost $$ in many deployments and we're trying to keep that down. 

Though as a heads up: this is confusingly named, but any services that use Cleanup Tasks (and thus Stream Stores) will invoke hard delete transactions, and thus register the metrics (so the savings might be less than we expect).

jkozlowski: Internal product creates the stream store, but does not use it in most deployments, so that wouldn't actually register the metrics right? In which case we need the laziness here and also making sure the scrubber doesn't put, say, zeros for those metrics (and thus register them) when it had nothing to do.

jkozlowski: @felixdesouza +1 for tracking something to make this easier, don't think it needs to block this PR.

clockfort: @jkozlowski Internal product uses the stream store for storing media, most all deployments would be using it.

I switched the new scrub metrics over to being lazily registered, though better upstream metrics library support transparent to the user would be better and surely result in additional cost savings for other less-used instrumented classes.

jkozlowski: @clockfort haha which internal product? the one I'm talking about only actually uses the stream store when installed on-prem, and the cost of those metrics we do not care about. Changes look good.

jeremyk-91: (@jkozlowski was probably referring to the fjord product, @clockfort the large internal product!)

clockfort: I know lol

PR #4401, opened 2019-11-11
Do not delegate for new API with default implementations

gmaretic:
**Goals (and why)**:
This was broken, as the generated code will call the default method of the delegate instead of the enclosing class


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -316,6 +316,7 @@ default boolean isInitialized() {
     TimelockService getTimelockService();
 
     // todo(gmaretic): implement
+    @DoNotDelegate
```
felixdesouza: technically, this is fine right? but for consistency sake/future proofing I'm guessing?

PR #4402, opened 2019-11-12
Cassandra 3 integration tests

leonz:
**Goals (and why)**:
To verify that Atlasdb also supports Cassandra 3.11.

The `build.gradle` for each module is the same as for the original modules, with the exception of using the Cassandra 3 dependency.  I added a gradle task to copy the source of the tests on build.

The containers I chose for the tests are arbitrary.  Let me know if there's a better way you'd like these to be organized.


scripts/circle-ci/run-circle-tests.sh
```diff
@@ -2,7 +2,7 @@
 
 set -x
 
-BASE_GRADLE_ARGS="--profile --continue"
+BASE_GRADLE_ARGS="--profile --no-daemon --continue"
```
felixdesouza: ?

leonz: I was getting random flakes without this (Exit code 137).  Did a bit of searching online, apparently can be caused if run without `--no-daemon` and Gradle docs specifically recommend running any CI with this option.

https://docs.gradle.org/3.3/userguide/gradle_daemon.html#when_should_i_not_use_the_gradle_daemon

PR #4403, opened 2019-11-13
[Conjure Java Runtime] Part 21: Global Conjuration

jeremyk-91:
**Goals (and why)**:
- Switch services using leader blocks to use CJR by default. 
- After this PR, all usage of AtlasDB, apart from tests or where explicitly configured, should go through CJR.

- Use CJR automatically if you are using a leader block.
- Change leader block codepaths to go through `RemoteLockServiceAdapter` where needed, as we cannot accept `null`s.

- We have many existing tests that run with leader blocks.

- Are the existing tests sufficient?

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -1042,7 +1042,7 @@ public abstract class TransactionManagers {
                 .sslConfiguration(leaderConfig.sslConfiguration())
                 .build();
         ServiceCreator creator = ServiceCreator.noPayloadLimiter(
-                metricsManager, () -> serverListConfig, userAgent, () -> RemotingClientConfigs.ALWAYS_USE_LEGACY);
+                metricsManager, () -> serverListConfig, userAgent, () -> RemotingClientConfigs.ALWAYS_USE_CONJURE);
         LockService remoteLock = creator.createService(LockService.class);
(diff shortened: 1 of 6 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -972,8 +973,13 @@ private static LockAndTimestampServices getLockAndTimestampServices(
             Optional<ClientLockDiagnosticCollector> lockDiagnosticCollector) {
         ServiceCreator creator = ServiceCreator.withPayloadLimiter(
                 metricsManager, timelockServerListConfig, userAgent, remotingConfigSupplier);
-        LockService lockService = RemoteLockServiceAdapter.create(
-                creator.createService(LockRpcClient.class), timelockNamespace);
+
+        // I'd use instrumentTimed(), but the timed annotations are not visible in LockService. Also, most practical
```
felixdesouza: I think you'll merge conflict with #4404 no?

felixdesouza: Or once #4404 merges, just change it here? no harm in it really

jeremyk-91: This shouldn't actually conflict, just that this won't use the smart features in #4404

jeremyk-91: Switched this call to `instrumentTimed()`.

PR #4404, opened 2019-11-13
[Conjure Java Runtime] Part 22: Martin Lewis's Muse

jeremyk-91:
**Goals (and why)**:
- Don't use the deprecated `instrument(...)` throughout `Leaders.java` which was a little unpleasant.
- Minor money-savings from deployments of products with internal leader-blocks.

- Move `@Timed` to `commons-annotations`. Previously `@Timed` was in `atlasdb-api`, which was sufficient for internal fjord product but not for us.
- Annotate various interfaces with the annotation. Reviewers should check these:
  - `CoordinationService`: `getLastKnownLocalValue` is not timed because it is a local read. The others have come up in past P0s before when debugging coordination.
  - `LeaderElectionService`: `markNotEligibleForLeadership` is a local atomic write. `stepDown` can involve network calls as it involves a Paxos proposal, but shouldn't be used frequently and definitely not on performance critical workflows. The others can do network calls and are useful for diagnosing leader election perf problems.
  - `PaxosAcceptor`, `PaxosLearner`, `PaxosProposer`, `PingableLeader`: All methods are timed. There aren't many, and can execute over remote contexts - and are generally helpful for debugging Paxos problems.

None.

- Did I accidentally mess up dependencies?
- Sadly this doesn't get to tagged metrics just yet. There'll be a second part which saves more money, I think both from TimeLock and from internal fjord product.

Each of the classes as named above, then `Leaders` and the other parts.

Whenever.


leader-election-api/src/main/java/com/palantir/leader/LeaderElectionService.java
```diff
@@ -51,13 +53,15 @@
      *
      * @return a leadership token to be used with {@link #isStillLeading}
      */
+    @Timed
```
felixdesouza: don't think this is interesting, as it includes time where you'd never become the leader, so metric is itself never useful.

leader-election-api/src/main/java/com/palantir/paxos/PaxosProposer.java
```diff
@@ -42,10 +45,12 @@
      * {@link PaxosProposer#propose(long, byte[])}. However, this value is proposed anonymously, with a fresh
      * proposer ID. This may be useful for relinquishing leadership in algorithms for that based on Paxos.
      */
+    @Timed
     byte[] proposeAnonymously(long seq, @Nullable byte[] proposalValue) throws PaxosRoundFailureException;
 
     /**
      * Returns a unique string identifier for the proposer.
      */
+    @Timed
```
felixdesouza: this is an in memory read

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -132,10 +132,10 @@ public static LocalPaxosServices createInstrumentedLocalServices(
                 leadershipObserver,
                 ImmutableList.of());
 
-        PaxosAcceptor ourAcceptor = AtlasDbMetrics.instrument(metricsManager.getRegistry(),
+        PaxosAcceptor ourAcceptor = AtlasDbMetrics.instrumentTimed(metricsManager.getRegistry(),
                 PaxosAcceptor.class,
```
jkozlowski: I won't check them all, but just be weary of those pesky autodelegate implementations...

leader-election-api/src/main/java/com/palantir/leader/LeaderElectionService.java
jkozlowski: yeah, we probably want something more targetted towards "how long until we elected a new leader"

jeremyk-91: good call, will remove

leader-election-api/src/main/java/com/palantir/paxos/PaxosProposer.java
jeremyk-91: doh, I thought this was a remote call, for some reason. Will remove

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
jeremyk-91: Yep. None of the 6 classes here are annotated with `@AutoDelegate`

PR #4405, opened 2019-11-14
[Timelock Partitioning] Part 46: Test reorganisation/refactor

felixdesouza:
**Goals (and why)**:
The current tests heavily make use of an implicit default namespace. That namespace is then used to do a lot of things, but more importantly it's as part of some readiness check i.e. check if the cluster is ready by grabbing a fresh timestamp etc.

I would like to reuse as many tests as possible, but run them under different configurations i.e. both single leader and multi leader configurations.

In the multi leader configuration, a default namespace now no longer serves as a viable readiness check. This is paving the way so that we can adapt existing tests/test setup 

They're some minor changes but still pretty close to source in terms of readability, but intent/semantics should be a lot clearer and it'd be harder to screw up etc.

* Move tests that don't really need multi-node configurations into the single node version
* Move tests that are really only applicable to single leader timelock into its own test class. Arguably, we could use `assume`? didn't really dig too deep into this, happy to try a better approach.
* Remove dead code
* Free up test specific cruft from `TestableTimelockCluster` and `TestableTimelockServer` by moving it into the corresponding test class.
* `TestableTimelockServer` and `TestableTimelockCluster` now require a namespace to be specified in order to do things with the cluster/server resp.
* Introduce some shared logic for constructing clients that also offers conveniences.
* Got rid of references to the old timestamp service. Can't really use it under normal operation, so everything just goes through timelock service.
* Also try to the use the upper most client where it makes sense i.e. use `TimelockService` vs `TimelockRpcClient` where possible. 
  * It's closer to how it will actually be used
  * It's no longer susceptible to rpc client "breaks", timelock service is still nice and high level "api" that we try to keep consistent.

This is all test code.

Some tests seemed to be doing weird things i.e. locking a lock with a random namespace, and then appearing to unlock that lock but only with the default namespace, which could explain why some tests take longer than they need to etc.

* `TestableTimelockCluster`
* `TestableTimelockServer`
* `TestProxies`
* Everything else, this will require a test by test scan, for the most part it's quite mechanical, but it's worth just double checking the semantics are still good.

ASAP.

### DIFF
--- a/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
+++ b/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -31,6 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import com.google.common.collect.ImmutableList;
@@ -38,6 +40,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Iterables;
 import com.palantir.atlasdb.transaction.impl.TransactionConstants;
+import com.palantir.common.concurrent.PTExecutors;
 import com.palantir.lock.HeldLocksGrant;
 import com.palantir.lock.HeldLocksToken;
 import com.palantir.lock.LockClient;
@@ -54,10 +57,10 @@ import com.palantir.lock.v2.LockRequest;
 import com.palantir.lock.v2.LockResponse;
 import com.palantir.lock.v2.LockResponseV2;
 import com.palantir.lock.v2.LockToken;
-import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionRequest;
 import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
 import com.palantir.lock.v2.StartTransactionRequestV4;
 import com.palantir.lock.v2.StartTransactionResponseV4;
+import com.palantir.lock.v2.TimelockService;
 import com.palantir.lock.v2.WaitForLocksRequest;
 import com.palantir.lock.v2.WaitForLocksResponse;
 import com.palantir.timestamp.TimestampRange;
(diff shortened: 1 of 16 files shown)


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -67,79 +76,105 @@
...
+        namespace.namespacedTimelockRpcClient().unlock(ImmutableSet.of(response2.getLock()));
     }
 
     @Test
     public void immutableTimestampIsGreaterThanFreshTimestampWhenNotLocked() {
-        long freshTs = cluster.getFreshTimestamp();
-        long immutableTs = cluster.timelockService().getImmutableTimestamp();
+        String randomNamespace = randomNamespace();
+        long freshTs = cluster.client(randomNamespace).getFreshTimestamp();
+        long immutableTs = cluster.client(randomNamespace).timelockService().getImmutableTimestamp();
```
jeremyk-91: I think we can just use the `namespace` constant here? I'd imagine we'd use that for anything where we aren't trying to test specific interactions across namespaces. Unless are we concerned about other tests doing shady things with the immutable timestamp

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -158,82 +193,82 @@ public void canRetrieveImmutableTimestampViaSyncApiForSyncServiceOrAsyncWithoutS
...
+                namespace.getFreshTimestamp(),
+                namespace.timelockService().getFreshTimestamp());
+
+        assertThat(timestamps).isSorted();
     }
 
     @Test
     public void canGetBatchTimestamps() {
-        TimestampRange range1 = cluster.getFreshTimestamps(5);
-        TimestampRange range2 = cluster.timelockService().getFreshTimestamps(5);
```
jeremyk-91: 🦆

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -188,228 +156,121 @@ public void lockServiceShouldNotAllowUsToRefreshLocksFromDifferentNamespaces() t
...
-        getFortyTwoFreshTimestamps(timestampService1);
+        getFortyTwoFreshTimestamps(namespace1.timelockService());
 
-        long firstServiceSecondTimestamp = timestampService1.getFreshTimestamp();
-        long secondServiceSecondTimestamp = timestampService2.getFreshTimestamp();
+        long firstServiceSecondTimestamp = namespace1.getFreshTimestamp();
+        long secondServiceSecondTimestamp = namespace2.getFreshTimestamp();
 
         assertThat(firstServiceSecondTimestamp - firstServiceFirstTimestamp).isGreaterThanOrEqualTo(FORTY_TWO);
         assertThat(secondServiceSecondTimestamp - secondServiceFirstTimestamp).isBetween(0L, (long) FORTY_TWO);
```
jeremyk-91: non-actionable: sigh, in practice this looks like a flake waiting to happen

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
```diff
@@ -188,228 +156,121 @@ public void lockServiceShouldNotAllowUsToRefreshLocksFromDifferentNamespaces() t
...
         assertThatThrownBy(() -> lockService.getMinLockedInVersionId(CLIENT_1))
                 .isInstanceOf(Exception.class);
     }
 
-    private static void getFortyTwoFreshTimestamps(TimestampService timestampService) {
+    private static void getFortyTwoFreshTimestamps(TimelockService timelockService) {
         for (int i = 0; i < FORTY_TWO; i++) {
-            timestampService.getFreshTimestamp();
+            timelockService.getFreshTimestamp();
         }
```
jeremyk-91: 😢

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
felixdesouza: I was getting test failures because we couldn't acquire the lock. I'm thinking to actually make `namespace` be a randomNamespace at the beginning, then we don't have to worry about sharing stuff.

jeremyk-91: OK, yeah so looks like the other tests _do_ do shady things

felixdesouza: changed so that each test has a random namespace to do stuff

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/PaxosTimeLockServerIntegrationTest.java
felixdesouza: hasn't flaked so far!!

PR #4408, opened 2019-11-14
Upgrade CJR

gmaretic:


felixdesouza:
Can you amend the description to include why we're bumping it outside of our broken upgrade dependencies excavator? :P

jeremyk-91:
I updated the changelog!

PR #4409, opened 2019-11-14
[Conjure Java Runtime] Part 24: Removing Feign From Jepsen, TM Test and ADHC Test

jeremyk-91:
**Goals (and why)**:
- Switch all remaining test-only paths (plus a usage in the large internal product) away from Feign.
- After this PR merges, all uses of AtlasDB on develop should use Conjure Java Runtime.

- Switch the last two methods in `AtlasDbHttpClients` to use CJR.
- Clean up necessary tests.

For the most part we can rely on what's already there. Though see the Concerns section

- `TransactionManagersTest`: I removed the test for the Payload Limiter (because CJR does not use them). I also removed the test for canUseCJR since the tests now use that, and I don't think it's necessary to maintain a canUseFeign codepath. 
- (possibly more contentious) I also removed the test for the ProxyConfiguration since we just pass this through to CJR, and I had a lot of difficulty getting WireMock to work nicely.
- Jepsen: Please sanity check my Clojure syntax, it's been a long while...

**Where should we start reviewing?**: `AtlasDbHttpClients`

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
@@ -21,21 +21,16 @@ import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 import com.palantir.atlasdb.config.AuxiliaryRemotingParameters;
 import com.palantir.atlasdb.config.ServerListConfig;
 import com.palantir.atlasdb.http.VersionSelectingClients.VersionSelectingConfig;
 import com.palantir.atlasdb.http.v2.ConjureJavaRuntimeTargetFactory;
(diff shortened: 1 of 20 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -21,22 +21,17 @@
...
 import com.google.common.annotations.VisibleForTesting;
 import com.palantir.atlasdb.config.AuxiliaryRemotingParameters;
 import com.palantir.atlasdb.config.ServerListConfig;
 import com.palantir.atlasdb.http.VersionSelectingClients.VersionSelectingConfig;
 import com.palantir.atlasdb.http.v2.ConjureJavaRuntimeTargetFactory;
-import com.palantir.atlasdb.util.AtlasDbMetrics;
 import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.conjure.java.config.ssl.TrustContext;
 
 public final class AtlasDbHttpClients {
```
felixdesouza: soon this class will disappear in its entirety 💃

jeremyk-91:
We paired on the fix-up commits. Thanks for your help here! 🚀

PR #4410, opened 2019-11-15
Fun with AutoDelegates <3

gmaretic:
**Goals (and why)**:
Yet again, we were bypassing delegates' implementations 🗡


gmaretic:
Thanks @jeremyk-91 !

PR #4413, opened 2019-11-15
[Conjure Java Runtime] Part 25: Fast Failover Proxy Needs To Be Used On Non-Live-Reloadable Paths

jeremyk-91:
**Goals (and why)**:
- Stop build flakes
- Ensure large internal product usage on non-live-reloadable paths.

- Wrap proxies with failover.

- Should reduce number of flakes. I can run a few builds to test this.

- Nothing in particular.


felixdesouza:
💃

Issue #4412, opened 2019-11-15
Reduce connect timeout in ClientOptions


gmaretic:
The current connect timeout in ClientOptions for CJR clients is 10 seconds, this is consistent with legacy behaviour, but there is no real reason to keep it that high since if a timelock node goes down, this can make us stuck for 10 seconds before we even try another node.

jkozlowski:
+1, this should have rather large upside

PR #4416, opened 2019-11-17
Lower timeouts.

jkozlowski:
**Goals (and why)**:

Do not merge as is, this is for discussion. 

### DIFF
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
@@ -36,8 +36,8 @@ public abstract class ClientOptions {
     // TODO (jkong): Re-enable client QoS after response body leaks are handled correctly.
     // Throws after expected outages of 1/2 * 0.01 * (2^11 - 1) = 10.24 s
     public static final ClientOptions DEFAULT_RETRYING = ImmutableClientOptions.builder()
-            .connectTimeout(Duration.ofSeconds(10))
-            .readTimeout(Duration.ofSeconds(65))
+            .connectTimeout(Duration.ofSeconds(1))
+            .readTimeout(Duration.ofSeconds(10))
             .backoffSlotSize(Duration.ofMillis(10))
             .failedUrlCooldown(Duration.ofMillis(100))
             .maxNumRetries(11)


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -36,8 +36,8 @@
     // TODO (jkong): Re-enable client QoS after response body leaks are handled correctly.
     // Throws after expected outages of 1/2 * 0.01 * (2^11 - 1) = 10.24 s
     public static final ClientOptions DEFAULT_RETRYING = ImmutableClientOptions.builder()
-            .connectTimeout(Duration.ofSeconds(10))
-            .readTimeout(Duration.ofSeconds(65))
+            .connectTimeout(Duration.ofSeconds(1))
```
jkozlowski: This is as per https://github.com/palantir/atlasdb/issues/4412; I would probably have an even lower timeout, most timelock calls take a few single digit milliseconds.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -36,8 +36,8 @@
...
+            .readTimeout(Duration.ofSeconds(10))
```
jkozlowski: Same here, we should probably lower even more, as this prolongs any outage: timelock is generally fast, so even seconds here are probably too high.

jkozlowski: So the problem here is that lock V1 API blocks until the lock is granted, so it is legitimate.

felixdesouza: they don't use conjure or atlasdb-feign, they have their own remoting layer

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
jeremyk-91: This is fine, yeah I'd have a lower timeout.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
jeremyk-91: @felixdesouza Lock V1 is actually used by anyone who does `runTaskWithLocks(...)` - there are some users in internal metal product that explicitly take out lock-v1 locks too.

felixdesouza:
Can we fill in the PR description to give backstory as to why we're making these changes, what we've seen/why the current thing isn't working as well as it shouldn't etc.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
jkozlowski: I think this is might be the actual thing that prolongs outages though. Since Timelock runs on http2, the connection would've been already established. I think we need a bit of testing of the client side to see what this actually does.

felixdesouza:
This has been addressed in later PRs #4517 #4519 #4520

PR #4425, opened 2019-11-19
Fix develop

felixdesouza:
**Goals (and why)**:
Handling a semantic merge conflict between #4405 and #4340. The latter hooks up closing/shutting down of resources. The former actually makes the timelock server spun up in integration tests run that code, both which weren't present 

`DisruptorAutobatcher` doesn't shutdown `ExecutorService`s :(. Instead it waits for the backlog of tasks to complete, and then stops the handler. This is problematic if the tasks contains blocking actions i.e. `LeaderElectionService#blockOnBecomingLeader`.

There doesn't seem to be a way to signal to a handler effectively that it needs to shutdown, I imagine due to perf reasons. Not trying to solve this problem just yet.

With the introduction of `stepDown` and leader ineligibility, it's one more point to ensure that the API in `LeaderElectionService` is respected. I moved the eligibility check further up so we could stop pinging if we're ineligible, reason being if we're not eligible and we fail to ping, there's nothing we can do anyway. However this goes in a loop with what calls it. In reality it needs to throw an `InterruptedException` since that's the contract of `LeaderElectionService#blockOnBecomingLeader` and the only way to signal that something has not gone right.

* If `AwaitingLeadershipProxy` is closed, including the underlying `LeaderElectionService` it will keep trying to gain leadership which will be fruitless, gives back some cpu.

* Since we're now stepping down when we close, we want to ensure that we can't break shutting down: if we fail to step down e.g. not enough nodes online, we will then just log and continue the shutdown process.

making existing tests pass.

Does the `PaxosLeaderElectionService` change "respect" the contract?

Commit by commit.

ASAP if you wanna merge anything in.

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -19,7 +19,11 @@ package com.palantir.atlasdb.autobatch;
 import java.io.Closeable;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.AsyncFunction;
@@ -28,6 +32,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.lmax.disruptor.EventHandler;
 import com.lmax.disruptor.RingBuffer;
+import com.lmax.disruptor.TimeoutException;
 import com.lmax.disruptor.dsl.Disruptor;
 import com.palantir.logsafe.Preconditions;
 
@@ -37,6 +42,8 @@ import com.palantir.logsafe.Preconditions;
  */
 public final class DisruptorAutobatcher<T, R>
         implements AsyncFunction<T, R>, Function<T, ListenableFuture<R>>, Closeable {
+
+    private static final Logger log = LoggerFactory.getLogger(DisruptorAutobatcher.class);
 
     /*
         By memoizing thread factories per loggable purpose, the thread names are numbered uniquely for multiple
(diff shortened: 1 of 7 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -81,7 +88,12 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
     @Override
     public void close() {
         closed = true;
-        disruptor.shutdown();
+        try {
+            disruptor.shutdown(10, TimeUnit.SECONDS);
```
gmaretic: Not sure it even makes sense to do the shutdown then, but this is fine since we can see some logging if we are hitting it at least

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -123,7 +123,7 @@ private void proposeLeadershipOrWaitForBackoff(LeadershipState currentState) thr
             if (leaderEligibilityLoggingRateLimiter.tryAcquire()) {
                 log.debug("Not eligible for leadership");
             }
-            return;
+            throw new InterruptedException("leader no longer eligible");
```
gmaretic: Document this behaviour on the interface?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/LeadershipComponents.java
```diff
@@ -103,7 +103,11 @@ private LeadershipContext createNewLeadershipContext(Client client) {
 
     private static void shutdownLeaderElectionService(LeaderElectionService leaderElectionService) {
         leaderElectionService.markNotEligibleForLeadership();
-        leaderElectionService.stepDown();
+        try {
+            leaderElectionService.stepDown();
```
gmaretic: Should the implementation ever throw?

leader-election-impl/src/test/java/com/palantir/paxos/PaxosConsensusFastTest.java
```diff
@@ -261,7 +263,11 @@ public void markNotEligibleForLeadership() throws InterruptedException {
         assertThat(state.leader(0).isStillLeading(token)).isEqualTo(StillLeadingStatus.NOT_LEADING);
         assertThat(state.leader(1).isStillLeading(token2)).isEqualTo(StillLeadingStatus.LEADING);
 
-        assertThat(future.cancel(true)).isTrue();
+        assertThat(future).isDone();
```
gmaretic: ccc

PR #4426, opened 2019-11-19
[Timelock Partitioning] Part 48: Make use of templates for config whilst testing.

felixdesouza:
**Goals (and why)**:
There are quite a few configurations that we will end up testing. In the three node configuration, as it stands we have 3 separate configuration files for each node in the cluster and then for each minor modification/variation another set. This leads to configuration file rot where if we fix a bug/add useful logging/loggers, only a particular set of tests get it vs the other.

* Use freemarker to do the templating.
* You pass in `TemplateVariables` to the cluster, it will then set up the right number of nodes, and generate the correct set of templates that does the right thing.

* Existing tests should pass.

It's a bit "magic" but then again so is referencing random files and hoping they all work. It would be nice to serialise the config as is, but there's a lot of cruft which I don't think is serialisable and is also annoying to write, this feels like a good middle ground.

* `TemporaryConfigurationHolder`
* `TestableTimelockCluster`
* `TemplateVariables`
* `PaxosSuite`
* `paxosSingleServer.ftl`, `paxosMultiServer.ftl`

This week please.


timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TemporaryConfigurationHolder.java
```diff
@@ -16,59 +16,59 @@
...
+        Template template = TEMPLATE_CONFIG.getTemplate(templateName);
+        template.process(
+                variables.withDataDirectory(temporaryFolder.newFolder().getAbsolutePath()),
+                new FileWriter(temporaryConfigFile));
     }
 
     String getTemporaryConfigFileLocation() {
         return temporaryConfigFile.getPath();
     }
+
```
OStevan: Is this intentional?

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TestableTimelockCluster.java
```diff
@@ -56,35 +52,28 @@
...
+        this(name, configFileTemplate, ImmutableList.copyOf(variables));
     }
 
-    public TestableTimelockCluster(ClusterName clusterName, String baseUri, String... configFileTemplates) {
-        this.clusterName = clusterName.get();
-        this.configs = Arrays.stream(configFileTemplates)
-                .map(this::getConfigHolder)
+    public TestableTimelockCluster(String name, String configFileTemplate, Iterable<TemplateVariables> variables) {
+        this.name = name;
+        this.configs = Streams.stream(variables)
```
OStevan: Can we change this to `Set<TemporaryConfigurationHolder>`?

felixdesouza: any particular reason why?

felixdesouza: @OStevan ping on this

PR #4427, opened 2019-11-21
[WIP] Checkstyle update

OStevan:
**Goals (and why)**:
Update check style definitions to be usable with IDE check style plugin.

whenever


OStevan:
Discussed with @felixdesouza this PR is not easily reviewable since the commit history is not sufficiently clear for a big diff required by this PR.

PR #4430, opened 2019-11-21
[LW] Part ???: The Boring Part (APIs, wiring)

gmaretic:
Atlas side things will still change, but this is mostly to separate the plumbing part from the actual meat of the server side impl


lock-api/src/main/java/com/palantir/lock/watch/LockEvent.java
```diff
@@ -0,0 +1,44 @@
...
+public abstract class LockEvent implements LockWatchEvent {
+    public abstract LockToken lockToken();
+    public abstract Set<LockDescriptor> lockDescriptors();
+
+    @Override
+    public void accept(LockWatchEventVisitor visitor) {
+        visitor.visit(this);
+    }
+
+    public static Function<Long, LockWatchEvent> fromSeq(LockToken lockToken, Set<LockDescriptor> lockDescriptors) {
```
OStevan: Does it make more sense to have a parallel `Builder` hierarchy for `LockWatchEvents`?

OStevan: The previous comment was a bit short and not very informative. We can use the fact that generated code with `@Value.Immutable` will generate a `ImmutableLockEvent.Builder` which would implement a `LockEvent.Builder` so instead of `fromSeq` we can have a static `builder` method. This would at least help us to remove `Function<Long, LockWatchEvent>` for something more intuitive.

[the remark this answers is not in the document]
felixdesouza: is it not sufficient to just have 3 parameters?

Or as @OStevan says, have a `Builder` hierarchy, and have this return a general `Builder`, which just needs the sequence to be completed?

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEvent.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.watch;
+
+public interface LockWatchEvent {
```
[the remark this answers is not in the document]
felixdesouza: it would look like this:

```java
@Data
abstract class LockWatchEvent {
  interface EventTypes<R> {
    R lockWatchCreatedEvent(long sequence, LockWatchRequest request);
    R lockEvent(long sequence, LockToken lockToken, Set<LockDescriptor> lockDescriptors);
    R unlockEvent(long sequence, LockToken lockToken, Set<LockDescriptor> lockDescriptors);
  }

  public abstract <R> R match(EventTypes<R> eventTypes);

  @Override
  public abstract int hashCode();

  @Override
  public abstract boolean equals(Object obj);

  @Override
  public abstract String toString();
}
```

It generates you all the different classes, and gives you convenient accessors, although you can just define methods here if you need as well.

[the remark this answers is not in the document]
felixdesouza: From `LeaderPingResult`, here's an example usage:

```java
    public boolean isSuccessful() {
        return LeaderPingResults.caseOf(this)
                .pingReturnedTrue_(true)
                .otherwise_(false);
    }
```

It has useful lens accessors e.g. for `LeaderPingResult` 
```java
static Optional<Throwable> getException(LeaderPingResult leaderPingResult);
static Function<LeaderPingResult, LeaderPingResult> setException(Throwable newException);
static Function<LeaderPingResult, LeaderPingResult> modException(
      Function<Throwable, Throwable> exceptionMod);
```

`getException` is optional because it's not on each case, but if an attribute *is* on each case, then you get the unwrapped value, lens ftw.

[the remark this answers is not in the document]
felixdesouza: reference material: https://github.com/derive4j/derive4j

lock-api/src/main/java/com/palantir/lock/watch/LockEvent.java
gmaretic: Maybe, but I don't really want to expose a generic builder. Maybe if we made it a staged builder

lock-api/src/main/java/com/palantir/lock/AtlasLockDescriptorRanges.java
```diff
@@ -0,0 +1,75 @@
...
+    public static byte[] createEndNameForPrefixScan(byte[] rowName) {
+        for (int i = rowName.length - 1; i >= 0; i--) {
+            if ((rowName[i] & 0xff) != 0xff) {
+                byte[] ret = new byte[i + 1];
+                System.arraycopy(rowName, 0, ret, 0, i + 1);
+                ret[i]++;
+                return ret;
+            }
+        }
+        throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
gmaretic: Sure, this is unreachable if you use it with a `String`, so I didn't bother too much

lock-api/src/main/java/com/palantir/lock/watch/LockWatchingRpcClient.java
```diff
@@ -40,5 +38,5 @@
     // as usual, GET is allowed to be cached, and we are not allowed to use a stale result
     @POST
     @Path("watch-state")
-    LockWatchState getWatchState(@PathParam("namespace") String namespace, UUID serviceId);
+    LockWatchStateUpdate getWatchState(@PathParam("namespace") String namespace);
```
OStevan: Should this be renamed to `getWatchStateUpdate`?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java
```diff
@@ -76,6 +80,12 @@ public TimestampRange getFreshTimestamps(@Safe @QueryParam("number") int numTime
         return timelock.getFreshTimestamps(numTimestampsRequested);
     }
 
+    @POST
+    @Path("commit-timestamp")
```
OStevan: Maybe `watched-commit-timestamp` ?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/AsyncLockService.java
```diff
@@ -47,6 +48,7 @@
     private final ImmutableTimestampTracker immutableTsTracker;
     private final LeaderClock leaderClock;
     private final LockLog lockLog;
+    private final LockWatchingService lockWatchingService = null;
```
jeremyk-91: Would it be reasonable to return a null object rather than null? i.e. something that does nothing and throws on getWatchState()

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TableElement.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.api.watch;
+
+import com.google.common.collect.Range;
+import com.palantir.lock.LockDescriptor;
+
+public interface TableElement {
```
j-baker: this doesn't seem like the right name for this type - it's a really generic type name to describe really specific behaviour. I don't know quite what the right name is, but how about LockRange?

j-baker: also - is this expected to be json serializable?

j-baker: also - I've found the Guava Range types to cause unnecessary pain, because once you've created one you can't actually back out what the range is describing.

It makes sense as an implementation detail of timelock, but feels weirder as something to slot in the Atlas API - it's like, we probably don't want anyone who's not us using this implementation - and it's not as though people can define their own ranges.

For that reason, maybe worth creating a visitor in this package and converting into Ranges within timelock, rather than in the atlasdb-api package?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TableWatchingService.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Registers watches.
+     */
+    @Idempotent
+    void registerWatches(Set<TableElement> lockWatchEntries);
+
+    /**
+     * Deregisters watches.
+     */
+    void deregisterWatches(Set<TableElement> lockWatchEntries);
```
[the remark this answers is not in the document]
j-baker: I think we said we'd probably not implement the ability to deregister watches, thus far?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/KvsLockWatchingService.java
```diff
@@ -16,34 +16,25 @@
...
-import com.palantir.lock.watch.WatchId;
+import com.palantir.lock.watch.LockWatchRequest;
 
 public interface KvsLockWatchingService {
     /**
-     * Registers watches for a set of rows in a single table.
-     * @param tableRef table to register watches for.
-     * @param rowNames rows to register watches for.
-     * @return a mapping of {@link WatchId}s to the {@link RowReference} for which the watches have been registered
+     * Registers watches.
```
j-baker: at this point, there's no KvsLockWatchingService, right? shoudl this PR delete that class?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/LockWatchState.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.api;
+
+public interface LockWatchState {
```
j-baker: you have two of these in this PR

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -155,7 +156,8 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
 
     @Override
     public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition) {
-        StartTransactionWithWatchesResponse responseWithWatches = timelockService.startTransactionWithWatches();
+        StartTransactionWithWatchesResponse responseWithWatches = timelockService
+                .startTransactionWithWatches(OptionalLong.empty());
```
j-baker: I don't think TimelockService should gain a method that's related to lock watches - rather we should be maintaining this state as some part of the hierarchy _below_ timelock service.

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEvent.java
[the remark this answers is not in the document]
j-baker: don't think there's much value from this

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventVisitor.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.watch;
+
+public interface LockWatchEventVisitor {
```
j-baker: it's typical to place visitors in the parent class e.g. this should be LockWatch.Visitor

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -86,14 +87,19 @@ default void tryUnlock(Set<LockToken> tokens) {
 
     // todo(gmaretic): implement
     @DoNotDelegate
-    default StartTransactionWithWatchesResponse startTransactionWithWatches() {
-        return StartTransactionWithWatchesResponse.of(
-                startIdentifiedAtlasDbTransaction(), LockWatchState.of(ImmutableMap.of()));
+    default StartTransactionWithWatchesResponse startTransactionWithWatches(OptionalLong lastKnownState) {
```
j-baker: I don't think TimelockService needs to know about lock watches. The separation we have at the moment is a TimelockService class which requires methods used by our API consumers, and the methods required within txns which are held separately. Given SnapshotTransaction and similar should not need lock watches (we are layering this above that level), we needn't make these changes here.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TableElements.java
```diff
@@ -0,0 +1,122 @@
...
+    @Value.Immutable
+    @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+    abstract static class RowPrefix implements TableElement {
+        @Value.Parameter
+        abstract TableReference tableRef();
+        @Value.Parameter
+        abstract byte[] row();
+
+        @Override
+        public Range<LockDescriptor> getAsRange() {
```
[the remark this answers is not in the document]
gmaretic: It's a matter of preference, I guess. We don't really make methods final unless there is a good reason to do so

lock-api/src/main/java/com/palantir/lock/AtlasLockDescriptorRanges.java
```diff
@@ -0,0 +1,72 @@
...
+                AtlasRowLockDescriptor.of(qualifiedTableName, startInc),
+                AtlasRowLockDescriptor.of(qualifiedTableName, endExc));
+    }
+
+    public static Range<LockDescriptor> exactRow(String qualifiedTableName, byte[] row) {
+        LockDescriptor descriptor = AtlasRowLockDescriptor.of(qualifiedTableName, row);
+        return Range.closed(descriptor, descriptor);
+    }
+
+    public static Range<LockDescriptor> exactCell(String qualifiedTableName, byte[] row, byte[] col) {
```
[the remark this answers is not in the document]
gmaretic: As Jeremy mentioned, we don't have the dependency here

lock-api/src/main/java/com/palantir/lock/AtlasLockDescriptorRanges.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock;
+
+import com.google.common.collect.Range;
+
+public final class AtlasLockDescriptorRanges {
+    private AtlasLockDescriptorRanges() {
```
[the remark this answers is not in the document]
gmaretic: I mean, we could move it, but this is really the correct place for it, given that it is also where the LockDescriptor classes for row and cell live

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TableWatchingService.java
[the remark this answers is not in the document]
gmaretic: This was here before the decision was made, current implementation just throws UOE, but I'll remove it to simplify

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimeLockServices.java
```diff
@@ -27,17 +28,20 @@ static TimeLockServices create(
...
                 .timestampManagementService(timestampManagementService)
                 .timelockService(timelockService)
+                .lockWatchingResource(lockWatchingResource)
                 .build();
     }
 
-    TimestampManagementService getTimestampManagementService();
     TimestampService getTimestampService();
-    AsyncTimelockResource getTimelockService();
     LockService getLockService();
```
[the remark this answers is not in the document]
gmaretic: Changing the ordering of arguments in a factory method, especially when some of them are supertypes of others, just seems unnecessarily fancy and dangerous

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TableWatchingService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Returns the current state of all registered watches.
+     */
+    TableWatchState getLockWatchState(Optional<Long> lastKnownState);
```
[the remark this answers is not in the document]
gmaretic: Yeah, I changed this in a subsequent PR and forgot to propagate all the way back

lock-api/src/main/java/com/palantir/lock/AtlasLockDescriptorRanges.java
```diff
@@ -0,0 +1,72 @@
...
+    public static Range<LockDescriptor> exactCell(String qualifiedTableName, byte[] row, byte[] col) {
+        LockDescriptor descriptor = AtlasCellLockDescriptor.of(qualifiedTableName, row, col);
+        return Range.closed(descriptor, descriptor);
+    }
+
+    private static byte[] bytesForTableName(String tableName) {
+        return tableName.getBytes();
+    }
+
+    private static byte[] createExclusiveEndNameForPrefixScan(byte[] prefix) {
```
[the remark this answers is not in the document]
gmaretic: I'm still not convinced that's the proper place for this class

lock-api/src/main/java/com/palantir/lock/v2/TimelockService.java
gmaretic: Makes sense to only have this on the RPC client and wire it into TransactionStarter and have something appropriate for commit as well

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventVisitor.java
gmaretic: 👍

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/AsyncLockService.java
gmaretic: I didn't bother because there is already a subsequent PR that wires in the impl

lock-api/src/main/java/com/palantir/lock/watch/LockWatchReferences.java
```diff
@@ -0,0 +1,198 @@
...
+            @JsonSubTypes.Type(value = ImmutableEntireTable.class, name = EntireTable.TYPE),
+            @JsonSubTypes.Type(value = ImmutableRowPrefix.class, name = RowPrefix.TYPE),
+            @JsonSubTypes.Type(value = ImmutableRowRange.class, name = RowRange.TYPE),
+            @JsonSubTypes.Type(value = ImmutableExactRow.class, name = ExactRow.TYPE),
+            @JsonSubTypes.Type(value = ImmutableExactCell.class, name = ExactCell.TYPE)})
+    public interface LockWatchReference {
+        <T> T accept(Visitor<T> visitor);
+    }
+
+    public interface Visitor<T> {
```
jeremyk-91: Discussed offline. This is not exactly what I would expect in terms of the class hierarchy, but is fine and we can change later on if needed.

PR #4431, opened 2019-11-21
[PDS-10{3841,4895}] Disable Cleanup Of Unmarked Streams

jeremyk-91:
**Goals (and why)**:
- Although we have not identified a specific bug in the cleaning of unmarked streams (introduced in #4201), it has correlated with internal issues. In the interest of risk mitigation, we should disable this.
- We have a hypothesis for what might be going wrong, and have written some code to evaluate that as we go.

- Only clean streams that weren't stored, as per legacy behaviour before #4201 
- Implement both the method of cleanup used in large internal product and in #4201 but purely for diagnostic purposes. Add logging comparing their results.

- Existing tests should suffice, I added an ignore on the test that actually checks this behaviour

- The feature being effectively reverted is still needed for internal shopping product, so it needs to come back in some form; though shopping product is able to mitigate it safely.

**Where should we start reviewing?**: StreamStoreRenderer

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
@@ -651,6 +651,8 @@ public class StreamStoreRenderer {
                 line();
                 line("public class ", MetadataCleanupTask, " implements OnCleanupTask {"); {
                     line();
+                    line("private static final Logger log = LoggerFactory.getLogger(", MetadataCleanupTask, ".class);");
+                    line();
                     line("private final ", TableFactory, " tables;");
                     line();
                     line("public ", MetadataCleanupTask, "(Namespace namespace) {"); {
@@ -658,6 +660,8 @@ public class StreamStoreRenderer {
                     } line("}");
                     line();
                     cellsCleanedUp();
+                    line();
+                    streamDiagnostics();
                 } line("}");
             }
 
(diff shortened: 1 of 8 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -695,29 +704,92 @@ private void cellsCleanedUp() {
                         line("rows.add(", StreamMetadataRow, ".BYTES_HYDRATOR.hydrateFromBytes(cell.getRowName()));");
                     } line("}");
                     line(StreamIndexTable, " indexTable = tables.get", StreamIndexTable, "(t);");
-                    line("Set<", StreamIndexRow, "> indexRows = rows.stream()");
+                    line("executeUnreferencedStreamDiagnostics(indexTable, rows);");
+                    line("Map<", StreamMetadataRow, ", StreamMetadata> currentMetadata = metaTable.getMetadatas(rows);");
+                    line("Set<", StreamId, "> toDelete = Sets.newHashSet();");
+                    line("for (Map.Entry<", StreamMetadataRow, ", StreamMetadata> e : currentMetadata.entrySet()) {"); {
+                        line("if (e.getValue().getStatus() != Status.STORED) {"); {
```
gsheasby: It looks like this would never delete unreferenced-but-stored streams. I don't think we could use this in the internal shopping product without the feature flag we discussed.

felixdesouza:
I'm a bit confused, is there a revert of the original thing? Being a tiny bit pedantic here, but it would be clearer to see the revert commit and then adding the diagnostics separately if possible.

gmaretic:
It makes sense, looking at the generated code. Would be cleaner to do what @felixdesouza is saying though, as this is quite error-prone

jeremyk-91:
I will reopen this.

PR #4433, opened 2019-11-21
[LW] Lock Event Log Impl

gmaretic:
This is still going to be cleaned up, but it's the general idea after we discounted all existing solutions that we could think of


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.OptionalLong;
+
+import com.google.common.math.LongMath;
+import com.google.common.primitives.Ints;
+import com.palantir.lock.watch.LockWatchEvent;
+
+public class ArrayLockEventSlidingWindow {
```
jeremyk-91: The concurrency model of this class needs a write-up / comments.

It looks like `add` performs a volatile write to `nextSequence` while `getFromVersion` performs a volatile read. Then, we do check anything we return. So _I think_ this is fine, but let's talk through.

jeremyk-91: Specifically, the cases I was concerned about involve liveness (e.g. if nextSequence was non-volatile it is possible to get into a state where the sliding window rejects all queries).

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -0,0 +1,91 @@
...
+        LockWatchEvent event = eventBuilder.build(nextSequence);
+        buffer[LongMath.mod(nextSequence, maxSize)] = event;
+        nextSequence++;
+    }
+
+    /**
+     * Returns a list of all events that occurred immediately after the requested version up to the most recent version,
+     * ordered by consecutively increasing sequence numbers. If the list cannot be created, either a priori or because
+     * new events are added to the window during execution of this method causing eviction an event before it was read,
+     * the method will return a singleton list containing only the most recent event, if it exists.
```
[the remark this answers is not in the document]
gmaretic: I have forgotten to update the java doc after changing the behaviour 😨

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -0,0 +1,91 @@
...
+
+        if (versionInTheFuture(version, lastWrittenSequence) || versionTooOld(version, lastWrittenSequence)) {
+            return Optional.empty();
+        }
+
+        int startIndex = LongMath.mod(version + 1, maxSize);
+        int windowSize = Ints.saturatedCast(lastWrittenSequence - version);
+        List<LockWatchEvent> events = new ArrayList<>(windowSize);
+
+        for (int i = startIndex; events.size() < windowSize; i = (i + 1) % maxSize) {
```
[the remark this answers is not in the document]
gmaretic: avoided using mod every time

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
gmaretic: I've added some documentation about it. PLMK if it can be improved!

jeremyk-91:
Thanks! 🎉

PR #4436, opened 2019-11-21
PTExecutors does not propagate state to recurring tasks

carterkozak:
Previously both tracing and ExecutorInheritableThreadLocal state
were retained by recurring tasks, which lead to background activity
referencing state from when a task is first scheduled. There are
several places where we lazily start scheduled background tasks,
this inadvertently attributes background work to the first user
to exercise the codepath.

### DIFF
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/InternalForwardingScheduledExecutorService.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/InternalForwardingScheduledExecutorService.java
@@ -1,0 +1,93 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


commons-executors/src/main/java/com/palantir/common/concurrent/InternalForwardingScheduledExecutorService.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Similar to {@link AbstractForwardingScheduledExecutorService} except this implementation allows
+ * implementors to wrap one-shot and recurring tasks differently. Consider tracing, for example,
+ * where we want to propagate trace-id to executors, we generally do not want recurring tasks
+ * to retain information from the originating thread.
+ *
+ * Intentionally package private, not meant for consumption outside of PTExecutors.
+ */
+abstract class InternalForwardingScheduledExecutorService extends AbstractExecutorService
```
carterkozak: Added a new (package-private) class to avoid potentially breaking existing implementations of the AbstractForwardingScheduledExecutorService

PR #4437, opened 2019-11-21
[PDS-10{3841,4895}] Part 3: Framework For Runtime-Configurable Cleanup Tasks

jeremyk-91:
**Goals (and why)**:
- Provide a mechanism for cleanup tasks to admit runtime configuration.
- This is required as part of internal shopping product's use case for stream stores.
- Outline precisely what configuration would be required.

- Pass config into executions of a cleanup task.

- Added a test for how the config should look like.

- The abstraction is kind of poor. Not entirely keen on `CleanupFollowerConfig` living in `atlasdb-api`, but unfortunately it's needed as it may be passed to the cleanup tasks.

**Where should we start reviewing?**: `CleanupFollowerConfig`

**Priority (whenever / two weeks / yesterday)**: this week would be great, if not early next week. **While this is outstanding, shopping product cannot take AtlasDB upgrades.**


jeremyk-91:
Closing following discussion on PDS-103841. We can reopen if needed.

PR #4438, opened 2019-11-21
[PDS-10{3841,4895}] Part 4: Delete Unreferenced Streams, But Only If Authorized

jeremyk-91:
**Goals (and why)**:
- Allow shopping product to continue its workflows at large deployments where it needs the old cleanup functionality (and it has suitable safety mechanisms)
- While still generally ensuring that products which can't take such hits are protected.

- Wire the configuration option through.

[] TODO jkong add tests
The 'no' version is tested by the `TargetedSweepEteTest`. The `yes` version is untested, @jeremyk-91 to add this. Nonetheless most of the PR should be reviewable in this state.

- Nothing in particular actually. Are we VERY sure this is safe, even assuming that the stream IDs we get back from findUnreferencedStreams may be wrong?

**Where should we start reviewing?**: `StreamStoreRenderer`

**Priority (whenever / two weeks / yesterday)**: this week would be nice, if not early next - this blocks AtlasDB upgrades on shopping product while outstanding


jeremyk-91:
Closing following discussion on PDS-103841. We can reopen if needed.

PR #4439, opened 2019-11-22
Number of hosts in C* config fix.

OStevan:
**Goals (and why)**:
Internal deployments have different requirements for number of host.

- renamed the method to reflect the semantic content
- for CQL hosts used a visitor where needed

- `CassandraServersConfigs`

ASAP


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/DefaultCassandraAsyncKeyValueServiceFactory.java
```diff
@@ -52,9 +55,21 @@ public AsyncKeyValueService constructAsyncKeyValueService(
                 config,
                 initializeAsync);
 
+        int numberOfHosts = config.servers().accept(new Visitor<Integer>() {
```
jkozlowski: Meh

PR #4441, opened 2019-11-23
reduce sleep between scrub runs

clockfort:
This will cause extra work to be performed in some cases, especially on initial startup.
This change is intended to fix issues where the scrub queue can easily become permanently behind.

This currently happens on our dev stack for a service that rolls every time a commit is made to a part of the codebase (that happens to need scrub for it to function, at least currently)

It is anticipated that [future deployments that are in the cloud and have cloud-y constant upgrades / upgrade management ] instead of the current user of this product [fairly restrictive on-prem sites that can only occasionally take upgrades] also might run into the same problem as the dev stack, where constant rolling of services can be at odds with this setting.


clockfort:
Picture is worth 1000 words (this is from the new metrics)
image

jeremyk-91:
Apologies for slow iteration, I'm on this now

PR #4446, opened 2019-11-27
Tracing for futures processing.

OStevan:
**Goals (and why)**:
Better tracing for query result processing.

- wraps `Callable` with tracing information

- `AtlasFutures`

today


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
```diff
@@ -92,4 +96,18 @@ public void close() {
...
+    /**
+     * Constructs a {@link Callable} which will be traced such that the trace id of the current trace is maintained when
+     * the wrapped {@link Callable} is passed to an {@link java.util.concurrent.Executor}.
+     *
+     * @param callable to be traced
+     * @param operation name of the passed callable
+     * @param <R> result type of he passed {@link Callable}
+     * @return traced {@link Callable}
+     */
+    public static <R> Callable<R> tracedCallable(Callable<R> callable, String operation) {
```
felixdesouza: Does this need to be public? Unclear where else it's used. On a separate note, will you not need to call it at points in three atlas stack where you've put things on executors as opposed to just the internal ski resort product?

jkozlowski: Yeah, should be private

jkozlowski: And also, just Tracers.wrap

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/CqlClientImpl.java
```diff
@@ -139,23 +141,26 @@ public void close() {
             Executor executor,
             RowStreamAccumulator<V> rowStreamAccumulator) {
         return Futures.transformAsync(
-                session.executeAsync(executableStatement),
+                AtlasFutures.tracingFuture(() -> session.executeAsync(executableStatement), "CqlClientImpl: execute"),
```
jkozlowski: Btw, this is why I suggested you just create a specialized Executor, so you don't need to do this manually everywhere.

jkozlowski: So let's switch to that

jkozlowski: Canonically, we want something like:

```
DeferredTracer deferredTracer = new DeferredTracer();
        Executor executor = command -> executorService.execute(() -> {
            deferredTracer.withTrace(() -> { command.run(); return null;});
        });
        return Futures.whenAllSucceed(inputToListenableFutureMap.values())
                .call(() -> KeyedStream.stream(inputToListenableFutureMap)
                                .map(AtlasFutures::getDone)
                                .filter(Optional::isPresent)
                                .map(Optional::get)
                                .collectToMap(),
                        executor);
```

OStevan: Refactored to this structure.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
OStevan: Removed.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
```diff
@@ -92,4 +96,12 @@ public void close() {
             throw Throwables.rewrapAndThrowUncheckedException(e);
         }
     }
+
+    public static Executor traceRestoringExecutor(Executor executorService, String operation) {
```
jkozlowski: executorService -> executor.

jkozlowski: private

jkozlowski: Maybe currentTraceExecutor? or a comment explaining how the heck this works, because this is super subtle.

versions.props
```diff
@@ -29,7 +29,7 @@ com.palantir.docker.proxy:docker-proxy-rule = 0.8.1
 com.palantir.remoting2:* = 2.6.3
 com.palantir.safe-logging:* = 1.11.0
 com.palantir.tokens:auth-tokens = 3.6.1
-com.palantir.tracing:* = 3.3.0
+com.palantir.tracing:* = 4.0.0
```
jkozlowski: Do you still need the bump?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
OStevan: It is also used in the CqlClientImpl, unless we want to remove that.

versions.props
OStevan: Not really, will bump in separate PR.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
OStevan: Renamed.

jkozlowski: Discussed offline: we don't care about it so much in CqlClientImpl, since that stuff is happening directly on the netty threads, it's unlikely we'll do any logging there.

PR #4448, opened 2019-11-28
[Timelock Partitioning] Part 44b: Ensure same instance of `LeaderElectionService` is reused

felixdesouza:
**Goals (and why)**:
As part of #4340, I switched over the wiring logic for timelock and how it does paxos in preparation for "leader for each namespace" timelock.

However, a new instance of a `LeaderElectionService` was being made for each client, despite the cluster only having *one* leader. As such, in this mode, we'd have `n` `LeaderElectionService`, and each would have a corresponding `AwaitingLeadershipProxy`. We're not running with any batching, so it's disastrous.

Similar to `LocalPaxosComponents`, we want to have a class that we store instances of a `LeaderElectionService` keyed on its `Client`.

Local testing to ensure we haven't blown up number of threads. 

Anywhere.

ASAP, unblocks atlasdb upgrades on internal timelock repo.

### DIFF
--- a/changelog/@unreleased/pr-4448.v2.yml
+++ b/changelog/@unreleased/pr-4448.v2.yml
@@ -1,7 +1,7 @@
 type: fix
 fix:
   description: 'Ensure we use the same instance of `LeaderElectionService` when in
-    `leader-for-all clients` mode, this brings request rates back to the same rate
+    `leader-for-all-clients` mode, this brings request rates back to the same rate
     before PR 4340 merged. This addresses internal issue PDS-105806.'
   links:
   - https://github.com/palantir/atlasdb/pull/4448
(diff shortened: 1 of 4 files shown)


changelog/@unreleased/pr-4448.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: fix
+fix:
+  description: 'Ensure we use the same instance of `LeaderElectionService` when in
+    `leader-for-all clients` mode, this brings request rates back to the same rate
```
jeremyk-91: Is this missing a hyphen?

jkozlowski:
Was this just a perf hit or did each client get a different paxos instance (and hence was a correctness problem?

felixdesouza:
Not a correctness issue. Just a perf hit.

Issue #4449, opened 2019-11-28
Immutable timestamp lock isn't always cleaned up by LegacyTimelockService


GrahamDennis:
`LegacyTimelockService.startIdentifiedAtlasDbTransaction()` can take an immutable timestamp lock and not free it if it is interrupted during the call to `getFreshTimestamp()`.

The startIdentifiedAtlasDbTransaction method looks like [this](https://github.com/palantir/atlasdb/blob/0.174.4/lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java#L105):

```
    @Override
    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
        return StartIdentifiedAtlasDbTransactionResponse.of(
                lockImmutableTimestamp(),
                TimestampAndPartition.of(getFreshTimestamp(), 0));
    }
```

Looking at the decompiled code, `lockImmutableTimestamp()` is called before `getFreshTimestamp()`.  If `getFreshTimestamp()` throws an exception because the current thread is interrupted, then the immutable timestamp lock taken by `lockImmutableTimestamp()` is not released.

Here's a stacktrace where this happened:

```
java.lang.RuntimeException: java.lang.InterruptedException
	at com.palantir.timestamp.RequestBatchingTimestampService.getFreshTimestamps(RequestBatchingTimestampService.java:77)
	at com.palantir.timestamp.RequestBatchingTimestampService.getFreshTimestamp(RequestBatchingTimestampService.java:65)
	at com.palantir.tritium.proxy.InstrumentedTimestampService_CloseableTimestampService$4.getFreshTimestamp(Unknown Source)
	at com.palantir.lock.impl.LegacyTimelockService.getFreshTimestamp(LegacyTimelockService.java:69)
	at com.palantir.lock.impl.LegacyTimelockService.startIdentifiedAtlasDbTransaction(LegacyTimelockService.java:108)
	at com.palantir.lock.v2.TimelockService.startTransactionWithWatches(TimelockService.java:91)
	at com.palantir.atlasdb.transaction.impl.SnapshotTransactionManager.setupRunTaskWithConditionThrowOnConflict(SnapshotTransactionManager.java:158)
	at com.palantir.atlasdb.transaction.impl.SerializableTransactionManager.setupRunTaskWithConditionThrowOnConflict(SerializableTransactionManager.java:57)
[...]

PR #4452, opened 2019-12-04
Fix immutable timestamps from not being cleaned up

GrahamDennis:
**Goals (and why)**:
If LegacyTimestampService.startIdentifiedAtlasDbTransaction was interrupted
during the call to getFreshTimestamp, then the immutable timestamp lock
would not be cleaned up.

Decompiling the code indicates that `lockImmutableTimestamp()` is called before `getFreshTimestamp()`.  It's unclear to me whether we need to call `getFreshTimestamp()` because `lockImmutableTimestamp()` itself just got a new fresh timestamp.

Fixes https://github.com/palantir/atlasdb/issues/4449

* added try/catch around the call to `lockImmutableTimestamp()`.

* No manual testing has occurred.

* is it necessary to actually call `getFreshTimestamp()` or can the timestamp of the immutable timestamp lock be re-used?

This is causing lost immutable timestamp locks, however this doesn't appear to be causing any user-visible problems currently.

### DIFF
--- a/lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
+++ b/lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
@@ -119,7 +119,7 @@ public class LegacyTimelockService implements TimelockService {
             } catch (Throwable unlockThrowable) {
                 throwable.addSuppressed(unlockThrowable);
             }
-            throw Throwables.rewrapAndThrowUncheckedException(throwable);
+            throw throwable;
         }
     }
 
--- a/lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
+++ b/lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
@@ -96,7 +96,11 @@ public class LegacyTimelockService implements TimelockService {
                     LockTokenConverter.toTokenV2(lock));
         } catch (Throwable e) {
             if (lock != null) {
-                lockService.unlock(lock);
+                try {
+                    lockService.unlock(lock);
+                } catch (Throwable unlockThrowable) {
+                    e.addSuppressed(unlockThrowable);
+                }
             }
             throw Throwables.rewrapAndThrowUncheckedException(e);
         }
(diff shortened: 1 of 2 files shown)


lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -103,9 +104,15 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
...
     public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        return StartIdentifiedAtlasDbTransactionResponse.of(
-                lockImmutableTimestamp(),
-                TimestampAndPartition.of(getFreshTimestamp(), 0));
+        LockImmutableTimestampResponse immutableTimestamp = lockImmutableTimestamp();
+        try {
+            return StartIdentifiedAtlasDbTransactionResponse.of(
+                    immutableTimestamp,
+                    TimestampAndPartition.of(getFreshTimestamp(), 0));
+        } catch (Throwable throwable) {
```
carterkozak:
```suggestion
        } catch (RuntimeException | Error e) {
```

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -103,9 +104,15 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
...
+            unlock(ImmutableSet.of(immutableTimestamp.getLock()));
```
carterkozak: Unlock should probably be wrapped in a try/catch, on `unlock` failure, the exception should be added as a suppressed exception to the first exception above, otherwise we lose the root cause :-)

GrahamDennis: done

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
GrahamDennis: done

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -95,17 +96,31 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
...
+            return StartIdentifiedAtlasDbTransactionResponse.of(
+                    immutableTimestamp,
+                    TimestampAndPartition.of(getFreshTimestamp(), 0));
+        } catch (RuntimeException | Error throwable) {
+            try {
+                unlock(ImmutableSet.of(immutableTimestamp.getLock()));
+            } catch (Throwable unlockThrowable) {
+                throwable.addSuppressed(unlockThrowable);
+            }
+            throw Throwables.rewrapAndThrowUncheckedException(throwable);
```
carterkozak: nit: `throwable` is unchecked, so we can throw it without a rewrap

carterkozak:
This looks reasonable to me, but I don't have approval on this repository.

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
GrahamDennis: done

GrahamDennis:
Can you review please @jeremyk-91?

jeremyk-91:
Reviewing today. Apologies, I was out on Tuesday and most of Wednesday

GrahamDennis:
ping @jeremyk-91!

GrahamDennis:
This was fixed by https://github.com/palantir/atlasdb/pull/4484

PR #4454, opened 2019-12-04
Refactor HostLocation discovery

OStevan:
**Goals (and why)**:
Make the host location discovery easier to propagate and split the logic.

- `HostLocationSupplier` is an interface
- `CombiningHostLocationSupplier` deals with location override
- `Ec2AwareHostLocationSupplier` does the dc aware logic

- tests are mostly the same but are split based on the tested class

- standard coding style

- `HostLocationSupplier`

- tomorrow

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
@@ -17,14 +17,25 @@
 package com.palantir.atlasdb.keyvalue.cassandra.pool;
 
 import java.util.Optional;
-import java.util.function.Supplier;
 
-public interface HostLocationSupplier extends Supplier<Optional<HostLocation>> {
+final class CombiningHostLocationSupplier implements HostLocationSupplier {
+    private final HostLocationSupplier locationSupplier;
+    private final HostLocationSupplier overrideLocation;
 
-    static HostLocationSupplier createForTests() {
-        return Optional::empty;
+    public CombiningHostLocationSupplier(
+            HostLocationSupplier overrideLocationSupplier,
+            HostLocationSupplier locationSupplier) {
+        this.overrideLocation = overrideLocationSupplier;
+        this.locationSupplier = locationSupplier;
     }
 
     @Override
-    Optional<HostLocation> get();
+    public Optional<HostLocation> get() {
+        Optional<HostLocation> location = overrideLocation.get();
+        if (location.isPresent()) {
+            return location;
+        }
+
+        return locationSupplier.get();
+    }
 }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
@@ -19,19 +19,19 @@ package com.palantir.atlasdb.keyvalue.cassandra.pool;
 import java.util.Optional;
 
 final class CombiningHostLocationSupplier implements HostLocationSupplier {
-    private final HostLocationSupplier overrideLocation;
+    private final HostLocationSupplier overrideLocationSupplier;
     private final HostLocationSupplier locationSupplier;
 
     CombiningHostLocationSupplier(
             HostLocationSupplier overrideLocationSupplier,
             HostLocationSupplier locationSupplier) {
-        this.overrideLocation = overrideLocationSupplier;
+        this.overrideLocationSupplier = overrideLocationSupplier;
         this.locationSupplier = locationSupplier;
     }
 
     @Override
     public Optional<HostLocation> get() {
-        Optional<HostLocation> location = overrideLocation.get();
+        Optional<HostLocation> location = overrideLocationSupplier.get();
         if (location.isPresent()) {
             return location;
         }
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
@@ -19,10 +19,10 @@ package com.palantir.atlasdb.keyvalue.cassandra.pool;
 import java.util.Optional;
 
 final class CombiningHostLocationSupplier implements HostLocationSupplier {
-    private final HostLocationSupplier locationSupplier;
     private final HostLocationSupplier overrideLocation;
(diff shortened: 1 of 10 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public Optional<HostLocation> get() {
+        Optional<HostLocation> location = overrideLocation.get();
+        if (location.isPresent()) {
+            return location;
+        }
+
+        return locationSupplier.get();
```
jeremyk-91: non-actionable: ... and this is where I complain about not having Java 9 `Optional#or` which does exactly what you need

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import java.util.Optional;
+
+final class CombiningHostLocationSupplier implements HostLocationSupplier {
+    private final HostLocationSupplier overrideLocation;
+    private final HostLocationSupplier locationSupplier;
```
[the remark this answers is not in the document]
OStevan: Standardized.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningHostLocationSupplier.java
OStevan: 😢

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CombiningAwareHostLocationSupplierTests.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.palantir.atlasdb.keyvalue.cassandra.pool;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Optional;
+
+import org.junit.Test;
+
+public class CombiningAwareHostLocationSupplierTests {
```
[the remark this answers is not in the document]
OStevan: Renamed.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2AwareHostLocationSupplier.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private final Supplier<String> snitchSupplier;
+    private final Supplier<HostLocation> ec2Supplier;
+
+    public static HostLocationSupplier createMemoized(Supplier<String> snitchSupplier) {
+        HostLocationSupplier delegate = new Ec2AwareHostLocationSupplier(
+                snitchSupplier,
+                new Ec2HostLocationSupplier());
+
+        return new HostLocationSupplier() {
```
felixdesouza: this is new, why do we want this?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
```diff
@@ -19,53 +19,12 @@
...
-            Optional<HostLocation> overrideLocation) {
-        this.snitchSupplier = Suppliers.memoize(snitchSupplier::get);
-        this.ec2Supplier = Suppliers.memoize(ec2Supplier::get);
-        this.overrideLocation = overrideLocation;
-    }
-
-    public HostLocationSupplier(Supplier<String> snitchSupplier,
-            Optional<HostLocation> overrideLocation) {
-        this(snitchSupplier, new Ec2HostLocationSupplier(), overrideLocation);
+    static HostLocationSupplier createForTests() {
```
felixdesouza: just inline this :'(

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2AwareHostLocationSupplier.java
OStevan: I wanted to have a static method will create a `HostLocationSupplier` which will have the same underlying sematics as the `Ec2AwareHostLocationSupplier` but memoized. And then this static method looked like a nice way to do it.

felixdesouza:
Sorry, should have been clearer, why do we need to memoize all of a sudden? We weren’t before right?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplier.java
OStevan: Inlined.

OStevan:
So what I wanted to do with `createMemoized` was to memoize the whole `get` of that class. The original solution was to memoize the building blocks and then by doing that remove the network overhead of contacting the endpoints. While still constantly logging messages every time even though the result is the same, this way we log once what happened and we are done. 

What might be confusing is that I still memoize the building blocks which can be removed.

felixdesouza:
Have a separate thing i.e. `MemoizingHostLocationSupplier` - or whatever naming scheme you've got - since it's not tied to that instance and the anonymous class is strange. Get rid of the building blocks memoization too. 

As an extra, when you remove the memoization of the building blocks, I'd double check whether these classes are used in the internal ski resort product and make sure that we're memoizing using the new stuff.

jkozlowski:
@OStevan do you have a full implementation of the location aware stuff for the CQL part? Cause I feel like we might be doing refactorings here before we understand how this is actually going to work.

e.g. how is the CqlClient (which is where you need it) pull out the supplier of the location, since that's tied to stuff that's instantiated in CassandraKeyValueServiceImpl?

OStevan:
@jkozlowski this is the wip [branch](https://github.com/palantir/atlasdb/tree/sognjanovic/refactor) which I used to explore plumbing through and whats what.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/Ec2AwareHostLocationSupplier.java
OStevan: Resolved.

felixdesouza:
closing this out, can reopen when ready to pick up cql work again/whatever made this PR necessary

PR #4457, opened 2019-12-05
[Timelock Partitioning] Part 44d: Revert partitioning refactors

felixdesouza:
Loss of metrics for the moment, don't have time to investigate and don't want to block other work. Reverting for now.


jkozlowski:
writeChangelog is still failing, do we need the changelog or something?

felixdesouza:
there's something weird going on where @gmaretic is editing the changelog comment and unticking things, but it's unclear to both of us how or why that's happening

jkozlowski:
Ok looks like it worked now

PR #4460, opened 2019-12-06
[LW] Part 5: Client Lock Watch Event Tracker

gmaretic:
**Goals (and why)**:
When starting a transaction, we have to request a log of lock watch events since the "last known version". The tracker supplies and updates this last known version of lock watch state.

The last known version should be as fresh as possible, to minimise the amount of events timelock has to send back, and to avoid the server-side log window to progress too far.
Only one update runs at a time, the new state is computed, and is then atomically set. Until that moment, all requests to get the current state will just get the slightly older version instead of waiting on the update.
 
Tests that cover all the basics. Feel free to suggest more elaborate tests if you think they are needed.

Is the code clear? Do we need more detailed docs or more extensive tests?

`VersionedLockWatchStateImpl` and tests.

Early next week?


lock-api/src/main/java/com/palantir/lock/watch/LockWatchStateUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+    public Void visit(LockWatchOpenLocksEvent openLocksEvent) {
+        for (LockDescriptor descriptor : openLocksEvent.lockDescriptors()) {
+            lockWatchState.put(descriptor, LockWatchInfo.of(LockWatchInfo.State.LOCKED, openLocksEvent.sequence()));
+        }
+        unmatchedOpenLocksEvents.add(openLocksEvent.lockWatchId());
+        return null;
+    }
+
+    @Override
+    public Void visit(LockWatchCreatedEvent lockWatchCreatedEvent) {
```
jeremyk-91: Need to check that the obvious race condition works:
- there is an open-locks event, associated with watch ID `Q`
- then there is a lock locked that is related to `Q`
- then we get the created event

I guess we still have it in the Info, even if we aren't sure how it ended up there at that point, so that looks fine. Would like to sanity check this thing next time we sync up though.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchEventTrackerImplTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Test
+    public void lockWatchCreatedWithoutOpenLockGivesNoInformationOnLocks() {
+        UUID leaderId = UUID.randomUUID();
+        LockWatchStateUpdate update = LockWatchStateUpdate.update(leaderId, 0, ImmutableList.of(
+                LockWatchCreatedEvent.builder(ROW_PREFIX_REQ, UUID.randomUUID()).build(0)));
+
+        tracker.updateState(update);
+        VersionedLockWatchState state = tracker.currentState();
+        assertThat(state.version()).isEqualTo(OptionalLong.of(0));
```
jeremyk-91:
```suggestion
        assertThat(state.version()).hasValue(0);
```

And so on, throughout the test class.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchEventTrackerImplTest.java
```diff
@@ -0,0 +1,203 @@
...
+        assertThat(state.leaderId()).isEqualTo(leaderId);
+        assertThat(state.lockWatchState(DESCRIPTOR)).isEqualTo(LockWatchInfo.NOT_WATCHED);
```
jeremyk-91: Preferably not in this PR as it is already huge, but one of those custom AssertJ assertions would make the tests smoother to read.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchEventTrackerImplTest.java
```diff
@@ -0,0 +1,203 @@
...
+        assertThat(state.lockWatchState(DESCRIPTOR)).isEqualTo(LockWatchInfo.of(LOCKED, OptionalLong.of(0)));
+        assertThat(state.lockWatchState(OTHER_DESCRIPTOR)).isEqualTo(LockWatchInfo.of(UNLOCKED, OptionalLong.empty()));
+    }
+
+    @Test
+    public void lockCanBeUpdated() {
+        UUID leaderId = UUID.randomUUID();
+        LockWatchStateUpdate update = LockWatchStateUpdate.update(leaderId, 1, ImmutableList.of(
+                        LockEvent.builder(ImmutableSet.of(DESCRIPTOR)).build(0),
+                        UnlockEvent.builder(ImmutableSet.of(DESCRIPTOR)).build(1)));
```
jeremyk-91: The assumption here, I guess, is that these events are already filtered to be relevant to things that are being watched - here this would mean that the registration of some watch covering `DESCRIPTOR` has already taken place in the past.

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        long currentTrackedVersion = lastKnownVersion.getAsLong();
+        if (updateNotPossible(update, currentTrackedVersion)) {
+            return currentState();
+        }
+
+        runUpdate(false, update);
+        return currentState();
+    }
```
jeremyk-91: Some comments/explanation probably needed here, I'm having trouble understanding the difference between `resetAllState` vs `runUpdate(true, x)` without staring at what the implementation actually does.

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * Multiple threads may simultaneously try to update the tracker, potentially with overlapping updates. At the same
+ * time, it is allowed to return a slightly stale version on calls to currentState, so we do not want to block on
+ * updates. This is implemented as follows.
+ *
+ * 1. Updates are synchronized, and the values of watches, singleLocks, openLocksEvents, lastKnownVersion, and leaderId
+ *    are only changed within the synchronized block. This ensures that two threads are not going to conflict during an
+ *    update, or do unnecessary work.
+ * 2. Furthermore, changes to the values of watches, singleLocks, openLocksEvents, lastKnownVersion are also guarded by
+ *    a write lock. This allows calls to currentState() to return a consistent version of lock watch state even if an
```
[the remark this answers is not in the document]
gmaretic: It's a bit more complex than that, since within the synchronized block we do not have to have any sort of lock to do reads; but that is the essence.

lock-api/src/main/java/com/palantir/lock/watch/LockWatchStateUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+        for (LockDescriptor descriptor : lockEvent.lockDescriptors()) {
+            lockWatchState.put(descriptor, LockWatchInfo.of(LockWatchInfo.State.LOCKED, lockEvent.sequence()));
+        }
+        return null;
+    }
+
+    @Override
+    public Void visit(UnlockEvent unlockEvent) {
+        for (LockDescriptor descriptor : unlockEvent.lockDescriptors()) {
+            OptionalLong lastLocked = lockWatchState.getOrDefault(descriptor, LockWatchInfo.NOT_WATCHED).lastLocked();
```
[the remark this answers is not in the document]
gmaretic: Hopefully more clear now

lock-api/src/main/java/com/palantir/lock/watch/LockWatchStateUpdater.java
gmaretic: Yes, that's the intended behaviour -- having open locks means we didn't miss any currently held locks, then we register the lock watch created event, but maybe some of the locks have been taken in the meantime -- they will, however, be logged, and we will correctly identify that they are in fact locked once the lock watch created event is processed.

The purpose of the LWCE is to let us know that everything in the lock watch request that is not explicitly locked is in fact not locked. Added a test for the case you mention.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchEventTrackerImplTest.java
gmaretic: I knew there must be a better way :(

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
gmaretic: Tried to make it more understandable, also changed the case for no known last state, as we can use the update then

lock-api/src/main/java/com/palantir/lock/watch/VersionedLockWatchStateImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public UUID leaderId() {
+        return leaderId;
+    }
+
+    @Override
+    public LockWatchInfo lockWatchState(LockDescriptor lockDescriptor) {
+        if (locks.containsKey(lockDescriptor)) {
+            return locks.get(lockDescriptor);
```
jkozlowski: Should be done in one lookup

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Override
+    public void setLockWatchStateForStartTimestamp(long startTimestamp, VersionedLockWatchState lockWatchState) {
+        startTsToLockWatchState.put(startTimestamp, lockWatchState);
+    }
+
+    @Override
+    public VersionedLockWatchState getLockWatchStateForStartTimestamp(long startTimestamp) {
+        return startTsToLockWatchState.get(startTimestamp);
```
jkozlowski: This can return a null right?

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Override
+    public VersionedLockWatchState removeLockWatchStateForStartTimestamp(long startTimestamp) {
+        return startTsToLockWatchState.remove(startTimestamp);
```
jkozlowski: How is this going to be used?

gmaretic: You could, but then you force your transactions to abort. Like this you keep it in memory until you get to commit time, when it gets removed

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
gmaretic: If you use it wrong, it can, if you use it correctly, no

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
jkozlowski: Oh wait, I just noticed that this map is for startTimestamp -> state

gmaretic:
Replaced by #4504

PR #4461, opened 2019-12-06
Remove ThrowingCqlClientImpl

OStevan:
**Goals (and why)**:
Make it easier for users of library to deal with wrong configs in a graceful manner.
- removed `ThrowingCqlClientImpl`
- Optional.empty instead of `ThrowingCqlClientImpl`

today


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
```diff
@@ -62,22 +63,22 @@ public DefaultCqlClientFactory() {
...
+            public Optional<CqlClient> visit(CassandraServersConfigs.DefaultConfig defaultConfig) {
+                return Optional.empty();
             }
 
             @Override
-            public CqlClient visit(CassandraServersConfigs.CqlCapableConfig cqlCapableConfig) {
+            public Optional<CqlClient> visit(CassandraServersConfigs.CqlCapableConfig cqlCapableConfig) {
                 if (!cqlCapableConfig.validateHosts()) {
                     log.warn("Your CQL capable config is wrong, the hosts for CQL and Thrift are not the same, using "
                             + "async API will result in an exception.");
```
jeremyk-91: nit: This log message is no longer accurate after the change (we can now say using the async API may result in undefined behaviour: I know that officially we will use the sync API, but I don't think we necessarily need to guarantee that).

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsAsyncFallbackMechanismsTests.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @ClassRule
+    public static final CassandraResource CASSANDRA_RESOURCE = new CassandraResource();
+
+    private KeyValueService keyValueService;
+    private Proxy socksProxy;
+    private AsyncKeyValueService asyncKeyValueService;
+    private CassandraAsyncKeyValueServiceFactory factory;
+
+    private static ImmutableCassandraKeyValueServiceConfig getConfigWithGcGraceSeconds(
```
jeremyk-91: nit: ok for tests, but please return the abstract type

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
OStevan: Changed.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsAsyncFallbackMechanismsTests.java
```diff
@@ -0,0 +1,120 @@
...
+        when(factory.constructAsyncKeyValueService(any(), any(), eq(AtlasDbConstants.DEFAULT_INITIALIZE_ASYNC)))
+                .thenReturn(Optional.empty());
+
+        CassandraKeyValueServiceConfig config = ImmutableCassandraKeyValueServiceConfig.builder()
+                .from(CASSANDRA_RESOURCE.getConfig())
+                .asyncKeyValueServiceFactory(factory)
+                .build();
+
+        keyValueService = CassandraKeyValueServiceImpl.createForTesting(config);
+        keyValueService.createTable(TEST_TABLE, AtlasDbConstants.GENERIC_TABLE_METADATA);
```
jeremyk-91: non-actionable: This seems a bit strange here, though I gather that you need it for the test to pass since it _actually_ hits the sync endpoint.

OStevan: It didn't make sense to throw so I did it like this.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKvsAsyncFallbackMechanismsTests.java
OStevan: This happens when you copy code 😄

Issue #4411, opened 2019-11-15
Force a decision on whether to delegate implemented methods with AutoDelegate


gmaretic:
We've burned ourselves for the n-th time with delegating a default method. We should fail on generating an AutoDelegate implementation if there are any default methods that have not been annotated with either `@DoNotDelegate` or `@DoDelegate`

PR #4462, opened 2019-12-06
Explicit do delegate annotation

sudiksha27:
[Reference Issue](https://github.com/palantir/atlasdb/issues/4411)

**Goals**:
- Fail on generating an AutoDelegate implementation if there are any default methods that have not been annotated with either @DoNotDelegate or @DoDelegate
- Do not autoDelegate CassandraKeyValueServiceConfig

- Validation check is added (validateMethodsToBeAutoDelegated) in the AutoDelegateProcessor.
- It enforces decision on whether a default method should be autoDelegated or not.
- In case where a default method is not explicitly annotated, autoDelegation for the interface fails.

Work in progress.

AutoDelegateProcessor

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
@@ -15,14 +15,19 @@
  */
 package com.palantir.atlasdb.cassandra;
 
+import java.net.InetSocketAddress;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.google.common.base.MoreObjects;
+import com.palantir.atlasdb.keyvalue.cassandra.async.CassandraAsyncKeyValueServiceFactory;
+import com.palantir.atlasdb.keyvalue.cassandra.pool.HostLocation;
 import com.palantir.atlasdb.spi.KeyValueServiceRuntimeConfig;
+import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
 
-public class CassandraReloadableKvsConfig implements AutoDelegate_CassandraKeyValueServiceConfig {
+public class CassandraReloadableKvsConfig implements CassandraKeyValueServiceConfig {
     private final CassandraKeyValueServiceConfig config;
     private final Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfigSupplier;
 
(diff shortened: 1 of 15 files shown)


atlasdb-processors/src/main/java/com/palantir/processors/AutoDelegateProcessor.java
```diff
@@ -139,6 +141,20 @@ private static TypeElement validateAnnotatedElement(Element annotatedElement) th
         return (TypeElement) annotatedElement;
     }
 
+    private static void validateMethodsToBeAutoDelegated(Element annotatedElement, TypeToExtend typeToExtend)
+            throws ProcessingException {
+        Set<ExecutableElement> methodsToBeDelegated = typeToExtend.getMethods();
+
+        for (ExecutableElement methodElement : methodsToBeDelegated) {
+            if (methodElement.getModifiers().contains(Modifier.DEFAULT)
+                && methodElement.getAnnotation(DoDelegate.class) == null
```
Jolyon-S: I haven't checked, but is there a neater method for doing `methodElement.getAnnotation(DoDelegate.class) == null`, like maybe `checkAnnotationPresent` or something like that?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -62,12 +157,102 @@ public int fetchBatchCount() {
...
+    @Override
+    public boolean usingSsl() {
+        return config.usingSsl();
+    }
+
+    @Override
+    public void check() {
+
+    }
+
```
Jolyon-S: this feels like a lot of copy and pasted methods, but I'm not sure if there is a better way of doing this while avoiding delegates.

OStevan: I don't really think we need this here. The worst thing that can happen is that we call the check as defined in the `CassandraKeyValueServiceConfig` which also in a way makes sense.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -62,12 +157,102 @@ public int fetchBatchCount() {
...
+    @Override
+    public boolean usingSsl() {
+        return config.usingSsl();
+    }
+
+    @Override
+    public void check() {
+
```
jeremyk-91: nit: Might be good to leave a comment explaining why `check()` is empty. _In this case, this is correct._

atlasdb-processors/src/main/java/com/palantir/processors/AutoDelegateProcessor.java
```diff
@@ -139,6 +142,19 @@ private static TypeElement validateAnnotatedElement(Element annotatedElement) th
...
     }
 
+    private static void validateMethodsToBeAutoDelegated(Element annotatedElement, TypeToExtend typeToExtend)
+            throws ProcessingException {
+        Set<ExecutableElement> methodsToBeDelegated = typeToExtend.getMethods();
+
+        for (ExecutableElement methodElement : methodsToBeDelegated) {
+            if (methodElement.getModifiers().contains(Modifier.DEFAULT)
+                    && methodElement.getAnnotation(DoDelegate.class) == null) {
+                throw new ProcessingException(annotatedElement, "Default methods must be annotated with "
```
jeremyk-91: Would be good to explain why we don't need to check for `DoNotDelegate` here.

PR #4466, opened 2019-12-09
Block default method auto delegate

sudiksha27:
[Reference Issue](https://github.com/palantir/atlasdb/issues/4411)

**Goals (and why)**:
- Completely block auto delegating the default methods.
- Remove DoNotDelegate annotation

- Updated the filter (interfaceMethodFilter) to block default methods from being auto delegated
- Safely removed doNotDelegate 

- TypeToExtend.interfaceMethodFilter


felixdesouza:
I'm potentially missing some context but this doesn't block compilation looking at circle.

I believe you need to fail to compile since both options are reasonable i.e. `@DelegateDefault` and `@DoNotDelegateDefault`.

A counter example for actually wanting to delegate default methods, say you had an interface:
```java
@AutoDelegate
interface Bla {
  void doSomething();
  default void doSomethingManyTimes(int n) {
	IntStream.range(n).forEach(Bla::doSomething);
  }
}

class SuperSpecialBlaImpl implements Bla {
  void doSomething() {...}
  void doSomethingManyTimes(int n) {
    specialBatchFunction(n);
  }
}

class HalfImplementedInterface implements AutoDelegate_Bla {
  private final Bla delegate = new SuperSpecialBlaImpl();

  Bla delegate() { return delegate; }
  
  @Override
  void doSomething() { ... }
}
```

if I call `HalfImplementedInterface#doSomethingManyTimes`, I believe if you don't block compilation, it won't delegate, and instead call the default method on the auto delegate interface but not on the implementation and then you lose the nice specialisation!

Where we were bitten previously is when we actually only ever wanted the behaviour on the top level interface, to provide a utility/adaptation method. 

Happy to discuss in person, @gmaretic I believe this is the counter example I was looking for but couldn't remember.

jkozlowski:
@felixdesouza I think this is the exact problem from here: https://github.com/palantir/atlasdb/pull/4463. But it's ever worse, because you don't even need auto delegates to break this. You just need one implementation of Bla that doesn't override the default implementation and everything breaks.

So my proposal was we generally disallow default implementations for anything marked as AutoDelegate (which most places where we do this delegation flow should be doing) and eat the potential pain or having to implement #isInitialized everywhere.

felixdesouza:
I'm not sure I'm following why it happens outside of auto delegate? That's assuming you have a composition based class right?

PR #4469, opened 2019-12-11
fix infinite loop in LocalOrRemoteProxy

tjwilson90:
previously LocalOrRemoteProxy would infinitely retry its call as long as
the delegate was throwing a NotCurrentLeaderException, regardless of
whether the thread was interrupted. Calling a local service behind a
AwaitingLeadershipProxy on an interrupted thread would trigger a
NotCurrentLeaderException without clearing the interrupt, leading to an
infinte loop.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
@@ -24,7 +24,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import com.google.common.reflect.AbstractInvocationHandler;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.palantir.leader.NotCurrentLeaderException;
 
 /**
@@ -63,7 +62,7 @@ final class LocalOrRemoteProxy<T> extends AbstractInvocationHandler {
             } catch (InvocationTargetException e) {
                 Throwable targetException = e.getTargetException();
                 if (targetException instanceof NotCurrentLeaderException) {
-                    Uninterruptibles.sleepUninterruptibly(10, TimeUnit.MILLISECONDS);
+                    Thread.sleep(10);
                     continue;
                 }
                 throw targetException;
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
@@ -24,7 +24,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import com.google.common.reflect.AbstractInvocationHandler;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.palantir.leader.NotCurrentLeaderException;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 
 /**
  * A proxy that chooses between using a local or remote proxy based on a task that may take a while to complete.
@@ -57,12 +59,16 @@ final class LocalOrRemoteProxy<T> extends AbstractInvocationHandler {
     @Override
     protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
         while (true) {
+            if (Thread.currentThread().isInterrupted()) {
+                // do not clear the interrupt flag
+                throw new SafeIllegalStateException("interrupted");
+            }
             try {
                 return method.invoke(delegate(), args);
             } catch (InvocationTargetException e) {
                 Throwable targetException = e.getTargetException();
                 if (targetException instanceof NotCurrentLeaderException) {
-                    Thread.sleep(10);
+                    Uninterruptibles.sleepUninterruptibly(10, TimeUnit.MILLISECONDS);
                     continue;
                 }
                 throw targetException;
--- a/atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
+++ b/atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
@@ -17,6 +17,7 @@ package com.palantir.atlasdb.factory;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.fail;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
@@ -39,12 +40,16 @@ import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
 
 import java.io.IOException;
+import java.lang.reflect.UndeclaredThrowableException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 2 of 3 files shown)


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -464,6 +472,36 @@ public void keyValueServiceMetricsDoNotContainUserAgent() {
                 .anyMatch(metricName -> metricName.contains(USER_AGENT_NAME)), is(false));
     }
 
+    @Test
+    public void interruptingLocalLeaderElectionTerminates() throws IOException {
+        Path tempDir = Files.createTempDirectory("foo");
```
felixdesouza: please use the `temporaryFolder` that's been setup for the test

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -464,6 +472,36 @@ public void keyValueServiceMetricsDoNotContainUserAgent() {
...
+                        .build(),
+                () -> ImmutableLeaderRuntimeConfig.builder().build(),
+                USER_AGENT);
+        LeaderElectionService leader = localPaxosServices.leaderElectionService();
+        LockService lockService = LockServiceImpl.create();
+        LockService leadershipLock = AwaitingLeadershipProxy.newProxyInstance(
+                LockService.class, () -> lockService, leader);
+        LockService localOrRemoteLock = LocalOrRemoteProxy.newProxyInstance(
+                LockService.class, leadershipLock, null, CompletableFuture.completedFuture(true));
+        try {
```
felixdesouza: prefer to use more canonical assertJ e.g.:

```java
assertThatCode(localOrRemoteLock::currentTimeMillis)
  .as("proxy correctly handles interrupts")
  .hasCauseInstanceOf(InterruptedException.class);
```

changelog/@unreleased/pr-4469.v2.yml
```diff
@@ -0,0 +1,13 @@
+type: fix
+fix:
+  description: |-
+    fix infinite loop in LocalOrRemoteProxy
```
felixdesouza:
```suggestion
```

changelog/@unreleased/pr-4469.v2.yml
```diff
@@ -0,0 +1,13 @@
...
+
+    previously LocalOrRemoteProxy would infinitely retry its call as long as
+    the delegate was throwing a NotCurrentLeaderException, regardless of
+    whether the thread was interrupted. Calling a local service behind a
+    AwaitingLeadershipProxy on an interrupted thread would trigger a
+    NotCurrentLeaderException without clearing the interrupt, leading to an
+    infinte loop.
```
felixdesouza:
```
Services running with a leader block no longer get into infinite loops when interrupted and now deal with interrupts correctly.
```

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
```diff
@@ -63,7 +62,7 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
             } catch (InvocationTargetException e) {
                 Throwable targetException = e.getTargetException();
                 if (targetException instanceof NotCurrentLeaderException) {
-                    Uninterruptibles.sleepUninterruptibly(10, TimeUnit.MILLISECONDS);
+                    Thread.sleep(10);
```
felixdesouza: seems a bit fragile to rely on `Thread.sleep` to do the interrupted checking for you considering that there's a `while(true)` loop. If anything I'd check at the top of the while loop whether the thread is interrupted, reset the flag and then react accordingly.

Also since this is a proxy, it's weird to throw a checked exception `InterruptedException` that people can never catch (it's a compile error), especially if there is behaviour attached. Hence why your test references `UndeclaredThrowableException`. 

Tempted to say we should just throw `NotCurrentLeaderException` when we get to this case. Since that's unchecked, and you can "expect" to get that exception at some point when there is no leader for instance.

tjwilson90: If we're not going to propagate an InterruptedException, we should not reset the interrupt flag, otherwise the caller won't know we've been interrupted and respond appropriately.

I'm fine with throwing a different exception, but I don't think it makes much sense to propagate NotCurrentLeaderException specifically. This code is currently designed so that it never throws NotCurrentLeaderException, and it's not actually (necessarily) true that the current node isn't the leader. I think a generic illegal state exception makes the most sense (along with a set interrupted flag).

felixdesouza: Sure the caller won't know, however, if we don't reset the flag, if they call other blocking methods afterwards, they won't clean up, they'll just continue going. 

I suppose what I'm getting at is, making sure that if I call a method on a the proxy which doesn't throw interrupts, but needs that behaviour to be handled i.e. get out of our loop, we close properly. Then proceed to the next statement as if nothing happened.

I suppose there's also a separate piece, where we should also detect when a method is actually throwing an interrupted exception and we do the right thing there e.g. `LockService#lockWithFullResponse`.

I'm happy to take over PR to get this through considering timezones. Let me know what works for you.

tjwilson90: Not clear to me whether we're just talking past each other with different meanings of reset, so I'll clarify.

We need to have the interrupt flag set to true when we throw an exception (assuming we can't throw an InterruptedException), so our caller knows they were interrupted and can respond accordingly. This is what the current patch set does.

I'm fine with you taking over and finishing this.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
```diff
@@ -58,6 +59,10 @@ private LocalOrRemoteProxy(T localService, T remoteService, Future<Boolean> useL
     @Override
     protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
         while (true) {
+            if (Thread.currentThread().isInterrupted()) {
+                // do not clear the interrupt flag
+                throw new SafeIllegalStateException("interrupted");
```
jeremyk-91: May be a misunderstanding: In the event that a `LocalOrRemoteProxy` is interrupted between iterations
(e.g. during the sleep), we will throw a `SafeIllegalStateException` instead of `InterruptedException`, and the interrupt flag is still set so the next interruptible thing will throw an `InterruptedException`. This feels a bit strange - is this what we want?

I'm thinking something like

```
            if (Thread.currentThread().isInterrupted()) {
                // do not clear the interrupt flag
                if (Arrays.stream(method.getExceptionTypes()).anyMatch(p -> p == InterruptedException.class)) {
                    throw new InterruptedException("interrupted");
                }
                throw new SafeIllegalStateException("interrupted");
```

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
felixdesouza: on second thought, the first part doesn't make sense since we need to throw, please ignore, apologies.

tested it with @jeremyk-91 and the interrupted thing is fine already, there was something along the lines of a race that could occur, where we end up throwing the wrong exception even though we *do* want to throw the `InterruptedException`., @jeremyk-91 said he'd comment so there's that.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/LocalOrRemoteProxy.java
tjwilson90: Yes, that is how thread interruption is supposed to work. You want the next blocking call to throw an InterruptedException. Something interrupted this thread and that interrupt has not been handled yet.

Whether we throw an InterruptedException or another type of exception + a set interrupt flag is not super important. Both ultimately result in an InterruptedException and signal to our caller that an interrupt has occurred.

jeremyk-91: Yep. I didn't think it was super important (the basis of the above comment is basically whether we want `InterruptedException` or another type + set flag)

jeremyk-91: I'd probably have written the alternative method, though not a big point as mentioned

PR #4473, opened 2019-12-12
Change logging level of Rollback Other Transaction Logline

Jolyon-S:
Change the level of logging from info to debug for the message `This isn't a bug but it should be very infrequent. Two transactions tried to roll back someone else's request with start: {}` from `SnapshotTransaction` as we already have logging for when that happens.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1960,11 +1960,6 @@ private boolean rollbackOtherTransaction(long startTs, TransactionService transa
             transactionOutcomeMetrics.markRollbackOtherTransaction();
             return true;
         } catch (KeyAlreadyExistsException e) {
-            log.info("This isn't a bug but it should be very infrequent. Two transactions tried to roll back someone"
```
jkozlowski: Let's just drop to DEBUG

PR #4475, opened 2019-12-12
[LW] Part 6: Client Side Wiring

gmaretic:
**Goals (and why)**:
Wiring in lock watches on the client side into transaction manager.

- `TransactionManager` exposes a `TableWatchingService` which can be used to register lock watches.
- `TransactionStarter` uses the new `startTransactionsWithWatches` API, the lock watch information for each transaction started is stored in the `LockWatchEventTracker` and accessible through the `TableWatchingService` using the start timestamp.
- to commit a transaction, `SnapshotTransaction` calls the `getCommitTimestampWithLockInfo` method on the `TableWatchingService`, which then gets a commit timestamp and computes a set of all watched locks that have been taken out since the transaction started, ignoring the locks the transaction itself took.
- `PrecommitCondition` has been replaced by `PreCommitConditionWithWatches`, where the condition can be checked on the object returned by the above call. Existing conditions can easily be converted to a `PreCommitConditionWithWatches`.
- if we are not using timelock in `TransactionManagers`, the implementation of `TableWatchingService` only delegates calls and reports that nothing is watched.
 
Existing tests pass, but will opena follow-up PR with ETE tests

Did I miss anything obvious in the wiring (especially in `TMs.java`)? Does the TWS interface for getting LWS at the beginning of transaction and getting the commit timestamp make sense, or is there a better way of doing this? 

`TransactionStarter`, `SnapshotTransaction`, `TableWatchingServiceImpl`

First work week of next year?


jeremyk-91:
Discussed offline: Please fill in the template normally, it's not easy to evaluate this PR without a better understanding of what you're attempting here.

gmaretic:
Superseded by #4636

PR #4476, opened 2019-12-13
[WIP] Learn leader URL

jkozlowski:
This isn't as easy as I thought. I think there is only a single approach here, which is that learners learn the UUID -> URL association through accepting proposals. At that point we have a problem of how will the proposing leader send the URL: can we add stuff to PaxosValue without breaking the wire? It feels like that should be possible, as long as it is nullable.

### DIFF
--- a/leader-election-api/src/main/java/com/palantir/leader/LeaderElectionService.java
+++ b/leader-election-api/src/main/java/com/palantir/leader/LeaderElectionService.java
@@ -16,6 +16,7 @@
 package com.palantir.leader;
 
 import java.io.Serializable;
+import java.net.URL;
 import java.util.Optional;
 
 import com.palantir.atlasdb.metrics.Timed;
@@ -81,4 +82,6 @@ public interface LeaderElectionService {
      * @return true if and only if this node was able to cause itself to lose leadership
      */
     boolean stepDown();
+
+    Optional<URL> getCurrentLeader();
 }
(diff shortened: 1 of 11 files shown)


felixdesouza:
Don't think this is the right approach, you likely want to tie it to pinging the leader, keeping it mainly inside the paxos leader election service and awaiting leadership proxy.

Paxos acceptor has no concept of leader, just a proposer, e.g. Timelock timestamp paxos.

I have a pretty good wiring story for this anyway.

jkozlowski:
Ah, great, how did I miss that? I was looking for this association.

felixdesouza:
I've spent far too much in general with `PaxosLeaderElectionService` and wiring lol. I'm currently whipping something up.

leader-election-api/src/main/java/com/palantir/leader/LeaderElectionService.java
```diff
@@ -81,4 +82,6 @@
      * @return true if and only if this node was able to cause itself to lose leadership
      */
     boolean stepDown();
+
+    Optional<URL> getCurrentLeader();
```
jeremyk-91: This is too strongly named - I'd prefer `getCurrentSuspectedLeader`, as this may not give you the true identity of the leader (e.g. consider what happens if a leadership proposal is not learned on all learners). Probably also want javadoc explaining what this doesn't guarantee.

leader-election-api/src/main/java/com/palantir/paxos/PaxosValue.java
```diff
@@ -39,6 +39,9 @@
     @Nullable
     final byte[] data;
     final String leaderUuid;
+    @Nullable
+    final String leaderUrl;
```
jeremyk-91: Probably better for offline discussion: I'm not comfortable with adding this to PaxosValue, the concept of a URL doesn't quite exist at this level.

PR #4481, opened 2019-12-16
Ensure that `NotCurrentLeaderExceptionMapper` can leverage the leader hint.

### LINKED ISSUE #4471: Client recovery after leader switch can fail
Refer to internal PDS-107023

felixdesouza:
**Goals (and why)**:
Fixes #4471. Instead of just randomly redirecting when we *do* know the leader, we use the leader hint that's gathered in the background whilst pinging the leader. When we don't know the leader, we still bias towards fast redirects by picking a random leader vs a 503 which will invoke backoff (we don't want this).

Idea here is that we ping the leader and we know who that leader is every 50 milliseconds. We just need to associate a `HostAndPort`/`URL` when we *do* ping. We also want to encode recency since it's not necessarily the case that the logs are up to date, so we only return the `HostAndPort` if we've successfully pinged the leader within some time frame.

Amended current tests to be aware of host and port etc.

Is remembering for 1 second too much? should it be 500ms? Bear in mind that we ping 20 times a second by default in timelock. I have made it 15 seconds in leader-block mode.

* `NotCurrentLeaderExceptionMapper`
* `LeaderElectionService`, `PaxosLeaderElectionService`
* `RetryRedirectTargeter`
* `SingleLeaderPinger`
* `AutobatchingPingableLeaderPingerFactory` if you're interested :P
* `PaxosRuntimeConfiguration` for the cache ttl defaults.

At some point this week, if not then over the christmas break.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
@@ -32,7 +32,6 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.base.Preconditions;
-import com.palantir.conjure.java.api.config.service.HumanReadableDuration;
 import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
 import com.palantir.logsafe.SafeArg;
 
@@ -100,8 +99,8 @@ public abstract class LeaderConfig {
     }
 
     @Value.Default
-    public HumanReadableDuration leaderAddressCacheTtl() {
-        return HumanReadableDuration.seconds(15);
+    public Duration leaderAddressCacheTtl() {
+        return Duration.ofSeconds(15);
     }
 
     @Value.Check
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
@@ -32,6 +32,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.base.Preconditions;
+import com.palantir.conjure.java.api.config.service.HumanReadableDuration;
 import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
 import com.palantir.logsafe.SafeArg;
 
@@ -96,6 +97,11 @@ public abstract class LeaderConfig {
     @Value.Auxiliary
     public Duration leaderPingResponseWait() {
         return Duration.ofMillis(leaderPingResponseWaitMs());
+    }
+
+    @Value.Default
+    public HumanReadableDuration leaderAddressCacheTtl() {
+        return HumanReadableDuration.seconds(15);
     }
 
     @Value.Check
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
@@ -98,6 +98,7 @@ public abstract class LeaderConfig {
         return Duration.ofMillis(leaderPingResponseWaitMs());
     }
 
+    @JsonIgnore
     @Value.Default
     public Duration leaderAddressCacheTtl() {
         return Duration.ofSeconds(15);
--- a/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
@@ -48,26 +48,23 @@ public class ServiceNotAvailableException extends RuntimeException {
         super(message, cause);
         this.serviceHint = Optional.empty();
         this.serviceHintUrl = Optional.empty();
-
     }
 
     public ServiceNotAvailableException(String message) {
         super(message);
         this.serviceHint = Optional.empty();
         this.serviceHintUrl = Optional.empty();
-
     }
 
     public ServiceNotAvailableException(Throwable cause) {
         super(cause);
         this.serviceHint = Optional.empty();
         this.serviceHintUrl = Optional.empty();
-
     }
 
     public ServiceNotAvailableException(String message, Throwable cause, URL leaderHint) {
         super(message, cause);
-        this.serviceHint = Optional.empty();
+        this.serviceHint = Optional.of(HostAndPort.fromParts(leaderHint.getHost(), leaderHint.getPort()));
         this.serviceHintUrl = Optional.of(leaderHint);
     }
 
--- a/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
@@ -15,6 +15,7 @@
  */
 package com.palantir.common.remoting;
 
+import java.net.URL;
 import java.util.Optional;
 
 import com.google.common.net.HostAndPort;
@@ -29,33 +30,52 @@ public class ServiceNotAvailableException extends RuntimeException {
     private static final long serialVersionUID = 2L;
 
     private final Optional<HostAndPort> serviceHint;
+    private final Optional<URL> serviceHintUrl;
 
     public ServiceNotAvailableException(String message, Throwable cause, HostAndPort serviceHint) {
         super(message, cause);
         this.serviceHint = Optional.of(serviceHint);
+        this.serviceHintUrl = Optional.empty();
     }
 
     public ServiceNotAvailableException(String message, HostAndPort serviceHint) {
         super(message);
         this.serviceHint = Optional.of(serviceHint);
+        this.serviceHintUrl = Optional.empty();
     }
 
     public ServiceNotAvailableException(String message, Throwable cause) {
         super(message, cause);
         this.serviceHint = Optional.empty();
+        this.serviceHintUrl = Optional.empty();
+
     }
 
     public ServiceNotAvailableException(String message) {
         super(message);
         this.serviceHint = Optional.empty();
+        this.serviceHintUrl = Optional.empty();
+
     }
 
     public ServiceNotAvailableException(Throwable cause) {
         super(cause);
         this.serviceHint = Optional.empty();
+        this.serviceHintUrl = Optional.empty();
+
+    }
+
+    public ServiceNotAvailableException(String message, Throwable cause, URL leaderHint) {
+        super(message, cause);
+        this.serviceHint = Optional.empty();
+        this.serviceHintUrl = Optional.of(leaderHint);
     }
 
     public Optional<HostAndPort> getServiceHint() {
         return serviceHint;
+    }
+
+    public Optional<URL> getServiceHintAsUrl() {
+        return serviceHintUrl;
     }
 }
--- a/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
@@ -15,7 +15,6 @@
  */
 package com.palantir.common.remoting;
 
-import java.net.URL;
 import java.util.Optional;
 
 import com.google.common.net.HostAndPort;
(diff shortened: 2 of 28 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
```diff
@@ -98,6 +99,11 @@ public Duration leaderPingResponseWait() {
         return Duration.ofMillis(leaderPingResponseWaitMs());
     }
 
+    @Value.Default
+    public HumanReadableDuration leaderAddressCacheTtl() {
+        return HumanReadableDuration.seconds(15);
```
jkozlowski: Do we need to have this be configurable?

jkozlowski: Also, I don't see this actually is used?

felixdesouza: Perhaps initially, whilst we iron it out. It may not need to go all the way to the config, but for tests it's a bit easier if it's configurable at the lower levels.

jkozlowski: Yeah, it looks like you missed wiring in Leaders#createInstrumentedLocalServices.

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosRuntimeConfiguration.java
```diff
@@ -68,6 +68,11 @@ default Duration leaderPingResponseWait() {
         return Duration.ofMillis(leaderPingResponseWaitMs());
     }
 
+    @JsonProperty("leader-address-cache-ttl")
+    default Duration leaderAddressCacheTtl() {
```
jkozlowski: Same here, is this used?

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
jkozlowski: I think it's fine for it to be injectable to the class as a param, so in tests you can turn it off. But then there should just be a default in the default constructor that's used by prod code.

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -61,32 +65,38 @@
     private final LeaderPinger leaderPinger;
     private final PaxosLearnerNetworkClient learnerClient;
 
-    private final long updatePollingRateInMs;
-    private final long randomWaitBeforeProposingLeadership;
+    private final Duration updatePollingRate;
```
jkozlowski: Looks like unrelated refactoring?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -126,8 +136,8 @@ private void proposeLeadershipOrWaitForBackoff(LeadershipState currentState) thr
             return;
         }
 
-        if (pingLeader(currentState.greatestLearnedValue()).isSuccessful()) {
-            Thread.sleep(updatePollingRateInMs);
+        if (isSuccessfulPing(pingLeader(currentState.greatestLearnedValue()))) {
```
jkozlowski: I think this name is misleading, "isSuccessfulPing suggests that you're just checking some invariant, but in fact you're also caching the result of the lookup. I would say that doing the caching in pingLeader is more appropriate.

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -42,13 +43,13 @@
 
     private static final Logger log = LoggerFactory.getLogger(SingleLeaderPinger.class);
 
-    private final ConcurrentMap<UUID, PingableLeader> uuidToServiceCache = Maps.newConcurrentMap();
-    private final Map<PingableLeader, ExecutorService> leaderPingExecutors;
+    private final ConcurrentMap<UUID, LeaderPingerContext<PingableLeader>> uuidToServiceCache = Maps.newConcurrentMap();
```
jkozlowski: Curious why this isn't a cache too? It doesn't look like anything ever expires values from here?

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -58,44 +59,49 @@ public SingleLeaderPinger(
...
+            return getLeaderPingResult(uuid, pingFuture);
         } catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
             return LeaderPingResults.pingCallFailure(ex);
         }
     }
 
-    private static LeaderPingResult getAndRecordLeaderPingResult(
-            @Nullable Future<Map.Entry<PingableLeader, Boolean>> pingFuture) throws InterruptedException {
+    private static LeaderPingResult getLeaderPingResult(
```
jkozlowski: Would it work if instead we just had a method to return HostAndPort for a UUID? then we wouldn't need to change the LeaderPingResult at all, and I feel the diff would be much smaller.

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
felixdesouza: no reason for it not to be, just been like that since whenever really

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
felixdesouza: fair, bit difficult to describe since I want to say "is this successful and if it is cache it, but I don't want to know whether or not the cache operation is successful"

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosRuntimeConfiguration.java
felixdesouza: turns out I didn't wire it in, thanks!

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
jkozlowski: I feel like pingLeader is basically where this should happen, then you don't need a name.

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
felixdesouza: still need it to be on `LeaderConfig`, I've made it `@JsonIgnore` however, so it's only done through code. I've removed it off of `PaxosRuntimeConfiguration`.

It only needs to be on `LeaderConfig` because timelock creates a leader config and calls `Leaders#createInstrumentedLocalServices` as opposed to using the builder itself. This goes away with my partitioning work, since that path is completely changed. Once I bring that back, I can remove it off of `LeaderConfig`.

leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingerContext.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.palantir.paxos;
+
+import org.immutables.value.Value;
+
+import com.google.common.net.HostAndPort;
+
+@Value.Immutable
+@Value.Style(allParameters = true)
+public interface LeaderPingerContext<T> {
```
jkozlowski: Do you need the generic param here?

jkozlowski: This looks like a lot of typing all around the place, maybe we could restructure SingleLeaderPinger somehow, so maybe it just uses HostAndPort everywhere as keys, instead of this thing.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPingableLeaderFactory.java
```diff
@@ -22,36 +22,39 @@
...
+import com.palantir.paxos.LeaderPingerContext;
 
 public class AutobatchingPingableLeaderFactory implements Closeable {
 
     private static final Logger log = LoggerFactory.getLogger(AutobatchingPingableLeaderFactory.class);
 
     private final Collection<? extends Closeable> closeables;
     private final DisruptorAutobatcher<UUID, Optional<ClientAwarePingableLeader>> uuidToRemoteAutobatcher;
-    private final Map<? extends ClientAwarePingableLeader, ExecutorService> executors;
+    private final Map<LeaderPingerContext<BatchPingableLeader>, ExecutorService> executors;
```
jkozlowski: As an aside, all those executor services feel weird, can we not use retrofit and just get non-blocking futures?

felixdesouza: probably, it's a pretty large change, and something I have wanted to do previously, but relatively low priority

leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingerContext.java
felixdesouza: it's not just `SingleLeaderPinger` that uses this. Granted, could probably just make two separate immutable objects that contain `PingableLeader` and `HostAndPort` and `BatchPingableLeader` and `HostAndPort`.

The issue here is that we need to keep the association between the rpc class we just invoked ping on and the the host and port information. It's easier to keep them as a "tuple" as opposed to having another map (in both impls) to associate the rpc class and the address info.

In an ideal world with retrofit, this would be pretty straightforward since we'd just get the information from `Call<T>`. But alas, we are not there.

jkozlowski: I feel like HostAndPort is the unique key though? So maybe the BatchPingableLeader and PingableLeader should be part of the value, not key?

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
felixdesouza: Not necessarily, to do that, we'd need `pinghLeaderWithUuid` to record the same amount of inforamtion it's doing now, and save it locally, and then expose that it up via a method you method. Similarly for `AutobatchingPingableLeaderFactory`. In a sense, it'd look like ping, except that it would return host and port, this feels pretty weird imo.

Throughout, you need to keep the association, so you can't really get rid of it.

leader-election-impl/src/main/java/com/palantir/paxos/LeaderPingerContext.java
felixdesouza: It is unique, but not useful by itself. The `PaxosQuorumChecker` takes in `Map<RPC, Executor>`. If you want to just have a single key (`RPC`), you will need a separate map to keep the association between `HostAndPort` and `RPC` in all the places you will use this.

atlasdb-commons/src/main/java/com/palantir/common/remoting/ServiceNotAvailableException.java
```diff
@@ -29,33 +30,52 @@
...
     public ServiceNotAvailableException(String message, HostAndPort serviceHint) {
         super(message);
         this.serviceHint = Optional.of(serviceHint);
+        this.serviceHintUrl = Optional.empty();
     }
 
     public ServiceNotAvailableException(String message, Throwable cause) {
         super(message, cause);
         this.serviceHint = Optional.empty();
+        this.serviceHintUrl = Optional.empty();
```
felixdesouza: don't really like this, it's somewhat awful, it's also difficult to change because `serviceHint` is used/passed in big internal product.

service hint is now bifurcated depending on which one you created the exception with, I'd prefer using `HostAndPort` over this tbh.

jkozlowski: Yeah, not ideal but we need a URL coming into RetryOther right? do you want to manually hardcode the https part there?

felixdesouza: sure, I'd rather pass the `HostAndPort` around and then establish the mapping inside the only place we want it.

jkozlowski: Sure, that's what it was. I didn't realize that large internal product was setting this so we can't just change the exception. In that case your plan sounds solid

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
```diff
@@ -98,6 +98,12 @@ public Duration leaderPingResponseWait() {
         return Duration.ofMillis(leaderPingResponseWaitMs());
     }
 
+    @JsonIgnore
+    @Value.Default
+    public Duration leaderAddressCacheTtl() {
```
jkozlowski: If we could remove this config, it would be nice.

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -58,44 +59,48 @@ public SingleLeaderPinger(
...
+            LeaderPingerContext<PingableLeader> remote,
+            @Nullable Future<Map.Entry<LeaderPingerContext<PingableLeader>, Boolean>> pingFuture) {
         if (pingFuture == null) {
             return LeaderPingResults.pingTimedOut();
         }
 
         try {
-            boolean isLeader = pingFuture.get().getValue();
-            return LeaderPingResult.fromBooleanResult(isLeader);
+            if (Futures.getDone(pingFuture).getValue()) {
```
jkozlowski: This is great API in MultiplexingCompletionService

atlasdb-api/src/main/java/com/palantir/atlasdb/config/LeaderConfig.java
felixdesouza: I can remove it when I've reintroduced partitioning, which skips this codepath entirely.

felixdesouza:
> * The value of recency to me at least in terms of avoiding failing loops is to check that the node that we are redirecting to is still alive (rather than that it is the leader).
Being a bit pedantic, I would say being alive should be defined as the `AwaitingLeadershipProxy` for that namespace being alive and running in the background.

Issue #4124, opened 2019-07-12
[Generated Code] Fresh Persister Object Created On Every Hydration


jeremyk-91:
Currently, the implementation of the bytes hydrator for a user-specified persister is to create a new instance of the persister on each hydration, and then decompress the values. This leads to unnecessary object allocations when the persister is not stateful, and has been flagged as possibly causing issues for internal shopping product.

For example, for schema code:

```java
        schema.addTableDefinition("table", new TableDefinition() {{
            rowName();
            rowComponent("row", ValueType.STRING);
            columns();
            column("column", "c", JPersister.class);
        }});
```

We get in the generated code:

```
        public static final Hydrator<Column> BYTES_HYDRATOR = new Hydrator<Column>() {
            @Override
            public Column hydrateFromBytes(byte[] bytes) {
                bytes = CompressionUtils.decompress(bytes, Compression.NONE);
                return of(new com.palantir.atlasdb.todo.JPersister().hydrateFromBytes(com.palantir.atlasdb.compress.CompressionUtils.decompress(bytes, com.palantir.atlasdb.table.description.ColumnValueDescription.Compression.NONE)));
            }
        };
```

Note that in general this needs to be configurable and default to off to avoid issues with users whose persisters are stateful (and thus creating a new one on every hydration is required).

felixdesouza:
stateful persisters???

PR #4483, opened 2019-12-17
Configurable perister creation

sudiksha27:
[Referance Issue](https://github.com/palantir/atlasdb/issues/4124)

**Goals (and why)**:
- Add ability to make persister reusable and not to create a new instance on each hydration.  

- `Reusable` annotation is used on persister classes which are stateless and new instance is not required on each hydration.
- The default behaviour is to create a new instance of persister on each hydration to avoid issues with persisters which are stateful.

- Test added to verify generated code in case of reusable persisters.

- So far, the changes cover code generation for named and dynamic column.

- ColumnValueDescription#getHydrateCode

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/annotation/Reusable.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/annotation/Reusable.java
@@ -1,3 +1,18 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


sudiksha27:
When the persister is reusable - 

reusable_persister_hydrator

When the persister is not reusable - 

hydrator_stateful_persister

atlasdb-client/src/main/java/com/palantir/atlasdb/annotation/Reusable.java
```diff
@@ -0,0 +1,11 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface Reusable {
+}
```
sudiksha27: add comments

changelog/@unreleased/pr-4483.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: "Add ability to make persister reusable and not to create a new
+    instance on each hydration. The default behaviour is to create
+    a new instance of persister on each hydration to avoid issues with persisters
+    which are stateful."
```
jeremyk-91: Might be worth mentioning that this is done through specifying the `@Reusable` annotation, and that if users want to enable this they will need to regenerate their schemas.

Also we should say that the persisters are only reusable if they are stateless and thread-safe.

PR #4485, opened 2019-12-23
[CJR] Refresh experimental clients every 30 minutes

gmaretic:
**Goals (and why)**:
Sometimes CJR clients end up in a bad state, and given that now is a rough time to root cause and fix, this is a stop gap measure to alleviate the issue by recreating CJR clients every 30 minutes.

Pretty straightforward

Unit test that verifies the recreation happens in ERP

Is this enough, and is it correct -- or did I miss something obvious?

Anywhere

ASAP

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
@@ -110,10 +110,11 @@ public final class ExperimentRunningProxy<T> extends AbstractInvocationHandler {
         }
     }
 
-    private synchronized void possiblyRefreshClient() {
-        if (Instant.now(clock).compareTo(lastClientRefresh.get().plus(CLIENT_REFRESH_INTERVAL)) > 0) {
+    private void possiblyRefreshClient() {
+        Instant currentLastRefresh = lastClientRefresh.get();
+        if (Instant.now(clock).compareTo(currentLastRefresh.plus(CLIENT_REFRESH_INTERVAL)) > 0 &&
+        lastClientRefresh.compareAndSet(currentLastRefresh, currentLastRefresh.plus(CLIENT_REFRESH_INTERVAL))) {
             experimentalService.set(experimentalServiceSupplier.get());
-            lastClientRefresh.set(Instant.now(clock));
         }
     }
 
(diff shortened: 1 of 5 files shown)


atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
```diff
@@ -100,6 +110,13 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
         }
     }
 
+    private synchronized void possiblyRefreshClient() {
```
jkozlowski: The way this is written, all requests will end up ordered here.

atlasdb-commons/src/main/java/com/palantir/common/proxy/ExperimentRunningProxy.java
```diff
@@ -36,42 +37,48 @@
 public final class ExperimentRunningProxy<T> extends AbstractInvocationHandler {
     private static final Logger log = LoggerFactory.getLogger(ExperimentRunningProxy.class);
     static final Duration REFRESH_INTERVAL = Duration.ofMinutes(10);
+    static final Duration CLIENT_REFRESH_INTERVAL = Duration.ofMinutes(30);
 
-    private final T experimentalService;
+    private final AtomicReference<T> experimentalService;
+    private final Supplier<T> experimentalServiceSupplier;
```
felixdesouza: thoughts on just making this a `Suppliers.memoizeWithExpiration`? the semantics are the same no and should be less intrusive?

gmaretic: Derp, I wanted to make it smarter, but then changed my mind and forgot there was an easy way to do it

gmaretic:
@felixdesouza should be good now

atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
```diff
@@ -52,15 +52,30 @@ private VersionSelectingClients() {
             TargetFactory.InstanceAndVersion<T> legacyClient,
             VersionSelectingConfig versionSelectingConfig,
             Class<T> clazz) {
-        T instrumentedNewClient = instrumentWithClientVersionTag(
-                metricsManager.getTaggedRegistry(), newClient, clazz);
+        return createVersionSelectingClientWithRefreshingNewClient(
+                metricsManager,
+                () -> newClient,
```
jkozlowski: Hmm, is this going to work?

jkozlowski:
So, we know that this has likely been working, cause that's been the default previously, pre-cjr. But I really wonder how: both the connection pool and the dispatcher in cjr's OkHttpClients are static, so I don't know how this would actually force a reset of connections.

```
    /** Shared dispatcher with static executor service. */
    private static final Dispatcher dispatcher;

    /** Shared connection pool. */
    private static final ConnectionPool connectionPool = new ConnectionPool(100, 10, TimeUnit.MINUTES);
```

PR #4487, opened 2020-01-06
Off-heap timestamp cache implementation

OStevan:
**Goals (and why)**:
Implementation of the timestamp cache using RocksDB along with tests.

Implementation of an off heap timestamp cache using persistent storage. Important design decision is that when the cache is full concurrent thread will all try to swap the namespace used for the cache and if another thread changed it in the meantime it will back-off. On the other hand calling clear on the cache will always succeed.

- `OffHeapTimestampCacheTests` tests that the concurrency model is correct
-  `OffHeapTimestampCacheIntegrationTests` tests that everything works in sequential world when using RocksDB

- are the tests as they are bit of an overkill
- concurrency model when the cache is full

- `OffHeapTimestampCache`
- `OffHeapTimestampCacheIntegrationTests`
- `OffHeapTimestampCacheTests`

by Wednesday


OStevan:
Implemented a simplified solution with batching writes. Closing off this PR.

PR #4488, opened 2020-01-07
[LW] Part 7: ETE Tests

gmaretic:
**Goals (and why)**:
Add ETE tests for lock watches, also add JsonSerialize to classes that actually need to be serializable.

Pretty self explanatory.

The tests are a bit hacky in the sense that we control the transaction workflow by manually starting and committing transactions through the timelock service. In particular, we do the same unwrapping of leased lock tokens that we do in `SnapshotTransaction`.

We can add tests that run through the use of TM, but then the test has to be a part of the resource implementation which is not great either. Should we do it anyway? Is there a better solution?

`EteLockWatchResource.java`

This week pls

### DIFF
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/lock/watch/LockWatchEteTest.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/lock/watch/LockWatchEteTest.java
@@ -1,0 +1,87 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/lock/watch/LockWatchEteTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+    @Test
+    public void doesNotFilerOutOurOwnLockIfLockedTwice() {
+        lockWatcher.registerLockWatch(FULL_TABLE);
+        TransactionHandler handler = startTransaction();
+        LockDescriptor ourLockDescriptor = getRandomRowDescriptor(TABLE);
+        LockToken token = lock(ourLockDescriptor);
+        lockWatcher.unlock(token);
+        LockToken ourToken = lock(ourLockDescriptor);
+        assertThat(handler.commitAndGetLocksIgnoring(ourToken)).containsExactly(ourLockDescriptor);
```
jeremyk-91: discussed offline: This manifests in practice when our transaction reads and writes things falling under the same lock descriptor. We could otherwise miss a write-write conflict. Thanks for the test, it's a good one 👍

PR #4491, opened 2020-01-07
[Timelock Partitioning] Part 44e: Bring back partitioning

felixdesouza:
**Goals (and why)**:
Bring back metrics for Timelock. My big hunch is that different names caused it to be borked 😅 . There's also the weird thing of proxies wrapping proxies, but gonna test it out to see if it's actually a problem.

* Pass the class formatted name when creating the metrics as opposed to a (imo nicer) slightly better descriptive name.

Should not affect correctness, just metrics.

Everywhere.

ASAP.


timelock-server/src/main/java/com/palantir/atlasdb/timelock/TimeLockServerLauncher.java
```diff
@@ -44,6 +45,10 @@
  * Provides a way of launching an embedded TimeLock server using Dropwizard. Should only be used in tests.
  */
 public class TimeLockServerLauncher extends Application<CombinedTimeLockServerConfiguration> {
+
+    private static final UserAgent USER_AGENT =
+            UserAgent.of(UserAgent.Agent.of("TimelockServerLauncher", "0.0.0"));
```
jeremyk-91: nit: please match the case

PR #4493, opened 2020-01-07
Add metrics for off-heap timestamp cache.

OStevan:
**Goals (and why)**:
Add metrics for off-heap cache.

- should we have any other metrics?
- would it make sense to split it to multiple classes

- by end of Wednesday


atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
```diff
@@ -126,6 +149,7 @@ private static CacheDescriptor createNamespaceAndConstructCacheProposal(
         @Override
         public Map<Map.Entry<Long, Long>, Map.Entry<Long, Long>> apply(Set<Map.Entry<Long, Long>> request) {
             if (offHeapTimestampCache.cacheDescriptor.get().currentSize().get() >= offHeapTimestampCache.maxSize) {
+                offHeapTimestampCache.taggedMetricRegistry.meter(CACHE_NUKE).mark();
```
jeremyk-91: Hmm. A meter is plausible here, but it registers a bunch of series that I'm not sure are interesting for an operation that I don't think is super common.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
```diff
@@ -126,6 +149,7 @@ private static CacheDescriptor createNamespaceAndConstructCacheProposal(
...
                 offHeapTimestampCache.clear();
```
jeremyk-91: I think this needs to reset `CACHE_SIZE` to 0, or the metric should be something like `CACHE_WRITES` rather than `CACHE_SIZE`.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
OStevan: Ack.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
OStevan: Discussed off-line, since counter does not provide a reset we will write a new `Gauge` which can be reseted.

OStevan:
Metrics regarding the performance of get/put in cache will be addressed during configuration.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
OStevan: Added a resettable counter implementation of a `Gauge`.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
```diff
@@ -84,6 +105,7 @@ public void clear() {
         if (previous != null) {
             persistentTimestampStore.dropNamespace(previous.storeNamespace());
         }
+        taggedMetricRegistry.gauge(CACHE_SIZE, cacheSizeGauge);
```
jeremyk-91: this repeatedly re-registers the gauges

OStevan: Now register once.

PR #4494, opened 2020-01-08
[Timelock Partitioning] Part 44f: More metrics fixes

felixdesouza:
**Goals (and why)**:
So, introduction of the new code path for partitioning makes use of hierarchical metric registries. `LeadershipEvents`, is a class that just logs/tracks metrics of events pertaining to leadership. 

The issue here is that loggers are not hierarchical, and it would be nice to know the context we're in + the client we're talking for. Hence we pass both the `paxosUseCase` and the `client` as `namespaceArgs`. However, the metric registry that is passed in *is* a child registry keyed by `paxosUseCase`, so including the `paxosUseCase` as a safe tag to a metric that will be inserted into its parent with keyed by that is an error and we get the following exception:

```
c.p.logsafe.e.SIAE: Base must not contain the extra key that is to be added
	at com.palantir.logsafe.Preconditions.checkArgument(Preconditions.java:52)
	at com.palantir.tritium.metrics.registry.ExtraEntrySortedMap.<init>(ExtraEntrySortedMap.java:53)
	at com.palantir.tritium.metrics.registry.RealMetricName.create(RealMetricName.java:81)
	at com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry.lambda$getMetrics$3(AbstractTaggedMetricRegistry.java:156)
	at com.google.common.collect.RegularImmutableMap.forEach(RegularImmutableMap.java:154)
	at com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry.lambda$getMetrics$4(AbstractTaggedMetricRegistry.java:155)
	at java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1597)
	at com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry.getMetrics(AbstractTaggedMetricRegistry.java:154)
	...
```

* Leadership events now takes in a logging safe args and a metrics safe args. A bit janky, but don't want to leak timelock implementation details into `leader-election-impl`.

Added the ability to get the `TaggedMetricRegistry` from Dropwizard. Added a smoke test/sanity check that we're able to iterate through the tagged metric registry. This was failing before I added the fix.

ASAP

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
@@ -48,8 +48,11 @@ class LeadershipEvents {
     private final Meter leaderPingReturnedFalse;
     private final Object[] contextArgs;
 
-    LeadershipEvents(TaggedMetricRegistry metrics, List<SafeArg<String>> contextArgs) {
-        Map<String, String> safeTags = safeTags(contextArgs);
+    LeadershipEvents(
+            TaggedMetricRegistry metrics,
+            List<SafeArg<String>> safeLoggingArgs,
+            List<SafeArg<String>> safeMetricArgs) {
+        Map<String, String> safeTags = safeTags(safeMetricArgs);
         gainedLeadership = metrics.meter(withName("leadership.gained", safeTags));
         lostLeadership = metrics.meter(withName("leadership.lost", safeTags));
         noQuorum = metrics.meter(withName("leadership.no-quorum", safeTags));
@@ -58,7 +61,7 @@ class LeadershipEvents {
         leaderPingFailure = metrics.meter(withName("leadership.ping-leader.failure", safeTags));
         leaderPingTimeout = metrics.meter(withName("leadership.ping-leader.timeout", safeTags));
         leaderPingReturnedFalse = metrics.meter(withName("leadership.ping-leader.returned-false", safeTags));
-        this.contextArgs = contextArgs.toArray(new Object[0]);
+        this.contextArgs = safeLoggingArgs.toArray(new Object[0]);
     }
 
     void proposedLeadershipFor(long round) {
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
@@ -130,6 +130,7 @@ public final class Leaders {
                 metricsManager.getTaggedRegistry(),
                 leaderUuid.toString(),
                 leadershipObserver,
+                ImmutableList.of(),
                 ImmutableList.of());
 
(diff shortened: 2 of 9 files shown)


leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -48,8 +48,11 @@
     private final Meter leaderPingReturnedFalse;
     private final Object[] contextArgs;
 
-    LeadershipEvents(TaggedMetricRegistry metrics, List<SafeArg<String>> contextArgs) {
-        Map<String, String> safeTags = safeTags(contextArgs);
+    LeadershipEvents(
+            TaggedMetricRegistry metrics,
```
jkozlowski: TaggedMetricRegistries are nice because of the:

```
void addMetrics(String safeTagName, String safeTagValue, TaggedMetricSet metrics)
```

Method, which basically allows for establishing metric hierarchies. Using this, this class would not have to deal with tags at all, because it would be passed a TaggedMetricRegistry that is scoped to the exact thing it needs, and ALL metrics created underneath it would be tagged correctly.

jkozlowski: Deregistration in the TimelockLeadershipMetrics should also hopefully become just a simple removal of the scoped TaggedMetricSet, so that code should be less error prone.

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -58,7 +61,7 @@
         leaderPingFailure = metrics.meter(withName("leadership.ping-leader.failure", safeTags));
         leaderPingTimeout = metrics.meter(withName("leadership.ping-leader.timeout", safeTags));
         leaderPingReturnedFalse = metrics.meter(withName("leadership.ping-leader.returned-false", safeTags));
-        this.contextArgs = contextArgs.toArray(new Object[0]);
+        this.contextArgs = safeLoggingArgs.toArray(new Object[0]);
     }
 
     void proposedLeadershipFor(long round) {
```
jkozlowski: This logging also feels rather cumbersome: it would be great to use MDC for this. The only problem is we can't assume anything about the logging framework... :(

felixdesouza: yeah, it would

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
felixdesouza: so I'm using that above, I didn't do another level for reasons unbeknownst to me, perhaps I was worried about perf of multi level hierarchies, but I can give it a try.

jkozlowski: If it's not performant enough, we should fix this in the library and not workaround here.

PR #4497, opened 2020-01-09
[LW] Part 8: Limit LWEL window size wrt number of Lock Descriptors

gmaretic:
**Goals (and why)**:
Limit the size of lock watch event log window in terms of the number of lock descriptors and lock watch requests that are tracked.

We maintain the size of all events in the window, when a new event is added that makes us go over the limit, we evict old events until we are under the limit again.
 
Added more tests for both the log and the sliding window classes.

The concurrency is now slightly more complex than before, is it still understandable?

Do we want to do something along the lines of always maintaining a few events in the window, regardless of large events? Currently, a single large event evicts everything else, and then gets evicted itself on the first new event added, almost certainly causing clients to miss out on the updates.

The maximum size of 1000 is a random choice, we should actually figure out what is a reasonable value for this constant.

Either dig directly into `ArrayLockEventSlidingWindow.java` or read through the tests first

After the previous links in the chain

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
@@ -23,6 +23,8 @@ import java.util.OptionalLong;
 
 import javax.annotation.concurrent.ThreadSafe;
 
+import org.immutables.value.Value;
+
 import com.google.common.math.LongMath;
 import com.google.common.primitives.Ints;
 import com.palantir.lock.watch.LockWatchEvent;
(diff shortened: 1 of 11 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -50,10 +57,29 @@ public OptionalLong getVersion() {
...
+                deletingIndex = incrementAndMod(deletingIndex);
+            }
+        }
         nextSequence++;
     }
 
+    private void evictOldestEntryAtIndex(int index) {
+        currentSize = currentSize - buffer[index].size();
+        buffer[index] = PlaceholderLockWatchEvent.INSTANCE;
+        lastVisibleSequence++;
```
jeremyk-91: Possibly better for offline discussion: Should this move up before we actually null out the entry?

PR #4500, opened 2020-01-10
Actually disable concurrency limiting.

jkozlowski:
**Goals (and why)**:

*ClientOptions* currently has a field for clientQos to disable it. But the factory method #serverListToClient used by AtlasDB clients (so clients of timelock) did not actually propagate that setting to com.palantir.conjure.java.client.config.ClientConfiguration. Meaning that since this codepath became the default, concurrency limiting has been enabled. Timelock itself does not send 429s, but concurrency limiters also have an adaptive mechanism, where they slowly release new permits: effectively slowing us down, where we know we don't want this for timelock.

Therefore, the code was simply incomplete.

### DIFF
--- a/changelog/@unreleased/pr-4500.v2.yml
+++ b/changelog/@unreleased/pr-4500.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Actually disable concurrency limiting.
+  links:
+  - https://github.com/palantir/atlasdb/pull/4500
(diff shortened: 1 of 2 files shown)


felixdesouza:
Can you fill out the description? i.e. what caused this, why this is better etc.

jkozlowski:
@felixdesouza done

changelog/@unreleased/pr-4500.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Actually disable concurrency limiting.
```
felixdesouza:
```suggestion
  description: Actually disable concurrency limiting for connections to Timelock.
```

PR #4502, opened 2020-01-10
[LW] 2nd Prequel to Part 5: LockWatchStateUpdater

gmaretic:
**Goals (and why)**:
Visitor that updates currently known lock watch state information.

Pretty straightforward

Unit tests


lock-api/src/main/java/com/palantir/lock/watch/LockWatchStateUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+@NotThreadSafe
+public class LockWatchStateUpdater implements LockWatchEvent.Visitor<Void> {
+    private final RangeSet<LockDescriptor> watches;
+    private final Map<LockDescriptor, LockWatchInfo> lockWatchState;
+    private final Set<UUID> unmatchedOpenLocksEvents;
+
+    /**
+     * LockWatchStateUpdater is a LockWatchEvent visitor that updates the set of current lock watches, the state of
+     * specific locks, and tracks the information on encountered open locks events. It is the responsibility of the
+     * caller to visit lock watch events in the correct order as they occurred, without skipping events.
```
jeremyk-91: By this, do we mean that we make a new updater if we fall out of the window and/or there is a leader election?

lock-api/src/test/java/com/palantir/lock/watch/LockWatchStateUpdaterTest.java
```diff
@@ -0,0 +1,234 @@
...
+        secondOpenLocks.accept(updater);
+        firstEvent.accept(updater);
+        secondEvent.accept(updater);
+
+        assertThat(watches.asRanges())
+                .containsExactlyInAnyOrder(TABLE_1.accept(TO_RANGES_VISITOR), TABLE_3.accept(TO_RANGES_VISITOR));
+    }
+
+    @Test
+    public void overlappingWatchesAreMergedOnUpdates() {
```
jeremyk-91: not sure how I feel about this test, it feels dependent on the contract of `RangeSet` (which is that it *must* coalesce) rather than anything we're doing in the code

lock-api/src/main/java/com/palantir/lock/watch/LockWatchStateUpdater.java
jeremyk-91: Discussed offline: Usage is once per batch of updates, so that's ok - the calling class is responsible for invalidation in the election/too-much-lag cases.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchStateUpdaterTest.java
gmaretic: That's fair, but it's critical for us that re-registering watches does not blow up our ranges so I think it's fine to have this just to verify the invariant. Also, `RangeSet` is technically `@Beta`...

PR #4503, opened 2020-01-10
Off-heap timestamp cache and persistent storage integration

OStevan:
**Goals (and why)**:
Enable usage of off-heap persistent storage for timestamp caching.

- if `TimestampCache` is supplied we use the supplied one
- if persistent storage is not enabled use the in-memory one

- added integration tests to confirm that cache is working

- do we even need the comparing one or just parameterize
- `TransactionManagers`

today


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorages.java
```diff
@@ -17,20 +17,43 @@
...
+    private PersistentStorages() {}
+
+    /**
+     * Given the {@code absolutePath} deletes all sub-folders, sub-files and the folder pointed by it.
+     *
+     * @param absolutePath which we want to delete
+     * @throws IOException if there is an underlying exception
+     */
+    public static void deletePath(Path absolutePath) throws IOException {
+        try (Stream<Path> stream = Files.walk(absolutePath)) {
```
jeremyk-91: Some notes:
- should we validate that the path is an absolute path?
- it feels this method should be private, unless I'm missing something.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/ComparingTimestampCache.java
```diff
@@ -0,0 +1,84 @@
...
+            PersistentTimestampStore persistentTimestampStore) {
+        TimestampCache first = new DefaultTimestampCache(
+                metricRegistry.getRegistry(),
+                () -> AtlasDbConstants.DEFAULT_TIMESTAMP_CACHE_SIZE);
+
+        TimestampCache second = OffHeapTimestampCache.create(
+                persistentTimestampStore,
+                metricRegistry.getTaggedRegistry(),
+                () -> AtlasDbConstants.DEFAULT_TIMESTAMP_CACHE_SIZE
+        );
```
jeremyk-91: nit: indentation

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/ComparingTimestampCache.java
```diff
@@ -0,0 +1,84 @@
...
+        Long firstCommitTimestamp = first.getCommitTimestampIfPresent(startTimestamp);
+        Long secondCommitTimestamp = second.getCommitTimestampIfPresent(startTimestamp);
+        if (firstCommitTimestamp == null || secondCommitTimestamp == null) {
+            return Stream.of(firstCommitTimestamp, secondCommitTimestamp)
+                    .filter(Objects::nonNull)
+                    .findFirst()
+                    .orElse(null);
+        }
+        Preconditions.checkState(firstCommitTimestamp.equals(secondCommitTimestamp),
+                "There is a bug in cache implementation");
```
jeremyk-91: I would include as Args the start timestamp and commit timestamps reported by each of the services in this case, for easier debugging.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -168,6 +170,74 @@ public void shouldNotConflictIfImmutableTimestampLockExpiresIfNoReadsOrWrites()
...
+            return txn;
+        });
+
+        assertThat(timestampCache.getCommitTimestampIfPresent(writer.getTimestamp()))
+                .isNotNull()
+                .isEqualTo(transactionService.get(writer.getTimestamp()));
+    }
+
+    @Test
+    public void abortedReadTransactionStillCachesWritersCommitTimestamp() {
```
jeremyk-91: nit: maybe `readThenAbortStillCaches...`? A `ReadTransaction` is something else, and could lead to confusion

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -168,6 +170,74 @@ public void shouldNotConflictIfImmutableTimestampLockExpiresIfNoReadsOrWrites()
...
+        assertThat(timestampCache.getCommitTimestampIfPresent(writer.getTimestamp())).isNull();
+
+        txMgr.runTaskWithRetry(txn -> {
+            get(txn, TEST_TABLE, "row", "column");
+            return txn;
+        });
+
+        assertThat(timestampCache.getCommitTimestampIfPresent(writer.getTimestamp()))
+                .isNotNull()
+                .isEqualTo(transactionService.get(writer.getTimestamp()));
```
jeremyk-91: In general throughout this class, is `isNotNull()` really needed? I imagine the `isEqualTo` check should be good enough. `Transaction#getTimestamp` returns a primitive `long`, so non-null.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/TransactionManagerTest.java
```diff
@@ -168,6 +170,74 @@ public void shouldNotConflictIfImmutableTimestampLockExpiresIfNoReadsOrWrites()
         txnManagerWithMocks.runTaskThrowOnConflict(txn -> null);
     }
 
+    @Test
+    public void writeDoesNotPopulateTimestampCache() {
+        Transaction transaction = txMgr.runTaskWithRetry(txn -> {
+            put(txn, TEST_TABLE, "row", "column", "test");
+            return txn;
+        });
```
jeremyk-91: Might be cleaner in terms of minimum visibility to just return the timestamp from the transaction task.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -495,6 +514,19 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
+            MetricsManager metricsManager,
+            Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier,
+            Optional<PersistentTimestampStore> persistentTimestampStore) {
+        LongSupplier cacheSize = () -> runtimeConfigSupplier.get().getTimestampCacheSize();
+        Supplier<TimestampCache> timestampCacheSupplier = () ->
+                persistentTimestampStore.map(store ->
+                        OffHeapTimestampCache.create(store, metricsManager.getTaggedRegistry(), cacheSize))
+                        .orElseGet(() -> new DefaultTimestampCache(metricsManager.getRegistry(), cacheSize));
+
+        return config().timestampCache().orElseGet(timestampCacheSupplier);
```
jeremyk-91: I think we should test the logic in this class in this PR, especially since it's on a live path

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -495,6 +509,32 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         return transactionManager;
     }
 
+    @VisibleForTesting
+    static Optional<PersistentTimestampStore> constructPersistentTimestampStore(
```
jeremyk-91: nit: maybe `constructPersistentTimestampStoreIfConfigured`?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -495,6 +509,32 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
+            AtlasDbConfig config,
+            PersistentStorageFactory persistentStorageFactory,
+            @Output List<AutoCloseable> closeables) {
+        return initializeCloseable(
+                config.persistentStorage().map(storageConfig -> persistentStorageFactory
+                        .constructPersistentTimestampStore((RocksDbPersistentStorageConfig) storageConfig)),
```
jeremyk-91: I would `Preconditions.checkState` this for readability (instead of the class cast exception you'd get if we ship a bug here)

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -714,6 +722,88 @@ public void kvsDoesNotRecordSweepStatsIfSweepQueueWritesAndTargetedSweepEnabled(
...
+        Optional<PersistentTimestampStore> persistentTimestampStore =
+                TransactionManagers.constructPersistentTimestampStore(
+                        installConfig,
+                        new DefaultPersistentStorageFactory(),
+                        new LinkedList<>());
+
+        assertThat(persistentTimestampStore)
+                .isEmpty();
+    }
+
```
jeremyk-91: There's probably something around `persistentTimestampStoreNotConstructedEvenIfConfiguredIfExplicitlyProvided`, though I'm okay with passing on that if it's really hard to test.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
OStevan: Renamed.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
OStevan: Added.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
OStevan: Will be resolved in a following PR.

OStevan: After discussing offline we decided that if the config is provided the persistent storage is initialized so this is not useful anymore.

PR #4504, opened 2020-01-10
[LW] Part 5, v2: LockWatchEventTrackerImpl and tests

gmaretic:
**Goals (and why)**:
This PR replaces Part5. Implement `LockWatchEventTracker`. This has two purposes:

1. Serve as a source of latest known lock watch state to minimise the update size when requests are made to timelock.
2. Store and expose the lock watch state for each in-flight transaction.

1. We can try updating the tracker with each update we receive from timelock. In case we try updating with stale updates, this is mostly a no-op (except in the case of a leader election, but it will eventually settle). In case of a discontinuity (which cannot happen with correct usage), it is also safe to no-op as the update is best effort.
2. The implementation of this is trivial, the wiring in is not in the scope of this PR.

There are unit tests

The concurrency in here is controlled via a read/write lock mechanism, which I have reasoned through carefully. The rest should be pretty self-explanatory, but please do flag any concerns.

In particular, the update operation returning the most recent known lock watch state can be misleading in the case when the update is stale, so made the method void to explicitly force a call to `currentState()`. This means that clients will need to compute their own lock watch state from a call to timelock, as they know the state that was used for the call and the update.


lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventTrackerImpl.java
```diff
@@ -0,0 +1,187 @@
...
+@ThreadSafe
+public class LockWatchEventTrackerImpl implements LockWatchEventTracker {
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final AtomicReference<RangeSet<LockDescriptor>> watches = new AtomicReference<>(TreeRangeSet.create());
+    private final AtomicReference<Map<LockDescriptor, LockWatchInfo>> singleLocks = new AtomicReference<>(
+            ImmutableMap.of());
+    private final ConcurrentSkipListSet<UUID> openLocksEvents = new ConcurrentSkipListSet<>();
+    private final ConcurrentMap<Long, VersionedLockWatchState> startTsToLockWatchState = Maps.newConcurrentMap();
+    private volatile OptionalLong lastKnownVersion = OptionalLong.empty();
+    private volatile UUID leaderId = UUID.randomUUID();
```
jeremyk-91: I'm not entirely sure these actually need to be volatile (since readers take either the read lock or the object lock, and the release of the write or object locks after a write happens-before the read lock or object lock respectively can be acquired). I'm not 100% convinced that removing this is safe, so maybe we should leave as is

gmaretic:
Benching this for now, will probably revisit in the future

PR #4505, opened 2020-01-13
Update certs and fix a conflicting test

OStevan:
**Goals (and why)**:
Certificates have expired, this PR replaces the old ones with the new ones valid for the next month. In future this should be an excavator job. Also updates a test which can break others


felixdesouza:
Thanks!

felixdesouza:
Doing the Lord's work @OStevan 💃 😅

PR #4508, opened 2020-01-13
Better coalescing supplier

j-baker:
This is a more readable version of the present coalescing supplier. Won't describe present state of world (algorithm is confusing) but the new implementation is easier to understand, and in some cases demonstrates better performance.

For each round, either we are the first to arrive (execute and return)
or we are not.

In the case we are not, we await the current round ending and then
perform the check again. If not this time, we wait for the executor to
finish and return their result.

The improvement is hard to measure in synthetic benchmarks. With spinning threads it's actually slower than the present day (due to the fair queue, the present code maximizes the likelihood that the batch will be full, whereas the new code will likely ensure that the completing task starts the next batch.

With some random noise delays added to smooth out this effect, performance is very similar to at present, possibly slightly less variance. It's possible to tweak any benchmark by adding extraneous loading.

1. The delays we're seeing are unpredictable - the sorts of things that would be the fault of the OS scheduler.
2. This solution does not show the bad behaviour that we think we're seeing of a thread waking another thread up and that thread not being scheduled for a while.


atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
```diff
@@ -29,52 +29,61 @@
...
-    private void completeOrWaitForCompletion(CompletableFuture<T> future) {
-        fairLock.lock();
-        try {
-            resetAndCompleteIfNotCompleted(future);
-        } finally {
-            fairLock.unlock();
+    private final class Round {
+        private final AtomicBoolean hasStarted = new AtomicBoolean(false);
+        private final CompletableFuture<T> future = new CompletableFuture<>();
+        private volatile Round next;
```
j-baker: it's unclear to me based on Java semantics whether this needs to be volatile or not. But I'm gonna leave it as such.

atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
```diff
@@ -29,52 +29,61 @@
...
+
+        boolean isFirstToArrive() {
+            return !hasStarted.get() && hasStarted.compareAndSet(false, true);
```
j-baker: This is much faster than the equivalent 'hasStarted.compareAndSet(false, true)'. I believe this is because the proposed solution can do the read in a MESI shared state, whereas the CAS will always make it exclusive.

atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
```diff
@@ -29,52 +28,61 @@
...
-    }
 
-    private void resetAndCompleteIfNotCompleted(CompletableFuture<T> future) {
-        if (future.isDone()) {
-            return;
+        Round awaitDone() {
+            try {
+                future.join();
+            } catch (CompletionException e) {
+                // ignore
```
schlosna: why are we eating this and not at least logging at lower level?

j-baker: it's intended behaviour - we're literally just awaiting the conclusion of the future

j-baker: we just want it to be complete. We shouldn't do anything with the exception because it's never relevant in this method.

atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
jeremyk-91: I think not, since completing a future does happen before a join on that future. Not 100% confident in this one so agree with leaving it as such.

atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
jeremyk-91: Makes sense. I think we should document this in a comment here.

PR #4509, opened 2020-01-14
Improve logging for TransactionManagers

OStevan:
**Goals (and why)**:
Improve logging by splitting loggers for `TransactionManagers`.

Might break internal products.


felixdesouza:
Have you checked big internal product?

OStevan:
Big internal product does not use `AbstractTransactionManager`.

PR #4511, opened 2020-01-15
[Timelock Partitioning] Part 49: Granular health checks.

felixdesouza:
**Goals (and why)**:
The current health checks don't really make any sense with multiple leaders for different clients. 

There are two ways to think about it:
* Are a quorum of nodes up?
* Of the clients that I'm aware of (active nodes) is there a node who believes that they are the leader for it?

* I've gone with the second route.

Modified existing tests to do be in terms of clients. Added another test to see different clients tracked separately.

A bit of a break for internal timelock, but should be addressed later. Is there any reason why we don't just 

* `LeaderPingHealthCheck(Test)`
* `NamespaceTracker`

ASAP, since this is blocking/conflicting with merging the other stuff in.

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
@@ -74,7 +74,7 @@ public interface Dependencies {
         com.palantir.atlasdb.timelock.paxos.NetworkClientFactories networkClientFactories();
     }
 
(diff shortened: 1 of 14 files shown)


felixdesouza:
timelock calls this method like it currently does, and if it sees anything that's not `ONE_LEADER`, it will warn and then error (the same way it does now), concat messages for each bad case etc.

out of curiosity is there any reason why we just don't use the same mechanism for checking if we're still the leader? i.e. is latest bla bla bla

PR #4512, opened 2020-01-15
Persistent storage refactoring

OStevan:
**Goals (and why)**:
Refactor persistent storage for easier usage.

- split into physical and logical layer
- timestamp serialization and deserialization is not in `TimestampStore`

- tests are also refactored to reflect the new changes

- `PersistentStore`

- tomorrow


atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PersistentStore.java
```diff
@@ -27,76 +27,76 @@
...
      *
-     * @param storeNamespace handle to the namespace from which we want to retrieve the commit timestamp
-     * @param startTs        entry key for which we want to retrieve commit timestamp
-     * @return the associated timestamp or null if the entry is missing
-     * @throws com.palantir.logsafe.exceptions.SafeIllegalArgumentException when {@code storeNamespace} is a
-     * handle to a non existing namespace
+     * @param storeNamespace handle to the namespace from which we want to retrieve the value
+     * @param key            entry key for which we want to retrieve the value
+     * @return the associated value or null if the entry is missing
+     * @throws com.palantir.logsafe.exceptions.SafeIllegalArgumentException when {@code storeNamespace} is a handle to a
```
felixdesouza: can this be imported? it pushes everything to the far right

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
```diff
@@ -0,0 +1,79 @@
...
+    void multiPut(StoreNamespace storeNamespace, Map<K, V> toWrite);
+
+    /**
+     * Creates a {@link StoreNamespace} with the given name. Multiple calls with the same {@code name} will return
+     * different namespaces.
+     *
+     * @param name of the namespace
+     * @return handle to the underlying namespace
+     */
+    StoreNamespace createNamespace(@Nonnull String name);
```
felixdesouza: The name is misleading and unintuitive and not really comparable to a `TableReference`. It should be called something else.

felixdesouza: also, I'd drop the concept of namespace, it's currently overloaded i.e. it's not the same thing as a timelock namespace or a c* namespace/keyspace and it's an extra concept. Something along the lines of: `createHandleForNamespace` if that's the distinction you're going for.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import com.palantir.atlasdb.persistent.api.PersistentStore.StoreNamespace;
+
+public interface LogicalPersistentStore<K, V> {
```
felixdesouza: `Persistent` -> `Persistence`? 

Is this generic i.e. it's not necessarily for a `TimestampCache`? Do we want this outside of the original goal?

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Retrieve the value associated with the given {@code key}.
+     *
+     * @param storeNamespace from which to retrieve the value
+     * @param key            of the cache entry
+     * @return value associated or null if the entry is missing
+     */
+    @Nullable
```
felixdesouza: if you are making a new api, can it not be `@Nullable` and use `Optional`'s?

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
```diff
@@ -0,0 +1,79 @@
...
+    V get(StoreNamespace storeNamespace, @Nonnull K key);
+
+    /**
+     * Retreives values for all supplied {@code keys}.
+     *
+     * @param storeNamespace from which to retrieve the values
+     * @param keys           for which entries we are interested in
+     * @return map of key, value pairs containing only entries which are stored
+     */
+    Map<K, V> multiGet(StoreNamespace storeNamespace, List<K> keys);
```
felixdesouza: nit: will defer to @jeremyk-91, just overload `get` and `put`, as opposed to multiget?

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
```diff
@@ -0,0 +1,79 @@
...
+     * Creates a {@link StoreNamespace} with the given name. Multiple calls with the same {@code name} will return
+     * different namespaces.
+     *
+     * @param name of the namespace
+     * @return handle to the underlying namespace
+     */
+    StoreNamespace createNamespace(@Nonnull String name);
+
+    /**
+     * Drops the namespace.
```
felixdesouza: What does dropping even mean here? This class is super abstract and it's still unclear what it's for.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
```diff
@@ -132,8 +134,8 @@ public Long getCommitTimestampIfPresent(Long startTimestamp) {
     }
 
     private static CacheDescriptor createNamespaceAndConstructCacheProposal(
-            PersistentTimestampStore persistentTimestampStore) {
-        StoreNamespace proposal = persistentTimestampStore.createNamespace(TIMESTAMP_CACHE_NAMESPACE);
+            LogicalPersistentStore<?, ?> persistentStore) {
```
felixdesouza: what's with the question mark generics?

proposal/namespace/descriptor? can we agree on the terminology?

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/TimestampStore.java
```diff
@@ -0,0 +1,105 @@
...
+import com.palantir.atlasdb.persistent.api.PersistentStore;
+import com.palantir.atlasdb.persistent.api.PersistentStore.StoreNamespace;
+import com.palantir.atlasdb.table.description.ValueType;
+import com.palantir.common.streams.KeyedStream;
+
+/**
+ * Stores timestamps using delta encoding for commit timestamp.
+ */
+public class TimestampStore implements LogicalPersistentStore<Long, Long> {
+    private final PersistentStore physicalPersistentStore;
```
felixdesouza: If you're making the distinction between logical and physical, then the physical one should actually be `PhysicalPersistentStore`.

atlasdb-client/src/main/java/com/palantir/atlasdb/persistent/rocksdb/RocksDbPersistentStore.java
```diff
@@ -46,52 +46,44 @@
...
- * backed by RocksDB ColumnFamilies such that calling {@link RocksDbPersistentTimestampStore#createNamespace(String)}
- * with the same name will construct a new {@link ColumnFamilyHandle} for each call.
+ * Implementation of the {@link PersistentStore} using RocksDB as the underlying persistent storage. Created
+ * {@link StoreNamespace}s are backed by RocksDB ColumnFamilies such that calling
+ * {@link RocksDbPersistentStore#createNamespace(String)} with the same name will construct a new
+ * {@link ColumnFamilyHandle} for each call.
  */
-public final class RocksDbPersistentTimestampStore implements PersistentTimestampStore {
-    private static final Logger log = LoggerFactory.getLogger(RocksDbPersistentTimestampStore.class);
+public final class RocksDbPersistentStore implements PersistentStore {
```
felixdesouza: It's still unclear to me the purpose of the refactoring here, this can still be in terms of timestamps no? do we actually have another use case for rocksdb?

atlasdb-client/src/test/java/com/palantir/atlasdb/cache/OffHeapTimestampCacheIntegrationTests.java
```diff
@@ -29,8 +29,8 @@
 import org.rocksdb.RocksDB;
 import org.rocksdb.RocksDBException;
 
-import com.palantir.atlasdb.persistent.api.PersistentTimestampStore;
-import com.palantir.atlasdb.persistent.rocksdb.RocksDbPersistentTimestampStore;
+import com.palantir.atlasdb.persistent.api.PersistentStore;
+import com.palantir.atlasdb.persistent.rocksdb.RocksDbPersistentStore;
 import com.palantir.atlasdb.util.MetricsManagers;
 
 public final class OffHeapTimestampCacheIntegrationTests {
```
felixdesouza: Put rocks somewhere in the title, since this *is* an integration test of rocksdb

atlasdb-client/src/test/java/com/palantir/atlasdb/cache/TimestampStoreTests.java
```diff
@@ -0,0 +1,91 @@
...
+import org.rocksdb.RocksDB;
+import org.rocksdb.RocksDBException;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import com.palantir.atlasdb.persistent.api.PersistentStore.StoreNamespace;
+import com.palantir.atlasdb.persistent.rocksdb.RocksDbPersistentStore;
+
+public final class TimestampStoreTests {
```
felixdesouza: similarly, add rocksdb somewhere to the title

felixdesouza: or you can not use rocks here, and provide a mocked physical persistent store

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactory.java
```diff
@@ -17,14 +17,15 @@
...
     /**
-     * Constructs a {@link com.palantir.atlasdb.persistent.rocksdb.RocksDbPersistentTimestampStore} using the
-     * supplied configuration.
+     * Constructs a {@link RocksDbPersistentStore} using the supplied configuration.
+     *
      * @param config to use to configure the store
      * @return store to be used
      */
-    PersistentTimestampStore constructPersistentTimestampStore(RocksDbPersistentStorageConfig config);
+    PersistentStore constructPersistentStore(RocksDbPersistentStorageConfig config);
```
felixdesouza: As an aside, this is weird. Everything is quite generic until it isn't. Similar to the async c* stuff, this should be a `PersistentStorageConfig`.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PersistentStore.java
OStevan: Imported.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
OStevan: The way things are used in internal ski product I need to store things which might not be just Long -> Long, so to be able to support that I need generics.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactory.java
OStevan: This was intentional since there is only one config so this is to keep it simple. Can be address in following PRs.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/TimestampStore.java
OStevan: Renamed.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
OStevan: It is now optional.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapTimestampCache.java
OStevan: Proposal and descriptor are internal to cache.

OStevan:
Dropped in favor of a different approach.

OStevan:
Alternative approach is also involved @jeremyk-91

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
OStevan: Added a more detailed doc. The explanation is in the previous comment.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
jeremyk-91: Yep, overloading seems reasonable

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/LogicalPersistentStore.java
jeremyk-91: Agree, though I think we should track this on another PR (`StoreNamespace` is already used quite a lot through this project).

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PhysicalPersistentStore.java
```diff
@@ -0,0 +1,97 @@
...
+    Optional<byte[]> get(StoreNamespace storeNamespace, @Nonnull byte[] key) throws SafeIllegalArgumentException;
+
+    /**
+     * Gets the values associated with the entries specified by {@code keys}.
+     *
+     * @param storeNamespace handle to the namespace from which we want to retrieve the values
+     * @param keys           representing keys for which we want to retrieve the values
+     * @return a map from keys to values
+     */
+    Map<byte[], byte[]> multiGet(StoreNamespace storeNamespace, List<byte[]> keys);
```
jeremyk-91: based on @felixdesouza's suggestion above, we can just overload `get`/`put` here.

Also I'd prefer to try to avoid unqualified `Map<byte[], byte[]>` as it's error-prone because byte arrays by default go by reference equality

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PhysicalPersistentStore.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Gets the value associated with the entry specified by {@code key}.
+     *
+     * @param storeNamespace handle to the namespace from which we want to retrieve the value
+     * @param key            entry key for which we want to retrieve the value
+     * @return the {@link Optional} containing the value or empty if there is no associated value
+     * @throws SafeIllegalArgumentException when {@code storeNamespace} is a handle to a non existing namespace
+     */
+    Optional<byte[]> get(StoreNamespace storeNamespace, @Nonnull byte[] key) throws SafeIllegalArgumentException;
```
jeremyk-91: nit: No need to declare runtime exceptions

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PhysicalPersistentStore.java
```diff
@@ -0,0 +1,97 @@
...
+     * Stores the {@code value} for the associated {@code key} while overwriting the existing value in the specified
+     * {@code storeNamespace}.
+     *
+     * @param storeNamespace of the store to which we should store the entry
+     * @param key            entry key
+     * @param value          entry value
+     * @throws SafeIllegalArgumentException when {@code storeNamespace} is a handle to a non existing namespace
+     */
+    void put(StoreNamespace storeNamespace, @Nonnull byte[] key, @Nonnull byte[] value)
+            throws SafeIllegalArgumentException;
```
jeremyk-91: nit: runtime exception

atlasdb-client/src/main/java/com/palantir/atlasdb/persistent/rocksdb/RocksDbPhysicalPersistentStore.java
```diff
@@ -101,23 +92,19 @@ public Long get(StoreNamespace storeNamespace, @Nonnull Long startTs) {
                 Streams.zip(
                         keys.stream(),
                         byteValues.stream(),
-                        (key, value) -> Maps.immutableEntry(key, deserializeValue(key, value))))
+                        Maps::immutableEntry))
                 .filter(Objects::nonNull)
                 .collectToMap();
```
jeremyk-91: I'm a bit worried about maps with byte-array keys. Let's talk through this in person.

atlasdb-client/src/test/java/com/palantir/atlasdb/cache/RocksDbTimestampStoreTests.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void multiGetFilters() {
+        timestampStore.put(defaultNamespace, 1L, 2L);
+
+        assertThat(timestampStore.multiGet(defaultNamespace, ImmutableList.of(1L, 3L)))
+                .containsExactly(Maps.immutableEntry(1L, 2L));
+    }
+
```
jeremyk-91: nit: probably want multiget returning multiple values?

atlasdb-client/src/test/java/com/palantir/atlasdb/persistent/rocksdb/RocksDbPhysicalPersistentStoreTests.java
```diff
@@ -115,21 +119,21 @@ public void droppingTwoTimesFailsOnSecond() {
...
+        timestampMappingStore.put(defaultNamespace, KEY2, VALUE2);
 
-        assertThat(timestampMappingStore.multiGet(defaultNamespace, ImmutableList.of(1L, 2L, 3L)))
+        assertThat(timestampMappingStore.multiGet(defaultNamespace, ImmutableList.of(KEY, KEY2, "random".getBytes())))
                 .containsExactlyInAnyOrderEntriesOf(ImmutableMap.of(
-                        1L, 2L,
-                        3L, 4L)
+                        KEY, VALUE,
+                        KEY2, VALUE2)
                 );
```
jeremyk-91: This assertion feels fragile. In particular if I change the assertion to

```
        assertThat(timestampMappingStore.multiGet(defaultNamespace, ImmutableList.of(KEY, KEY2, "random".getBytes())))
                .containsExactlyInAnyOrderEntriesOf(ImmutableMap.of(
                        KEY, VALUE,
                        "key2".getBytes(), VALUE2)
                );
```

it fails. In general I think we want to either use ByteString for the return type of the persistent store, or use sorted maps.

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PhysicalPersistentStore.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Stores the entry pairs given in {@code toWrite}, overwriting the existing values.
+     *
+     * @param storeNamespace of the store to which we should store the entry
+     * @param toWrite        entry pairs to write
+     * @throws SafeIllegalArgumentException when {@code storeNamespace} is a handle to a non existing namespace
+     */
+    void multiPut(StoreNamespace storeNamespace, Map<byte[], byte[]> toWrite);
```
jeremyk-91: similarly, `Map<byte[], byte[]>` can be tricky

atlasdb-api/src/main/java/com/palantir/atlasdb/persistent/api/PhysicalPersistentStore.java
OStevan: Changed to `ByteString`.

atlasdb-client/src/test/java/com/palantir/atlasdb/cache/RocksDbTimestampStoreTests.java
OStevan: Added more values.

PR #4516, opened 2020-01-16
Log and limit path sanitization.

OStevan:
**Goals (and why)**:
Add logging and prevent uncontrollable deletion of folders. In cases when a service stops ungracefully the folders used by rocks will not be cleaned. This is the simplest way to prevent unlimited disk consumption.

- add a limit of two subfolders with UUID name pattern to be deleted
- log deleted subfolders

- added a test for limiting

this week

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
@@ -42,6 +42,7 @@ public final class PersistentStorageFactories {
             "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
 
     private static final Set<String> SANITIZED_PATHS = new HashSet<>();
+    private static final int DEFAULT_DELETION_LIMIT = 2;
 
     private PersistentStorageFactories() {}
 
@@ -91,13 +92,13 @@ public final class PersistentStorageFactories {
     }
 
     private static List<File> limitFoldersToDelete(List<File> folders) {
-        if (folders.size() > 2) {
+        if (folders.size() > DEFAULT_DELETION_LIMIT) {
             log.warn(
                     "You are trying to delete more that two UUID named folders during persistent storage start-up. "
                             + "Only two will be deleted.",
                     SafeArg.of("number", folders.size()));
         }
-        return folders.stream().limit(2).collect(Collectors.toList());
+        return folders.stream().limit(DEFAULT_DELETION_LIMIT).collect(Collectors.toList());
     }
 
     private static void deleteDirectories(List<File> directoryContentToDelete) {
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
@@ -33,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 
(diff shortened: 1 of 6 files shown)


felixdesouza:
What's the motivation for this? Why do you need it?

OStevan:
@felixdesouza added a more detailed explanation in the PR description.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
```diff
@@ -68,23 +73,50 @@ public static synchronized void sanitizeStoragePath(String storagePath) {
...
+    private static boolean folderFilter(File file) {
+        return file.isDirectory() && (UUID_PATTERN.matcher(file.getName()).matches());
+    }
+
+    private static List<File> limitFoldersToDelete(List<File> folders) {
+        if (folders.size() > DEFAULT_DELETION_LIMIT) {
+            log.warn(
+                    "You are trying to delete more that two UUID named folders during persistent storage start-up. "
+                            + "Only two will be deleted.",
+                    SafeArg.of("number", folders.size()));
```
jeremyk-91: nit: Some of our products still use the `{}` syntax, I think? @felixdesouza for sanity check - I know I'm not as up to date on logging SOTW

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
```diff
@@ -68,23 +73,50 @@ public static synchronized void sanitizeStoragePath(String storagePath) {
...
+        }
+        return folders.stream().limit(DEFAULT_DELETION_LIMIT).collect(Collectors.toList());
```
jeremyk-91: I think I'd prefer to delete 0 if there are >2 since the most likely state of the world then is that you're in a wrong directory

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
```diff
@@ -28,19 +29,24 @@
...
+import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 
 public final class PersistentStorageFactories {
+    private static final Logger log = LoggerFactory.getLogger(PersistentStorageFactories.class);
     private static final Pattern UUID_PATTERN = Pattern.compile(
             "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
 
     private static final Set<String> SANITIZED_PATHS = new HashSet<>();
```
felixdesouza: static mutable state...???

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
felixdesouza: not necessary, it's not parsed, everywhere is using the new logging format

felixdesouza:
Still not really following.

This either works as it did before the PR, or it does *less*, so how does it prevent unlimited disk consumption?

What's the rationale behind uncontrollable deletion?

What's the actual failure mode that happened that you're trying to fix.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
OStevan: Yes this very sad, I'll have to refactor this in a separate PR.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
OStevan: Now deletes 0.

OStevan:
@felixdesouza when I come back we can discuss in person.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStorageFactories.java
jkozlowski: Curious how this ended up on develop :)

OStevan:
PR description is completely different than the current implementation, therefor closed.

PR #4517, opened 2020-01-16
Improve TimeLock Client Recovery From Network Partitions

jeremyk-91:
**Goals (and why)**:
- Improve recovery times when a node is partitioned off from the network. Previously, if a leader was partitioned off, talking to it as a client would take about 50 seconds before the leader realised it was no longer the leader.
- Fail faster when a quorum of nodes is unavailable.

- Don't retry for the single-leader `PaxosAcceptor`, because this is handled at a higher level in `AwaitingLeadershipProxy`. Previously, we had multiple levels of retries meaning recovery time was often 50 seconds.

Run `MultiNodePaxosTimeLockServerIntegrationTest#leaderLosesLeadershipIfQuorumIsNotAlive`. Previously 55 seconds, now about 8. Sadly nothing automated and I'm not so keen on putting a time limit on the test.

- Is there a cleaner way to test this?

**Where should we start reviewing?**: PaxosRemoteClients.java

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
@@ -198,7 +198,7 @@ public final class AwaitingLeadershipProxy<T> extends AbstractInvocationHandler 
         Object delegate = delegateRef.get();
         StillLeadingStatus leading = null;
 
-        // Wait for at most 6.7 seconds in case of a momentary network partition
+        // Wait for at most 6.3 seconds in case of a momentary network partition
         for (int i = 0; i < MAX_NO_QUORUM_RETRIES; i++) {
             // TODO(nziebart): check if leadershipTokenRef has been nulled out between iterations?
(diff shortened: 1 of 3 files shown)


leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -195,12 +197,17 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
...
+
+        // Wait for at most 6.3 seconds in case of a momentary network partition
         for (int i = 0; i < MAX_NO_QUORUM_RETRIES; i++) {
             // TODO(nziebart): check if leadershipTokenRef has been nulled out between iterations?
             leading = leaderElectionService.isStillLeading(leadershipToken);
             if (leading != StillLeadingStatus.NO_QUORUM) {
                 break;
             }
+            if (i != MAX_NO_QUORUM_RETRIES - 1) {
+                Uninterruptibles.sleepUninterruptibly(700, TimeUnit.MILLISECONDS);
```
felixdesouza: haha nice!

PR #4519, opened 2020-01-16
Shorter Nonblocking Calls, Part 2: Configurable Support in Parameters

jeremyk-91:
**Goals (and why)**:
- Allow users to specify whether a proxy should support calls that block on the server (that is, have a long read timeout). This allows for faster failure given some of the awkward remoting issues we've run into.
- Note that this PR does not change anything on the production code paths yet.

- Instead of using two constants for (production versions of) `ClientOptions`, provide a function that converts `AuxiliaryRemotingParameters` into `ClientOptions`.
- Add a `shouldSupportBlockingOperations` boolean parameter to `AuxiliaryRemotingParameters`; read timeouts vary depending on whether this is specified.
- This is true by default, since `AtlasDbHttpClients` and `ServiceCreator` are public API and used by other services.

Added tests for the new functionality, and regression tests in light of the refactoring

- The plan is for _only_ the lock service to have this specified. Does this make sense?

**Where should we start reviewing?**: `ClientOptions.java`

### DIFF
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
@@ -38,11 +38,18 @@ public abstract class ClientOptions {
     private static final Duration CONNECT_TIMEOUT = Duration.ofMillis(500);
 
     @VisibleForTesting
+    // The read timeout controls how long the client waits to receive the first byte from the server before giving up,
+    // so in general read timeouts should not be set to less than what is considered an acceptable time for the server
+    // to give a suitable response.
+    // In the context of TimeLock, this timeout must be longer than how long an AwaitingLeadershipProxy takes to
+    // decide whether a node is the leader and still has a quorum.
     static final Duration NON_BLOCKING_READ_TIMEOUT = Duration.ofMillis(12566); // Odd number for debugging
+
+    // Should not be reduced below 65 seconds to support workflows involving locking.
     static final Duration BLOCKING_READ_TIMEOUT = Duration.ofSeconds(65);
 
+    // Under standard settings, throws after expected outages of 1/2 * 0.01 * (2^13 - 1) = 40.96 s
     private static final Duration STANDARD_BACKOFF_SLOT_SIZE = Duration.ofMillis(10);
-
     private static final int STANDARD_MAX_RETRIES = 13;
     private static final int NO_RETRIES = 0;
 
@@ -135,15 +142,6 @@ public abstract class ClientOptions {
                 .build();
     }
 
-    /**
-     * If {@link AuxiliaryRemotingParameters#shouldSupportBlockingOperations()} is configured to true, sets the read
-     * timeout of operations to {@link ClientOptions#BLOCKING_READ_TIMEOUT}; otherwise sets this to
-     * {@link ClientOptions#NON_BLOCKING_READ_TIMEOUT}. This controls how long the client waits to receive the first
-     * byte from the server before giving up, so in general read timeouts should not be set to less than what is
-     * considered an acceptable time for the server to give a suitable response.
-     *
-     * Connect timeouts are not affected by the value of {@param parameters}.
-     */
     private static void setupTimeouts(ImmutableClientOptions.Builder builder, AuxiliaryRemotingParameters parameters) {
         builder.connectTimeout(CONNECT_TIMEOUT)
                 .readTimeout(parameters.shouldSupportBlockingOperations()
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
@@ -38,7 +38,7 @@ public abstract class ClientOptions {
     private static final Duration CONNECT_TIMEOUT = Duration.ofMillis(500);
 
     @VisibleForTesting
-    static final Duration NON_BLOCKING_READ_TIMEOUT = Duration.ofSeconds(1);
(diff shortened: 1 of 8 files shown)


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -33,28 +35,21 @@
...
-    public static final ClientOptions DEFAULT_NO_RETRYING = ImmutableClientOptions.builder()
-            .connectTimeout(Duration.ofSeconds(10))
-            .readTimeout(Duration.ofSeconds(65))
-            .backoffSlotSize(Duration.ofMillis(100))
-            .failedUrlCooldown(Duration.ofMillis(1))
-            .maxNumRetries(0)
-            .clientQoS(ClientConfiguration.ClientQoS.DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS)
-            .build();
+    @VisibleForTesting
+    static final Duration NON_BLOCKING_READ_TIMEOUT = Duration.ofMillis(12566); // Odd number for debugging
```
felixdesouza: 💃

atlasdb-conjure/src/test/java/com/palantir/atlasdb/http/v2/ClientOptionsTest.java
```diff
@@ -0,0 +1,107 @@
...
+                .shouldSupportBlockingOperations(false)
+                .userAgent(USER_AGENT)
+                .build());
+
+        assertThat(clientOptions.readTimeout()).isEqualTo(ClientOptions.NON_BLOCKING_READ_TIMEOUT);
+    }
+
+    @Test
+    public void defaultRetryingOptionsShouldMatchLegacyBehaviour() {
+        ClientOptions clientOptions = ClientOptions.fromRemotingParameters(AuxiliaryRemotingParameters.builder()
```
felixdesouza: this test is kinda weird, there's no notion of default `AuxiliaryRemotingParameters`

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/AuxiliaryRemotingParameters.java
```diff
@@ -37,6 +37,17 @@
...
     boolean shouldRetry();
 
+    /**
+     * Whether this client should support operations that block on the server side.
+     * In practice, this means that read and idle connection timeouts may be longer.
+     *
+     * This is set to true by default, to support legacy behaviour.
+     */
+    @Value.Default
+    default boolean shouldSupportBlockingOperations() {
```
felixdesouza: I suppose for the lock service, it's only the lock endpoint that you want to be able to block no?

You could create an impl of the lock service which uses the "supports blocking operations" for the lock method, and then the standard for everything else.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -129,4 +124,27 @@ public static ClientOptions of(Duration connect, Duration read, Duration backoff
...
+        ImmutableClientOptions.Builder builder = ImmutableClientOptions.builder();
+
+        setupTimeouts(builder, parameters);
+        setupRetrying(builder, parameters);
+
+        return builder.clientQoS(ClientConfiguration.ClientQoS.DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS)
+                .build();
+    }
+
+    private static void setupTimeouts(ImmutableClientOptions.Builder builder, AuxiliaryRemotingParameters parameters) {
```
felixdesouza: some docs would be good, to describe what the effect each combination will have, so we don't have to do the same exercise we did today, or even if we do, it's a lot easier.

either here or the constants above.

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/AuxiliaryRemotingParameters.java
felixdesouza: oh, you did this in #4520 💃

atlasdb-conjure/src/test/java/com/palantir/atlasdb/http/v2/ClientOptionsTest.java
jeremyk-91: Sure, yep. I changed this to `minimallySpecified...` as that's what I meant here

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
jeremyk-91: Added a bit of explanation - I'm not sure exactly what you want, might be worth catching up if there are more specific things you'd like to be documented. Thanks!

felixdesouza: This'll do, I suppose I wanted a way to codify our investigation, something along the lines of, anything below a proxy probably shouldn't retry etc. What the current setup is, if there are separate timelock to timelock configurations vs client to timelock etc.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
felixdesouza: haha techincally it's not "odd" 😛

jeremyk-91:
Great, thanks!

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
jeremyk-91: The odd even number was intentional 😅

PR #4520, opened 2020-01-16
Shorter Nonblocking Calls, Part 3: Breaking the TimeLock Logjam

jeremyk-91:
**Goals (and why)**:
- Reduce read timeouts for the majority of calls. See #4416, though this accounts for the endpoints that are legitimately allowed to take closer to 65 seconds.

- Allow ServiceCreators to create whitelisted things as having the lower read timeout.
- Implement BlockingSensitiveTimelockRpcClient, which delegates between two TimelockRpcClients to the same cluster - one for the time-sensitive endpoints (lock, waitForLocks) and one for everything else.
- For (V1) LockService I just made everything use the old timeouts.

Added a test for the delegating timelock rpc client.

_Most of these concerns are around using two separate proxies that operate disjoint sets of endpoints._
- Metrics: will these work correctly? I believe so: they are backed by the same metrics registry, and furthermore the metrics that each of these report will be disjoint.
- ExperimentRunningProxy / Fallback Mechanism for legacy users: This may behave a bit funkily, but if it's bad both services will fallback, and we aren't monitoring the performance numbers for anyone right now.
- HTTP client internals may behave weirdly e.g. sticking to the leader found via 308s: This is a concern, but it's not obvious how we can configure these timeouts at the request level - and if so, we will likely need to have two proxies anyway.

**Where should we start reviewing?**: ServiceCreators or TransactionManagers

**Priority (whenever / two weeks / yesterday)**: early next week? Until this merges #4519 has no value for a majority of our consumers.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/BlockingSensitiveTimelockRpcClient.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/BlockingSensitiveTimelockRpcClient.java
@@ -1,0 +1,128 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


jeremyk-91:
Great, thanks @felixdesouza  for watching out for that. I'll finish this up today and merge.

PR #4521, opened 2020-01-17
[WIP] Extract off heap cache interface and logic

OStevan:
**Goals (and why)**:
Extract the logic for caching to a separate class to give us the ability to store arbitrary data. The first goal is to make it easier to work in environments like internal ski product and the second to able to reuse this logic.

- this is just a split refactor `OffHeapTimestampCache`
- `DefaultOffHeapCache` contains the same logic as the previous `OffHeapTimestampCache` however it is now generic

- current tests should cover the changes.

- are the interfaces and decomp correct 
- the diff is a bit strange because of renaming

- comparing previous `OffHeapTimestampCache` with `DefaultOffHeapCache`

would be great today


atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+                offHeapCache.persistentStore.put(
+                        cacheDescriptor.storeNamespace(),
+                        ImmutableMap.copyOf(request));
+
+                cacheDescriptor.currentSize().addAndGet(sizeIncrease);
+            } catch (SafeIllegalArgumentException exception) {
+                // happens when a store is dropped by a concurrent call to clear
+                log.warn("Clear called concurrently, writing failed", exception);
+            } finally {
+                offHeapCache.inFlightRequests.clear();
```
felixdesouza: is this correct?

if you have a request that comes in for key K, whilst the autobatcher is in here, it'll be in the next batch. If you clear, then another request for key K can come in (still in the same batch, since we haven't returned) because it's not an in flight request anymore.

if it doesn't matter, then the in flight request map doesn't matter because autobatcher will do the right thing (you're using the coalescing one).

if it does matter, then you're gonna have to fix this

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+            return KeyedStream.of(request.stream()).collectToMap();
+        }
+    }
+
+    @Value.Immutable
+    @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+    interface CacheDescriptor {
+        AtomicInteger currentSize();
+        StoreNamespace storeNamespace();
```
felixdesouza: are we not changing store namespace? I'm really quite against it.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapCache.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Caches entry pair.
+     *
+     * @param key of the cached entry
+     * @param value associated with a given key
+     */
+    void put(K key, V value);
+
+    /**
+     * Retrieves the value fot yhe given {@code key}.
```
felixdesouza: typo: "for the"

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapCache.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @param key for which we want to get a value
+     * @return optional containing the associated value or empty if the entry is not available
+     */
+    Optional<V> getIfPresent(K key);
```
felixdesouza: nit: get, put, clear in that order

also just `get`? the optional implies whether it's present or not

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/OffHeapCaches.java
```diff
@@ -0,0 +1,49 @@
...
+    static final String OFF_CACHE_NAMESPACE = "default_off_heap_cache";
+
+    private OffHeapCaches() {
+    }
+
+    public static <K, V> OffHeapCache<K, V> create(
+            LogicalPersistentStore<K, V> persistentStore,
+            TaggedMetricRegistry taggedMetricRegistry,
+            LongSupplier maxSize) {
+        StoreNamespace storeNamespace = persistentStore.createNamespace(OFF_CACHE_NAMESPACE);
```
felixdesouza: I still don't understand how the name is supposed to be used inside namespace

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+        if (inFlight != null) {
+            return Optional.of(inFlight);
+        }
+
+        return persistentStore.get(cacheDescriptor.get().storeNamespace(), key);
+    }
+
+    private static <K, V> CacheDescriptor createNamespaceAndConstructCacheProposal(
+            LogicalPersistentStore<K, V> store) {
+        StoreNamespace proposal = store.createNamespace(OFF_CACHE_NAMESPACE);
```
felixdesouza: `OFF_HEAP_CACHE_NAMESPACE`

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final String BATCHER_PURPOSE = "off-heap-cache";
+    private static final MetricName CACHE_HIT = constructCacheMetricName("cacheHit");
+    private static final MetricName CACHE_MISS = constructCacheMetricName("cacheMiss");
+    private static final MetricName CACHE_NUKE = constructCacheMetricName("cacheNuke");
+    private static final MetricName CACHE_SIZE = constructCacheMetricName("cacheSize");
+
+    private final LogicalPersistentStore<K, V> persistentStore;
+    private final LongSupplier maxSize;
+    private final AtomicReference<CacheDescriptor> cacheDescriptor = new AtomicReference<>();
+    private final TaggedMetricRegistry taggedMetricRegistry;
```
felixdesouza: don't think you need a reference to the registry, just have fields which are the corresponding metrics, also means you can't accidentally use two metrics for the same metric name.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+        if (value.isPresent()) {
+            taggedMetricRegistry.meter(CACHE_HIT).mark();
+        } else {
+            taggedMetricRegistry.meter(CACHE_MISS).mark();
+        }
+        return value;
+    }
+
+    private Optional<V> getValue(K key) {
+        V inFlight = inFlightRequests.get(key);
```
felixdesouza: use canonical `Optional` form for this entire method

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+    private static class WriteBatcher <K, V>
+            implements CoalescingRequestFunction<Map.Entry<K, V>, Map.Entry<K, V>> {
+        DefaultOffHeapCache<K, V> offHeapCache;
+
+        WriteBatcher(DefaultOffHeapCache<K, V> offHeapCache) {
+            this.offHeapCache = offHeapCache;
+        }
+
+        @Override
+        public Map<Map.Entry<K, V>, Map.Entry<K, V>> apply(Set<Map.Entry<K, V>> request) {
```
felixdesouza: have the second parameter be `Void`? the result is not actually used

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+            CacheDescriptor cacheDescriptor = offHeapCache.cacheDescriptor.get();
+            if (cacheDescriptor.currentSize().get() >= offHeapCache.maxSize.getAsLong()) {
+                offHeapCache.taggedMetricRegistry.counter(CACHE_NUKE).inc();
+                offHeapCache.clear();
+            }
+            cacheDescriptor = offHeapCache.cacheDescriptor.get();
+            try {
+                Map<K, V> response = offHeapCache.persistentStore.get(
+                        cacheDescriptor.storeNamespace(),
+                        request.stream().map(Map.Entry::getKey).collect(Collectors.toList()));
```
felixdesouza: variable

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+            return;
+        }
+        Futures.getUnchecked(valuePutter.apply(Maps.immutableEntry(key, value)));
+    }
+
+    @Override
+    public Optional<V> getIfPresent(K key) {
+        Optional<V> value = getValue(key);
+
+        if (value.isPresent()) {
```
felixdesouza: undecided but assuming you take into account the above:
```java
Meter cacheMeter = value.isPresent() ? cacheHitMeter : cacheMissMeter;
cacheMeter.mark();
...
```

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
felixdesouza: plus is this safe? there's no guarantee it was written no? is this not assuming the immutability of transactions table?

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+
+        CacheDescriptor previous = cacheDescriptor.getAndUpdate(prev -> proposedCacheDescriptor);
+        if (previous != null) {
+            persistentStore.dropNamespace(previous.storeNamespace());
+        }
+    }
+
+    @Override
+    public void put(K key, V value) {
+        if (inFlightRequests.putIfAbsent(key, value) != null) {
```
felixdesouza: similarly here, if this is general, I've lost updates...

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+            LongSupplier maxSize,
+            TaggedMetricRegistry taggedMetricRegistry) {
+        this.persistentStore = persistentStore;
+        this.cacheDescriptor.set(cacheDescriptor);
+        this.maxSize = maxSize;
+        this.taggedMetricRegistry = taggedMetricRegistry;
+        this.valuePutter = Autobatchers.coalescing(new WriteBatcher<>(this))
+                .safeLoggablePurpose(BATCHER_PURPOSE)
+                .build();
+        Gauge<Integer> cacheSizeGauge = () -> DefaultOffHeapCache.this.cacheDescriptor.get().currentSize().intValue();
```
felixdesouza: do you need to specify `DefaultOffHeapCache` out of curiosity? if no then get rid of it please.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -535,12 +538,23 @@ static TimestampCache timestampCache(
...
                         .orElseGet(() -> new DefaultTimestampCache(metricsManager.getRegistry(), cacheSize));
 
         return config.timestampCache().orElseGet(timestampCacheSupplier);
     }
 
+    private static TimestampCache constructOffHeapTimestampCache(
+                PhysicalPersistentStore persistentStore,
+                TaggedMetricRegistry taggedMetricRegistry,
+                LongSupplier cacheSize) {
+        OffHeapCache<Long, Long> offHeapCache = OffHeapCaches.create(
```
felixdesouza: This should live inside `OffHeapTimestampCache`. Otherwise it's easy to create something an offheap timestamp cache that can pass in different logical stores and what not. The abstraction is slowly falling apart here, but if you must have this, put it in `OffHeapTimestampCache`.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
```diff
@@ -0,0 +1,178 @@
...
+
+final class DefaultOffHeapCache<K, V> implements OffHeapCache<K, V> {
+    private static final Logger log = LoggerFactory.getLogger(DefaultOffHeapCache.class);
+    private static final String BATCHER_PURPOSE = "off-heap-cache";
+    private static final MetricName CACHE_HIT = constructCacheMetricName("cacheHit");
+    private static final MetricName CACHE_MISS = constructCacheMetricName("cacheMiss");
+    private static final MetricName CACHE_NUKE = constructCacheMetricName("cacheNuke");
+    private static final MetricName CACHE_SIZE = constructCacheMetricName("cacheSize");
+
+    private final LogicalPersistentStore<K, V> persistentStore;
```
felixdesouza: as an aside, I don't understand why you need to distinguish between `Logical` and `Physical`, if anything, the latter is an instance of the former where `K` and `V` are both `ByteString`...

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: I removed the `inflight`. As discussed this logic can be implemented for the specific timestamp cache.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: No longer a `StoreNamespace` but a `StoreHandle`. I'm open to a different name.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: No longer relevant.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: Removed.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: This is removed.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: No, removed.

atlasdb-client/src/main/java/com/palantir/atlasdb/cache/DefaultOffHeapCache.java
OStevan: Acked and removed.

OStevan:
Will be split in two PRs therefore closed.

PR #4523, opened 2020-01-20
Remove inaccurate log line

gmaretic:
**Goals (and why)**:
The claim that this log line is making is not true, since it is possible for a reader to roll us back. The case where this happens (that we are aware of) is quite a niche corner case, so we should investigate further/


jeremyk-91:
The edge case is

- T1 passes its lock checks
- T1 loses the immutable timestamp lock but not its other locks
- the immutable timestamp passes the start timestamp of T1
- a read transaction T2 reads a cell T1 wrote, and this read doesn't wait for the transaction lock because the immutable timestamp has passed it
- T2 rollbacks T1
- T1 logs the error

Are we sure we don't want to just downgrade this rather than remove it entirely?

PR #4529, opened 2020-01-22
Reset the Cassandra cluster after runs of Single/DoubleNodeDown ETE Test

leonz:
**Goals (and why)**:
Starting in [Cassandra 3.6](https://issues.apache.org/jira/browse/CASSANDRA-10134), Cassandra detects and fails to start if a different node previously existed at the same IP address.

The Multi Cassandra tests kill the docker container and then start it up again with the same IP, which is not valid since the node is technically no longer the same node (killing the docker container wipes the data).

At the end of each Multi Cassandra test, reset the state of the Cassandra cluster by killing _all_ of the nodes and starting them back up again.

I tested this fix on top of the Cassandra 3 development branch to prove that it worked in both Cassandra 2 and 3.  This is just a test fix, so no impact to production systems.

### DIFF
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/MultiCassandraUtils.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/MultiCassandraUtils.java
@@ -31,7 +31,7 @@ public final class MultiCassandraUtils {
         // utility
     }
 
-    /** Kill all of the Cassandra containers and start them again.  This wipes all data on the nodes. */
+    /* Kill all of the Cassandra containers and start them again.  This wipes all data on the nodes. */
     public static void resetCassandraCluster(Set<String> containerNames) {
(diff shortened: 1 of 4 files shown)


felixdesouza:
Not on my laptop, how much slower are builds?

leonz:
Long test now takes about 5:30 compared to 4:48 on develop.  Still not the slowest container though so not on the critical path.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/MultiCassandraUtils.java
```diff
@@ -30,6 +31,12 @@ private MultiCassandraUtils() {
         // utility
     }
 
+    /** Kill all of the Cassandra containers and start them again.  This wipes all data on the nodes. */
```
jeremyk-91: nit: I think JavaDoc usually is

```
/**
 * bla
 */
```

otherwise normally we'd just use `/*` rather than `/**` to open.

PR #4531, opened 2020-01-23
Introduce NanoClock

j-baker:
NanoClock represents a safety mechanism around NanoTime - it throws if
we ever spot NanoTime.now() go down (which we should never see, but
would lead to weird locking behaviour).

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
@@ -1,0 +1,46 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.common.time;
+
+import static com.google.common.base.Preconditions.checkState;
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+public enum NanoClock implements Supplier<NanoTime> {
+    INSTANCE;
+
+    private final AtomicReference<NanoTime> maxSeen = new AtomicReference<>();
+
+    @Override
+    public NanoTime get() {
+        NanoTime currentMax = maxSeen.get();
+        NanoTime now = NanoTime.now();
+        checkState(!now.isBefore(currentMax), "Clock reversal detected");
+        maybeSetNewMax(now);
+        return now;
+    }
+
+    private void maybeSetNewMax(NanoTime time) {
+        while (true) {
+            NanoTime current = maxSeen.get();
+            if (time.isBefore(current) || maxSeen.compareAndSet(current, time)) {
+                return;
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static final AtomicReferenceFieldUpdater<NanoClock, NanoTime> updater =
+            AtomicReferenceFieldUpdater.newUpdater(NanoClock.class, NanoTime.class, "maxSeen");
+    private volatile NanoTime maxSeen;
+
+    @Override
+    public NanoTime get() {
+        NanoTime currentMax = maxSeen;
+        NanoTime now = NanoTime.now();
+        checkState(currentMax == null || !now.isBefore(currentMax), "Clock reversal detected");
```
schlosna: would you want to include the previous/current if this fails to have a sense of magnitude of reversal?

felixdesouza:
Are you able to go into a bit more detail in the description as to why this change needs to be made? What prompted it etc. What are we trying to prevent in the slightly grander scheme of things.

jeremyk-91:
Probably want a changelog (and maybe a short description as @felixdesouza mentions)

atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.common.time;
+
+import static com.google.common.base.Preconditions.checkState;
```
felixdesouza:
```suggestion
import static com.palantir.logsafe.Preconditions.checkState;
```

atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
felixdesouza:
```suggestion
        checkState(currentMax == null || !now.isBefore(currentMax), "Clock reversal detected",
            SafeArg.of("currentMax", currentMax.time()),
            SafeArg.of("now", now.time()));
```

atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
```diff
@@ -0,0 +1,48 @@
...
+    public NanoTime get() {
+        NanoTime currentMax = maxSeen;
+        NanoTime now = NanoTime.now();
+        checkState(currentMax == null || !now.isBefore(currentMax), "Clock reversal detected");
+        maybeSetNewMax(now);
+        return now;
+    }
+
+    private void maybeSetNewMax(NanoTime time) {
+        while (true) {
```
felixdesouza: could you not just do a `AtomicReferenceFieldUpdater#getAndUpdate`?

felixdesouza: or are you trying to avoid going into the CAS if the predicate holds?

atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package com.palantir.common.time;
+
+import static com.google.common.base.Preconditions.checkState;
+
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Supplier;
+
+public enum NanoClock implements Supplier<NanoTime> {
```
felixdesouza: does this necessarily need to be an enum? I know the language allows fields to be mutable, but feels a bit weird. Is it not sufficient to just create a NanoClock as a normal class/object and pass that into `LeaderClock`? you'll still have one per namespace, or if we really want it to span everything, just inject it into `AsyncTimeLockServicesCreator`?

atlasdb-commons/src/main/java/com/palantir/common/time/NanoClock.java
felixdesouza: Unfortunately I can't seem to figure out off the top of my head a nice way to get an Instant from nano's without going through DateTime? Unclear.

PR #4532, opened 2020-01-23
Rewrite disruptorautobatcher to just use a plain blocking queue

j-baker:
We've seen bugs internally caused by the autobatcher. In particular, the
code here:
https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java#L136
is causing a high context switch rate. Since I mostly used this class
because it seemed cute, and not because it seemed necessary, it seems
reasonable to rewrite it. This means that a few of the other classes
also become simpler.

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -17,7 +17,13 @@
 package com.palantir.atlasdb.autobatch;
 
 import java.io.Closeable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
@@ -30,11 +36,7 @@ import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import com.lmax.disruptor.EventHandler;
-import com.lmax.disruptor.RingBuffer;
-import com.lmax.disruptor.TimeoutException;
-import com.lmax.disruptor.dsl.Disruptor;
-import com.palantir.logsafe.Preconditions;
+import com.palantir.common.concurrent.PTExecutors;
 
 /**
  * While this class is public, it shouldn't be used as API outside of AtlasDB because we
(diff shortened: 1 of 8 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -63,42 +65,48 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
-            disruptor.shutdown(10, TimeUnit.SECONDS);
-        } catch (TimeoutException e) {
-            log.warn("Disruptor took more than 10 seconds to shutdown. "
-                    + "Ensure that handlers aren't uninterruptibly blocking and ensure that they are closed.", e);
+            executorService.shutdownNow();
+            if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
+                log.warn("Disruptor took more than 10 seconds to shutdown. "
+                        + "Ensure that handlers aren't uninterruptibly blocking and ensure that they are closed.");
+            }
+            executorService.awaitTermination(10, TimeUnit.SECONDS);
```
jeremyk-91: I think you intended to delete this line - as written we wait 20 seconds

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -63,42 +65,48 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
                 .build();
     }
 
-    private final Disruptor<DefaultBatchElement<T, R>> disruptor;
```
jeremyk-91: I thought their claim was that they put up some really good perf numbers in their benchmarking, though I read the ticket and the code you linked, so I'm convinced it's worth trying this out.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -63,42 +65,48 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
+    private final BlockingQueue<DefaultBatchElement<T, R>> queue;
+    private final ExecutorService executorService;
     private volatile boolean closed = false;
 
-    DisruptorAutobatcher(
-            Disruptor<DefaultBatchElement<T, R>> disruptor,
-            RingBuffer<DefaultBatchElement<T, R>> buffer) {
-        this.disruptor = disruptor;
-        this.buffer = buffer;
+    DisruptorAutobatcher(BlockingQueue<DefaultBatchElement<T, R>> queue, ExecutorService executorService) {
```
jeremyk-91: nit: technically misleading now that we don't actually use a Disruptor

felixdesouza:
For posterity, what were these bugs? Are there links to said investigations. Is there a way to measure things actually being better? Are there any knobs in Disruptor that we can modify before rewriting? 

Can we put this behind a flag, and try it out in a few products first so that if it's worse we're not chasing down products to take an upgrade, especially if it means increased load on timelock etc.

I suppose it's just unclear to me why this is immediately better/worse without just taking your word for it.

felixdesouza:
closing in favour of already merged #4535. Can always revisit if necessary.

PR #4534, opened 2020-01-23
Prefixed path for persistent storage

OStevan:
**Goals (and why)**:
Use a magic fold to prefix all folders which will be used to store persistent storage data. The goal of this PR is to make the usage safer and to remove the static mutable state.

- using a magic keyword to prefix folders

- tests are updated

tomorrow


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -24,22 +24,30 @@
...
 import com.palantir.logsafe.SafeArg;
 
-public final class PersistentStorageFactories {
-    private static final Pattern UUID_PATTERN = Pattern.compile(
-            "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
+public final class PersistentStoragePathSanitizer {
+    private static final Logger log = LoggerFactory.getLogger(PersistentStoragePathSanitizer.class);
+    private static final PersistentStoragePathSanitizer INSTANCE = new PersistentStoragePathSanitizer();
+    @VisibleForTesting
+    static final String MAGIC_SUFFIX = "atlasdb-persistent-storage";
```
gmaretic: I'd shove this into `AtlasdbConstants`

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -24,22 +24,30 @@
...
-    private static final Set<String> SANITIZED_PATHS = new HashSet<>();
+    public static PersistentStoragePathSanitizer create() {
+        return INSTANCE;
+    }
 
-    private PersistentStorageFactories() {}
+    private final Set<String> sanitizedPaths = new HashSet<>();
 
+    @VisibleForTesting
+    PersistentStoragePathSanitizer() {}
```
gmaretic: Constructor should be the first method generally, also making it visible for testing is super sketch. We also don't seem to really need a singleton, but if you want to keep it singleton, why don't you make each test use a different storagePath instead?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -24,22 +24,30 @@
...
-    private static final Pattern UUID_PATTERN = Pattern.compile(
-            "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
+public final class PersistentStoragePathSanitizer {
+    private static final Logger log = LoggerFactory.getLogger(PersistentStoragePathSanitizer.class);
+    private static final PersistentStoragePathSanitizer INSTANCE = new PersistentStoragePathSanitizer();
+    @VisibleForTesting
+    static final String MAGIC_SUFFIX = "atlasdb-persistent-storage";
 
-    private static final Set<String> SANITIZED_PATHS = new HashSet<>();
+    public static PersistentStoragePathSanitizer create() {
```
gmaretic: Create is a bit misleading, maybe `getInstance`?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -24,22 +24,30 @@
...
+import com.google.common.annotations.VisibleForTesting;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 
-public final class PersistentStorageFactories {
-    private static final Pattern UUID_PATTERN = Pattern.compile(
-            "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$");
+public final class PersistentStoragePathSanitizer {
+    private static final Logger log = LoggerFactory.getLogger(PersistentStoragePathSanitizer.class);
+    private static final PersistentStoragePathSanitizer INSTANCE = new PersistentStoragePathSanitizer();
```
felixdesouza: as discussed offline, let's just do the straightforward thing since this will be called at most once, for the places where multiple transaction managers are used without proper sync, we'd need to sort it out ourselves, but that's easy since there are not that many.

felixdesouza: this can just be a static method

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/DefaultPersistentStorageFactory.java
```diff
@@ -34,6 +35,11 @@
  */
 public final class DefaultPersistentStorageFactory implements PersistentStorageFactory {
     private static final Logger log = LoggerFactory.getLogger(DefaultPersistentStorageFactory.class);
+    private final PersistentStoragePathSanitizer persistentStoragePathSanitizer;
+
+    public DefaultPersistentStorageFactory(PersistentStoragePathSanitizer persistentStoragePathSanitizer) {
```
felixdesouza: given this is a singleton, why are we passing this in?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
OStevan: Removed now.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
OStevan: in @felixdesouza we believe, so still a magic but now it is private

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -21,70 +21,63 @@
...
      * @param storagePath to the proposed storage location
      */
-    public static synchronized void sanitizeStoragePath(String storagePath) {
-        if (SANITIZED_PATHS.contains(storagePath)) {
-            return;
-        }
+    public static Path sanitizeStoragePath(String storagePath) {
+        File storageDirectory = new File(storagePath, MAGIC_SUFFIX);
+
+        Preconditions.checkArgument(
```
felixdesouza: what happens if `storagePath` doesn't exist, does this catch it?

OStevan: It is caught.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizerTests.java
```diff
@@ -0,0 +1,78 @@
...
+        File file = testFolder.newFile();
+
+        assertThatThrownBy(() -> PersistentStoragePathSanitizer.sanitizeStoragePath(file.getAbsolutePath()))
+                .isInstanceOf(SafeIllegalArgumentException.class)
+                .hasMessageContaining("has to point to a directory");
+    }
+
+    @Test
+    public void removesMagicFolder() {
+        PersistentStoragePathSanitizer.sanitizeStoragePath(testFolderPath).toFile().mkdir();
```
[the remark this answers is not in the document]
OStevan: Added.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/PersistentStoragePathSanitizer.java
```diff
@@ -21,70 +21,63 @@
...
-            SANITIZED_PATHS.add(storagePath);
-            return;
+            return storageDirectory.toPath().toAbsolutePath();
         }
 
-        Preconditions.checkArgument(
+        Preconditions.checkState(
                 storageDirectory.isDirectory(),
-                "Storage path has to point to a directory",
+                "Persistent storage path magic atlas subfolder path points to a directory",
```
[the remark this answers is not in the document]
OStevan: Fixed.

PR #4539, opened 2020-01-27
Bump tests to use Cassandra 3

leonz:
**Goals (and why)**:
Start testing against Cassandra 3!  Marking as "do not merge" until we have a concrete timeline for using Cassandra 3.

We're no longer testing against Cassandra 2.  Though I don't think the implementation is significantly different enough that we should need this kind of coverage for both.

Whenever.


gmaretic:
@leonz is it the case that multiget multislice is broken on C3? If yes, we should add tests that actually catch this

leonz:
MGMS bug was tracked down to an upstream issue and fixed with the version bump.

My theory on why the most recent commit (bump to PT Cass 3.11.6) is failing with `sstablemetadata` not found is an upstream change in the Docker file which no longer installs the supplementary tools to `/usr/bin`: https://github.com/docker-library/cassandra/commit/de006e2143c7685102fcc790aec32ebdf1cf7789

Fixing this in Palantir Cassandra docker file, will push up a commit shortly.

Jolyon-S:
I'm going to close this since this is 10 months old, but do feel free to reopen if this becomes a reality.

PR #4540, opened 2020-01-27
Coalescing Suppliers: Explicit Round Ordering and Max Accumulation

jeremyk-91:
**Goals (and why)**:
- Fix what I think is an issue with our `CoalescingSupplier` implementation. If I'm not mistaken, multiple executions of `execute()` on different rounds _can_ concurrently happen, and there is a particularly nasty way in which they can interact. 

coalescing-supplier

- Basically, future completion and `round` getting updated are non-atomic, so a bad GC at the wrong time can cause us to be stuck in an old round, if two new rounds happened during the GC. Thanks to @gmaretic and @felixdesouza who helped a lot in isolating this.
- See PDS-109071 for more details: this explains our metrics woes, among other concerns.

- Add a numeric counter to rounds. The idea is to achieve an ordering.
- Have the next round be the _highest claimed next round_ in terms of sequence, and not just a volatile assignment.

- None, other than that the existing tests still pass. Happy to take suggestions/ideas on how to test this better.

- Am I mistaken in thinking the interaction is bad?
- Is it more prudent to just revert to the old implementation?
- Are there other categories of issues that tracking rounds won't fix?

**Where should we start reviewing?**: `CoalescingSupplier.java`

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
@@ -82,8 +82,9 @@ public class CoalescingSupplier<T> implements Supplier<T> {
                 future.complete(delegate.get());
             } catch (Throwable t) {
                 future.completeExceptionally(t);
+            } finally {
+                round.accumulateAndGet(next, (r1, r2) -> r1.roundNumber > r2.roundNumber ? r1 : r2);
             }
-            round.accumulateAndGet(next, (r1, r2) -> r1.roundNumber > r2.roundNumber ? r1 : r2);
         }
 
         T getResult() {
--- a/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
@@ -15,9 +15,11 @@
  */
 package com.palantir.common.concurrent;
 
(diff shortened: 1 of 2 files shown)


felixdesouza:
I think we should revert the commit before we do anything else, let it propagate. If we're going to spend time on this, we should also make time for it instead of having to react to it, it's consumed a lot of dev hours. 

What should follow is a more careful rollout plan since this affects timelock, consumers of atlas, and also big internal product via both atlas consumption and leader election library consumption. That is, potentially a gauge on the current round, some checks to make sure things don't go back in "rounds".

j-baker:
Apologies for this. I think I originally put the volatile set _before_ completing the future and then realised that didn’t work for different reasons (concurrent computations arose).

Felix, I don’t think we should ever test for concurrency correctness with a gauge - either this thing is reliable and we should not care about it ever, or it is unreliable and we should not use it - like there’s no gap in the middle for ‘correctness is broken and we debugged it via a metric’, we have to be able to rely on this shit. Likewise this isn’t a place for careful rollout to be the solution - this is way too binary.

Jeremy - two other solutions that might be simpler:

instead of the accumulate you might want a compareAndSwap(this, next)? You’re really trying to do a cas, though your rounds concept is a convenient mental model.

Or, you could have the compute function return a new completable future with the result in, and then complete the future with that _after_ you set next to nextRound.

I think reverting the original PR and recommitting with these changes makes sense - no need to have a bad develop.

jeremyk-91:
closing out in favour of #4541

PR #4541, opened 2020-01-28
Fix coalescing supplier

j-baker:
Let's just take out a lock - and write a test that covers it. It's more
obviously correct.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
@@ -29,6 +29,8 @@ import com.google.common.base.Throwables;
  */
 public class CoalescingSupplier<T> implements Supplier<T> {
     private final Supplier<T> delegate;
+
+    private final Object lock = new Object();
     private volatile Round round = new Round();
 
     public CoalescingSupplier(Supplier<T> delegate) {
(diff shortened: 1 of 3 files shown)


atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
```diff
@@ -69,13 +71,15 @@ Round awaitDone() {
         }
 
         void execute() {
-            next = new Round();
-            try {
-                future.complete(delegate.get());
-            } catch (Throwable t) {
-                future.completeExceptionally(t);
+            synchronized (lock) {
```
j-baker: this is an uncontended lock

atlasdb-commons/src/test/java/com/palantir/common/concurrent/CoalescingSupplierTest.java
```diff
@@ -114,6 +121,41 @@ public void exceptionsArePropagatedForCoalescedCalls() {
...
+        int poolSize = 1024;
+        ListeningExecutorService executorService =
+                MoreExecutors.listeningDecorator(PTExecutors.newFixedThreadPool(poolSize));
+        AtomicLong counter = new AtomicLong(0);
+        Supplier<Long> supplier = new CoalescingSupplier<>(() -> {
+            sleep(2);
+            return counter.incrementAndGet();
+        });
+        List<ListenableFuture<?>> futures = IntStream.range(0, poolSize)
+                .mapToObj(index -> executorService.submit(() -> assertIncreasing(supplier)))
```
gmaretic: This is neat!

PR #4543, opened 2020-01-30
Better log when rolled back by someone else

gmaretic:
**Goals (and why)**:
We still want to log something when this occurs, but we don't want to panic users by logging at warn or saying that this cannot happen in atlas normally (as it can)

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -140,8 +140,6 @@ import com.palantir.common.streams.MoreStreams;
 import com.palantir.lock.AtlasCellLockDescriptor;
 import com.palantir.lock.AtlasRowLockDescriptor;
 import com.palantir.lock.LockDescriptor;
-import com.palantir.lock.client.LeasedLockToken;
-import com.palantir.lock.client.LockTokenShare;
 import com.palantir.lock.v2.ImmutableLockRequest;
 import com.palantir.lock.v2.LockRequest;
 import com.palantir.lock.v2.LockResponse;
@@ -2232,15 +2230,6 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             throws TransactionFailedException {
         Preconditions.checkArgument(commitTimestamp > getStartTimestamp(), "commitTs must be greater than startTs");
         try {
-            if (immutableTimestampLock.isPresent()) {
-                log.error("This transaction has been rolled back by someone else.",
-                        immutableTimestampLock.get().toSafeArg(("immutableTimestampLock")),
-                        locksToken.toSafeArg("commitLocksToken"));
-            } else {
-                log.error("This transaction has been rolled back by someone else.",
-                        SafeArg.of("immutableTimestampLock", "not present"),
-                        locksToken.toSafeArg("commitLocksToken"));
-            }
             transactionService.putUnlessExists(getStartTimestamp(), commitTimestamp);
         } catch (KeyAlreadyExistsException e) {
             handleKeyAlreadyExistsException(commitTimestamp, e, locksToken);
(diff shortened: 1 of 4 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2260,8 +2260,7 @@ private void handleKeyAlreadyExistsException(
                         + " because our locks timed out. startTs: " + getStartTimestamp() + ".  "
                         + getExpiredLocksErrorString(commitLocksToken, expiredLocks), ex);
             } else {
-                log.debug("Possible bug: Someone rolled back our transaction but"
-                        + " our locks were still valid; Atlas code is not allowed to do this, though others can.",
+                log.info("This transaction has been rolled back by someone else.",
                         SafeArg.of("immutableTimestampLock", immutableTimestampLock),
```
jkozlowski: This will break logging again: LeasedLockToken is not loggable by our logging infra. So either get rid of the args in this logline or figure out how to collect the stuff you need in a way that's loggable.

jkozlowski: For this to be useful, probably startTimestamp would be pretty useful as an arg?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2260,8 +2260,7 @@ private void handleKeyAlreadyExistsException(
                         + " because our locks timed out. startTs: " + getStartTimestamp() + ".  "
                         + getExpiredLocksErrorString(commitLocksToken, expiredLocks), ex);
             } else {
-                log.debug("Possible bug: Someone rolled back our transaction but"
-                        + " our locks were still valid; Atlas code is not allowed to do this, though others can.",
+                log.info("This transaction has been rolled back by someone else.",
```
jeremyk-91: maybe we can say that this should not happen frequently?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2260,10 +2269,10 @@ private void handleKeyAlreadyExistsException(
                         + " because our locks timed out. startTs: " + getStartTimestamp() + ".  "
                         + getExpiredLocksErrorString(commitLocksToken, expiredLocks), ex);
             } else {
-                log.debug("Possible bug: Someone rolled back our transaction but"
-                        + " our locks were still valid; Atlas code is not allowed to do this, though others can.",
-                        SafeArg.of("immutableTimestampLock", immutableTimestampLock),
-                        SafeArg.of("commitLocksToken", commitLocksToken));
+                log.info("This transaction has been rolled back by someone else.",
+                        SafeArg.of("immutableTimestampLock",
```
jkozlowski: @gmaretic would you be able to quickly test manually using an internal product to see that this logline works?

gmaretic:
@jkozlowski tested out locally by butchering some ete tetsts

jkozlowski:
Is should've said that: this is rather easy to test in internal product, I've done that in the past. Ensures you have the right classpath and webservers and whatnot.

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -24,7 +24,7 @@
 import com.palantir.lock.v2.LockToken;
 import com.palantir.logsafe.Preconditions;
 
-final class LockTokenShare implements LockToken {
+public final class LockTokenShare implements LockToken {
```
jkozlowski: At this point maybe we should just add something to the LockToken interface that is going to return a String that can be used for logging? Or make it extend https://github.com/palantir/safe-logging/blob/develop/safe-logging/src/main/java/com/palantir/logsafe/SafeLoggable.java?

gmaretic:
@jkozlowski is this what you had in mind? Anyway, tested on internal product and works fine

lock-api/src/main/java/com/palantir/lock/client/LockTokenShare.java
```diff
@@ -23,8 +23,9 @@
 
 import com.palantir.lock.v2.LockToken;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 
-final class LockTokenShare implements LockToken {
+public final class LockTokenShare implements LockToken {
```
jkozlowski: These publics are not necessary anymore?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2260,10 +2260,11 @@ private void handleKeyAlreadyExistsException(
                         + " because our locks timed out. startTs: " + getStartTimestamp() + ".  "
                         + getExpiredLocksErrorString(commitLocksToken, expiredLocks), ex);
             } else {
-                log.debug("Possible bug: Someone rolled back our transaction but"
-                        + " our locks were still valid; Atlas code is not allowed to do this, though others can.",
-                        SafeArg.of("immutableTimestampLock", immutableTimestampLock),
-                        SafeArg.of("commitLocksToken", commitLocksToken));
+                log.info("This transaction has been rolled back by someone else.",
```
jeremyk-91: We should say we still have our locks, otherwise it's not clear what the args are meant to mean.

gmaretic:
Fair enough, improved the message. Left leased lock token public as we need it for lock watches anyway

PR #4544, opened 2020-01-30
Docs for conservative to thorough + python3 changes.

felixdesouza:
**Goals (and why)**:
Docs for migrating from conservative to thorough + some background as to why things can be bad. This is here for posterity instead of lurking in tickets.

* Updated to python3 (hopefully).

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -56,9 +56,9 @@ source_suffix = ['.rst', '.md']
 master_doc = 'index'
 
 # General information about the project.
-project = u'OSS AtlasDB'
-copyright = u'2017, Palantir Technologies'
-author = u'Palantir Technologies'
+project = 'OSS AtlasDB'
+copyright = '2017, Palantir Technologies'
+author = 'Palantir Technologies'
 
 def get_last_release_from_gradle():
   """ The version that was last released before this commit.
@@ -266,8 +266,8 @@ latex_elements = {
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-  (master_doc, 'AtlasDB.tex', u'AtlasDB Documentation',
-   u'Palantir Technologies', 'manual'),
+  (master_doc, 'AtlasDB.tex', 'AtlasDB Documentation',
+   'Palantir Technologies', 'manual'),
 ]
 
 # The name of an image file (relative to this directory) to place at the top of
@@ -296,7 +296,7 @@ latex_documents = [
 # One entry per manual page. List of tuples
 # (source start file, name, description, authors, manual section).
 man_pages = [
-    (master_doc, 'atlasdb', u'AtlasDB Documentation',
+    (master_doc, 'atlasdb', 'AtlasDB Documentation',
      [author], 1)
 ]
 
@@ -310,7 +310,7 @@ man_pages = [
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-  (master_doc, 'AtlasDB', u'AtlasDB Documentation',
+  (master_doc, 'AtlasDB', 'AtlasDB Documentation',
    author, 'AtlasDB', 'One line description of project.',
    'Miscellaneous'),
 ]
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -37,7 +37,7 @@ extensions = [
     'code-block-with-version-replacement'
 ]
 
-# Included 
+# Included
 source_parsers = {
   '.md': CommonMarkParser,
 }
@@ -84,7 +84,7 @@ def get_last_release_from_gradle():
   codebase_root = os.path.join(conf_py_directory, '..', '..')
   p = subprocess.Popen(['./gradlew', 'printLastVersion'],
                        cwd=codebase_root, stdout=subprocess.PIPE)
-  stdout = p.communicate()[0]
+  stdout = p.communicate()[0].decode("utf-8")
   matcher = re.search('\d+\.\d+\.\d+[\.\-\+\w]*', stdout)
   return matcher.group(0)
 
--- a/docs/source/troubleshooting/index.rst
+++ b/docs/source/troubleshooting/index.rst
@@ -183,3 +183,68 @@ compare-and-set operation here.
          True
 
 Clients should be able to take the backup lock again after this step.
+
+Migrating from conservatively swept tables to thoroughly swept tables
+=====================================================================
+
+Background and Motivation
+-------------------------
+
+Long running read transactions need to be able to distinguish between cells that were never written to, and cells that
+have been overwritten and are now being cleaned by sweep. A solution to this is to leave a sentinel that will always be
+read no matter what timestamp the transaction is running at when deleting a cell along with the normal Atlas-level
+delete (empty byte-array). This is how tables running with conservative sweep work. The downside is that for data that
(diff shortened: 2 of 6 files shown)


docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+
+Background and Motivation
+-------------------------
+
+Long running read transactions need to be able to distinguish between cells that were never written to, and cells that
+have been overwritten and are now being cleaned by sweep. A solution to this is to leave a sentinel that will always be
+read no matter what timestamp the transaction is running at when deleting a cell along with the normal Atlas-level
+delete (empty byte-array). This is how tables running with conservative sweep work. The downside is that for data that
+is transient - i.e. gets written to and deleted fairly quickly - this metadata stays behind forever, and even if the
+table is empty from an Atlas level, it will still have all of this metadata in the underlying key value service,
```
Jolyon-S: This sentence is very long and doesn't flow very well

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+resulting in unbounded disk usage growth the more cells are written to and deleted from. It also means that more data
+is potentially loaded from the underlying KVS when doing range scans making queries worse as time goes on.
```
Jolyon-S: `is potentially loaded from the underlying KVS when doing range scans, making queries worse as time goes on.` (note the comma)

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+
+An alternative solution is to take out locks on the immutable timestamp, such that sweep cannot progress further than
+any in progress transaction. This means we needn't write any of the sentinels for long running read transactions to
```
Jolyon-S: `in-progress transaction` (hyphenate)

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+find. This is how sweep interacts with tables that are marked with the "thorough" sweep strategy.
+
+Unless explicitly specified, tables will have the "conservative" sweep strategy, and as a result will accumulate a lot
+of cruft especially for the high traffic tables. Migrating from conservative to thorough in a HA setup is possible but
```
Jolyon-S: `of cruft, especially for high traffic tables.`

Jolyon-S: and I believe it is correct to use `an HA setup` instead of `a HA setup`

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+not trivial as it is possible to run into data integrity concerns. If one node is writing sentinels, and the other is
+just taking out locks and *not* writing sentinels, a long running read transaction on the first node cannot tell the
+difference between an overwritten/deleted cell that occurred in a transaction on the new node was deleted or a cell
```
Jolyon-S: The last part of this sentence is a bit unclear

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+delete (empty byte-array). This is how tables running with conservative sweep work. The downside is that for data that
+is transient - i.e. gets written to and deleted fairly quickly - this metadata stays behind forever, and even if the
+table is empty from an Atlas level, it will still have all of this metadata in the underlying key value service,
+resulting in unbounded disk usage growth the more cells are written to and deleted from. It also means that more data
+is potentially loaded from the underlying KVS when doing range scans making queries worse as time goes on.
+
+An alternative solution is to take out locks on the immutable timestamp, such that sweep cannot progress further than
+any in progress transaction. This means we needn't write any of the sentinels for long running read transactions to
+find. This is how sweep interacts with tables that are marked with the "thorough" sweep strategy.
+
```
jeremyk-91: In general for this section: I think this should not be described as an alternative approach, as the immutable timestamp is actually checked in both cases. 

Conservative Sweep _additionally_ wants to support read transactions, and if we do this we need to make sure sweep does not get rid of values that may be read for long running read transactions, hence the need for the unreadable timestamp (which you describe in the first paragraph). Actually conservative sweep uses for its sweep timestamp the minimum of the immutable and unreadable timestamps.

docs/source/troubleshooting/index.rst
jeremyk-91: nit: I'd clarify that the range scans here include both row and column range scans.

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+   ``SweepStrategy.THOROUGH``.
+
+The combinations of versions that can run concurrently i.e. during a rolling upgrade are restricted and are as follows:
+
+* ``A`` and ``B``
+  * Sentinels are still being written, at the same time the immutable timestamp lock is being taken out on ``B``. This
+    adds a bit of overhead and is somewhat redundant since the sentinels are still being written, but it is an essential
+    step in the migration.
+* ``B`` and ``C``
+  * Sentinels are still being written but by *``B``* only. The immutable timestamp lock is being taken out on both
```
jeremyk-91: Hmm, I think you intend italicised monospace here, but it doesn't look like RST renders this as such.

docs/source/conf.py
```diff
@@ -56,9 +56,9 @@
...
 
 # General information about the project.
-project = u'OSS AtlasDB'
-copyright = u'2017, Palantir Technologies'
-author = u'Palantir Technologies'
+project = 'OSS AtlasDB'
+copyright = '2017, Palantir Technologies'
+author = 'Palantir Technologies'
 
 def get_last_release_from_gradle():
```
jeremyk-91: `subprocess.Popen.communicate` is changed in Python 3 to return byte strings, so we'll need to decode the output before applying the regex matcher against it.

docs/source/troubleshooting/index.rst
```diff
@@ -183,3 +183,68 @@ compare-and-set operation here.
...
+    nodes. Should ``C`` delete a cell that ``B`` is currently writing to, sweep won't clear it underneath ``B`` since
+    the immutable timestamp lock is still being held.
+
+``A`` and ``C`` are illegal combinations as ``A`` is not taking out the immutable timestamp lock, so sweep can continue
+unhindered. ``C`` is also not writing sentinels, so when sweep deletes the cell from the KVS, read transactions on
+``A`` will treat the cell as empty as opposed to deleted.
+
+.. warning::
+
+    Once you've migrated, no __**new**__ sentinels will be written to the KVS, however any existing sentinels will be
```
jeremyk-91: Doesn't seem like `__**bla**__` renders as bold italic (which I assume is the intention)

docs/source/conf.py
jeremyk-91: The `stdout = p.communicate()[0]` in this function needs to be something like `stdout = p.communicate()[0].decode("utf-8")` (maybe check if the encoding is necessarily utf-8 first)

felixdesouza: 💃 thanks!

felixdesouza:
thorough to conservative is sketchy, conservative to thorough is okay, there are more steps to ensure that this is okay. It's just not worth it, migrating to a new table is certainly the best way here. @gmaretic for SA. I was looking at #4607.

Issue #4479, opened 2019-12-16
Automatically use Nathan's optimization on single-column table range scans


jeremyk-91:
For more context, see 
- https://github.com/palantir/atlasdb/pull/2480
- internal shopping product, issue 5480
- PDS-106328

@sudiksha27 this is probably a good one for us to pair on in the new year 😄

jeremyk-91:
Fixed in #4558

PR #713, opened 2016-07-23
be smarter about column selection for all columns in fixed column tab…

tpetracca:
…les; resolves #678 

CC @tmgordeeva @clockfort 

It's worth noting that the only real code change here is in the TableRenderer class.  The rest is just generated code that changed as a result.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/atlasdb/713)

PR #4546, opened 2020-02-03
Optimize range scans for single column tables 

sudiksha27:
**Goals (and why)**:
Fixes - [Reference Issue](https://github.com/palantir/atlasdb/issues/4479)

When performing a range scan for a single column table, we simply grab the latest version for the column for each row, rather than pulling all the data for each row.

- Fetch only the latest version for the column when exactly column is specified in the range scan request (existing behavior)
- When columns are not specified, check the number of columns in the table. 
- If there is exactly one column, update the slice predicate to fetch the latest version of the column for each row

- Test added to ensure correctness

- Does this actually improve perf?

CassandraKeyValueServiceImpl#getRangeWithPageCreator

By the end of this week.

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
@@ -45,6 +45,7 @@ import org.apache.cassandra.thrift.Compression;
 import org.apache.cassandra.thrift.ConsistencyLevel;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.thrift.TException;
+import org.junit.Assert;
 import org.junit.ClassRule;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -309,19 +310,6 @@ public class CassandraKeyValueServiceIntegrationTest extends AbstractKeyValueSer
 
     @Test
     public void testRangeLoader() throws Exception {
-//        int numColumns = 5;
-//        TableReference tableReference = TableReference.createFromFullyQualifiedName(
-//                "test." + RandomStringUtils.randomAlphanumeric(16));
-//        List<NamedColumnDescription> columns = new ArrayList<>();
-//        for (int i = 1; i <= numColumns; ++i) {
-//            columns.add(new NamedColumnDescription(
-//                    "c" + i, "column" + i, ColumnValueDescription.forType(ValueType.BLOB)));
-//        }
-//        keyValueService.createTable(tableReference, TableMetadata.builder()
-//                .columns(new ColumnMetadataDescription(columns))
-//                .nameLogSafety(TableMetadataPersistence.LogSafety.SAFE)
-//                .build()
-//                .persistToBytes());
         TableReference tableReference =
                 TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
         keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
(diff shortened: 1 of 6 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Dec 18 16:02:46 GMT 2019
```
felixdesouza: let's undo this

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ColumnMetadataDescription.java
```diff
@@ -66,6 +69,21 @@ public DynamicColumnDescription getDynamicColumn() {
         return ret;
     }
 
+    public Set<byte[]> getAllColumnNames() {
+        Set<byte[]> ret = Sets.newTreeSet(UnsignedBytes.lexicographicalComparator());
```
felixdesouza: nit:

```java
if (dynamicColumn != null) {
  return dynamicColumn.getColumnNameDesc().getRowParts().stream()
    .map(rowPart -> rowPart.getComponentName().getBytes())
    .collect(toCollection(() -> Sets.newTreeSet(UnsignedBytes.lexicographicalComparator())));
} else {
  return namedColumns.stream()
    .map(column -> getShortName().getBytes())
    .collect(toCollection(() -> Sets.newTreeSet(UnsignedBytes.lexicographicalComparator())));
}
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TableMetaDataProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.palantir.atlasdb.keyvalue.cassandra;
+
+import com.palantir.atlasdb.keyvalue.api.KeyValueService;
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.impl.KeyValueServices;
+import com.palantir.atlasdb.table.description.TableMetadata;
+
+public class TableMetaDataProvider {
```
felixdesouza: why do we need this class? it feels like its purpose is to be used in tests as some way of mocking i.e. not possible to mock static calls, but you don't have any reference to this in your test. Basically I'm questioning whether this indirection is necessary/worth it?/why not just be a private method inside the `CassandraKeyValueServiceImpl`.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RangeLoader.java
```diff
@@ -65,20 +69,38 @@ public void setConsistencyLevel(ConsistencyLevel consistencyLevel) {
...
+            predicate = getSlicePredicateForSingleColumn(rangeRequest, startTs);
         } else {
-            // TODO(nziebart): optimize fetching multiple columns by performing a parallel range request for
-            // each column. note that if no columns are specified, it's a special case that means all columns
-            predicate = SlicePredicates.create(SlicePredicates.Range.ALL, SlicePredicates.Limit.NO_LIMIT);
+            TableMetadata tableMetadata = metaDataProvider.getMetaDataFromRef(tableRef);
+            // todo (sudiksha) : add null check, also refactor
+            Set<byte[]> allColumnNames = tableMetadata.getColumns().getAllColumnNames();
+            if (allColumnNames.size() == 1) {
+                RangeRequest newRangeRequest = RangeRequest.builder().startRowInclusive(rangeRequest.getStartInclusive())
```
felixdesouza: non-actionable as it might actually be worse, would it make sense to basically call this method but with the range request you just created? @jeremyk-91 what do you think?

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +308,47 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
...
+                rangeRequest,
+                (-89L));
+
+        assertThat(results_1.hasNext());
+        assertThat(results_2.hasNext());
+
+        byte[] contents_1 = results_1.next().getOnlyColumnValue().getContents();
+        byte[] contents_2 = results_2.next().getOnlyColumnValue().getContents();
+
+        Assert.assertEquals(contents_1, contents_2);
```
felixdesouza: unless I'm mistaken, isn't it hard to test this performance optimisation or is the claim that if you didn't add what you added, then this test would fail because it would return too much data?

felixdesouza: if so, can we have a comment?

felixdesouza:
didn't see the "work in progress" label, sorryyyy

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TableMetaDataProvider.java
felixdesouza: I've been silly, turns out this class is used inside `RangeLoader` **not** `CassandraKeyValueServiceImpl`. Don't necessarily want to have a reference to KVS inside `RangeLoader` especially when there's a relation the other way around.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RangeLoader.java
```diff
@@ -65,20 +69,36 @@ public void setConsistencyLevel(ConsistencyLevel consistencyLevel) {
...
+
         if (rangeRequest.getColumnNames().size() == 1) {
-            byte[] colName = rangeRequest.getColumnNames().iterator().next();
-            predicate = SlicePredicates.latestVersionForColumn(colName, startTs);
+            predicate = getSlicePredicateForSingleColumn(rangeRequest, startTs);
         } else {
-            // TODO(nziebart): optimize fetching multiple columns by performing a parallel range request for
-            // each column. note that if no columns are specified, it's a special case that means all columns
-            predicate = SlicePredicates.create(SlicePredicates.Range.ALL, SlicePredicates.Limit.NO_LIMIT);
+            TableMetadata tableMetadata = metaDataProvider.getMetaDataFromRef(tableRef);
```
j-baker: It's a bit shakey to rely on table metadata here. Basically, this has never been relied upon for correctness of the sort you're trying to use it for, and so in classic Atlas fashion it will almost surely break somewhere (I can give you an internal example where I know it will).

More reliable is using the code generation, because it's at an API layer which will ignore any additional columns returned. This is the PR that probably missed it the first time around.

https://github.com/palantir/atlasdb/pull/713/files

j-baker: or, to copy here my internal comment

```
    /**
     * {@link NameMetadataDescription} and {@link ColumnMetadataDescription} are only used by atlasdb-cli to interact
     * with the internal representation of tables. They do not specify actual shape of tables.
     */
```

and so you probably don't get to assume that the information is correct :P

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RangeLoader.java
jeremyk-91: That said, see James's comment: this rewriting should be done at the codegen layer, in which case you'll probably be calling transactional/KVS methods with the different range request in any case

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RangeLoader.java
tpetracca: because of #713 do we expect this change (the whole PR) to actually change performance for single column tables? as in is the codepath we're optimizing here even possible to hit? the generated code should sense that for any "all columns" request that it should specify the specific set of columns

jeremyk-91: In general, different nodes can end up diverging on metadata information, since they read it once on startup and then never again (unless that's changed). There can be fun issues with `noColumns()` tables here as well...

PR #4548, opened 2020-02-04
[Timelock Partitioning] Part 50b: Fix develop

felixdesouza:
**Goals (and why)**:
As part of #4517, we disabled retries of `PaxosAcceptor` since it's generally in a loop elsewhere, either in Proposer loop or loop in `AwaitingLeadershipProxy`.

We did not set `retry=false` for `BatchPaxosAcceptor`, whilst simultaneously lowering the read timeout, this resulted in socket timeout in the face of partitions instead of a `RetryOther` exception (what we expect).

Change parameter from `true` to `false`.

Existing tests pass. They pass for single leader which does not use the `BatchPaxosAcceptor`, and should pass when we run a multi-leader configuration.

None

Everywhere

As important as develop being green 💃


felixdesouza:
You know it ;) Awkward moment when develop is still broken..

PR #4549, opened 2020-02-04
Timelock Metric Fixes

felixdesouza:
**Goals (and why)**:
Metrics on Timelock are somewhat hard to decipher and in some places incorrect. Small fixes to improve 

* We were both instrumenting the local remotes and combination of both the local and remote `Paxos{Acceptor|Learner}`, meaning we had two entries for the local acceptor/learners. We only instrument the "logical" `BatchPaxos{Acceptor|Learner}` that wraps the (already instrumented) remote `BatchPaxos{Acceptor|Learner}RpcClient`.
* The interface to consumers is through `BatchPaxos{Acceptor|Learner}NetworkClient`, however that itself isn't instrumented. This gives the true indication of how long a "call" took taking into account quoroms.
  * Only downside is that it's a result type which encapsulates errors, so an error - which isn't acceptable state - will count as a normal invocation, not entirely sure it's worth the effort to work around that
* `AtlasDbHttpClients#createExperimentallyWithFallback` instruments the proxy that gets returned, which is potentially further instrumented (in timelock's case, this is true). This results in double reported metrics which inflates/confuses what's actually going on. We really want to use scoped registries, but because `AtlasDbHttpClients` is static and has no real state, it's somewhat difficult (since we can't get scoped registries from a tagged metric registry). It's also not necessary, since we're not falling back to atlasdb-feign as the experiment is over.

None.

None.

Commit 👏 by 👏 commit 💃 

Relatively soon pls

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
@@ -37,17 +37,11 @@ public final class AtlasDbHttpClients {
     }
 
     public static <T> T createProxy(
-            MetricsManager metricsManager,
             Optional<TrustContext> trustContext,
             String uri,
             Class<T> type,
             AuxiliaryRemotingParameters parameters) {
-        return createExperimentallyWithFallback(
-                metricsManager,
-                () -> ConjureJavaRuntimeTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters),
-                () -> AtlasDbFeignTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters),
-                type,
-                parameters);
+        return ConjureJavaRuntimeTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters).instance();
     }
 
     /**
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
@@ -42,11 +42,13 @@ abstract class SingleLeaderNetworkClientFactories implements
                     .wrap(useCase(), remoteClients())
                     .apply(client);
             PaxosAcceptor localAcceptor = components().acceptor(client);
+
             LocalAndRemotes<PaxosAcceptor> allAcceptors = LocalAndRemotes.of(localAcceptor, remoteAcceptors);
-            return new SingleLeaderAcceptorNetworkClient(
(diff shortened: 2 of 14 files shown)


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
```diff
@@ -42,11 +42,7 @@
                     .wrap(useCase(), remoteClients())
                     .apply(client);
             PaxosAcceptor localAcceptor = components().acceptor(client);
-            LocalAndRemotes<PaxosAcceptor> allAcceptors = metrics().instrumentLocalAndRemotesFor(
-                    PaxosAcceptor.class,
-                    localAcceptor,
-                    remoteAcceptors,
-                    client);
+            LocalAndRemotes<PaxosAcceptor> allAcceptors = LocalAndRemotes.of(localAcceptor, remoteAcceptors);
```
jeremyk-91: enhanceRemotes?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
```diff
@@ -63,11 +59,8 @@
                     .wrap(useCase(), remoteClients())
                     .apply(client);
             PaxosLearner localLearner = components().learner(client);
-            LocalAndRemotes<PaxosLearner> allLearners = metrics().instrumentLocalAndRemotesFor(
-                    PaxosLearner.class,
-                    localLearner,
-                    remoteLearners,
-                    client);
+
+            LocalAndRemotes<PaxosLearner> allLearners = LocalAndRemotes.of(localLearner, remoteLearners);
```
jeremyk-91: as above, need to enhanceRemotes

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -37,17 +37,11 @@ private AtlasDbHttpClients() {
...
             String uri,
             Class<T> type,
             AuxiliaryRemotingParameters parameters) {
-        return createExperimentallyWithFallback(
-                metricsManager,
-                () -> ConjureJavaRuntimeTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters),
-                () -> AtlasDbFeignTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters),
-                type,
-                parameters);
+        return ConjureJavaRuntimeTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters).instance();
```
jeremyk-91: (non-actionable) we verified that no-one actually is still running the experiment

PR #4552, opened 2020-02-05
[PDS-109527 FLUP] TimeLock Invariant Enforcement Part 1: Only One Leader Allowed

jeremyk-91:
**Goals (and why)**:
- TimeLock should be able to detect its corruption reasonably quickly and stop itself.
- Of course, any such mechanism needs to avoid false positives.

- Whenever timelock checks its health, if it detects that there were multiple leaders for a given client `C`, it performs the following:
  - it requests a fresh timestamp for the namespace `C` on each service node, in the order `S1, S2 ... Sn`
  - it then verifies that if multiple timestamps were returned, `S1 < S2 < ... < Sn`, and `S1 > Q` where `Q` was the last value of `Sn` - if this fails, we declare an error state
  - if only one node replies, we accept that timelock is consistent
  - otherwise, we try again; after five cases where multiple nodes reply and everything is consistent, we give up: the cluster might legitimately be heavily loaded with leader elections, and there's no reason to believe that there is corruption anyway.
- The aforementioned checks take place on a separate thread, to avoid having health checks block for long amounts of time.

- New classes have unit tests added.
- I manually verified that removing the leadership proxy from TimeLock will cause the error message to be printed in `MultiNodePaxosTimeLockServerIntegrationTest`, though in general testing this is messy.

- It is possible for verifications to queue up in the event that a node is very slow to respond (along the lines of the client lock-ups ticket). Could this be a problem? Should the executor have a bounded work queue, and/or coalesce results for each client? Is this too much work?
- In the case where we don't have multiple leaders, might it make sense to test a random client? Possibly not in the scope of this PR.

**Where should we start reviewing?**: NoSimultaneousServiceCheck

**Priority (whenever / two weeks / yesterday)**: this week or early next week.


timelock-agent/src/main/java/com/palantir/timelock/invariants/ServerKiller.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private ServerKiller() {
+        // no
+    }
+
+    public static Error killMeNow(Throwable error) {
+        log.error("Something bad happened and we can't continue safely, so we're preemptively killing the server.",
+                error);
+        System.exit(1);
+        throw new SafeIllegalStateException("We should have exited before we get to this point", error);
```
gmaretic: Is this actually reachable under some conditions or is it to make IntelliJ happy?

timelock-agent/src/main/java/com/palantir/timelock/invariants/TimeLockActivityCheckerFactory.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public TimeLockActivityCheckerFactory(
+            TimeLockInstallConfiguration installConfiguration, MetricsManager metricsManager, UserAgent userAgent) {
+        this.installConfiguration = installConfiguration;
+        this.metricsManager = metricsManager;
+        this.userAgent = userAgent;
+    }
+
+    public List<TimeLockActivityChecker> getTimeLockActivityCheckers() {
+        return installConfiguration.cluster()
```
gmaretic: Ah, so the TACs are single node clients, which will throw NCLEs when we request a timestamp if they are not leaders, correct?

timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
```diff
@@ -0,0 +1,126 @@
...
+        });
+    }
+
+    private void performCheckOnSpecificClientUnsafe(Client client) {
+        // Only fail on repeated violations, since it is possible for there to be a leader election between checks that
+        // could legitimately cause false positives if we failed after one such issue. However, given the number of
+        // checks it is unlikely that *that* many elections would occur.
+        for (int attempt = 1; attempt <= REQUIRED_CONSECUTIVE_VIOLATIONS_BEFORE_FAIL; attempt++) {
+            long numberOfNodesServingTimestamps = timeLockActivityCheckers.stream()
+                    .map(timeLockActivityChecker ->
```
gmaretic: just filter instead of map + filter

timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
```diff
@@ -0,0 +1,126 @@
...
+                + " checks on these specific clients now.", SafeArg.of("clients", clientsWithMultipleLeaders));
+        clientsWithMultipleLeaders.forEach(this::scheduleCheckOnSpecificClient);
+    }
+
+    private void scheduleCheckOnSpecificClient(Client client) {
+        executorService.submit(() -> {
+            try {
+                performCheckOnSpecificClientUnsafe(client);
+            } catch (Exception e) {
+                log.info("No-simultaneous service check failed, suppressing exception to allow future checks", e);
```
gmaretic: Is this actually useful? We don't need to suppress because there are not scheduled. Also if something went wrong with TACs but we keep trying to check, we will never really figure out something is wrong

timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
```diff
@@ -0,0 +1,126 @@
...
+                    .count();
+            if (numberOfNodesServingTimestamps <= 1) {
+                // Accept 0: the cluster being in such a bad state is not a terminal condition, could just be a
+                // network partition or legitimate no-quorum situation. No reason to kill the server then.
+                log.info("We don't think services were simultaneously serving timestamps for client {}",
+                        SafeArg.of("client", client));
+                return;
+            }
+
+            if (attempt < REQUIRED_CONSECUTIVE_VIOLATIONS_BEFORE_FAIL) {
```
gmaretic: You are already in a for loop, you don't need the if/else

timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
```diff
@@ -97,30 +103,36 @@ private void performCheckOnSpecificClientUnsafe(Client client) {
...
-                        SafeArg.of("backoffMillis", BACKOFF.toMillis()));
-                Uninterruptibles.sleepUninterruptibly(BACKOFF.toMillis(), TimeUnit.MILLISECONDS);
-            } else {
+            if (!Ordering.natural().isStrictlyOrdered(timestamps) || timestamps.get(0) <= timestampBound) {
                 failureMechanism.accept(client.value());
+                return;
             }
+
+            timestampBound = timestamps.get(timestamps.size() - 1);
+            log.info("We observed on attempt that multiple services were serving timestamps, but the timestamps were"
```
jeremyk-91: I think we want the attempt number?

changelog/@unreleased/pr-4552.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: TimeLock now logs ERROR messages if multiple nodes are detected to
```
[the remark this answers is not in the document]
jeremyk-91: Yes, will change the changelog. That's a good flag

timelock-agent/src/main/java/com/palantir/timelock/invariants/ServerKiller.java
jeremyk-91: Don't think it's reachable: System.exit() should never return normally (it could throw a `SecurityException`)

PR #4555, opened 2020-02-06
[PDS-110262] Meter for Cassandra pool exhaustion events

jeremyk-91:
**Goals (and why)**:
- Allow easy visualisation of when requests are being dropped because our connection pools are getting exhausted. See internal ticket PDS-110262, this would have helped figure things out more quickly.

- Add a meter from the metrics registry that is marked whenever the pool is exhausted on a request

None, admittedly (not breaking existing tests is progress, of course)

- The goal here is to add only one meter per namespace. Was this done correctly?
- Are there cases where this over or under counts the number of request failures?
- Not really a fan of passing the `CassandraClientPoolMetrics` objects around. Should I have passed a callback instead?

**Where should we start reviewing?**: CassandraClientPoolMetrics

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Meter;
 import com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPool;
 import com.palantir.atlasdb.keyvalue.cassandra.CassandraClientPoolingContainer;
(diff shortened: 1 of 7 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
```diff
@@ -28,12 +28,18 @@
 
 public class CassandraClientPoolMetrics {
     private final MetricsManager metricsManager;
-    private final RequestMetrics aggregateMetrics;
+    private final RequestMetrics aggregateRequestMetrics;
     private final Map<InetSocketAddress, RequestMetrics> metricsByHost = new HashMap<>();
 
+    // Tracks occurrences of client pool exhaustions.
+    // Not bundled in with request metrics, as we seek to not produce host-level metrics for economic reasons.
+    private final Meter poolExhaustionMeter;
```
jkozlowski: Just a Counter should suffice?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
```diff
@@ -61,10 +67,14 @@ public void recordConnectionExceptionOnHost(CassandraClientPoolingContainer host
         updateMetricOnAggregateAndHost(hostPool, RequestMetrics::markRequestConnectionException);
     }
 
+    public void recordPoolExhaustion(CassandraClientPoolingContainer unused) {
```
jkozlowski: Do we need to pass the CassandraClientPoolingContainer if it's unused?

changelog/@unreleased/pr-4555.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: AtlasDB now publishes a metric indicating the count of requests that
+    were dropped owing to saturated Cassandra connection pools.
```
jkozlowski: I don't think connections get dropped, they still retry. Maybe "AtlasDB now publishes a metric that can be used to monitor occurences of Cassandra connection pool exhaustion"

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
jeremyk-91: Yep, that's right - will change.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolMetrics.java
jeremyk-91: I did this more for consistency with the existing methods. Though yeah it might be worth not passing it to not give people the impression that it would be reported

changelog/@unreleased/pr-4555.v2.yml
jeremyk-91: Yep, it's a bit confusing - it's the count of requests _internal to the Cassandra client pool_ that are dropped. At a higher level (i.e. `CassandraClientPool` as opposed to the individual pooling container) they will get retried.

Issue #4351, opened 2019-10-28
CassandraKeyValueServiceImpl executor thread pool never removes idle threads


OStevan:
Even though the current implementation sets the keep alive timer for removing idle threads to 1 minute on the `executor` it also sets the core pool size and max pool size to the same value. As a result threads are never removed from the pool.

If we in fact want the pool to remove some threads one possible solution is to have a config option which will specify the fraction of threads from the pool which can be removed. The default value for this would be 0 so that users have to explicitly set it if they think they need such feature.

jkozlowski:
I think the tradeoff here is how quickly the threads are started when you need the capacity. We can probably get metrics on the utilization of those threadpools and come up with a good initial value.

PR #4556, opened 2020-02-07
Make max threadpool size for Cass calls track connection pool sizes.

jkozlowski:
**Goals (and why)**:

This threadpool controls the number of concurrent Cassandra calls we make. Each call is effectively a retrying thing that can call any number of Cassandra hosts, but ideally should be doing just 1. Underneath it there are per host pools that have min and burst size (by default 30/100). If one of those pools is exhausted, the calls will sleep inside of this threadpool to backoff.

Letting this number go up might increase the overall congestion for the connection pools. 


jkozlowski:
Fixes https://github.com/palantir/atlasdb/issues/4351

PR #4560, opened 2020-02-10
[PDS-109527 FLUP] TimeLock Invariant Enforcement, Part 2: The ServerKiller is out for blood

jeremyk-91:
**DO NOT MERGE until we are confident that we aren't exposing ourselves to too many false positives.**

**Goals (and why)**:
- Actually shut down rogue timelocks faster.

- Change the penalty for an invariant violation from an ERROR log to actually taking out the cluster.

- None beyond existing testing to validate that in the normal case things still work.

- Nothing in particular.

**Where should we start reviewing?**: `NoSimultaneousServiceCheck.java`

**Priority (whenever / two weeks / yesterday)**: If there are zero instances of the ERROR state logs by Monday next week, I think we should be good to merge.

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
@@ -34,6 +34,7 @@ import com.google.common.util.concurrent.Uninterruptibles;
 import com.palantir.atlasdb.timelock.paxos.Client;
 import com.palantir.common.concurrent.PTExecutors;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.timelock.TimeLockStatus;
 import com.palantir.timelock.paxos.HealthCheckDigest;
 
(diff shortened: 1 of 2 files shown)


timelock-agent/src/main/java/com/palantir/timelock/invariants/NoSimultaneousServiceCheck.java
```diff
@@ -64,15 +65,11 @@ public static NoSimultaneousServiceCheck create(List<TimeLockActivityChecker> ti
...
-                    // TODO (jkong): Gather confidence and then change to ServerKiller, so that we ACTUALLY shoot
-                    // ourselves in the head.
-                    log.error("We observed that multiple services were consistently serving timestamps, for the"
-                            + " client {}. This is potentially indicative of SEVERE DATA CORRUPTION, and should"
-                            + " never happen in a correct TimeLock implementation. If you see this message, please"
-                            + " check the frequency of leader elections on your stack: if they are very frequent,"
-                            + " consider increasing the leader election timeout. Otherwise, please contact support -"
-                            + " your stack may have been compromised",
-                            SafeArg.of("client", client));
+                    throw ServerKiller.killMeNow(new SafeIllegalStateException("We observed that multiple services "
```
felixdesouza: ignore me, I'm testing out intellij gh integration with comments :dancer:

felixdesouza: it works 💃 although can't do reviews :(

jeremyk-91:
We have validated that running for 1.5 weeks there have been no false positives result.

If this *does* happen, remediation procedures are along the lines of your standard data corruption thing - restore from backup.

jeremyk-91:
Probably worth re-working this PR then, agree that disabling actual service and reporting an e.g. ERROR health check (which will page us) is probably a better idea than just killing the server.

jeremyk-91:
The validation mechanism won't change, and we can be certain that that is in a safe place.

jkozlowski:
@jeremyk-91 feel like you want to keep this around, or at least link internally to some tasks we're working on?

PR #4561, opened 2020-02-11
[Conjure Java Runtime] Part 26: Story of Thanatosis (Removing AtlasDB-Feign Entirely)

jeremyk-91:
**Goals (and why)**:
- Remove the `atlasdb-feign` project from the codebase.
- Generally leave people fewer complications to think about in the remoting stack (e.g. James thought that we could still fallback when beginning to write the Conjure migration). I thought I'd give a 1 hour stab at this, and it looks like it wasn't as hard as I thought.

- Remove `atlasdb-feign`, fix compile and checkstyle issues that arise from that.

- Existing tests still pass.
- The code compiling serves as verification that it isn't referenced any more.

- I wanted to eliminate `RemotingClientConfig` as well, but wasn't too happy with the API break/mess there.
- Should the experiment running proxy code be retained?
- There are a few trouble spots:
  - Leader block mechanism to elide remote calls in a 1 node cluster used Feign but have now been switched to CJR. I've updated the failure mechanism to match, but is this correct?
  - Benchmarking infrastructure had some fancy magic with `FeignOkHttpClients`, which of course isn't used in CJR. Does this still work?
- Are there some edge cases that I've missed out?

**Where should we start reviewing?**: Top down, looking at places which are actually `+/-` as opposed to just `-` is probably a good approach.

**Priority (whenever / two weeks / yesterday)**: Whenever


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -56,54 +51,24 @@ private AtlasDbHttpClients() {
...
             Class<T> type,
             AuxiliaryRemotingParameters parameters) {
-        return createExperimentallyWithFallback(
-                metricsManager,
-                () -> ConjureJavaRuntimeTargetFactory.DEFAULT.createProxyWithFailover(
-                        serverListConfig, type, parameters),
-                () -> AtlasDbFeignTargetFactory.DEFAULT.createProxyWithFailover(serverListConfig, type, parameters),
-                type,
-                parameters);
+        return VersionSelectingClients.instrumentWithClientVersionTag(
```
jkozlowski: I haven't read the PR, but we need to make sure we don't miss the code in ExperimentRunningProxy:

```
this.refreshingExperimentalServiceSupplier = Suppliers.memoizeWithExpiration(
                experimentalServiceSupplier::get, CLIENT_REFRESH_INTERVAL.toMinutes(), TimeUnit.MINUTES);  
```

Without this we loose the 30 minute client reload thing, meaning we'll get the lockups back again.

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/AuxiliaryRemotingParameters.java
```diff
@@ -48,7 +48,11 @@ default boolean shouldSupportBlockingOperations() {
         return true;
     }
 
+    /**
+     * @deprecated Conjure is now always used, and supplying alternative configs is not possible.
```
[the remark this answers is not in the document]
felixdesouza: I think we should go through and purge the config stuff properly. Deprecations don't really do anything if we never remove them.

I also don't think we want to deprecate this *anyway*. Since if we do need to open up other configurations i.e. if we want to add some configurability for our own testing purposes, this is the right place to do it.

Starting from the config `RemotingClientConfig`, we can delete the two properties, and if we're super concerned we can do what we did in `AtlasDbConfig`, where as a class level annotation, we do `@JsonIgnoreProperties({"maximumConjureRemotingProbability", "enableLegacyClientFallback"}` and then we delete those properties.

We should probably add a `RemotingClientConfigs.DEFAULT` configuration, and then have the old thing `RemotingClientConfigs.ALWAYS_USE_CONJURE` refer to that, and change all of our references to that.

Then we can propagate the deletion of the java properties everywhere (i.e. and continue to get rid of `VersionSelectingClients`.

timelock-agent/src/test/java/com/palantir/timelock/paxos/LeaderPingHealthCheckTest.java
```diff
@@ -188,7 +187,7 @@ private static HealthCheckPinger getMockOfPingableLeaderWherePingReturns(Set<Cli
 
     private static HealthCheckPinger getMockOfPingableLeaderWherePingThrows() {
         HealthCheckPinger mockLeader = mock(HealthCheckPinger.class);
-        when(mockLeader.apply(ALL_CLIENTS)).thenThrow(mock(AtlasDbRemoteException.class));
+        when(mockLeader.apply(ALL_CLIENTS)).thenThrow(mock(IllegalStateException.class));
```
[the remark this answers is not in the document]
jeremyk-91: Yep, sorry about this 😞

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -83,11 +87,23 @@ private AtlasDbHttpClients() {
...
-        Supplier<T> clientFactory = () -> VersionSelectingClients.instrumentWithClientVersionTag(
+        Supplier<T> clientFactory = () -> instrument(
                 metricsManager.getTaggedRegistry(),
                 ConjureJavaRuntimeTargetFactory.DEFAULT.createProxyWithQuickFailoverForTesting(
                         serverListConfig, type, parameters),
                 type);
         return SelfRefreshingProxy.create(clientFactory, type);
     }
+
+    private static <T> T instrument(
```
felixdesouza: iirc, instrumentation has always been separate from creating clients. i.e. `ServiceCreator` and in timelock (before and after my changes). Unless they're all changed, we're going to double report, and likely we'll have wrong metrics on our dashboards.

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/RemotingClientConfigs.java
```diff
@@ -21,8 +21,5 @@ private RemotingClientConfigs() {
         // Constants
     }
 
-    public static final RemotingClientConfig ALWAYS_USE_CONJURE = ImmutableRemotingClientConfig.builder()
```
felixdesouza: note to self/you: you'll have to change the places where this is referred to explicitly namely big internal product and it's related projects.

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
jeremyk-91: good catch!

atlasdb-remoting-api/src/main/java/com/palantir/atlasdb/config/RemotingClientConfigs.java
jeremyk-91: Yeah, that's right. Thanks

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
jeremyk-91: Discussed offline: This is messy, because just removing will mean that we lose some metrics - not true that all `ServiceCreator` created things are instrumented. Will do as a follow up to this PR.

PR #4563, opened 2020-02-11
Make createProxy retain refreshing capabilities.

felixdesouza:
**Goals (and why)**:
As part of figuring out why Timelock is OOMing, rolling back potentially contentious changes that occurred between 0.185.3 and 0.185.4 of Atlas

Restore the client refreshing behaviour but drop the extra metrics.

Existing tests should pass

is `null` the worst thing here?

Everywhere

ASAP 💃

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
@@ -24,6 +24,7 @@ import java.util.function.Supplier;
 import org.immutables.value.Value;
 
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
 import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.config.RemotingClientConfig;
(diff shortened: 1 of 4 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/http/VersionSelectingClients.java
```diff
@@ -83,6 +84,19 @@ private VersionSelectingClients() {
                 errorMetric::increment);
     }
 
+    static <T> T createRefreshingNewClient(
+            Supplier<TargetFactory.InstanceAndVersion<T>> newClientSupplier,
+            Class<T> clazz) {
+
+        return ExperimentRunningProxy.newProxyInstance(
+                Suppliers.compose(TargetFactory.InstanceAndVersion::instance, newClientSupplier::get),
+                null,
```
jeremyk-91: Is ok, since `useExperimental` is true and `useFallback` false. Is not great, but maybe exploding if we get into this kind of state that we don't expect is possible is not the worst reaction.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/LeadersTest.java
```diff
@@ -116,7 +116,7 @@ public void createProxyAndLocalListCreatesSingletonListIfNoRemoteAddressesProvid
         verifyNoMoreInteractions(localAcceptor);
     }
 
-    @Test(expected = IllegalStateException.class)
+    @Test(expected = IllegalArgumentException.class)
```
jeremyk-91: wat 🦆

changelog/@unreleased/pr-4563.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
```
jeremyk-91: nit: fix?

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/LeadersTest.java
felixdesouza: 🤷‍♂ weird, I know

PR #4565, opened 2020-02-11
[Timelock Partitioning] Part 51: Wiremock integration tests for multi leader

felixdesouza:
**Goals (and why)**:
So previously, when #4515 merged/was introduced it didn't really assert anything regarding the multi leader capabilities. It just confirmed that for a single leader, are we still "correct".

With this PR we add specific tests exercise/assert that we're doing the right things. As it stands they're not exhaustive, but with the framework I've added it should make for a pretty easy way to add extra tests.

* We use wiremock as a proxy in front of timelock. This allows us to arbitrarily spy/reject requests to timelock that enables forcing timelock into a particular state. i.e. if we have decided on an allocation for leaders for timelock, we can then block all requests to that namespace, and ensure that the clients we use in the test all go through the proxy that's wired up to the wiremock server.
* We add some basic smoke tests to confirm this works as we want i.e. if I were to speak to a node directly (not via failover) it'll 308 when it's not the leader etc.

This is all tests.

Tests here aren't exhaustive, happy to pair to add some more.

Commit by commit really.

ASAP 💃


timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiLeaderMultiNodePaxosTimeLockIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        TestableTimelockServer allocatedLeader = Iterables.getOnlyElement(
+                ImmutableSetMultimap.copyOf(allocations).inverse().get(randomNamespace));
+
+        assertThat(cluster.currentLeaderFor(randomNamespace))
+                .isEqualTo(allocatedLeader);
+
+        cluster.nonLeaders(randomNamespace).get(randomNamespace).forEach(nonLeader ->
+                assertThatThrownBy(() -> nonLeader.client(randomNamespace).getFreshTimestamp())
+                        .as("non leaders should return 503")
```
jeremyk-91: nit: 308

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiLeaderMultiNodePaxosTimeLockIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+                .map(NamespacedClients::throughWireMockProxy)
+                .forEach(NamespacedClients::getFreshTimestamp);
+
+        SetMultimap<TestableTimelockServer, String> namespacesByLeader =
+                ImmutableSetMultimap.copyOf(cluster.currentLeaders(allNamespaces)).inverse();
+
+        assertThat(namespacesByLeader)
+                .isEqualTo(allocations);
+
+        cluster.servers().forEach(TestableTimelockServer::allowAllNamespaces);
```
jeremyk-91: Discussed offline: The reason this test works is because in the setup, if we have assigned namespace `NAMESPACE` to node `A`, then only `A` knows about `NAMESPACE`. So only `A` will propose in the Paxos sequence for `NAMESPACE`, and thus `A` should become the leader. Also, before we `allowAllNamespaces` the other nodes will just reply with 503s that are in front of the direct proxy, so it's not possible that there will be another election in the other namespaces causing `cluster.currentLeaders(...)` to change.

This is pretty clever! Maybe worth a comment somewhere.

felixdesouza: with some of my other work, we can probably use the takeover primitive instead to more reliably deal with this.

PR #4575, opened 2020-02-13
[Conjure Java Runtime] Part 28: Modernisation and Recovery of Jepsen

jeremyk-91:
**Goals (and why)**:
- Get the Jepsen tests working again. They would have been useful for an issue like PDS-109527: I haven't tested this, but am quite confident that running timestamp verification with the `partition-random-halves` nemesis could have exposed the split-brain scenario that we were vulnerable to.
- This PR does NOT include ways of making the Jepsen test results more visible, though that is something I would like to do in the medium term. I think we stopped looking at it because the idea was that no one should ever touch core TimeLock again, but there are a lot of pieces in AtlasDB that we don't always realise influence core TimeLock - or we do (as in PDS-109527) but don't see awkward race conditions - this is what fuzzing is great at doing.

- This PR restores operation of the Jepsen infrastructure through some problematic changes for Conjure Java Runtime and other relevant breaks:

1. **Oracle changed their policy around the use of Java 8 JDKs.** To fix this, I switched to OpenJDK. Unfortunately this was not entirely trivial because the Docker image for `tjake/jepsen` doesn't seem to be updated to deal with this (even though Jepsen itself has), and `java-8-openjdk` is part of `stretch` (think stable) but not `jessie` (think super-stable) debian repositories, which is what we were on.
2. **Conjure Java Runtime requires the use of SSL.** Fair enough, but again nontrivial because the old servers weren't set up to use SSL at all, and the standard certificates we have will fail hostname verification (since the nodes are hardcoded to `n1` through `n5`).
3. **There were considerable breaks to the API endpoints to allow for namespace-agnostic clients.** This was much easier to address than the first two issues.

- I ran the Jepsen tests through gradle and verified that connections were being made, actual nontrivial responses were being returned (i.e. numbers for timestamp, lock tokens for lock), and the tests passed.

- This works on my machine, is there a reason to believe it won't on internal test stack?

**Where should we start reviewing?**: Follow the flow of the breaks described above.

**Priority (whenever / two weeks / yesterday)**: should have been 4 months ago, but this week isn't too bad 🔥 

@j-baker for SA, we discussed this offline.
@Jolyon-S I think you might find this one interesting, though would like to have Felix or Grgur take a pass before we merge


atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/http/TimelockUtils.java
```diff
@@ -43,21 +46,26 @@ private TimelockUtils() {
...
+        ServerListConfig serverListConfig = ImmutableServerListConfig.builder()
+                .addAllServers(endpointUris)
+                .sslConfiguration(SSL_CONFIGURATION)
+                .build();
+        // lock technically blocking, but OK to run with short timeout given the way the tests work.
+        AuxiliaryRemotingParameters build = AuxiliaryRemotingParameters.builder()
+                .shouldRetry(false)
+                .shouldLimitPayload(false)
+                .shouldSupportBlockingOperations(false)
+                .userAgent(UserAgent.of(UserAgent.Agent.of("atlasdb-jepsen", "1.2.3")))
```
Jolyon-S: Why don't we use `UserAgent.Agent.DEFAULT_VERSION` anymore?

felixdesouza: I imagine it's slightly more identifiable, but can't think of another reason other than that 🤷‍♂

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/http/TimelockUtils.java
```diff
@@ -43,21 +46,26 @@ private TimelockUtils() {
...
-        return Lists.transform(hosts, host -> String.format("https://%s:%d/%s", host, PORT, NAMESPACE));
+        return hosts.stream().map(host -> String.format("https://%s:%d", host, PORT)).collect(Collectors.toList());
     }
 
     private static <T> T createFromUris(MetricsManager metricsManager, List<String> endpointUris, Class<T> type) {
+        ServerListConfig serverListConfig = ImmutableServerListConfig.builder()
+                .addAllServers(endpointUris)
+                .sslConfiguration(SSL_CONFIGURATION)
+                .build();
+        // lock technically blocking, but OK to run with short timeout given the way the tests work.
```
felixdesouza: why is this okay? what happens if you make it the same as how it's used in prod?

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/http/TimelockUtils.java
```diff
@@ -43,21 +46,26 @@ private TimelockUtils() {
...
+        AuxiliaryRemotingParameters build = AuxiliaryRemotingParameters.builder()
```
felixdesouza: nit: `build` -> `parameters`

jeremyk-91:
TODO: Further extend self signed cert validity

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/http/TimelockUtils.java
jeremyk-91: Ah, that's fair, I forgot that existed - changed to that.

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/http/TimelockUtils.java
jeremyk-91: Is also fine, just that I'm not really a fan of duplicating that logic here, and it seems a bit unpleasant to expose internals of `TransactionManagers` to simulate that wrapping. Let me know what you think.

felixdesouza: I suppose I'm asking from a perspective of "will jepsen say there's an illegal trace, when really it would never happen in prod"?

PR #4580, opened 2020-02-17
[Timelock OOMs] Memoized supplier with timeout client

felixdesouza:
**Goals (and why)**:
From the bisections, fairly certain that the issue lies with metrics PR #4549. The only meaningful thing that comes from that PR is the fact that we have a client that is refreshing behind the scenes via `ExperimentRunningProxy` etc. and in that PR we don't. 

~Currently testing the hypothesis that HTTP1 does not suffer from this issue i.e. some weird interop between okhttp and undertow when dealing with h2.~ Turns out that it's equal opportunity OOMs :(.

If this experiment works, then will defer to #4561 for the better implementation.

other FLUPs:
* when we fail when we've gained leadership and we're trying to create the delegate, we repeatedly and aggressively try to create the delegate with no backoff (backoff isn't the issue here per se). In doing so, we also create resources that will never be used i.e. threads, may not take up processing time, but they do take up memory. In some places saw around 30k threads at one time (couldn't get a JStack as the node was unresponsive), but in other places, you can definitely see elevated thread counts (this does happen independently of the GC storm IIRC). 
  * FLUP at bottom: ensure we shutdown any auxiliary resources created as part of failed createDelegate calls
* undo old support work i.e. Targeted Sweep rate limiter, this adds threads per client, should be light, but it's resources 🙄 

* Suppliers.memoizeWithExpiration etc.

About to test on an RC. 

Is 10 minutes too frequent?

Everywhere.

I'm just talking to myself at this point.


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -41,7 +44,12 @@ private AtlasDbHttpClients() {
             String uri,
             Class<T> type,
             AuxiliaryRemotingParameters parameters) {
-        return ConjureJavaRuntimeTargetFactory.DEFAULT.createProxy(trustContext, uri, type, parameters).instance();
+        return Suppliers.memoizeWithExpiration(
```
jkozlowski: So I thought this was the first thing we tried and with 30 minutes expiration it was OOMing. Why is the 30 minutes fine on develop, but not fine here?

felixdesouza: Unclear, I'm chalking it down to me doing it wrong the first time and trying to get it to work with everything that exists already. I'm struggling to figure out an alternate hypothesis tbh.

jkozlowski: ok, so I would say we probably want to keep the 30 minute period here, in case your original PR just didn't do the wiring correctly. Then it's a representative test case

felixdesouza: cool

felixdesouza: well no build is working because certs 😠

PR #4582, opened 2020-02-17
Update certs again.

felixdesouza:
**Goals (and why)**:
#4505 Updated Certs but the certs went only to 15 Feb 2020, now they should go to somewhere in 2023

ASAP, develop is broken without it.


felixdesouza:
Well if you had to make the certs yourself, I imagine it's fine because you'll have the expiry is far into the future. Out of curiosity, how long did you put the expiry for those custom certs for?

jeremyk-91:
I had them as the default, which is 180 days. I'll extend them

felixdesouza:
3 years 💃

PR #4583, opened 2020-02-17
Remove targeted sweep rate limiting from Timelock.

felixdesouza:
**Goals (and why)**:
We introduced targeted sweep lock rate limiting in timelock to stem the bleeding from a bad configuration change. #4096 #4110 and #4115. Now that we've remediated, we should get rid of it, especially as it can consume resources (albeit not very much) but the point still stands.

Too many conflicts when reverting, so I just deleted all references to anything `TargetedSweepAsyncLock` etc.

This should be covered by existing tests.

Have I fundamentally changed how locks work?

Anywhere, mostly deletes.

At some point this week.


timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockRuntimeConfiguration.java
```diff
@@ -30,6 +30,7 @@
 @JsonDeserialize(as = ImmutableTimeLockRuntimeConfiguration.class)
 @JsonSerialize(as = ImmutableTimeLockRuntimeConfiguration.class)
 @Value.Immutable
+@JsonIgnoreProperties("targeted-sweep-locks")
```
jeremyk-91: 💯

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/LockCollection.java
```diff
@@ -45,7 +43,7 @@ public OrderedLocks getAll(Set<LockDescriptor> descriptors) {
             locks.add(getLock(descriptor));
         }
 
-        return locksDecorator.decorate(orderedDescriptors, locks);
+        return OrderedLocks.fromOrderedList(locks);
```
jeremyk-91: sanity checked this matches the world before #4096

PR #4584, opened 2020-02-17
Change conjure client replacement logic to replace ASAP

j-baker:


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -92,10 +92,10 @@ private AtlasDbHttpClients() {
             AuxiliaryRemotingParameters clientParameters) {
         TargetFactory.InstanceAndVersion<T> fallbackProxy = fallbackProxySupplier.get();
         try {
-            return VersionSelectingClients.createVersionSelectingClientWithRefreshingNewClient(
+            return VersionSelectingClients.createVersionSelectingClient(
                     metricsManager,
-                    experimentalProxySupplier,
                     fallbackProxy,
+                    experimentalProxySupplier.get(),
```
jeremyk-91: The ordering of the arguments has to be the other way (i.e. you still want `experimentalProxySupplier.get()` before `fallbackProxy`).

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -103,14 +107,32 @@ private ConjureJavaRuntimeTargetFactory() {
...
-                clientConfiguration);
+                clientConfiguration));
         return decorateFailoverProxy(type, () -> client);
     }
 
+    private static <T> T wrapConjureClientWithOkHttpTimeoutHandling(Class<T> iface, Supplier<T> clientFactory) {
+        return ReplaceIfExceptionMatchingProxy.newProxyInstance(
+                iface,
+                Suppliers.memoizeWithExpiration(clientFactory::get, 20, TimeUnit.MINUTES),
+                ConjureJavaRuntimeTargetFactory::isPossiblyOkHttpTimeoutBug);
```
jeremyk-91: What this does:
- The client is replaced not more than once every 20 minutes.
- In the event we get a SocketTimeoutException or something with that somewhere in the causal chain, we replace the client (subject to the above limitation).

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,64 @@
...
+            replaceIfNecessary(cause);
+            throw cause;
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private synchronized void replaceIfNecessary(Throwable thrown) {
+        if (shouldReplace.test(thrown)) {
+            delegate = delegateFactory.get();
```
felixdesouza: is there a nice way to prevent the factory being called n times if we get a whole batch of failing requests? Or is this handled at a higher level?

j-baker: method is synchronized

jkozlowski:
Merged in another PR

PR #4587, opened 2020-02-19
[Timelock OOMs] Stop `PaxosQuorumChecker` cancellation

felixdesouza:
**Goals (and why)**:
As part of our annoyingly long running investigation into Timelock OOMs, we stopped cancelling remaining calls in the `PaxosQuorumChecker`. The benefit of this was two-fold:
* We don't get those spurious log lines about leaked connections or at the very least the volume is significantly reduced and we only really see them when a GC happens.
  * This makes sense since that leak logger is implemented with Garbage collection.
  * Supposedly this all changes in 3.14.x where it should be fixed, but internally timelock is using 3.13.1, a potential follow up would be to upgrade and see whether we're just better.
* We don't OOM! 🎉 💃 

* Since we control Timelock and operate it internally, we have a bit more control over how we want change cancellations. However, the same cannot be said about big internal product and anyone on the leader block. They may be running in an environment that makes it a lot more difficult/consume more resources.

Existing tests should pass.

The wiring of this sucks, and feels like the `PaxosQuorumChecker` should no longer be a static class, and have a bit of state. At the very least it would make code that uses it a bit clearer. It's somewhat out of scope, if someone else wants to pick it up sure, but I think it's okay to take on this tech debt, I just can't afford to pay it right now.

Commit by commit, first is the refactor, the next is the config change.

ASAP. Unblocks Atlas releases on Timelock.

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
@@ -221,7 +221,9 @@ public final class PaxosQuorumChecker {
             log.warn("paxos request interrupted", e);
             interrupted = true;
         } finally {
-            cancelOutstandingRequestsAfterTimeout(cancelRemainingCalls, allFutures);
+            if (cancelRemainingCalls) {
+                cancelOutstandingRequestsAfterTimeout(allFutures);
+            }
 
             if (interrupted) {
                 Thread.currentThread().interrupt();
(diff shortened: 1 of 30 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosQuorumChecker.java
```diff
@@ -249,9 +243,11 @@ private static boolean timedOut(Future<?> responseFuture) {
     }
 
     private static <SERVICE, RESPONSE extends PaxosResponse> void cancelOutstandingRequestsAfterTimeout(
+            boolean shouldCancel,
             List<Future<Map.Entry<SERVICE, RESPONSE>>> responseFutures) {
+
         boolean areAllRequestsComplete = responseFutures.stream().allMatch(Future::isDone);
-        if (areAllRequestsComplete) {
+        if (!shouldCancel || areAllRequestsComplete) {
```
jeremyk-91: nit: ok. I'd probably just not call this method if `shouldCancel` is false above, but this is fine.

felixdesouza:
Don't merge just yet, getting rid of the configurability from the timelock side.

jkozlowski:
Can we just make sure that this is better and not provide the configuration? Providing the ability to configure this feels painful, I would not expect people to know how/if they should configure this. If we're worried about the threadpool utilisation, we should add monitoring for it or even better switch to a Retrofit client and get rid of the specific threadpool altogether. I just did a quick check and the timeouts on those calls are not that long?

felixdesouza:
As per offline comment from @j-baker

jkozlowski:
Lol, added the comment at almost exactly the same time

PR #4588, opened 2020-02-19
Safer workflow for getting rows from CassandraKvs

sudiksha27:
**Goals (and why)**:
The `getRows` endpoint fetches all the cells (all the versions) for a given list of rows  from Cassandra if the columns are not explicitly specified. This endpoint thus causes a lot of very large reads on Cassandra when reading a list of rows from a table with many rows that are quite wide.

Instead of reading all the cells (and all the versions of these cells) from Cassandra in one request, split the very large read query into number of serial requests with limited reads. It is maintained that all the cells are read and only (some) older versions of cells may be skipped while reading from Cassandra.

- Modify `getRowsForSingleHost` endpoint to split one very large read query into number of serial requests with limited reads.

- Tests added to verify correctness

- Tuning required depending on individual deployment
- Do hit and trial in order to determine the read limit per row per request.

- CassandraKeyValueServiceImpl#getRowsForSingleHost

- EOD - 24 Feb

### DIFF
--- a/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
+++ b/atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
@@ -40,6 +40,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
 
 import org.apache.cassandra.thrift.CfDef;
 import org.apache.cassandra.thrift.Compression;
@@ -59,7 +61,6 @@ import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.ListMultimap;
 import com.google.common.io.BaseEncoding;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cassandra.CassandraKeyValueServiceConfig;
@@ -84,6 +85,7 @@ import com.palantir.atlasdb.table.description.ValueType;
 import com.palantir.atlasdb.transaction.api.ConflictHandler;
 import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.atlasdb.util.MetricsManagers;
+import com.palantir.common.streams.KeyedStream;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 
@@ -238,7 +240,8 @@ public class CassandraKeyValueServiceIntegrationTest extends AbstractKeyValueSer
     private static ImmutableCassandraKeyValueServiceConfig getConfigWithGcGraceSeconds(int gcGraceSeconds) {
         return ImmutableCassandraKeyValueServiceConfig
                 .copyOf(CASSANDRA.getConfig())
-                .withGcGraceSeconds(gcGraceSeconds);
+                .withGcGraceSeconds(gcGraceSeconds)
+                .withFetchReadLimitPerRow(100);
     }
 
     private static String getInternalTestTableName() {
@@ -304,7 +307,7 @@ public class CassandraKeyValueServiceIntegrationTest extends AbstractKeyValueSer
     }
 
     @Test
-    public void testGetRows() {
+    public void testGetRows_highlyVersionedCells() {
         TableReference tableReference =
                 TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
         keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
@@ -313,27 +316,53 @@ public class CassandraKeyValueServiceIntegrationTest extends AbstractKeyValueSer
 
         Cell CELL_WITH_VERSIONS = Cell.create(row(1), column(1));
         Cell CELL_WITH_SAME_ROW = Cell.create(row(1), column(2));
-        Cell CELL_WITH_DIFFERENT_ROW = Cell.create(row(2), column(1));
 
         ImmutableListMultimap<Cell, Value> tableValues = ImmutableListMultimap.<Cell, Value>builder()
-                .putAll(CELL_WITH_VERSIONS, valueWithTimestamps(data, 57L, 72L, 100L))
-                .putAll(CELL_WITH_SAME_ROW, valueWithTimestamps(data, 63L))
-                .putAll(CELL_WITH_DIFFERENT_ROW, valueWithTimestamps(data, 42L, 81L))
+                .putAll(CELL_WITH_VERSIONS, valueWithNumberOfTimestamps(data, 250L))
+                .putAll(CELL_WITH_SAME_ROW, valueWithNumberOfTimestamps(data, 200L))
                 .build();
-
 
         keyValueService.putWithTimestamps(tableReference, tableValues);
 
         Map<Cell, Value> result = keyValueService.getRows(
                 tableReference,
-                ImmutableList.of(row(1), row(2)),
+                ImmutableList.of(CELL_WITH_VERSIONS.getRowName(), CELL_WITH_SAME_ROW.getRowName()),
                 ColumnSelection.all(),
                 STARTING_ATLAS_TIMESTAMP - 1);
 
         assertThat(result).containsOnly(
-                entry(CELL_WITH_VERSIONS, Value.create(data, 100L)),
-                entry(CELL_WITH_SAME_ROW, Value.create(data, 63L)),
-                entry(CELL_WITH_DIFFERENT_ROW, Value.create(data, 81L)));
+                entry(CELL_WITH_VERSIONS, Value.create(data, 250L)),
+                entry(CELL_WITH_SAME_ROW, Value.create(data, 200L)));
+    }
+
+    @Test
+    public void testGetRows_manyColumnRows() {
+        TableReference tableReference =
+                TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
+        keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
+
+        byte[] data = PtBytes.toBytes("data");
+
+        ImmutableListMultimap.Builder builder = ImmutableListMultimap.<Cell, Value>builder();
+
+        IntStream stream = IntStream.rangeClosed(1, 1000);
+
+        byte[] row1 = row(1);
+
+        ImmutableListMultimap<Cell, Value> tableValues = stream.mapToObj(
+                col -> Cell.create(row1, column(col)))
+                .collect(ImmutableListMultimap.toImmutableListMultimap(
+                        cell -> cell, cell -> Value.create(data, 1L)));
+
+        keyValueService.putWithTimestamps(tableReference, tableValues);
+
+        Map<Cell, Value> result = keyValueService.getRows(
+                tableReference,
+                ImmutableList.of(row1),
+                ColumnSelection.all(),
+                STARTING_ATLAS_TIMESTAMP - 1);
+
+        assertThat(result).isEqualTo(KeyedStream.stream(tableValues).collectToMap());
     }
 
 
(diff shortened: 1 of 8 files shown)


atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +301,54 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
         assertThat(results).doesNotContainKey(CELL);
     }
 
+    @Test
+    public void testGetRowsForDeletion() throws Exception {
```
felixdesouza: the deletion at this point isn't really important, in your test you have no deletion code anyway (not that I'm asking you to do so either), should name this along the lines of "testGetRowsDoesNotReadOldData" or something along those lines

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +301,54 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
...
+        Cell CELL_1 = Cell.create(PtBytes.toBytes("row1"), PtBytes.toBytes("column"));
+        Cell CELL_4 = Cell.create(PtBytes.toBytes("row1"), PtBytes.toBytes("column1"));
+        Cell CELL_5 = Cell.create(PtBytes.toBytes("row1"), PtBytes.toBytes("column2"));
+
+
+        Cell CELL_2 = Cell.create(PtBytes.toBytes("row2"), PtBytes.toBytes("column"));
+        Cell CELL_3 = Cell.create(PtBytes.toBytes("row3"), PtBytes.toBytes("column"));
+
+
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL, Value.create(data, 8L)));
```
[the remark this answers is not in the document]
felixdesouza: if it's alright to put it into one, then I think the latter style works quite nicely i.e. single multimap with `valueWithTimestamps` method.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +301,54 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
...
+            add(PtBytes.toBytes("row2"));
+            add(PtBytes.toBytes("row3"));
+            add(PtBytes.toBytes("row4"));
+            add(PtBytes.toBytes("row5"));
+
+        }};
+
+        Map<Cell, Value> result = keyValueService.getRows(tableReference, rows, ColumnSelection.all(), STARTING_ATLAS_TIMESTAMP - 1);
+
+        assertThat(result).containsKeys(CELL_1, CELL_2, CELL_3, CELL_4, CELL_5);
```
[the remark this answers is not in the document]
felixdesouza: if I've understood correctly this should verify that the rows don't contain old timestamps?

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +301,54 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
...
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_5, Value.create(moreData, 98L)));
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_1, Value.create(moreData, 90L)));
+
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_2, Value.create(moreData, 73L)));
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_2, Value.create(moreData, 75L)));
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_2, Value.create(moreData, 79L)));
+        keyValueService.putWithTimestamps(tableReference, ImmutableListMultimap.of(CELL_3, Value.create(moreData, 51L)));
+
+
+        Iterable<byte[]> rows = new ArrayList() {{
```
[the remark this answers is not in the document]
felixdesouza: or better still, in order to not have to scroll up to figure out what `row1` means, just use `CELL.getRowName`, or have the signature be `Iterable<byte[]> rowNames(Cell... cells)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +633,97 @@ public String toString() {
...
+
+            //todo refactor
+            for(Entry<ByteBuffer, List<ColumnOrSuperColumn>> cellsForRow: partialResult.entrySet()) {
+
+                List<ColumnOrSuperColumn> cells = cellsForRow.getValue();
+
+                if (!cells.isEmpty()) {
+                    ByteBuffer row = cellsForRow.getKey();
+                    result.getOrDefault(row, new ArrayList<>()).addAll(cells);
+                    query.add(new KeyPredicate()
```
felixdesouza: this feels oddly similar to `translateRowsToKeyPredicates`

felixdesouza: perhaps instead of mutating the `query`, how about just setting it at the bottom? it would be nice to just make a new query object each time but we probably don't want recursion here.. so need some mutation somewhere.

felixdesouza: that way, I *think* you can reuse `translateRowsToKeyPredicates`, although I may be mistaken

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +633,97 @@ public String toString() {
...
+                                    SafeArg.of("cells", query.size()),
+                                    LoggingArgs.tableRef(tableRef),
+                                    SafeArg.of("timestampClause", ""),
+                                    SafeArg.of("startTs", startTs),
+                                    SafeArg.of("host", CassandraLogHelper.host(host)));
+                        }
+
+                        Map<ByteBuffer, List<List<ColumnOrSuperColumn>>> results = wrappingQueryRunner.multiget_multislice(
+                                "getRows", client, tableRef, query, readConsistency);
+                        Map<ByteBuffer, List<ColumnOrSuperColumn>> aggregatedResults = Maps.transformValues(results,
```
felixdesouza: just return  `Maps.transformValues(...)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +633,97 @@ public String toString() {
...
+                        return "multiget_multislice(" + host + ", "
+                                + query.size() + " cells" + ")";
+                    }
+
+                }
+        );
+    }
+
+    //todo(Sudiksha): refactor | names
+    private static SlicePredicate getSlicePredicate(List<ColumnOrSuperColumn> columns) {
```
felixdesouza: at some point, you want to comment on the optimisation is that you're doing, it's easy to get lost in `CassandraKeyValueServiceImpl`. If you're able to encode that in your method names, even better, but might make sense to just comment/javadoc.

felixdesouza:
still gotta rebase onto develop, certs are broken on the build unfortunately.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraConstants.java
```diff
@@ -64,6 +64,8 @@
     // TODO (jkong): Review this limit, it seems like we are making very big requests to Cassandra even at this value
     public static final int DEFAULT_SINGLE_QUERY_LOAD_BATCH_LIMIT = 50_000;
 
+    public static final int DEFAULT_READ_LIMIT_PER_ROW = 200;
+
```
sudiksha27: To be modified

sudiksha27: Long.MAX_VAlUE

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +634,94 @@ public String toString() {
         }
     }
 
+    private Map<Cell, Value> fetchAllCellsForRows(final InetSocketAddress host,
```
jeremyk-91: This method is actually fetching the _values_ as well, not just the cells. I would name it something like `getAllCellsForRows`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +634,94 @@ public String toString() {
...
+
+    private static KeyPredicate keyPredicate(ByteBuffer row, SlicePredicate predicate) {
+        return new KeyPredicate().setKey(row).setPredicate(predicate);
+    }
+
+    private static SlicePredicate allPredicateWithLimit(int limit) {
+        return SlicePredicates.create(Range.ALL, Limit.of(limit));
+    }
+
+    private Map<ByteBuffer, List<ColumnOrSuperColumn>> fetchCellsForKeyPredicates(final InetSocketAddress host,
```
jeremyk-91: as above, I'd prefer `get` instead of `fetch` because we are getting the values. Strictly speaking it should be `getForCellsForKeyPredicates` but I find that a bit clumsy, so would be ok with just changing the first word.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -652,6 +634,94 @@ public String toString() {
...
+                                "getRows", client, tableRef, query, readConsistency);
+
+                        return Maps.transformValues(results,
+                                lists -> Lists.newArrayList(Iterables.concat(lists)));
+                    }
+
+                    @Override
+                    public String toString() {
+                        return "multiget_multislice(" + host + ", " + tableRef +
+                                + query.size() + " cells" + ")";
```
jeremyk-91: nit: `... + " cells)";` rather than two separate strings

jeremyk-91: And we probably want a comma before saying the number of cells. The idea is to get something like `multiget_multislice(host, table, 1000 cells)`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -290,6 +290,11 @@ default int rangesConcurrency() {
         return 32;
     }
 
+    @Value.Default
+    default int fetchReadLimitPerRow() {
+        return CassandraConstants.DEFAULT_READ_LIMIT_PER_ROW;
+    }
```
jeremyk-91: Let's move this to the runtime config, so it can be reloaded more easily.

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceIntegrationTest.java
```diff
@@ -300,6 +306,66 @@ public void deletionTakesPlaceAtFreshTimestamp() throws Exception {
...
+
+    @Test
+    public void testGetRows_manyColumnRows() {
+        TableReference tableReference =
+                TableReference.createFromFullyQualifiedName("test." + RandomStringUtils.randomAlphanumeric(16));
+        keyValueService.createTable(tableReference, AtlasDbConstants.GENERIC_TABLE_METADATA);
+
+        byte[] data = PtBytes.toBytes("data");
+
+        ImmutableListMultimap.Builder builder = ImmutableListMultimap.<Cell, Value>builder();
```
jeremyk-91:
```suggestion

```

It's unused.

PR #4592, opened 2020-02-21
[LW] Snapshot Updates

gmaretic:
**Goals (and why)**:
This PR reworks lock watch updates and registration, to avoid the awkward workflow of having to re-register lock watches every time we fall behind. In addition, lock watch registration is calculated in a similar way, removing the need for separate open locks and lock watch registered events and simplifying client side logic.

- `LockWatchStateUpdate` was completely reworked: there are now three types of updates
  - `Success`: if we are able to get all events from last known version, which includes the list of events
  - `Snapshot`: if we are not able to get events from last known version, which includes references of current watches and currently held locks that are watched
  - `Failed`: if we are unable to calculate the snapshot

Additionally, `OptionalLong.empty()` now really stands for unknown version, so the responses return a long instead (-1 means log is empty). 
- `LockWatchCreatedEvent` now includes references that were added (as already watched references are filtered out), as well as the set of all currently held locks matching those watches
- `LockWatchingServiceImpl` now tracks both lock watch references and ranges to better enable the above workflow
- Removed `LockWatchStateUpdater` as we moved away from that approach and required some effort to reflect the changes
- Similarly removed the implementation of `LockWatchTestingService` as it wasn't trivial to fix, but did not remove it altogether as it will be useful in the future.

Fixed existing tests, and added some new tests for the lock watch log. The added tests are not very exhaustive, so if you have concerns, open to elaborating them.

One point of contention is that it is possible (although quite unreasonable) that the creation and therefore logging of the lock watch created event fails. In this case, the lock watch will still be updated, but there will never be a corresponding event. Consequently, further lock watch creation requests will not log it either because the watches will be registered already. This is fine, as we can always request a snapshot though.

`LockEventLogImpl`?

Next week. I suggest you look at it first, and then we can pair on any questions, say Wednesday or earlier?


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/lock/watch/LockEventLogImplTest.java
```diff
@@ -0,0 +1,178 @@
...
+        LockWatchStateUpdate update = log.getLogDiff(OptionalLong.of(100L));
+
+        LockWatchStateUpdate.Snapshot snapshot = UpdateVisitors.assertSnapshot(update);
+        assertThat(snapshot.lastKnownVersion()).isEqualTo(-1L);
+        assertThat(snapshot.locked()).isEqualTo(ImmutableSet.of(DESCRIPTOR_2, DESCRIPTOR_3));
+        assertThat(snapshot.lockWatches()).containsExactly(entireTable);
+    }
+
+    @Test
+    public void snapshotReflectsEventsOcurringDuringOpenLocksCalculation() {
```
jeremyk-91: nice!

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchingServiceImplTest.java
```diff
@@ -105,6 +105,29 @@ public void registeringWatchWithWiderScopeLogsAlreadyWatchedLocksAgain() {
         assertLoggedEvents(expectedEvents);
     }
 
+    @Test
+    public void registeringWatchWithOverlappingScopeLogsAlreadyWatchedLocksInScopeAgain() {
```
jeremyk-91: yep, this is exactly what I was looking for. Thanks!

PR #4596, opened 2020-02-24
[LW] Conjure Undertow

gmaretic:
**Goals (and why)**:
Use CJR undertow for lock watch APIs

Did it in a pretty straightforward and lazy way, open to suggestions -- this was meant to get the ball rolling

None

1. Should the start transactions with lock watches just straight up replace the start transactions endpoint? Seems like there is no use in having both since we don't have to worry about backwards compatibility.
2. Is using `SafeLong`s sensible here? Or should I just use longs and hide the fact by using custom objects?
3. Haven't yet wired up the lock watching resource part, that will be a separate PR.
 
timelock-api.yml

It's not super urgent, but better early than late 💃


jkozlowski:
I can try to get to reviewing here tomorrow

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -33,7 +41,25 @@ types:
...
+          known version is not available anymore, we will attempt to return a snapshot update.
+      StartTransactionsWithWatchesResponse:
+        fields:
+          immutableTimestamp: LockImmutableTimestampResponse
+          timestamps: PartitionedTimestamps
+          lease: Lease
+          lockWatchUpdate: LockWatchStateUpdate
+      GetCommitTimestampResponse:
+        fields:
+          commitTimestamp: safelong
```
jeremyk-91: `safelong` actually only goes up to 2^53, so I don't think we should use it for (AtlasDB) timestamps.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -47,8 +73,44 @@ services:
...
+          indicative of poorly registered lock watches.
       leaderTime:
         http: POST /lt/{namespace}
         args:
           namespace: string
-        returns: LeaderTime
\ No newline at end of file
+        returns: LeaderTime
+      getCommitTimestampWithWatches:
+        http: POST /gctww/{namespace}
```
jeremyk-91: Not sure how much this came out in the RFC/discussions, but I think we might need some form of autobatching around this operation. Otherwise this will cause the load on TimeLock to increase considerably, since currently `getFreshTimestamps` used in committing is autobatched.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -47,8 +73,44 @@ services:
...
+      startTransactionsWithWatches:
+        http: POST /stww/{namespace}
+        args:
+          namespace: string
+          request: StartTransactionsWithWatchesRequest
+        returns: StartTransactionsWithWatchesResponse
+        docs: |
+          The returned lock watch state the state taken atomically along with issuing the start timestamps. A failed
+          update means the lock watch state changed too much while start timestamps were being issued, which is
+          indicative of poorly registered lock watches.
```
jeremyk-91: This is reasonable, but is there a reason we don't want to or can't use existing `startTransactions`? This could add some overhead if clients aren't using the watches, but I imagine that will be changing soon.

timelock-api/src/main/conjure/timelock-api.yml
gmaretic: There is a world where we decide that it's fine to limit timestamps to 2^53, but I wanted to check that we decided where that is not OK

timelock-api/src/main/conjure/timelock-api.yml
gmaretic: Makes sense, will make this endpoint batched and add autobatching to get commit timestamp

timelock-api/src/main/conjure/timelock-api.yml
j-baker: Yes, there should not need to be any new endpoint, here.

gmaretic:
Remaining part of this is in #4676

PR #4597, opened 2020-02-25
[LW] Synchronise Issuing Start Timestamps with Lock Event Log

gmaretic:
**Goals (and why)**:
We need to have a consistent view of lock watch state as start timestamps are issued. This avoids some convoluted race conditions that could impact correctness.

Pretty straightforward, mostly wiring: after locking the immutable timestamp lock we enter a synchronised block in the lock event log (ensuring nothing can get logged), take out the start timestamps and get the current log version. After exiting the synchronised block, we get the list of events between last known version and the version that came with the timestamps.

No additional testing since it's mostly wiring

Other than the perf concerns that we will test out, nothing really.

small

Wednesday?


jeremyk-91:
Wednesday sounds good as a deadline here.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchingService.java
```diff
@@ -27,7 +28,8 @@
 public interface LockWatchingService {
     void startWatching(LockWatchRequest locksToWatch);
     LockWatchStateUpdate getWatchStateUpdate(OptionalLong lastKnownVersion);
-
+    LockWatchStateUpdate getWatchStateUpdate(OptionalLong lastKnownVersion, long endVersion);
+    <T> ValueAndVersion<T> runTaskAndAtomicallyReturnVersion(Supplier<T> task);
```
jeremyk-91: maybe verbose, but would prefer `...WatchStateVersion` perhaps given that this is on the super-interface  `AsyncTimelockService`

PR #4598, opened 2020-02-25
Bump Atlas retry timeout to 15 seconds

Jolyon-S:
**Goals (and why)**:
PDS-111849 - caused by Timelock being down for >10 seconds during a leader election. Increasing this retry timeout to 15 seconds should vastly reduce the chance of there being a problem.

Changed hard-coded variable from 10 to 15 seconds

N/A

Verify that there are no other places where this limit needs to be changed.

`FastFailoverProxy`

Today

### DIFF
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
@@ -48,6 +48,10 @@ public abstract class ClientOptions {
     // Should not be reduced below 65 seconds to support workflows involving locking.
     static final Duration BLOCKING_READ_TIMEOUT = Duration.ofSeconds(65);
 
+    // Time limit for retrying requests which in turn retry 308s that have reached their 20 retry limit defined
+    // in conjure-java-runtime
+    static final Duration REDIRECT_RETRY_TIME_LIMIT = Duration.ofSeconds(15);
+
     // Under standard settings, throws after expected outages of 1/2 * 0.01 * (2^13 - 1) = 40.96 s
     private static final Duration STANDARD_BACKOFF_SLOT_SIZE = Duration.ofMillis(10);
     private static final int STANDARD_MAX_RETRIES = 13;
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/FastFailoverProxy.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/FastFailoverProxy.java
@@ -20,7 +20,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.time.Clock;
-import java.time.Duration;
 import java.time.Instant;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -41,7 +40,6 @@ import feign.RetryableException;
  * leader election is still taking place.
  */
 public final class FastFailoverProxy<T> extends AbstractInvocationHandler {
-    private static final Duration TIME_LIMIT = Duration.ofSeconds(15);
 
     private final Supplier<T> delegate;
     private final Clock clock;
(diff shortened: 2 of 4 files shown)


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/FastFailoverProxy.java
```diff
@@ -41,7 +41,7 @@
  * leader election is still taking place.
  */
 public final class FastFailoverProxy<T> extends AbstractInvocationHandler {
-    private static final Duration TIME_LIMIT = Duration.ofSeconds(10);
+    private static final Duration TIME_LIMIT = Duration.ofSeconds(15);
```
jkozlowski: Can we add a metric and alerting on when this deadline is exceeded?

changelog/@unreleased/pr-4598.v2.yml
```diff
@@ -0,0 +1,30 @@
...
+    `FastFailoverProxy`
+
+    **Priority (whenever / two weeks / yesterday)**:
+    Today
+
+    <!---
+    Please remember to:
+    - Add any necessary release notes (including breaking changes)
+    - Make sure the documentation is up to date for your change
+    --->
```
jeremyk-91: Probably not all of this in the changelog

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/FastFailoverProxy.java
felixdesouza: does something in `ClientOptions` need to be changed also? if so @jeremyk-91, we really need to centralise this stuff, otherwise it's too easy to break assumptions here

jeremyk-91: Adding metrics/alerting makes sense, though I would not block this on that given the issue recurred and caused another batch of errors today - as long as we're tracking it somewhere.

ClientOptions: good flag. I'm guessing you're worried on `SocketTimeoutException`? Those get wrapped and become retryable, so if we were facing that we would fire one off, get blocked for 10 seconds, and if that's bad because we haven't had 15 seconds fire off another one (which would again get blocked for 10 seconds). If that's the only case then that should be fine, though I might have missed something else.

jkozlowski: Don't want to block, but I'd like this to be the next change, and it's tracked. @Jolyon-S can you make sure we track this?

Jolyon-S: I will add an issue and make sure we track it to completion

Jolyon-S: @jeremyk-91 I can't see any reference to 10s in `ClientOptions` - is this in a related class?

felixdesouza: not 10s, but regards to timeouts in general and how they interplay. This will only retry when 308s occur i.e. you've been speaking to the non-leader or you're in the middle of a leader election as opposed to "we're not getting anything from the socket" for which we won't retry if that limit is reached.

At the very least, having this constant being in `ClientOptions` and a comment on what it represents and how it interplays with some of the other constants in that class would be good here.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java
```diff
@@ -48,6 +48,10 @@
     // Should not be reduced below 65 seconds to support workflows involving locking.
     static final Duration BLOCKING_READ_TIMEOUT = Duration.ofSeconds(65);
 
+    // Time limit for retrying requests which in turn retry 308s that have reached their 20 retry limit defined
+    // in conjure-java-runtime
+    static final Duration REDIRECT_RETRY_TIME_LIMIT = Duration.ofSeconds(15);
```
jeremyk-91: Unclear if this should go here, this feels like it is at a different level of abstraction (the other things here are actually configurations on the CJR client, while this one is not)

Jolyon-S: Given that we found that the retry limit we thought was possibly not the cause, I think we should talk about what we actually want to increase before merging anyway.

PR #4600, opened 2020-02-25
Make KeyValueServiceRuntimeConfig jackson discoverable

sudiksha27:
**Goals (and why)**:
KeyValueServiceRuntimeConfig must be added to the list of Jackson discoverable classes for successful serialization/ deserialization of/to AtlasDbRuntimeConfig.
This is the only way to tell Jackson that ImmutableCassandraKeyValueServiceRuntimeConfig is a concrete implementation of KeyValueServiceRuntimeConfig.

- KeyValueServiceRuntimeConfig added to io.dropwizard.jackson.Discoverable 

- Modifying an existing test to ensure deserialization is working.

- Does this change completely fix the bug?

- io.dropwizard.jackson.Discoverable 

As soon as possible


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfigsTest.java
```diff
@@ -115,8 +116,8 @@ public void canDeserializeRuntimeConfig() throws IOException {
 
     @Test
     public void canParseRuntimeDeprecatedConfigType() throws IOException {
-        CassandraKeyValueServiceRuntimeConfig config =
-                AtlasDbConfigs.OBJECT_MAPPER.readValue("type: CassandraKeyValueServiceRuntimeConfig",
-                        CassandraKeyValueServiceRuntimeConfig.class);
+        KeyValueServiceRuntimeConfig config =
```
felixdesouza: I think you want to still use the old form for this test since that's the deprecated version.

I would also change all the tests to deserialise to `KeyValueServiceRuntimeConfig.class` instead of the semi-concrete `CassandraKeyValueServiceRuntimeConfig` test to ensure we're doing the right thing everywhere

atlasdb-config/src/test/resources/runtime-config-block.yml
```diff
@@ -26,3 +26,7 @@ sweep:
 
 internalSchema:
   targetTransactionsSchemaVersion: 1
+
+keyValueService:
+  type: cassandra
+  fetchReadLimitPerRow: 1000
```
felixdesouza: I think you need a new line at the end of the file, things seem to scream generally

jkozlowski:
@sudiksha27 can you provide some context here? Why do we need to update a dropwizard SPI stuff here?

sudiksha27:
KeyValueServiceRuntimeConfig is an interface with semi-concrete implementation in CassandraKeyValueServiceRuntimeConfig.

The organization of modules is such that we cannot specify CassandraKVSRuntimeConfig as JsonSubType in KVSRuntimeConfig (circular dependency).

Jackson has to know that CassandraKVSRuntimeConfig is a subclass of KVSRuntimeConfig for successful deserialization and this information is registered by DiscoverableSubtypeResolver.

Code path - AtlasDbConfigs.OBJECT_MAPPER#setSubtypeResolver

Turns out KeyValueServiceRuntimeConfig is not being registered right now which is why the deserialization to AtlasDbRuntimeConfig#keyValueService is failing.

This change aims to fix that (and some test are modified to verify the fix).

PR #4604, opened 2020-02-26
[LW] Expose LW Info in TM

gmaretic:
**Goals (and why)**:
Expose an interface for users of atlasdb to get lock watch information (and register watches) about their transactions from the TM. This might still change, but is what we currently need.

This is just the interface

Might be we need something different, but we'll just change it if that's the case

small

ASAP pls, to unblock other work


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TransactionsLockWatchEvents.java
```diff
@@ -0,0 +1,80 @@
...
+ * Represents a condensed view of lock watch events occurring between some known version and a set of start transaction
+ * calls.
+ */
+public interface TransactionsLockWatchEvents {
+    <T> T accept(Visitor<T> visitor);
+
+    interface Visitor<T> {
+        T visit(Success success);
+        T visit(Failure failure);
+    }
```
jeremyk-91: These names are a bit close to the `LockWatchStateUpdate` ones, and a little confusingly there a `Snapshot` becomes here a `Failure`.  Maybe `Events` rather than `Success`, and something like `Snapshot` or `State` rather than `Failure`?

PR #4607, opened 2020-02-26
SweepStrategyManagers only load one table metadata on cache miss

mswintermeyer:
**Goals (and why)**:

Do not have to reload all tables' metadata (which could include a LOT of tables) every single time there is a cache miss.

Rather than have a supplier that recalculates the entire map each time, just get and cache each entry. But still add all initial tables at the beginning as a perf optimization.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
@@ -43,8 +43,7 @@ public class SweepStrategyManagers {
             if (cache.estimatedSize() == 0) {
                 cache.putAll(getSweepStrategies(kvs));
             }
-            SweepStrategy strategy = cache.get(tableRef);
-            return Preconditions.checkNotNull(strategy, "unknown table", SafeArg.of("tableRef", tableRef));
+            return cache.get(tableRef);
         };
     }
 
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
@@ -17,6 +17,8 @@ package com.palantir.atlasdb.transaction.impl;
 
 import java.util.Map;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
@@ -33,13 +35,15 @@ public class SweepStrategyManagers {
     }
 
     public static SweepStrategyManager createDefault(KeyValueService kvs) {
-        RecomputingSupplier<Map<TableReference, SweepStrategy>> supplier =
-                RecomputingSupplier.create(() -> getSweepStrategies(kvs));
+        // On a cache miss, load metadata only for the relevant table. Helpful when many dynamic tables.
+        LoadingCache<TableReference, SweepStrategy> cache = Caffeine.newBuilder()
+                .build(tableRef -> getSweepStrategy(kvs.getMetadataForTable(tableRef)));
         return tableRef -> {
-            SweepStrategy strategy = supplier.get().get(tableRef);
-            if (strategy == null) {
-                strategy = supplier.recompute().get(tableRef);
+            // Add all existing tables the first time this is called, to optimize for cases when mostly non-dynamic tables.
+            if (cache.estimatedSize() == 0) {
+                cache.putAll(getSweepStrategies(kvs));
             }
+            SweepStrategy strategy = cache.get(tableRef);
             return Preconditions.checkNotNull(strategy, "unknown table", SafeArg.of("tableRef", tableRef));
         };
     }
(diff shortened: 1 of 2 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -33,13 +35,15 @@ private SweepStrategyManagers() {
...
+                .build(tableRef -> getSweepStrategy(kvs.getMetadataForTable(tableRef)));
         return tableRef -> {
-            SweepStrategy strategy = supplier.get().get(tableRef);
-            if (strategy == null) {
-                strategy = supplier.recompute().get(tableRef);
+            // Add all existing tables the first time this is called, to optimize for cases when mostly non-dynamic tables.
+            if (cache.estimatedSize() == 0) {
+                cache.putAll(getSweepStrategies(kvs));
             }
+            SweepStrategy strategy = cache.get(tableRef);
```
tjwilson90: The way you wrote this, this will never return null.

mswintermeyer: It could be if we queried for a tableRef for a table that had already been deleted by the time Atlas initialized, right?

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -33,13 +35,15 @@ private SweepStrategyManagers() {
...
+        // On a cache miss, load metadata only for the relevant table. Helpful when many dynamic tables.
+        LoadingCache<TableReference, SweepStrategy> cache = Caffeine.newBuilder()
+                .build(tableRef -> getSweepStrategy(kvs.getMetadataForTable(tableRef)));
         return tableRef -> {
-            SweepStrategy strategy = supplier.get().get(tableRef);
-            if (strategy == null) {
-                strategy = supplier.recompute().get(tableRef);
+            // Add all existing tables the first time this is called, to optimize for cases when mostly non-dynamic tables.
+            if (cache.estimatedSize() == 0) {
+                cache.putAll(getSweepStrategies(kvs));
```
felixdesouza: in this initial case doesn't this degrade to the version on the left on initial load? or is that an acceptable compromise?

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
felixdesouza: `getSweepStrategy` never returns `null` though right? Similarly `TableMetadata.BYTES_HYDRATOR.hydrateFromBytes(tableMeta).getSweepStrategy()`.

changelog/@unreleased/pr-4607.v2.yml
```diff
@@ -0,0 +1,28 @@
+type: fix
+fix:
+  description: |-
+    SweepStrategyManagers only load one table metadata on cache miss
```
felixdesouza: please fix up the changelog

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
mswintermeyer: Oh yeah.. argh, I need coffee before I deal with support tickets in the morning

changelog/@unreleased/pr-4607.v2.yml
mswintermeyer: What fixing up do you want? What's the usual Atlas formatting for these changelogs?

felixdesouza: Just a short description of the change akin to a commit message if you will. This has copied the entire PR description which is usually just bloat.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
mswintermeyer: If you are asking if it's okay to load all tables a single time: Even before https://github.com/palantir/atlasdb/pull/4542, we loaded the sweep strategy for all tables initially, and hadn't heard complaints around perf. So my _guess_ is this is just a problem when dynamic tables get created often, and thus now we limit loading all tables to only happen once.

If your concern is instead about how we no longer clear deleted table entries from the cache after seeing a new table entry: I hope that's fine.

berler: There's still a potential problem here where if loading all the sweep strategies takes a long time, you could have many threads all hit this case (as the cache size would still be 0 while the first thread is still loading).

felixdesouza: hmm yeah, might we want to have a `Suppliers.memoize` on line 42?

mswintermeyer: If memoizing, we'd want that to happen outside of the lambda, right?

felixdesouza: oop I can't read lol, but yes

mswintermeyer: Pushed https://github.com/palantir/atlasdb/pull/4625

PR #4609, opened 2020-02-27
[Conjure Undertow, Part 2] Upgrade timelock dependency and inline

jeremyk-91:
**Goals (and why)**:
- Actually enable work done in #4545, see if we can get performance numbers

- Update product dependency to 0.137.0, please validate this is correct as part of this review
- Inline previous guarded sections on `HAVE_ROLLED_OUT_CONJURE_CHANGES_INTERNALLY`

- None, beyond existing manual test done on internal herby stack.

- Nothing in particular. Manual test was done on internal herby stack which IMO reduces the risk.

**Where should we start reviewing?**: build.gradles


gmaretic:
I don't think we want to do this and then immediately deprecate as we are wiring in start transactions with watches. I suggest we just get the dependency to 0.141.0 and use the start transactions with watches endpoint, and then get that on conjure undertow next week

j-baker:
Grgur, there should be no new API for start transactions with watches in the Conjure API?

jeremyk-91:
Old clients will understand and correctly deserialise start transaction responses even with the watches in a conjure world. There could be abit of perf overhead but I doubt it's significant

PR #4610, opened 2020-02-27
If a method returned by AwaitingLeadershipProxy returns ListenableFuture, the leadership check is also asynchronous

j-baker:
Checkpoint from pairing with Jeremy


atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
```diff
@@ -40,16 +45,30 @@ public CoalescingSupplier(Supplier<T> delegate) {
 
     @Override
     public T get() {
+        try {
+            return getAsync().get();
```
j-baker: a consequence of this is that if the get() method is used, we will still always execute on a thread pool.

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -50,8 +58,12 @@
 
     private static final Logger log = LoggerFactory.getLogger(AwaitingLeadershipProxy.class);
 
-    private static final long MAX_NO_QUORUM_RETRIES = 10;
+    private static final int MAX_NO_QUORUM_RETRIES = 10;
```
j-baker: This class should probably do with some better unit tests, now, before this merges.

atlasdb-commons/src/main/java/com/palantir/common/concurrent/CoalescingSupplier.java
j-baker: There's a refactor we could do where the old method still remains, but behaves differently. I don't know it's important, though.

atlasdb-commons/src/test/java/com/palantir/common/concurrent/CoalescingSupplierTest.java
```diff
@@ -35,25 +35,57 @@
...
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Object[] getParameters() {
+        return new Object[][] {
+                {"blocking", (Function<CoalescingSupplierTest, Integer>) test -> test.coalescing.get() },
+                {"async", (Function<CoalescingSupplierTest, Integer>) test -> unwrap(test.coalescing.getAsync()) }
+        };
+    }
+
+    private static <T> T unwrap(ListenableFuture<T> future) {
```
gmaretic: Extract this to some utility class, as it's effectively duplicated from ConjureTimelockResource, and also exists in ALP

atlasdb-commons/src/test/java/com/palantir/common/concurrent/CoalescingSupplierTest.java
```diff
@@ -35,25 +35,57 @@
...
     private final Supplier<Integer> delegate = mock(Supplier.class);
     private final FreezableSupplier freezableDelegate = new FreezableSupplier(delegate);
-    private final CoalescingSupplier<Integer> supplier = new CoalescingSupplier<>(freezableDelegate);
+    private final CoalescingSupplier<Integer> coalescing = new CoalescingSupplier<>(freezableDelegate);
+
+    private final Supplier<Integer> supplier;
+
+    public CoalescingSupplierTest(String name, Object parameter) {
+        Function<CoalescingSupplierTest, Integer> factory = (Function) parameter;
+        supplier = () -> factory.apply(this);
```
gmaretic: nice!

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java
```diff
@@ -128,7 +131,19 @@ public StartTransactionResponseV4 startTransactions(StartTransactionRequestV4 re
...
+
+    private static <T> T unwrapListenableFuture(ListenableFuture<T> future) {
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        } catch (ExecutionException e) {
+            Throwables.throwIfUncheckedException(e.getCause());
+            throw new RuntimeException(e.getCause());
```
gmaretic: Again as before (except this time we get the cause if we reach here, which doesn't seem an intentional dif)

PR #4612, opened 2020-02-27
[Async ALP] Make getLeadershipToken return ListenableFuture and chain it to avoid throwing

jeremyk-91:
Note: This PR is pointed at #4610, not at develop.

**Goals (and why)**:
- Normally if I call a method `blahAsync()` returning `ListenableFuture<T>` and the underlying call fails, I expect a failed future, not my async thing throwing an exception.

- Change `getLeadershipToken()` to use `ListenableFuture` and chain it accordingly.

- Added a test that would have thrown the `NCLE` directly in the implementation James and I paired on.
- Test failures happen in the base branch for the paxos consensus fast test.

- At points I just get the value of the future, but I think this makes sense, as when these occur on the async path they're part of future-transforms.

**Where should we start reviewing?**: AwaitingLeadershipProxy?

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
@@ -34,7 +34,6 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.reflect.AbstractInvocationHandler;
--- a/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
@@ -22,7 +22,9 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.time.Duration;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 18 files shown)


leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -225,18 +224,19 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
             return null;
         }
 
-        final LeadershipToken leadershipToken = getLeadershipToken();
+        ListenableFuture<LeadershipToken> leadershipToken = getLeadershipToken();
```
j-baker: I don't know you'd want this - it seems reasonable to throw if you're knowably not the leader

j-baker: but does add some complexity?

j-baker: Idk - it's not something i'd like to be part of the contract, if that makes sense

jeremyk-91: Yep, the main reason for this was that I didn't want to throw in the async case. I did consider the alternative of splitting here to decide whether to throw immediately or return immediateFailedFuture on whether we were on an async path - didn't really like either solution. No strong opinions, wouldn't mind changing if you think that works better.

It does seem that we should handle the exceptions here similar to how we get them in the `isStillLeading` part (that could just be that I know my round is not the latest, for instance), which I would imagine is throwing for sync and returning a failed future for async.

jeremyk-91:
Discussed offline: the original API should be fine, it is similar to Guava `AsyncFunction` in that throwing is allowed and has the same semantics as a `Futures.immediateFailedFuture(e)`.

PR #4614, opened 2020-02-28
[LW] Add LW to conjure start transactions

gmaretic:
**Goals (and why)**:
Start transaction requests must specify a last known version for lock watch state.

Pretty straightforward

No new tests

This is not API breaking, but will require clients to have a not yet released version of timelock. Generally fine, but need to make sure that the version dependency is correctly specified.

Let's decide either way today


timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -28,11 +36,13 @@ types:
           requestId: uuid
           requestorId: uuid
           numTransactions: integer
+          lastKnownVersion: LastKnownVersion
       ConjureStartTransactionsResponse:
         fields:
           immutableTimestamp: LockImmutableTimestampResponse
           timestamps: PartitionedTimestamps
           lease: Lease
+          lockWatchUpdate: LockWatchStateUpdate
```
jkozlowski: This is not a breaking change, you are adding an optional field. It's just that this will be hidden from conjure backcompat the way it is written

gmaretic: Yeah, the response is fine, but the request is the break

gmaretic:
Alternatively, we could make last known version to be `optional<safelong>`? This should still give us plenty of breathing room and should be correctly filled in if missing by timelock

timelock-api/src/main/conjure/timelock-api.yml
j-baker: the request parameter is _also_ optional, and so won't be a break either.

changelog/@unreleased/pr-4614.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Start transactions requests now return lock watch state information
```
jeremyk-91: Please add information about the new product dependency.

PR #4616, opened 2020-02-28
[LW] Batched get commit timestamps endpoint

gmaretic:
**Goals (and why)**:
We need to batch getting the commit timestamps, as before that got batched with fresh timestamp requests. This is the server side change

The implementation is pretty straightforward, also removed the old jersey stuff as it wasn't being used.

None

Did I miss anything obvious?

*.yml

Would like to get it in today, so that we can wire up the client side asap


lock-api-objects/src/main/java/com/palantir/lock/watch/CommitTimestampsWithWatches.java
```diff
@@ -20,15 +20,16 @@
...
+import com.palantir.timestamp.TimestampRange;
 
 @Value.Immutable
-@JsonSerialize(as = ImmutableTimestampWithWatches.class)
-@JsonDeserialize(as = ImmutableTimestampWithWatches.class)
-public interface TimestampWithWatches {
-    long timestamp();
+@JsonSerialize(as = ImmutableCommitTimestampsWithWatches.class)
+@JsonDeserialize(as = ImmutableCommitTimestampsWithWatches.class)
+public interface CommitTimestampsWithWatches {
```
felixdesouza: define this in conjure? seems like a fairly simple object no?

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -20,6 +20,14 @@ types:
       base-type: any
       external:
         java: com.palantir.lock.v2.Lease
+    CommitTimestampsWithWatches:
+      base-type: any
+      external:
+        java: com.palantir.lock.watch.CommitTimestampsWithWatches
+    Long:
```
felixdesouza: I believe you're going to need to rebase, @j-baker got a similar thing in with #4622

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -20,6 +20,14 @@ types:
       base-type: any
       external:
         java: com.palantir.lock.v2.Lease
+    CommitTimestampsWithWatches:
+      base-type: any
+      external:
+        java: com.palantir.lock.watch.CommitTimestampsWithWatches
```
felixdesouza: as above, it's probably just easier to define this object here instead of external, unless you have strong reservations as to why not.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceImpl.java
```diff
@@ -176,8 +176,10 @@ public StartTransactionResponseV5 startTransactionsWithWatches(StartTransactionR
     }
 
     @Override
-    public TimestampWithWatches getCommitTimestampWithWatches(OptionalLong lastKnownVersion) {
-        return TimestampWithWatches.of(getFreshTimestamp(), getWatchStateUpdate(lastKnownVersion));
+    public GetCommitTimestampsResponse getCommitTimestamps(int numTimestamps, OptionalLong lastKnownVersion) {
+        TimestampRange freshTimestamps = getFreshTimestamps(numTimestamps);
+        return GetCommitTimestampsResponse.of(freshTimestamps.getLowerBound(), freshTimestamps.getUpperBound(),
+                getWatchStateUpdate(lastKnownVersion));
```
felixdesouza:
```suggestion
        return GetCommitTimestampsResponse.of(
                freshTimestamps.getLowerBound(), 
                freshTimestamps.getUpperBound(),
                getWatchStateUpdate(lastKnownVersion));
```

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -20,8 +20,12 @@ types:
       base-type: any
       external:
         java: com.palantir.lock.v2.Lease
+    LockWatchStateUpdate:
```
felixdesouza: I think I missed this earlier, any reason to *not* use Conjure unions here? you should still get the visitors and everything that you want etc.

gmaretic: Discussed offline -- requires nontrivial refactoring since the update is used outside of timelock api

PR #4620, opened 2020-03-02
Refactor out an AsyncRetrier to allow for better testing

j-baker:
This is part of #4610 - figured it could be split off and made its own thing. Class that takes a supplier of listenable future and returns a future that retries it up to n times, with a delay between attempts.


leader-election-impl/src/main/java/com/palantir/leader/proxy/AsyncRetrier.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public ListenableFuture<T> retryUntilSatistfied(Supplier<ListenableFuture<T>> supplier) {
+        return retryUntilSatistfied(supplier, maxAttempts);
+    }
+
+    private ListenableFuture<T> retryUntilSatistfied(Supplier<ListenableFuture<T>> supplier, int retriesRemaining) {
+        return Futures.transformAsync(Futures.submitAsync(supplier::get, executionExecutor),
+                result -> {
+                    int newRetriesRemaining = retriesRemaining - 1;
+                    if (predicate.test(result) || newRetriesRemaining == 0) {
```
Jolyon-S: This `if` statement feels a bit odd: it exits when the predicate is satisfied (which makes sense), but also when you run out of retries. However, running out of retries will return the result even though it doesn't satisfy the predicate - meaning the name of the method `retryUntilSatisfied` is a little misleading, as it can return a result when the predicate is _not_ satisfied. I think this is a bit misleading.

j-baker: Renamed to execute()

PR #4621, opened 2020-03-02
Improve observability of DisruptorAutobatcher.

felixdesouza:
**Goals (and why)**:
It can be useful to disambiguate different instances of the autobatcher especially when the resources they communicate with have varying latencies i.e. in memory access vs rpc.

It's also useful to have tracing to see when batches are being slow etc. Doesn't seem immediately clear how to marry the two with internal tools just yet, but individually they both can be useful.

* `AutobatcherBuilder` takes in a `safeTag` option.
* `AutobatcherBuilder` takes in an `observability` option.

* simple test checking that the tags are actually present.

None, concern about cardinality lies in the specific usage of these classes.

Everywhere.

Soonish please.


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -98,12 +103,33 @@ private AutobatcherBuilder(Function<Integer, EventHandler<BatchElement<I, O>>> h
...
+        public AutobatcherBuilder<I, O> observability(Observability observabilityParam) {
+            this.observability = observabilityParam;
+            return this;
+        }
+
         public DisruptorAutobatcher<I, O> build() {
             Preconditions.checkArgument(purpose != null, "purpose must be provided");
             EventHandler<BatchElement<I, O>> handler = this.handlerFactory.apply(DEFAULT_BUFFER_SIZE);
 
+            EventHandler<BatchElement<I, O>> tracingHandler = (event, sequence, endOfBatch) -> {
```
Jolyon-S: I think it would be interesting to test this to make sure that this handler does the right thing (i.e. same behaviour as before except it has the new trace).

felixdesouza: addressed, see new commit.

atlasdb-autobatch/src/test/java/com/palantir/atlasdb/autobatch/TracingEventHandlerTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Mock
+    private EventHandler<BatchElement<Integer, Long>> delegate;
+
+    @Test
+    public void nonFlushesDoNotHaveTraces() throws Exception {
+        TracingEventHandler<Integer, Long> tracingHandler =
+                new TracingEventHandler<>(delegate, "test", Observability.SAMPLE);
+
+        tracingHandler.onEvent(new TestBatchElement(), 45, false);
```
Jolyon-S: Could you also add a pass-through test to assert that `tracingHandler.onEvent(x,y,z)` calls the correct delegate `handler`?

Jolyon-S: I'll leave the addition of this test at your discretion.

PR #4623, opened 2020-03-02
[Timelock Partitioning] Part 52: ManualBatchingLeaderPinger

felixdesouza:
**Goals (and why)**:
In `LEADER_PER_CLIENT` mode, each namespace has its own `LeaderElectionService` and at least one `AwaitingLeadershipProxy`, which means that each `LeaderElectionService` tries to ping nodes for the leader.

I had previously put in an autobatcher, but the batches were relatively small, and we were pinging far more than the default 20 times a second (ping rate 50ms). In addition there is no benefit in doing so.

Instead, you register your desire to check leadership, and then there is a single guava `Service` per host that will ping the node with all registered clients every `pingRate`. This is because there is no additional benefit to checking  more often but with a different subset of clients.

In the steady state, you're sending over the same amount of bytes - the set of clients you want to check leadership for - 20 times a second. Could consider doing something similar to the `BatchingPaxosLatestSequenceCache`, where instead you send some token that represents your view of the cache at a particular time, and save on the bytes transferred etc. 

Considering that the cache itself gets way more requests per second and we're still not seeing amazing perf just yet, it's probably worth looking there first.

* `ManualBatchingPingableLeader`
* `ManualBatchingPingableLeaderTest`
* `AutobatchingPingableLeaderFactory`
ASAP please.

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/ManualBatchingPingableLeader.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/ManualBatchingPingableLeader.java
@@ -38,6 +38,7 @@ import com.google.common.util.concurrent.FluentFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.paxos.LeaderPingResult;
 import com.palantir.paxos.LeaderPingResults;
 import com.palantir.paxos.LeaderPingerContext;
@@ -51,6 +52,7 @@ final class ManualBatchingPingableLeader extends AbstractScheduledService implem
     private final LeaderPingerContext<BatchPingableLeader> remoteClient;
     private final Duration leaderPingRate;
     private final Duration leaderPingResponseWait;
+    private final UUID nodeUuid;
 
     private final Map<Client, SettableFuture<Void>> hasProcessedFirstRequest = Maps.newConcurrentMap();
     private final AtomicReference<LastResult> lastResult = new AtomicReference<>();
@@ -64,6 +66,8 @@ final class ManualBatchingPingableLeader extends AbstractScheduledService implem
         this.remoteClient = remoteClient;
         this.leaderPingRate = leaderPingRate;
         this.leaderPingResponseWait = leaderPingResponseWait;
+        this.nodeUuid = nodeUuid;
+
         // TODO(fdesouza): for comparison whilst testing
         MetricName metricName = MetricName.builder()
                 .safeName("atlasdb.autobatcherMeter")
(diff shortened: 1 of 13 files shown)


felixdesouza:
@Jolyon-S thanks for the review, will address comments when @jeremyk-91 also reviews, too much context switching today 🤯

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/ManualBatchingPingableLeader.java
```diff
@@ -0,0 +1,183 @@
...
+            Instant after = Instant.now();
+
+            Duration pingDuration = Duration.between(before, after);
+            if (pingDuration.compareTo(leaderPingResponseWait) >= 0) {
+                log.info("Ping took more than ping response wait, any waiters will report that ping timed out",
+                        SafeArg.of("pingDuration", pingDuration),
+                        SafeArg.of("leaderPingResponseWait", leaderPingResponseWait));
+            }
+            LastResult newResult = ImmutableLastResult.of(after, clientsThisNodeIsTheLeaderFor);
+            this.lastResult.set(newResult);
```
jeremyk-91: checking: safe because of guarantees of the guava Service class, right?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/ManualBatchingPingableLeader.java
```diff
@@ -0,0 +1,183 @@
...
+import com.google.common.util.concurrent.SettableFuture;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.paxos.LeaderPingResult;
+import com.palantir.paxos.LeaderPingResults;
+import com.palantir.paxos.LeaderPingerContext;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.SharedTaggedMetricRegistries;
+
+final class ManualBatchingPingableLeader extends AbstractScheduledService implements ClientAwarePingableLeader {
```
jeremyk-91: I find this class a little confusing, perhaps because of the naming similarity to `Autobatchers` and co. In particular, if a given client is requested once, it will continue to be requested for the life of the object, while in most things that batch I would not expect the request to be sent after the first time it is processed.

The logic makes sense (as you do mention in the comments, the worry here is serialization cost - but I think it's worth a shot, we can test with an RC if you'd prefer).

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/ManualBatchingPingableLeaderTest.java
```diff
@@ -0,0 +1,174 @@
...
+                .isEqualTo(LeaderPingResults.pingReturnedTrue(HOST_UUID, HOST_AND_PORT));
+
+        Future<LeaderPingResult> clientWhoIsLedSecondRequest =
+                manualBatchingPingableLeader.ping(HOST_UUID, CLIENT_WHO_IS_LED);
+
+        assertThat(Futures.getUnchecked(clientWhoIsLedSecondRequest))
+                .isEqualTo(LeaderPingResults.pingReturnedTrue(HOST_UUID, HOST_AND_PORT));
+
+        verify(batchPingableLeader, only()).ping(ImmutableSet.of(CLIENT_WHO_IS_LED));
+    }
```
jeremyk-91: There's a test that I think is quite important that's missing here:
- `runOneIteration` will ping (and continue to ping) services that haven't been requested since the last iteration - this is the behaviour I was confused by, but given its usage I think would be good to codify it

PR #4625, opened 2020-03-02
SweepStrategyManagers first init is synchronous

mswintermeyer:
**Goals (and why)**:

Follow up to https://github.com/palantir/atlasdb/pull/4607, returning initialization perf to the state before https://github.com/palantir/atlasdb/pull/4542. This way we avoid a scenario where multiple queries all try to load all tables' metadata at the same time.


mswintermeyer:
Is a changelog here necessary? or not really so close to the earlier PR?

felixdesouza:
already been a release with the previous code in unfortunately

mswintermeyer:
I'm running into continual Atlas build infra flakes, so I'm going to force this through.. hope that's fine!

PR #4634, opened 2020-03-05
Fix old endpoint

j-baker:
Accidentally broke this old endpoint in refactors. It shouldn't be used (unused for over a year) but is being caught in some integration tests internally, so I'm fixing it. We can't just downgrade internally, because people need the new endpoints.


felixdesouza:
a shame, the `AsyncResponse` makes it impossible to see what types an endpoint is supposed to return :(

j-baker:
Good thing we're not using them anymore!

PR #4636, opened 2020-03-05
[LW] PrecommitCondition With Watches

gmaretic:
**Goals (and why)**:
Precommit conditions can now be specified in terms of a `CommitUpdate`, instead of just the commit timestamp.

Implementation changed quite a bit since this PR was first opened. This is the current view:

- Went with the least breaking change, by adding a new default method `throwIfConditionInvalid(CommitUpdate update)` to the `PreCommitCondition` interface. This lets us keep it as a functional interface and not break existing precommit conditions. The drawbacks of this approach are that it's slightly non-intuitive if we want a precommit condition that uses commit updates, and that whenever we check precommit conditions, we have to call both methods.
- Introduced `LockWatchService` (not to be confused with lock watching service -- ideas for a better name?) that is wired into snapshot transaction manager. Lock watch service extends `LockWatchManager`, but also gives a transaction its commit update. The purpose of this is to minimize the surface exposed to users, and it also enables us to keep the default method on the `TransactionManager` interface without returning a broken LWM.
- Implemented autobatching `BatchingCommitTimestampGetter`, which coalesces requests for commit timestamps, and updates the lock watch event cache with received updates.
- `LeasedLockToken` is now package private again, but added a method to the lock token interface that lets us verify if a token was derived from another. So we can check if a leased lock token has X for its server token, but not get its server token.
- `CommitUpdate` currently explicitly lists all locks that were taken out. The assumption here is that lock watches are reasonable and this number will not be too large.

Added a test for BatchingCommitTimestampGetter, do we need more?

Does the overall design make sense, did I miss anything?

`CommitUpdate.java`, `SnapshotTransaction.java`, `BatchingCommitTimestampGetter.java`

This week pls


felixdesouza:
Haven't fully gone through this just yet, but I would be hesitant to do the dev break in one fell swoop, live migrations relies pretty heavily on `PreCommitCondition` and I'd be worried what would happen if atlas is upgraded and live migrations is not. 

As annoying as it would be, I think  having an intermediate version would be useful/almost necessary.

It would also be interesting to run rev-api and see what the extent of the dev breaks are.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
```diff
@@ -37,4 +39,6 @@
      * @param lastKnownVersion exclusive start version to get events from
      */
     TransactionsLockWatchEvents getEventsForTransactions(Set<Long> startTimestamps, IdentifiedVersion lastKnownVersion);
+
+    CommitUpdate getCommitUpdate(long startTimestamp, UUID requestIdToExclude);
```
j-baker: Instead of this being a requestIdToExclude, please can this actually be a LockToken?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConditionAwareTransactionTask.java
```diff
@@ -15,6 +15,6 @@
  */
 package com.palantir.atlasdb.transaction.api;
 
-public interface ConditionAwareTransactionTask<T, C extends PreCommitCondition, E extends Exception> {
+public interface ConditionAwareTransactionTask<T, C extends PreCommitConditionWithWatches, E extends Exception> {
```
j-baker: is this backwards compatible? There's no point not breaking if it's not

j-baker: i feel like the type erasure will end up with C being a PrecommitConditioNWithWatches, which will definitely be a binary breaking change.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/PreCommitConditionWithWatches.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.api;
+
+import com.palantir.lock.watch.CommitUpdate;
+
+@FunctionalInterface
+public interface PreCommitConditionWithWatches {
```
j-baker: Please don't add 'WithWatches' etc everywhere in Atlas. Atlas has traditionally suffeered from a pattern where you have a Foo, then you add feature Bar so you have a FooWithBar, and then you add a Baz so you end up with a FooWithBarAndBaz and it all ends up being extremely confusing over time - it lacks abstraction.

Instead, please can we change PreCommitCondition so that it covers this behaviour?

I would suggest that we instead add the new method, have it have a default (empty) implementation, and then have the old method also have a default (empty) implementation. This gives us space to upgrade internal users.

j-baker: It does mean that it's no longer a functional interface - but that's a source and not binary breaking change.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -319,10 +302,7 @@ default boolean isInitialized() {
      */
     TimelockService getTimelockService();
 
-    @DoNotDelegate
-    default LockWatchManager getLockWatchManager() {
```
j-baker: what happened here?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -232,28 +231,12 @@ default boolean isInitialized() {
...
-     * This is the same as {@link #runTaskWithConditionWithRetry(Supplier, ConditionAwareTransactionTask)}, but instead
-     * takes in a Guava supplier. This is deprecated in favour of the aforementioned method.
-     *
-     * @deprecated use {@link #runTaskWithConditionWithRetry(Supplier, ConditionAwareTransactionTask)} instead.
-     * @see #runTaskWithConditionWithRetry(Supplier, ConditionAwareTransactionTask)
-     */
-    @DoNotDelegate
-    @Deprecated
-    @Timed
-    default <T, C extends PreCommitCondition, E extends Exception> T runTaskWithConditionWithRetry(
```
j-baker: We shouldn't be deleting methods from this interface!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1640,6 +1646,14 @@ private void commitWrites(TransactionService transactionService) {
         });
     }
 
+    private UUID getRequestId(LockToken token) {
+        if (token instanceof LeasedLockToken) {
```
j-baker: As discussed yesterday - please can we not do this, and make the .serverToken() call package private again (as it should always have been)?

j-baker: (we can chat about alternatives, but that method is very much designed to be opaque, so making it public and castable needs some extra refactoring if we're going to do it - instanceof checks should be a flag to you).

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -27,10 +32,15 @@
 
 public final class LockWatchManagerImpl implements LockWatchManager {
     private final NamespacedLockWatchingRpcClient lockWatchingRpcClient;
+    private final NamespacedConjureTimelockService timelock;
     private final LockWatchEventCache cache;
 
-    public LockWatchManagerImpl(NamespacedLockWatchingRpcClient lockWatchingRpcClient, LockWatchEventCache cache) {
+    public LockWatchManagerImpl(
```
j-baker: I think this interface is weird to be putting on TransactionManager - we've tried hard to shy away from exposing yet more internal implementation details of AtlasDB to the world.

At most, clients need access to registerWatches - the other two methods should remain somewhere where they aren't available. Maybe getEventsForTransactions and getCommitUpdate could be package private? Or otherwise in a different class, where only registerWatches is visible here?

gmaretic: The commit update can be hidden as it's only used by the TM internally, but we need some kind of interface for clients to register watches and get the watches for their transactions if they are to use this information to implement caching themselves.

I am open to suggestions on how to handle it better, but package private is not going to work and exposing it at the transaction level is even worse.

jkozlowski: Proposed solution here: https://github.com/palantir/atlasdb/pull/4799

jkozlowski:
Ok, @j-baker I think:

1. Should now have been addressed.
3. Should now have been addressed.
2. I don't quite understand the problem, so need to do a bit of digging.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/NoOpInternalLockWatchManager.java
```diff
@@ -19,16 +19,20 @@
...
 import com.palantir.lock.watch.LockWatchReferences;
 import com.palantir.lock.watch.NoOpLockWatchEventCache;
 import com.palantir.lock.watch.TransactionsLockWatchEvents;
 
 public final class NoOpInternalLockWatchManager implements InternalLockWatchManager {
-    public static final InternalLockWatchManager INSTANCE = new NoOpInternalLockWatchManager();
 
-    private NoOpInternalLockWatchManager() {
-        // ...
+    private final CommitUpdateGetter commitUpdateGetter;
```
j-baker, line 32: this seems wrong - surely it should just return a random response?

j-baker: ah, I see, you construct it elsewhere. I think you should just explicitly return empty here - it shouldn't need an actual commit update getter

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/InternalLockWatchManager.java
```diff
@@ -32,4 +34,6 @@
      */
     TransactionsLockWatchEvents getEventsForTransactions(Set<Long> startTimestamps,
             Optional<IdentifiedVersion> lastKnownVersion);
+
+    CommitUpdate getCommitUpdate(long startTimestamp, LockToken commitLocksToken);
```
j-baker, line 38: I'm kind of surprised that we'd need an extra type for this. Ideally I'd expect that the methods on this class be:

```java
IdentifiedVersion getVersion(long timestamp);
LockWatchEvents getLockWatchUpdates(Optional<IdentifiedVersion> from, Optional<IdentifiedVersion> to);
```

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -217,6 +217,9 @@
     @Idempotent
     void commit(TransactionService transactionService) throws TransactionFailedException;
 
+    @Idempotent
+    void commit(CommitRequest commitRequest) throws TransactionFailedException;
```
j-baker, line 221: this won't be the right interface, because users almost never call commit directly themselves - the transaction manager does it. If you want Transaction to be aware of this, I think it gets to be aware of the lock watches that are registered and accept the read ranges in a method.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -320,10 +319,12 @@ default boolean isInitialized() {
...
-    @DoNotDelegate
-    default LockWatchManager getLockWatchManager() {
-        return NoOpInternalLockWatchManager.INSTANCE;
-    }
+    /**
+     * Returns the lock watch manager used by this transaction manager.
+     *
+     * @return the lock watch manager service for this transaction manager
+     */
+    LockWatchManager getLockWatchManager();
```
j-baker, line 327: this is just a binary breaking change - will affect a number of consumers internally

jkozlowski:
Closing in favour of https://github.com/palantir/atlasdb/pull/4811

PR #4639, opened 2020-03-05
[Timelock Partitioning] Part 53: Latest sequence cache concurrent

felixdesouza:
**Goals (and why)**:
`BatchingPaxosLatestSequenceCache` is stateful, I correctly identified it as not thread-safe but gave an incorrect justification as to why. i.e. the claim was that it’s in an `Autobatcher` so it’s fine. But it’s actually launched in `PaxosQuorumChecker` *from* an `Autobatcher`, so it’s not fine. i.e. with `B` and `C` as followers, if `C` is slow, `Autobatcher` will have returned a result from itself and `B`, and is free to send another `PaxosQuorumChecker` call => two concurrent threads to a class that’s explicitly not thread safe.

A separate approach that I considered:
* Actually have another Autobatcher that we call instead of calling the cache directly. This would allow us to use `BatchingPaxosLatestSequenceCache` as is without any changes because then it would serialise calls into the cache (autobatcher has only one thread). I did not go with this approach because I want to try getting rid of threads generated in `PaxosQuorumChecker` that leverages async retrofit.

Current approach:
* Each cache key is timestamped (it's timestamped internally, just not exposed over the wire). This is a break, but this is fine, this hasn't gone out anywhere.
  * Multiple threads can come in and make requests, if they get back differing cache keys because updates have occurred in between, which do we pick? The timestamp is what orders them all.
  * Similarly still process all sequences returned, but merge them with what already exists in the map.
* Use concurrent collections for safe access
* Make it easier to expose the timestamp from the server side, see `AcceptorCacheImpl`

Existing tests pass, not sure how to go about the concurrent pieces since there aren't that many levers I can pull to force certain orderings.

Am I missing anything?

* `BatchingPaxosLatestSequenceCache`
* `AcceptorCacheImpl`

ASAP please.

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
@@ -50,9 +50,18 @@ final class BatchingPaxosLatestSequenceCache implements CoalescingRequestFunctio
 
     private final BatchPaxosAcceptor delegate;
 
+    // we accumulate all clients that we've seen so far such that if we need to invalidate the cache and request a new
+    // update, we receive a cache update that's consistent with the clients we've seen e.g. in the face of leader
+    // elections
     private final Set<Client> clientsSeenSoFar = Sets.newConcurrentHashSet();
 
+    // represents the latest cache digest that this client side cache has processed
     private final AtomicReference<TimestampedAcceptorCacheKey> latestCacheKey = new AtomicReference<>();
+
+    // each value refers to a materialised version of the cache in terms of namespaces and the latest sequence at that
+    // particular cache timestamp
+    // as an optimisation to avoid copying, we can share the same materialised view for a string of cache keys issued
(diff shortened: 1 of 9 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
```diff
@@ -16,92 +16,159 @@
...
+                        newCachedEntries.merge(client, paxosLong, BatchingPaxosLatestSequenceCache::max));
+
+        // for a *new* mapping, setting the cache key must happen *after* we've setup the mapping, so that concurrent
+        // clients will not reference an in-progress populating map which can be empty.
+        maybeSetNewCacheKey(newCacheKey);
+    }
+
+    private void maybeSetNewCacheKey(TimestampedAcceptorCacheKey newCacheKey) {
+        while (true) {
+            TimestampedAcceptorCacheKey current = cacheKey.get();
```
felixdesouza: should I have passed in the cache key that generated this request? I think it's probably fine, but might make sense

PR #4642, opened 2020-03-10
Allow easy migration from/to thorough sweep

j-baker:
This PR adds a new sweep strategy, thorough_migration.
It forces the lock checking of thorough sweep, whilst actually behaving
like conservative sweep.

In this way, to migrate to thorough sweep, you simply:

1. Make sure all nodes are running code that knows about thorough_migration (via e.g. schema version bump).
1. Change to thorough_migration.
2. Make sure all nodes are on the new version (via e.g. schema version
bump).
3. Change to thorough.

This can also be tested on a single stack. Here, we'd add a config point for the sweep strategy, and when initializing the schema object, pass it in. Provided one guarantees that one goes through the intermediate stage (via e.g. observing), this is a safe test.

Would also flag that the current method that we recommend for migrating to thorough is *wrong*. If you look at SnapshotTransaction.requiresImmutableTimestampLocking, you can see that it only applies the locking for tables that it thinks should be thoroughly swept, so the docs and internal procedures are incorrect and provide no safety guarantees https://github.com/palantir/atlasdb/pull/4544.


atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/ReadTransaction.java
```diff
@@ -127,7 +126,7 @@ public Transaction delegate() {
 
     private void checkTableName(TableReference tableRef) {
         SweepStrategy sweepStrategy = sweepStrategies.get(tableRef);
-        if (sweepStrategy == SweepStrategy.THOROUGH) {
+        if (sweepStrategy.mustCheckImmutableLockAfterReads()) {
             throw new SafeIllegalStateException(
                     "Cannot read from a table with a thorough sweep strategy in a read only transaction.");
         }
```
sudiksha27: Could we change the message to something like - "Can only read from a table with a no/conservative sweep strategy in a read only transaction." ?

j-baker:
This is kind of sad. If we had used protos, there'd have been no issues. The field would come out as null, and we default such cases typically to saying 'conservative' which is perfect.

In practice, this isn't the case. So, it will come out as null. On the transaction codepath this is absolutely fine, because the comparison is to thorough sweep, and so it will correctly say false.

But, there's a second codepath, in the targeted sweeper, where it compares to conservative sweep, and so on this codepath it will thorough sweep everything.

So, given there's a no-op migration, we should just doc to have a no-op migration before you do anything, too.

changelog/@unreleased/pr-4642.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Allow easy migration from/to thorough sweep
```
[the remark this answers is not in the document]
j-baker: done

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/ReadTransaction.java
j-baker: done

PR #4645, opened 2020-03-11
[PDS-113120] Block Enforcing (Client-Side) Lock Service

jeremyk-91:
**Goals (and why)**:
- The V2 lock service has a point of fragility: if you ever make a lock request that blocks for longer than any underlying timeout we have in the remoting stack and can't acquire the lock, you'll get a runtime exception.
- This was addressed with the `BlockingTimeoutException` mechanism in V1, but remains an open bug in V2.

- Retry timed-out responses and SocketTimeoutExceptions up until the full duration of a user's blocking timeout, _as measured from the client side_.

- Added tests for the key retrying piece, and two previously failing integration tests.

- I went for a different design than in V1 because that created a lot of unpleasant coupling between remote timeout values and the actual lock service.
- I think there is no starvation risk even if we replay some requests, because lock requests have UUIDs and so retrying the *same* lock request *at this level* is fine (compare: retrying above from outside Atlas code is *not ok* because the request id will be different), but we should sanity check this.
- There is a bit of an edge case where we don't actually spend the full blocking timeout blocking on the server because of the mechanism I use to calculate timeouts.
- Will this cause more load on TimeLock than we like?
- I shoved WaitForLocksRequest through the LockLeaseService, otherwise we need to have the BlockEnforcing one in two places.

**Where should we start reviewing?**: RemoteTimeoutRetryer


lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
```diff
@@ -0,0 +1,164 @@
...
+    @Test
+    public void blockingDurationsDecreaseAsTimePasses() {
+        setupDefaultClock();
+        when(query.apply(any())).thenReturn(5);
+
+        retryForTenMillisAcceptingOnlyFortyTwo();
+
+        ArgumentCaptor<IntAndDuration> captor = ArgumentCaptor.forClass(IntAndDuration.class);
+        verify(query, times(3)).apply(captor.capture());
+        assertThat(captor.getAllValues()).containsExactly(
```
felixdesouza: this is a bit weird, I think if you have to have this, have something along the lines of:


```
List<Duration> durations = captor.getAllValues().stream().map(IntAndDuration::duration).collect(toList());
assertThat(durations)
  .as("duration is decreasing")
  .isSortedAccordingTo(Comparator.reverseOrder())
```

lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
```diff
@@ -0,0 +1,164 @@
...
+                ImmutableIntAndDuration.of(QUERY_VALUE, Duration.ofMillis(10)),
+                ImmutableIntAndDuration.of(QUERY_VALUE, Duration.ofMillis(6)),
+                ImmutableIntAndDuration.of(QUERY_VALUE, Duration.ofMillis(2)));
+        verifyNoMoreInteractions(query);
+    }
+
+    @Test
+    public void recoversFromFailureIfWeBecomeSuccessful() {
+        setupDefaultClock();
```
felixdesouza: the tests here are weird because there is nothing tying together a request timing out and the actual time, one is handled by the `setupDefaultClock` and the other is just if the query returns a correct response, as such it makes this example a bti contrived and hard to follow.

If anything, I'd probably have a method which was like: `returnResponseAfterDuration(T response, Duration duration)`. That sets up the clock and also the query mock.

```java
private static BoolAndDuration successfulAfter(Duration duration) { ... }
private static BoolAndDuration failedAfter(Duration duration) { ... }
private static void setupStuff(BoolAndDuration... durations) {
  // create arrays of the flag and stick it into the query mock
  // create arrays of the duration and stick it into the clock mock
}

public void test() {
  setupStuff(
    failedAfter(Duration.ofMillis(4)), 
    successfullAfter(Duration.ofMillis(5));

  bool result = retryForTenMillisWaitingForSuccess();
  ...
}
```

I'd probably go further and get rid of ints and just do this purely on booleans.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -193,4 +197,29 @@ public void canCreateNewClientsDynamically() {
...
+            LockResponse response = client.lock(LockRequest.of(LOCKS, 15_000));
+            assertThat(response.wasSuccessful()).isFalse();
+        } finally {
+            client.unlock(token);
+        }
+    }
+
+    @Test
+    public void waitForLocksRequestCanBlockForTheFullTimeout() {
+        // Test proxy timeout is 12.5 seconds, so 15 seconds suffices.
```
felixdesouza: bit obscure, is there a way we can refer to `ClientOptions.NON_BLOCKING_READ_TIMEOUT.plus(Duration.ofSeconds(1))` as the lock timeout here?

felixdesouza: as an aside, whilst trying to find that option, it's unclear to me why we have `clientQoS` enabled as a default, why is that?

felixdesouza:
Something else to think about: if we have this, we technically don't need the distinction between blocking and non-blocking anymore right? we can just do everything in terms of the non-blocking timeout as long as we have this retrying thing no? Although that's a separate PR of course.

felixdesouza:
also note: `waitForLocks` is the only one with a request id in the request that's accessible, so I think retrying of lock requests will result in separate request id's btw. Namely `ConjureLockRequests#toConjure(LockRequest)`, so you might want to parameterise that. 

Upon closer inspection, there's `ConjureLockRequests#toConjure(WaitForLocksRequest)` that actually ignores the `requestId` field from `WaitForLocksRequest` which is also bad.

lock-api/src/main/java/com/palantir/lock/client/RemoteTimeoutRetryer.java
```diff
@@ -0,0 +1,73 @@
...
+        Instant now = clock.instant();
+        Instant deadline = now.plus(durationExtractor.apply(request));
+
+        while (now.isBefore(deadline)) {
+            Duration remainingTime = Duration.between(now, deadline);
+            S durationLimitedInput = durationLimiter.apply(request, remainingTime);
+
+            try {
+                T response = query.apply(durationLimitedInput);
+                if (successfulResponseEvaluator.test(response)) {
```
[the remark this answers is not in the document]
jeremyk-91: I've moved this into an inner class: as discussed offline the predicate should really be `definitiveResponse` or something along those lines if we wanted it to be fully precise.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    static BlockEnforcingLockService create(NamespacedConjureTimelockService namespacedConjureTimelockService) {
+        return new BlockEnforcingLockService(namespacedConjureTimelockService, RemoteTimeoutRetryer.createDefault());
+    }
+
+    LockResponse lock(LockRequest request) {
+        return timeoutRetryer.attemptUntilTimeLimitOrException(
+                request,
+                req -> Duration.ofMillis(req.getAcquireTimeoutMs()),
```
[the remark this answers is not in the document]
jeremyk-91: I think I still want to keep the `Duration` as an input, just because we have a `BiFunction<S, Duration, S>` and I'm not really keen on having both duration and a millis-thing in the same API. I've changed this to not pass in an extractor though.

lock-api/src/main/java/com/palantir/lock/client/RemoteTimeoutRetryer.java
```diff
@@ -0,0 +1,73 @@
...
+            BiFunction<S, Duration, S> durationLimiter,
+            Function<S, T> query,
+            Predicate<T> successfulResponseEvaluator,
+            T defaultResponse) {
+        Instant now = clock.instant();
+        Instant deadline = now.plus(durationExtractor.apply(request));
+
+        while (now.isBefore(deadline)) {
+            Duration remainingTime = Duration.between(now, deadline);
+            S durationLimitedInput = durationLimiter.apply(request, remainingTime);
```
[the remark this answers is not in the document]
jeremyk-91: Changed to iteratively supply inputs (though no longer keeping it general).

lock-api/src/main/java/com/palantir/lock/client/RemoteTimeoutRetryer.java
```diff
@@ -0,0 +1,73 @@
...
+        return new RemoteTimeoutRetryer(Clock.systemUTC());
+    }
+
+    <S, T> T attemptUntilTimeLimitOrException(
+            S request,
+            Function<S, Duration> durationExtractor,
+            BiFunction<S, Duration, S> durationLimiter,
+            Function<S, T> query,
+            Predicate<T> successfulResponseEvaluator,
+            T defaultResponse) {
```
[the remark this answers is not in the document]
jeremyk-91: Discussed offline: We now save the response if the evaluator fails and return that.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: Yep, I'm OK with making that public.

ClientQoS: I don't know as well, other than that the normal default for that in most uses outside of Atlas should be enabled.

lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Test
+    public void returnsDefaultIfAlwaysFailing() {
+        setupDefaultClock();
+        when(query.apply(any())).thenReturn(5);
+
+        int result = retryForTenMillisAcceptingOnlyFortyTwo();
+        assertThat(result).isEqualTo(DEFAULT_VALUE);
+
+        verify(query, times(3)).apply(any(IntAndDuration.class));
```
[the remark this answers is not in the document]
jeremyk-91: The request is 10 ms, and the clock advances in steps of 4 ms, so ceil(10/4) = 3.

lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
jeremyk-91: I probably should rename the test. Not entirely happy with that assertion: `10, 9, 8` (or for that matter `10, 10, 10`) would pass when they don't.

lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
jeremyk-91: The BDD style thing could work. I think this is somewhat more complicated than what you've written above though: there are 4 cases, not 2 (retryable and non-retryable exceptions) so we can't just use `thenReturn`, there'll be some tricky conversions involved.

jeremyk-91:
Updated. The most important point of feedback needed is verification on the handling of lock request IDs.

I _think_ at a high level `BlockEnforcingLockServiceTest` is more readable now, though some of the underlying orchestration is nightmarish.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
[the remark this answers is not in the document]
felixdesouza: yeah, I assumed that anyway, strongly typed params 💪

lock-api/src/test/java/com/palantir/lock/client/RemoteTimeoutRetryerTest.java
felixdesouza: that's fair

felixdesouza:
Yeah didn't realise how involved the test change would be, namely the extra test cases. Appreciate you making the extra effort, it's clearer now! The only thing that remains in the lock request ids, it might be the case that you need the lock call to be in terms of `ConjureLockRequest` instead of `LockRequest`. I think in the past perhaps there was `IdentifiedLockRequest` used on the client side, but I'm not so sure.

Another test that would be in the ete's that I'm curious about: suppose we wanted to use the same request id, server side do we get back the same async lock? does the async lock keep track of the async response to write to? will it try to do it to all open connection requests that are for the same request id? what happens if that socket is closed etc. Important because depending on how we answer/repeat we could end up throwing/not responding etc.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -0,0 +1,168 @@
...
+        // The addition of a UUID takes place only at the Conjure level, so we must retry the same request.
+        return timeoutRetryer.attemptUntilTimeLimitOrException(
+                ConjureLockRequests.toConjure(request),
+                Duration.ofMillis(request.getAcquireTimeoutMs()),
+                BlockEnforcingLockService::clampLockRequestToDeadline,
+                this::performSingleLockRequest,
+                response -> !response.wasSuccessful());
+    }
+
+    WaitForLocksResponse waitForLocks(WaitForLocksRequest request) {
```
j-baker: waitForLocks should not need to be handled here

j-baker: actually, ignore, this is still better...

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -0,0 +1,168 @@
...
+                Duration.ofMillis(request.getAcquireTimeoutMs()),
+                BlockEnforcingLockService::clampLockRequestToDeadline,
+                this::performSingleWaitForLocksRequest,
+                response -> !response.wasSuccessful());
+    }
+
+    private static ConjureLockRequest clampLockRequestToDeadline(ConjureLockRequest request, Duration remainingTime) {
+        return ConjureLockRequest.builder()
+                .from(request)
+                .acquireTimeoutMs(Ints.saturatedCast(remainingTime.toMillis()))
```
j-baker: can we Ints.checkedCast this? This is only likely to be a long if we have a bug.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -0,0 +1,168 @@
...
+            while (now.isBefore(deadline)) {
+                Duration remainingTime = Duration.between(now, deadline);
+                currentRequest = durationLimiter.apply(currentRequest, remainingTime);
+
+                try {
+                    currentResponse = query.apply(currentRequest);
+                    if (!isTimedOutResponse.test(currentResponse)) {
+                        return currentResponse;
+                    }
+                } catch (Exception e) {
```
j-baker: can this be a RuntimeException? Should not see any checked exceptions

lock-api/src/main/java/com/palantir/lock/client/ConjureLockRequests.java
```diff
@@ -47,7 +47,7 @@ public static ConjureLockRequest toConjure(WaitForLocksRequest request) {
         return ConjureLockRequest.builder()
                 .lockDescriptors(toConjure(request.getLockDescriptors()))
                 .clientDescription(request.getClientDescription())
-                .requestId(UUID.randomUUID())
+                .requestId(request.getRequestId())
```
j-baker: is the line above also broken?

j-baker: answer - no

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
jeremyk-91: Sure, yeah, agree `checkedCast` is more appropriate.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
jeremyk-91: Yep, good point. Changed

PR #4646, opened 2020-03-11
[PDS-113496] Martin Lewis's Muse Part 1: Snapshot Transaction

jeremyk-91:
**Goals (and why)**:
- See internal ticket.
- Having fewer metrics saves money.

- Counters don't publish a 1m rate metric while meters do, but a lot of our 1m rate metrics are nto being used.
- Change cell filter based metrics and broad cells-read metrics to use counters, not meters.

- The cells read metric _is_ actually used, but you really should be using `per_second(count)` in internal infrastructure. It's true that that is different from the 1m rate, but I don't think the difference is material here.

**Where should we start reviewing?**: no preference, it's small

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -2375,13 +2375,6 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 metricsManager.getTableNameTagFor(tableRef));
     }
 
-    private Meter getMeter(String name, TableReference tableRef) {
-        return metricsManager.registerOrGetTaggedMeter(
-                SnapshotTransaction.class,
-                name,
-                metricsManager.getTableNameTagFor(tableRef));
-    }
-
     private Counter getCounter(String name, TableReference tableRef) {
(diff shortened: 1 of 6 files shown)


j-baker:
Just to make sure you know - our internal infra does not actually collect 1m rates for meters.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2380,4 +2381,11 @@ private Meter getMeter(String name, TableReference tableRef) {
                 name,
                 metricsManager.getTableNameTagFor(tableRef));
     }
+
+    private Counter getCounter(String name, TableReference tableRef) {
```
felixdesouza: `getMeter` should now be unused right? can we get rid of that?

jeremyk-91:
@j-baker we discussed offline: I thought so too, but the data in internal metrics framework clearly includes 1ms, and includes 1ms from relatively recent stacks

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: yep, removed `getMeter`.

PR #4648, opened 2020-03-13
Change conjure client replacement logic to replace ASAP 

j-baker:
Rebase of #4584


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -106,4 +108,24 @@ private AtlasDbHttpClients() {
...
+                Suppliers.memoizeWithExpiration(supplier::get, 20, TimeUnit.MINUTES),
+                AtlasDbHttpClients::isPossiblyOkHttpTimeoutBug);
+    }
+
+    private static boolean isPossiblyOkHttpTimeoutBug(Throwable throwable) {
+        if (throwable instanceof SocketTimeoutException) {
+            return true;
+        }
+        Throwable cause = throwable.getCause();
+        return cause != null && isPossiblyOkHttpTimeoutBug(throwable);
```
jkozlowski: I think you want isPossiblyOkHttpTimeoutBug(cause) otherwise it will SO.

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,70 @@
...
+            replaceIfNecessary(cause);
+            throw cause;
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private synchronized void replaceIfNecessary(Throwable thrown) {
+        if (shouldReplace.test(thrown)) {
+            log.info("Replacing underlying proxy due to thrown exception", thrown);
```
jkozlowski: Putting the log line here will basically print it for every request that fails with this, since shouldReplace will be true for all of them. So we either need to put it in the supplier or remove it. Sorry for confusing things

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,70 @@
...
+        this.delegate = delegateFactory.get();
+        this.shouldReplace = shouldReplace;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        try {
+            return method.invoke(delegate, args);
+        } catch (InvocationTargetException e) {
+            Throwable cause = e.getCause();
```
jkozlowski: Checked that getCause is the same as getTargetException.

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,70 @@
...
+            replaceIfNecessary(cause);
+            throw cause;
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException(e);
```
jkozlowski: Is handling this separately necessary?

jeremyk-91: We probably don't want to throw the cause in this case (seems like a bug in our reflection so we probably want to see it)

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
j-baker: yeah getTargetException is from the time before Java had getCause

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
j-baker: i fixed logging

jkozlowski: I checked in other places in this repo, and I don't really see it being handled so I'm curious.

j-baker: yeah, this was just an artifact of me removing 'throws Throwable', handling the cases and then realizing i needed to rethrow the cause

jkozlowski: Yeah, this is what's failing the build

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,71 @@
...
+        try {
+            return method.invoke(delegate, args);
+        } catch (InvocationTargetException e) {
+            Throwable cause = e.getCause();
+            replaceIfNecessary(cause);
+            throw cause;
+        }
+    }
+
+    private synchronized void replaceIfNecessary(Throwable thrown) {
```
carterkozak: Should we defer synchronization until the predicate has evaluated true?

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
```diff
@@ -0,0 +1,71 @@
...
+    private ReplaceIfExceptionMatchingProxy(Supplier<T> delegateFactory, Predicate<Throwable> shouldReplace) {
+        this.delegateFactory = delegateFactory;
+        this.delegate = delegateFactory.get();
+        this.shouldReplace = shouldReplace;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        try {
+            return method.invoke(delegate, args);
```
carterkozak: I think we need to compare the delegate that we actually invoked the method on, not the value after an exception is caught. Concurrent failures could cause us to roll several times.

jkozlowski: This is handled outside of this class. The supplier given here is memoized.

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
jkozlowski: Possibly, but it's probably ok to keep it simple here. @jeremyk-91

j-baker: doesn't matter - this is only used for RPCs, and they aren't expected to ever throw

carterkozak: :+1:

atlasdb-commons/src/main/java/com/palantir/common/proxy/ReplaceIfExceptionMatchingProxy.java
carterkozak: I don't understand, in that case wouldn't we always replace the delegate with itself? I just rolled out of bed though, and might not be reading thoroughly as I should...

I think we need something like:

```java
T localDelegate = delegate;
try {
  return method.invoke(localDelegate, args);
} catch (ITE e) {
  Throwable cause = e.getCause();
  replaceIfNecessary(cause, localDelegate);
  throw cause;
}
```
Then update `replaceIfNecessary`
```java
   private synchronized void replaceIfNecessary(Throwable thrown, T invocationTarget) {
        if (shouldReplace.test(thrown)) {
            // cas on fail
            if (invocationTarget != delegate) {
                // already replaced
                return;
            }
            T replacement = delegateFactory.get();
            if (delegate != replacement) {
                log.info("Replacing underlying proxy due to thrown exception", thrown);
                delegate = delegateFactory.get();
            }
        }
    }
```

jkozlowski: Ok so reasoning through this:

The supplier passed here is:

```
Suppliers.memoizeWithExpiration(supplier::get, 20, TimeUnit.MINUTES)
```

So when we hit the bug, all requests will enter the synchronized block and one by one, check they hit the bug, then call the factory. Only the first one will actually refresh the client (and the supplier will memoize it for the next 20 minutes). Next one will call the factory again and get the same one and so not replace it.

The only thing I'm wonder is why is the delegateFactory called twice

jkozlowski: Talked offline: the proposed code from @carterkozak is to make sure that if we ever switch the supplier passed to here, we don't shoot ourselves in the foot. I will keep as is, as we hope to remove this shortly.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/LeadersTest.java
```diff
@@ -116,19 +115,6 @@ public void createProxyAndLocalListCreatesSingletonListIfNoRemoteAddressesProvid
         verifyNoMoreInteractions(localAcceptor);
     }
 
-    @Test(expected = IllegalArgumentException.class)
```
jkozlowski: This is basically testing some code in Feign, so I'll say not worth it.

jeremyk-91: Yes. This looks like it originally sought to test that the AtlasDbFeignTargetFactory which doesn't exist anymore wouldn't do silly things

jeremyk-91: (When we had a lot of custom code there, but now we don't)

jkozlowski:
There are still failures in ETE tests CassandraMultinodeTestSuite and DbKvsTestSuite that are happening all the time, so need to figure that out.

PR #4651, opened 2020-03-13
Make #close safer.

jkozlowski:
**Goals (and why)**:

Currently any exception thrown from cleaning up TransactionManager resources will abort the running of closing callbacks.

1. I have removed the suppressed exception propagation, since we already log the individual exceptions.
2. I made the log messages more generic.
3. Extracted a small utility class to track the failure.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -327,24 +326,24 @@ public void registerClosingCallback(Runnable closingCallback) {
...
-            super.close();
-            cleaner.close();
-            keyValueService.close();
-            shutdownExecutor(deleteExecutor);
-            shutdownExecutor(getRangesExecutor);
-            closeLockServiceIfPossible();
-
-            List<Throwable> suppressedExceptions = new ArrayList<>();
+
+            ShutdownRunner shutdownRunner = new ShutdownRunner();
```
schlosna: curious if we should just use Guava `Closer` for this:

```
            try (Closer closer = Closer.create()) {
                closer.register(super::close);
                closer.register(cleaner::close);
                closer.register(keyValueService::close);
                closer.register(() -> shutdownExecutor(deleteExecutor));
                closer.register(() -> shutdownExecutor(getRangesExecutor));
                closer.register(this::closeLockServiceIfPossible);

                for (Runnable callback : Lists.reverse(closingCallbacks)) {
                    closer.register(callback::run);
                }

                closer.register(metricsManager::deregisterMetrics);

                log.info("Shutting down transaction manager");
            } catch (IOException e) {
                throw new SafeRuntimeException("Close failed.", e);
            }
```

jkozlowski: The way you implemented it it changes the semantics of the closing callbacks. From javadoc of #registerClosingCallback:

```
Registers a Runnable that will be run after the transaction manager is closed.
```

I am not a fan of those giant try-with-resources, and since the semantics are so funky (always run closing callbacks AFTER you've closed everything else), I think I prefer my solution.

jkozlowski: But I amended it to make the ShutdownRunner implement autocloseable, that is nice.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -326,26 +325,23 @@ public void registerClosingCallback(Runnable closingCallback) {
...
+            shutdownRunner.shutdownSafely(super::close);
+            shutdownRunner.shutdownSafely(cleaner::close);
+            shutdownRunner.shutdownSafely(keyValueService::close);
+            shutdownRunner.shutdownSafely(() -> shutdownExecutor(deleteExecutor));
+            shutdownRunner.shutdownSafely(() -> shutdownExecutor(getRangesExecutor));
+            shutdownRunner.shutdownSafely(this::closeLockServiceIfPossible);
+
             for (Runnable callback : Lists.reverse(closingCallbacks)) {
-                runShutdownCallbackSafely(callback).ifPresent(suppressedExceptions::add);
+                shutdownRunner.shutdownSafely(callback);
```
Jolyon-S: Given that there is a test for this class, would you be able to add any tests that verify this new behaviour (i.e. what happens when an exception is thrown during close)?

jkozlowski: I will see what tests I can add, although there's already a whole bunch of tests for SnapshotTransactionManager to do with #close behavior

Jolyon-S: If I am not mistaken, none of them handle errors during the close.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
felixdesouza: I believe you actually want `Closer`, which does this.

felixdesouza:
From an above comment, you want Guava's `Closer` I believe.

jkozlowski:
@felixdesouza I would still prefer my implementation, with Closer I'd need to revert the order here and it would be rather weird to read?

felixdesouza:
ahh, this is the actual `close`. It would make more sense to use `Closer` and register at resource creation time, and then `close` just calls `Closer#close`.

felixdesouza:
but sure, either works

jkozlowski:
So I could: use Closer for the KeyValueService and related things. But then to handle the callbacks, I'd still need custom code? Closer#close also throws an Exception, so I'd need to make sure I handle that correctly.

And if I wanted to use Closer for everything, I'd need to make sure this stuff happens in reverse order. as said above.

Unless there's an approach I'm missing, those 2 feel like overcomplicating things, this is a nice small type I've implemented to handle exactly what we want.

schlosna:
The benefits of try-with-resources or Guava `Closer` are that if more than one thing blows up during close, you get a single error log with the full stacktrace with all suppressed exceptions to understand the impact/potential causes.

jkozlowski:
Those log messages should be under the same traceId, so should not be a problem. I would argue that an exception with multiple suppressed exceptions will be harder to figure out than a bunch of disagreggated log messages.

jeremyk-91:
📖 0:20

I'd hold off a little over removing the stacked exceptions: large internal product has some legacy places where internal structured logging specification isn't used (so no traceIds), so untangling those could prove annoying.

No strong opinions on `Closer` vs. try with resources vs. `ShutdownRunner`, actually given the callbacks and weird non-close close operations I think this is probably easier.

Generally looks good, though note the first point.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
jkozlowski: Tests added, good shout.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -326,26 +326,23 @@ public void registerClosingCallback(Runnable closingCallback) {
...
-            shutdownExecutor(getRangesExecutor);
-            closeLockServiceIfPossible();
-
-            List<Throwable> suppressedExceptions = new ArrayList<>();
+        if (!isClosed.compareAndSet(false, true)) {
+            return;
+        }
+
+        try (ShutdownRunner shutdownRunner = new ShutdownRunner()) {
+            shutdownRunner.shutdownSafely(super::close);
```
jkozlowski: Diff here is much better in IntelliJ. The order should be exactly the same, just wrapped with shutdownSafely calls

PR #4655, opened 2020-03-19
Set conjure client configuration enable-http2

carterkozak:
**Goals (and why)**:

This provide a safety buffer to prevent behavior changes if the
default configuration values change.

Upgrade the minimal client-configuration conjure library rather than all CJR to avoid fighting the jakarta upgrade in this change.


versions.props
```diff
@@ -27,6 +27,7 @@ com.palantir.conjure:conjure = 4.7.0
 com.palantir.conjure.java.api:* = 2.9.0
 com.palantir.conjure.java:* = 5.2.0
 com.palantir.conjure.java.runtime:* = 4.43.0
+com.palantir.conjure.java.runtime:client-config:q = 4.70.1
```
felixdesouza: do you want `:q`?

jkozlowski:
what's jakarta upgrade?

carterkozak:
> what's jakarta upgrade?

renaming maven coordinates from `javax.*` to `jakarta.*`.

versions.props
```diff
@@ -27,6 +27,7 @@ com.palantir.conjure:conjure = 4.7.0
 com.palantir.conjure.java.api:* = 2.9.0
 com.palantir.conjure.java:* = 5.2.0
 com.palantir.conjure.java.runtime:* = 4.43.0
+com.palantir.conjure.java.runtime:client-config = 4.70.1
```
jeremyk-91: I guess we don't expect any inconsistencies with `4.43.0`. Doesn't really feel great to have this gap, though I guess this is to some extent our fault that we're 27+ versions behind 😅

Maybe hard question: are we aware of any large pitfalls in changing the line above to reference `4.70.1`?

jkozlowski: @jeremyk-91 I guess I should've written this down: upgrading the whole thing is a bit of work, since we need to change all usages of javax.* to jakarta.* in atlas. This is a bit of work apparently, but I'd really like us to just do it and then merge this PR cleanly.

carterkozak: If we upgrade all cjr libraries to use 4.70.1 we'll run into dependency issues due to the jakarta maven coordinate changes from `javax.*` to `jakarta.*`. The configuration itself is backwards compatible, so that shouldn't be an issue, and consumers of this library already depend on the newer version :-)

I'd recommend upgrading everything, but I want to make sure this change lands asap to prevent potential fallout from changes to the defaults.

jeremyk-91:
Cool, I think we ended up doing the full migration in #4677 (including the explicit setting of http2 to be enabled) so we can close this out. Thanks for your help @carterkozak!

PR #4656, opened 2020-03-19
Reduce retained memory in delete executor work queue

onursatici:
**Goals (and why)**:
reduce the retained size of the `deleteExecutor` work queue. Each runnable in `deleteExecutor` currently has a reference to `this`, which is a `SnapshotTransaction` and can be potentially huge.

- make `deleteCells`, the only method passed to `deleteExecutor::execute` in a lambda, be static so the lambda object will lose the `SnapshotTransaction` reference, reducing its retained memory size.

### DIFF
--- a/changelog/@unreleased/pr-4656.v2.yml
+++ b/changelog/@unreleased/pr-4656.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: |
+    The retained memory in delete executor work queue is reduced. Previously, each runnable in deleteExecutor had SnapshotTransaction reference, which could mean a lot of memory was unnecessarily retained.
+  links:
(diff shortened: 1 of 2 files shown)


changelog/@unreleased/pr-4656.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: |
+    The retained memory in delete executor work queue is reduced. Previously, each runnable in deleteExecutor had SnapshotTransaction reference, which can be potentially huge.
```
jeremyk-91, line 4: minor nits/technicalities: the reference to a SnapshotTransaction can't be huge, it's just the word size. Maybe instead something like `which could mean a lot of memory was unnecessarily retained`?

PR #4658, opened 2020-03-19
[PDS-111849] Trace Parts of AwaitingLeadershipProxy

jeremyk-91:
**Goals (and why)**:
Part of PDS-111849, and also there are some perf characteristics about TimeLock I'd be interested in understanding more about - this would be a first step.

Add some tracing to the leadership check and delegate execution pieces of ALP

None, is there a good way of testing this?

- Not much testing: is there a way to test I've set this up correctly?
- For the Async things, is the method I'm using correct? The intention is to trace how long the operation takes to get done, and not how long it takes to return the future.

**Where should we start reviewing?**: ALP, it's small


felixdesouza:
Have a look at https://github.com/palantir/tracing-java#tracing-test-utils if you want to be able to validate your changes, it's quite cool!

felixdesouza:
I imagine, as well as looking at the nice html report you get back, the golden master thing would be pretty useful if the `RenderTracingRule` is hooked up to `AwaitingLeadershipProxyTest`.

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -214,7 +218,8 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
         Object maybeValidDelegate = delegateRef.get();
 
         ListenableFuture<StillLeadingStatus> leadingFuture =
-                statusRetrier.execute(() -> leaderElectionService.isStillLeading(leadershipToken));
+                Tracers.wrapListenableFuture("validate-leadership",
```
felixdesouza: do we want one on each invocation that it retries?

felixdesouza: this would still remain as the overall trace for the `validate-leadership` portion, but I imagine that each invocation inside `statusRetrier` should be wrapped in one where it's like `validate-leadership attempt 0` etc.

leader-election-impl/src/test/java/com/palantir/leader/proxy/AwaitingLeadershipProxyTest.java
```diff
@@ -64,6 +65,9 @@
 
     @Rule public final ExpectedException expect = ExpectedException.none();
 
+    @Rule
+    public final RenderTracingRule rule = new RenderTracingRule();
```
felixdesouza: do you think it's worth doing the golden master thing? my proposal is that it could be, because this codepath is fragile, and we want to make sure that we're not undoing any async work as a result by accident (which is quite hard to test atm). Perhaps we don't need it for this PR, but it's something I've been thinking about with all the partitioning/perf work i.e. how do we avoid regressing things that aren't easily tested?

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -234,14 +239,21 @@ protected Object handleInvocation(Object proxy, Method method, Object[] args) th
...
             }
         } else {
             return FluentFuture.from(delegateFuture)
-                    .transformAsync(delegate -> (ListenableFuture<Object>) method.invoke(delegate, args),
+                    .transformAsync(delegate ->
+                                    Tracers.wrapListenableFuture("execute-on-delegate-async", () -> {
+                                        try {
+                                            return (ListenableFuture<Object>) method.invoke(delegate, args);
+                                        } catch (IllegalAccessException | InvocationTargetException e) {
+                                            return Futures.immediateFailedFuture(e);
```
felixdesouza: Do we not want to handle the `IAE` below? not that I think it'll come up that often, I don't think we handled it previously anyway, but seems a bit odd, and it'll come out as a 500 iirc.

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
felixdesouza: fair enough, not strictly necessary tbf

leader-election-impl/src/test/java/com/palantir/leader/proxy/AwaitingLeadershipProxyTest.java
felixdesouza: you needn't migrate big bang, that'd be way too much effort, could be a similar policy to assertj i.e. use it moving forward etc. or just for this rule if we think it has value

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
felixdesouza: unclear, if it's unhandled before, and we haven't really seen instances of it, I'll leave it up to you if you want to change it, it just caught my eye with the try-catch in the new diff but was surprised it was missing from the old diff, not fussed either way

leader-election-impl/src/test/java/com/palantir/leader/proxy/AwaitingLeadershipProxyTest.java
jeremyk-91: The issue was more bumping the Guava dependency, though we checked internally and determined that this was almost certainly acceptable.

jeremyk-91: Turns out this causes more problems than expected because e.g. of dev breaks within this codebase, so not proceeding with this right now.

jeremyk-91:
Thanks!

PR #4659, opened 2020-03-20
[Draft] Batching Txns

Jolyon-S:
**Goals (and why)**:
Start a batch of txns in one call.

Carved a hole from `SnapshotTransacstionManager` downwards to see what adding the option to start a batch of transactions looks like.

N/A.

Is this the right approach? Don't worry about details for now.

If at all, snapshotTransactionManager

No need to review, draft PR


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -433,6 +434,10 @@ default LockWatchManager getLockWatchManager() {
     @Timed
     TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
 
+    @Timed
+    List<TransactionAndImmutableTsLock> setupRunTaskWithConditionThrowOnConflictBatch(PreCommitCondition condition,
```
jkozlowski: I think we'd like to see the internal Skiing product's use of this.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -85,6 +88,18 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
         return result;
     }
 
+    public List<ListenableFuture<R>> applyBatch(int count) {
+        Preconditions.checkState(!closed, "Autobatcher is already shut down");
+        return IntStream.range(0, count).mapToObj($ -> {
+            SettableFuture<R> result = SettableFuture.create();
+            buffer.publishEvent((refresh, sequence) -> {
```
j-baker: is there a 'publishEvents' version of this, or otherwise a Disruptor mechanism of writing multiple events? Also, I don't think this makes sense as a primitive - null arguments are uncommon - rather I expect you'd want to pass a list of arguments (which in our case is a list of nulls).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ReadOnlyTransactionManager.java
```diff
@@ -161,6 +162,12 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
         throw new UnsupportedOperationException("Not supported on this transaction manager");
     }
 
+    @Override
+    public List<TransactionAndImmutableTsLock> setupRunTaskWithConditionThrowOnConflictBatch(
+            PreCommitCondition condition, int count) {
```
j-baker: as in the other case, the PreCommitCondition is likely (although not guaranteed) to be identical between all requests. So I'd recommend you instead pass a _list_ of PreCommitConditions.

jkozlowski: You absolutely need to pass a list of them.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -165,6 +167,21 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+    public List<TransactionAndImmutableTsLock> setupRunTaskWithConditionThrowOnConflictBatch(
+            List<PreCommitCondition> conditions) {
+        List<StartIdentifiedAtlasDbTransactionResponse> transactionResponse
+                = timelockService.startIdentifiedAtlasDbTransactionBatch(conditions.size());
+
+        return Streams.zip(transactionResponse.stream(), conditions.stream(),
+                (response, condition) -> wrap(condition, response)).collect(Collectors.toList());
+    }
+
+    private TransactionAndImmutableTsLock wrap(PreCommitCondition condition,
```
jkozlowski: You are not dealing with any exception here correctly. If #wrap fails for one of the transactions in the batch, you'll leak everything.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -79,6 +80,21 @@ StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
...
 
+    List<StartIdentifiedAtlasDbTransactionResponse> startIdentifiedAtlasDbTransactionBatch(int count) {
+        return autobatcher.applyBatch(
+                IntStream.range(0, count).mapToObj($ -> (Void) null).collect(Collectors.toList()))
+                .stream()
+                .map(result -> {
+                    try {
+                        return result.get();
+                    } catch (ExecutionException e) {
+                        throw Throwables.throwUncheckedException(e.getCause());
```
jkozlowski: Again, I think you can't just throw exceptions here: @gmaretic can correct me, but I think timelock can 1. choose to return you less transactions than you asked for. 2. Dunno if the disruptor can guarantee that all your requested number of transactions will end up in a single batch to timelock.

Therefore, I think you should assume that each of those futures can fail independently, therefore you need to decide how to handle that. At the very least you need to make sure that you do not leak resources: StartIdentifiedAtlasDbTransactionResponse contains the immutable timestamp lock (again someone more versed in this can explain what happens when that is leaked), which in the #wrap method is explicitly unlocked when something fails.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -85,6 +89,22 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
         return result;
     }
 
+    public List<ListenableFuture<R>> applyBatch(List<T> arguments) {
+        Preconditions.checkState(!closed, "Autobatcher is already shut down");
+        List<ListenableFuture<R>> results = new ArrayList<>();
+        buffer.publishEvents((EventTranslator<DefaultBatchElement<T, R>>[]) arguments.stream().map(argument -> {
```
Jolyon-S: I think it is possible that `buffer.publishEvents` does not call every one of its arguments; this is the only case that I can see where we would end up with less transactions than expected (in the sense that, elsewhere, we will have the right number of `Optional<Thing>`. It is possible that we may want it to return `Optionals` at even this level - i.e. if `results.size() < arguments.size()` or something akin to that.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -85,6 +89,22 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
+            SettableFuture<R> result = SettableFuture.create();
```
jkozlowski: DisruptorFuture, this code has moved on since you raised the PR

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -85,6 +89,22 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
+            EventTranslator<DefaultBatchElement<T, R>> translator = (refresh, sequence) -> {
+                refresh.result = result;
+                refresh.argument = argument;
+            };
+            results.add(result);
+            return translator;
+        }).collect(Collectors.toList()).toArray());
```
jkozlowski: This is a bit shady, have you written a test?

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -85,6 +89,22 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
+
```
jkozlowski: Probably something like this:

```
        buffer.publishEvents(arguments.stream().map(argument -> {
            DisruptorFuture<R> result = new DisruptorFuture<>(safeLoggablePurpose);
            EventTranslator<DefaultBatchElement<T, R>> translator = (refresh, sequence) -> {
                refresh.result = result;
                refresh.argument = argument;
            };
            results.add(result);
            return translator;
        }).toArray(EventTranslator[]::new));
```

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
Jolyon-S: No tests on anything yet, but can be remedied

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
jkozlowski: It should be pretty clear from documentation what this method does, having to return Optionals here would be wrong API for us.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
jkozlowski: You need to think about the tracing requirements here

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -165,6 +167,39 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+    @Override
+    public List<Optional<TransactionAndImmutableTsLock>> setupRunTaskWithConditionThrowOnConflictBatch(
+            List<PreCommitCondition> conditions) {
+        List<Optional<StartIdentifiedAtlasDbTransactionResponse>> transactionResponses
+                = timelockService.startIdentifiedAtlasDbTransactionBatch(conditions.size());
+
+        if (transactionResponses.stream().allMatch(response -> !response.isPresent())) {
+            // (todo) jshah - We may or may not want to throw in the case that all transactions failed to start;
+            //      to be decided
+            throw new RuntimeException("entire batch failed");
```
jkozlowski: It's pointless to have this in the API level down if you're just going to throw.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -79,6 +80,18 @@ StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
         }
     }
 
+    List<Optional<StartIdentifiedAtlasDbTransactionResponse>> startIdentifiedAtlasDbTransactionBatch(int count) {
+        return autobatcher.applyBatch(IntStream.range(0, count).mapToObj($ -> (Void) null).collect(Collectors.toList()))
```
jkozlowski: I vote that this be implemented to return List<StartIdentifiedAtlasDbTransactionResponse> and throw Atlas's TransactionFailedRetriableException if not the whole batch is happy.

Jolyon-S: The thing I was trying to avoid throughout is that you possibly increase the failure rate if you fail a whole batch based on possibly a single transaction's outcome; however, it may be that in general use we don't expect many transactions at all to fail to start, and therefore this is a reasonable assumption to make.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -165,6 +166,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
         StartIdentifiedAtlasDbTransactionResponse transactionResponse
                 = timelockService.startIdentifiedAtlasDbTransaction();
+        return wrap(condition, transactionResponse);
+    }
+
+    @Override
+    public List<TransactionAndImmutableTsLock> setupRunTaskWithConditionThrowOnConflictBatch(
+            List<PreCommitCondition> conditions) {
+        // now, this can throw if any fail, but the resources are cleaned up (no promises
+        // about the passed in conditions though
```
jkozlowski: I think we should make this do promises about conditions though, it feels like much better API.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
jkozlowski: I checked, and it does call. If you wanted it to fail if it can't do the whole batch, there's try* version of this method.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -79,6 +80,25 @@ StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
...
+                    .forEach(result -> {
+                        try {
+                            responses.add(result.get());
+                        } catch (Throwable t) {
+                            throw new RuntimeException(t);
+                        }
+                    });
+            return responses;
+        } catch (RuntimeException e) {
+            responses.forEach(response -> unlock(ImmutableSet.of(response.immutableTimestamp().getLock())));
```
jkozlowski: What if one of these throws?

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -79,6 +80,25 @@ StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
         }
     }
 
+    List<StartIdentifiedAtlasDbTransactionResponse> startIdentifiedAtlasDbTransactionBatch(int count) {
```
jkozlowski: There could be a world here where we return Optionals though: it feels almost as if the layer above is better placed to unlock these in an efficient manner (without requiring a request to timelock). But it sure is a rather leaky API

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -165,6 +166,34 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
     public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition) {
         StartIdentifiedAtlasDbTransactionResponse transactionResponse
                 = timelockService.startIdentifiedAtlasDbTransaction();
+        return wrap(condition, transactionResponse);
+    }
+
+    @Override
+    public List<TransactionAndImmutableTsLock> setupRunTaskWithConditionThrowOnConflictBatch(
```
jkozlowski: I would like to know if we could make TransactionAndImmutableTsLock have a #close method that correctly invokes the #finishMethod

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -318,11 +347,11 @@ public void registerClosingCallback(Runnable closingCallback) {
     /**
      * Frees resources used by this SnapshotTransactionManager, and invokes any callbacks registered to run on close.
      * This includes the cleaner, the key value service (and attendant thread pools), and possibly the lock service.
-     *
-     * Concurrency: If this method races with registerClosingCallback(closingCallback), then closingCallback
-     * may be called (but is not necessarily called). Callbacks registered before the invocation of close() are
-     * guaranteed to be executed (because we use a synchronized list) as long as no exceptions arise. If an exception
-     * arises, then no guarantees are made with regard to subsequent callbacks being executed.
+     * <p>
```
jkozlowski: Ideally no extraneous diff

Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/4711

PR #4661, opened 2020-03-23
Async Metrics

felixdesouza:
**Goals (and why)**:
As part of some of the conjure-undertow changes, async versions of requests were added. The sad part is, the metrics as we know currently are off since they measure the time it takes to be given back the `ListenableFuture` vs how long the `ListenableFuture` takes to complete.

Ideally this would go in https://github.com/palantir/tritium, but given that we've hacked together the `TaggedMetricsInvocationHandler` for our own (nefarious?) gains, I didn't want to unwind that change as I'm working on Timelock Partitioning etc. @schlosna for SA. My proposed impl would be that it goes in `ByteBuddyInstrumentationAdvice` or `InvocationEventProxy` then every `InvocationHandler` gets async metrics stuff for free!

* Backport some of the metrics changes in the `TaggedMetricsInvocationHandler` wrt metric name cache etc. Since the metric name doesn't change for most invocations we're wasting time doing the string concat and the allocations for `MetricName`. For the ones with the extra tags, have a specialised cache for those calls.
* Check if the return type is `ListenableFuture` and if so, add a `listener` that will update the metric. Otherwise we leave it as is.

* Hard to test properly since we can't change the future that gets returned, we can only act on it => we can have a race condition where if we were to wait for the future to be done, it's not necessarily the case that the metrics would have been tracked correctly. Only fix I see for this is in upstream in tritium.

* `TaggedMetricsEventInvocationHandler`
* `AtlasDbMetrics`

ASAP! Some metrics in our internal dashboard is a lie until fixed.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
@@ -75,12 +75,10 @@ public final class AtlasDbMetrics {
             TaggedMetricRegistry taggedMetrics,
             Class<T> serviceInterface,
             U service) {
-        return Instrumentation.builder(serviceInterface, service)
-                .withHandler(new TaggedMetricsInvocationEventHandler(
-                        taggedMetrics,
-                        MetricRegistry.name(serviceInterface)))
-                .withPerformanceTraceLogging()
-                .build();
+        return buildInstrumentation(
+                serviceInterface,
+                service,
+                new TaggedMetricsInvocationEventHandler(taggedMetrics, MetricRegistry.name(serviceInterface)));
     }
 
     public static <T, U extends T> T instrumentWithTaggedMetrics(
@@ -88,9 +86,21 @@ public final class AtlasDbMetrics {
             Class<T> serviceInterface,
             U service,
             Function<InvocationContext, Map<String, String>> tagFunction) {
-        String name = MetricRegistry.name(serviceInterface);
+        return buildInstrumentation(
+                serviceInterface,
+                service,
+                new TaggedMetricsInvocationEventHandler(
+                        taggedMetrics,
+                        MetricRegistry.name(serviceInterface),
+                        tagFunction));
+    }
+
+    private static <T, U extends T> T buildInstrumentation(
+            Class<T> serviceInterface,
+            U service,
+            TaggedMetricsInvocationEventHandler handler) {
         return Instrumentation.builder(serviceInterface, service)
-                .withHandler(new TaggedMetricsInvocationEventHandler(taggedMetrics, name, tagFunction))
+                .withHandler(handler)
                 .withPerformanceTraceLogging()
                 .build();
     }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
@@ -75,10 +75,12 @@ public final class AtlasDbMetrics {
             TaggedMetricRegistry taggedMetrics,
             Class<T> serviceInterface,
             U service) {
-        return buildInstrumentation(
-                serviceInterface,
-                service,
-                new TaggedMetricsInvocationEventHandler(taggedMetrics, MetricRegistry.name(serviceInterface)));
+        return Instrumentation.builder(serviceInterface, service)
+                .withHandler(new TaggedMetricsInvocationEventHandler(
+                        taggedMetrics,
+                        MetricRegistry.name(serviceInterface)))
+                .withPerformanceTraceLogging()
+                .build();
     }
 
     public static <T, U extends T> T instrumentWithTaggedMetrics(
@@ -86,21 +88,11 @@ public final class AtlasDbMetrics {
             Class<T> serviceInterface,
             U service,
             Function<InvocationContext, Map<String, String>> tagFunction) {
-        return buildInstrumentation(
-                serviceInterface,
-                service,
-                new TaggedMetricsInvocationEventHandler(
+        return Instrumentation.builder(serviceInterface, service)
+                .withHandler(new TaggedMetricsInvocationEventHandler(
                         taggedMetrics,
                         MetricRegistry.name(serviceInterface),
-                        tagFunction));
-    }
-
-    private static <T, U extends T> T buildInstrumentation(
-            Class<T> serviceInterface,
-            U service,
-            TaggedMetricsInvocationEventHandler handler) {
-        return Instrumentation.builder(serviceInterface, service)
-                .withHandler(handler)
+                        tagFunction))
                 .withPerformanceTraceLogging()
                 .build();
     }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
@@ -74,15 +74,24 @@ public final class AtlasDbMetrics {
     public static <T, U extends T> T instrumentWithTaggedMetrics(
             TaggedMetricRegistry taggedMetrics,
             Class<T> serviceInterface,
+            U service) {
+        return Instrumentation.builder(serviceInterface, service)
+                .withHandler(new TaggedMetricsInvocationEventHandler(
+                        taggedMetrics,
+                        MetricRegistry.name(serviceInterface)))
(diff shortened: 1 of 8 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
```diff
@@ -71,18 +71,27 @@ private AtlasDbMetrics() {}
...
+            Class<T> serviceInterface,
+            U service) {
+        return Instrumentation.builder(serviceInterface, service)
+                .withHandler(new TaggedMetricsInvocationEventHandler(
+                        taggedMetrics,
+                        MetricRegistry.name(serviceInterface)))
+                .withPerformanceTraceLogging()
+                .build();
+    }
+
```
Jolyon-S: This implementation is very nearly identical to the one below - is there a common method that can be extracted?

Also, below you factored out `String name = MetricRegistry.name(serviceInterface);`, but not in this block.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -48,7 +58,26 @@
...
+    private final ConcurrentMap<Method, Timer> untaggedTimerCache = Maps.newConcurrentMap();
+
+    @Nullable
+    private final Function<MethodWithExtraTags, Timer> onSuccessTimerMappingFunctionExtraTags;
+    private final ConcurrentMap<MethodWithExtraTags, Timer> extraTagsTimerCache = Maps.newConcurrentMap();
+
+    public TaggedMetricsInvocationEventHandler(TaggedMetricRegistry taggedMetricRegistry, String serviceName) {
+        super(InstrumentationUtils.getEnabledSupplier(serviceName));
+        this.taggedMetricRegistry = Preconditions.checkNotNull(taggedMetricRegistry, "metricRegistry");
+        this.serviceName = Preconditions.checkNotNull(serviceName, "serviceName");
```
Jolyon-S: Repeated code from below; is there a better way this can be done?

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -72,12 +106,54 @@ public void onSuccess(@Nullable InvocationContext context, @Nullable Object resu
             return;
         }
 
+        if (result != null && ListenableFuture.class.isAssignableFrom(context.getMethod().getReturnType())) {
+            Futures.addCallback((ListenableFuture<?>) result, new FutureCallback<Object>() {
```
Jolyon-S: It looks like you are casting `result` from `Object` to `ListenableFuture<?>`; can you do this without a check? (It looks to me like the check above is checking `context`'s return type, but I don't know if that is guaranteed to be the same as the type of `result`).

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -95,7 +171,6 @@ public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable ca
         taggedMetricRegistry.meter(MetricName.builder().safeName(
                 MetricRegistry.name(failuresMetricName, cause.getClass().getName())).build())
                 .mark();
-
```
Jolyon-S: nit: unnecessary diff?

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
```diff
@@ -91,6 +122,77 @@ public void instrumentWithCustomNameAll() {
...
+            // transformed future
+            Awaitility.await()
+                    .atMost(ASYNC_DURATION_TTL.toMillis(), TimeUnit.MILLISECONDS)
+                    .until(() -> taggedMetrics.timer(metricName).getSnapshot().size() > 0);
+            return true;
+        }, MoreExecutors.directExecutor());
+
+        assertThat(done.get()).isEqualTo(true);
+
+        assertThat(done.get()).isEqualTo(true);
```
Jolyon-S: Should there be two of the exact same statement here? If so, please may you leave a comment as to why?

nit: maybe also group them together w.r.t spacing?

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
```diff
@@ -91,6 +122,77 @@ public void instrumentWithCustomNameAll() {
...
+                .as("in the event of scheduling issues, and despite having 10 concurrent futures, we complete "
+                        + "everything with 2*ttl")
+                .isBefore(now.plus(ASYNC_DURATION_TTL.plus(ASYNC_DURATION_TTL)));
+        Snapshot snapshot = taggedMetrics.timer(metricName).getSnapshot();
+        assertThat(Duration.ofNanos(snapshot.getMin())).isGreaterThan(ASYNC_DURATION_TTL);
+        assertThat(snapshot.size()).isEqualTo(10);
+    }
+
+    @Test
+    public void instrumentTaggedAsyncFunctionWithExtraTags() throws InterruptedException, ExecutionException {
```
Jolyon-S: Copied code from above - can a common method be factored out?

atlasdb-client/src/main/java/com/palantir/atlasdb/util/AtlasDbMetrics.java
felixdesouza: can do, not really sure it buys anything though, this is what it looks like:

```java
    public static <T, U extends T> T instrumentWithTaggedMetrics(
            TaggedMetricRegistry taggedMetrics,
            Class<T> serviceInterface,
            U service) {
        return buildInstrumentation(
                serviceInterface,
                service,
                new TaggedMetricsInvocationEventHandler(taggedMetrics, MetricRegistry.name(serviceInterface)));
    }

    public static <T, U extends T> T instrumentWithTaggedMetrics(
            TaggedMetricRegistry taggedMetrics,
            Class<T> serviceInterface,
            U service,
            Function<InvocationContext, Map<String, String>> tagFunction) {
        return buildInstrumentation(
                serviceInterface,
                service,
                new TaggedMetricsInvocationEventHandler(
                        taggedMetrics,
                        MetricRegistry.name(serviceInterface),
                        tagFunction));
    }

    private static <T, U extends T> T buildInstrumentation(
            Class<T> serviceInterface,
            U service,
            TaggedMetricsInvocationEventHandler handler) {
        return Instrumentation.builder(serviceInterface, service)
                .withHandler(handler)
                .withPerformanceTraceLogging()
                .build();
    }
```

The factoring was so it would not kill my line count aha.

Jolyon-S: Yeah this isn't particularly pretty. Don't mind either way, so long as you are consistent (i.e. the factoring thing is either done or not done for both.

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
felixdesouza: deleted, can't remember, was likely an accident

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
felixdesouza: some code can be factored out, I won't go the full length because it's important to see what's actually in the test without having to jump around everywhere.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
felixdesouza: they should be equivalent, I suppose I wanted to adhere to the interface in question as opposed to whatever type we may get back, I can add another check to the object itself.

atlasdb-client/src/test/java/com/palantir/atlasdb/util/AtlasDbMetricsTest.java
Jolyon-S: That's fair, it just seemed that a fair amount was exact copy/paste (but I don't mind things with small variations not being factored out exactly).

PR #4662, opened 2020-03-23
[Timelock Partitioning] Part 54: Add namespace distribution tracker

felixdesouza:
**Goals (and why)**:
In order to validate any sort of balancing strategies that we might have/employ in the future, we need a way to introspect that the distribution is. Having this also can sometimes give a more detailed view as to why a leader might be performing a lot slower relative to its peers i.e. heavily loaded etc.

* Expose a metric that just emits how many namespaces a particular node is the leader for.
* Expose a tracker object that will be consumed by a health check in internal timelock. This will not only tell you how many namespaces a particular leader is leading, but it will also tell you which ones, which is a bit harder to grok in our internal metrics platform.

Class is fairly simple, didn't come up whilst I was iterating, but not sure there would be a valuable test.

None

* `MultiLeaderNamespaceDistributionTracker`

By the end of this week.

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
@@ -1,0 +1,55 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.paxos;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.immutables.value.Value;
+
+import com.palantir.timelock.paxos.NamespaceTracker;
+
+public final class MultiLeaderNamespaceDistributionTracker {
+    private final NamespaceTracker namespaceTracker;
+    private final BatchPingableLeader localBatchPingableLeader;
+
+    public MultiLeaderNamespaceDistributionTracker(
+            NamespaceTracker namespaceTracker,
+            BatchPingableLeader localBatchPingableLeader) {
+        this.namespaceTracker = namespaceTracker;
+        this.localBatchPingableLeader = localBatchPingableLeader;
+    }
+
+    public NamespaceDistribution getCurrentNodeNamespaceDistribution() {
+        Set<Client> trackedNamespaces = namespaceTracker.trackedNamespaces();
+        List<Client> clientsLedByThisNode = localBatchPingableLeader.ping(trackedNamespaces).stream()
+                .sorted(Comparator.comparing(Client::value))
+                .collect(Collectors.toList());
+        return ImmutableNamespaceDistribution.builder()
+                .addAllNamespacesLed(clientsLedByThisNode)
+                .numberOfTrackedNamespaces(trackedNamespaces.size())
+                .build();
+    }
+
+    @Value.Immutable
+    public interface NamespaceDistribution {
+        List<Client> namespacesLed();
+        int numberOfTrackedNamespaces();
+    }
+}
--- a/timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
@@ -28,6 +28,7 @@ import com.codahale.metrics.InstrumentedThreadFactory;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Suppliers;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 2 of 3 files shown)


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+    public NamespaceDistribution getCurrentNodeNamespaceDistribution() {
+        Set<Client> trackedNamespaces = namespaceTracker.trackedNamespaces();
+        List<Client> clientsLedByThisNode = localBatchPingableLeader.ping(trackedNamespaces).stream()
+                .sorted(Comparator.comparing(Client::value))
+                .collect(Collectors.toList());
+        return ImmutableNamespaceDistribution.builder()
+                .addAllNamespacesLed(clientsLedByThisNode)
+                .numberOfTrackedNamespaces(trackedNamespaces.size())
+                .build();
+    }
```
jeremyk-91: Do we want any tests for this class? Probably something on the number being correct and something around sorting.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Value.Immutable
+    public interface NamespaceDistribution {
```
jeremyk-91: This probably needs to be JSON serializable if it's going to be used as a param within internal health-checking framework.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+            NamespaceTracker namespaceTracker,
+            BatchPingableLeader localBatchPingableLeader) {
+        this.namespaceTracker = namespaceTracker;
+        this.localBatchPingableLeader = localBatchPingableLeader;
+    }
+
+    public NamespaceDistribution getCurrentNodeNamespaceDistribution() {
+        Set<Client> trackedNamespaces = namespaceTracker.trackedNamespaces();
+        List<Client> clientsLedByThisNode = localBatchPingableLeader.ping(trackedNamespaces).stream()
+                .sorted(Comparator.comparing(Client::value))
```
jeremyk-91: This makes sense! ❤️

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -171,6 +179,23 @@ private void createAndRegisterResources() {
                 this::createInvalidatingTimeLockServices,
                 Suppliers.compose(TimeLockRuntimeConfiguration::maxNumberOfClients, runtime::get));
 
+        if (install.paxos().leaderMode() == PaxosLeaderMode.LEADER_PER_CLIENT) {
+            LocalPaxosComponents paxosComponents = Preconditions.checkNotNull(
+                    paxosResources.leadershipBatchComponents().get(PaxosUseCase.LEADER_FOR_EACH_CLIENT),
+                    "Timelock setup has gone awfully wrong, this is a bug.");
+
```
sudiksha27: Could this check be moved to a public method in  paxosResources?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+                .addAllNamespacesLed(clientsLedByThisNode)
+                .numberOfTrackedNamespaces(trackedNamespaces.size())
+                .build();
+    }
+
+    @Value.Immutable
+    public interface NamespaceDistribution {
+        List<Client> namespacesLed();
+        int numberOfTrackedNamespaces();
+    }
```
sudiksha27: Why do we need numberOfTrackedNamespaces?

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.immutables.value.Value;
+
+import com.palantir.timelock.paxos.NamespaceTracker;
+
+public final class MultiLeaderNamespaceDistributionTracker {
```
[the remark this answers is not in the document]
felixdesouza: will add a comment

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/MultiLeaderNamespaceDistributionTracker.java
felixdesouza: so the difference between namespaces led and tracked namespaces is that tracked namespaces are all namespaces that via their service have hit this node at least *once*, whether or not this node is the leader is a given by `namespacesLed`. Again it's just a sanity check on are people hitting this node etc.  and a quick and easy way to see "I'm leading `n/m` namespaces".

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
felixdesouza: I don't believe so, in this portion of the code, I need to get a particular set of batch components, whereas in `PaxosResources` it will populate whatever config has been setup. So in any case, if I fetch from the map, I need to ensure that the entry exists.

Alternatively, I wanted this additiong to go inside `PaxosResources(Factory)` but that involved opening up too many holes/trying to satisfy dependency injection, so I gave up (circular dependency exists).

PR #4664, opened 2020-03-23
[Timelock Partitioning/Perf] Part 56/?? Execution environments

felixdesouza:
**Goals (and why)**:
To service a request in timelock, we spin up multiple threads (or at least transfer control to numerous active threads). This isn't free and definitely adds latency to the original request in question. 

It first started with a simple observation, in some places where local computation is done (i.e. no remote calls), we would spin up a thread, wait for it to be scheduled and then return. The calling thread in question i.e. in `PaxosQuorumChecker` would *not* be doing anything besides blocking!. The local computation should finish relatively quickly compared to its remote counter parts, so we avoid the cost of a context switch and put to good use a thread that would have blocked anyway.

In general, the number of context switches/changing of threads on the hot path, means that it has to contend with all the other requests that also consume threads and need to run that also start threads etc., such that things can become more sluggish that they need to.

Adding the ability to have a separate "execution environment" allows us to contrast and compare the different strategies i.e. "thread for each service" vs "thread for remote services but use local thread for local computation". Down the line, it also opens up the ability to get rid of spinning up threads if we're going fully async.

* Refactor `PaxosQuorumChecker` to be aware of an `ExecutionEnvironment`. `AllRequestsOnSeparateThreads` is the current impl.
* Add an execution environment that uses the current thread for local computation
* Wire this change in everywhere (this proved to be pretty beneficial when deployed).

Somewhat scarce, both prior to this PR and this PR itself. Happy to add some if people have ideas as to what would be useful to test.

Commit-by-commit is probably easier.

By the end of the week please.


gmaretic:
Reviewed this, looks good. Merge conflicts are far from trivial, so would be great if @jeremyk-91 can have another go after I resolve them

PR #4666, opened 2020-03-23
[Timelock Partitioning] Part 53b: Fix insiduous cache bug

felixdesouza:
**Goals (and why)**:
> NB: I'm very sad about this class now. Cost of making errors in this class is really quite high, and needs extra attention.

So a combination of errors: a removal listener in a cache and an atomic reference to the latest cache key PLUS loading cache with automatic loads result in a terrifying time.

If for whatever reason a cache key goes out of the cache, there is still a reference to it via the `latestCacheKey`, given that we get from the cache before every call, the entry might be phased out, and then we have an *empty* set of latest sequences per client 🔥 . Since there is a mismatch, we'll then request (with the same cache key) but with all the namespaces we've seen so far. 

> Which you'd think would get us out of this snag. But no.

Since that cache key itself is up to date, we'll get back an *empty* digest and never populate our map to be something consistent. Meaning any results we send back will be based on an empty map/default value of -1 sequence number (disastrous).

Scary part is that there is no further validation up the chain, not for leadership, and within timelock not for timestamps (although if timestamps go backwards, clients will stop themselves from progressing).

I'm genuinely quite sad about this class now, and wish I could go back to the single threaded approach (which I may be able to on the grpc branch, but unsure for the normal operation).

* Remove loading cache and require it to be explicit. If you want to put a value in the cache you have to either `put` or pass the `get` with a loader.
* The only place where you can instantiate an empty map, is when you're in the `populateNewCache` portion of code
* If an entry has been evicted, don't do anything with the null map, just start again, and also clear the `latestCacheKey` reference.

I'm at odds with how to test this class now, which is rather scary all things considered. Most of the bugs haven't been with the semantics but with the concurrency, and caching etc.

* Is there a useful test I can write?
* `AcceptorCacheImpl` also suffers from the same fate of a cache key going out of the cache, but the latest reference still referring to the key, (it's less dire here) but it does mean that before this change, you'd never get out of this loop until you'd done a prepare/accept and even then there's no guarantee that what state you calculate is consistent.
* Are there any extra guard rails we can add especially if we might be changing this code in the future?

`BatchingPaxosLatestSequenceCache`.

Should go in before we refactor main code to use batching logic. No real impact aside from internal testing stack since that's the only place running this code.

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/BatchingPaxosLatestSequenceCache.java
@@ -30,8 +30,8 @@ import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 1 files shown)


felixdesouza:
I think we almost certainly need something that gives us safety checks e.g. in the bound store and in the leadership, something like the `ErrorCheckingTimestampService` where are sequence numbers shouldn't go back in time. At the very least it should give us confidence to change these parts in the future if need be.

felixdesouza:
Regarding how I found it, I was running some tests locally, and then I realised that after a while the cached method wasn't being called. Obviously still haunted by the `CoalesingSupplier` bug, I Thought are we skipping them? But the non-cached version was being called instead. I thought this is also fine, but it turns out not because you call non-cached endpoint with the latest cache key, and obviously that's fine because it's up to date, and then you never update your digest to be non-empty, which is disastrous.

PR #4667, opened 2020-03-23
[Proof of Concept] TimeLock Partitioning: Single Leader Batched Paxos

jeremyk-91:
_More of a proof of concept: not intended for style review, or merging as is. If we decide this is the right approach, I can own breaking this into more manageable pieces and refactoring accordingly._

**Goals (and why)**:
- Allow testing of @felixdesouza's new smart clients for leadership checks, without necessarily introducing timelock partitioning entirely. **The method in this PR is sufficient for testing with a shutdown migration: it may NOT be suitable for rolling from existing single leader. Doing so may cause SEVERE DATA CORRUPTION™️ - need to check.  🔥**
- The idea is that we are able to evaluate a majority of the new features e.g. acceptor caches without committing to timelock partitioning. 
- It is also the goal, and I believe realistic, that we are able to get into this state without any downtime migrations (though no op schema bumps may be required). As mentioned what is written here as is requires a downtime migration!

- Use all the new infra, but still establish a single leader. The `PinningFoo` classes are versions of `BatchingFoo` except the factories always point you at `PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT`. Note that we can't change `BatchingFoo` because the behaviour of using whatever client is named is correct for timestamp bound Paxoses. There was copypasting here: we would probably refactor this more elegantly if we decide this approach is what we want to go with. I think this is still safe, since the use case is passed in as the legacy one.
- There is some nasty magic with `LegacyPaxos(Accepto|Learne)rShim` because of the way the JAX-RS algorithm works where it identifies which resource should be matched first, meaning that `LeadershipResource` would cause a lot of 404s.

I verified MNPTLSIT for single leader (and clients using single leader methods, which the testsuite does).

Not so much at this stage. Is this along the lines of what we were expecting?

**Where should we start reviewing?**: `PaxosResourcesFactory`

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
@@ -129,10 +129,10 @@ public final class PaxosResourcesFactory {
                 TimelockPaxosMetrics.of(PaxosUseCase.LEADER_FOR_ALL_CLIENTS, metrics.getTaggedRegistry());
 
         Factories.LeaderPingHealthCheckFactory healthCheckPingersFactory = dependencies -> {
-            PingableLeader local = dependencies.components().pingableLeader(PaxosUseCase.PSEUDO_LEADERSHIP_CLIENT);
-            List<PingableLeader> remotes = dependencies.remoteClients().nonBatchPingableLeaders();
+            BatchPingableLeader local = dependencies.components().batchPingableLeader();
+            List<BatchPingableLeader> remotes = dependencies.remoteClients().batchPingableLeaders();
             return Stream.concat(Stream.of(local), remotes.stream())
-                    .map(SingleLeaderHealthCheckPinger::new)
+                    .map(MultiLeaderHealthCheckPinger::new)
                     .collect(Collectors.toList());
         };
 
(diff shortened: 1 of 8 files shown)


jeremyk-91:
In the way of verification, fails only seem to be checkstyle.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PinningNetworkClientFactories.java
```diff
@@ -0,0 +1,89 @@
...
+    @Value.Auxiliary
+    @Value.Derived
+    @Override
+    public List<Closeable> closeables() {
+        return ImmutableList.of(acceptorNetworkClientFactory(), learnerNetworkClientFactory());
+    }
+
+    @Override
+    public Factory<PaxosAcceptorNetworkClient> acceptor() {
+        return $ -> metrics().instrument(
```
jeremyk-91: when implementing this, probably memoize: one instance across all clients would be enough

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
```diff
@@ -150,10 +150,18 @@ private static PaxosResources configureLeaderForAllClients(
```
jeremyk-91: the above network client factory builder and leader pinger factory builder should use the `PinningX`, not the `SingleLeaderX`.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
```diff
@@ -143,8 +143,8 @@ private static PaxosResources configureLeaderForAllClients(
                 .runtime(paxosRuntime)
                 .useCase(PaxosUseCase.LEADER_FOR_ALL_CLIENTS)
                 .metrics(timelockMetrics)
-                .networkClientFactoryBuilder(ImmutablePinningNetworkClientFactories.builder())
-                .leaderPingerFactoryBuilder(ImmutableBatchingLeaderPingerFactory.builder())
+                .networkClientFactoryBuilder(ImmutableSingleLeaderNetworkClientFactories.builder())
+                .leaderPingerFactoryBuilder(ImmutableSingleLeaderPingerFactory.builder())
```
jeremyk-91: 👎

jeremyk-91:
Closing out as @gmaretic and I have merged similar changes to this effect: this was an attempt to show how the end state should look like

PR #4671, opened 2020-03-25
Increase default buffer size to 4096

azarum:
After #4535 we still see LockSupport.park(1 nanosecond), trying to bump this buffer further


azarum:
Done from a non-fork here #4675

felixdesouza:
Might it be worth making this configurable? Since this will affect any batches on timelock, especially taking into account your partitioning refactor @jeremyk-91 ? Or at the very least, have a separate size for client side, and have timelock maintain a separate one if necessary.

PR #4672, opened 2020-03-25
[Timelock Partitioning/Perf] Reduce context switches

felixdesouza:
**Goals (and why)**:
Currently, as a request comes into timelock:
1. it heads into the `AwaitingLeadershipProxy` (`ALP`)
1. `ALP` then checks whether or not it is the leader
1. Since many requests will be coming in (under single leader) you run it through the `CoalescingSupplier` via `CoalescingPaxosLatestRoundVerifier`.
1. This then calls through to the `PaxosAcceptorNetworkClient`, when we're doing multi-leader, we're running through the `AutobatchingPaxosAcceptorNetworkClient`.
1. This is another batcher by the way of `DisruptorAutobatcher`.
1. Then it makes its way to the actual `BatchPaxosAcceptor`.

It doesn't make sense to batch twice, and in any case we pay for it in context switches, which has been killing our `server.response.p95` latency.

The first part (at least for multi-leader) stops us from going through the `CoalescingPaxosLatestRoundVerifier` and just straight to the batcher.

Inside `AwaitingLeadershipProxy` as part of a few async changes that occurred earlier, to check we were still the leader we now have an async retrier for when we get `NO_QUORUM`, however all attempts are immediately done on a separate executor than the calling thread. For the most part it will likely succeed on the first attempt, so we should at least try on the first attempt without switching.

We do as much work as we can on the direct executor and only farm it off when it's off the happy path and might take longer than usual etc.

* Skip `CoalescingPaxosLatestRoundVerifier` if we're running with the multi-leader batching logic.
* `AsyncRetrier` only executes stuff on a separate thread on the second retry, transforms within the same logical request will happen in the same thread.
* `ALP` tries to do as much work in the direct executor, and dispatches when we're no longer on the hot path.	

Ran the `AwaitingLeadershipProxyBenchmarks`, seemed to indicate an improvement, relative to their previous results. 

The `ALP` change really shines when you have a proper async network client, as the call chain to get to that network client happens in a single thread wit hno switches

Commit-by-commit

Whenever

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/proxy/AsyncRetrier.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/proxy/AsyncRetrier.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.ListeningScheduledExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
 
 /**
  * Class that uses ListenableFuture primitives to provide a simple delay-retry loop. Should be kept very simple.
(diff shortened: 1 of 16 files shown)


leader-election-impl/src/main/java/com/palantir/leader/proxy/AsyncRetrier.java
```diff
@@ -53,20 +54,33 @@
...
+    In a similar vein to reduce context switches, when the supplier::get call returns, the processing should be fairly
+    lightweight, so we should be able to do it in whatever thread ran/took control of the original supplier::get call,
+    hence the usage of directExecutor.
+
+    If we do have to retry, then we're likely "slow" enough that we should give up the initial calling thread and
+    offload the work onto a separate executor.
+     */
     private ListenableFuture<T> execute(Supplier<ListenableFuture<T>> supplier, int retriesRemaining) {
-        return Futures.transformAsync(Futures.submitAsync(supplier::get, executionExecutor),
+        return Futures.transformAsync(Futures.submitAsync(supplier::get, MoreExecutors.directExecutor()),
```
schlosna: nit: should we remove the inner `transformAsync` as it should be unnecessary if we're getting from the supplier on the current thread?
```suggestion
        return Futures.transformAsync(supplier.get(),
```

felixdesouza: A good shout, thanks!

felixdesouza:
This is from `AwaitingLeadershipProxyBenchmark` on develop on my work laptop:

```
Benchmark                                             Mode    Cnt     Score    Error  Units
AwaitingLeadershipProxyBenchmark.benchmarkAsync     sample    525    44.814 ±  0.893  us/op
AwaitingLeadershipProxyBenchmark.benchmarkBlocking  sample  79830  6430.013 ± 25.145  us/op
```

After these changes:
```
Benchmark                                             Mode     Cnt     Score   Error  Units
AwaitingLeadershipProxyBenchmark.benchmarkAsync     sample     896    26.570 ± 0.619  us/op
AwaitingLeadershipProxyBenchmark.benchmarkBlocking  sample  234451  2183.729 ± 2.761  us/op
```

The figure for async feels kinda weird, but relative to its previous result it's much better. Similarly blocking is closer to the 2ms limit (if I've understood the benchmark correctly) for how long an operation should take. In both of these, number of threads in the blocking case is 256.

leader-election-impl/src/main/java/com/palantir/leader/proxy/AsyncRetrier.java
felixdesouza: Had to wrap the supplier so we don't throw but return failed futures, but I think it's fine.

gmaretic:
This looks much cleaner, thanks!

PR #4673, opened 2020-03-25
Include extra tags for failure metrics

felixdesouza:
**Goals (and why)**:
If we've included extra tags, if for any reason they fail, it'd be nice to see the failures grouped by those extra tags as well.

Add tags to `MetricName` in the failure case.

Whenever, fairly tiny.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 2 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/TaggedMetricsInvocationEventHandler.java
```diff
@@ -169,9 +170,16 @@ public void onFailure(@Nullable InvocationContext context, @Nonnull Throwable ca
         }
 
         String failuresMetricName = InstrumentationUtils.getFailuresMetricName(context, serviceName);
-        taggedMetricRegistry.meter(MetricName.builder().safeName(failuresMetricName).build()).mark();
-        taggedMetricRegistry.meter(MetricName.builder().safeName(
-                MetricRegistry.name(failuresMetricName, cause.getClass().getName())).build())
+        Map<String, String> tags = tagFunction.map(f -> f.apply(context)).orElse(ImmutableMap.of());
```
Jolyon-S: `orElseGet(() -> ...` (as CircleCI kindly pointed out).

felixdesouza:
We *do* have them, they're bundled away in the metric name cache, whole issue is that when something goes wrong, you can't tell *where*/by-which-tag something had gone wrong, whereas I'm able to tell where something had gone "right" by virtue of success tags.

PR #4674, opened 2020-03-26
[Timelock Perf] Disruptor future

felixdesouza:
**Goals (and why)**:
Not immediately clear where time is spent when a task is submitted to an autobatcher. Add tracing to numerous parts.

Changed the future such that it tracks the different stages a task goes through.
When the future is created, it'll start a `DetachedSpan` so this will inherit the trace of the thread that *submitted* to the autobatcher, so we should have provenance from client request all the way to before the rpc (which can only have a single parent, but that's fine I suppose).

I removed starting new traces around the actual handler, as  I think this approach is easier to navigate with our tracing infra.

Used `RenderTracingRule` to verify what's going on.

`DisruptorAutobatcher`.

Whenever, would be nice to get it in fairly soon.

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -22,19 +22,24 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
+import javax.annotation.Nullable;
+
+import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.SettableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.lmax.disruptor.EventHandler;
 import com.lmax.disruptor.RingBuffer;
 import com.lmax.disruptor.TimeoutException;
 import com.lmax.disruptor.dsl.Disruptor;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.tracing.DetachedSpan;
 
 /**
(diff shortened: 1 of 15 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -98,19 +106,60 @@ public void close() {
...
         }
     }
 
+    public static final class DisruptorFuture<R> extends AbstractFuture<R> {
+
+        private final DetachedSpan parent;
+        private final DetachedSpan waitingSpan;
+
+        @Nullable
+        private DetachedSpan runningSpan = null;
```
jeremyk-91: Does this need to be volatile? Specifically I'm thinking of a race condition where one thread assigns `runningSpan`, but another cancels the future and runs the listener: thread 2 won't see thread 1's write. Maybe this is ok if GC is sufficient to clean out 'leaked' spans

atlasdb-autobatch/src/test/java/com/palantir/atlasdb/autobatch/TracingEventHandlerTest.java
```diff
@@ -16,107 +16,66 @@
...
 
     @Rule
-    public TraceCapturingRule traceRule = new TraceCapturingRule();
-
-    @Mock
-    private EventHandler<BatchElement<Integer, Long>> delegate;
-    private static final TestBatchElement ELEMENT = new TestBatchElement();
-
-    @Test
-    public void nonFlushesDoNotHaveTraces() throws Exception {
```
jeremyk-91: nicely done

changelog/@unreleased/pr-4674.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: Disruptor Autobatchers now has a special future that allows it track
+    the different phases of the batching process.
```
jeremyk-91: nit: I'd word this a bit more directly towards how users will consume the tracking.

```suggestion
  description: Disruptor Autobatchers now trace their calls. Time spent waiting for the
    batch to start and for the batch to be processed are also tracked in child spans.
```

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
felixdesouza: nah, this is only updated in the disruptor thread, so it's safe, see the `TracingEventHandler`.

PR #4676, opened 2020-03-26
[LW] Remove old client and resource and use undertow

gmaretic:
**Goals (and why)**:
Since this is not used yet, we can reimplement using conjure undertow

Basically the same as for conjure timelock.

No tests. Is there something we want to test? I will work on ETE tests next which will catch if there is something wrong in the wiring.

Is it correct, and did I miss removing something that's now unused?

`timelock-api.yml`

This week I hope


lock-api/src/main/java/com/palantir/lock/client/NamespacedConjureLockWatchingService.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final AuthHeader AUTH_HEADER = AuthHeader.valueOf("Bearer omitted");
+
+    private final ConjureLockWatchingService lockWatcher;
+    private final String namespace;
+
+    public NamespacedConjureLockWatchingService(ConjureLockWatchingService lockWatcher, String namespace) {
+        this.lockWatcher = lockWatcher;
+        this.namespace = namespace;
+    }
+
```
felixdesouza: nit: double space

timelock-agent/src/main/java/com/palantir/timelock/paxos/AsyncTimeLockServicesCreator.java
```diff
@@ -78,7 +77,6 @@ public TimeLockServices createTimeLockServices(
 
         AsyncTimelockResource asyncTimelockResource =
                 new AsyncTimelockResource(maybeEnhancedLockLog, asyncTimelockService);
-        LockWatchingResource lockWatchingResource = new LockWatchingResource(asyncTimelockService);
```
felixdesouza: ~not too sure about this, I think you still want stuff behind the `ATR`, but I'll think about it whilst goign through the rest of the PR.~

Turns out `ATR` extends `LWS` so it's all good.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -28,6 +28,10 @@ types:
       base-type: any
       external:
         java: java.lang.Long
+    LockWatchReference:
+      base-type: any
+      external:
+        java: com.palantir.lock.watch.LockWatchReferences.LockWatchReference
```
felixdesouza: not particularly fussed but not sure where we stand re implementing the union in conjure?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/ConjureLockWatchingResource.java
```diff
@@ -0,0 +1,78 @@
...
+    private final Function<String, AsyncTimelockService> timelockServices;
+
+    private ConjureLockWatchingResource(
+            RedirectRetryTargeter redirectRetryTargeter,
+            Function<String, AsyncTimelockService> timelockServices) {
+        this.exceptionHandler = new ConjureResourceExceptionHandler(redirectRetryTargeter);
+        this.timelockServices = timelockServices;
+    }
+
+    public static UndertowService undertow(
```
felixdesouza: sad that we're gonna have to do this for each resource :(

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -174,17 +174,16 @@ private void createAndRegisterResources() {
...
-                resource::getLockWatchingResource);
+
         registrar.accept(resource);
 
-        Function<String, AsyncTimelockService> serviceCreator =
-                namespace -> namespaces.get(namespace).getTimelockService();
+        Function<String, AsyncTimelockService> creator = namespace -> namespaces.get(namespace).getTimelockService();
         if (undertowRegistrar.isPresent()) {
-            undertowRegistrar.get().accept(ConjureTimelockResource.undertow(redirectRetryTargeter(), serviceCreator));
+            undertowRegistrar.get().accept(ConjureTimelockResource.undertow(redirectRetryTargeter(), creator));
```
felixdesouza: sad about this :( but I can't seem to think of a way without having some sort of conjure-undertow generation support for handling exceptions or having a proxy etc.

gmaretic: 😢

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/ConjureLockWatchingResource.java
gmaretic: It's a nuisance, yeah

timelock-api/src/main/conjure/timelock-api.yml
gmaretic: Yeah, would be nice, but LWReferences are already included in `LockWatchStateUpdate`, so that change would break the in-use start transaction endpoint 🐛

gmaretic:
@felixdesouza gtg?

PR #4679, opened 2020-03-26
Generate dialogue bindings for lock API

ferozco:
**Goals (and why)**:
Clear out prerequisites to using dialogue for client -> lock communications

Just some upgrade

Its all build time stuff, if we pass CI we're good

N/A


versions.props
```diff
@@ -50,6 +51,7 @@ io.netty:* = 4.1.39.Final
 javax.annotation:javax.annotation-api = 1.3.2
 javax.validation:validation-api = 1.1.0.Final
 javax.ws.rs:javax.ws.rs-api = 2.0.1
+jakarta.ws.rs:jakarta.ws.rs-api = 2.1.6
```
jkozlowski: Given this, let's just see what comes out of https://github.com/palantir/atlasdb/pull/4677

ferozco: I think we can safely merge this separately from #4677 since jakarta would only appear on the classpath of the dialogue jars which are currently not being referenced inside of the repo.

Otherwise, I'm happy to start stacking the PRs that actually adopt dialogue on top of this one

jkozlowski: Ok, that stuff is now cleaned up. I've merge develop let's see if build is green

versions.props
```diff
@@ -24,9 +24,10 @@ com.lmax:disruptor = 3.4.2
 com.netflix.feign:* = 8.18.0
 com.palantir.common:streams = 1.9.1
 com.palantir.config.crypto:encrypted-config-value-module = 2.1.0
-com.palantir.conjure:conjure = 4.7.0
+com.palantir.conjure:conjure = 4.10.1
```
jkozlowski: Changes here are just bumps, and explicit dependency on dialogue; @ferozco will excavator keep bumping the dialogue version or do we need to do something special?

ferozco: Dialogue will be automatically upgraded by excavator :)

PR #4680, opened 2020-03-27
[PDS-114184] Increase transaction read timeout in CLI to 8 hours

Jolyon-S:
**Goals (and why)**:
The validation step of a KVS migration is performed within a single transaction for a given table. For large tables, this step will fail as the transaction will timeout after one hour.

The aim here is to improve this by increasing the transaction timeout for the CLI.

* Overrides `TransactionReadTimeoutMillis` in the AtlasDBConfigs provided to the CLI

Not tested.

* My approach is to hard-code the value - may be more desirable to provide the ability to config (but in reality, how often will people configure this?)
* Should there be additional logging to map progress, so that people know the validation step is running successfully (but is just taking a long time)?

`KvsMigrationCommand`

Within a week.

### DIFF
--- a/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/KvsMigrationCommand.java
+++ b/atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/KvsMigrationCommand.java
@@ -27,6 +27,7 @@ import com.palantir.atlasdb.cli.output.OutputPrinter;
 import com.palantir.atlasdb.config.AtlasDbConfig;
 import com.palantir.atlasdb.config.AtlasDbConfigs;
 import com.palantir.atlasdb.config.AtlasDbRuntimeConfig;
+import com.palantir.atlasdb.config.ImmutableAtlasDbConfig;
 import com.palantir.atlasdb.schema.KeyValueServiceMigrator;
 import com.palantir.atlasdb.schema.KeyValueServiceValidator;
 import com.palantir.atlasdb.services.AtlasDbServices;
@@ -40,6 +41,7 @@ import io.airlift.airline.OptionType;
 @Command(name = "migrate", description = "Migrate your data from one key value service to another.")
 public class KvsMigrationCommand implements Callable<Integer> {
     private static final OutputPrinter printer = new OutputPrinter(LoggerFactory.getLogger(KvsMigrationCommand.class));
+    private static final int TRANSACTION_READ_TIMEOUT_MILLIS_OVERRIDE = 8 * 60 * 60 * 1000;
 
     @Option(name = {"-fc", "--fromConfig"},
             title = "CONFIG PATH",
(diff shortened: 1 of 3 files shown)


atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/KvsMigrationCommand.java
```diff
@@ -40,6 +41,7 @@
 @Command(name = "migrate", description = "Migrate your data from one key value service to another.")
 public class KvsMigrationCommand implements Callable<Integer> {
     private static final OutputPrinter printer = new OutputPrinter(LoggerFactory.getLogger(KvsMigrationCommand.class));
+    private static final int TRANSACTION_READ_TIMEOUT_MILLIS_OVERRIDE = 8 * 60 * 60 * 1000;
```
jkozlowski: Why is it hard to make the verification step use multiple transactions, same as migration.

Jolyon-S: https://github.com/palantir/atlasdb/pull/4681

atlasdb-cli/src/main/java/com/palantir/atlasdb/cli/command/KvsMigrationCommand.java
```diff
@@ -150,21 +152,29 @@ private AtlasDbConfig makeOfflineIfNecessary(AtlasDbConfig atlasDbConfig) {
...
+                        ? AtlasDbConfigs.load(toConfigFile, configRoot, AtlasDbConfig.class)
+                        : AtlasDbConfigs.loadFromString(inlineConfig, null, AtlasDbConfig.class));
         ServicesConfigModule scm = ServicesConfigModule.create(
                 makeOfflineIfNecessary(toConfig), AtlasDbRuntimeConfig.withSweepDisabled());
         return DaggerAtlasDbServices.builder().servicesConfigModule(scm).build();
     }
 
+    private AtlasDbConfig overrideTransactionTimeoutMillis(AtlasDbConfig config) {
+        return ImmutableAtlasDbConfig.builder().from(config)
+                .transactionReadTimeoutMillis(TRANSACTION_READ_TIMEOUT_MILLIS_OVERRIDE)
```
jkozlowski: If we're going to do it this way, we should probably also disable retries in the validator; if you hit a problem after a couple hours it doesn't really buy you much to retry, becuase nobody will sit there for that long.

changelog/@unreleased/pr-4680.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: '[PDS-114184] Increase transaction read timeout in CLI to 8 hours'
```
jkozlowski: Update changelog

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -115,7 +115,7 @@ private void validateTables(Set<TableReference> tables) {
     private void validateTable(final TableReference table) {
         final int limit = getBatchSize(table);
         // read only, but need to use a write tx in case the source table has SweepStrategy.THOROUGH
-        validationFromTransactionManager.runTaskWithRetry(
+        validationFromTransactionManager.runTaskThrowOnConflict(
```
jkozlowski: Put a comment why you don't want retryes

PR #4681, opened 2020-03-27
[PDS-114184] Make Atlas CLI Validate Step Use Many Transactions

Jolyon-S:
**Goals (and why)**:
CLI performs all validation for a given table in a single transaction. This PR aims to separate into many PRs instead. This is an alternative approach to https://github.com/palantir/atlasdb/pull/4680

* Perform each `getRange` in a new transaction.

N/A

* This will create lots of transactions (will this cause problems)?
* Correctness - I assume that validation is only ever run when DB is offline, so the DB state should not change between transactions, but would like confirmation.

`KeyValueServiceValidator`

Within a week


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -114,31 +113,27 @@ private void validateTables(Set<TableReference> tables) {
...
+            return validationFromTransactionManager.runTaskWithRetry(t1 ->
+                    validationToTransactionManager.runTaskWithRetry(t2 -> {
+                                RangeRequest range = RangeRequest.builder()
+                                        .batchHint(limit)
+                                        .startRowInclusive(nextRowName)
+                                        .build();
+                                return validateAndGetNextRowName(table, limit, t1, t2, range);
+                            }
+                    ));
+        } catch (Exception e) {
```
jkozlowski: Do you need this try-catch?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -114,31 +113,27 @@ private void validateTables(Set<TableReference> tables) {
 
     private void validateTable(final TableReference table) {
         final int limit = getBatchSize(table);
-        // read only, but need to use a write tx in case the source table has SweepStrategy.THOROUGH
```
jkozlowski: Can you keep this comment, it's pretty important.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -114,31 +113,27 @@ private void validateTables(Set<TableReference> tables) {
...
-        RangeRequest.Builder builder = RangeRequest.builder().batchHint(limit);
-        byte[] nextRowName = new byte[0];
-        while (nextRowName != null) {
-            RangeRequest range = builder.startRowInclusive(nextRowName).build();
-            nextRowName = validateAndGetNextRowName(table, limit, t1, t2, range);
+    private byte[] validateNextRow(TableReference table, int limit, byte[] nextRowName) {
+        try {
+            return validationFromTransactionManager.runTaskWithRetry(t1 ->
+                    validationToTransactionManager.runTaskWithRetry(t2 -> {
+                                RangeRequest range = RangeRequest.builder()
```
jkozlowski: Feels like creating this range should belong in validateAndGetNextRowName

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -114,31 +113,27 @@ private void validateTables(Set<TableReference> tables) {
...
-
-    private void validateTable(TableReference table, int limit, Transaction t1, Transaction t2) {
-        RangeRequest.Builder builder = RangeRequest.builder().batchHint(limit);
-        byte[] nextRowName = new byte[0];
-        while (nextRowName != null) {
-            RangeRequest range = builder.startRowInclusive(nextRowName).build();
-            nextRowName = validateAndGetNextRowName(table, limit, t1, t2, range);
+    private byte[] validateNextRow(TableReference table, int limit, byte[] nextRowName) {
+        try {
+            return validationFromTransactionManager.runTaskWithRetry(t1 ->
```
jeremyk-91: Please retain the comment on why this is runTaskWithRetry: it's non-obvious.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
```diff
@@ -114,31 +113,27 @@ private void validateTables(Set<TableReference> tables) {
...
-                });
-    }
-
-    private void validateTable(TableReference table, int limit, Transaction t1, Transaction t2) {
-        RangeRequest.Builder builder = RangeRequest.builder().batchHint(limit);
-        byte[] nextRowName = new byte[0];
-        while (nextRowName != null) {
-            RangeRequest range = builder.startRowInclusive(nextRowName).build();
-            nextRowName = validateAndGetNextRowName(table, limit, t1, t2, range);
+    private byte[] validateNextRow(TableReference table, int limit, byte[] nextRowName) {
```
jeremyk-91: nit: `validateNextBatchOfRows` / `startRow` for the param?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
jkozlowski: Actually, it's probably fine.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
jkozlowski: This change means that with default batch size, we'll do a transaction each 100 rows, so it will be probably quite slow.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
Jolyon-S: gradle complains if I don't

jkozlowski:
Closing for now to try https://github.com/palantir/atlasdb/pull/4680 first.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/schema/KeyValueServiceValidator.java
jkozlowski: This should be fine, most people override this anyway. And it's run with embedded timelock usually, so overhead should be minimal.

jkozlowski:
What sort of test coverage is there for this CLI?

PR #4685, opened 2020-03-30
[WIP] Use Dialogue for Timelock clients

ferozco:
**Goals (and why)**:
atlas currently relies on c-j-r clients, which use okhttp under the hood. We have plenty of complaints with c-j-r's stack of okhttp, feign and retrofit. Rather than investing in fixing the laundry list of problems with conjure-java-runtime, all effort is being directed into making dialogue the best possible client for palantir. Specifically, we're hoping for better observability, throughput, latency and wire-format correctness (see Dialogue rollout quip).

cc @carterkozak @iamdanfox . Happy to discuss here or in #dev-rpc.

dialogue contains many of the same ideas as c-j-r, but all fresh implementations, so we're probably going to see some subtle emergent effects/odd behaviour as a result of the new implementation (e.g. the queue and concurrency limiters are both slightly different). Only products with an appetite for risk should opt-in to this (ideally giving us a heads up in #dev-rpc).


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/BlockingSensitiveConjureTimelockService.java
```diff
@@ -37,12 +37,12 @@
  * Given two proxies to the same set of underlying TimeLock servers, one configured to expect longer-running operations
  * on the server and one configured not to, routes calls appropriately.
  */
-public final class BlockingSensitiveConjureTimelockService implements ConjureTimelockService {
-    private final ConjureTimelockService blocking;
-    private final ConjureTimelockService nonBlocking;
+public final class BlockingSensitiveConjureTimelockService implements ConjureTimelockServiceBlocking {
+    private final ConjureTimelockServiceBlocking blocking;
+    private final ConjureTimelockServiceBlocking nonBlocking;
```
iamdanfox: this type & variable name combo is fantastically misleading... IIRC the two fields have different timeouts?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DefaultTargetFactory.java
```diff
@@ -0,0 +1,139 @@
...
+import com.palantir.dialogue.hc4.ApacheHttpClientChannels;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import com.palantir.util.CachedTransformingSupplier;
+
+public enum DefaultTargetFactory implements TargetFactory {
+    INSTANCE;
+
+    private static final HostMetricsRegistry HOST_METRICS_REGISTRY = new HostMetricsRegistry();
```
iamdanfox: might as well just use a `NoOpHostEventsSink` if they're never wiring this up to anything...

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DefaultTargetFactory.java
```diff
@@ -0,0 +1,139 @@
...
+        return JaxRsClient.create(
+                clazz,
+                clientConfiguration.userAgent()
+                        .orElseThrow(() -> new SafeIllegalStateException("User agent must be provided")),
+                HOST_METRICS_REGISTRY,
+                clientConfiguration);
+    }
+
+    private static <T> T dialogueClient(Class<T> clazz, ClientConfiguration clientConfiguration) {
+        return callStaticFactoryMethod(clazz, ApacheHttpClientChannels.create(clientConfiguration));
```
iamdanfox: Every call to `ApacheHttpClientChannels.create(clientConfiguration)` creates a new connection pool. I think when timelock is blue/greening it would be cool if we could use the `updateUris` method and avoid re-opening a ton of connections?

iamdanfox:
Is the plan still to wire in a config switch to flip between c-j-r and dialogue?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DefaultTargetFactory.java
ferozco: baby steps Dan. Just trying to get it to run first

ferozco:
Yes, this was meant to just be a draft PR to start iterating against the internal CI

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/BlockingSensitiveConjureTimelockService.java
j-baker: Macro vs micro blocking :)

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DefaultTargetFactory.java
jkozlowski: ya, I started working on wiring this up at some point last month

ferozco:
Closing in favour of #4706

PR #4686, opened 2020-03-30
[PDS-111849] Rolling in the Deep, Part 1: Failing Test & Rerouting Internode Calls Through Proxies

jeremyk-91:
**Goals (and why)**:
- See PDS-111849.
- We have landed on a very plausible RCA, and it's usually good to validate this kind of thing by writing a test that demonstrates what we may have fixed. Here, this test (idea originally from @j-baker) effectively puts a timelock into a mode where it's still alive, but shutting down. Furthermore, I changed it so it takes two seconds before failing, to replicate conditions observed on some internal servers. This is unfortunately a bit more invasive than what I would have liked, but it'll do.

- Write a failing test (see `stressTest`: the number of threads explodes under these circumstances).
- Rewire the `MultiNodePaxosTimeLockIntegrationTest` to go exclusively through WireMock proxies that proxy to the relevant server endpoints. This is basically a glorified, but importantly, dynamically changeable in test filter.

- Doesn't really change production code. Existing tests passing would be validation that I haven't broken the world in my refactoring.

- We now use hard-coded https ports for the wiremock servers. Is that really bad? I think we already depended on the hard-coded timelock ones, so this is just another set of numbers.

**Where should we start reviewing?**: `MultiNodePaxosTimeLockServerIntegrationTest.java`

**Priority (whenever / two weeks / yesterday)**: Ideally by Wednesday

### DIFF
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -48,6 +48,7 @@ import com.palantir.atlasdb.timelock.api.ConjureUnlockRequest;
 import com.palantir.atlasdb.timelock.api.SuccessfulLockResponse;
 import com.palantir.atlasdb.timelock.api.UnsuccessfulLockResponse;
 import com.palantir.atlasdb.timelock.suite.MultiLeaderPaxosSuite;
+import com.palantir.atlasdb.timelock.suite.SingleLeaderPaxosSuite;
 import com.palantir.atlasdb.timelock.util.ExceptionMatchers;
 import com.palantir.atlasdb.timelock.util.ParameterInjector;
 import com.palantir.lock.LockDescriptor;
@@ -130,7 +131,7 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
 
     @Test
     public void canUseNamespaceStartingWithTlOnLegacyEndpoints() {
-        cluster.client("tl" + "suffix").getFreshTimestamp();
+        cluster.client("tl" + "suffix").throughWireMockProxy().getFreshTimestamp();
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -282,6 +285,30 @@ public void multipleLockRequestsWithTheSameIdAreGranted() {
...
 
+    @Test
+    @Ignore // TODO (jkong): Fix this test by reworking the threading model.
+    public void stressTest() {
+        TestableTimelockServer nonLeader = Iterables.getFirst(cluster.nonLeaders(client.namespace()).values(), null);
+        int startingNumThreads = ManagementFactory.getThreadMXBean().getThreadCount();
+        try {
+            for (int i = 0; i < 10_000; i++) { // Needed as it takes a while for the thread buildup to occur
+                client.getFreshTimestamp();
+                assertThat(ManagementFactory.getThreadMXBean().getThreadCount())
```
gmaretic: This test will fail even if the buildup occurs in the warmup period. Maybe at least indicate in the reason for failure if it occurred during warmup or after we simulated shutting down?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -282,6 +285,30 @@ public void multipleLockRequestsWithTheSameIdAreGranted() {
...
+                        .as("should not additionally spin up too many threads")
+                        .isLessThanOrEqualTo(startingNumThreads + 200);
+                if (i == 1_000) {
+                    nonLeader.serverHolder().wireMock().register(
```
gmaretic: Extract a method indicating that this simulates the node shutting down?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: Alright, will do. Yeah, that seems unlikely, but good to catch :)

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: Will do!

PR #4668, opened 2020-03-24
Optimize getRowsColumnRange

j-baker:
We're seeing perf issues internally on the random getRowsColumnRange
implementation that works better on Postgres and Oracle. This is the way
to make it both simpler, and faster.

This is an alternative to https://github.com/palantir/atlasdb/pull/4326.

Needs tests.

PR #4687, opened 2020-03-30
getRowsColumnRange always has cells grouped by row

mswintermeyer:
Ensure that throughout the iterator manipulation that happens
internally to getRowsColumnRange that we maintain the invariant
that all cells for a given row will be grouped together at the end.

**Goals (and why)**:
After https://github.com/palantir/atlasdb/pull/4668 (though maybe before as well if the BatchSizeIncreasingIterator didn't kick in?), getRowsColumnRange could return an iterator where the cells for a given row are not always grouped together, which could break assumptions made by clients.

Two potential implementation options, and I'm curious what people think is cleanest: we can sort the rows passed in, so that they're in the same order that we internally sort Cells in, or we should purely rely on the ordering provided by the KVS.

Changing the existing tests to shuffle the order of rows passed in, which was enough to break things somewhat. But that combined with having rows not have a cell count that's a multiple of the number of the batch size is where the test demonstrates what could could actually go bad.

What option to fix this is cleanest?

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -380,9 +380,13 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             return Collections.emptyIterator();
         }
         hasReads = true;
+        // N.B. Rows must be sorted, to ensure that partitionByRow can correctly assume that all of the cells
+        // in a given row are already all grouped together.
+        List<byte[]> sortedRows = Lists.newArrayList(rows);
+        sortedRows.sort(UnsignedBytes.lexicographicalComparator());
         RowColumnRangeIterator rawResults =
                 keyValueService.getRowsColumnRange(tableRef,
-                                                   rows,
+                                                   sortedRows,
                                                    columnRangeSelection,
                                                    batchHint,
                                                    getStartTimestamp());
@@ -488,6 +492,9 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                             tableRef,
                             raw,
                             x -> x));
+            // Note: the alternative fix here is to keep the original ordering, like this line would do
+            //       and changing the test's expectedCells to only reorder based on the new shuffledRows order
+            // return Iterables.filter(batch, entry -> postFiltered.containsKey(entry.getKey())).iterator();
             return postFiltered.entrySet().iterator();
         }));
     }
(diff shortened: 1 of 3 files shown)


mswintermeyer:
After taking another look, I think I'm more decided on fixing this by ensuring the rows stay returned in their original order, rather than by trying sorting the rows as they're passed in. Some callers pass in a List of rows and assume that the resulting cells will come back in the same row-order.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -476,19 +477,19 @@ public void disableReadWriteConflictChecking(TableReference tableRef) {
     private Iterator<Map.Entry<Cell, Value>> getRowColumnRangePostFiltered(
             TableReference tableRef, RowColumnRangeIterator iterator, int batchHint) {
         return Iterators.concat(Iterators.transform(Iterators.partition(iterator, batchHint), batch -> {
-            ImmutableMap.Builder<Cell, Value> rawBuilder = ImmutableMap.builder();
-            batch.forEach(rawBuilder::put);
-            Map<Cell, Value> raw = rawBuilder.build();
+            // N.B. This batch could be spread across multiple rows, and those rows might extend into other
+            // batches. We are given cells for a row grouped together, so easiest way to ensure they stay together
+            // is to preserve the original order.
+            Map<Cell, Value> raw = new LinkedHashMap<>();
```
j-baker: this refactor is unnecessary - ImmutableMap is in insertion order

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1184,6 +1226,7 @@ public void testRowsColumnRangesSingleIteratorVersion() {
         runTestForGetRowsColumnRangeSingleIteratorVersion(10, 10, 5);
         runTestForGetRowsColumnRangeSingleIteratorVersion(10, 10, 10);
         runTestForGetRowsColumnRangeSingleIteratorVersion(100, 100, 99);
+        runTestForGetRowsColumnRangeSingleIteratorVersion(100, 11, 0);
```
j-baker, line 1229: can you comment why this test is there?

mswintermeyer: done! Wanted to test when a row's cells are spread across multiple batches

PR #4690, opened 2020-03-31
[PDS-111849] Rolling in the Deep, Part 2: Bounded Executors for Acceptors (Single and Batched)

jeremyk-91:
**Goals (and why)**:
- Enforce some concurrency limitation of parallel requests: specifically, restrict `PaxosAcceptor` and `AutobatchingPaxosAcceptor` Functions to only be allowed to open up to 100 threads that are doing work on each separate remote.
- See PDS-111849 for how not having this manifests as badness.
- Don't have ignored tests :)

- Single Leader
  - Don't use the shared executor on acceptor codepaths, instead use a separately generated map of executors that maps local to a direct executor, and the remotes to bounded thread pools of core size 1 and burst size 100.
- Multiple Leaders
  - This case is more difficult: I wired all `PaxosAcceptor` functions inside an `AutobatchingPaxosAcceptorNetworkClientFactory` to share a set of executors - across these, the number of threads permitted per remote is 100.
- Un-ignore the stress test in MNPTLSIT, and add a couple of executor tests.

New tests validate that executor construction should hopefully be sane.

- Ideally the core pool size for the multi-leader case should be 3, not 1. I did think about making it configurable, but don't think we lose anything significant going this way as we'll end up keeping a couple of threads alive after the first requests come in. Not really a fan of hard-coding a number here as well, since it depends on the number of distinct autobatcher functions running. Is this a reasonable decision?
- Limiting Acceptors is likely to be the most critical piece of this work, since that is the endpoint that tends to have the highest concurrency. Pingers and Learners are hit much less so I didn't include them in this PR - and still our stress tests pass. Would it make sense to add this to the other classes as well? The caveat here is that the other endpoints are never really barraged, and so I think a shared executor is actually not unreasonable for efficiency.
- I've wired up the executors so they'll throw a `RejectedExecutionException` if the pools get filled. This is largely handled by `PaxosQuorumChecker`, and it's pretty clear to see that this is correct in single leader. I'm less confident in multi-leader (I'm relatively less comfortable with that part of the code).
- Is 100 a reasonable number? Is this too high, since in theory most of these calls get autobatched? Didn't want to be too strict.
- I had to make the stress test a bit more lenient because in a multi-leader world we get 100 threads blocked on the new node plus another 100 threads blocking in the autobatcher. This may be reducible to some extent after #4664. 

**Where should we start reviewing?**: `TimeLockPaxosExecutors`

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
@@ -54,7 +54,7 @@ public interface Dependencies {
         /**
          * Caution! The shared executor should only be used for tasks that are expected to complete quickly.
          * DO NOT use the shared executor when the concurrency of requests may be very high (e.g. for Paxos round
-         * verification). This may lead to thread explosion.
+         * verification).
          */
         ExecutorService sharedExecutor();
     }
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
@@ -50,6 +50,12 @@ public interface Dependencies {
         PaxosRemoteClients remoteClients();
         LocalPaxosComponents components();
         int quorumSize();
+
+        /**
+         * Caution! The shared executor should only be used for tasks that are expected to complete quickly.
+         * DO NOT use the shared executor when the concurrency of requests may be very high (e.g. for Paxos round
+         * verification). This may lead to thread explosion.
+         */
         ExecutorService sharedExecutor();
     }
 
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
@@ -35,6 +35,7 @@ final class TimeLockPaxosExecutors {
     static final int MAXIMUM_POOL_SIZE = 100;
 
     private static final Duration THREAD_KEEP_ALIVE = Duration.ofSeconds(5);
+    private static final int SINGLE_THREAD_FOR_MOSTLY_AUTOBATCHED_OPERATIONS = 1;
 
     private TimeLockPaxosExecutors() {
         // no
@@ -55,14 +56,23 @@ final class TimeLockPaxosExecutors {
         return remoteExecutors;
     }
 
+    /**
+     * Creates a bounded executor for handling operations on remotes.
+     * These executors are typically called as part of Paxos verification from autobatched contexts, *but* the
+     * individual executions are <b>not</b> autobatched. This means that if one node is performing slowly, calls
+     * pending on that node may build up over time, eventually leading to thread explosion or OOMs. We thus limit
+     * the size to {@link TimeLockPaxosExecutors#MAXIMUM_POOL_SIZE}.
+     *
+     * Users of such an executor should be prepared to handle {@link java.util.concurrent.RejectedExecutionException}.
+     */
     private static ExecutorService createBoundedExecutor(MetricRegistry metricRegistry, String useCase) {
         return new InstrumentedExecutorService(
                 PTExecutors.newThreadPoolExecutor(
-                        1, // Many operations are autobatched, so under ordinary circumstances 1 thread will do
-                        MAXIMUM_POOL_SIZE, // Want to bound the number of threads that might be stuck
+                        SINGLE_THREAD_FOR_MOSTLY_AUTOBATCHED_OPERATIONS,
+                        MAXIMUM_POOL_SIZE,
                         THREAD_KEEP_ALIVE.toMillis(),
                         TimeUnit.MILLISECONDS,
-                        new SynchronousQueue<>(), // Prefer to avoid OOM risk if we get hammered
+                        new SynchronousQueue<>(),
                         new ThreadFactoryBuilder()
                                 .setNameFormat("timelock-executors-" + useCase + "-%d")
                                 .setDaemon(true)
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
@@ -16,6 +16,7 @@
 
 package com.palantir.atlasdb.timelock.paxos;
 
+import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
@@ -23,12 +24,18 @@ import java.util.concurrent.TimeUnit;
 
 import com.codahale.metrics.InstrumentedExecutorService;
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.palantir.common.concurrent.PTExecutors;
 import com.palantir.common.streams.KeyedStream;
 
 final class TimeLockPaxosExecutors {
+    @VisibleForTesting
+    static final int MAXIMUM_POOL_SIZE = 100;
+
+    private static final Duration THREAD_KEEP_ALIVE = Duration.ofSeconds(5);
+
     private TimeLockPaxosExecutors() {
         // no
     }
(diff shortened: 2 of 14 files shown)


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/Dependencies.java
```diff
@@ -50,6 +50,12 @@
         PaxosRemoteClients remoteClients();
         LocalPaxosComponents components();
         int quorumSize();
+
+        /**
+         * Caution! The shared executor should only be used for tasks that are expected to complete quickly.
+         * DO NOT use the shared executor when the concurrency of requests may be very high (e.g. for Paxos round
+         * verification). This may lead to thread explosion.
+         */
```
j-baker: can we slap a limit on the executor?

j-baker: like, fix it at max 256 threads or sth?

jeremyk-91: Yeah, that sounds like a good thing to do

jeremyk-91: Since the claim is that it makes sense to share the executors for the other endpoints as they are explicitly _not_ heavily barraged

jeremyk-91: I'll have a look on internal metrics framework to decide what a good threshold might be (though might still need monitoring further since that will include acceptor activity which is the bulk of the activity).

jeremyk-91: Yeah, internal metrics not super helpful - they do spike above 400+ in asymptomatic operation (and over 1000 on the relevant ticket), but I'm pretty confident a majority of these are acceptor operations.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosQuorumCheckingCoalescingFunction.java
```diff
@@ -96,27 +108,28 @@ public PaxosQuorumCheckingCoalescingFunction(
...
             int quorumSize,
             Function<SERVICE, F> functionFactory) {
-        return services.stream()
-                .map(functionFactory)
-                .collect(collectingAndThen(
-                        toList(), functions -> new PaxosQuorumCheckingCoalescingFunction<>(
-                                functions,
-                                Maps.toMap(functions, $ -> executor),
-                                quorumSize)));
+        // Not 100% sure if ordering is important, but assuming so here
```
gmaretic, line 114: We should remove this comment before merging. +1 for keeping the ordering

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosQuorumCheckingCoalescingFunction.java
```diff
@@ -96,27 +108,28 @@ public PaxosQuorumCheckingCoalescingFunction(
...
+        List<FunctionAndExecutor<F>> functionsAndExecutors = KeyedStream.of(services)
+                .map(executors::get)
+                .mapKeys(functionFactory)
+                .entries()
+                .<FunctionAndExecutor<F>>map(entry -> ImmutableFunctionAndExecutor.of(entry.getKey(), entry.getValue()))
+                .collect(Collectors.toList());
+        return new PaxosQuorumCheckingCoalescingFunction<>(functionsAndExecutors, quorumSize);
```
gmaretic, line 121: I think we went a bit overboard with fanciness to keep it fluent as opposed to
```suggestion
        List<F> functions = new ArrayList<>(services.size());
        Map<F, ExecutorService> executorMap = new HashMap<>(services.size());
        for (SERVICE service: services) {
            F function = functionFactory.apply(service);
            functions.add(function);
            executorMap.put(function, executors.get(service));
        }
        return new PaxosQuorumCheckingCoalescingFunction<>(functions, executorMap, quorumSize);```

File: timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java [hunk 4]
```diff
@@ -286,7 +285,6 @@ public void multipleLockRequestsWithTheSameIdAreGranted() {
     }
 
     @Test
-    @Ignore // TODO (jkong): Fix this test by reworking the threading model.
```
gmaretic, line 289: 🎉

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -140,11 +147,17 @@ private TimeLockAgent(MetricsManager metricsManager,
...
+                        MAX_SHARED_EXECUTOR_THREADS,
+                        5,
+                        TimeUnit.SECONDS,
+                        new SynchronousQueue<>(),
                         new InstrumentedThreadFactory(new ThreadFactoryBuilder()
                                 .setNameFormat("paxos-timestamp-creator-%d")
                                 .setDaemon(true)
-                                .build(), metricsManager.getRegistry())),
+                                .build(), metricsManager.getRegistry()),
+                        new ThreadPoolExecutor.CallerRunsPolicy()), // Be resilient-ish to overloading
```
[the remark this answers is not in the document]
gmaretic: Maybe have a policy that will log or update a metric if this ever happens, so we can detect if this is actually being hit and our choice for the bound is suspicious

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
```diff
@@ -0,0 +1,73 @@
...
+                .map(remote -> createBoundedExecutor(metricRegistry, useCase))
+                .collectToMap();
+        remoteExecutors.put(localAndRemotes.local(), MoreExecutors.newDirectExecutorService());
+        return remoteExecutors;
+    }
+
+    private static ExecutorService createBoundedExecutor(MetricRegistry metricRegistry, String useCase) {
+        return new InstrumentedExecutorService(
+                PTExecutors.newThreadPoolExecutor(
+                        1, // Many operations are autobatched, so under ordinary circumstances 1 thread will do
```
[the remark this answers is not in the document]
jeremyk-91: OK, will do. Yep, makes sense

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
```diff
@@ -0,0 +1,73 @@
...
+                        MAXIMUM_POOL_SIZE, // Want to bound the number of threads that might be stuck
```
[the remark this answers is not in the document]
jeremyk-91: Yep, I'll move the comments to the top.

timelock-agent/src/test/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutorsTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    public void remoteExecutorsAreLimitedSeparately() {
+        for (int i = 0; i < TimeLockPaxosExecutors.MAXIMUM_POOL_SIZE; i++) {
+            executors.get(remote1).submit(SLEEP_FOR_ONE_SECOND);
+        }
+        assertThatCode(() -> executors.get(remote2).submit(SLEEP_FOR_ONE_SECOND))
+                .doesNotThrowAnyException();
+    }
+}
```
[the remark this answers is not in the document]
jeremyk-91: Yep, makes sense. I was a bit torn on whether to include this test (it didn't seem like that was necessarily something I was guaranteeing), but fine to have it.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosQuorumCheckingCoalescingFunction.java
jeremyk-91: Fair enough, when I threaded the needle I (probably wrongly) thought it was clear enough 😅 but agree this version is clearer.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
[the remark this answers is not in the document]
jeremyk-91: Yep, added a metric to the execution handler.

PR #4695, opened 2020-04-02
[Timelock Partitioning] Allow Switching to Batched Single Leader

gmaretic:
Partially paired with @jeremyk-91 on this

**Goals (and why)**:
Allow switching to batched single leader for paxos acceptor/learner via runtime config.

Use `PredicateSwitchedProxy` to switch between legacy client and the batched client depending on the runtime config

Added a third parameter to `SingleLeaderMultiNodePaxosTimeLockIntegrationTest` that runs the tests in batched single leader mode and sanity checked we are hitting the correct endpoints.

- I would have preferred to do the above tests by changing the runtime config during tests, but didn't see an easy way to bake that into the dropwizard server. Probably not worth investing the time into this.
- The client creation in TimelockPaxos(Acceptor|Learner)Adapters for single leader assumes that the clients in `remoteClients.batch(Acceptor|Learner)()` and `remoteClients.singleLeaderAcceptor|Learner)()` are in consistent order, given that they are lists. This is indeed the case, but is it safe to assume this invariant? The alternative is to make the whole list switch on the proxy, but that is not foolproof in case of runtime config change anyway.
- is there a better way to implementat `BatchTimelockPaxosLearnerAdapter. getGreatestLearnedValue()` or is tracking what we've seen so far the only option?

`PaxosResourcesFactory.java`

Early next week


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
```diff
@@ -26,21 +27,25 @@
...
-import com.palantir.timelock.paxos.TimelockPaxosLearnerAdapter;
+import com.palantir.timelock.paxos.TimelockPaxosAcceptorAdapters;
+import com.palantir.timelock.paxos.TimelockPaxosLearnerAdapters;
 
 @Value.Immutable
 abstract class SingleLeaderNetworkClientFactories implements
         NetworkClientFactories, Dependencies.NetworkClientFactories {
 
+    @Value.Default
+    Supplier<Boolean> useBatchedEndpoints() {
```
jeremyk-91, line 38: uncertain: Might be better to use primitive `BooleanSupplier`, though I wouldn't block if you need `Supplier<T>` for something down the line.

timelock-agent/src/main/java/com/palantir/timelock/paxos/BatchTimelockPaxosLearnerAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+        Preconditions.checkState(result.size() <= 1,
+                "Unexpected result {} in a call for client {}.",
+                SafeArg.of("result", result),
+                SafeArg.of("client", client));
+    }
+
+    private void updateLastKnownSequence(long seq) {
+        if (seq > lastKnownSequence) {
+            lastKnownSequence = seq;
+        }
```
jeremyk-91, line 106: Let's use an `AtomicLong` and `accumulateAndGet`: not _wrong_ because of how it's used, but there could be unwanted race conditions

timelock-agent/src/main/java/com/palantir/timelock/paxos/BatchTimelockPaxosLearnerAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+                .reduce((fst, snd) -> fst.getRound() > snd.getRound() ? fst : snd);
+        greatestIfExists.ifPresent(paxosValue -> updateLastKnownSequence(paxosValue.getRound()));
+        return greatestIfExists;
+    }
+
+    @Override
+    public Collection<PaxosValue> getLearnedValuesSince(long seq) {
+        Set<PaxosValue> result = rpcClient.getLearnedValuesSince(paxosUseCase, ImmutableMap.of(client, seq))
+                .get(client);
+        result.stream().map(PaxosValue::getRound).mapToLong(x -> x).max().ifPresent(this::updateLastKnownSequence);
```
jeremyk-91, line 92: Nice! Many ways to go about this, and looks like you picked the one that touches the volatile/atomic thing as little as possible.

timelock-agent/src/main/java/com/palantir/timelock/paxos/BatchTimelockPaxosLearnerAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+        return Optional.of(Iterables.getOnlyElement(result));
+    }
+
+    @Override
+    public Optional<PaxosValue> getGreatestLearnedValue() {
+        Set<PaxosValue> result = rpcClient
+                .getLearnedValuesSince(paxosUseCase, ImmutableMap.of(client, lastKnownSequence))
+                .get(client);
+        Optional<PaxosValue> greatestIfExists = result.stream()
+                .reduce((fst, snd) -> fst.getRound() > snd.getRound() ? fst : snd);
```
jeremyk-91, line 83: Equivalent, but I think I like the readability of this better:

```suggestion
                .max(Comparator.comparingLong(PaxosValue::getRound));
```

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosAcceptorAdapters.java
```diff
@@ -0,0 +1,70 @@
...
+    public static List<PaxosAcceptor> create(
+            PaxosUseCase paxosUseCase,
+            PaxosRemoteClients remoteClients,
+            Supplier<Boolean> useBatchedSingleLeader,
+            Client client) {
+        switch (paxosUseCase) {
+            case LEADER_FOR_ALL_CLIENTS:
+                return Streams.zip(
+                        remoteClients.batchAcceptor().stream(),
+                        remoteClients.singleLeaderAcceptor().stream(),
```
jeremyk-91, line 46: Discussed offline: Here, iteration order is very important. It's *essential* that we go through these in the same order - and while these are both lists, there is an underlying set and technically no guarantee that the iteration order is consistent. Thus, I'd suggest making `PaxosResourcesFactory#remoteUris` return a list rather than a set.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimelockPaxosLearnerAdapters.java
```diff
@@ -0,0 +1,69 @@
...
+import com.palantir.atlasdb.timelock.paxos.Client;
+import com.palantir.atlasdb.timelock.paxos.PaxosRemoteClients;
+import com.palantir.atlasdb.timelock.paxos.PaxosUseCase;
+import com.palantir.common.proxy.PredicateSwitchedProxy;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.paxos.PaxosLearner;
+
+public final class TimelockPaxosLearnerAdapters {
+    private TimelockPaxosLearnerAdapters() {
+        // how many times
```
jeremyk-91, line 34: 💯

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/suite/SingleLeaderPaxosSuite.java
```diff
@@ -43,20 +43,29 @@
...
             "paxosMultiServer.ftl",
             generateThreeNodeTimelockCluster(9080, builder ->
-                    builder.clientPaxosBuilder(builder.clientPaxosBuilder().isUseBatchPaxos(false))
+                    builder.clientPaxosBuilder(builder.clientPaxosBuilder().isUseBatchPaxosTimestamp(false))
                             .leaderMode(PaxosLeaderMode.SINGLE_LEADER)));
 
     public static final TestableTimelockCluster BATCHED_TIMESTAMP_PAXOS = new TestableTimelockCluster(
             "batched timestamp paxos single leader",
             "paxosMultiServer.ftl",
             generateThreeNodeTimelockCluster(9083, builder ->
```
jeremyk-91, line 52: These should be `9086` and `9092`, just because of wiremock we might actually use six ports, not three.

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/SingleLeaderNetworkClientFactories.java
gmaretic: PredicateSwitchedProxy uses `Supplier<Boolean>` so just more consistent

timelock-agent/src/test/java/com/palantir/timelock/paxos/PaxosRemotingUtilsTest.java
```diff
@@ -141,14 +136,14 @@ public void addProtocolAddsHttpsIfSslPresent() {
...
+        assertThat(PaxosRemotingUtils.addProtocols(NO_SSL_TIMELOCK, ImmutableList.of("foo:1", "bar:2")))
+                .isEqualTo(ImmutableList.of("http://foo:1", "http://bar:2"));
     }
 
     @Test
     public void addProtocolsAddsHttpsIfSslPresent() {
-        assertThat(PaxosRemotingUtils.addProtocols(SSL_TIMELOCK, ImmutableSet.of("foo:1", "bar:2")))
-                .containsExactlyInAnyOrder("https://foo:1", "https://bar:2");
+        assertThat(PaxosRemotingUtils.addProtocols(SSL_TIMELOCK, ImmutableList.of("foo:1", "bar:2")))
+                .isEqualTo(ImmutableList.of("https://foo:1", "https://bar:2"));
```
jeremyk-91, line 146: I would have thought `containsExactly` (without the `inAnyOrder` bit) would have been enough?

PR #4696, opened 2020-04-03
fix ServiceLoader in ServiceDiscoveringAtlasSupplier

tjwilson90:
ServiceLoader is not thread safe


tjwilson90:
The behavior I saw in concurrent tests was that the service loader occasionally appeared empty.

PR #4698, opened 2020-04-06
[PDS-111849] Rolling in the Deep, Part 3: Fix Metrics Plughole

jeremyk-91:
**Goals (and why)**:
- Actually allow legacy metrics to be propagated from TimeLock codepaths. Previous impl claimed that we don't, but we now do indirectly with instrumented executor services.

- Rewire TimelockPaxosMetrics to take a MetricsManager instead.

- Not much, unfortunately.

- Did I miss another blackhole?

**Where should we start reviewing?**: TimelockPaxosMetrics.java

### DIFF
--- a/changelog/@unreleased/pr-4698.v2.yml
+++ b/changelog/@unreleased/pr-4698.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Executors used internally by TimeLock for its Paxos protocol now publish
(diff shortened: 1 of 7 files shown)


jeremyk-91:
I don't think this PR works. I'll need to investigate further.

jeremyk-91:
actually it does: I forgot to restart a box

PR #4701, opened 2020-04-08
PTExecutors improvements for cached executor resource utilization

carterkozak:
**Goals (and why)**:

We have a ton of threads at any given point in time. While we have a ton of threads, we're also actively creating and destroying threads across existing thread pools where we could reuse the resources we've already created.

PTExecutors cached executors use a short timeout duration of 5 seconds (3 second for BasicSQL.java) which contributes heavily to the aforementioned thread churn, especially in real world environments that can't keep the executors saturated like synthetic benchmarking.

* PTExecutors thread factory uses daemon threads by default. Most consumers prefer daemon threads to avoid propping up JVMs that want to die.
* PTExecutors.newCachedThreadPool returns ExecutorService instances instead of ThreadPoolExecutor concrete types. This is an ABI break. Please use WC executor factories wherever possible (note this optimization is not applied there quite yet, on on my list)
* Cached pool factories that don't take a ThreadFactory provide a decorated view onto a shared cached executors so that threads may flow freely between all cached executors.

Existing testing covers behavior of the refactored code, added test coverage for new functionality.

¯\_(ツ)_/¯

¯\_(ツ)_/¯

¯\_(ツ)_/¯

### DIFF
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorView.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorView.java
@@ -49,19 +49,23 @@ final class CachedExecutorView extends AbstractExecutorService {
     private static final int ACTIVE_COUNT_MASK = (1 << 30) - 1;
 
     // Hoist expensive lambda allocations
-    private static final IntUnaryOperator executeStateUpdater = current -> {
-        // This is safe in a compareAndUpdate because SHUTDOWN_MASK is never unset.
-        if (isShutdown(current)) {
-            throw new RejectedExecutionException("Executor has been shut down");
+    private static final IntUnaryOperator incrementActiveStateUpdater = current -> {
+        int activeCount = getActiveCount(current);
+        // Defensive check. If there are a million active tasks something has gone terribly wrong.
+        if (activeCount == ACTIVE_COUNT_MASK) {
+            throw new SafeIllegalStateException("There are already " + ACTIVE_COUNT_MASK + " active tasks");
         }
-        validateForIncrement(current);
-        int updatedActiveCount = getActiveCount(current) + 1;
+        int updatedActiveCount = activeCount + 1;
         return updatedActiveCount | (current & ~ACTIVE_COUNT_MASK);
     };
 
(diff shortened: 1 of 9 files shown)


commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorView.java
```diff
@@ -0,0 +1,178 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
carterkozak, line 2: wrong license, fix incoming

commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorView.java
```diff
@@ -0,0 +1,226 @@
...
+        int stateSnapshot = stateUpdater.updateAndGet(this, decrementActiveStateUpdater);
+        if (isTerminated(stateSnapshot)) {
+            synchronized (shutdownLock) {
+                shutdownLock.notifyAll();
+            }
+        }
+    }
+
+    private static void validateForIncrement(int state) {
+        // Defensive check. If there are a million active tasks something as gone terribly wrong.
```
carterkozak, line 176:
```suggestion
        // Defensive check. If there are a million active tasks something has gone terribly wrong.
```

commons-executors/src/main/java/com/palantir/common/concurrent/CachedExecutorView.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Override
+    public void execute(Runnable task) {
+        boolean submittedTask = false;
+        stateUpdater.updateAndGet(this, executeStateUpdater);
+        try {
+            if (isShutdown()) {
+                throw new RejectedExecutionException("Executor has been shut down");
+            }
```
carterkozak, line 156: these 3 lines should be deleted

carterkozak: resolved in the other direction

jeremyk-91:
We can hold off for now, I’d like to move the implementation to a shared library since we’re already using it in <large internal product>

PR #4702, opened 2020-04-08
[PDS-115789] API Contracts for Snapshot Transaction Methods

jeremyk-91:
**Goals (and why)**:
- The lack of precision on what is and isn't expected of these methods was probably a contributing factor to PDS-115789.
- Make it very clear what, as AtlasDB developers, we need to check when these methods are changed
- Make it clear what, as AtlasDB users, we can and cannot expect from these endpoints.

- Add comments to the four methods at the top of SnapshotTransaction

- No new prod code here.

- Are these contracts correctly defined given (1) what the code actually does, and (2) current usage in the field?
- Some of these contracts may be stronger than necessary, in terms of the guarantees we offer consumers. Do we want to continue offering such strong guarantees?
- A little less confident on users seeing these (they use the generated code), but a responsible user should check; also, this is still useful for Atlas dev.

_I would not like this PR to merge without a sanity check from @j-baker and @mswintermeyer (or someone from large internal product)._

**Where should we start reviewing?**: SnapshotTransaction

**Priority (whenever / two weeks / yesterday)**: this week or early next.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
@@ -41,18 +41,68 @@ import com.palantir.common.base.BatchingVisitable;
  */
 public interface Transaction {
 
+    /**
+     * Returns a mapping of rows to {@link RowResult}s within {@code tableRef} for the specified {@code rows}, loading
+     * columns according to the provided {@link ColumnSelection}.
+     *
+     * The returned {@link SortedMap} is sorted on the byte order of row keys; the ordering of the input parameter
+     * {@code rows} is irrelevant.
+     *
+     * If there are rows with no cells matching the provided {@link ColumnSelection}, they will not be present in the
+     * {@link Map#keySet()} of the output map at all.
+     *
+     * @param tableRef table to load rows from
+     * @param rows rows to be loaded
+     * @param columnSelection columns to load from the given rows
+     * @return a mapping of rows to the columns matching the provided column selection.
+     */
     @Idempotent
     SortedMap<byte[], RowResult<byte[]>> getRows(
             TableReference tableRef,
             Iterable<byte[]> rows,
             ColumnSelection columnSelection);
 
+    /**
+     * Returns a mapping of rows to {@link BatchingVisitable}s (which may be thought of as iterators over columns)
+     * within {@code tableRef} for the specified {@code rows}, where the columns fall within the provided
+     * {@link BatchColumnRangeSelection}. The single provided {@link BatchColumnRangeSelection} applies to all of the
+     * rows.
+     *
+     * The returned {@link BatchingVisitable}s are guaranteed to return cells matching the predicate, sorted on
+     * (ascending) byte ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has the same elements as {@code rows},
+     * even if there are rows where no columns match the predicate.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply the column range selection to
+     * @param columnRangeSelection range of columns and batch size to load for each of the rows provided
+     * @return a mapping of rows to cells matching the predicate in the row, following the ordering outlined above
+     * @throws IllegalArgumentException if {@code rows} contains duplicates
+     */
     @Idempotent
     Map<byte[], BatchingVisitable<Map.Entry<Cell, byte[]>>> getRowsColumnRange(
             TableReference tableRef,
             Iterable<byte[]> rows,
             BatchColumnRangeSelection columnRangeSelection);
 
+    /**
+     * Returns a single iterator over the cell-value pairs in {@code tableRef} for the specified {@code rows}, where the
+     * columns fall within the provided {@link ColumnRangeSelection}. The single provided {@link ColumnRangeSelection}
+     * applies to all of the rows.
+     *
+     * If the provided {@link Iterable} of {@code rows} has a stable ordering, the returned iterator is guaranteed
+     * to return cell-value pairs in a lexicographic ordering over rows and columns, where rows are sorted according to
+     * the stable ordering of {@code rows}, and columns are sorted on byte ordering. If {@code rows} does not have a
+     * stable ordering, behaviour is undefined.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply the column range selection to
+     * @param columnRangeSelection range of columns to load for each of the rows provided
+     * @param batchHint number of columns that should be loaded from the underlying database at once
+     * @return an iterator over cell-value pairs, guaranteed to follow the ordering outlined above
+     * @throws IllegalArgumentException if {@code rows} contains duplicates
+     */
     @Idempotent
     Iterator<Map.Entry<Cell, byte[]>> getRowsColumnRange(
             TableReference tableRef,
@@ -60,6 +110,23 @@ public interface Transaction {
             ColumnRangeSelection columnRangeSelection,
             int batchHint);
 
+    /**
+     * Returns a mapping of rows to {@link Iterator}s over cell-value pairs within {@code tableRef} for the specified
+     * {@code rows}, where the columns fall within the provided {@link BatchColumnRangeSelection}. The single provided
+     * {@link BatchColumnRangeSelection} applies to all of the rows.
+     *
+     * The returned {@link Iterator}s are guaranteed to return cells matching the predicate, sorted on (ascending) byte
+     * ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has the same elements as {@code rows},
+     * even if there are rows where no columns match the predicate.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply the column range selection to
+     * @param columnRangeSelection range of columns and batch size to load for each of the rows provided
+     * @return a mapping of rows to cells matching the predicate in the row, following the ordering outlined above
+     * @throws IllegalArgumentException if {@code rows} contains duplicates
+     */
     @Idempotent
     Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> getRowsColumnRangeIterator(
             TableReference tableRef,


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,92 @@
...
+     * Returns a mapping of rows to {@link BatchingVisitable}s (which may be thought of as iterators over columns)
+     * within {@code tableRef} for the specified {@code rows}, where the columns fall within the provided
+     * {@link BatchColumnRangeSelection}. The single provided {@link BatchColumnRangeSelection} applies to all of the
+     * rows.
+     *
+     * The returned {@link BatchingVisitable}s are guaranteed to return cells matching the predicate, sorted on
+     * (ascending) byte ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has the same elements as {@code rows},
+     * even if there are rows where no columns match the predicate.
```
jeremyk-91, line 75: @jkozlowski notes: does same elements here mean the same *objects*, or just arrays that are equal in terms of `Arrays#equals()`? I intended the former, historically the behaviour of getRowsColumnRange is as follows:

```
    @Test
    public void identity() {
        txMgr.runTaskWithRetry((tx) -> {
            put(tx, "tom", "aaa", "42");
            return null;
        });

        txMgr.runTaskWithRetry((tx) -> {
            byte[] b1 = PtBytes.toBytes("tom");
            Map<byte[], BatchingVisitable<Map.Entry<Cell, byte[]>>> iterators = tx.getRowsColumnRange(TEST_TABLE,
                    ImmutableList.of(b1),
                    BatchColumnRangeSelection.create(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 10));

            byte[] b1v2 = PtBytes.toBytes("tom");
            assertThat(iterators.get(b1)).isNotNull(); // Passes
            assertThat(iterators.get(b1v2)).isNotNull(); // Fails
            return null;
        });
    }
```

We should probably be stricter/more explicit that the identity of the rows *must* be equal, will incorporate this.

jeremyk-91: To be clear, unless people feel strongly about this I have no intention of codifying the failure of the second assertion above - but will say that the map must be accessed by the original rows passed in, and using other rows (even if equal in terms of byte arrays to an original row) is undefined behaviour.

jkozlowski: Yes, we should make it clear that pulling things out by the original byte arrays is fine.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,92 @@
...
+     * If there are rows with no cells matching the provided {@link ColumnSelection}, they will not be present in the
+     * {@link Map#keySet()} of the output map at all.
+     *
+     * @param tableRef table to load rows from
+     * @param rows rows to be loaded
+     * @param columnSelection columns to load from the given rows
+     * @return a mapping of rows to the columns matching the provided column selection
+     */
     @Idempotent
     SortedMap<byte[], RowResult<byte[]>> getRows(
```
jkozlowski, line 60: For this one though, looking up in the returned SortedMap should obviously be allowed with the original byte array, as well as a new byte array with same contents.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Agreed. My point is also that pulling things out by means other than the byte arrays is a bad idea, though I don't want to guarantee that it _will_ fail.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Yep. Pretty sure that's the behaviour, this one is governed by `UnsignedBytes.lexicographicalComparator`. I can specify some of that.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
...
 
+    /**
+     * Returns a mapping of rows to {@link RowResult}s within {@code tableRef} for the specified {@code rows}, loading
+     * columns according to the provided {@link ColumnSelection}.
+     *
+     * The returned {@link SortedMap} is sorted on the byte order of row keys; the ordering of the input parameter
+     * {@code rows} is irrelevant. Access to the returned {@link SortedMap} is permitted both with the original byte
+     * array, as well as with byte arrays that are equal in terms of {@link java.util.Arrays#equals(byte[], byte[])}.
+     *
+     * If there are rows with no cells matching the provided {@link ColumnSelection}, they will not be present in the
```
j-baker, line 52: It's worth a test for this, I think. Specifically, what do I get back if I e.g. delete a cell in a txn and then read the row?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
  */
 public interface Transaction {
 
+    /**
+     * Returns a mapping of rows to {@link RowResult}s within {@code tableRef} for the specified {@code rows}, loading
+     * columns according to the provided {@link ColumnSelection}.
+     *
+     * The returned {@link SortedMap} is sorted on the byte order of row keys; the ordering of the input parameter
+     * {@code rows} is irrelevant. Access to the returned {@link SortedMap} is permitted both with the original byte
```
j-baker, line 49: 'with any byte array that is equivalent according to Arrays.equals(byte[], byte[])'.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
...
             ColumnSelection columnSelection);
 
+    /**
+     * Returns a mapping of rows to {@link BatchingVisitable}s (which may be thought of as iterators over columns)
+     * within {@code tableRef} for the specified {@code rows}, where the columns fall within the provided
+     * {@link BatchColumnRangeSelection}. The single provided {@link BatchColumnRangeSelection} applies to all of the
+     * rows.
+     *
+     * The returned {@link BatchingVisitable}s are guaranteed to return cells matching the predicate, sorted on
+     * (ascending) byte ordering.
```
j-baker, line 73: of columns?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
...
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has the same elements as {@code rows} by
+     * identity, even if there are rows where no columns match the predicate. Random access for the returned map
+     * should ONLY be performed through the original byte arrays that were passed in.
```
j-baker, line 77: Can we tighten this up? It'd be easy to (under the hood) just wrap with an ImmutableSortedMap and so have a consistent contract between this and getRows.

j-baker: Like, arguably we should have a consistent contract between all our methods.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
...
 
+    /**
+     * Returns a single iterator over the cell-value pairs in {@code tableRef} for the specified {@code rows}, where the
+     * columns fall within the provided {@link ColumnRangeSelection}. The single provided {@link ColumnRangeSelection}
+     * applies to all of the rows.
+     *
+     * If the provided {@link Iterable} of {@code rows} has a stable ordering, the returned iterator is guaranteed
+     * to return cell-value pairs in a lexicographic ordering over rows and columns, where rows are sorted according to
+     * the stable ordering of {@code rows}, and columns are sorted on byte ordering. If {@code rows} does not have a
+     * stable ordering, behaviour is undefined.
```
j-baker, line 99: Can we tighten this up as well? You're right, of course, but all hell breaks loose if this happens. So when you see something like this, I think it's on us to actually clean the contract. In this case, internally we should do ImmutableList.copyOf.

j-baker: (assuming that this is not lazy, in which case the contract might need tidying up).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -41,25 +41,95 @@
...
+     * Returns a mapping of rows to {@link Iterator}s over cell-value pairs within {@code tableRef} for the specified
+     * {@code rows}, where the columns fall within the provided {@link BatchColumnRangeSelection}. The single provided
+     * {@link BatchColumnRangeSelection} applies to all of the rows.
+     *
+     * The returned {@link Iterator}s are guaranteed to return cells matching the predicate, sorted on (ascending) byte
+     * ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has the same elements as {@code rows} by
+     * identity, even if there are rows where no columns match the predicate. Random access for the returned map
+     * should ONLY be performed through the original byte arrays that were passed in.
```
j-baker, line 125: Likewise. Basically, if we get each method to the point where you can always lookup in the map via any equivalent byte array, then we can just shove it in a single comment at the top and avoid it on each method.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Yep, of course the same array is equal to itself!

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Yep, will state that explicitly.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Yep, changed this (and added appropriate tests).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Yeah, added an internal ImmutableList.copyOf, and also a funky test that actually gives unstable iteration order.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: As above: this was done.

jeremyk-91:
Still need to clarify precise semantics for getRows (including the case James flags). The row-column-range comments have been addressed.

jeremyk-91:
Should be ready for a second pass.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
```diff
@@ -1309,6 +1318,149 @@ public void testGetRangesPaging() {
...
+        t = startTransaction();
+        SortedMap<byte[], RowResult<byte[]>> result =
+                t.getRows(TEST_TABLE, ImmutableList.of(rowKey), ColumnSelection.all());
+        assertThat(result.get(rowKey))
+                .as("it should be possible to get a row from getRows with a passed-in byte array")
+                .isNotNull()
+                .satisfies(rowResult ->
+                        assertThat(rowResult.getOnlyColumnValue()).isEqualTo(value));
+
+        byte[] rowKeyCopy = row(0);
```
j-baker, line 1339: nit: i think this is more clear as rowKey.clone()

j-baker: i suppose, same as in other places

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+public class UnstableOrderedIterable<T> implements Iterable<T> {
```
j-baker, line 27: :rofl:

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterableTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public void orderingIsUnstable() {
+        // Strobes once in 1000000! times. We can live with that.
+        List<Integer> numbers = IntStream.range(0, 1000000).boxed().collect(Collectors.toList());
+        Iterable<Integer> numbersIterable = new UnstableOrderedIterable<>(numbers);
+
+        List<Integer> iterationOrder = Lists.newArrayList();
+        numbersIterable.iterator().forEachRemaining(iterationOrder::add);
+
+        List<Integer> secondIterationOrder = Lists.newArrayList();
+        numbersIterable.iterator().forEachRemaining(secondIterationOrder::add);
```
j-baker, line 41: List<Integer> iterationOrder = ImmutableList.copyOf(numbersIterable);
List<Integer> secondIterationOrder = ImmutableList.copyOf(numbersIterable);

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterableTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+
+public class UnstableOrderedIterableTest {
+    @Test
+    public void orderingIsUnstable() {
+        // Strobes once in 1000000! times. We can live with that.
+        List<Integer> numbers = IntStream.range(0, 1000000).boxed().collect(Collectors.toList());
```
j-baker, line 34: nope nope nope. Nondeterministic tests are bad. In this case you have to create 1 million objects.

Instead, UnstableOrderedIterable should take as a field a Random, and you can instantiate that deterministically here, use the version of shuffle that takes a Random and guarantee it 100% of the time.

j-baker: with 2 elements

j-baker: Actually, you're doing shuffling here, instead you should have UnstableOrderedIterable return OrderedPermutations.



the test I'd like to see here is one that basically reads:

```java
Iterable<Integer> numbers = new UnstableOrderedIterable<>(ImmutableList.of(1, 2));
assertThat(ImmutableList.copyOf(numbers)).containsExactly(1, 2);
assertThat(ImmutableList.copyOf(numbers)).containsExactly(2, 1);
```

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+    private final List<T> underlying;
+
+    public UnstableOrderedIterable(Collection<T> underlying) {
+        this.underlying = ImmutableList.copyOf(underlying);
+    }
+
+    @Override
+    public Iterator<T> iterator() {
```
j-baker, line 35: Recommend this, makes for an easier test and guarantees the property you want.

```java
public class UnstableOrderedIterable<T extends Comparable<? super T>> implements Iterable<T> {
    private final Iterator<ImmutableList<T>> underlying;

    public UnstableOrderedIterable(Collection<T> underlying) {
        this.underlying = Iterables.cycle(Collections2.orderedPermutations(underlying));
    }

    @Override
    public Iterator<T> iterator() {
        return underlying.next();
    }
}
```

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
+ *
+ * In general: users may assume that if maps (including sorted maps) keyed on byte[] are returned to the user,
+ * they may be accessed, including via random access, via any byte array that is equivalent in terms of
+ * {@link java.util.Arrays#equals(byte[], byte[])}.
+ *
  * @see TransactionManager
  */
 public interface Transaction {
 
+    /**
```
j-baker, line 49: and, just to make sure... do other Transaction implementations such as CachedTransaction obey these constraints? I know we have AbstractTransactionTests, do those get run on CachedTransactions? Personally I'd be minded to write a 'contract validating transaction' which we can wrap our transactions in.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
jeremyk-91: Alright, yeah agree that reads better. Will change.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterableTest.java
jeremyk-91: Will change (in the end followed your above outline). Nice

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/UnstableOrderedIterable.java
jeremyk-91: Yep, this is nice. I implemented this (some small changes to allow a `Comparator`, but the basic idea is what you've got).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Is AbstractTransactionTest run on CachingTransactions: good flag. Not previously, but now it is - we've tested for almost all of the contractual nuances by now, and it passed all of the tests.

jeremyk-91: Most of the others just delegate (or in the case of ReadTransaction delegate the relevant things), from a quick pass.

jeremyk-91: (Turns out there are some weird CircleCI issues that I can't reproduce locally around TMs being shut. I wouldn't block on this currently)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
+     * Returns a mapping of rows to {@link RowResult}s within {@code tableRef} for the specified {@code rows}, loading
+     * columns according to the provided {@link ColumnSelection}.
+     *
+     * The returned {@link SortedMap} is sorted on the byte order of row keys; the ordering of the input parameter
+     * {@code rows} is irrelevant.
+     *
+     * If there are rows with no cells matching the provided {@link ColumnSelection}, they will not be present in the
+     * {@link Map#keySet()} of the output map at all. This accounts for local writes and deletes: a row written
```
jkozlowski, line 57: I think "local" might be an overloaded term for this. I would also maybe not mention key value service.

What about: "this accounts for writes and deletes done in this transaction: a row written to in this transaction will be present, and a row which is deleted in this transaction will be absent"

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
+     * @return a mapping of rows to the columns matching the provided column selection
+     */
     @Idempotent
     SortedMap<byte[], RowResult<byte[]>> getRows(
             TableReference tableRef,
             Iterable<byte[]> rows,
             ColumnSelection columnSelection);
 
+    /**
+     * Returns a mapping of rows to {@link BatchingVisitable}s (which may be thought of as iterators over columns)
```
jkozlowski, line 73: nit: "which may be thought of as iterators over columns". Unclear of the value of putting this here? you are linking to the interface, perhaps the interface should have this sentence?

jkozlowski: Just an idea, so can be noop: perhaps you don't actually want to use "BatchingVisitable" of "BatchColumnRangeSelection" and explain it thus:

"Returns a mapping of requested rows to corresponding columns in this table. Only columns matching the single column range selection will be returned." @Jolyon-S  can provide better english.

Maybe this reads worse, hard to say.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
             BatchColumnRangeSelection columnRangeSelection);
 
+    /**
+     * Returns a single iterator over the cell-value pairs in {@code tableRef} for the specified {@code rows}, where the
+     * columns fall within the provided {@link ColumnRangeSelection}. The single provided {@link ColumnRangeSelection}
+     * applies to all of the rows.
+     *
+     * If the provided {@link Iterable} of {@code rows} has a stable ordering, the returned iterator is guaranteed
+     * to return cell-value pairs in a lexicographic ordering over rows and columns, where rows are sorted according to
+     * the stable ordering of {@code rows}, and columns are sorted on byte ordering. If {@code rows} does not have a
```
jkozlowski, line 103: Why does iteration order of the passed iterable need to be stable? Can we not pick one format for the return order and stick to it?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
             Iterable<byte[]> rows,
             ColumnRangeSelection columnRangeSelection,
             int batchHint);
 
+    /**
+     * Returns a mapping of rows to {@link Iterator}s over cell-value pairs within {@code tableRef} for the specified
+     * {@code rows}, where the columns fall within the provided {@link BatchColumnRangeSelection}. The single provided
+     * {@link BatchColumnRangeSelection} applies to all of the rows.
+     *
+     * The returned {@link BatchingVisitable}s are guaranteed to return cells matching the predicate. These are sorted
```
jkozlowski, line 127: We don't return BatchingVisitable here?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
+     * by column, with ascending byte ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
+     * input {@code rows}, even if there are rows where no columns match the predicate.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply the column range selection to
```
jkozlowski, line 134: What happens when these rows are not unique?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,124 @@
...
     @Idempotent
     Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> getRowsColumnRangeIterator(
             TableReference tableRef,
             Iterable<byte[]> rows,
             BatchColumnRangeSelection columnRangeSelection);
 
+    /**
+     * Gets the values associated for each cell in {@code cells} from table specified by {@code tableRef}.
+     *
+     * @param tableRef the table from which to get the values
```
jkozlowski, line 148: Should this also have something around Cell lookup from this map? I guess that's implied because Cell has an #equals we control.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Local is used in some of the internal methods, though yeah that's more of an internal implementation term, so will change to "in this transaction".

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: good spot, bad copy-paste here 😓

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: I could add it, though yeah I didn't because Cell#equals() behaves predictably

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: The return order for the stable case has to be the same order as was passed in (this was the core issue in PDS-115789, where an internal user did `getRowsColumnRange(table, [B, A], cols, batch)` and expected the cells for `B` to come back first). 

Generalising this to iterables without a consistent ordering seemed awkward. We _could_ guarantee something stronger (i.e. `if rows does not have a stable ordering, the returned iterator will return cell-value pairs where rows are sorted according to the order of the first iteration over it, and columns ...`), but it seems like an implementation detail I wouldn't want users to rely on, and it seems easy to break inadvertently.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: I think changing this to the version without the classes seems reasonable.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jkozlowski: How about saying "Returned iterator is in the order of rows iterable then cells lexicographically. If your iterable does not have a stable iterator (i.e. iteration order can change between different calls to .iterator()) the cells are ordered lexicographically, but the order of rows is undefined."

As in I would put emphasis on the correct case, but explain that things will explode if that doesn't hold.

As a broader question: could we deprecate this method and automatically try to rewrite user's code (or rather suggest a rewritten version, to be reviewed by devs).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
 
 /**
  * Provides the methods for a transaction with the key-value store.
+ *
+ * In general: users may assume that if maps (including sorted maps) keyed on byte[] are returned to the user,
+ * they may be accessed, including via random access, via any byte array that is equivalent in terms of
```
jkozlowski, line 42: What does "including via random access" mean?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
...
+ * {@link java.util.Arrays#equals(byte[], byte[])}.
+ *
  * @see TransactionManager
  */
 public interface Transaction {
 
+    /**
+     * Returns a mapping of rows to {@link RowResult}s within {@code tableRef} for the specified {@code rows}, loading
```
jkozlowski, line 50: Meta comment: is this javadoc copied to the generated code? If not, this feels like nobody will actually read this and understand...

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
...
+     * columns according to the provided {@link ColumnSelection}.
+     *
+     * The returned {@link SortedMap} is sorted on the byte order of row keys; the ordering of the input parameter
```
jkozlowski, line 53: Sometimes you say "sorted on the byte order" and sometimes you add "with ascending byte ordering".

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -408,16 +409,16 @@ public void disableReadWriteConflictChecking(TableReference tableRef) {
         hasReads = true;
         Map<byte[], RowColumnRangeIterator> rawResults = keyValueService.getRowsColumnRange(tableRef, rows,
                 columnRangeSelection, getStartTimestamp());
-        Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> postFilteredResults =
-                Maps.newHashMapWithExpectedSize(rawResults.size());
+        ImmutableSortedMap.Builder<byte[], Iterator<Map.Entry<Cell, byte[]>>> postFilteredResults =
+                ImmutableSortedMap.orderedBy(UnsignedBytes.lexicographicalComparator());
```
jkozlowski, line 413: nit: I think in other places we use PtBytes#BYTES_COMPARATOR shim to this method. Just something I noticed.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -381,10 +381,11 @@ public void disableReadWriteConflictChecking(TableReference tableRef) {
         if (Iterables.isEmpty(rows)) {
```
jkozlowski, line 381: Iterables.isEmpty might create an iterator already. So I believe you need to create a copy before this if statement.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
...
+     *
+     * @param tableRef table to load rows from
+     * @param rows rows to be loaded
+     * @param columnSelection columns to load from the given rows
+     * @return a mapping of rows to the columns matching the provided column selection
+     */
     @Idempotent
     SortedMap<byte[], RowResult<byte[]>> getRows(
             TableReference tableRef,
             Iterable<byte[]> rows,
```
jkozlowski, line 69: What about duplicates here? Should we encode this behavior?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
...
+     * by column, with ascending byte ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
+     * input {@code rows}, even if there are rows where no columns match the predicate.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply the column range selection to
+     * @param columnRangeSelection range of columns and batch size to load for each of the rows provided
+     * @return a mapping of rows to cells matching the predicate in the row, following the ordering outlined above
+     * @throws IllegalArgumentException if {@code rows} contains duplicates
```
jkozlowski, line 87: Where is this checked in the code?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
```diff
@@ -698,12 +707,12 @@ public void testColumnRangePagingTransaction_batchingVisitable() {
         verifyMatchingResult(expected, row, columnRange);
 
         columnRange =
-                t.getRowsColumnRange(TEST_TABLE, ImmutableList.of(row), BatchColumnRangeSelection.create(PtBytes.toBytes("col"), PtBytes.EMPTY_BYTE_ARRAY, 1));
+                t.getRowsColumnRange(TEST_TABLE, ImmutableList.of(PtBytes.toBytes("row1")), BatchColumnRangeSelection.create(PtBytes.toBytes("col"), PtBytes.EMPTY_BYTE_ARRAY, 1));
```
jkozlowski, line 710: Maybe you want a separate scoped test for this, as opposed to putting this here?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
```diff
@@ -1309,6 +1318,165 @@ public void testGetRangesPaging() {
...
+        Transaction t = startTransaction();
+        byte[] row0 = row(0);
+        byte[] row1 = row(1);
+        byte[] col0 = column(0);
+        t.put(TEST_TABLE, ImmutableMap.of(Cell.create(row0, col0), value(0), Cell.create(row1, col0), value(1)));
+        t.commit();
+
+        t = startTransaction();
+        SortedMap<byte[], RowResult<byte[]>> readRows =
+                t.getRows(TEST_TABLE, ImmutableList.of(row0, row1), ColumnSelection.all());
```
jkozlowski, line 1359: Do we test somewhere the ColumnSelection and ordering of columns?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -37,41 +37,123 @@
...
 
+    /**
+     * Returns a mapping of requested {@code rows} to corresponding columns from the queried table.
+     * Only columns matching the provided predicate will be returned, and the single predicate provided applies across
+     * all of the rows.
+     *
+     * The returned {@link BatchingVisitable}s are guaranteed to return cells matching the predicate. These are sorted
+     * by column, with ascending byte ordering.
+     *
+     * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
```
jkozlowski, line 80: Is there a test for this behavior?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Random access here refers to `get(x)` though to be fair the next clause kind of renders this one redundant.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: There'll be a next part to this. These comments were written more for AtlasDB devs than for users (so we won't think it's acceptable to e.g. change the order of rows)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: I'll standardise to one of them

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: First: Putting emphasis on the normal case makes sense.

Second question: I don't think so - this method is used in production in a number of places, and in some KVS implementations does something better than "concat the iterators from the Map returning methods"

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: You're right - it does. So if you give an inconsistent iterable we actually rely on the ordering we get the *second* time we get an iterator. We're saying the ordering doesn't matter in that case anyway, so this is fine for the guarantees we've given.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: I added a test, and noted that they are allowed but of course are only represented once in the returned mapping.

jeremyk-91: I realise it's inconsistent with gRCR but can't guarantee we don't have users providing this

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: This is in the builder of the ImmutableSortedMap. I'm not sure if there are existing tests, I added one.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
jeremyk-91: Fair enough, added one.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractTransactionTest.java
jeremyk-91: Not that I'm aware of, I added one test.

jeremyk-91:
Probably the addition of tests and some fairly nuanced semantics made this bigger than expected (the 50-75 lines of docs in Transaction are probably particularly dense)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: There wasn't, now there is (also added one for the other Map-returning method)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jkozlowski: Cool

PR #4706, opened 2020-04-14
Expose flag to use dialogue clients in production

ferozco:
**Goals (and why)**:
atlas currently relies on c-j-r clients, which use okhttp under the hood. We have plenty of complaints with c-j-r's stack of okhttp, feign and retrofit. Rather than investing in fixing the laundry list of problems with conjure-java-runtime, all effort is being directed into making dialogue the best possible client for palantir. Specifically, we're hoping for better observability, throughput, latency and wire-format correctness (see Dialogue rollout quip).

cc @carterkozak @iamdanfox . Happy to discuss here or in #dev-rpc.

In tests I've configured the transaction manager to use the dialogue clients

dialogue contains many of the same ideas as c-j-r, but all fresh implementations, so we're probably going to see some subtle emergent effects/odd behaviour as a result of the new implementation (e.g. the queue and concurrency limiters are both slightly different). Only products with an appetite for risk should opt-in to this (ideally giving us a heads up in #dev-rpc).


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureClientFactory.java
```diff
@@ -0,0 +1,121 @@
...
+import com.palantir.conjure.java.client.config.ClientConfiguration;
+import com.palantir.conjure.java.client.jaxrs.JaxRsClient;
+import com.palantir.conjure.java.okhttp.HostEventsSink;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import com.palantir.tritium.Tritium;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+public class ConjureClientFactory {
```
iamdanfox, line 37: would prefer this to not be public otherwise people are gonne start depending on it by accident instead of WC's ConjureClientFactory.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureClientFactory.java
```diff
@@ -0,0 +1,121 @@
...
+    private final UserAgent userAgent;
+    private final HostEventsSink hostEventsSink;
+    private final DialogueClientFactory dialogueClientFactory;
+
+    public ConjureClientFactory(
+            Supplier<ServerListConfig> clientConfig,
+            TaggedMetricRegistry taggedMetricRegistry,
+            UserAgent userAgent,
+            HostEventsSink hostEventsSink,
+            boolean useDialogueJaxrs) {
```
iamdanfox, line 53: can we call this something like `useDialogueFeignShim`?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -998,7 +1029,12 @@ private static LockAndTimestampServices getLockAndTimestampServices(
                 RemoteLockServiceAdapter.create(lockRpcClient, timelockNamespace));
 
         ConjureTimelockService conjureTimelockService = new BlockingSensitiveConjureTimelockService(
-                BlockingAndNonBlockingServices.create(creator, ConjureTimelockService.class));
+                ImmutableBlockingAndNonBlockingServices.<ConjureTimelockService>builder()
+                        .nonBlocking(clientFactory.client(ConjureTimelockService.class))
```
iamdanfox, line 1033: this means we'd end up using the feign shim right? is the plan still to get to the full zero-reflection codegen version? (presumably would need to open-source carter's ConjureServiceAdapter proxy thingy.)

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -239,6 +243,11 @@ boolean allSafeForLogging() {
         return false;
     }
 
+    @Value.Default
+    boolean useDialogue() {
```
iamdanfox, line 247: might want to call this `enableDialogue` so that if the atlasdb folks want to have some auto-reverting logic in here we couldn't get into the state where config said "useDialogue" but old clients were being used?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/CachedComposingSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+        this.inputSupplier = inputSupplier;
+        this.outputFunction = outputFunction;
+    }
+
+    public <V> CachedComposingSupplier<U, V> map(Function<U, V> map) {
+        return new CachedComposingSupplier<>(this, map);
+    }
+
+    @Override
+    public synchronized U get() {
```
iamdanfox, line 43: I put together a lock-free version of this because I was concerned about the SupplierChannel hitting a `synchronized` block: https://internal-github/foundry/aaaa-client-java/pull/910/files

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureClientFactory.java
ferozco: Ya it would be nice to keep it package private, but we need to make it visible for a couple packages. I'll rename it `AtlasClientFactory` to avoid any confusion.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
ferozco: Ya originally the plan was to go straight to full dialogue, but I wanted to avoid copying all of carter's proxy so I went for the shim as a starting point

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/CachedComposingSupplier.java
ferozco: haha nice thanks! Carter and I were chatting about pulling out Refreshable so that we can get rid of all of these approximations

iamdanfox: amusingly it seems like we did in fact start this process: https://github.com/palantir/refreshable

jkozlowski: Great, can we get that in, would love to get rid of all the permutations of this code.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski, line 247: This class is directly changed in code, meaning that rollout might be harder?

When do rollouts like these, we make changes to AtlasDbRuntimeConfig (which clients wire into their service's config), and specifically I think something like RemotingClientConfig (which was used to rollout http-remoting -> CJR switch) might be useful to you here. I am happy to start with a DialogueClientConfig that has a boolean flag to just enable it, if we use it for testing in a dev environment. Having a dedicated DialogueClientConfig should hopefully make it easier to remove later.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -361,7 +370,8 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                 managedTimestampSupplier,
                 atlasFactory.getTimestampStoreInvalidator(),
                 userAgent(),
-                lockDiagnosticInfoCollector());
+                lockDiagnosticInfoCollector(),
```
jkozlowski, line 373: If it's in config, you won't need to thread it manually like this.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -985,7 +1006,17 @@ private static LockAndTimestampServices getLockAndTimestampServices(
             Supplier<RemotingClientConfig> remotingConfigSupplier,
             UserAgent userAgent,
             String timelockNamespace,
-            Optional<ClientLockDiagnosticCollector> lockDiagnosticCollector) {
+            Optional<ClientLockDiagnosticCollector> lockDiagnosticCollector,
+            boolean useDialogue) {
+
+        ConjureClientFactory clientFactory = new ConjureClientFactory(
```
jkozlowski, line 1012: I'm assuming you need to wire the flag here?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
ferozco: Hmm not sure I understand what the issue with adding the flag to the transactionManager directly. It seems just as easy to remove/change the default and avoids an extra level of indirection

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/CachedComposingSupplier.java
ferozco: Sure, i'll update to use a refreshable

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
ferozco: It is: https://github.com/palantir/atlasdb/pull/4706/files/390da9fb497272cb708864adff500acddd9e7098#diff-4d9676a5ce96e75de18356e3d09145f6R1018

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -197,11 +204,24 @@
...
     private static final int LOGGING_INTERVAL = 60;
     private static final Logger log = LoggerFactory.getLogger(TransactionManagers.class);
+    private static final Duration REFRESH_INTERVAL = Duration.ofSeconds(1);
 
     public static final LockClient LOCK_CLIENT = LockClient.of("atlas instance");
 
     abstract AtlasDbConfig config();
 
+    @Value.Default
+    Refreshable<Optional<AtlasDbRuntimeConfig>> runtimeConfig() {
```
jkozlowski, line 214: So this is great, for internal ski product that creates a lot of TransactionManagers I can pass this directly and hence not create a threadpool per TransactionManager.

But if it's not passed in, you need to make sure that this threadpool is cleaned up. TransactionManagers have a #close method, so I think you want to create this threadpool somewhere further down where it could be cleaned up.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -197,11 +204,24 @@
...
+        return Refreshables.create(runtimeConfigSupplier()::get, exception -> {
+            log.error("Failed to load runtime config",
+                    exception);
+        }, PTExecutors.newSingleThreadScheduledExecutor(), REFRESH_INTERVAL);
+    }
+
+    /**
+     * @deprecated use {@link #runtimeConfig} instead.
+     */
+    @Deprecated
```
jkozlowski, line 224: Can you add a #check to make sure both are never provided?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Refreshables.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import com.palantir.refreshable.DefaultRefreshable;
+import com.palantir.refreshable.Refreshable;
+
+final class Refreshables {
```
jkozlowski, line 31: Can you push this class into the refreshables repo please? I think I will need the same code in internal ski product.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Ok, I didn't explain properly.

What it means is that you need any application that picks this up to push special code that sets this flag.

They can either:
1. Risk that everything will be ok and just hard set it. If they hit issues, they need to rollback in production, then revert your change and cut a new release so they're not blocking develop.
2. Add custom configuration block to their config so that you can switch this via config.

I don't want anyone to do 1, or 2, because AtlasDbRuntimeConfiguration is how changeable parameters to TransactionManagers should be handled.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Refreshables.java
ferozco: @iamdanfox you felt strongly about removing this method in the first place, what do you think?

iamdanfox: this is only used here to adapt from a supplier to a refreshable right? I don't think you need this in internal ski product.

ferozco: no ski product would not need this, but any library that would like to create refreshables would.

iamdanfox: I think I actually do feel strongly about keeping this out. I don't want open-sourcing refreshable to mean that we find a bazillion usages of refreshable everywhere, I'm ok with there being a little friction in creating entirely new ones, as I still want them to _primarily_ come from WC.

jkozlowski: Ack, I forgot we can adapt the WC one.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Refreshables.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private Refreshables() {}
+
+    /** Creates a refreshable that updates based on the result of invoking the {@link Callable} at the given period. */
+    @SuppressWarnings("FutureReturnValueIgnored")
+    static <T> Refreshable<T> create(
+            Callable<T> callable,
+            Consumer<Throwable> exceptionHandler,
```
jkozlowski, line 39: Can we make this slightly more generic, it will end up being called once

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/SupplierChannel.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.palantir.dialogue.Channel;
+import com.palantir.dialogue.Endpoint;
+import com.palantir.dialogue.Request;
+import com.palantir.dialogue.Response;
+
+final class SupplierChannel implements Channel {
+    private final Supplier<Channel> channelSupplier;
+
+    SupplierChannel(Supplier<Channel> channelSupplier) {
```
jkozlowski, line 18: This is same as in internal ski product's library? Should this live upstream?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasClientFactory.java
```diff
@@ -0,0 +1,121 @@
...
+            StaticClientConfiguration.builder().build();
+
+    private final boolean useDialogueJaxrs;
+    private final Refreshable<Optional<ServerListConfig>> clientConfig;
+    private final TaggedMetricRegistry taggedMetricRegistry;
+    private final UserAgent userAgent;
+    private final HostEventsSink hostEventsSink;
+    private final DialogueClientFactory dialogueClientFactory;
+
+    public AtlasClientFactory(
```
jkozlowski, line 49: Have you seen ServiceCreator and AtlasDbHttpClients? I feel like this should be folded somewhere there, instead of creating another codepath through which to go, just to create ConjureTimelockService.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasClientFactory.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private <T> T jaxrsInternal(Class<T> serviceClass, StaticClientConfiguration staticClientConfig) {
+        Function<ServerListConfig, T> clientFactory = jaxrsFactory(serviceClass,
+                staticClientConfig);
+        Supplier<Optional<T>> client = clientConfig.map(conf -> conf.map(clientFactory));
+        T alwaysThrowingService = serviceProxy(serviceClass, (proxy, method, args) -> {
+            throw new SafeIllegalStateException("Service not configured");
+        });
+
+        return serviceProxy(serviceClass, (proxy, method, args) -> {
```
jkozlowski, line 89: All this exists in AtlasDbHttpClients, and I think it has additional protections against OkHttp bugs we had to put in.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasClientFactory.java
jkozlowski: Also I think in that path, ConjureJavaRuntimeTargetFactory has a whole bunch of overrides to do with QoS and whatnot.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Refreshables.java
```diff
@@ -0,0 +1,69 @@
...
+import com.palantir.refreshable.DefaultRefreshable;
+import com.palantir.refreshable.Refreshable;
+
+final class Refreshables {
+
+    private Refreshables() {}
+
+    /** Creates a refreshable that updates based on the result of invoking the {@link Callable} at the given period. */
+    @SuppressWarnings("FutureReturnValueIgnored")
+    static <T> Refreshable<T> create(
```
jkozlowski, line 37: In fact, I think we should split a smaller PR that switches to Refreshables: this should then be propagated down to ConjureJavaRuntimeTargetFactory, and would allow us to remove CachedComposedSupplier.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueClientFactory.java
```diff
@@ -0,0 +1,174 @@
...
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+final class DialogueClientFactory {
+    private static final DefaultConjureRuntime DIALOGUE_RUNTIME =
+            DefaultConjureRuntime.builder().build();
+    private static final String SERVICE_NAME = "lock";
+    private final TaggedMetricRegistry taggedMetricRegistry;
+    private final Refreshable<Optional<ServerListConfig>> config;
+    private final UserAgent userAgent;
+    private volatile ConcurrentHashMap<CacheKey, DialogueChannel> cachedChannels = new ConcurrentHashMap<>();
```
jkozlowski, line 51: Why is this volatile?

jkozlowski: Why do we have this map in the first place? Is this so that you can reuse the channels if we have, say, multiple services on the same host we talk to (so ConjureTimelockService and something else in the future?)

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueClientFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                .userAgent(userAgent)
+                .build();
+    }
+
+    static boolean isDialogue(Class<?> serviceInterface) {
+        return getStaticOfMethod(serviceInterface).isPresent();
+    }
+
+    private static String toChannelName(CacheKey cacheKey) {
+        return "atlasdb-dialogue-channel-" + SERVICE_NAME + staticConfigToString(cacheKey.staticConfig());
```
jkozlowski, line 121: Does similar code exist in internal webserver?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/StaticClientConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+package com.palantir.atlasdb.http.v2;
+
+import java.util.Optional;
+
+import org.immutables.value.Value;
+
+import com.palantir.conjure.java.client.config.ClientConfiguration;
+
+@Value.Immutable
+public interface StaticClientConfiguration {
```
jkozlowski, line 26: I think this is also similar to AuxiliaryRemotingParameters?

versions.props
```diff
@@ -99,7 +100,13 @@ org.rocksdb:rocksdbjni = 6.4.6
...
 org.freemarker:freemarker = 2.3.29
+com.palantir.refreshable:refreshable = 0.1.0
 
 com.google.errorprone:error_prone_refaster = 2.3.3
 com.palantir.baseline:baseline-refaster-testing = 2.17.0
 com.google.truth:truth = 1.0
+
+
+# conflict resolution
+org.apache.httpcomponents:httpclient = 4.5.12
```
jkozlowski, line 111: Where is the conflict coming from? Is there a better way to resolve it? it would be good to instruct gradle to just take the dialogue's version.

jkozlowski: Unless this is coming from some Oracle related code, in which case we'll need to check those versions make sense.

versions.props
```diff
@@ -27,8 +27,9 @@ com.palantir.config.crypto:encrypted-config-value-module = 2.1.0
 com.palantir.conjure:conjure = 4.10.1
 com.palantir.conjure.java.api:* = 2.9.0
 com.palantir.conjure.java:* = 5.11.2
-com.palantir.dialogue:* = 1.2.0
-com.palantir.conjure.java.runtime:* = 4.70.1
+com.palantir.dialogue:* = 1.11.4
+com.palantir.conjure.java.runtime:* = 5.3.0
+com.github.tomakehurst:wiremock-standalone = 2.26.3
```
jkozlowski, line 32: We already have this version slightly higher, this library is used for some tests.

jkozlowski:
Ok, I think I went through everything. What I see is this:

1. Send the Refreshable switcheroo in a separate PR and propagate it down to AtlasDbHttpClients and ConjureJavaRuntimeTargetFactory. This should hopefully get rid of CachedTransformingSupplier.
2. Move all the config down to AtlasDbRuntimeConfig: I don't want the manual threading of a single boolean flag down to the appropriate layer. AtlasDbRuntimeConfig is the appropriate place for it.
3. Combine your dialogue bits with AtlasDbHttpClients (basically make it switch between ConjureJavaRuntimeTargetFactory and your DialogueClientFactory. As I understand the code, your AtlasClientFactory is able to detect if the class is Dialogueable, so I don't see blockers if we can only really do ConjureTimelockService.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueClientFactory.java
jkozlowski: We need to make sure that this is closed whenever TransactionManager is closed.

ferozco:
Caught up with @jkozlowski offline and we've agreed to the following work:

1. Wire up all closable state (Refreshable executor, clientFactory, etc.) to TM close method.
1. Ensure all workaround for okhttp bugs (ex: https://github.com/palantir/atlasdb/blob/a7ef819be080c85b64edc06673095fbd084b4d01/atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java#L109) are applied on the CJR path of the client factory.
1. Keep proposed flag on TM for using Dialogue clients to avoid operators inadvertently swapping services to use the new clients hence minimizing the blast zone of any issues.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -279,6 +296,12 @@ LockWatchingCache lockWatchingCache() {
         thread.start();
     };
 
+    @Value.Check
+    protected void check() {
+        Preconditions.checkState(runtimeConfigSupplier().isPresent() ^ runtimeConfig().isPresent(),
+                "Either Refreshable or Supplier of runtime config must be provide, but not both");
```
jkozlowski, line 302: nit: "must be provided"

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -301,15 +324,17 @@ public static TransactionManager createInMemory(Set<Schema> schemas) {
...
                 .serializable();
     }
 
     @JsonIgnore
-    @Value.Derived
+    @Value.Lazy
     public TransactionManager serializable() {
         List<AutoCloseable> closeables = Lists.newArrayList();
 
+
```
jkozlowski, line 337: whitespace

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: FIxed

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Fixed

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -197,15 +205,19 @@
...
-    Supplier<Optional<AtlasDbRuntimeConfig>> runtimeConfigSupplier() {
-        return Optional::empty;
-    }
+    abstract Optional<Refreshable<Optional<AtlasDbRuntimeConfig>>> runtimeConfig();
+
+    /**
+     * @deprecated use {@link #runtimeConfig} instead.
+     */
+    @Deprecated
+    abstract Optional<Supplier<Optional<AtlasDbRuntimeConfig>>> runtimeConfigSupplier();
```
jkozlowski, line 220: Hmm, I wonder if this would break people?

jkozlowski: Nope, shouldn't break, the builder should have the same method as begore.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -337,13 +361,21 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         MetricsManager metricsManager = MetricsManagers.of(globalMetricsRegistry(), globalTaggedMetricRegistry());
 
         AtlasDbRuntimeConfig defaultRuntime = AtlasDbRuntimeConfig.defaultRuntimeConfig();
-        Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier =
-                () -> runtimeConfigSupplier().get().orElse(defaultRuntime);
+        Refreshable<AtlasDbRuntimeConfig> runtime = runtimeConfig()
+                .orElseGet(() -> {
+                    ScheduledExecutorService refreshableExecutor = PTExecutors.newSingleThreadScheduledExecutor();
```
jkozlowski, line 366: From my reading of the code, this will not get closed on TransactionManager#close. As the code is written, the closeables list is only cleaned up on exception.

jkozlowski: See https://github.com/palantir/atlasdb/pull/4723

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasClientFactory.java
```diff
@@ -0,0 +1,135 @@
...
+        } else if (useDialogueJaxrs) {
+            return dialogueClientFactory.jaxrsClient(serviceClass, staticConfig);
+        }
+        return jaxrsInternal(serviceClass, staticConfig);
+    }
+
+    private <T> T jaxrsInternal(Class<T> serviceClass, StaticClientConfiguration staticClientConfig) {
+        Function<ServerListConfig, T> clientFactory = jaxrsFactory(serviceClass,
+                staticClientConfig);
+        Supplier<Optional<T>> client = clientConfig.map(maybeConf ->
```
jkozlowski, line 86: Can you explain why you couldn't use the AtlasDbHttpClients codepath? It's not just ReplaceIfExceptionMatchingProxy that's there, there's also FastFailoverProxy that you need to wire up. I suspect it also needs to be wired up in the Dialogue version.

jkozlowski:
Found another interesting corollary to using a boolean switch in TransactionManagers: with the current setup, we'd need to create the same flag in internal live-migration library, since that is actually how most clients create AtlasDB transaction managers.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasClientFactory.java
jkozlowski: Changed it over. We should have this whole codepath be either on or off.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueStaticClientConfigurations.java
```diff
@@ -0,0 +1,71 @@
...
+    // Should not be reduced below 65 seconds to support workflows involving locking.
+    private static final Duration BLOCKING_READ_TIMEOUT = Duration.ofSeconds(65);
+
+    static ClientConfiguration apply(StaticClientConfiguration staticClientConfig, ServerListConfig serverConfig) {
+        ClientConfiguration.Builder mixed = ClientConfiguration.builder()
+                .from(ClientConfigurations.of(toServiceConfiguration(serverConfig)))
+                .connectTimeout(CONNECT_TIMEOUT)
+                .enableGcmCipherSuites(true)
+                .enableHttp2(true)
+                .nodeSelectionStrategy(NodeSelectionStrategy.PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE);
```
carterkozak, line 50: We may want to add this as an optional field on the static configuration if the client factory is used for non-paxos targets.

jkozlowski: I don't think we talk to anything other than timelock. Maybe timelock to timelock might be affected, but I think that has client per URL

versions.props
jeremyk-91: These come from cassandra-thrift (including latest version). I'm checking with @leonz if this is safe and/or what we can do here.

jeremyk-91: Turns out this is tricky to bump. But given this combination has _already_ been running in a fair few places and the [discrepancy](https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt) is small I think we can take on this risk.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueClientFactory.java
```diff
@@ -0,0 +1,249 @@
...
+import com.palantir.dialogue.core.DialogueChannel;
+import com.palantir.dialogue.hc4.ApacheHttpClientChannels;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import com.palantir.refreshable.Refreshable;
+import com.palantir.tritium.Tritium;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+
+final class DialogueClientFactory implements AutoCloseable {
```
iamdanfox, line 49: @jeremyk-91 FYSA I've got a dialogue client factory coming in https://github.com/palantir/dialogue/pull/694 which I think will replace this.

iamdanfox: Ok 1.35.0 is out go go go https://circleci.com/workflow-run/add38174-7f42-4100-bc21-e28e80bddc58

atlasdb-console/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
     compile group: 'org.codehaus.groovy', name: 'groovy'
     compile group: 'org.codehaus.groovy', name: 'groovy-groovysh'
     compile group: 'org.codehaus.groovy', name: 'groovy-json'
+    compile 'com.palantir.refreshable:refreshable'
```
iamdanfox, line 12: implementation pls?

iamdanfox:
Closing out because @jkong incorporated these commits into another PR

PR #4707, opened 2020-04-14
Executor quality of life and instrumentation

carterkozak:
**Goals (and why)**:

Preparation for #4701 This will provide additional data to discover and remediate thread bloat.

* PTExecutors thread factory uses daemon threads by default. Most consumers prefer daemon threads to avoid propping up JVMs that want to die.
* PTExecutors.newCachedThreadPool returns ExecutorService instances instead of ThreadPoolExecutor concrete types. This is an ABI break. Please use WC executor factories!
* Instrument NamedThreadFactory using the singleton TaggedMetricRegistry instance to collect data describing thread bloat.

Existing test coverage is sufficient.

¯_(ツ)_/¯


lock-impl/src/test/java/com/palantir/lock/impl/ClientAwareLockTest.java
```diff
@@ -48,8 +47,7 @@
 @ShouldRetry
 public final class ClientAwareLockTest {
 
-    private static final ExecutorService executor = PTExecutors.newCachedThreadPool(
-            new NamedThreadFactory(ClientAwareLockTest.class.getName(), true));
+    private static final ExecutorService executor = PTExecutors.newCachedThreadPool("ClientAwareLockTest");
```
jeremyk-91, line 50: nit: Don't feel strongly about which way this falls, but we should probably be consistent on whether we prefer the string literal or `ThisClass.class.getName()` for these executors.

changelog/@unreleased/pr-4707.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
```
jeremyk-91, line 1: We could mark this as `break` if we want to make the ABI break more obvious.

carterkozak: Good call, version numbers are cheap :-)

lock-impl/src/test/java/com/palantir/lock/impl/ClientAwareLockTest.java
carterkozak: that's fair, I'll revert these out to use class.getName(), no reason to churn on these in this change.

carterkozak:
Thanks @jeremyk-91!

PR #4708, opened 2020-04-14
[PDS-111849] Rolling in the Deep, Part 5: Only One Dedicated Executor

jeremyk-91:
**Goals (and why)**:
- Understand why the number of acceptor threads on a leader at a deployment went up by >500 when a non-leader was bounced, when it should really only increase by 100.
- The original implementation in this chain of PRs spins up dedicated executors for **every** time the acceptor factory is called, which is every time a new client is presented (see `LeadershipComponents#createNewLeadershipContext`). The previous PR thus limited thread explosion from infinite to (100 * num_clients). In practice because of autobatching and the like it generally won't be anywhere near that cap, but I think we wanted 100 _in general_, not 100 per client.

- Rewire things so that each remote has a canonical executor for leadership acceptor operations across clients. This is much closer to the intention of the original fix I had.

- Existing tests pass, including the stress test.
- I validated through the debugger that you only get one creation of these thread pools per remote now, while previously we would stop at multiple points to create the bounded executors.

- Is there anything wired wrongly? The existing tests passing and debugger validation give me some confidence, but always worth a check.
- Will we run into problems with rejected executions? I think rejecting is actually the desired behaviour, and if the cluster is stable we should get a response from the other node (and if it's not we shouldn't proceed).

**Where should we start reviewing?**: SingleLeaderNetworkClientFactories

**Priority (whenever / two weeks / yesterday)**: ideally earlier this week, this is part of the big issue


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/WithDedicatedExecutor.java
```diff
@@ -0,0 +1,40 @@
...
+package com.palantir.atlasdb.timelock.paxos;
+
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Function;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface WithDedicatedExecutor<T> {
```
gmaretic, line 26: ❤️

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -46,7 +50,23 @@
     public abstract PaxosResourcesFactory.TimelockPaxosInstallationContext context();
 
     @Value.Parameter
-    public abstract TaggedMetricRegistry metrics();
+    public abstract MetricsManager metrics();
+
+    @Value.Derived
+    Map<String, ExecutorService> dedicatedExecutors() {
+        Set<String> remoteUris = context().remoteUris();
+        int executorIndex = 0; // Acceptable for Set, we just want to differentiate the elements
```
gmaretic, line 58: we don't need this comment

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -113,7 +138,17 @@
         return createInstrumentedRemoteProxies(clazz, shouldRetry).values().collect(Collectors.toList());
     }
 
-    private <T> KeyedStream<HostAndPort, T> createInstrumentedRemoteProxies(Class<T> clazz, boolean shouldRetry) {
+    private <T> List<WithDedicatedExecutor<T>> createInstrumentedRemoteProxiesAndAssignDedicatedExecutors(
+            Class<T> clazz,
+            boolean shouldRetry) {
+        return createInstrumentedRemoteProxies(clazz, shouldRetry)
```
gmaretic, line 144: Nice!

PR #4710, opened 2020-04-15
Remove qos-service-impl.

jkozlowski:
**Goals (and why)**:


gmaretic:
Probably should also remove docs/source/services/qos_service/index.rst

PR #4711, opened 2020-04-16
[ETE] Batching Txns, Attempt 2

Jolyon-S:
Edit: I'm merging in https://github.com/palantir/atlasdb/pull/4750 and https://github.com/palantir/atlasdb/pull/4748 to this branch, so they need to be merged before we do final merge on this.

**Goals (and why)**:
Start a batch of txns in one call.

Added a batch start transaction endpoint to the transaction manager. Guarantees to return either a full batch or throw an exception.

Have not implemented any tests yet

Is this the right approach? Don't worry about details for now.

`SnapshotTransactionManager`, `TransactionStarter`

whenever - this is a draft PR, so please only review decomp


lock-api-objects/src/main/java/com/palantir/lock/v2/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,61 @@
...
+
+package com.palantir.lock.v2;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+
+public final class ExceptionHandlingRunner implements AutoCloseable {
```
Jolyon-S, line 25: This class is copied from bottom of SnapshotTransactionManager - probably want to merge the two (if I can find a place where it makes sense to live in the project hierarchy).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -78,6 +89,21 @@ public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransacti
...
+    private <T, R> T checkAndUpdateLowerBoundBatch(Supplier<T> timestampContainerSupplier,
+            Function<T, List<R>> responseExtractor,
+            ToLongFunction<R> lowerBoundExtractor,
+            ToLongFunction<R> upperBoundExtractor) {
+        long threadLocalLowerBound = lowerBound.get();
+        T timestampContainers = timestampContainerSupplier.get();
+        List<R> responses = responseExtractor.apply(timestampContainers);
+        responses.forEach(timestampContainer -> {
+            checkTimestamp(threadLocalLowerBound, lowerBoundExtractor.applyAsLong(timestampContainer));
+            updateLowerBound(upperBoundExtractor.applyAsLong(timestampContainer));
```
Jolyon-S, line 102: I need to spend a bit of time confirming what these do and determine the most efficient way to do this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
+                    (response, condition) -> wrapResponse(condition, response)).collect(Collectors.toList());
+            return responses.successful(transactions);
+        }
+    }
+
+    private TransactionAndImmutableTsLock wrapResponse(PreCommitCondition condition,
+            StartIdentifiedAtlasDbTransactionResponse transactionResponse) {
+        LockToken immutableTsLock = transactionResponse.immutableTimestamp().getLock();
+        long immutableTs = transactionResponse.immutableTimestamp().getImmutableTimestamp();
+        recordImmutableTimestamp(immutableTs);
```
Jolyon-S, line 202: We can probably optimise this in the batched case (only update once), if that is a performance concern.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -86,11 +89,31 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
     public ListenableFuture<R> apply(T argument) {
         Preconditions.checkState(!closed, "Autobatcher is already shut down");
         DisruptorFuture<R> result = new DisruptorFuture<R>(safeLoggablePurpose);
-        buffer.publishEvent((refresh, sequence) -> {
+        buffer.publishEvent(createTranslator(argument, result));
+        return result;
+    }
+
+    public List<ListenableFuture<R>> applyBatch(List<T> arguments) {
+        Preconditions.checkState(!closed, "Autobatcher is already shut down");
```
jkozlowski, line 97: Refactor this check to a method.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -86,11 +89,31 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
...
+        EventTranslator<DefaultBatchElement<T, R>>[] translators = arguments.stream().map(argument -> {
+            DisruptorFuture<R> result = new DisruptorFuture<>(safeLoggablePurpose);
+            EventTranslator<DefaultBatchElement<T, R>> translator = createTranslator(argument, result);
+            results.add(result);
+            return translator;
+        }).toArray(EventTranslator[]::new);
+
+        buffer.publishEvents(translators);
+
+        return results;
```
jkozlowski, line 109: Tests for this method, and I suppose for this whole class? The batching code is tricky, so definitely need tests for batching code.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -86,11 +89,31 @@ private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
     public ListenableFuture<R> apply(T argument) {
         Preconditions.checkState(!closed, "Autobatcher is already shut down");
         DisruptorFuture<R> result = new DisruptorFuture<R>(safeLoggablePurpose);
-        buffer.publishEvent((refresh, sequence) -> {
+        buffer.publishEvent(createTranslator(argument, result));
+        return result;
+    }
+
+    public List<ListenableFuture<R>> applyBatch(List<T> arguments) {
```
jkozlowski, line 96: Would it be a lot of overhead for removing the non-batched method here?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -67,6 +68,13 @@ public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransacti
                 r -> r.startTimestampAndPartition().timestamp());
     }
 
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
```
jkozlowski, line 72: Again, would it make sense to only keep the batched method? count=1 should be simple to handle.

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    /**
+     * Calling close with no failures should be a no-op; equally, calling close multiple times will re-throw a runtime
+     * exception with the same suppressed errors (plus any additional errors suppressed since the last close call).
+     */
+    @Override
+    public void close() {
+        if (!failures.isEmpty()) {
```
jkozlowski, line 65: Running close should be enforced to only ever happen once. Also you should probably not allow adding more callbacks after close is called.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockService.java
```diff
@@ -67,6 +68,13 @@ public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransacti
...
+        return checkAndUpdateLowerBound(() -> delegate.startIdentifiedAtlasDbTransactionsBatch(count),
+                StartIdentifiedAtlasDbTransactionResponseBatch::getMinTimestamp,
+                StartIdentifiedAtlasDbTransactionResponseBatch::getMaxTimestamp);
+    }
+
     private <T> T checkAndUpdateLowerBound(Supplier<T> timestampContainerSupplier,
             ToLongFunction<T> lowerBoundExtractor,
```
jkozlowski, line 79: 2 arguments of the same type, wink wink 😉

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
         } catch (Throwable e) {
             timelockService.tryUnlock(ImmutableSet.of(transactionResponse.immutableTimestamp().getLock()));
             throw Throwables.rewrapAndThrowUncheckedException(e);
         }
     }
 
+    @Override
+    public List<TransactionAndImmutableTsLock> setupRunTaskBatchWithConditionThrowOnConflict(
+            List<PreCommitCondition> conditions) {
+        Preconditions.checkArgument(!conditions.isEmpty(), "Trying to start an empty batch");
```
jkozlowski, line 182: Generally it's nicer to allow fall through like this (so allow empty list), I think this is what the usage of this in internal ski product assumed.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
+        try (StartIdentifiedAtlasDbTransactionResponseBatch responses =
+                timelockService.startIdentifiedAtlasDbTransactionsBatch(conditions.size())) {
+            if (responses.size() != conditions.size()) {
```
jkozlowski, line 185: This feels like a weird thing to check, the method you call should not return you a small response, that doesn't feel like a great API.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
+                throw new TransactionBatchFailedRetriableException(
```
jkozlowski, line 186: Shouldn't this method or something lower from here retry this instead? Feels like a bad thing to ask the users of this API to handle.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
jkozlowski: @jeremyk-91 to help verify; also probably need metrics to help answer this

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
-            return TransactionAndImmutableTsLock.of(transaction, immutableTsLock);
+            return wrapResponse(condition, transactionResponse);
         } catch (Throwable e) {
             timelockService.tryUnlock(ImmutableSet.of(transactionResponse.immutableTimestamp().getLock()));
             throw Throwables.rewrapAndThrowUncheckedException(e);
         }
     }
 
+    @Override
+    public List<TransactionAndImmutableTsLock> setupRunTaskBatchWithConditionThrowOnConflict(
```
jkozlowski, line 180: Thoughts on returning something like TransactionAndImmutableTsLock, but that additionally has some sort of close method that users of this can call instead of using #finishRunTaskWithLockThrowOnConflict?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -433,6 +434,17 @@ default LockWatchManager getLockWatchManager() {
...
 
+    /**
+     * This method exposes the ability to start a batch of transactions with pre-commit conditions in a single call.
+     * The result is similar to calling {@link #setupRunTaskWithConditionThrowOnConflict(PreCommitCondition)}, except
+     * the code has been specifically optimised with respect to batching and error handling.
+     *
+     * @return a list of transactions and their associated immutable timestamp locks for the tasks
+     */
+    @Timed
+    List<TransactionAndImmutableTsLock> setupRunTaskBatchWithConditionThrowOnConflict(
```
jkozlowski, line 445: Should be @Deprecated

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -433,6 +434,17 @@ default LockWatchManager getLockWatchManager() {
     @Timed
     TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
 
+    /**
+     * This method exposes the ability to start a batch of transactions with pre-commit conditions in a single call.
+     * The result is similar to calling {@link #setupRunTaskWithConditionThrowOnConflict(PreCommitCondition)}, except
+     * the code has been specifically optimised with respect to batching and error handling.
```
jkozlowski, line 440: Please document that PreCommitConditions do not get cleaned up on errors.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -433,6 +434,17 @@ default LockWatchManager getLockWatchManager() {
     @Timed
     TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
```
jkozlowski, line 435: We need to remove this method, since it is only supposed to be used by internal ski product.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -166,26 +169,50 @@ public TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(Pr
...
-            Supplier<Long> startTimestampSupplier = Suppliers.ofInstance(
-                    transactionResponse.startTimestampAndPartition().timestamp());
-
-            Transaction transaction = createTransaction(
-                    immutableTs,
-                    startTimestampSupplier,
-                    immutableTsLock,
-                    condition);
-            return TransactionAndImmutableTsLock.of(transaction, immutableTsLock);
+            return wrapResponse(condition, transactionResponse);
```
jkozlowski, line 172: Please check that this code you're moving is covered by extensive tests first, especially around resource cleanup on errors.

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
```diff
@@ -85,6 +85,10 @@ public void registerLock(LockToken token) {
         tokensToRefresh.add(token);
     }
 
+    public void registerLocks(Collection<LockToken> tokens) {
```
jkozlowski, line 88: Again, how bad would it be to only have the batched method?

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -116,6 +117,12 @@ public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransacti
         return runTaskTimed("startIdentifiedAtlasDbTransaction", delegate::startIdentifiedAtlasDbTransaction);
     }
 
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
+        return runTaskTimed("startIdentifiedAtlasDbTransactionsBatch",
```
jkozlowski, line 122: Batched method only.

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -123,6 +125,21 @@ public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransacti
         }
     }
 
+    @Override
+    public StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
```
jkozlowski, line 129: Only batched method

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -70,12 +71,17 @@ static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEve
...
-        try {
-            return autobatcher.apply(null).get();
-        } catch (ExecutionException e) {
-            throw Throwables.throwUncheckedException(e.getCause());
-        } catch (Throwable t) {
-            throw Throwables.throwUncheckedException(t);
+        return AtlasFutures.getUnchecked(autobatcher.apply(null));
+    }
+
+    StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
```
jkozlowski, line 77: Batched only.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -70,12 +71,17 @@ static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEve
     }
 
     StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        try {
-            return autobatcher.apply(null).get();
-        } catch (ExecutionException e) {
-            throw Throwables.throwUncheckedException(e.getCause());
-        } catch (Throwable t) {
-            throw Throwables.throwUncheckedException(t);
+        return AtlasFutures.getUnchecked(autobatcher.apply(null));
```
jkozlowski, line 74: Can someone explain why do we have this bizarre interface here with passing nulls?

jkozlowski: Hmm, I guess we sometimes need to pass an actual argument here.

jkozlowski: I think there's an argument for rewriting this to pass actual number of requested transactions to start? So that internal ski product's disruptors would actually become extremely tiny! Then the batch processor could just count number of transactions requested across the batch and send that to timelock?

1. we would save on a bunch of allocations for futures and lists for args here
2. We would get rid of the weird "null" passing of arguments?

@jeremyk-91 interested in your thoughts

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -70,12 +71,17 @@ static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEve
...
+    }
+
+    StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
+        try (StartIdentifiedAtlasDbTransactionResponseBatch.Builder batchBuilder =
+                new StartIdentifiedAtlasDbTransactionResponseBatch.Builder(response ->
+                        unlock(ImmutableSet.of(response.immutableTimestamp().getLock())))) {
+            List<Void> inputs = IntStream.range(0, count).mapToObj($ -> (Void) null).collect(Collectors.toList());
```
jkozlowski, line 81: Yeah, this is hopefully unnecessary.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -70,12 +71,17 @@ static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEve
...
+            autobatcher.applyBatch(inputs).forEach(
+                    response -> batchBuilder.safeAddToBatch(() -> AtlasFutures.getUnchecked(response)));
+            return batchBuilder.build();
```
jkozlowski, line 84: So in this class we call #getStartTransactionResponses which is obviously already batched. But it's implementation can call timelock multiple times, without any cleanup of resources between calls: @jeremyk-91 does StartIdentifiedAtlasDbTransactionResponse not require any cleanup if we fail mid batch in this method?

lock-api-objects/src/main/java/com/palantir/lock/v2/TimelockService.java
```diff
@@ -53,6 +53,8 @@ default boolean isInitialized() {
 
     WaitForLocksResponse waitForLocks(WaitForLocksRequest request);
 
+    StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count);
```
jkozlowski, line 56: Again, batched only.

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
jkozlowski: Oh wait this is legacy; @jeremyk-91 this is mostly unused now right so don't need to worry about perf?

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -70,12 +71,17 @@ static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEve
...
-            throw Throwables.throwUncheckedException(e.getCause());
-        } catch (Throwable t) {
-            throw Throwables.throwUncheckedException(t);
+        return AtlasFutures.getUnchecked(autobatcher.apply(null));
+    }
+
+    StartIdentifiedAtlasDbTransactionResponseBatch startIdentifiedAtlasDbTransactionsBatch(int count) {
+        try (StartIdentifiedAtlasDbTransactionResponseBatch.Builder batchBuilder =
+                new StartIdentifiedAtlasDbTransactionResponseBatch.Builder(response ->
+                        unlock(ImmutableSet.of(response.immutableTimestamp().getLock())))) {
```
jkozlowski, line 80: What's the cost of this #unlock call here, should it be batched?

jkozlowski:
This review took me ⌛ 57 minutes 🤭

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
Jolyon-S: The purpose of the Batch class is that if this does fail midway, we will call unlock (hence the passing in of the closing callback on line 80), unless you are referring to something else?

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
Jolyon-S: Would you mean that, instead of passing `null`, you pass in say `count`, and then the autobatcher batches several `count`s together into a larger batch and fans those out? If so, that would definitely reduce the size of the autobatcher (or at least, the ring buffer underneath).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
Jolyon-S: Either I'm missing something, or we don't currently perform any retries in the old method - I was following suit in this one (but that can be changed if that is our desired behaviour, i.e. to retry on certain exceptions).

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
jeremyk-91: Fine by me: this is largely used in embedded contexts. There is large internal product, but in any case the algorithm here is not really worse than what people will probably end up writing anyway

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
jeremyk-91: Rewriting to add numbers makes sense - as noted `null` is just because we need to pass _something_. This isn't really a thing raw AtlasDB users would think about as much, but I see how it's useful in ski product (though let's do that in a separate self-contained piece please)

lock-api-objects/src/main/java/com/palantir/lock/v2/StartIdentifiedAtlasDbTransactionResponseBatch.java
```diff
@@ -0,0 +1,118 @@
...
+        private final ExceptionHandlingRunner runner = new ExceptionHandlingRunner();
+        private final Consumer<StartIdentifiedAtlasDbTransactionResponse> cleaner;
+
+        public Builder(Consumer<StartIdentifiedAtlasDbTransactionResponse> cleaner) {
+            this.cleaner = cleaner;
+        }
+
+        public void safeAddToBatch(
+                Supplier<StartIdentifiedAtlasDbTransactionResponse> supplier) {
+            Optional<StartIdentifiedAtlasDbTransactionResponse> response = runner.supplySafely(supplier);
```
jkozlowski, line 95: This isn't entirely safe implementation: if someone later decides to add something other than the response.ifPresent below here, you'll leak. You should somehow tie the below line to the supplySafely

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
jkozlowski: Cool, I added tracking for this, we'll fix this next.

lock-api-objects/src/main/java/com/palantir/lock/v2/StartIdentifiedAtlasDbTransactionResponseBatch.java
Jolyon-S: As in:
`runner.runSafely(() -> {
    BlahResponse x = supplier.get();
    responses.add(x);
});`?

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -433,6 +434,17 @@ default LockWatchManager getLockWatchManager() {
...
 
+    /**
+     * This method exposes the ability to start a batch of transactions with pre-commit conditions in a single call.
+     * The result is similar to calling {@link #setupRunTaskWithConditionThrowOnConflict(PreCommitCondition)}, except
+     * the code has been specifically optimised with respect to batching and error handling.
+     *
+     * @return a list of transactions and their associated immutable timestamp locks for the tasks
```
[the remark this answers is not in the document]
Jolyon-S: Read the javadoc for the old method - was simply mirroring what was said there but in plural form.

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
Jolyon-S: I think so - it is used in quite a few places.

We could instead just have `apply` call `applyBatch` instead.

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/TimestampCorroboratingTimelockServiceTest.java
```diff
@@ -71,14 +72,28 @@ public void getFreshTimestampsShouldFail() {
...
 
-        when(rawTimelockService.startIdentifiedAtlasDbTransaction())
-                .thenReturn(startIdentifiedAtlasDbTransactionResponse);
+        when(rawTimelockService.startIdentifiedAtlasDbTransactionBatch(1))
+                .thenReturn(ImmutableList.of(startIdentifiedAtlasDbTransactionResponse));
 
-        assertThrowsOnSecondCall(() -> timelockService.startIdentifiedAtlasDbTransaction());
+        assertThrowsOnSecondCall(() -> timelockService.startIdentifiedAtlasDbTransactionBatch(1));
     }
+//
```
Jolyon-S, line 82: Test, will fixup later.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -163,29 +168,52 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
-        } catch (Throwable e) {
-            timelockService.tryUnlock(ImmutableSet.of(transactionResponse.immutableTimestamp().getLock()));
-            throw Throwables.rewrapAndThrowUncheckedException(e);
+            return wrapResponseBatch(conditions, responses);
+        } catch (Throwable t) {
+            timelockService.tryUnlock(
+                    responses.stream()
+                            .map(response -> response.immutableTimestamp().getLock())
+                            .collect(Collectors.toSet()));
+            throw Throwables.rewrapAndThrowUncheckedException(t);
```
Jolyon-S, line 182: The interesting thing here is that I've moved away from doing the approach requiring a specialised object, and I believe the reason it is OK to do so is:
1. The batch will either be in one piece from `TransactionStarter` or it won't (and therefore will have thrown itself);
2. If we fail mid-batch, we still unlock all of them - and there is only a single call, so we don't have to handle cases where the closing itself fails midway, because it either fails or succeeds.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/WriteBatchingTransactionService.java
```diff
@@ -123,7 +123,7 @@ public void close() {
      *
      * Retrying does theoretically mean that in the worst case with N transactions in our batch, we may actually
      * require N calls to the database, though this is extremely unlikely especially because of the semantics of
-     * {@link TimelockService#startIdentifiedAtlasDbTransaction()}.
+     * {@link TimelockService#startIdentifiedAtlasDbTransactionBatch(int)}.
```
Jolyon-S, line 126: Can clean up when it comes to it

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -138,16 +131,21 @@ public void close() {
...
 
-            for (int i = 0; i < numTransactions; i++) {
-                batch.get(i).result().set(startTransactionResponses.get(i));
+            int start = 0;
+            int end;
+            for (BatchElement<Integer, List<StartIdentifiedAtlasDbTransactionResponse>> batchElement
+                    : batch) {
+                end = start + batchElement.argument();
+                batchElement.result().set(startTransactionResponses.subList(start, end));
+                start = end;
```
Jolyon-S, line 148: This code is a bit messy, for sure - but more a proof of concept of how it would work.

Should definitely go in its own PR.

lock-impl/src/main/java/com/palantir/lock/impl/LegacyTimelockService.java
```diff
@@ -107,15 +109,17 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
...
-                    TimestampAndPartition.of(getFreshTimestamp(), 0));
+            IntStream.range(0, count).forEach(
+                    $ -> responses.add(StartIdentifiedAtlasDbTransactionResponse.of(lockImmutableTimestamp(),
+                            TimestampAndPartition.of(getFreshTimestamp(), 0))));
+            return responses;
         } catch (RuntimeException | Error throwable) {
             try {
-                unlock(ImmutableSet.of(immutableTimestamp.getLock()));
+                unlock(responses.stream().map(response -> response.immutableTimestamp().getLock()).collect(
+                        Collectors.toSet()));
```
Jolyon-S, line 122: Again, this is now a single call, so there is no handling for what happens if there is a failure half-way through closing, as such a failure cannot occur.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -162,27 +168,40 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
-            return TransactionAndImmutableTsLock.of(transaction, immutableTsLock);
-        } catch (Throwable e) {
-            timelockService.tryUnlock(ImmutableSet.of(transactionResponse.immutableTimestamp().getLock()));
-            throw Throwables.rewrapAndThrowUncheckedException(e);
+            return Streams.zip(
+                    responses.stream(),
+                    conditions.stream(),
+                    (response, condition) -> {
+                        LockToken immutableTsLock = response.immutableTimestamp().getLock();
+                        cleaner.punch(response.startTimestampAndPartition().timestamp());
```
jkozlowski, line 186: Haven't looked at puncher in a while, so unclear if you can do this once for the whole batch same as immutable timestamp.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -153,7 +158,8 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
         checkOpen();
         try {
             TransactionAndImmutableTsLock txAndLock =
-                    runTimed(() -> setupRunTaskWithConditionThrowOnConflict(condition), "setupTask");
+                    runTimed(() -> Iterables.getOnlyElement(
+                            setupRunTaskBatchWithConditionThrowOnConflict(ImmutableList.of(condition))), "setupTask");
```
jkozlowski, line 162: I quite like this, it's just an unwrapping and you're done. You could have a private method for setup*, but I'm not fussed either way.

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
```diff
@@ -112,8 +113,8 @@ public LockImmutableTimestampResponse lockImmutableTimestamp() {
     }
 
     @Override
-    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        return runTaskTimed("startIdentifiedAtlasDbTransaction", delegate::startIdentifiedAtlasDbTransaction);
+    public List<StartIdentifiedAtlasDbTransactionResponse> startIdentifiedAtlasDbTransactionBatch(int count) {
+        return runTaskTimed("startIdentifiedAtlasDbTransactionBatch", () -> delegate.startIdentifiedAtlasDbTransactionBatch(1));
```
jkozlowski, line 117: You need to propagate the count?

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
```diff
@@ -88,16 +91,19 @@ public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
...
     @Override
-    public StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        StartIdentifiedAtlasDbTransactionResponse response = executeOnTimeLock(
-                delegate::startIdentifiedAtlasDbTransaction);
-        lockRefresher.registerLock(response.immutableTimestamp().getLock());
-        return response;
+    public List<StartIdentifiedAtlasDbTransactionResponse> startIdentifiedAtlasDbTransactionBatch(int count) {
+        List<StartIdentifiedAtlasDbTransactionResponse> responses = executeOnTimeLock(
+                () -> delegate.startIdentifiedAtlasDbTransactionBatch(count));
+        lockRefresher.registerLocks(responses
```
jkozlowski, line 102: You're relying on this not throwing. I think it's fine in this refactoring, but need to track fixing this, same as the other bug.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -45,38 +44,32 @@
...
  * service also handles creating {@link LockTokenShare}'s to enable multiple transactions sharing a single immutable
  * timestamp.
- *
+ * <p>
  * Callers of this class should use {@link #unlock(Set)} and {@link #refreshLockLeases(Set)} for returned lock tokens,
  * rather than directly calling delegate lock service.
  */
 final class TransactionStarter implements AutoCloseable {
-    private final DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher;
+    private final DisruptorAutobatcher<Integer, List<StartIdentifiedAtlasDbTransactionResponse>> autobatcher;
```
jkozlowski, line 52: This is a bit sucky, since we're having to allocate Integers, while disruptor is optimised for that and reuses the actual instances of arguments, but I don't think it's worth obsessing about this. @jeremyk-91

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -45,38 +44,32 @@
...
 
-    StartIdentifiedAtlasDbTransactionResponse startIdentifiedAtlasDbTransaction() {
-        try {
-            return autobatcher.apply(null).get();
-        } catch (ExecutionException e) {
-            throw Throwables.throwUncheckedException(e.getCause());
-        } catch (Throwable t) {
-            throw Throwables.throwUncheckedException(t);
-        }
+    List<StartIdentifiedAtlasDbTransactionResponse> startIdentifiedAtlasDbTransactionBatch(int count) {
```
jkozlowski, line 71: Probably want a test what happens when you push 0 here. Also should throw on negative.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -138,16 +131,20 @@ public void close() {
...
             List<StartIdentifiedAtlasDbTransactionResponse> startTransactionResponses =
                     getStartTransactionResponses(lockLeaseService, lockWatchEventCache, numTransactions);
 
-            for (int i = 0; i < numTransactions; i++) {
-                batch.get(i).result().set(startTransactionResponses.get(i));
+            int start = 0;
+            for (BatchElement<Integer, List<StartIdentifiedAtlasDbTransactionResponse>> batchElement
+                    : batch) {
+                int end = start + batchElement.argument();
+                batchElement.result().set(startTransactionResponses.subList(start, end));
```
jkozlowski, line 146: Hmm, just think very carefully about what this subList does here. @jeremyk-91

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
Jolyon-S: I think I won't do that in this PR - we can maybe do that improvement in a separate task.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
Jolyon-S: From the javadoc:
```
Indicate that the given timestamp has just been created. This must be called frequently (preferably on each transaction commit) so that the Cleaner can keep track of the wall-clock/timestamp mapping. If it is never called, semantically nothing goes wrong, but the sweeper won't sweep, since it cannot know what things are old enough to be swept.
```
I interpret this as something we could conceivably do once per batch. @jeremyk-91 to sanity check that.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/service/WriteBatchingTransactionService.java
Jolyon-S, line 126: Just want to check that this simple substitution still makes sense in the context above.

lock-api/src/main/java/com/palantir/lock/client/ProfilingTimelockService.java
Jolyon-S: Good catch - autoreplaced the code and I obviously missed this.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
Jolyon-S: Yes, since all of the returned lists are a view of the original list `startTransactionResponses`, we can get `ConcurrentModificationException`s or other such things if the lists returned are abused (i.e. elements are removed). I don't think this will be a problem given that we don't do that, but we could change this to return a fresh list for each result if it is a concern.

jeremyk-91: @Jolyon-S The tricky thing here is that subList keeps a reference to the original list, so the entire response will remain live in terms of memory as long as any consumer keeps a reference to their chunk of the list. This _looks_ like it would probably be safe from tracing through the code, but I'd prefer to just create a new list in case there's something silly around threads contending over `updateAndGet()` or things like that.

lock-api/src/test/java/com/palantir/lock/client/TransactionStarterTest.java
```diff
@@ -98,49 +101,99 @@ public void after() {
...
+        List<List<StartIdentifiedAtlasDbTransactionResponse>> responses = requestBatches(sizes);
+        Streams.forEachPair(responses.stream(), sizes.stream(),
+                (response, size) -> assertThat(response)
+                        .hasSize(size)
+                        .allSatisfy(startTxnResponse -> assertDerivableFromBatchedResponse(startTxnResponse,
+                                batchResponse)));
+
+        assertThat(flattenResponses(responses)).satisfies(
+                TransactionStarterTest::assertThatStartTransactionResponsesAreUnique);
+    }
```
jeremyk-91, line 161: ❤️ Nice!

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/AsyncTimelockServiceIntegrationTest.java
```diff
@@ -458,44 +458,49 @@ public void waitForLockRequestsAreIdempotent() {
...
-                .isEqualTo(secondResponse.startTimestampAndPartition().partition());
+        assertThatStartIdentifiedTransactionResponseTimestampsInSamePartition(firstResponse, secondResponse);
+    }
+
+    @Test
+    public void startIdentifiedAtlasDbTransactionBatchGivesUsStartTimestampsInTheSamePartition() {
+        List<StartIdentifiedAtlasDbTransactionResponse> responses =
+                namespace.timelockService().startIdentifiedAtlasDbTransactionBatch(2);
+
+        assertThatStartIdentifiedTransactionResponseTimestampsInSamePartition(responses.get(0), responses.get(1));
```
jeremyk-91, line 491: good reuse here :)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
```diff
@@ -418,24 +419,24 @@ default LockWatchManager getLockWatchManager() {
...
      *
+     * @return a batch of transactions with associated immutable timestamp locks
      * @deprecated Similar functionality will exist, but this method is likely to change in the future
-     *
-     * @return the transaction and associated immutable timestamp lock for the task
      */
     @Deprecated
     @Timed
-    TransactionAndImmutableTsLock setupRunTaskWithConditionThrowOnConflict(PreCommitCondition condition);
+    List<TransactionAndImmutableTsLock> setupRunTaskBatchWithConditionThrowOnConflict(
```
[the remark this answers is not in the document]
jkozlowski: hmmm, feels to me the interface is pretty simple, the ordering should absolutely match; so at index 1, returned transaction should be matched to the precondition at index 1 in the request list.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -162,27 +168,43 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+            return ImmutableList.of();
+        }
+
+        List<StartIdentifiedAtlasDbTransactionResponse> responses =
+                timelockService.startIdentifiedAtlasDbTransactionBatch(conditions.size());
         try {
-            LockToken immutableTsLock = transactionResponse.immutableTimestamp().getLock();
-            long immutableTs = transactionResponse.immutableTimestamp().getImmutableTimestamp();
+            long immutableTs = Collections.max(responses.stream()
+                    .map(response -> response.immutableTimestamp().getImmutableTimestamp())
```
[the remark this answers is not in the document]
Jolyon-S: I was assuming that it is possible for the batch to be made out of several `ConjureStartTransactionsResponse`s, and therefore there being a possibility that the immutable timestamp differs, but if it is not a problem then we could equally just take the first element's value.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/TransactionManager.java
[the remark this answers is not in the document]
jeremyk-91: I think I meant more in terms of the timestamps (e.g. can you assume response[0].getTimestamp() is the min?)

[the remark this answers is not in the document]
Jolyon-S: Ah, I see what you mean. I don't think we should make that assumption anywhere (although I suspect, looking through the code at `TransactionStarter`, that it is the case anyway). Perhaps a bit of additional javadoc would cover this?

PR #4714, opened 2020-04-16
[WIP] Coalesce calls to PaxosStateLogImpl#getLeastLogEntry and #getGreatestLogEntry

jkozlowski:
**Goals (and why)**:

We have had issues with those methods getting slower and slower, since timelock does not currently have a way of removing old files in the paxos state log directory.


jkozlowski:
Not convinced we need this, but rather a better approach. Manual remediation stemmed the bleeding.

PR #4715, opened 2020-04-16
Use structured locks for observability

carterkozak:
**Goals (and why)**:

PaxosStateLogImpl and PaxosLeaderElectionService use structured
locks (the synchronized keyword) for better observability.
This makes it easier to discover which thread is holding locks
while others wait from a jstack.

Structured locks allow the jit to perform optimziations that
aren't possible for unstructured locks, for example lock coarsening[1]
and lock elision[2], on top of generally better performance of
`synchronized` on modern JVMs.

1. https://shipilev.net/jvm/anatomy-quarks/1-lock-coarsening-for-loops/
2. https://shipilev.net/jvm/anatomy-quarks/19-lock-elision/

`synchronized`

No behavior change, existing testing is sufficient.

### DIFF
--- a/changelog/@unreleased/pr-4715.v2.yml
+++ b/changelog/@unreleased/pr-4715.v2.yml
@@ -1,0 +1,9 @@
+type: improvement
+improvement:
+  description: |-
+    PaxosStateLogImpl and PaxosLeaderElectionService use structured
+    locks (the synchronized keyword) for better observability.
+    This makes it easier to discover which thread is holding locks
+    while others wait from a jstack.
+  links:
+  - https://github.com/palantir/atlasdb/pull/4715
--- a/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
@@ -19,7 +19,6 @@ import java.time.Duration;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.immutables.value.Value;
 import org.slf4j.Logger;
@@ -60,7 +59,7 @@ public class PaxosLeaderElectionService implements LeaderElectionService {
     // stored here to be consistent when proposing to take over leadership
     private static final byte[] LEADERSHIP_PROPOSAL_VALUE = null;
 
-    private final ReentrantLock lock = new ReentrantLock();
+    private final Object lock = new Object();
     private final PaxosLatestRoundVerifier latestRoundVerifier;
 
     private final PaxosProposer proposer;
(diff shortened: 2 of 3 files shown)


carterkozak:
I don't have permissions to merge this, anything I should do here?

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -187,24 +186,24 @@ private LeaderPingResult pingLeader(Optional<PaxosValue> maybeGreatestLearnedVal
...
     private void proposeLeadershipAfter(Optional<PaxosValue> value) {
-        lock.lock();
-        try {
-            log.debug("Proposing leadership with value [{}]", SafeArg.of("paxosValue", value));
-            if (!isLatestRound(value)) {
-                // This means that new data has come in so we shouldn't propose leadership.
-                // We do this check in a lock to ensure concurrent callers to blockOnBecomingLeader behaves correctly.
-                return;
-            }
+        synchronized (lock) {
```
jkozlowski, line 189: why not synchronized on the method?

changelog/@unreleased/pr-4715.v2.yml
```diff
@@ -0,0 +1,9 @@
+type: improvement
+improvement:
+  description: |-
+    PaxosStateLogImpl and PaxosLeaderElectionService use structured
+    locks (the synchronized keyword) for better observability.
+    This makes it easier to discover which thread is holding locks
+    while others wait from a jstack.
```
jkozlowski, line 7: Why is this not handled correctly by the JVM? Feels like a pretty big gaping hole if the locking classes hinder observability

jkozlowski:
We are going to rewrite this codepath and we already have metrics on the runtime of this method. I don't believe if it's worth making changes here

leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
carterkozak: External consumers can lock on the object, where nothing else can acquire this internal lock

changelog/@unreleased/pr-4715.v2.yml
carterkozak: I meant that jstack can point out which thread holds a given lock, and which are waiting for it.

jkozlowski: I don’t think we should ever be using external synchronisation

jkozlowski:
Work started on rewrite, so I don’t think we need this.

changelog/@unreleased/pr-4715.v2.yml
carterkozak: We shouldn't pass null values, but we still verify using Preconditions.checkNotNull ;-)
The tricky bit is when third party code or API consumers end up synchronizing on your objects.

jkozlowski: I don't think we run checkNotNull except on external API. This should be internal API, unless I'm mistaken.

PR #4718, opened 2020-04-17
Refactor out Shutdown Runner from SnapshotTransactionManager

Jolyon-S:
**Goals (and why)**:
Factor out the `ShutdownRunner` class from `SnapshotTransactionManager` into a common project so that other classes may use it too. Also add tests and additional functionality.

* `ShutdownRunner` moved to its own class and renamed `ExceptionHandlingRunner`
* Now also swallows exceptions for suppliers
* Can also be instantiated with a throwable therefore guaranteeing it will throw when closed (useful for the future batching PR).

* No tests previously existed for this class
* New test class added, testing the various cases

* Most important is correctness - does this capture the same behaviour as before, and do what we expect it to do?
* Secondarily is style - both of the code, and whether the class is in the right place, and the javadoc has been written correctly.

* `ExceptionHandlingRunner`

* Within a week

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
@@ -24,7 +24,8 @@ import java.util.function.Supplier;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 
 /**
- * Runs runnables and suppliers that may throw exceptions, and swallows those exceptions until the runner is closed.
+ * Runs runnables and suppliers that may throw exceptions, and swallows those exceptions until the runner is closed. If
+ * any exceptions occurred, they are suppressed in the order in which they were thrown.
  * Can be used as a resource in a try block, or delegated to by another resource.
  */
 public final class ExceptionHandlingRunner implements AutoCloseable {
--- a/atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
@@ -24,8 +24,7 @@ import java.util.function.Supplier;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 
 /**
- * Runs runnables and suppliers that may throw exceptions, and swallows those exceptions until the runner is closed. If
- * any exceptions occurred, they are suppressed in the order in which they were thrown.
+ * Runs runnables and suppliers that may throw exceptions, and swallows those exceptions until the runner is closed.
  * Can be used as a resource in a try block, or delegated to by another resource.
  */
 public final class ExceptionHandlingRunner implements AutoCloseable {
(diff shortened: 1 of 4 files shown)


atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,72 @@
...
+            return Optional.of(shutdownCallback.get());
+        } catch (Throwable throwable) {
+            failures.add(throwable);
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * Calling close with no failures should be a no-op; equally, calling close multiple times will re-throw a runtime
+     * exception with the same suppressed errors (plus any additional errors suppressed since the last close call).
```
jeremyk-91, line 61: OK. Do we want to make any guarantees on the ordering? Might be good to add a test to ratify whichever version we want (in the order executed, or just that it contains the same elements - I imagine the assertion picked is based on the guarantee we decide here).

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,72 @@
...
+            failures.add(throwable);
+        }
+    }
+
+    public <T> Optional<T> supplySafely(Supplier<T> shutdownCallback) {
+        try {
+            return Optional.of(shutdownCallback.get());
+        } catch (Throwable throwable) {
+            failures.add(throwable);
+            return Optional.empty();
```
jeremyk-91, line 55: Would be nice to have a test for the return values of `supplySafely` (even though this seems obvious)

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public void runSafely(Runnable shutdownCallback) {
+        try {
+            shutdownCallback.run();
+        } catch (Throwable throwable) {
+            failures.add(throwable);
+        }
+    }
+
+    public <T> Optional<T> supplySafely(Supplier<T> shutdownCallback) {
```
jeremyk-91, line 50: I'm guessing you have a plan for this method (right now it's unused)

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+
+/**
+ * Runs runnables and suppliers that may throw exceptions, and swallows those exceptions until later. Can be used as a
```
jeremyk-91, line 27: nit: I'd say when the exceptions are no longer swallowed: `until the runner is closed.`.

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
Jolyon-S: Yes, will use it later on

atlasdb-commons/src/main/java/com/palantir/util/ExceptionHandlingRunner.java
Jolyon-S: Intended behaviour is that they are added to the suppressed list in the order of being called - please may you verify that my test does indeed confirm this behaviour? Should I add this to the javadoc?

jeremyk-91: Yeah, the test does confirm that.

This class is in `atlasdb-commons` so I'd bias towards being very explicit with the API, someone in large internal product might actually pick it up and use it outside of Atlas code.

Jolyon-S:
We opted not to go for this route in the end, as the batching code did not need this handler.

PR #4719, opened 2020-04-17
[Timelock Partitioning] Tag remote hosts for internode communication

gmaretic:
**Goals (and why)**:
Tag remote hosts for internode communication

Add the tag at the rpc client level

None

We don't tag the clients at the level of `NetworkClientFactories`. We can easily do this, but it requires some more wiring, and increases our metrics cardinality. Is it worth it>

tiny

Now?

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
@@ -28,6 +28,7 @@ import org.immutables.value.Value;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.net.HostAndPort;
+import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.config.AuxiliaryRemotingParameters;
 import com.palantir.atlasdb.config.RemotingClientConfigs;
 import com.palantir.atlasdb.http.AtlasDbHttpClients;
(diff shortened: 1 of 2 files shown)


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosRemoteClients.java
```diff
@@ -160,7 +161,11 @@
                                 .remotingClientConfig(() -> RemotingClientConfigs.DEFAULT)
                                 .shouldSupportBlockingOperations(false)
                                 .build()))
-                .map(proxy -> AtlasDbMetrics.instrumentWithTaggedMetrics(metrics().getTaggedRegistry(), clazz, proxy));
+                .map((host, proxy) -> AtlasDbMetrics.instrumentWithTaggedMetrics(
+                        metrics().getTaggedRegistry(),
+                        clazz,
+                        proxy,
+                        any -> ImmutableMap.of(AtlasDbMetricNames.TAG_REMOTE_HOST, host)));
```
jeremyk-91, line 168: Is this risky in terms of metrics cardinality? I'm in the process of checking with the internal team that's responsible for this

jeremyk-91: Checked internally: no, this is fine.

carterkozak:
Have you considered using [metric-schema](https://github.com/palantir/metric-schema) here? A large internal project would really appreciate the generated documentation!

PR #4721, opened 2020-04-20
Serializable index need not lock anything

j-baker:
It's an index table, and so semantically equivalent locks were taken out on the main table. We can avoid taking all locks, and thus reduce some network traffic.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
@@ -82,7 +82,8 @@ public enum ConflictHandler {
      * on index tables. Read/write conflicts should still need to be checked, since we do not need to read the index
      * table with the main table.
      * <p>
-     * This conflict does not lock, because it's assumed that a semantically equivalent lock is taken out on the base table.
+     * This conflict does not lock, because it's assumed that a semantically equivalent lock is taken out on the base 
+     * table.
      */
     SERIALIZABLE_INDEX(false, false, false, true),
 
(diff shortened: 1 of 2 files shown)


j-baker:
Verified internally that no one abuses this strategy.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -82,10 +82,9 @@
      * on index tables. Read/write conflicts should still need to be checked, since we do not need to read the index
      * table with the main table.
      * <p>
-     * This conflict handler locks at cell level. As index tables use dynamic columns, locking at row level would
-     * result in higher contention.
+     * This conflict does not lock, because it's assumed that a semantically equivalent lock is taken out on the base table.
```
Jolyon-S, line 85: nit: fix line length

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -82,10 +82,10 @@
      * on index tables. Read/write conflicts should still need to be checked, since we do not need to read the index
      * table with the main table.
      * <p>
-     * This conflict handler locks at cell level. As index tables use dynamic columns, locking at row level would
-     * result in higher contention.
+     * This conflict does not lock, because it's assumed that a semantically equivalent lock is taken out on
```
jeremyk-91, line 85:
```suggestion
     * This conflict does not lock, because it's assumed that a semantically equivalent lock is taken out on
```

jeremyk-91:
The changelog exists, despite the check not reporting, so I'm going to merge this.

PR #4723, opened 2020-04-21
Use refreshable for runtime config

jkozlowski:
**Goals (and why)**:

Should allow us to remove CachedComposingSupplier class (in separate PR) and simplify diff in https://github.com/palantir/atlasdb/pull/4706.

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
@@ -70,7 +70,7 @@ public class TransactionPostMortemRunner {
             TransactionManager transactionManager,
             TableReference tableReference,
             AtlasDbConfig install,
-            AtlasDbRuntimeConfig runtime,
+            Refreshable<AtlasDbRuntimeConfig> runtime,
             ClientLockDiagnosticCollector clientLockDiagnosticCollector) {
         this.timelockNamespace = timelockNamespace(install);
         this.clientLockDiagnosticCollector = clientLockDiagnosticCollector;
@@ -177,9 +177,9 @@ public class TransactionPostMortemRunner {
 
     private static LockDiagnosticInfoService createRpcClient(
             AtlasDbConfig config,
-            AtlasDbRuntimeConfig runtimeConfig) {
+            Refreshable<AtlasDbRuntimeConfig> runtimeConfigSupplier) {
         Supplier<ServerListConfig> serverListConfigSupplier =
-                getServerListConfigSupplierForTimeLock(config, runtimeConfig);
+                getServerListConfigSupplierForTimeLock(config, runtimeConfigSupplier);
 
         timelockNamespace(config);
 
@@ -187,7 +187,7 @@ public class TransactionPostMortemRunner {
                 new MetricsManager(new MetricRegistry(), new DefaultTaggedMetricRegistry(), _unused -> true),
                 serverListConfigSupplier,
                 UserAgent.of(UserAgent.Agent.of("agent", "0.0.0")),
-                runtimeConfig::remotingClient);
+                () -> runtimeConfigSupplier.get().remotingClient());
 
         return serviceCreator.createService(LockDiagnosticInfoService.class);
     }
@@ -199,13 +199,11 @@ public class TransactionPostMortemRunner {
 
     private static Supplier<ServerListConfig> getServerListConfigSupplierForTimeLock(
             AtlasDbConfig config,
-            AtlasDbRuntimeConfig runtimeConfigSupplier) {
+            Refreshable<AtlasDbRuntimeConfig> runtimeConfigSupplier) {
         TimeLockClientConfig clientConfig = config.timelock()
                 .orElseGet(() -> ImmutableTimeLockClientConfig.builder().build());
         return ServerListConfigs.parseInstallAndRuntimeConfigs(
                 clientConfig,
-                Refreshable.only(runtimeConfigSupplier.timelockRuntime()));
+                runtimeConfigSupplier.map(AtlasDbRuntimeConfig::timelockRuntime));
     }
-
-
 }
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
@@ -48,6 +48,7 @@ import com.palantir.common.persist.Persistable;
 import com.palantir.common.streams.KeyedStream;
 import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.refreshable.Refreshable;
 import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
 import com.palantir.util.OptionalResolver;
 
@@ -69,7 +70,7 @@ public class TransactionPostMortemRunner {
             TransactionManager transactionManager,
             TableReference tableReference,
             AtlasDbConfig install,
-            Supplier<AtlasDbRuntimeConfig> runtime,
+            AtlasDbRuntimeConfig runtime,
             ClientLockDiagnosticCollector clientLockDiagnosticCollector) {
         this.timelockNamespace = timelockNamespace(install);
         this.clientLockDiagnosticCollector = clientLockDiagnosticCollector;
@@ -176,9 +177,9 @@ public class TransactionPostMortemRunner {
 
     private static LockDiagnosticInfoService createRpcClient(
             AtlasDbConfig config,
-            Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier) {
+            AtlasDbRuntimeConfig runtimeConfig) {
         Supplier<ServerListConfig> serverListConfigSupplier =
-                getServerListConfigSupplierForTimeLock(config, runtimeConfigSupplier);
+                getServerListConfigSupplierForTimeLock(config, runtimeConfig);
 
         timelockNamespace(config);
 
@@ -186,7 +187,7 @@ public class TransactionPostMortemRunner {
                 new MetricsManager(new MetricRegistry(), new DefaultTaggedMetricRegistry(), _unused -> true),
                 serverListConfigSupplier,
                 UserAgent.of(UserAgent.Agent.of("agent", "0.0.0")),
-                () -> runtimeConfigSupplier.get().remotingClient());
+                runtimeConfig::remotingClient);
 
         return serviceCreator.createService(LockDiagnosticInfoService.class);
     }
@@ -198,12 +199,12 @@ public class TransactionPostMortemRunner {
 
     private static Supplier<ServerListConfig> getServerListConfigSupplierForTimeLock(
             AtlasDbConfig config,
-            Supplier<AtlasDbRuntimeConfig> runtimeConfigSupplier) {
+            AtlasDbRuntimeConfig runtimeConfigSupplier) {
         TimeLockClientConfig clientConfig = config.timelock()
                 .orElseGet(() -> ImmutableTimeLockClientConfig.builder().build());
-        return () -> ServerListConfigs.parseInstallAndRuntimeConfigs(
+        return ServerListConfigs.parseInstallAndRuntimeConfigs(
                 clientConfig,
-                () -> runtimeConfigSupplier.get().timelockRuntime());
+                Refreshable.only(runtimeConfigSupplier.timelockRuntime()));
     }
 
 
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
@@ -85,7 +85,7 @@ final class AtlasDbRuntimeConfigRefreshable implements AutoCloseable {
                         refreshable.update(config.get());
                     } catch (Throwable e) {
                         // This should not occur in practice
-                        log.error("Failed to load runtime config", e);
+                        log.error("Failed to reload runtime config", e);
                     }
                 },
                 REFRESH_INTERVAL.toNanos(),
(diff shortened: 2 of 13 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ConfigRefreshable.java
```diff
@@ -0,0 +1,91 @@
...
+    static ConfigRefreshable createPolling(Supplier<Optional<AtlasDbRuntimeConfig>> config) {
+        DefaultRefreshable<Optional<AtlasDbRuntimeConfig>> refreshable = new DefaultRefreshable<>(call(config));
+
+        ScheduledExecutorService executor = PTExecutors.newSingleThreadScheduledExecutor();
+        executor.scheduleWithFixedDelay(
+                () -> {
+                    try {
+                        refreshable.update(config.get());
+                    } catch (Throwable e) {
+                        // This should not occur in practice
```
jkozlowski, line 73: This isn't ideal, but if this occurs the application is doing something really rather bad.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -202,10 +202,13 @@
...
-    Supplier<Optional<AtlasDbRuntimeConfig>> runtimeConfigSupplier() {
-        return Optional::empty;
-    }
+    abstract Optional<Refreshable<Optional<AtlasDbRuntimeConfig>>> runtimeConfig();
+
+    /**
+     * @deprecated use {@link #runtimeConfig} instead.
+     */
+    @Deprecated
+    abstract Optional<Supplier<Optional<AtlasDbRuntimeConfig>>> runtimeConfigSupplier();
```
jkozlowski, line 211: This change should be fine, the autogenerated immutables class should still have the same #runtimeConfigSupplier(Supplier) method, so should not be a break for users.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -279,6 +282,12 @@ LockWatchingCache lockWatchingCache() {
         thread.start();
     };
 
+    @Value.Check
+    protected void check() {
+        Preconditions.checkState(!(runtimeConfigSupplier().isPresent() && runtimeConfig().isPresent()),
+                "Either Refreshable or Supplier of runtime config must be provided or none, but not both");
```
jkozlowski, line 288: This maintains the current setup, where it's possible to not provide the runtime config (used for InMemory configuration).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -375,11 +388,11 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
 
             // Even if sweep queue writes are enabled, unless targeted sweep is enabled we generally still want to
             // at least retain the option to perform background sweep, which requires updating the priority table.
-            if (!targetedSweepIsFullyEnabled()) {
+            if (!targetedSweepIsFullyEnabled(config(), runtime)) {
                 kvs = SweepStatsKeyValueService.create(kvs,
                         new TimelockTimestampServiceAdapter(lockAndTimestampServices.timelock()),
-                        Suppliers.compose(SweepConfig::writeThreshold, sweepConfig::get),
-                        Suppliers.compose(SweepConfig::writeSizeThreshold, sweepConfig::get)
+                        sweepConfig.map(SweepConfig::writeThreshold),
```
jkozlowski, line 394: I removed all uses of Suppliers.compose in this class.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: See test in TransactionManagers I added, it's using this method.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ConfigRefreshable.java
```diff
@@ -0,0 +1,92 @@
...
+    public Refreshable<Optional<AtlasDbRuntimeConfig>> refreshable() {
+        return delegate;
+    }
+
+    @Override
+    public void close() {
+        closer.run();
+    }
+
+    static ConfigRefreshable wrap(Refreshable<Optional<AtlasDbRuntimeConfig>> delegate) {
```
ferozco, line 59: Seems kinda odd to wrap the refreshable in this way. I think it was expose a way to convert a supplier to an optional using some executor, and handle the lifecycle of the executor seperately

ferozco: The suggestion is to just get rid of the extra level of abstraction and manage the lifecycle of the executor used to update the refreshables separately. This is especially beneficial since it allows you to reuse the executor across refreshables and allows you to avoid unnecessarily wrapping existing refreshables.

Hence:
```
ConfigRefreshable<T> config1 = ConfigRefreshable.createPolling(mySupplier1);
ConfigRefreshable<T> config2 = ConfigRefreshable.createPolling(mySupplier2);
...

config1.close();
config2.close();
```

becomes:
```
ScheduledExecutorService refreshableExecutor = PTExecutors.newSingleThreadScheduledExecutor();

Refreshable<T> config1 = Refreshables.create(mySupplier1, refreshableExecutor);
Refreshable<T> config2 = Refreshables.create(mySupplier2, refreshableExecutor);
...

refreshableExecutor.shutdown()
```

jkozlowski: I don’t expect to be using refreshables anywhere else. Handling this separately in TransactionManagers will be messy, because you don’t create the executor if someone provides the Refreshable. I can see what it would look like though

jkozlowski: So I tried to come up with something that's not messy and I can't: the fact that you only sometimes create the executor, and then you need to add it to the closeables list AND THEN, but only after TransactionManager is created, call #shutdownCallbacks with it makes me not want to do it.

This is actually a nice small abstraction of a closeable refreshable, and everything falls through nicely.

jkozlowski: Instead, I don't expose any special wrap methods, but instead the only way to create this special refreshable is from the TransactionManagers builder.

jkozlowski: And if we excavate this so we can remove the deprecated method, we'll just remove all this code.

ferozco: sure, whatever you're happy with

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
```diff
@@ -0,0 +1,110 @@
...
+    private static AtlasDbRuntimeConfigRefreshable createPolling(Supplier<Optional<AtlasDbRuntimeConfig>> config) {
+        DefaultRefreshable<Optional<AtlasDbRuntimeConfig>> refreshable = new DefaultRefreshable<>(call(config));
+
+        ScheduledExecutorService executor = PTExecutors.newSingleThreadScheduledExecutor();
+        executor.scheduleWithFixedDelay(
+                () -> {
+                    try {
+                        refreshable.update(config.get());
+                    } catch (Throwable e) {
+                        // This should not occur in practice
```
jkozlowski, line 87: Famous last words: the danger here is that we write some sort of weird update code (since DefaultRefreshable.update propagates to all the children) and this throws, and we won't find out about it. This is a bit of a change in behavior, but I don't feel strongly that this is such high risk.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: I guess my bigger concern is if someone calls `txMgr.runtimeConfigSupplier()` though a code search suggests that no one does so. 

If someone did (especially if that someone is one of our friendly supporting libs) that would be a spicy break 🔥but thankfully looks like that's not the case.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: nit: probably enough to just say `Cannot provide both Refreshable and Supplier of runtime config`?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -538,9 +550,11 @@ TransactionConfig withConsolidatedGrabImmutableTsLockFlag(TransactionConfig tran
...
     }
 
-    private boolean targetedSweepIsFullyEnabled() {
-        return config().targetedSweep().enableSweepQueueWrites()
-                && runtimeConfigSupplier().get().map(config -> config.targetedSweep().enabled()).orElse(false);
+    private static boolean targetedSweepIsFullyEnabled(
+            AtlasDbConfig installConfig,
+            Supplier<AtlasDbRuntimeConfig> runtime) {
+        return installConfig.targetedSweep().enableSweepQueueWrites()
+                && runtime.get().targetedSweep().enabled();
```
jeremyk-91, line 557: I think you drive-by fixed a bug here! Previously if you don't specify a runtime config, targeted sweep is enabled _in practice_, but the check here thinks it is disabled. Now it correctly recognises that targeted sweep is enabled by default.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                },
+                REFRESH_INTERVAL.toNanos(),
+                REFRESH_INTERVAL.toNanos(),
+                TimeUnit.NANOSECONDS);
+
+        return new AtlasDbRuntimeConfigRefreshable(refreshable, () -> {
+            if (!MoreExecutors.shutdownAndAwaitTermination(
+                    executor, GRACEFUL_SHUTDOWN.toMillis(), TimeUnit.MILLISECONDS)) {
+                log.error("Executor did not terminate within graceful shutdown duration");
```
jeremyk-91, line 98: nit: I'd prefer `warn`, this is probably not super actionable for us?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
jeremyk-91: Not sure I follow: do we mean a subscriber? If a subscriber throws it does look like the world could be in an inconsistent state, but right now I think we're fine

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Hmmm, I don't think runtimeConfigSupplier() would've compiled before? immutables shouldn't generate it, it only generates them when the type is Optional of something.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbRuntimeConfigRefreshable.java
jkozlowski: Yes, I mean that Refreshables remember everyone who called #map on them and propagate the updates. Yes you could be in an inconsistent state, but I think this is by design of refreshables.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Yep

atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
```diff
@@ -69,7 +70,7 @@ public TransactionPostMortemRunner(
             TransactionManager transactionManager,
             TableReference tableReference,
             AtlasDbConfig install,
-            Supplier<AtlasDbRuntimeConfig> runtime,
```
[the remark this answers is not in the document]
jkozlowski: This is only instantiated in test code from what I can see? So it didn't make sense to keep it this way, and it was calling ServerListConfigs.parseInstallAndRuntimeConfigs, which now takes a Refreshable.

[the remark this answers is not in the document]
jkozlowski: Ah, nevermind, this is called in an internal product at runtime. Ok, lemme rejig this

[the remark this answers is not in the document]
jkozlowski: I'll change into Refreshable and we'll need to fixup internal product to do this properly.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Done

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ConfigRefreshable.java
jkozlowski: Or we are really.

atlasdb-config/src/main/java/com/palantir/atlasdb/debug/TransactionPostMortemRunner.java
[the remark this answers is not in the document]
jkozlowski: Donezo

PR #4724, opened 2020-04-21
[PaxosLog] Sqlite implementation without namespaces

gmaretic:
**Goals (and why)**:
This is the basic psqlite implementation, without namespacing which will be added in a subsequent PR.

Relatively straightforward, although there is some magic with how the result sets are interpreted when there are no results

Unit tests using in memory implementation

I think it's fine

Tests?

Now

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
@@ -21,13 +21,11 @@ import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.function.Supplier;
 
-import com.google.common.base.Suppliers;
-
 /**
  * This class is responsible for creating Sqlite connections to an instance.
  * There should be one instance per timelock.
  */
-public class SqliteConnections {
+public final class SqliteConnections {
     private SqliteConnections() {
         // no
     }
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
@@ -27,17 +27,16 @@ import com.google.common.base.Suppliers;
  * This class is responsible for creating Sqlite connections to an instance.
  * There should be one instance per timelock.
  */
-public class Sqlites {
-    private Sqlites() {
+public class SqliteConnections {
+    private SqliteConnections() {
         // no
     }
 
-    public static Supplier<Connection> createDatabaseForTests() {
-        Supplier<Connection> memoized = Suppliers.memoize(createSqliteDatabase(":memory:")::get);
-        return memoized;
+    public static Supplier<Connection> createDatabaseForTest() {
+        return createSqliteDatabase(":memory:");
     }
 
-    public static Supplier<Connection> createSqliteDatabase(String path) {
+    private static Supplier<Connection> createSqliteDatabase(String path) {
         String target = String.format("jdbc:sqlite:%s", path);
         return () -> {
             try {
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
@@ -16,12 +16,15 @@
 
 package com.palantir.paxos;
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.function.Supplier;
 
+import com.google.common.io.ByteStreams;
 import com.palantir.common.persist.Persistable;
 import com.palantir.exception.PalantirSqlException;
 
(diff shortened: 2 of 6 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -0,0 +1,49 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.function.Supplier;
+
+import com.google.common.base.Suppliers;
+
+/**
+ * This class is responsible for creating Sqlite connections to an instance.
+ * There should be one instance per timelock.
```
jeremyk-91, line 28: Not now, and we'll change this: should we *enforce* this?

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+import com.google.common.base.Suppliers;
+
+public class SqlitePaxosStateLogTest {
+    private PaxosStateLog<PaxosValue> stateLog;
+
+    @Before
+    public void setup() {
+        Supplier<Connection> connectionSupplier = Suppliers.memoize(SqliteConnections.createDatabaseForTest()::get);
+        stateLog =  SqlitePaxosStateLog.createInitialized(connectionSupplier);
```
jeremyk-91, line 37:
```suggestion
        stateLog = SqlitePaxosStateLog.createInitialized(connectionSupplier);
```

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public static <V extends Persistable & Versionable> PaxosStateLog<V> createInitialized(Supplier<Connection> conn) {
+        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(conn);
+        log.initialize();
+        return log;
+    }
+
+    private void initialize() {
+        executeVoid("CREATE TABLE IF NOT EXISTS paxosLog (seq BIGINT, val BLOB, CONSTRAINT pk_dual PRIMARY KEY (seq))");
```
jeremyk-91, line 45: will be fixed in next PR: the constraint is funky

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+            preparedStatement.setBytes(2, round.persistToBytes());
+            preparedStatement.execute();
+        } catch (SQLException e) {
+            throw Throwables.rewrapAndThrowUncheckedException(e);
+        }
+    }
+
+    @Override
+    public byte[] readRound(long seq) {
+        return executeStatement(String.format("SELECT val FROM paxosLog WHERE seq = %s;", seq))
```
jeremyk-91, line 63:
```suggestion
        return executeStatement(String.format("SELECT val FROM paxosLog WHERE seq = %s", seq))
```

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+            preparedStatement.setBytes(2, round.persistToBytes());
+            preparedStatement.execute();
+        } catch (SQLException e) {
+            throw Throwables.rewrapAndThrowUncheckedException(e);
+        }
+    }
+
+    @Override
+    public byte[] readRound(long seq) {
+        return executeStatement(String.format("SELECT val FROM paxosLog WHERE seq = %s", seq))
```
j-baker, line 63: please can we use jooq here? Direct Java sql is kind of shitty...

j-baker: or JDBI - JDBI's pretty great.

j-baker: Explicitly, something like this:

```java
public interface UserDao {
    @SqlUpdate("CREATE TABLE user (id INTEGER PRIMARY KEY, name VARCHAR)")
    void createTable();

    @SqlUpdate("INSERT INTO user(id, name) VALUES (?, ?)")
    void insertPositional(int id, String name);

    @SqlUpdate("INSERT INTO user(id, name) VALUES (:id, :name)")
    void insertNamed(@Bind("id") int id, @Bind("name") String name);

    @SqlUpdate("INSERT INTO user(id, name) VALUES (:id, :name)")
    void insertBean(@BindBean User user);

    @SqlQuery("SELECT * FROM user ORDER BY name")
    @RegisterBeanMapper(User.class)
    List<User> listUsers();
}
```

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private void executeVoid(String statement) {
+        try {
+            connectionSupplier.get().prepareStatement(statement).execute();
+        } catch (SQLException e) {
+            throw Throwables.rewrapAndThrowUncheckedException(e);
+        }
+    }
+
+    private Optional<ResultSet> executeStatement(String statement) {
```
j-baker, line 95: Connection is closeable :)

j-baker: JDBI handles that for you :)

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public byte[] readRound(long seq) {
+        return executeStatement(String.format("SELECT val FROM paxosLog WHERE seq = %s", seq))
+                .map(SqlitePaxosStateLog::getByteArrayUnchecked)
+                .orElse(null);
+    }
+
+    @Override
+    public long getLeastLogEntry() {
+        return executeStatement("SELECT MIN(seq) FROM paxosLog")
```
jkozlowski, line 70: Do we not need an index?

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.function.Supplier;
+
+import com.google.common.io.ByteStreams;
+import com.palantir.common.base.Throwables;
+import com.palantir.common.persist.Persistable;
+
+public final class SqlitePaxosStateLog<V extends Persistable & Versionable> implements PaxosStateLog<V> {
+    private final Supplier<Connection> connectionSupplier;
+
+    private SqlitePaxosStateLog(Supplier<Connection> connectionSupplier) {
```
jkozlowski, line 34: Who closes the connection?

PR #4725, opened 2020-04-21
[PaxosLog] MigrationPaxosStateLog

gmaretic:
**Goals (and why)**:
Have a dual writing paxos state log

At initialization time, seed the new log with the state from the old log. Do all the writes to both logs, do all the reads from both, log if results are not consistent, then return the value from old one.

Unit tests

Currently, we reseed the new log every time. Maybe we should do something smarter? The simple idea with using get least, get greatest to check the current state does not guarantee consistency unfortunately.

Tests?

tomorrow

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/MigrationPaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/MigrationPaxosStateLog.java
@@ -18,11 +18,11 @@ package com.palantir.paxos;
 
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.Optional;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
 
+import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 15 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/MigrationPaxosStateLog.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    @Override
+    public byte[] readRound(long seq) throws IOException {
+        lock.readLock().lock();
+        try {
+            byte[] legacyResult = legacyLog.readRound(seq);
+            byte[] currentResult = currentLog.readRound(seq);
+            if (!Arrays.equals(legacyResult, currentResult)) {
```
jkozlowski, line 89: I think this validation should run only on startup

leader-election-impl/src/main/java/com/palantir/paxos/MigrationPaxosStateLog.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Persistable.Hydrator<V> hydrator;
+
+    private MigrationPaxosStateLog(Settings<V> settings) {
+        this.legacyLog = settings.sourceOfTruth();
+        this.currentLog = settings.secondary();
+        this.hydrator = settings.hydrator();
+    }
+
+    public static <V extends Persistable & Versionable> PaxosStateLog<V> create(Settings<V> settings) {
+        seedCurrentFromLegacy(settings);
```
jkozlowski, line 49: How will you handle remembering to run this only once?

Also, how would we make sure someone doesn’t downgrade from when we start writing only to new store to code that only reads from old store

jeremyk-91: I think the current approach _is_ intentionally to do it everytime. This is probably mitigated by us being non-leader when starting up, generally speaking, so not terrible. Though it could be O(minutes) at our big deployments which seems less than ideal. 

We should think if there's a way to only do this for values that weren't already copied (possibly bounded by something in the learner sequence). If I'm not mistaken overwrites _can_ happen at previous sequence numbers, so we do need to be careful.

Downgrade is probably protected via schema versions in internal framework.

jeremyk-91:
There are probably two separate concerns here: _validation of correctness_, and _data migration_. This is probably a fair approach for the first, though we need to think carefully about the second.

gmaretic:
Will reopen as a new PR

PR #4726, opened 2020-04-21
[PaxosLog] Part 3: Namespacing SqlitePaxosStateLog

jeremyk-91:
**Goals (and why)**:
- Allow multiple distinct sequences that are stored in the same SQLite instance.

- Add namespacing for our SQLite store: allow a string namespace to be taken in
- Provide a mapping from PaxosUseCases, clients and roles (acceptor or learner) to distinct string table identifiers for SQLite.
- Partially wire this mapping through to generate table names that would be appropriate for SQLite usage. _These would be wired into log creation in PaxosAcceptor or Learner._

Not much. Existing tests passing validate a lack of regressions here. Some extra tests around namespaces being distinct but also referenceable.

- `LocalPaxosComponents#createComponents()` is currently fragile - the mechanism we use in CentOS Migration for detecting if a client is original is whether it has a paxos data directory. We need to be careful with this.
- Is the wiring of clients correct, in that a `(paxosUseCase, client, paxosRole)` combination maps to a `paxosUseCase!client!paxosRole` table in SQLite? Could there ever be clashes?
- This creates a fair number of tables (basically with N clients 2N tables), though it keeps each table fairly small. There are alternatives: is this a reasonable design choice? Note that we don't plan on doing joins on these tables. I don't think it's that bad since we're still looking at <300 tables in general even at large deployments.

**Where should we start reviewing?**: PaxosStorageParameters


leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -30,26 +30,34 @@
...
+        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(conn, logNamespace);
         log.initialize();
         return log;
     }
 
     private void initialize() {
-        executeVoid("CREATE TABLE IF NOT EXISTS paxosLog (seq BIGINT, val BLOB, CONSTRAINT pk_dual PRIMARY KEY (seq))");
+        executeVoid(
+                String.format(
+                        "CREATE TABLE IF NOT EXISTS %s (seq BIGINT, val BLOB, CONSTRAINT pk_%s PRIMARY KEY (seq))",
```
jkozlowski, line 51: Table per namespace, interesting.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -30,26 +30,34 @@
...
-    private SqlitePaxosStateLog(Supplier<Connection> connectionSupplier) {
+    private SqlitePaxosStateLog(Supplier<Connection> connectionSupplier, String logNamespace) {
         this.connectionSupplier = connectionSupplier;
+        this.logNamespace = logNamespace;
     }
 
-    public static <V extends Persistable & Versionable> PaxosStateLog<V> createInitialized(Supplier<Connection> conn) {
-        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(conn);
+    public static <V extends Persistable & Versionable> PaxosStateLog<V> createInitialized(
+            Supplier<Connection> conn,
```
jkozlowski, line 41: Can you create as many sqlite instances as you want if it's backed by the filesystem? Does it maintain some on-disk locks or something?

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -30,26 +30,34 @@
 
 public final class SqlitePaxosStateLog<V extends Persistable & Versionable> implements PaxosStateLog<V> {
     private final Supplier<Connection> connectionSupplier;
+    private final String logNamespace;
 
-    private SqlitePaxosStateLog(Supplier<Connection> connectionSupplier) {
+    private SqlitePaxosStateLog(Supplier<Connection> connectionSupplier, String logNamespace) {
```
jkozlowski, line 35: Do we need to enforce something about the logNamespace, I don't suspect sqlite likes any odd table names?

PR #4729, opened 2020-04-22
[PaxosLog] Use JDBI and add namespacing

gmaretic:
**Goals (and why)**:
Address comments on previous PR and clean up the implementation

Straightforward, also added namespacing to avoid having to redo work

Existing tests pass

To use in memory sqlite we have to make sure our connection does not get closed. This is done by creating the Jdbi object in tests and passing it in as an argument to the log. This factory is exposed to allow testing multiple namespaces, but the factory that should generally be called is the one with the connection supplier, as that will create and close connections appropriately. 

anywhere

asap

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
@@ -42,11 +42,8 @@ public final class SqlitePaxosStateLog<V extends Persistable & Versionable> impl
 
     public static <V extends Persistable & Versionable> PaxosStateLog<V> create(String namespace,
             Supplier<Connection> connectionSupplier) {
-        return create(namespace, Jdbi.create(connectionSupplier::get));
-    }
-
-    static <V extends Persistable & Versionable> PaxosStateLog<V> create(String namespace, Jdbi jdbi) {
-        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(namespace, jdbi.installPlugin(new SqlObjectPlugin()));
+        Jdbi jdbi = Jdbi.create(connectionSupplier::get).installPlugin(new SqlObjectPlugin());
+        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(namespace, jdbi);
         log.initialize();
         return log;
     }
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
@@ -45,7 +45,7 @@ public final class SqlitePaxosStateLog<V extends Persistable & Versionable> impl
         return create(namespace, Jdbi.create(connectionSupplier::get));
     }
 
-    public static <V extends Persistable & Versionable> PaxosStateLog<V> create(String namespace, Jdbi jdbi) {
+    static <V extends Persistable & Versionable> PaxosStateLog<V> create(String namespace, Jdbi jdbi) {
         SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(namespace, jdbi.installPlugin(new SqlObjectPlugin()));
         log.initialize();
         return log;
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
@@ -16,105 +16,92 @@
 
 package com.palantir.paxos;
 
-import java.io.IOException;
 import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.util.Optional;
+import java.util.OptionalLong;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
-import com.google.common.io.ByteStreams;
-import com.palantir.common.base.Throwables;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.sqlobject.SingleValue;
+import org.jdbi.v3.sqlobject.SqlObjectPlugin;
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogTest.java
```diff
@@ -102,4 +127,34 @@ private static PaxosValue valueForRound(long round) {
...
+
+    private static Supplier<Connection> bypassCloseOnConnection(Supplier<Connection> connectionSupplier) {
+        return Suppliers.compose(conn ->
+                        (Connection) Proxy.newProxyInstance(Connection.class.getClassLoader(),
+                                new Class<?>[] {Connection.class},
+                                new CloseIgnoringInvocationHandler(conn)),
+                connectionSupplier::get);
+    }
+
+    private static final class CloseIgnoringInvocationHandler extends AbstractInvocationHandler {
```
jeremyk-91, line 145: We talked offline so I know why this is needed, but probably best to document that.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -16,105 +16,89 @@
...
-                .orElse(PaxosAcceptor.NO_LOG_ENTRY);
+        return execute(x -> x.getLeastLogEntry(namespace)).orElse(PaxosAcceptor.NO_LOG_ENTRY);
     }
 
     @Override
     public long getGreatestLogEntry() {
-        return executeStatement("SELECT MAX(seq) FROM paxosLog")
-                .map(SqlitePaxosStateLog::getLongResultUnchecked)
-                .orElse(PaxosAcceptor.NO_LOG_ENTRY);
+        return execute(x -> x.getGreatestLogEntry(namespace)).orElse(PaxosAcceptor.NO_LOG_ENTRY);
```
jeremyk-91, line 72: nit: standardise `x` vs `dao`

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -16,105 +16,89 @@
...
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.Define;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
 import com.palantir.common.persist.Persistable;
 
 public final class SqlitePaxosStateLog<V extends Persistable & Versionable> implements PaxosStateLog<V> {
-    private final Supplier<Connection> connectionSupplier;
+    private final String namespace;
```
jkozlowski, line 35: Is there a validation/parsing of the namespace somewhere in timelock? I wonder what restrictions sqllite places on table names.

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogTest.java
```diff
@@ -102,4 +127,39 @@ private static PaxosValue valueForRound(long round) {
...
+    private static Supplier<Connection> bypassCloseOnConnection(Supplier<Connection> connectionSupplier) {
+        return Suppliers.compose(conn ->
+                        (Connection) Proxy.newProxyInstance(Connection.class.getClassLoader(),
+                                new Class<?>[] {Connection.class},
+                                new CloseIgnoringInvocationHandler(conn)),
+                connectionSupplier::get);
+    }
+
+    /**
+     * JDBI closes the connection after executing a query. This is desired behaviour, but does not play nicely with in
```
jkozlowski, line 146: Is there no way to amend this behavior? How about SQLiteDataSource, does that do it better? And then also SQLiteConnectionPoolDataSource,

jkozlowski: Or how about we use the disk in tests too? Shouldn't really be a problem I suspect? We can give it temp directories

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -16,105 +16,89 @@
...
+    private SqlitePaxosStateLog(String namespace, Jdbi jdbi) {
+        this.namespace = namespace;
+        this.jdbi = jdbi;
     }
 
-    public static <V extends Persistable & Versionable> PaxosStateLog<V> createInitialized(Supplier<Connection> conn) {
-        SqlitePaxosStateLog<V> log = new SqlitePaxosStateLog<>(conn);
+    public static <V extends Persistable & Versionable> PaxosStateLog<V> create(String namespace,
+            Supplier<Connection> connectionSupplier) {
+        Jdbi jdbi = Jdbi.create(connectionSupplier::get).installPlugin(new SqlObjectPlugin());
```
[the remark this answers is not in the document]
gmaretic: No, this does not take up any resources that need to be cleaned up later

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogTest.java
gmaretic: OK, we use disk

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
gmaretic: Yeah, we have to pass in a valid namespace which we take care of elsewhere

jkozlowski: What about sqlite restrictions? are all namespaces valid for sqlite tables?

jkozlowski: Can we have a Namespace wrapper type? so that the validation done is clear here?

PR #4732, opened 2020-04-24
[PDS-117161] Add additional logging on close of Snapshot Transaction Manager

Jolyon-S:
**Goals (and why)**:
As per PDS-117161, there are times where a transaction manager is closed but we do not necessarily call all of the closing callbacks. This PR should help with debugging this case.

* Add logging for when the Snapshot transaction manager closes;
* Add logging for when this close has been called before (i.e. transaction manager is already closed).

N/A

Any other places where logging should be added? I know that internal skiing product doesn't (or should not) close transaction managers too frequently, but if there are other uses that do, this could add more logging noise.

`SnapshotTransactionManager`

Today would be nice.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
@@ -345,7 +345,9 @@ import com.palantir.timestamp.TimestampService;
             }
 
             shutdownRunner.shutdownSafely(metricsManager::deregisterMetrics);
+            log.info("Close callbacks complete in snapshot transaction manager");
         }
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -326,7 +326,9 @@ public void registerClosingCallback(Runnable closingCallback) {
      */
     @Override
     public void close() {
+        log.info("Calling close on snapshot transaction manager");
         if (!isClosed.compareAndSet(false, true)) {
+            log.info("Snapshot transaction manager has already been closed, performing no action");
             return;
         }
```
jkozlowski, line 333: Maybe logging also towards the end of the method before we exit?

PR #4735, opened 2020-04-27
[PaxosLog] VerifyingPaxosStateLog

gmaretic:
**Goals (and why)**:
Have a dual writing / read verifying implementation of paxos state log

This class assumes the inital states of both the current and the experimental log are consistent. We then simply do all the writes to both logs, do all the reads from both, log if results are not consistent, and return the result from the current log.

Unit tests

Pretty straightforward

small

asap?


leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -0,0 +1,130 @@
...
+        lock.readLock().lock();
+        try {
+            byte[] result = currentLog.readRound(seq);
+            byte[] experimentalResult = experimentalLog.readRound(seq);
+            if (!Arrays.equals(result, experimentalResult)) {
+                log.error("Mismatch in reading round for sequence number {} between legacy and current "
+                        + "implementations. Legacy result {}, current result {}.",
+                        SafeArg.of("sequence", seq),
+                        UnsafeArg.of("legacy", hydrator.hydrateFromBytes(result)),
+                        UnsafeArg.of("current", hydrator.hydrateFromBytes(experimentalResult)));
```
jeremyk-91, line 78: discussed: This needs to be resilient to `null` results.

leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -0,0 +1,130 @@
...
+        this.hydrator = settings.hydrator();
+    }
+
+    @Override
+    public void writeRound(long seq, V round) {
+        lock.writeLock().lock();
+        try {
+            currentLog.writeRound(seq, round);
+            experimentalLog.writeRound(seq, round);
+        } finally {
```
jeremyk-91, line 62: This is a bit nasty, but I'd prefer

```
lock();
try {
  current.write(seq, round);
  try {
    experiment.write(seq, round);
  } catch (RuntimeException ex) {
    log.warn("I wrote to current but not experimental, please be careful with future errors for sequence number {}", SafeArg.of(seq), ex);
     // intentionally succeed, because we don't want to degrade service
  }
} finally {
  unlock();
}
```

leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -0,0 +1,130 @@
...
+            lock.writeLock().unlock();
+        }
+    }
+
+    @Override
+    public byte[] readRound(long seq) throws IOException {
+        lock.readLock().lock();
+        try {
+            byte[] result = currentLog.readRound(seq);
+            byte[] experimentalResult = experimentalLog.readRound(seq);
```
jeremyk-91, line 72: similar to above: should we try to maintain service if the experimental log throws stuff?

leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -0,0 +1,130 @@
...
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    private long getExtremeLogEntry(Function<PaxosStateLog<V>, Long> extractor) {
+        lock.readLock().lock();
+        try {
+            long result = extractor.apply(currentLog);
+            long experimentalResult = extractor.apply(experimentalLog);
```
jeremyk-91, line 111: and here for experimental protection

leader-election-impl/src/test/java/com/palantir/paxos/VerifyingPaxosStateLogTest.java
```diff
@@ -0,0 +1,111 @@
...
+    public void noMoreInteractions() {
+        verifyNoMoreInteractions(current, experimental);
+    }
+
+    @Test
+    public void writeWritesToBoth() {
+        long sequence = 15L;
+        log.writeRound(sequence, valueForRound(sequence));
+        verify(current).writeRound(sequence, valueForRound(sequence));
+        verify(experimental).writeRound(sequence, valueForRound(sequence));
```
jeremyk-91, line 70: Non-blocking: Consider `InOrder`?

jeremyk-91: It looks something like
```
Mockito.inOrder(current, experimental)
  .verify(current).write(bla);
  .verify(experimental).write(bla);
```

leader-election-impl/src/test/java/com/palantir/paxos/VerifyingPaxosStateLogTest.java
```diff
@@ -0,0 +1,111 @@
...
+        verify(experimental).getGreatestLogEntry();
+    }
+
+    @Test
+    public void truncateBoth() {
+        log.truncate(7L);
+        verify(current).truncate(7L);
+        verify(experimental).truncate(7L);
+    }
+
```
jeremyk-91, line 101: Might want to add a few tests for the exception protection for experimental, once we've implemented that!

PR #4737, opened 2020-04-27
[PaxosLog] Expose batch write method

gmaretic:
**Goals (and why)**:
Batched writes are going to be useful for migration and benchmarks

There is an efficient implementation for SqlitePaxosStateLog.

There is a unit test

Is this polluting the interface? Alternatively we could have a different interface that exposes batched methods, but feels overengineered

tiny

asap

### DIFF
--- a/leader-election-api/src/main/java/com/palantir/paxos/PaxosStateLog.java
+++ b/leader-election-api/src/main/java/com/palantir/paxos/PaxosStateLog.java
@@ -16,8 +16,6 @@
 package com.palantir.paxos;
 
 import java.io.IOException;
-import java.util.Collection;
-import java.util.List;
 
 import com.palantir.common.persist.Persistable;
 
@@ -35,7 +33,7 @@ public interface PaxosStateLog<V extends Persistable & Versionable> {
      */
     void writeRound(long seq, V round);
 
-    default void writeBatchOfRounds(Collection<PaxosRound<V>> rounds) {
+    default void writeBatchOfRounds(Iterable<PaxosRound<V>> rounds) {
         rounds.forEach(round -> writeRound(round.sequence(), round.value()));
     }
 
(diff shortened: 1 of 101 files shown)


leader-election-api/src/main/java/com/palantir/paxos/PaxosStateLog.java
```diff
@@ -33,6 +33,10 @@
      */
     void writeRound(long seq, V round);
 
+    default void writeBatchOfRounds(Iterable<PaxosRound<V>> rounds) {
```
jeremyk-91, line 36:
```suggestion
    /**
     * Implementations of this method MUST obey the following:
     * 1) The end state of the log should be equivalent to applying the operations in 
     *    iteration order.
     * 2) Implementations need not be atomic. However, if they fail, they should apply only
     *    prefixes of the input iterable.
     */
    default void writeBatchOfRounds(Iterable<PaxosRound<V>> rounds) {
```

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -100,5 +109,9 @@ public void truncate(long toDeleteInclusive) {
 
         @SqlUpdate("DELETE FROM <table> WHERE seq <= :seq")
         boolean truncate(@Define("table") String table, @Bind("seq") long seq);
+
+        @SqlBatch("INSERT OR REPLACE INTO <table> (seq, val) VALUES (:round.sequence, :round.valueBytes)")
+        <V extends Persistable & Versionable> boolean[] writeBatchOfRounds(@Define("table") String table,
```
jeremyk-91, line 114: Discussed why this must be `boolean[]` offline: required because of the signature of `execute()`

PR #4738, opened 2020-04-27
[PDS-117310] Add trace logging when updating the namespace mapping

jeremyk-91:
**Goals (and why)**:
- See PDS-117310: the namespace table is queried a lot, and we don't really understand why.

- Add trace logging to `KvTableMappingService`.

- Nothing in particular: this change should not affect behaviour.

- Is there a potential perf problem? Given this we're only doing this when TRACE logging is enabled, I think this is fine as long as it _only_ affects perf when TRACE logging _is_ enabled.

**Where should we start reviewing?**: KvTableMappingService


jeremyk-91:
Yep, I checked that this is the only place where we write to the namespace table, so we should be covered there.

PR #4739, opened 2020-04-28
Revert revert stream corruption

sudiksha27:
**Goals (and why)**:
This PR reimplements the fix made in https://github.com/palantir/atlasdb/pull/4201 which was reverted in https://github.com/palantir/atlasdb/pull/4434. 

- Read the index table via both the method used in large internal method and the method used in #4201. 

- Proceed with deleting the unreferenced streams when the response from two methods matches.

- Do not delete unreferenced streams and log the discrepancies are observed otherwise.

Existing ETE tests should suffice.

Is there a case where we delete things that should not be deleted?

StreamStoreRenderer.java

End of this week.

### DIFF
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TargetedSweepEteTest.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TargetedSweepEteTest.java
@@ -23,7 +23,6 @@ import java.util.concurrent.TimeUnit;
 import org.awaitility.Awaitility;
 import org.junit.After;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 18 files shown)


atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TargetedSweepEteTest.java
```diff
@@ -106,6 +105,7 @@ public void targetedSweepCleansUpUnmarkedStreamsTest() {
     }
 
     @Test
+    @Ignore // TODO (jkong): This is obviously not the desired behaviour, but we are doing this for safety.
```
jeremyk-91, line 108: No need to ignore, just kill this test - it documents the broken behaviour at the time.

PR #4740, opened 2020-04-28
[PaxosLog] Part 6: Namespace Tracking for SqlitePaxosStateLog

jeremyk-91:
**Goals (and why)**:
- Keep track of all AtlasDB namespaces via a table in SQLite. Possibly contentious: please read the Concerns section.

- Create the SqlitePaxosStateLog through a Factory that tracks namespaces.

- Added some small tests.

- An alternative is to parse the namespaces from string table names. I'm not entirely happy about that. Is that a better choice?
- Do we need delete/decommission operations?
- Is there any other tracking we should include?

**Where should we start reviewing?**: SqlitePaxosStateLogFactory

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
@@ -20,6 +20,7 @@ import java.sql.Connection;
 import java.util.List;
 import java.util.function.Supplier;
 
+import org.immutables.value.Value;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 import org.jdbi.v3.sqlobject.customizer.Bind;
@@ -46,13 +47,19 @@ public class SqlitePaxosStateLogFactory<V extends Persistable & Versionable> {
         return new SqlitePaxosStateLogFactory<>(connectionSupplier, jdbi);
     }
 
-    public PaxosStateLog<V> createPaxosStateLog(String namespace) {
-        jdbi.withExtension(Queries.class, queries -> queries.registerNamespace(namespace));
-        return SqlitePaxosStateLog.create(namespace, connectionSupplier);
+    public PaxosStateLog<V> createPaxosStateLog(PaxosStateLogCreationContext context) {
+        jdbi.withExtension(Queries.class, queries -> queries.registerNamespace(context.atlasNamespace()));
+        return SqlitePaxosStateLog.create(context.physicalTableName(), connectionSupplier);
     }
 
     public List<String> getAllRegisteredNamespaces() {
         return jdbi.withExtension(Queries.class, Queries::getAllRegisteredNamespaces);
+    }
+
+    @Value.Immutable
+    public interface PaxosStateLogCreationContext {
+        String atlasNamespace();
+        String physicalTableName();
     }
 
     public interface Queries {
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
@@ -62,7 +62,7 @@ public class SqlitePaxosStateLogFactory<V extends Persistable & Versionable> {
         String physicalTableName();
     }
 
-    public interface Queries {
+    private interface Queries {
         @SqlUpdate("CREATE TABLE IF NOT EXISTS " + NAMESPACE_TABLE + " (namespace TEXT PRIMARY KEY)")
         boolean createNamespaceTable();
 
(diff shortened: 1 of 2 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
```diff
@@ -0,0 +1,75 @@
...
+        return SqlitePaxosStateLog.create(context.physicalTableName(), connectionSupplier);
+    }
+
+    public List<String> getAllRegisteredNamespaces() {
+        return jdbi.withExtension(Queries.class, Queries::getAllRegisteredNamespaces);
+    }
+
+    @Value.Immutable
+    public interface PaxosStateLogCreationContext {
+        String atlasNamespace();
```
jkozlowski, line 61: I think I asked this on previous PR: can we have a Namespace type that enforces whatever naming conventions we have, so it's easy to check whether we're somehow encroaching on these with these internal only tables?

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
```diff
@@ -0,0 +1,75 @@
...
+        String physicalTableName();
```
jkozlowski, line 62: How is this going to be generated?

j-baker: please can we just have the physical table name be a column? As in, one table for all namespaces

j-baker: and then the other question is something like: does

`SELECT DISTINCT(namespace) AS namespace FROM log` give us what we need?

jeremyk-91: The generation was meant to happen inside `timelock-impl` or `TimeLockAgent`, we'd write a function `(PaxosUseCase, Client, PaxosFunction) -> String` (where `PaxosFunction` is acceptor or learner).

@j-baker We benchmarked that approach against separate tables, and we found that the separate tables gave us (marginal but reproducibly) better write perf as well as (~30%) less disk usage, which is why we went for it. 

That said it looks like there are complexity costs here (SELECT DISTINCT needs benchmarking, and should hopefully be fast since there would be a primary key index, plus it is what we want) though.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
jeremyk-91: Can do. This one is controlled by the client regex, but no harm being defensive here.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
gmaretic: The suggested solution is cleaner, but I prefer this approach due to the perf benefits

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    public static <V extends Persistable & Versionable> SqlitePaxosStateLogFactory<V> create(
+            Supplier<Connection> connectionSupplier) {
+        Jdbi jdbi = Jdbi.create(connectionSupplier::get).installPlugin(new SqlObjectPlugin());
+        jdbi.withExtension(Queries.class, Queries::createNamespaceTable);
+        return new SqlitePaxosStateLogFactory<>(connectionSupplier, jdbi);
+    }
+
+    public PaxosStateLog<V> createPaxosStateLog(PaxosStateLogCreationContext context) {
+        jdbi.withExtension(Queries.class, queries -> queries.registerNamespace(context.atlasNamespace()));
+        return SqlitePaxosStateLog.create(context.physicalTableName(), connectionSupplier);
```
gmaretic, line 52: We should at least make `SqlitePaxosStateLog.create` package private, as if you don't go through the factory you will fail to register the namespace

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private interface Queries {
+        @SqlUpdate("CREATE TABLE IF NOT EXISTS " + NAMESPACE_TABLE + " (namespace TEXT PRIMARY KEY)")
+        boolean createNamespaceTable();
+
+        @SqlUpdate("INSERT OR REPLACE INTO " + NAMESPACE_TABLE + " VALUES (:namespace)")
+        boolean registerNamespace(@Bind("namespace") String namespace);
+
+        @SqlQuery("SELECT namespace FROM " + NAMESPACE_TABLE)
+        List<String> getAllRegisteredNamespaces();
```
gmaretic, line 73: nit: `Set` since order is not guaranteed?

jeremyk-91:
closing out in favour of #4747

PR #4741, opened 2020-04-28
Standardize on NamedThreadFactory for observability

carterkozak:
Instead of using a mix of both ThreadFactoryBuilder and
NamedThreadFactory, we standardize on the latter for the provided
instrumentation.

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -33,11 +33,11 @@ import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.lmax.disruptor.EventHandler;
 import com.lmax.disruptor.RingBuffer;
 import com.lmax.disruptor.TimeoutException;
 import com.lmax.disruptor.dsl.Disruptor;
+import com.palantir.common.concurrent.NamedThreadFactory;
 import com.palantir.logsafe.Preconditions;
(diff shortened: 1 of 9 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -61,11 +61,7 @@ private static ThreadFactory threadFactory(String safeLoggablePurpose) {
     }
 
     private static ThreadFactory createThreadFactory(String safeLoggablePurpose) {
-        String namePrefix = String.format("autobatcher.%s-", safeLoggablePurpose);
-        return new ThreadFactoryBuilder()
-                .setDaemon(true)
-                .setNameFormat(namePrefix + "%d")
-                .build();
+        return new NamedThreadFactory("autobatcher." + safeLoggablePurpose, true);
```
jkozlowski, line 64: Flag that for internal ski product, this will basically report weird metrics? We start many TransactionManagers, so if you hoped this would give you a total count and whatnot, don't think this will work.

jkozlowski: That said I don't think it's necessarily worse than no-instrumentation, just don't know if this will get us what we want. Also, this will be a few extra time series.

jkozlowski: Actually, had a look at the metrics, and looks like this will be cumulative, so all good.

jkozlowski:
+2 @jeremyk-91 to review

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
carterkozak: The underlying metric objects are shared so duplicate names are okay, if the safeLoggablePurpose differs that gives me even more specificity. Either way it solves my problem :-)

PR #4742, opened 2020-04-28
[PaxosLog] Part 7: Configuration Sketch :fire: CONTENTIOUS :fire:

jeremyk-91:
**Goals (and why)**:
- Outline how users can enable SQLite.

```
paxos:
  # (no more data directory - if present and conflicting with persistence config, explode)
  persistence:
    type: FILE
    data-directory: var/data/paxos
  is-new-service: false
```

or

```
paxos:
  # (no more data directory - if present and *equal to* persistence config, explode)
  persistence:
    type: SQLITE
    data-directory: var/data/paxos
    magicDatabaseTuning: 42 # this is an advantage of method 2
  is-new-service: false
```

- This is more complicated than it looks, because existing mechanisms e.g. is-new-service now need to rely on potentially two directories, depending on the persistence type!
- Note that users can't actually configure SQLite yet: the config doesn't deserialize.

- Added several unit tests

- There is a simpler alternative:

```
paxos:
  data-directory: var/data/paxos
  persistence-mode: FILE # or SQLITE
  is-new-service: false
```

And the data directory is reusable (since Paxos uses directories, we can coordinate so that we use a SQLite file in the root). This is hackier though.

**Where should we start reviewing?**: PaxosInstallConfiguration.java

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/timelock/config/FilePaxosPersistenceConfiguration.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/config/FilePaxosPersistenceConfiguration.java
@@ -1,0 +1,24 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.timelock.config;
+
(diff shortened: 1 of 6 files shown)


j-baker:
Can we have another path exist for the sqlite data? Don't want it in the same dir as the legacy stuff.

j-baker:
aand i don't think we need to config the sqlite validation vs sqlite with no validation. It's like, we roll out with validation, if it looks good, we delete the validation code.

jeremyk-91:
Fair enough - I guess we could just say that from some version `FILE` _always_ runs with sqlite validation. I was thinking if we wanted to turn on validation first at internal herby stack before elsewhere, but that can be controlled via RCs of the timelock server.

Main thing to be aware of is that the validation currently does a non-live migration, but we could just blacklist to bail out of that if needed.

jeremyk-91:
Bleh, I'm not super happy with this approach: it is nice in some ways, but I think actually just something like

```
data-directory: /var/data/paxos
sqlite-persistence: # this is optional
  enabled: false
  data-directory: /var/data/paxosv2 # throws if equal
```

would be enough. Eventually, we would need to validate against the existence of either directory, but this doesn't feel quite right.

I was hoping to have something elegant, but this is not it. I'll have a new version of this up shortly.

PR #4743, opened 2020-04-28
AtlasFutures no longer destroys thread interruption state

carterkozak:
**Goals (and why)**:

Interrupted threads shouldn't forget they're interrupted.

Standard InterruptedException handling.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
+++ b/atlasdb-commons/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
@@ -93,6 +93,10 @@ public final class AtlasFutures {
             return listenableFuture.get();
         } catch (ExecutionException e) {
             throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            listenableFuture.cancel(true);
+            throw Throwables.rewrapAndThrowUncheckedException(e);
         } catch (Exception e) {
             throw Throwables.rewrapAndThrowUncheckedException(e);
         }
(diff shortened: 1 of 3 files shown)


atlasdb-commons/src/main/java/com/palantir/atlasdb/futures/AtlasFutures.java
```diff
@@ -93,6 +93,10 @@ public void close() {
             return listenableFuture.get();
         } catch (ExecutionException e) {
             throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
j-baker, line 97: I think that Throwables.rewrapAndThrow handles interrupted exception?

j-baker: Calls down to:

```java
    private static RuntimeException createPalantirRuntimeException(String newMessage, Throwable ex) {
        if (ex instanceof InterruptedException || ex instanceof InterruptedIOException) {
            return new PalantirInterruptedException(newMessage, ex);
        }
        return new PalantirRuntimeException(newMessage, ex);
    }
```

carterkozak: Good call, that's deep in there.

j-baker: yeah, it wasn't my decision to make the PalantirInterruptedException, but I do know it exists :)

carterkozak:
The jersey/undertow adapter cannot cancel resulting futures, but you may choose to implement a more specialized version there. The expection wrapping behavior in that case is likely to result in problems as well. Ideally that codepath would be removed for a single implementation though.

PR #4744, opened 2020-04-30
PG-148649: AgnosticLightResultSetImpl doesn't fill the finalizer queue

carterkozak:
Objects which implement `finalize` end up on the single-threaded
finalizer queue which causes servers to oom.

**Goals (and why)**
oom is bad.

Don't implement finalize.

yesterday

### DIFF
--- a/changelog/@unreleased/pr-4744.v2.yml
+++ b/changelog/@unreleased/pr-4744.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fix OOM caused by AgnosticLightResultSetImpl filling the finalizer
+    queue under heavy load.
+  links:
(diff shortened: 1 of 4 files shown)


carterkozak:
Thanks, Jeremy!

carterkozak:
@jeremyk-91 another flake on `AutobatchingPingableLeaderFactoryTests.canPingLeaderMatchingUuidAndClient` with `org.junit.ComparisonFailure: expected:<pingReturned[True(b8ed452f-5269-4a99-90c8-474f18888221, localhost:8080])> but was:<pingReturned[False(])>`

jeremyk-91:
Hmm, the latest one sounds like a flake. I'll keep an eye on the build.

PR #4746, opened 2020-04-30
[PaxosLog] Part 7b: Configuration Sketch :fire: CONTENTIOUS :fire:

jeremyk-91:
**Goals (and why)**:
- Allow us to turn on the SQLite log.

- Configuration looks like (though for now it won't deserialise):

```
install:
  data-directory: var/data/paxos #file 
  sqlite-persistence: # separate block in case we want more config options
    data-directory: var/data/meh
```
- Ban combinations of install and sqlite data directory that are equal, or where one is a subdirectory of the other.
- There is a minor change in the treatment of `isNewService`

- Added a good number of tests for the interactions between multiple data directories.

- Does this design make sense?
- My plan here was in some versions of TimeLock > X, setting SQLite characteristics enables validation mode (but will still always treat the file based record as the master). Then in some versions of TimeLock > Y > X, startup consists of executing the SQLite migration, this time for good, and then only using SQLite.

**Where should we start reviewing?**: PaxosInstallConfiguration.java


timelock-agent/src/main/java/com/palantir/timelock/config/SqlitePaxosPersistenceConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Value.Immutable
+// TODO (jkong): Allow serialization/deserialization, once we want this to actually be configurable
+public interface SqlitePaxosPersistenceConfiguration {
+    /**
+     * Indicates where the SQLite paxos state log files should be stored.
+     * Should not conflict with any other data directories that may be used by TimeLock to store Paxos state logs.
+     */
+    @JsonProperty("data-directory")
+    File dataDirectory();
```
jkozlowski, line 33: Is there any other config that you need here? Would it not make sense to inline this?

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
```diff
@@ -39,6 +44,9 @@ default File dataDirectory() {
         return new File("var/data/paxos");
     }
 
+    @JsonProperty("sqlite-persistence")
+    Optional<SqlitePaxosPersistenceConfiguration> sqlitePersistence();
```
jkozlowski, line 48: What if this was required and you actually checked state (meaning whether you already migrated or not/running validation) not just by looking if directories exist, but actually checking an internal sqlite table?

timelock-agent/src/main/java/com/palantir/timelock/config/SqlitePaxosPersistenceConfiguration.java
jeremyk-91: I'm not aware of anything right now, but I also don't know that I won't need anything more

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
jeremyk-91: The purpose of `isNewService` is not to determine whether migration/validation has taken place - it's to avoid accidental misconfiguration of the directories. 

That said an inclusive OR is probably not ideal. While we're running in validation mode it's probably best to just test the base directory. Once we are migrating:

- If both directories exist, accept.
- If the data directory exists and SQLite does not, accept: migration maybe hasn't taken place. The SQLite migration will poison the data directory so we will be safe overall
- Otherwise, reject: we create the data directory every time when is new service _is_ true (*).

Long term solution / if we don't want to keep (*) around:
- If both directories are absent, reject.
- If SQLite exists and data directory does not, check if SQLite has migrated, and accept if yes, reject if no.

We could make this required, but I would not want to run the validation/migration everywhere immediately (because of its possible effect on startup times).

timelock-agent/src/main/java/com/palantir/timelock/config/SqlitePaxosPersistenceConfiguration.java
jkozlowski: I would be surprised if we wanted more config, the only thing I can think of is tuning any sqlite options, but instead we should just hardcode them and make sure they work correctly.

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
jkozlowski: That makes sense. I would challenge if we actually should run this everywhere straight away: our infra should just downgrade if we overrun the startup budget, which means we'll need to fix that anyway

jeremyk-91: Cool. I was thinking of using the canaries as a mechanism for avoiding a possible rollback at large deployment (I'd assume the migration is O(N log N) with number of entries) but maybe that's too much effort, and I should just do the math here when we do pub lish the release and abort if we believe things will explode (e.g. if Herb takes 5 minutes, Brimstone will definitely be a no-go)

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
```diff
@@ -76,21 +84,46 @@ default void checkLeaderModeIsNotInAutoMigrationMode() {
...
         }
     }
+    default boolean doesDirectoryAlreadyExist() {
+        return dataDirectory().isDirectory() || sqlitePersistence()
+                .map(SqlitePaxosPersistenceConfiguration::dataDirectory)
+                .map(File::isDirectory).orElse(false);
+    }
 
+    @Value.Check
+    default void checkSqliteAndFileDataDirectoriesAreNotPossiblyShared() {
```
gmaretic, line 111: This is really tedious, but necessary 🌵

timelock-agent/src/test/java/com/palantir/timelock/config/PaxosInstallConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    public void serviceIsNewIfNoDataDirectoriesPresent() {
+        ImmutablePaxosInstallConfiguration.Builder partialConfiguration = createPartialConfiguration(
+                fileSubdirectory, sqliteSubdirectory);
+        assertThatCode(() -> partialConfiguration.isNewService(true).build()).doesNotThrowAnyException();
+        assertThatThrownBy(() -> partialConfiguration.isNewService(false).build())
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("The timelock data directories do not appear to exist.");
```
gmaretic, line 50: I't kinda funky that this is not an IllegalStateException tbh

timelock-agent/src/test/java/com/palantir/timelock/config/PaxosInstallConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Test
+    public void serviceIsNotNewIfBaseDataDirectoryIsPresent() {
+        ImmutablePaxosInstallConfiguration.Builder partialConfiguration = createPartialConfiguration(
+                getAndCreateRandomSubdirectory(), sqliteSubdirectory);
+        assertIsNotNewService(partialConfiguration);
+    }
+
+    @Test
+    public void serviceIsNotNewIfSqliteDataDirectoryIsPresent() {
```
gmaretic, line 61: Both of these tests have both base and sqlite?

timelock-agent/src/main/java/com/palantir/timelock/config/PaxosInstallConfiguration.java
jkozlowski: yeah, if we can get away without special config I'd say it's worth a try. Worst thing that happens as I understand it is that you just timeout the upgrade, and infra should rollback

gmaretic:
Yeah, I think this is fine now

timelock-agent/src/test/java/com/palantir/timelock/config/PaxosInstallConfigurationIntegrationTest.java
jeremyk-91: Yep, I'd have used `State` if fresh, but preserving legacy behaviour.

timelock-agent/src/test/java/com/palantir/timelock/config/PaxosInstallConfigurationIntegrationTest.java
jeremyk-91: Spoke offline: The confusion here was that `baseDirectory` and `sqliteDirectory` were path constants _that were not created_. We renamed them in the pairing session on Friday.

timelock-agent/src/main/java/com/palantir/timelock/config/SqlitePaxosPersistenceConfiguration.java
jeremyk-91: Going ahead with this for now, we can change later on if needed: I'm not confident there is a single correct set of parameters given variance in request loads.

PR #4747, opened 2020-04-30
[PaxosLog] Part 6b: Namespace Loader via Large Table

jeremyk-91:
See also #4740. This PR achieves a similar effect through a different schema. This schema was originally dropped because of perf and disk space considerations, but the implementation of the namespace tracking required for #4740 is probably much nicer.


leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -27,91 +28,109 @@
...
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 
 import com.palantir.common.persist.Persistable;
 
 public final class SqlitePaxosStateLog<V extends Persistable & Versionable> implements PaxosStateLog<V> {
-    private final String namespace;
+    private final Client namespace;
+    private final String seqId;
```
jkozlowski, line 39: For my knowledge, what is seqId?

gmaretic: Yes, this is not super descriptive, I imagine it's a combination of use case and acceptor/learner?

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -27,91 +28,109 @@
...
-        <V extends Persistable & Versionable> boolean[] writeBatchOfRounds(@Define("table") String table,
+        @SqlBatch("INSERT OR REPLACE INTO paxosLog (namespace, seqId, seq, val) VALUES ("
+                + ":namespace.value, :seqId, :round.sequence, :round.valueBytes)")
+        <V extends Persistable & Versionable> boolean[] writeBatchOfRounds(
+                @BindPojo("namespace") Client namespace,
+                @Bind("seqId") String seqId,
                 @BindPojo("round") Iterable<PaxosRound<V>> rounds);
+
+        @SqlQuery("SELECT DISTINCT(namespace) FROM paxosLog")
+        List<String> getAllNamespaces();
```
gmaretic, line 134: As in the other PR, should be 'Set'

gmaretic: Also would be nice if this returned 'Set<Client>' but doesn't seem to be trivial to implement, so this is fine

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/management/SqliteNamespaceLoader.java
```diff
@@ -0,0 +1,49 @@
...
+        this.jdbi = jdbi;
+    }
+
+    public static SqliteNamespaceLoader create(Supplier<Connection> connectionSupplier) {
+        Jdbi jdbi = Jdbi.create(connectionSupplier::get).installPlugin(new SqlObjectPlugin());
+        jdbi.withExtension(SqlitePaxosStateLog.Queries.class, SqlitePaxosStateLog.Queries::createTable);
+        return new SqliteNamespaceLoader(jdbi);
+    }
+
+    public List<Client> getAllRegisteredNamespaces() {
```
gmaretic, line 43: As above, make it a set?

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
jeremyk-91: Right, yeah. The problem is that `PaxosUseCase` as a concept only exists in TimeLock (this project is also used by embedded services and clustered lock in large internal product), which is why I opted to be generic and just describe it as a sequence ID. It _is_ meant to be `(PaxosUseCase, Acceptor|Learner)`, yes.

jeremyk-91: I think we discussed offline and went for `useCase` - this does run the risk of people conflating it with `PaxosUseCase`, but it is general enough still.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
jeremyk-91: Changing to Set works. As discussed a full change to `Set<Client>` is nontrivial

PR #4748, opened 2020-05-01
Fix bug in transaction starter if it fails midway through the batch

Jolyon-S:
**Goals (and why)**:
* Transaction starter uses an autobatcher to batch start transaction requests. It continually takes batches out until it has enough requests, but if it fails midway through this process, it will drop some immutable TS locks on the floor.
* This PR covers this case (and moves some code around in TransactionStarter).

* Add a try/catch around starting transactions in the batching method in TransactionStarter, and unlock all currently obtained locks if we fail.

* TransactionStarterTests focuses on successful starts, haven't added a test for this case (yet).

* Does this catch the case we are concerned about?

* TransactionStarter

Within a week.

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
@@ -53,16 +53,20 @@ final class TransactionStarter implements AutoCloseable {
     private final DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher;
     private final LockLeaseService lockLeaseService;
 
-    private TransactionStarter(LockLeaseService lockLeaseService, LockWatchEventCache lockWatchEventCache) {
-        this.autobatcher = Autobatchers
-                .independent(consumer(lockWatchEventCache))
-                .safeLoggablePurpose("transaction-starter")
-                .build();
+    private TransactionStarter(
+            DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher,
+            LockLeaseService lockLeaseService) {
+        this.autobatcher = autobatcher;
         this.lockLeaseService = lockLeaseService;
     }
 
     static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEventCache lockWatchEventCache) {
-        return new TransactionStarter(lockLeaseService, lockWatchEventCache);
+        DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher = Autobatchers
+                .independent(consumer(lockLeaseService, lockWatchEventCache))
+                .safeLoggablePurpose("transaction-starter")
+                .build();
+        return new TransactionStarter(autobatcher,
+                lockLeaseService);
     }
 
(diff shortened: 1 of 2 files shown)


lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -53,20 +53,16 @@
...
         this.lockLeaseService = lockLeaseService;
     }
 
     static TransactionStarter create(LockLeaseService lockLeaseService, LockWatchEventCache lockWatchEventCache) {
-        DisruptorAutobatcher<Void, StartIdentifiedAtlasDbTransactionResponse> autobatcher = Autobatchers
-                .independent(consumer(lockLeaseService, lockWatchEventCache))
-                .safeLoggablePurpose("transaction-starter")
-                .build();
-        return new TransactionStarter(autobatcher,
-                lockLeaseService);
```
Jolyon-S, line 69: I don't know what doing it this way gave us, but collapsing into the constructor allows us to make the consumer non-static so that it can call the unlock method.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -152,18 +148,25 @@ public void close() {
         };
     }
 
-    private static List<StartIdentifiedAtlasDbTransactionResponse> getStartTransactionResponses(
```
Jolyon-S, line 155: Non-static to call the unlock method.

Jolyon-S:
Shouldn't need to rework when the batch piece is merged.

Jolyon-S:
> Just for my understanding, have we considered making `unlock` static instead (with lockLeaseService as a parameter). It requires fewer changes and the utility can be tested independently.

While technically possible, I think the use of statics in that class were only because of the instantiation - those methods were never called outside of `TransactionStarter`, and it was only because of the `create` method that they needed to be static at all. I think it would be a bit strange for `unlock` to be static, but that's just my feeling on it. I think the removal of static is the better change, even if it is a larger diff (and from the test diff, you can see that it is still easy to test this way too).

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
jeremyk-91: The general idea here is that through much of Atlas (at least recently written parts), constructors don't generally create new resources: normally this allows better use of e.g. mocks or stubs in tests, though admittedly less so in this case. I'd still prefer this because conventionally `create` alerts the user that there is some not-completely-trivial construction going on, which there is in this case with the disruptor.

I'd suggest we keep making the consumer in a static way, though without changing the public API of `TransactionStarter`: if we extract a private static method that does what `unlock` does with a passed-in `LockLeaseService` (e.g. `determineAndExecuteUnlocks()`) it should be simple to call that from `getStartTransactionResponses`, and `unlock` can call that as well.

Jolyon-S: Interesting - that gives some context on why it was written this way (whereas my instinct is to use a constructor).

PR #4752, opened 2020-05-04
Add namespace validation as a param to KvTableMappingService

safreiberg:
Currently products are subclassing KvTableMappingService
to override namespace validation. This exposes the validation
as an optional parameter, with the default remaining the
'strict' validator.

**Goals (and why)**:


gmaretic:
LGTM

PR #4757, opened 2020-05-06
Implement better TM closing

gmaretic:
**Goals (and why)**:
The current hypothesis for why sometimes closing the transaction manager fails to close all the background tasks is that we end up blocking during closing one of them. This addresses the issue by timeboxing each.

We attempt to synchronously run the closing runnable, then attempt to cancel after 20 seconds. We defend against being unable to interrupt the task by using a fixed threadpool of size 5 just in case.

Unit tests

Should be fine, but shout if I missed something

tests?

ASAP

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
@@ -1,0 +1,67 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.util;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import com.palantir.common.concurrent.NamedThreadFactory;
+import com.palantir.common.concurrent.PTExecutors;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+
+public class SafeShutdownRunner implements AutoCloseable {
+    private final ExecutorService executor = PTExecutors
+            .newFixedThreadPool(5, new NamedThreadFactory("safe-shutdown-runner", true));
+    private final List<Throwable> failures = new ArrayList<>();
+    private final Duration timeoutDuration;
+
+    public SafeShutdownRunner(Duration timeoutDuration) {
+        this.timeoutDuration = timeoutDuration;
+    }
+
+    public void shutdownSafely(Runnable shutdownCallback) {
+        Future<?> future = executor.submit(shutdownCallback);
+        try {
+            future.get(timeoutDuration.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            failures.add(e);
+        } catch (ExecutionException e) {
+            failures.add(e.getCause());
+        } catch (TimeoutException e) {
+            future.cancel(true);
+            failures.add(e);
+        }
+    }
+
+    @Override
+    public void close() {
+        executor.shutdown();
+        if (!failures.isEmpty()) {
+            RuntimeException closeFailed = new SafeRuntimeException(
+                    "Close failed. Please inspect the code and fix the failures");
+            failures.forEach(closeFailed::addSuppressed);
+            throw closeFailed;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
```diff
@@ -0,0 +1,67 @@
...
+    public void close() {
+        executor.shutdown();
+        if (!failures.isEmpty()) {
+            RuntimeException closeFailed = new SafeRuntimeException(
+                    "Close failed. Please inspect the code and fix the failures");
+            failures.forEach(closeFailed::addSuppressed);
+            throw closeFailed;
+        }
+    }
+}
```
jeremyk-91, line 67:
```suggestion
}

```

atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
```diff
@@ -0,0 +1,128 @@
...
+        assertThatCode(runner::close).doesNotThrowAnyException();
+    }
+
+    @Test
+    public void exceptionsAreSuppressedAndReportedWhenClosing() {
+        SafeShutdownRunner runner = new SafeShutdownRunner(Duration.ofSeconds(1));
+
+        assertThatCode(() -> runner.shutdownSafely(throwingRunnable)).doesNotThrowAnyException();
+        assertThatThrownBy(runner::close)
+                .isInstanceOf(RuntimeException.class)
```
jeremyk-91, line 71: Can we have a message assert on this? I think the point here is that the exception thrown by the runner close isn't the exception that happened inside the shutdown operation

atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import com.palantir.common.concurrent.NamedThreadFactory;
+import com.palantir.common.concurrent.PTExecutors;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+
+public class SafeShutdownRunner implements AutoCloseable {
+    private final ExecutorService executor = PTExecutors
+            .newFixedThreadPool(5, new NamedThreadFactory("safe-shutdown-runner", true));
```
jeremyk-91, line 34: I guess this is fine. A service with tons of TransactionManagers could end up having a bunch of stuck threads though, especially if things are re-opened and closed (though if things hang uninterruptibly I guess there isn't much we can do).

Note: This does not actually create the threads until needed so most of the time you only create one thread even though we're making a fixed thread pool of 5

atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
```diff
@@ -0,0 +1,128 @@
...
+        runner.shutdownSafely(verySlowUninterruptibleRunnable);
+        runner.shutdownSafely(verySlowUninterruptibleRunnable);
+        runner.shutdownSafely(verySlowUninterruptibleRunnable);
+        runner.shutdownSafely(verySlowUninterruptibleRunnable);
+        runner.shutdownSafely(mockRunnable);
+
+        verify(verySlowUninterruptibleRunnable, times(5)).run();
+        verify(mockRunnable, never()).run();
+
+        closeAndAssertNumberOfTimeouts(runner, 7);
```
jeremyk-91, line 116: Interesting. I guess it never gets executed, and we drop out after the time limit.

atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
jkozlowski: I would almost say we want cached here? we want to close whatever we can.

jkozlowski: And we should make internal ski product notify us when it gets an exception from here so we can investigate immediately.

jeremyk-91: Ah, you're right that if we run, e.g.

```
<thing that takes 30 seconds>
<fast task>
<thing that hangs>
```

then this keeps around 2 threads for no reason. Cached is probably fine in that the number of threads is at worst bounded by number of callbacks. If we wanted to have a hard limit we could also do a ThreadPoolExecutor with core 0 and max 5 - I think either is fine

gmaretic: Agreed, cached is probably fine here

atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
gmaretic: That is implied by having the exception that happened inside the shutdown operation be the suppressed exception, although I guess you could self reference if you really wanted to make me sad. I can change to this being an instance of SRE, which will disambiguate

atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
gmaretic: Once all the threads are stuck in uninterruptible tasks, everything gets put on the queue and never gets to run as we time out waiting to pick it up from the queue

gmaretic: This is no longer applicable with the cached thread pool

atlasdb-commons/src/main/java/com/palantir/util/SafeShutdownRunner.java
jkozlowski: Also, it's bounded by number of bad tasks, which I suspect is 1. Either way is fine

PR #4758, opened 2020-05-07
[PaxosStateLog] Parallel reads and processing

gmaretic:
**Goals (and why)**:
Reading entries one by one is going to be too slow for our migration, at least if we wish to proceed with migration on startup. This PR allows us to parallelise the reads.

We use a fixed thread pool for the lifetime of reader, and do the reads in parallel

Unit tests

Should we be more defensive, not allow specifying the number of threads? Anything else?

Tests?

Early next week

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
@@ -1,0 +1,78 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.paxos;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.atlasdb.futures.AtlasFutures;
+import com.palantir.common.base.Throwables;
+import com.palantir.common.concurrent.NamedThreadFactory;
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
```diff
@@ -0,0 +1,78 @@
...
+                                .collect(Collectors.toList())))
+                .stream()
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(Collectors.toList());
+    }
+
+    private Optional<PaxosRound<V>> singleRead(long sequence) {
+        try {
+            return Optional.ofNullable(delegate.readRound(sequence))
```
jkozlowski, line 67: Doesn't #readRound have a exclusive lock in PaxosStateLogImpl?

gmaretic: Yeah, that's something we have to discuss, but it's a solvable problem just outside the scope of this PR

leader-election-impl/src/test/java/com/palantir/paxos/PaxosStateLogBatchReaderTest.java
```diff
@@ -0,0 +1,125 @@
...
+                    if (sequence % 2 == 0) {
+                        return null;
+                    }
+                    return valueForRound(sequence).persistToBytes();
+                });
+
+        try (PaxosStateLogBatchReader<PaxosValue> reader = createReader()) {
+            assertThat(reader.readBatch(START_SEQUENCE, BATCH_SIZE))
+                    .isEqualTo(EXPECTED_ROUNDS.stream()
+                            .filter(round -> round.sequence() % 2 != 0)
```
jeremyk-91, line 89: nit: might be worth extracting a local predicate

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
jeremyk-91: Yep, I think we should be good to change that to a R/W lock (though be careful when reading that class to make sure it's safe!)

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
```diff
@@ -50,7 +51,7 @@
 
 public class PaxosStateLogImpl<V extends Persistable & Versionable> implements PaxosStateLog<V> {
 
-    private final ReentrantLock lock = new ReentrantLock();
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
     private final Map<Long, Long> seqToVersionMap = Maps.newHashMap();
```
jeremyk-91, line 55: Hmm. Could we put `@GuardedBy("lock.writeLock()")`, or more defensively and probably better make this a concurrent hash map? Right now it's fine, but if later on a reader tries to use this (e.g. for caching) then memory visibility becomes a bit tricky.

PR #4759, opened 2020-05-07
[Proof of Concept] [Dialogue] Dialogue Migration for Atlas->TimeLock comms

jeremyk-91:
**Proof of concept only. Before merging, I want to add:**
- [ ] unit tests
- [ ] actually making dialogue configurable - this switches it on just so the ETE tests run with it
- [ ] sanity check all the TODOs I've dumped in
- [ ] implementation of some suggestions from @iamdanfox: changing `BlockingAndNonBlocking` stuff to `ShortAndLongTimeout` or something like that, also no need to specify `withFailedUrlCooldown` in the Dialogue service creator as it doesn't really matter. I might track the former separately.

**Goals (and why)**:
- We're broadly moving towards Dialogue. There is potentially big perf upside + better infra support, we should get on it.

- AtlasDbDialogueServiceProvider creates the services. We take in a `base` ReloadingFactory and do some decoration on top so that it matches the CJR semantics we used to have.
- PredicateSwitchingProxy to allow runtime config between one or the other.

Ran ETE tests using Dialogue, which worked! 
Didn't add unit tests, they will be added before merge.

- An alternative design ([jkong/dialogue-2](https://github.com/palantir/atlasdb/compare/jkong/dialogue-2)) exists that is more complex, but allows for better reuse of the ServiceCreator infra. I actually wrote that first, but I wasn't happy with it: it was too messy and it's fragile to Atlas devs making silly mistakes when instantiating services. It's worth a look for comparison - is the more restrictive but tighter design in this PR the right choice?
- Do my plans for multitenancy make sense? A multitenant service would need to instantiate an `AtlasDbDialogueServiceProvider` by themselves (but that's okay, they have access to atlas config). Then this can be passed in to the TransactionManager: if it doesn't exist we create our own, if it does we just use it.

**Where should we start reviewing?**: AtlasDbDialogueServiceProvider.java

### DIFF
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasDbDialogueFactories.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasDbDialogueFactories.java
@@ -1,0 +1,43 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.http.v2;
+
+import java.time.Duration;
+import java.util.Map;
+
+import com.palantir.conjure.java.client.config.ClientConfiguration;
+import com.palantir.conjure.java.client.config.NodeSelectionStrategy;
+import com.palantir.dialogue.clients.DialogueClients;
+import com.palantir.refreshable.Refreshable;
+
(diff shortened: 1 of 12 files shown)


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/AtlasDbDialogueFactories.java
```diff
@@ -0,0 +1,43 @@
...
+import com.palantir.refreshable.Refreshable;
+
+public final class AtlasDbDialogueFactories {
+    private static final Duration STANDARD_FAILED_URL_COOLDOWN = Duration.ofMillis(100);
+
+    private AtlasDbDialogueFactories() {
+        // no
+    }
+
+    public static DialogueClients.ReloadingFactory decorateForFailoverServices(
```
iamdanfox, line 34: If this method is only used in one place, why not just make it a `private static` in AtlasDbDialogueServiceProvider?

jeremyk-91: Fair point: this one's probably just a hangover from the previous aggressively customizable design. I'll roll it into ADDSP.

jeremyk-91:
Rough merging path:

- [ ] ShortAndLongTimeout
- [ ] DialogueClientOptions and AtlasDbDialogueServiceProvider
- [ ] Switching Proxy, Adapter, TransactionManager Wiring
- [ ] Runtime Configuration, defaulting to false
- [ ] ETE testing

PR #4760, opened 2020-05-11
Filter out leaderPaxos from list of namespaces

sudiksha27:
**Goals (and why)**:
leaderPaxos is not a namespace itself. It should be filtered out when loading namespaces from disk.

Filter out "leaderPaxos" while loading namespaces

None

None

DiskNamespaceLoader.java

Anytime


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private void createDirectoryForLeaderForEachClientUseCase(String namespace) {
+        new File(tempFolder.getRoot().toPath() + "/" +
+                Paths.get(PaxosTimeLockConstants.LEADER_PAXOS_NAMESPACE,
+                        PaxosTimeLockConstants.MULTI_LEADER_PAXOS_NAMESPACE).toString() + "/" +
+                namespace).mkdirs();
+    }
+
+    private void createDirectoryTimestampUseCase(String namespace) {
+        new File(tempFolder.getRoot().toPath() + "/" +
```
jeremyk-91, line 67: Some improvements I'd suggest here:
- The Path API has a cleaner way of doing this: `Paths.get(tempFolder.getRoot().toPath(), namespace).toFile()`. The main reason is that we can't guarantee `/` is always the path separator.
- `File::mkdirs` fails silently if creating the directory fails. It's probably better to surface such failures earlier in the test, so debugging is easier.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+        Set<String> namespaces = diskNamespaceLoader.getAllPersistedNamespaces().stream().map(client -> client.value()).collect(
+                Collectors.toSet());
+        assertThat(namespaces).containsExactlyInAnyOrder(NAMESPACE_1, NAMESPACE_2);
+    }
+
+    private void createDirectoryForLeaderForEachClientUseCase(String namespace) {
+        new File(tempFolder.getRoot().toPath() + "/" +
+                Paths.get(PaxosTimeLockConstants.LEADER_PAXOS_NAMESPACE,
+                        PaxosTimeLockConstants.MULTI_LEADER_PAXOS_NAMESPACE).toString() + "/" +
+                namespace).mkdirs();
```
jeremyk-91, line 63: (see comment on other mkdirs creation)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private void createDirectoryTimestampUseCase(String namespace) {
+        new File(tempFolder.getRoot().toPath() + "/" +
+                namespace).mkdirs();
+    }
+
+    @After
+    public void cleanup() {
+        tempFolder.delete();
```
jeremyk-91, line 73: Have a look at `TemporaryFolder.after()` 😄

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
+    private DiskNamespaceLoader diskNamespaceLoader;
+
+    @Before
+    public void setup() {
+        createDirectoryForLeaderForEachClientUseCase(NAMESPACE_1);
+        createDirectoryTimestampUseCase(NAMESPACE_2);
+        createDirectoryTimestampUseCase(PaxosTimeLockConstants.LEADER_PAXOS_NAMESPACE);
```
jeremyk-91, line 48: nit: maybe `createDirectoryInRootDataDirectory` to avoid confusion?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class DiskNamespaceLoaderTest {
+    private static final String NAMESPACE_1 = "namespace_1";
+    private static final String NAMESPACE_2 = "namespace_2";
+
+
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
+    private DiskNamespaceLoader diskNamespaceLoader;
```
jeremyk-91, line 42: nit: I _believe_ the Rule can be made final, and with that you can make this final and set it here as opposed to inside `setup()` (the test as written works, this is just general cleanliness).

sudiksha27: The DiskNamespaceLoader requires path to the tempFolder which is only available after the folder has been created. The only other option TemporaryFolder api gives is to provide the path of parent file. There is not a way to determine the path of the temp folder before it is created (which happens in the `before` block).

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.rules.TemporaryFolder;
+
+import com.palantir.atlasdb.timelock.paxos.PaxosTimeLockConstants;
+
+public class DiskNamespaceLoaderTest {
+    private static final String NAMESPACE_1 = "namespace_1";
+    private static final String NAMESPACE_2 = "namespace_2";
+
+
+    public final @Rule TemporaryFolder tempFolder = new TemporaryFolder();
```
jeremyk-91, line 37:
```suggestion
    @Rule
    public final TemporaryFolder tempFolder = new TemporaryFolder();
```

This is fine also, just for consistency with the other tests.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import com.palantir.atlasdb.timelock.paxos.PaxosTimeLockConstants;
+
+public class DiskNamespaceLoaderTest {
+    private static final String NAMESPACE_1 = "namespace_1";
+    private static final String NAMESPACE_2 = "namespace_2";
+
+
```
jeremyk-91, line 36: nit: 2 lines?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @Before
+    public void setup() {
+        diskNamespaceLoader = new DiskNamespaceLoader(tempFolder.getRoot().toPath());
+        createDirectoryForLeaderForEachClientUseCase(NAMESPACE_1);
+        createDirectoryInRootDataDirectory(NAMESPACE_2);
+    }
+
+    @Test
+    public void doesNotLoadLeaderPaxosAsNamespace() {
+        Set<String> namespaces = diskNamespaceLoader.getAllPersistedNamespaces().stream().map(client -> client.value()).collect(
```
jeremyk-91, line 50: nit: Let's use a method reference here (`Client::value` as opposed to `x -> x.value()`)

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -0,0 +1,71 @@
...
+                Collectors.toSet());
+        assertThat(namespaces).containsExactlyInAnyOrder(NAMESPACE_1, NAMESPACE_2);
+    }
+
+    private void createDirectoryForLeaderForEachClientUseCase(String namespace) {
+        if (Paths.get(tempFolder.getRoot().toPath().toString(),
```
jeremyk-91, line 56: nit: I think it's probably nicer to do `tempFolder.getRoot().toPath().resolve(LEADER_PAXOS_NAMESPACE).resolve(...)` which should behave in the same way

PR #4765, opened 2020-05-12
[PaxosStateLog] Migrator and unit tests

gmaretic:
**Goals (and why)**:
Implement PaxosStateLog migration
 
Pretty straightforward: use SqlitePaxosStateLogMigrationState to maintain migration state, read entries from the source log using 100 threads in parallel, then write a batch after each 10k reads

Unit tests for both state and migrator

Don't really have concerns about this, will add intergration tests going from file to sqlite in a follow up. This PR grew a bit, but it's mostly tests

Tests?

Tomorrow


leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -0,0 +1,83 @@
...
+import com.palantir.common.persist.Persistable;
+
+public final class PaxosStateLogMigrator<V extends Persistable & Versionable> {
+    @VisibleForTesting
+    static final int BATCH_SIZE = 10_000;
+
+    private final PaxosStateLog<V> sourceLog;
+    private final PaxosStateLog<V> destinationLog;
+    private final Persistable.Hydrator<V> hydrator;
+    private final SqlitePaxosStateLogMigrationState migrationState;
```
jeremyk-91, line 33: Non-actionable: sigh. This was _almost_ entirely agnostic of SQLite. Given we aren't going to be implementing another migrator any time soon though, I'd say leave it.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -0,0 +1,83 @@
...
+        migrator.runMigration();
+    }
+
+    private void runMigration() {
+        if (migrationState.hasAlreadyMigrated()) {
+            return;
+        }
+        destinationLog.truncate(destinationLog.getGreatestLogEntry());
+        long lowerBound = lowestSequenceToMigrate();
+        long upperBound = sourceLog.getGreatestLogEntry();
```
jeremyk-91, line 60: It's benign, but to avoid future confusion we should handle the edge case here: this method returns `PaxosAcceptor.NO_LOG_ENTRY` if there are zero entries in the source log, meaning that `numberOfBatches = (-1 - 0) / BATCH_SIZE + 1` which by Java semantics is actually `1`. Then there is a batch from zero to BATCH_SIZE entries that the reader will try and read.

leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -97,7 +96,7 @@ public void truncate(long toDeleteInclusive) {
         @SqlUpdate("CREATE TABLE IF NOT EXISTS paxosLog ("
                 + "namespace TEXT,"
                 + "useCase TEXT,"
-                + "seq BIGINT, "
+                + "seq BIGINT,"
```
jeremyk-91, line 99: 🎉

leader-election-impl/src/test/java/com/palantir/paxos/PaxosStateLogMigratorTest.java
```diff
@@ -0,0 +1,167 @@
...
+        PaxosStateLog<PaxosValue> mockLog = mock(PaxosStateLog.class);
+
+        when(mockLog.getLeastLogEntry()).thenReturn(lowerBound);
+        when(mockLog.getGreatestLogEntry()).thenReturn(upperBound);
+        when(mockLog.readRound(anyLong())).thenAnswer(invocation -> {
+            long sequence = (long) invocation.getArguments()[0];
+            if (sequence > upperBound || sequence < lowerBound) {
+                return null;
+            }
+            return valueForRound(sequence).persistToBytes();
```
jeremyk-91, line 125: nice!

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
gmaretic: 👍

PR #4766, opened 2020-05-12
[Dialogue] Part 4: Runtime Configuration and ETE Testing

jeremyk-91:
**Goals (and why)**:
- Users should be able to specify whether Dialogue should be used.
- Dialogue should be evaluated on an end-to-end basis.

- Specify `useDialogue` in `RemotingClientConfig` (a part of runtime config)
- Switch between clients dependent on runtime config.

- Run the existing ETEs with Dialogue.

- There isn't a test that the Dialogue versions of these ETE tests actually use Dialogue!
- A lot of copying.

**Where should we start reviewing?**: `TransactionManagers`


atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/CassandraTimeLockDialogueTestSuite.java
```diff
@@ -0,0 +1,49 @@
...
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+        TodoEteTest.class,
+        TargetedSweepEteTest.class,
+        CassandraTimestampsEteTest.class,
+        TimestampManagementEteTest.class,
+        CoordinationEteTest.class,
+        MultipleSchemaVersionsCoordinationEteTest.class,
+        LockWithTimelockEteTest.class
```
iamdanfox, line 38: I worry that lists of test suites like this will very quickly get out of sync with the c-j-r versions. Maybe OK if we can burn down the migration pretty fast, but i think if it takes longer than 2 weeks then this copy&pasting might be too much of a risk

jeremyk-91: This is a valid concern - though the rate at which AtlasDB's ETE test classes change is incredibly slow (for instance, this class was last changed in February 2019) so I wouldn't be too worried.

jeremyk-91: (And of course once the migration is done the non-Dialogue versions will go)

PR #4769, opened 2020-05-13
Gradle consistent versions

j-baker:
First stab at this. Unclear if it's actually going to work, I removed some important looking stuff.


gmaretic:
This looks good to me, @jeremyk-91 to approve?

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
```diff
@@ -181,55 +174,6 @@ public String renderToJson(byte[] name) {
         return sb.toString();
     }
 
-    public byte[] parseFromJson(String json, boolean allowPrefix) {
```
jeremyk-91, line 184: This is used by large internal product's command line interface to an AtlasDB database, so I'm not sure we can remove this?

jeremyk-91:
I will pick up fixing the removed method this week.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
jeremyk-91: This deletion is not needed after #4821

jeremyk-91:
Merging: build has passed, James wrote the PR and Grgur effectively approved it, I just resolved the merge conflicts.

PR #4770, opened 2020-05-13
Allow invalidating SweepStrategyManager cache

mswintermeyer:
**Goals (and why)**:

Useful to change the sweep strategy on an existing table without
bouncing the service.

Just expose a new method that allows cache invalidation.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
@@ -16,6 +16,7 @@
 package com.palantir.atlasdb.transaction.impl;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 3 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -51,7 +52,17 @@ public static SweepStrategyManager createDefault(KeyValueService kvs) {
...
+        return new SweepStrategyManager() {
+            @Override
+            public SweepStrategy get(TableReference tableRef) {
+                return sweepStrategySupplierLoadingCache.get().get(tableRef);
+            }
+
+            @Override
+            public void invalidateCaches(Set<TableReference> tableRefs) {
+                sweepStrategySupplierLoadingCache.get().invalidateAll(tableRefs);
+            }
```
jeremyk-91, line 64:
```
  /**
   * Discards any cached values for the {@code keys}. The behavior of this operation is undefined
   * for an entry that is being loaded (or reloaded) and is otherwise not present.
   *
   * @param keys the keys whose associated values are to be removed
   * @throws NullPointerException if the specified collection is null or contains a null element
   */
  void invalidateAll(@NonNull Iterable<?> keys);
```

Invalidating something that's not there looks like it could end up either way. It's a bit nasty, but should we take a read-write lock?

jeremyk-91: Discussed offline: The "undefined-ness" here just means that load and invalidate could race and end up either way. This is fine for the internal use case we have planned here.

j-baker:
This is pretty sketchy because it's not HA safe. Is this a problem?

mswintermeyer:
@j-baker good question. Discussed it with Jeremy, and it's not a problem for our use case at least. The node that calls this is the same node that will delete a cell and thus put it in the thorough sweep queue; the targeted sweeper background thread on a different node might not have its cache invalidated, but it trusts whatever's in the queue and will delete it appropriately. I was imagining that whatever application called this would be responsible for ensuring it gets called on all relevant nodes, but could add docs here to ensure future callers realize the implication. Does that sound reasonable?

PR #4772, opened 2020-05-13
[WIP] Retry after replace.

jkozlowski:
**Goals (and why)**:

If a request heroically detects the OkHttp bug, it will still cause an outage right now.


jkozlowski:
We will wait for dialogue, this provides marginal benefit since read timeouts are large.

PR #4764, opened 2020-05-12
[Draft] [LW] Client-side Lock Watch Event Log

Jolyon-S:
Draft PR for me to track my progress on this. Do not review. Will aim to make it much tidier on Thursday 14th May.

PR #4773, opened 2020-05-14
[Draft] [LW] Client-side Lock Watch Event Log approach 2

Jolyon-S:
I think this is better than the other approach in https://github.com/palantir/atlasdb/pull/4764.

edit: synced with Jakub, got a few things I need to refactor.


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchService.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.keyvalue.api.watch;
+
+public interface LockWatchService extends LockWatchManager {
```
Jolyon-S, line 19: just allows an extra method without changing API of LWM.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -225,6 +226,7 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
         } finally {
             postTaskContext = postTaskTimer.time();
             timelockService.tryUnlock(ImmutableSet.of(txAndLock.immutableTsLock()));
+            lockWatchService.removeTimestampFromCache(tx.getTimestamp());
```
Jolyon-S, line 229: Since we don't have the `CommitUpdate` / `PreCommitCondition` piece done, timestamps need to be removed from the cache at some point.

lock-api-objects/src/main/java/com/palantir/lock/watch/ClientLockWatchEventLogImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+        long startKey = logSnapshot.ceilingKey(oldestVersion);
+        long endKey = logSnapshot.floorKey(latestVersion);
+
+        return TransactionsLockWatchEvents.success(
+                new ArrayList<>(logSnapshot.subMap(startKey, endKey).values()),
+                timestampToVersion);
+    }
+
+    private synchronized void processSuccess(LockWatchStateUpdate.Success success) {
```
Jolyon-S, line 84: determining whether synchronized is necessary is one of the remaining pieces of work here.

lock-api-objects/src/main/java/com/palantir/lock/watch/ClientLockWatchEventLogImpl.java
```diff
@@ -0,0 +1,150 @@
...
+        setToLatestVersion(IdentifiedVersion.of(failed.logId(), Optional.empty()));
+        eventLog.clear();
+        seed = failedSnapshot(failed.logId());
+    }
+    
+    private static LockWatchStateUpdate.Snapshot failedSnapshot(UUID uuid) {
+        return LockWatchStateUpdate.snapshot(uuid, -1L, ImmutableSet.of(), ImmutableSet.of());
+    }
+
+    private void setToLatestVersion(IdentifiedVersion newVersion) {
```
Jolyon-S, line 105: This may be overly cautious, but I can imagine a case where a successful update A is actually done after B, even though B covers A and goes beyond, i.e.

A: |----|
B: |----------|

Don't want the version to go backwards on the same leader ID.

Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/4781

PR #4774, opened 2020-05-14
[PaxosStateLog] Implement Different Migration States

gmaretic:
**Goals (and why)**:
Add support for validation and migrated states

Pretty straightforward, although I would have liked to make it nicer

Added another test and updated existing ones

Any obvious better way to do this?

SqlitePaxosStateLogMigrationState.java

Tomorrow?

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogMigrationState.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogMigrationState.java
@@ -57,12 +57,20 @@ public final class SqlitePaxosStateLogMigrationState {
         execute(dao -> dao.migrateToVersion(namespace, useCase, States.VALIDATION.schemaVersion));
     }
 
+    public void migrateToMigratedState() {
+        execute(dao -> dao.migrateToVersion(namespace, useCase, States.MIGRATED.schemaVersion));
+    }
+
     public boolean hasMigratedFromInitialState() {
         return !Objects.equals(States.NONE.getSchemaVersion(), execute(dao -> dao.getVersion(namespace, useCase)));
     }
 
     public boolean isInValidationState() {
         return States.VALIDATION.getSchemaVersion().equals(execute(dao -> dao.getVersion(namespace, useCase)));
+    }
+
+    public boolean isInMigratedState() {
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLogMigrationState.java
```diff
@@ -52,12 +53,24 @@ private void initialize() {
         execute(Queries::createTable);
     }
 
-    public void finishMigration() {
-        execute(dao -> dao.finishMigration(namespace, useCase));
+    public void migrateToValidationState() {
+        execute(dao -> dao.migrateToVersion(namespace, useCase, States.VALIDATION.schemaVersion));
```
jeremyk-91, line 57: Discussed offline: Let's validate that we aren't in migrated state before doing this.

leader-election-impl/src/test/java/com/palantir/paxos/FileToSqlitePaxosStateLogIntegrationTest.java
```diff
@@ -54,7 +54,7 @@ public void setup() throws IOException {
     @Test
     public void emptyMigrationSucceeds() {
         migrate();
```
jeremyk-91, line 56: nit: let's check that it's false at some point

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogMigrationStateTest.java
```diff
@@ -44,43 +44,62 @@ public void setup() {
...
+        migrationState.migrateToMigratedState();
+        migrationState.migrateToMigratedState();
+        assertThat(migrationState.hasMigratedFromInitialState()).isTrue();
+        assertThat(migrationState.isInValidationState()).isFalse();
+        assertThat(migrationState.isInMigratedState()).isTrue();
+
+        migrationState.migrateToValidationState();
+        assertThat(migrationState.hasMigratedFromInitialState()).isTrue();
+        assertThat(migrationState.isInValidationState()).isTrue();
+        assertThat(migrationState.isInMigratedState()).isFalse();
```
jeremyk-91, line 81: Last 4 lines would be different if we're not allowing migrated -> validated state.

PR #4778, opened 2020-05-14
[Dialogue] Part 6: Road to Multitenancy

jeremyk-91:
**Goals (and why)**:
- Allow users to pass in a `DialogueClients.ReloadingFactory` for Atlas to use

- If a `ReloadingFactory` is configured, use it, otherwise create an Atlas-internal one.

- Added new tests around multiple users of a single underlying factory, both with the same and different TimeLocks.

- I'm not entirely happy with introducing a Dialogue class to `AtlasDbConfig`, but alternatives like allowing users to supply an `AtlasDbDialogueServiceProvider` also seemed awful - and if you're using `atlasdb-config` you will have Dialogue-using things on your classpath anyway.

**Where should we start reviewing?**: AtlasDbConfig.java

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
@@ -34,7 +34,6 @@ import com.palantir.atlasdb.cache.TimestampCache;
 import com.palantir.atlasdb.memory.InMemoryAtlasDbConfig;
 import com.palantir.atlasdb.spi.KeyValueServiceConfig;
 import com.palantir.atlasdb.sweep.queue.config.TargetedSweepInstallConfig;
-import com.palantir.dialogue.clients.DialogueClients;
 import com.palantir.exception.NotInitializedException;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
@@ -31,7 +31,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.cache.TimestampCache;
-import com.palantir.atlasdb.factory.AtlasDbDialogueServiceProvider;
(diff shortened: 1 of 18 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
```diff
@@ -271,6 +272,16 @@ public int getDefaultLockTimeoutSeconds() {
...
 
+    /**
+     * If set, a {@link com.palantir.dialogue.clients.DialogueClients.ReloadingFactory} that a
+     * {@link com.palantir.atlasdb.transaction.api.TransactionManager} based on this configuration should use.
+     * This may be useful for ensuring that connection pools are shared between multitenant services, for instance.
+     *
+     * WARNING: Providing reloading factories to multiple distinct TimeLock servers may result in undefined
+     * behaviour.
+     */
+    public abstract Optional<DialogueClients.ReloadingFactory> reloadingFactory();
```
j-baker, line 283: AtlasDbConfig is the wrong place, this should really be in TransactionManagers.

j-baker: AtlasDbConfig is json deserializable.

changelog/@unreleased/pr-4778.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: AtlasDB now allows users to pass in a `DialogueClients.ReloadingFactory`
+    in install configuration. This is likely to be useful for multitenant users of
```
j-baker, line 4: Please don't use 'multitenant' in this context, the noun is already 'taken' so it's likely to be confusing.

atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
jeremyk-91: Hmm, yeah it did feel like a better `TransactionManagers` fit, though I originally dumped it in `AtlasDbConfig` because the shared timestamp cache is there. I’ll move this across

changelog/@unreleased/pr-4778.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: AtlasDB now allows users to pass in a `DialogueClients.ReloadingFactory`
+    in install configuration. This is likely to be useful for AtlasDB-Proxy, which
```
gmaretic, line 4: 🔪

PR #4781, opened 2020-05-15
[Draft] [LW] Client-side Lock Watch Event Log approach 5

Jolyon-S:
Opened for personal tracking, WIP.


lock-api-objects/src/main/java/com/palantir/lock/watch/ClientLockWatchEventLogImpl.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.stream.Collectors;
+
+import com.google.common.collect.ImmutableList;
+
+public final class ClientLockWatchEventLogImpl implements ClientLockWatchEventLog {
```
j-baker, line 30: In real life, we expect there to be very little actual concurrency here (because we're batching start transactions at most like 2 or 3 threads will be in this class), but you've made it more complicated for yourself by having it. Please can you make this be a thread-unsafe class, where you access via a synchronized layer or something?

lock-api-objects/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Set;
 
 public interface LockWatchEventCache {
+    void removeTimestampFromCache(Long timestamp);
```
j-baker, line 22: long

Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/4798 and https://github.com/palantir/atlasdb/pull/4800

PR #4785, opened 2020-05-18
[PaxosStateLog] Implement locking for SqlitePSL

gmaretic:
**Goals (and why)**:
In a high concurrency setting we've been hitting Sqlite timeouts due to its locking mechanism. We now do our own locking to avoid that.

Quite straightforward

Added test that was failing on develop, now passes

Is the factory approach reasonable and clear enough?

Small

Today

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
@@ -19,6 +19,8 @@ package com.palantir.paxos;
 import java.sql.Connection;
 import java.util.OptionalLong;
 import java.util.Set;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SqlitePaxosStateLog.java
```diff
@@ -38,60 +40,95 @@
...
+    }
+
+    private <T> T executeRead(Function<Queries, T> call) {
+        sharedLock.readLock().lock();
+        try {
+            return execute(call);
+        } finally {
+            sharedLock.readLock().unlock();
+        }
     }
```
jeremyk-91, line 116: ⭐

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogTest.java
```diff
@@ -179,14 +186,27 @@ public void valuesAreDistinguishedAcrossSequenceIdentifiers() throws IOException
...
+        int numThreads = 100;
+        ExecutorService executor = PTExecutors.newFixedThreadPool(numThreads);
+        List<Future<?>> futures = IntStream.range(0, numThreads)
+                .mapToObj(ignore -> executor.submit(() -> {
+                    PaxosStateLog<PaxosValue> log = FACTORY.create(wrap(CLIENT_1, USE_CASE_1), conn);
+                    for (int i = 0; i < 30; i++) {
+                        log.writeRound(i, valueForRound(i));
+                    }
+                })).collect(Collectors.toList());
+        futures.forEach(future -> assertThatCode(() -> Futures.getUnchecked(future)).doesNotThrowAnyException());
```
jeremyk-91, line 207: This is kind of unpleasant but I didn't find a better way to write it

PR #4786, opened 2020-05-18
[PaxosStateLog] Reintroduce Migration and Verification with Locks

gmaretic:
**Goals (and why)**:
Now that we have locking, we can go ahead with the migration plans

Revert previous revert and wire in the locking mechanism

Existing tests pass

We are not locking in the migration state class. Doesn't seem necessary since we almost never write there, but we can add it if it's valuable

The revert is pretty straightforward, so mostly make sure nothing is strange and that the latest commit makes sense

Today, after the other PR

### DIFF
--- a/leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogMigrationStateTest.java
+++ b/leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogMigrationStateTest.java
@@ -41,7 +41,7 @@ public class SqlitePaxosStateLogMigrationStateTest {
     public void setup() {
         connSupplier = SqliteConnections
                 .createDefaultNamedSqliteDatabaseAtPath(tempFolder.getRoot().toPath());
-        migrationState = SqlitePaxosStateLogMigrationState.create(NAMESPACE_AND_USE_CASE, connSupplier);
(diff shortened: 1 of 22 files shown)


leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogMigrationStateTest.java
```diff
@@ -99,8 +99,8 @@ public void finishingMigrationForOneNamespaceDoesNotSetFlagForOthers() {
     public void finishingMigrationForOneUseCaseDoesNotSetFlagForOthers() {
         migrationState.migrateToValidationState();
 
-        SqlitePaxosStateLogMigrationState otherState = SqlitePaxosStateLogMigrationState
-                .create(ImmutableNamespaceAndUseCase.of(Client.of("namespace"), "other"), connSupplier);
+        SqlitePaxosStateLogMigrationState otherState = new SqlitePaxosStateLogFactory().createMigrationState(
+                ImmutableNamespaceAndUseCase.of(Client.of("namespace"), "other"), connSupplier);
```
jeremyk-91, line 103: These don't happen simultaneously with the other instances in the other tests in this test class, so while this _might_ be a bit sketchy it's ok

PR #4787, opened 2020-05-19
[PaxosStateLog] Do not migrate

gmaretic:
**Goals (and why)**:
Migration caused an outage due to taking too long. We will have to revisit the approach


gmaretic:
This was done in another PR

PR #4788, opened 2020-05-19
[PaxosStateLog] Validate without migration

gmaretic:
**Goals (and why)**:
Do validations but without migrating

First time we do a write, remember the sequence, then only report mismatch if it's a later sequence

None

Should we do something differently

small

Now

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
@@ -22,8 +22,8 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Objects;
+import java.util.OptionalLong;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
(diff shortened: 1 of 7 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -53,6 +56,7 @@
     private final PaxosStateLog<V> currentLog;
     private final PaxosStateLog<V> experimentalLog;
     private final Persistable.Hydrator<V> hydrator;
+    private final CompletableFuture<Long> firstEntry = new CompletableFuture<>();
```
jeremyk-91, line 59: discussed offline: I'd probably write `AtomicReference<OptionalLong>`, though this is fine as well

leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -175,12 +201,15 @@ private long getExtremeLogEntry(Function<PaxosStateLog<V>, Long> extractor) {
...
-            Long experimentalResult = safeReadFromExperimental(extractor::apply);
-            if (!Objects.equals(result, experimentalResult) && result != PaxosAcceptor.NO_LOG_ENTRY) {
-                log.error("Mismatch in getting the extreme log entry between legacy and current implementations."
-                                + " Legacy result {}, current result {}.",
-                        SafeArg.of("legacy", result),
-                        SafeArg.of("current", experimentalResult));
+            if (firstEntry.isDone()) {
+                Long experimentalResult = safeReadFromExperimental(extractor::apply);
+                if (!Objects.equals(result, experimentalResult) && result != PaxosAcceptor.NO_LOG_ENTRY
+                        && result >= Futures.getUnchecked(firstEntry)) {
```
jeremyk-91, line 207:
```suggestion
                        && result > Futures.getUnchecked(firstEntry)) {
```

I think this is needed, otherwise we might end up claiming assertions on the minimum after writing just the very first entry

jeremyk-91, line 207: We probably don't need the second term here, but this is fine

jeremyk-91: In hindsight, following a suggestion from @j-baker, it might be worth not doing this check actually, since this would be a good way to test a hypothesis that the old entries aren't actually particularly important. It might still be worth keeping track of the first entry and dumping that as an arg in the result.

leader-election-impl/src/main/java/com/palantir/paxos/VerifyingPaxosStateLog.java
```diff
@@ -117,13 +141,15 @@ public void writeRound(long seq, V round) {
...
         try {
             byte[] result = currentLog.readRound(seq);
-            byte[] experimentalResult = safeReadFromExperimental(psl -> psl.readRound(seq));
-            if (!Arrays.equals(result, experimentalResult)) {
-                log.error("Mismatch in reading round for sequence number {} between legacy and current "
-                        + "implementations. Legacy result {}, current result {}.",
-                        SafeArg.of("sequence", seq),
-                        UnsafeArg.of("legacy", hydrateIfNotNull(result)),
-                        UnsafeArg.of("current", hydrateIfNotNull(experimentalResult)));
+            if (firstEntry.isDone() && Futures.getUnchecked(firstEntry) <= seq) {
```
jeremyk-91, line 144: In hindsight, following a suggestion from @j-baker, it might be worth not doing this check actually, since this would be a good way to test a hypothesis that the old entries aren't actually particularly important. It might still be worth keeping track of the first entry and dumping that as an arg in the result.

PR #4790, opened 2020-05-19
[Dialogue] Part 7: Conjure for Selected Legacy Lock Endpoints

jeremyk-91:
**Goals (and why)**:
- We've seen that Dialogue is broadly beneficial for the performance of RPCs. However, it is difficult to benefit completely from Dialogue while our APIs do not conform to Conjure. `ConjureTimelockResource` covers 70 to 85 percent of traffic; most of the remainder is through the V1 `LockService`, and specifically three main endpoints: `refreshLockRefreshTokens`, `unlockSimple`, and `lockAndGetHeldLocks`. The first is about 2.3-2.5x more frequent than the latter two, which have about the same frequency.
- The legacy lock service is a bit scary and is a source of technical debt, so it would be good to move it to Conjure. There is however a wealth of classes and APIs which are not particularly well understood. Given broad usage across a multitude of consumers that is not well understood, it's also _very_ dangerous to break APIs here.

- Expose a conjure API on TimeLock that is able to delegate RLRT and US to the same lock service as exposed on the legacy endpoints. Do not actually stop exposing old endpoints.

Added a new test that going through either API _will_ affect the locks on the other side in the way we expect.

- I only did the first two endpoints: the last has a more complicated serialized form in terms of the request that needs to be sent, and should be forthcoming in a separate PR. (It's also only about ~20 percent of this part of the traffic.)
- Don't like the copypasta in timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java, but not sure how to do it nicer without either some weird abusive use of interfaces or reflection.

**Where should we start reviewing?**: lock-api.yml


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/ConjureLockV1Resource.java
```diff
@@ -0,0 +1,123 @@
...
+            ConjureSimpleHeldLocksToken request) {
+        return exceptionHandler.handleExceptions(() -> {
+            SimpleHeldLocksToken serverToken = new SimpleHeldLocksToken(
+                    request.getTokenId(), request.getCreationDateMs());
+            return Futures.immediateFuture(lockServices.apply(namespace).unlockSimple(serverToken));
+        });
+    }
+
+    @VisibleForTesting
+    static List<LockRefreshToken> getLegacyTokens(List<ConjureLockRefreshToken> request) {
```
gmaretic, line 88: Rename to `toLegacyTokens`?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/ConjureLockV1Resource.java
```diff
@@ -0,0 +1,123 @@
...
+        return request.stream()
+                .map(token -> new LockRefreshToken(token.getTokenId(), token.getExpirationDateMs()))
+                .collect(Collectors.toList());
+    }
+
+    @VisibleForTesting
+    static Set<ConjureLockRefreshToken> getConjureTokens(Set<LockRefreshToken> serverTokens) {
```
gmaretic, line 95: as above

gmaretic: Also, these two methods should just live in a separate utility class and be public imo

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -320,6 +326,42 @@ public void canGetAllNamespaces() {
                 .collect(Collectors.toSet());
     }
 
+    @Test
+    public void directLegacyAndConjureLockServicesInteractCorrectly() throws InterruptedException {
```
gmaretic, line 330: Test is a bit verbose, but thorough 🎉

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/ConjureLockV1Resource.java
jeremyk-91: That's a good call. I will do this refactor

PR #4791, opened 2020-05-20
Always send full last known version, including id.

jkozlowski:
**Goals (and why)**:

This is an obviously more correct way to specify this in the API. If the leader changes, the new leader should be forced to recompute a clean snapshot. Without pushing the client's view of the leader id, it is technically possible that timelock would not compute a snapshot if the log event sequences somehow overlap in weird ways.

Pass Optional<IdentifiedVersion> around, which is a bit more awkward, but I believe, correct.

This changes the conjure API, so **technically** a break. But the client's would've only pushed Optional.empty() there, so should be fine.

### DIFF
--- a/lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
+++ b/lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
@@ -25,18 +25,17 @@ import com.google.common.collect.ImmutableSet;
 @SuppressWarnings("FinalClass") // mocks
 public class NoOpLockWatchEventCache implements LockWatchEventCache {
     public static final LockWatchEventCache INSTANCE = new NoOpLockWatchEventCache();
-    private static final IdentifiedVersion FAKE = ImmutableIdentifiedVersion
-            .of(UUID.randomUUID(), Optional.empty());
+    private static final IdentifiedVersion FAKE = ImmutableIdentifiedVersion.of(UUID.randomUUID(), 0L);
     private static final TransactionsLockWatchEvents NONE = TransactionsLockWatchEvents.failure(
-            LockWatchStateUpdate.snapshot(UUID.randomUUID(), 0L, ImmutableSet.of(), ImmutableSet.of()));
+            LockWatchStateUpdate.snapshot(UUID.randomUUID(), -1L, ImmutableSet.of(), ImmutableSet.of()));
 
     private NoOpLockWatchEventCache() {
         // singleton
     }
 
     @Override
-    public IdentifiedVersion lastKnownVersion() {
-        return FAKE;
+    public Optional<IdentifiedVersion> lastKnownVersion() {
+        return Optional.empty();
     }
 
     @Override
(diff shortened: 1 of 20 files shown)


lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
```diff
@@ -25,18 +25,17 @@
...
 
     private NoOpLockWatchEventCache() {
         // singleton
     }
 
     @Override
-    public IdentifiedVersion lastKnownVersion() {
-        return FAKE;
+    public Optional<IdentifiedVersion> lastKnownVersion() {
+        return Optional.empty();
```
jkozlowski, line 38: Thinking about this a bit more: I think this is a change in behavior, in that every startTransactions call now will cause timelock to compute a snapshot.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -36,12 +36,18 @@ types:
   definitions:
     default-package: com.palantir.atlasdb.timelock.api
     objects:
+      ConjureIdentifiedVersion:
+        fields:
+          id: uuid
+          version: safelong
```
jeremyk-91, line 42: since we'd already declared the external java `Long` I'd suggest using it? I get that this usage is probably safe, but probably best to be consistent especially with places where we really don't want this to be `safelong` (i.e. timestamps)

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -36,12 +36,18 @@ types:
   definitions:
     default-package: com.palantir.atlasdb.timelock.api
     objects:
+      ConjureIdentifiedVersion:
```
jeremyk-91, line 39: nit: the Lock Watch stuff never really had a non-conjurized (?) form, so it's probably fine to just name this `IdentifiedVersion`

lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
jeremyk-91: I guess snapshotting right now is cheap since there aren't actually going to be events, but this could lead to unintended consequences later on. It seems safer to return `Optional.of(FAKE)`? - you still take a snapshot when there's a leader election, but that's relatively uncommon

jkozlowski: Exactly, I think snapshotting is cheap so I shouldn't worry about it.

jkozlowski: Returning FAKE is same as Optional.empty, see impl. I think this is semantically clearer, since we're simply not tracking the version.

timelock-api/src/main/conjure/timelock-api.yml
jkozlowski: So problem is that that IdentifiedVersion is already defined in the lock-api-objects. So until we clean that up, I think it makes sense to keep this as ConjureIdentifiedVersion.

timelock-api/src/main/conjure/timelock-api.yml
jkozlowski: Done

PR #4793, opened 2020-05-21
[PaxosStateLog] Splitting log

gmaretic:
**Goals (and why)**:
Implement log that delegates reads and writes based on a cutoff sequence

As discussed, mark if we see reads and writes at old sequences, otherwise delegate everything to new log.

Trivial

Do we want something else?

tiny

This week


leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public long getGreatestLogEntry() {
+        return currentLog.getGreatestLogEntry();
+    }
+
+    @Override
+    public void truncate(long toDeleteInclusive) {
+        currentLog.truncate(toDeleteInclusive);
```
jeremyk-91, line 79: Seems like we should truncate both? Or just throw UnsupportedOperationException given where we're deploying this

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,81 @@
...
+            return currentLog.readRound(seq);
+        } else {
+            markLegacyRead.run();
+            return legacyLog.readRound(seq);
+        }
+    }
+
+    @Override
+    public long getLeastLogEntry() {
+        return currentLog.getLeastLogEntry();
```
jeremyk-91, line 69: We should talk about this one: I think it needs careful thought. The acceptors for example are just going to fail requests with a TruncatedStateLogException (see `PaxosAcceptor.checkLogIfNeeded()`)

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,122 @@
...
+    public void truncate(long toDeleteInclusive) {
+        // noop
+    }
+
+    @Value.Immutable
+    abstract static class SplittingParameters<V extends Persistable & Versionable> {
+        abstract PaxosStateLog<V> legacyLog();
+        abstract PaxosStateLog<V> currentLog();
+        abstract Runnable markLegacyWrite();
+        abstract Runnable markLegacyRead();
```
jeremyk-91, line 108: Good to disambiguate the types

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,122 @@
...
+        abstract long cutoffInclusive();
+
+        @Value.Check
+        void cutoffEntryMustBeInCurrentLogOrEqualToNoEntry() {
+            try {
+                Preconditions.checkState(cutoffInclusive() == PaxosAcceptor.NO_LOG_ENTRY
+                                || currentLog().readRound(cutoffInclusive()) != null,
```
jeremyk-91, line 115: This does incur a read whenever creating this object.

This is constructed once per (client, use-case). Let's go ahead with this for now I think

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public static <V extends Persistable & Versionable> PaxosStateLog<V> create(SplittingParameters<V> parameters) {
+        return new SplittingPaxosStateLog<>(
+                parameters.legacyLog(),
+                parameters.currentLog(),
+                parameters.markLegacyWrite(),
+                parameters.markLegacyRead(),
+                parameters.cutoffInclusive(),
+                new AtomicLong(Math.min(parameters.legacyLog().getLeastLogEntry(), parameters.cutoffInclusive())));
```
jeremyk-91, line 61: Discussed offline: This is ok as long as we make sure internal server doesn't say that it is fully ready until this has completed for all clients.

jeremyk-91: We can also hack in an optimisation to make `getLeastLogEntry()` faster at a majority of installations.

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
```diff
@@ -0,0 +1,122 @@
...
+    public long getGreatestLogEntry() {
+        return currentLog.getGreatestLogEntry();
+    }
+
+    /**
+     * This implementation is a noop to ensure correctness of {@link #getLeastLogEntry()}.
+     */
+    @Override
+    public void truncate(long toDeleteInclusive) {
+        // noop
```
jeremyk-91, line 100: `log.warn("Tried to truncate with an impl that doesn't support it")` or something like that please, this may confuse unsuspecting users.

leader-election-impl/src/main/java/com/palantir/paxos/SplittingPaxosStateLog.java
jeremyk-91: discussing offline:
- Actually, we want to check that at least one entry exists in the new log after the cutoff - and this can be checked by verifying that the greatest log entry in `currentLog()` is at least the cutoff

PR #4794, opened 2020-05-21
[Dialogue] Part 9: Shims for Lesser Used Services: AtlasDB -> TimeLock

jeremyk-91:
**Goals (and why)**:
- Improve performance on lesser used services to some extent
- Don't go through the pain/effort of fully Conjurising these services.

- Allow the usage of the JaxRs client shim for Dialogue across lesser used services.

- Existing integration tests should suffice.

Nothing in particular. Does the organisation make sense?

**Where should we start reviewing?**: AtlasDbDialogueServiceProvider

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -1035,8 +1035,11 @@ public abstract class TransactionManagers {
         ServiceCreator creator = ServiceCreator.withPayloadLimiter(
                 metricsManager, timelockServerListConfig, userAgent, remotingClientConfig);
 
-        LockRpcClient lockRpcClient = new TimeoutSensitiveLockRpcClient(
+        LockRpcClient cjrLockRpcClient = new TimeoutSensitiveLockRpcClient(
                 ShortAndLongTimeoutServices.create(creator, LockRpcClient.class));
+        LockRpcClient dialogueLockRpcClient = serviceProvider.getLockRpcClient();
+        LockRpcClient lockRpcClient = createRuntimeConfigSwitchedProxy(
+                dialogueLockRpcClient, cjrLockRpcClient, remotingClientConfig, LockRpcClient.class);
 
         LockService lockService = AtlasDbMetrics.instrumentTimed(
                 metricsManager.getRegistry(),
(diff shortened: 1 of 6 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
```diff
@@ -80,6 +83,14 @@ private AtlasDbHttpClients() {
         return wrapWithOkHttpBugHandling(type, clientFactory);
     }
 
+    public static <T> T createDialogueProxy(TaggedMetricRegistry registry, Class<T> type, Channel channel) {
```
iamdanfox, line 86: I'd recommend putting the word 'shim' in here so that it's super clear which ones are using the 'full dialogue' codegen vs the feign thingy.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueShimFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import com.palantir.conjure.java.client.jaxrs.JaxRsClient;
+import com.palantir.conjure.java.dialogue.serde.DefaultConjureRuntime;
+import com.palantir.dialogue.Channel;
+
+public final class DialogueShimFactory {
+    private DialogueShimFactory() {
+        // Nö
+    }
+
+    public static <T> T create(Class<T> type, Channel channel) {
```
iamdanfox, line 28: only used in one place... inline?

jeremyk-91: I tried to write it that way at first... but most of the HTTP libs aren't visible in the `atlasdb-config` project (which seems kind of reasonable - that's the intention behind the `atlasdb-conjure` project) so this class was needed

atlasdb-config/src/main/java/com/palantir/atlasdb/http/AtlasDbHttpClients.java
jeremyk-91: Makes sense

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/DialogueShimFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.palantir.atlasdb.http.v2;
+
+import com.palantir.conjure.java.client.jaxrs.JaxRsClient;
+import com.palantir.conjure.java.dialogue.serde.DefaultConjureRuntime;
+import com.palantir.dialogue.Channel;
+
+public final class DialogueShimFactory {
+    private DialogueShimFactory() {
+        // Nö
```
gmaretic, line 25: 🎉

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1305,4 +1325,20 @@ default Runnable close() {
         TransactionService transactionService();
         Optional<TransactionSchemaInstaller> schemaInstaller();
     }
+
+    @Value.Immutable
+    @Value.Style(stagedBuilder = false)
+    public interface RemoteProxies<T> {
```
gmaretic, line 1331: 🎉

jeremyk-91: I actually didn't write this at first, I just had the two Ts, and then when doing a self-review I thought this would be nice

PR #4796, opened 2020-05-22
[PaxosStateLog] Migrate from Cutoff + Persist Cutoff

gmaretic:
**Goals (and why)**:
This allows us to specify the starting point for the migration, and remembers the cutoff point which is necessary for the migration and verification we discussed.

Pretty straightforward

Fixed existing tests and added some new ones

This shouldn't be contentious

PaxosStateLogMigrator.java

Monday


leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -83,5 +91,25 @@ private long lowestSequenceToMigrate() {
         PaxosStateLog<V> destinationLog();
         Persistable.Hydrator<V> hydrator();
         SqlitePaxosStateLogMigrationState migrationState();
+        OptionalLong migrateFrom();
```
jeremyk-91, line 94: discuss: either make this non-`Optional`, or add a safety buffer in the ranges above before doing the actual migration

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -29,52 +38,51 @@
...
-    private long lowestSequenceToMigrate() {
-        long leastLogEntry = sourceLog.getLeastLogEntry();
-        return leastLogEntry == PaxosAcceptor.NO_LOG_ENTRY ? 0L : leastLogEntry;
+    private void writeBatchRetryingUpToFiveTimes(PaxosStateLog<V> target, List<PaxosRound<V>> batch) {
+        for (int retryCount = 0; retryCount < 5; retryCount++) {
+            try {
+                target.writeBatchOfRounds(batch);
+                return;
+            } catch (Exception e) {
+                Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
```
jeremyk-91, line 82: For debugging: Would it make sense to log the exceptions here?

leader-election-impl/src/test/java/com/palantir/paxos/FileToSqlitePaxosStateLogIntegrationTest.java
```diff
@@ -79,9 +85,19 @@ private void migrateAndVerifyValuesForSequences(LongStream sequences) {
...
-
-        assertThat(migratedValues).isEqualTo(roundsToValues(rounds));
+        long cutoff = source.getGreatestLogEntry() - 10;
+        Map<Long, byte[]> targetEntries = readMigratedValuesFor(expectedValues);
+
+        targetEntries.entrySet().stream()
+                .filter(entry -> entry.getKey() < cutoff)
+                .map(Map.Entry::getValue)
+                .map(Assertions::assertThat)
+                .forEach(AbstractAssert::isNull);
```
jeremyk-91, line 95: 🚀

leader-election-impl/src/test/java/com/palantir/paxos/SqlitePaxosStateLogMigrationStateTest.java
```diff
@@ -105,4 +105,16 @@ public void finishingMigrationForOneUseCaseDoesNotSetFlagForOthers() {
         otherState.migrateToValidationState();
         assertThat(otherState.hasMigratedFromInitialState()).isTrue();
     }
+
+    @Test
+    public void defaultCutoffIsLongMinValue() {
+        assertThat(migrationState.getCutoff()).isEqualTo(Long.MIN_VALUE);
```
jeremyk-91, line 111: discussed offline: `PaxosAcceptor.NO_LOG_ENTRY` seems more appropriate

leader-election-impl/src/test/java/com/palantir/paxos/PaxosStateLogMigratorTest.java
```diff
@@ -66,75 +81,104 @@ public void emptyLogMigrationSuccessfullyMarksAsMigrated() {
     }
 
     @Test
-    public void logMigrationSuccessfullyMigratesEntries() {
+    public void logMigrationWithNoLowerBoundMigratesOnlyGreatest() {
```
jeremyk-91, line 84: note: This test is expected to change because we plan on making the migrator remember a buffer of entries before the one externally specified.

leader-election-impl/src/test/java/com/palantir/paxos/PaxosStateLogMigratorTest.java
```diff
@@ -66,75 +81,104 @@ public void emptyLogMigrationSuccessfullyMarksAsMigrated() {
...
+        LongStream.rangeClosed(lowerBound, expectedCutoff - 1)
+                .mapToObj(sequence -> readRoundUnchecked(target, sequence))
+                .map(Assertions::assertThat)
+                .forEach(AbstractAssert::isNull);
+        KeyedStream.of(LongStream.rangeClosed(expectedCutoff, upperBound).boxed())
+                .map(sequence -> getPaxosValue(target, sequence))
+                .mapKeys(PaxosStateLogTestUtils::valueForRound)
+                .entries()
+                .forEach(entry -> assertThat(entry.getKey()).isEqualTo(entry.getValue()));
     }
```
jeremyk-91, line 128: Let's document the behaviour where `expectedCutoff > upperBound` (which currently is not migrating anything, which is weird but we may not want to ban that).

gmaretic: Yeah, this actually doesn't play nice with splitting log so changed

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -33,8 +35,12 @@
 import com.palantir.common.persist.Persistable;
 
 public final class PaxosStateLogMigrator<V extends Persistable & Versionable> {
+    public static final Logger log = LoggerFactory.getLogger(PaxosStateLogMigrator.class);
```
jeremyk-91, line 38: nit: private

PR #4797, opened 2020-05-22
[PaxosStateLog] Use WAL and remove locking

gmaretic:
**Goals (and why)**:
Use WAL and higher timeout instead of our own locking

Pretty straightforward

Passes the more aggressive version of the concurrency test, and is much faster at it!

Timeout of 5 seconds reasonable?

SqliteConnections.java

Monday


leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -42,9 +43,17 @@ private SqliteConnections() {
...
+        config.setPragma(SQLiteConfig.Pragma.JOURNAL_MODE, SQLiteConfig.JournalMode.WAL.getValue());
+        config.setBusyTimeout(5000);
+        SQLiteConnectionPoolDataSource dataSource = new SQLiteConnectionPoolDataSource();
+        dataSource.setUrl(target);
+        dataSource.setConfig(config);
+
         return () -> {
             try {
-                return DriverManager.getConnection(target);
+                return dataSource.getConnection();
```
jeremyk-91, line 56: 💯 We end up avoiding having to do reflection!

PR #4798, opened 2020-05-26
[Draft] [LW] Client-side lock watch event cache for timestamps

Jolyon-S:
Still need to write tests and wire through before this is ready to review.

**Goals (and why)**:
Provide a means for the client to get a mapping from start timestamp to version for a batch of transactions.

* Implemented `LockWatchEventCache` methods
* Main processing methods `processStartTransactionUpdate` and `getEventsForTransactions` are synchronised as we expect them to be accessed in a single-threaded way.

TODO

* Do we leave ourselves open to any race conditions?
* Are there any ways that we can have a memory leak?
* Does this expose the necessary operations that we will need?
* [Edit]: Rewritten to not use a `Caffeine` cache, so that we can reliably get the min version.

* `LockWatchEventCacheImpl`

Once it is ready, within the week (but not yet!)


lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCacheImpl.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+
+public final class LockWatchEventCacheImpl implements LockWatchEventCache {
```
Jolyon-S, line 27: Going to split some of the internals of this class to make testing easier.

lock-api/src/main/java/com/palantir/lock/watch/NoOpClientLockWatchEventLog.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.palantir.lock.watch;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+import com.google.common.collect.ImmutableSet;
+
+public enum NoOpClientLockWatchEventLog implements ClientLockWatchEventLog {
```
Jolyon-S, line 25: May not be necessary if we merge #4800 at the same time, but sketched here for now.

lock-api/src/test/java/com/palantir/lock/watch/LockWatchEventCacheImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+@RunWith(MockitoJUnitRunner.class)
+public final class LockWatchEventCacheImplTest {
```
Jolyon-S, line 42: Work in progress; this is messy at the moment but will be nicer when I split some methods out from the underlying impl.

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCacheImpl.java
```diff
@@ -0,0 +1,158 @@
...
+        timestampsToDelete.forEach(timestampMap::remove);
+        earliestVersion = Optional.ofNullable(timestampMap.firstEntry()).map(Map.Entry::getValue);
+    }
+
+    @VisibleForTesting
+    Map<Long, IdentifiedVersion> getTimestampToVersionMap(Set<Long> startTimestamps) {
+        Map<Long, IdentifiedVersion> timestampToVersion = new HashMap<>();
+        // This may return a map of a different size to the input, if timestamps are missing (which we do not expect in
+        // practice).
+        startTimestamps.forEach(timestamp -> {
```
Jolyon-S, line 126: Throw!

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCacheImpl.java
```diff
@@ -0,0 +1,158 @@
...
+     */
+    @Override
+    public synchronized TransactionsLockWatchEvents getEventsForTransactions(
+            Set<Long> startTimestamps,
+            Optional<IdentifiedVersion> version) {
+        return lockWatchEventLog.getEventsForTransactions(getTimestampToVersionMap(startTimestamps), version);
+    }
+
+    @Override
+    public void removeTimestampFromCache(Long timestamp) {
```
Jolyon-S, line 106: Synchronised

Jolyon-S:

PR #4799, opened 2020-05-26
TransactionManager#startTransactions API.

jkozlowski:
**Goals (and why)**:

* Return OpenTransaction that knows how to close the transaction wrt. open resources.
* Remove start/finish pair of methods that is very C-like.
* Hide the immutable timestamp lock token.
* OpenTransactions can contain any lockwatch-related methods, so they do not have to be exposed outside of this flow, as it doesn't make sense otherwise.

* It feels like an improvement, but there might be something even more drastic.


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/StartTransactionRequest.java
```diff
@@ -13,23 +13,12 @@
...
-    LockToken immutableTsLock();
-
-    static TransactionAndImmutableTsLock of(Transaction transaction, LockToken immutableTsLock) {
-        return ImmutableTransactionAndImmutableTsLock.builder()
-                .transaction(transaction)
-                .immutableTsLock(immutableTsLock)
-                .build();
-    }
+public interface StartTransactionRequest {
+    PreCommitCondition preCommitCondition();
```
jkozlowski, line 23: I am thinking we could have a different Interface passed here here that is only lockwatch related, so we don't pollute PreCommitCondition.

jkozlowski, line 23: Possibly, this dealing with List<PreCommitCondition> might cleanup some more code (where we have CombinedCondition* type things), but I'd hold off for now.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -209,26 +214,74 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
+
+
+        @Override
+        public List<OpenTransaction> getTransactions() {
+            return transactions;
+        }
+
+        //    @Override
+        //    public TransactionsLockWatchEvents getEvents(Optional<IdentifiedVersion> lastKnownVersion) {
+        //        return null;
```
jkozlowski, line 246: This would have to work with an empty list of started transactions, and just send the update up to the latest we know about.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/OpenTransaction.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.api;
+
+import com.palantir.atlasdb.metrics.Timed;
+
+public interface OpenTransaction extends Transaction {
```
jkozlowski, line 21: Interested in another name, but this is how we call this in atlasdb-proxy.

jeremyk-91: I think this is fine, enough of us work on atlasdb-proxy that the naming should be obvious :)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/OpenTransactions.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.palantir.atlasdb.transaction.api;
+
+import java.util.List;
+import java.util.Optional;
+
+import com.palantir.lock.watch.IdentifiedVersion;
+import com.palantir.lock.watch.TransactionsLockWatchEvents;
+
+public interface OpenTransactions {
```
jeremyk-91, line 25: Think we discussed this one, I'm still slightly in favour of `StartTransactionsResponse`

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/OpenTransactions.java
```diff
@@ -0,0 +1,35 @@
...
+
+    List<OpenTransaction> getTransactions();
+
+    /**
+     * Returns a condensed view of new lock watch events since lastKnownVersion for the set of started transactions.
+     *
+     * @param lastKnownVersion exclusive start version to get events from
+     */
+    TransactionsLockWatchEvents getEvents(Optional<IdentifiedVersion> lastKnownVersion);
```
jeremyk-91, line 34: Should we document/put in the contract the behaviour around FAILUREs at -1 (for systems with no-op managers i.e. embedded)? I don't remember exactly how we handle that case

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/OpenTransactions.java
jkozlowski: Cool, will rename now.

jkozlowski: Renamed

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/OpenTransactions.java
jkozlowski: Discussed offline: this ends up just sending a dummy SNAPSHOT always, meaning lock watches are effectively off.

PR #4800, opened 2020-05-26
[Draft] [LW] Client-side lock watch event log to back cache

Jolyon-S:
Built on top of https://github.com/palantir/atlasdb/pull/4798. Some work still to be done to wire this in - creating several PRs to make it ETE first.

**Goals (and why)**:
* Provide a backing class to the `LockWatchEventCache` so that it can effectively get events between two timestamps.

* Used a `ConcurrentSkipListMap` to track mapping from `IdentifiedVersion -> LockWatchEvent`
* Old events are evicted when we are told we can (from the cache, i.e. events are no longer needed) - this depends on the cache not hanging on to very old transactions in order to not OOM ourselves.
* Snapshots are generated using the yet-to-be-implemented `SnapshotUpdater`
* Class is largely synchronised as this flow should be single-threaded in practice (as it is autobatched).

Yet to be implemented. TODO

Same as in #4798:
* Do we leave ourselves open to any race conditions?
* Are there any ways that we can have a memory leak?
* Does this expose the necessary operations that we will need?

`ClientLockWatchEventLogImpl`

Within the week that it is done, preferably.


Jolyon-S:

PR #4801, opened 2020-05-26
[Draft] [LW] Client-side lock watch snapshot replayer

Jolyon-S:
Built on top of https://github.com/palantir/atlasdb/pull/4800 and https://github.com/palantir/atlasdb/pull/4798 -  still in progress but split here to track.

**Goals (and why)**:
* If the user of Atlas is far behind in their version compared to the client-side event log, they need to get a snapshot independently of Atlas getting a snapshot from Timelock
* Therefore, there needs to be snapshot construction logic of some kind on the Atlas client side.

* Echoes the implementation in `LockEventLogImpl`, which essentially maintains two sets, one for currently held locks, and the other for currently watched locks, and updates depending on the events observed.

Yet to be implemented. TODO

* Same as the other two. The code here is pretty simple as everything in the class is single-threaded (this is guaranteed by the caller, but is enforced again here anyway).
* Also, the class names are really meh.

`ClientLockWatchEventSnapshotUpdaterImpl`

Not yet, not finished.


Jolyon-S:

PR #4802, opened 2020-05-26
Filter out leaderPaxos from list of namespaces

sudiksha27:
**Goals (and why)**:
Extension of https://github.com/palantir/atlasdb/pull/4760.
leaderPaxos is not a namespace itself and should be filtered out while loading namespaces via SqliteNamespaceLoader.

Filter out "leaderPaxos" while loading namespaces

Existing test modified

Today


changelog/@unreleased/pr-4802.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: fix
+fix:
+  description: |
+    `leaderPaxos` is filtered out from the set of persisted namespaces.
```
jeremyk-91, line 4: This is fine, but I would be a bit more specific on explaining the subset of users this actually applied to (i.e. previously *if, and only if, you had used SQLite*, leaderPaxos would be returned as part of the set of persisted namespaces). Otherwise it sounds like it was always being returned

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -17,38 +17,73 @@
...
         createDirectoryForLeaderForEachClientUseCase(NAMESPACE_1);
         createDirectoryInRootDataDirectory(NAMESPACE_2);
     }
 
     @Test
-    public void doesNotLoadLeaderPaxosAsNamespace() {
-        Set<String> namespaces = diskNamespaceLoader.getAllPersistedNamespaces().stream().map(Client::value).collect(
-                Collectors.toSet());
+    public void doesNotLoadLeaderPaxosAsNamespace() throws ExecutionException, InterruptedException {
+        Set<String> namespaces = timeLockManagementResource.getNamespaces(AUTH_HEADER).get();
```
jeremyk-91, line 86: Do we need to create the leader paxos directory somewhere before this?

sudiksha27: we do create one is the setup - `createDirectoryForLeaderForEachClientUseCase`, it is of `LeaderForEachClient` paxos use case pattern but I think that's sufficient for this test.

changelog/@unreleased/pr-4802.v2.yml
sudiksha27: 👍

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
jeremyk-91: I see. Yep, that makes sense

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/management/DiskNamespaceLoaderTest.java
```diff
@@ -17,38 +17,73 @@
...
     private static final String NAMESPACE_1 = "namespace_1";
     private static final String NAMESPACE_2 = "namespace_2";
+    private static final AuthHeader AUTH_HEADER = AuthHeader.valueOf("Bearer omitted");
+    @Mock private Function<String, TimeLockServices> serviceFactory;
+    @Mock private Supplier<Integer> maxNumberOfClientsSupplier;
+
+    private final MetricsManager metricsManager = new MetricsManager(
+            new MetricRegistry(),
+            DefaultTaggedMetricRegistry.getDefault(),
+            unused -> false);
```
jeremyk-91, line 58:
```suggestion
    private final MetricsManager metricsManager = MetricsManagers.createForTests();
```

PR #4803, opened 2020-05-26
[PDS-120786, PaxosLog] Avoid Overhead From Repeatedly Taking Locks And Starting New Connections

jeremyk-91:
Pairing with @gmaretic 

**Goals (and why)**:
- Address performance issues e.g. PDS-120786

- Most of the overhead involved opening/closing connections. This PR maintains a single connection that delegates to another connection, but is able to re-create it if it gets closed.
- Add some important SQLite pragmas.

- Proxy tests.

- If getting a new connection takes a long time this is not ideal: everyone waits for a new connection.
- ResilientDatabaseConnectionProxy is nasty, but Connection has a ton of methods that I'd otherwise need to manually write. AutoDelegate was probably a reasonable option and would be better perf, but tricky to reason about its correctness.

**Where should we start reviewing?**: `ResilientDatabaseConnectionProxy`


leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -18,46 +18,54 @@
...
-        return () -> {
-            try {
-                return dataSource.getConnection();
-            } catch (SQLException e) {
-                throw new RuntimeException(e);
-            }
-        };
+        HikariConfig hikariConfig = new HikariConfig();
+        hikariConfig.setDataSource(dataSource);
+        hikariConfig.setMaximumPoolSize(1);
```
j-baker, line 67: can we have this be more than 1? Say 16?

j-baker: Like 1 just seems wrong.

j-baker: (and maybe remove the locking mode exclusive pragma...

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -18,46 +18,54 @@
...
+
+    private static DataSource buildDataSource(Path path) {
         createDirectoryIfNotExists(path);
         String target = String.format("jdbc:sqlite:%s", path.resolve(DEFAULT_SQLITE_DATABASE_NAME).toString());
 
         SQLiteConfig config = new SQLiteConfig();
         config.setPragma(SQLiteConfig.Pragma.JOURNAL_MODE, SQLiteConfig.JournalMode.WAL.getValue());
-        config.setBusyTimeout(5000);
+        config.setPragma(SQLiteConfig.Pragma.LOCKING_MODE, SQLiteConfig.LockingMode.EXCLUSIVE.getValue());
+        config.setPragma(SQLiteConfig.Pragma.SYNCHRONOUS, SQLiteConfig.SynchronousMode.FULL.getValue());
```
j-baker, line 59: don't you want EXTRA here?

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -18,46 +18,54 @@
...
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
 
 /**
  * This class is responsible for creating Sqlite connections to an instance.
  * There should be one instance per timelock.
  */
 public final class SqliteConnections {
     private static final String DEFAULT_SQLITE_DATABASE_NAME = "sqliteData.db";
+    private static final LoadingCache<Path, DataSource> DATA_SOURCE_LOADING_CACHE = Caffeine.newBuilder()
```
j-baker, line 41: this seems like a code smell. Why can't the datasource be injected?

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -18,46 +18,54 @@
...
-            try {
-                return dataSource.getConnection();
-            } catch (SQLException e) {
-                throw new RuntimeException(e);
-            }
-        };
+        HikariConfig hikariConfig = new HikariConfig();
+        hikariConfig.setDataSource(dataSource);
+        hikariConfig.setMaximumPoolSize(1);
+        return new HikariDataSource(hikariConfig);
```
gmaretic, line 68: This is much better 🎉

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
j-baker: Like, just to be clear here: 1 means that you'll only have one query in flight to SQLite at a time, which seems strictly worse than shared locking.

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
jeremyk-91: The enum didn't exist in the java driver. I've set it to the string, not sure there's a better solution.

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
gmaretic: That's a good point, having more than 1 connection and removing exclusive locking means we can do parallel reads

jeremyk-91: Yeah, in hindsight our tests probably overindexed on writes (which might be fair given that migration specifically has been causing us pain).

leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
jeremyk-91: Ah, yes this makes sense. We should have been more aggressive about refactoring the structures we hackily built to get the benchmarks green.

jeremyk-91:
Merging: I can't approve my own PR, but this PR is the work of Grgur and I basically.

PR #4807, opened 2020-05-27
Remove TimeLockServerDownIntegrationTest.

jkozlowski:
* This test takes many minutes to run,
  the behavior is unit tested in TimeLockClientTest.


jkozlowski:
Maybe it doesn't help so much

PR #4808, opened 2020-05-27
Cleanup #validatePreCommitRequirementsOnReadIfNecessary.

jkozlowski:
**Goals (and why)**:

This method is only ever called with the start timestamp pre-having any other timestamp.


jkozlowski:
Will hold off on this for now.

PR #4809, opened 2020-05-27
[Dialogue] Part 10: :hot_pepper: TimeLock Internode by Shim

jeremyk-91:
_This is a little bit beyond the scope of what we planned internally, but it wasn't very hard and should give us a decent perf win, so why not?_

~_DO NOT MERGE until the existing pains we have around remoting are resolved please - we don't need to introduce yet another variable to the mix._~ Said ticket has been resolved! (mostly, at least we're unblocked from Atlas upgrades?)

**Goals (and why)**:
- Move internode TimeLock communications to use Dialogue by shim: we should go down from piles of different clients to just six.
- Unlock the path to pure Dialogue on TimeLock communications (even if that _does_ require jumping through quite a lot more hoops that are _not_ addressed in this PR).

- Allow creation of proxies with services of the form `timelock-<retry?>-<host>`.
- Update timelock code paths to use the TimeLockDialogueServiceProvider.

All TimeLock integration tests run with Dialogue. This isn't configurable, so TimeLock internode communication in all ETE tests that use TimeLock (independent of whether Atlas talks to TimeLock using Dialogue) also uses Dialogue.

- Is this too risky / should this be a config flag? I'd argue this is fine: we're confident in the lib generally speaking, the tests do run with it, and mechanisms exist to kill this if needed.

**Where should we start reviewing?**: TimeLockDialogueServiceProvider.java


timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -117,6 +129,28 @@ public static TimeLockAgent create(
...
+                .proxyConfiguration(install.cluster().cluster().proxyConfiguration())
+                .build();
+        return TimeLockDialogueServiceProvider.create(
+                metricsManager.getTaggedRegistry(),
+                baseFactory,
+                timeLockServerListConfig,
+                AuxiliaryRemotingParameters.builder()
+                        .userAgent(userAgent)
+                        .shouldLimitPayload(false)
+                        .shouldRetry(false) // Is subsequently overridden.
```
jeremyk-91, line 148: We could work around this by making it Optional

jeremyk-91:
Will need some re-basing after #4810 goes through, since we both enriched the Paxos creation code paths.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockDialogueServiceProvider.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private static Map<String, RemoteServiceConfiguration> createRemoteServiceConfigurations(
+            ServerListConfig serverListConfig, UserAgent versionedAgent, AuxiliaryRemotingParameters parameters) {
+        return KeyedStream.of(serverListConfig.servers())
+                .map(server -> ImmutableServerListConfig.builder()
+                        .from(serverListConfig)
+                        .servers(ImmutableList.of(server))
+                        .build())
+                .flatMapEntries((uri, singleServerConfig) -> {
+                    RemoteServiceConfiguration nonRetrying = ImmutableRemoteServiceConfiguration.builder()
```
gmaretic, line 73: nit: extract method

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
gmaretic: 👍

jeremyk-91: Turns out this causes a lot of unexpected headaches with the logic for detecting different remote clients, so I'm not going to block on this if that's okay?

gmaretic: Fair enough, can live with this

jeremyk-91: never mind, I coded a silly bug

PR #4810, opened 2020-05-28
[PaxosStateLog] Use Singlethreaded access to Sqlite

gmaretic:
**Goals (and why)**:
Tests were flaking due to timeouts on writes in a multithreaded approach. This avoids the problem by ensuring there is always only a single connection.

Use just a single connection. This is ideal for writes, but makes reads somewhat slower as reads are allowed to happen in parallel.
This made other tests fail, revealing that we in fact instantiate LocalPaxosComponents twice. We additionally created a connection to sqlite for the namespace loader, and we were creating multiple connections in PaxosTimelockIntegration tests on every bounce.

Existing tests

Discussed implementing our own read/write locking which would allow us to have parallel reads. This would give us faster reads, but decided against it for now as it makes the implementation more difficult and does not seem necessary.
The fixes for the test failures / bad instantiations of connections may be cleaner so let's discuss.
 
tiny

ASAP as fixes test flakes

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
@@ -19,8 +19,6 @@ package com.palantir.paxos;
 import java.io.IOException;
 import java.nio.file.Path;
 
-import javax.sql.DataSource;
-
 import org.apache.commons.io.FileUtils;
 import org.sqlite.SQLiteConfig;
 import org.sqlite.javax.SQLiteConnectionPoolDataSource;
@@ -41,7 +39,7 @@ public final class SqliteConnections {
         // no
     }
 
-    public static DataSource getPooledDataSource(Path path) {
+    public static HikariDataSource getPooledDataSource(Path path) {
         createDirectoryIfNotExists(path);
         String target = String.format("jdbc:sqlite:%s", path.resolve(DEFAULT_SQLITE_DATABASE_NAME).toString());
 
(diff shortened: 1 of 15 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SqliteConnections.java
```diff
@@ -41,22 +39,22 @@ private SqliteConnections() {
         // no
     }
 
-    public static DataSource getPooledDataSource(Path path) {
+    public static HikariDataSource getPooledDataSource(Path path) {
```
jeremyk-91, line 42: discussed offline: `CloseableDataSource` might be nicer in terms of least privilege, but this is already on the classpath everywhere and we don't really want to add another layer of delegation

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -113,7 +112,7 @@ public static TimeLockAgent create(
                 registrar,
                 paxosResources,
                 userAgent);
-        agent.createAndRegisterResources();
+        agent.createAndRegisterResources(paxosResources.leadershipContextFactory().sqliteDataSource());
```
jeremyk-91, line 115: Might be better if we create it here and inject it into both

timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosResourcesFactory.java
```diff
@@ -60,13 +64,16 @@ public static PaxosResources create(
             ExecutorService sharedExecutor) {
         PaxosRemoteClients remoteClients = ImmutablePaxosRemoteClients.of(install, metrics);
 
-        ImmutablePaxosResources.Builder resourcesBuilder =
-                setupTimestampResources(install, metrics, paxosRuntime, sharedExecutor, remoteClients);
+        HikariDataSource sqliteDataSource = SqliteConnections.getPooledDataSource(install.sqliteDataDirectory());
```
jeremyk-91, line 67: let's create this inside `TimeLockAgent` and inject it here

PR #4811, opened 2020-05-28
Lock watches wiring and exposing API for PreCommitCondition.

jkozlowski:
**Goals (and why)**:

Different approach to https://github.com/palantir/atlasdb/pull/4636.


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchTransactionManager.java
```diff
@@ -16,20 +16,13 @@
...
-    /**
-     * Returns a condensed view of new lock watch events since lastKnownVersion for a set of transactions identified by
-     * their start timestamps.
-     *
-     * @param startTimestamps  a set of start timestamps identifying transactions
-     * @param lastKnownVersion exclusive start version to get events from
-     */
-    TransactionsLockWatchEvents getEventsForTransactions(Set<Long> startTimestamps,
-            Optional<IdentifiedVersion> lastKnownVersion);
+public interface LockWatchTransactionManager extends TransactionManager {
```
j-baker, line 25: we've found extending txn manager to be a bit of a folly in the past. Would much appreciate if this method were not an interface method, or were otherwise on TransactionManager. At a minimum, it means you get to pick if SchemaAwareTransactionManager should implement LockWatchTransactionManager _facepalm_

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -331,6 +333,12 @@ public static TransactionManager createInMemory(Set<Schema> schemas) {
     @JsonIgnore
     @Value.Derived
     public TransactionManager serializable() {
+        return lockWatch();
+    }
+
+    @JsonIgnore
+    @Value.Derived
+    public LockWatchTransactionManager lockWatch() {
```
j-baker, line 341: yeah, this change can't happen - breaks too much binary backcompat internally

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPreCommitCondition.java
```diff
@@ -0,0 +1,59 @@
...
+        this.cache = cache;
+        this.watchCommitCondition = watchCommitCondition;
+    }
+
+    void initialize(Transaction newTxn) {
+        Preconditions.checkState(txn.compareAndSet(null, newTxn), "Already initialized");
+    }
+
+    @Override
+    public void throwIfConditionInvalid(long timestamp) {
```
j-baker, line 44: you only need the start timestamp for the txn. You might want to pass that to the precommit condition explicitly as a new (default) method.

j-baker: like, seems reasonable that we'd have method that initializes with the start timestamp... but fine with this for now.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPreCommitCondition.java
```diff
@@ -0,0 +1,59 @@
...
+        if (timestamp != getTransaction().getTimestamp()) {
+            CommitUpdate commitUpdate = cache.getCommitUpdate(getTransaction().getTimestamp());
```
j-baker, line 46: can we just get rid of the 'CommitUpdate' type? Like, to me there should be no difference between the commit update and the start update.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPrecommitCondition.java
```diff
@@ -0,0 +1,4 @@
+package com.palantir.atlasdb.transaction.impl.watch;
+
+public class WatchPrecommitCondition {
```
j-baker, line 3: kill?

lock-api-objects/src/main/java/com/palantir/lock/watch/CommitUpdate.java
```diff
@@ -0,0 +1,50 @@
...
+
+package com.palantir.lock.watch;
+
+import java.util.Set;
+
+import org.immutables.value.Value;
+
+import com.palantir.lock.LockDescriptor;
+
+public interface CommitUpdate {
```
j-baker, line 25: think this should not exist, rather this should be implemented at the level of the precommit condition.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/LockWatchTransactionManagerImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    public TransactionManager delegate() {
+        return transactionManager;
+    }
+
+    @Override
+    public StartTransactionsResponse startTransactions(Optional<IdentifiedVersion> lastKnownVersion,
+            List<StartTransactionRequest> requests) {
+        List<CombinedCondition> amendedConditions = requests.stream().map(
+                request -> new CombinedCondition(request.preCommitCondition(),
+                        new WatchPreCommitCondition(lockWatchEventCache, request.watchCommitCondition())))
```
j-baker, line 62: this API is a bit goofy. The point of a PrecommitCondition is that you should just be able to pass it in, it should be entirely pluggable.

Here, the reason you need the extra method is because you need to get the start timestamp of the transactions that have been passed in.

This would be much easier if we just passed the precommit condition the start timestamp at some point.

j-baker: like, the rough minimal thing i'm advocating for here is that I should be able to write a

```java
interface CacheLocator extends PrecommitCondition {}

interface CacheFactory {
    Cache getCache(CacheLocator initializedLocator, Transaction transactionForFallbacks);
    CacheLocator newLocator();
}

and then use it like:

CacheLocator locator = cacheFactory.newLocator();
txnManager.runWithConditionWithRetry(locator, txn -> {
    Cache cache = cacheFactory.getCache(locator, txn);
});
```

Of course, the consumer doesn't use this API, but in Atlas it should read like this.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchTransactionManager.java
jkozlowski: Removed

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jkozlowski: Removed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPreCommitCondition.java
jkozlowski: Added default method to initialize with startTs.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPreCommitCondition.java
jkozlowski: I think this was added for ease of use: it gives you the descriptors that had any events, so the cache dedupes everything for you. Could just return raw events if you want.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/WatchPrecommitCondition.java
jkozlowski: Gone

lock-api-objects/src/main/java/com/palantir/lock/watch/CommitUpdate.java
jkozlowski: As per previous comment, can remove if you want.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/watch/LockWatchManagerImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    public LockWatchManagerImpl(NamespacedConjureLockWatchingService lockWatchingService) {
+        this.lockWatchingService = lockWatchingService;
+    }
+
+    @Override
+    public synchronized void registerWatches(Set<LockWatchReferences.LockWatchReference> newLockwatches) {
+        lockWatchReferences.addAll(newLockwatches);
+    }
+
+    public synchronized void reregisterWatches() {
```
jkozlowski, line 41: I think I can just make this have a thread that calls this on timelock once every few seconds and not bother with being clever.

jkozlowski: Gonna do it

Jolyon-S:
Just had a really quick look, and the APIs exposed seem fine to me.

lock-api/src/main/java/com/palantir/lock/client/CommitTimestampGetter.java
```diff
@@ -0,0 +1,105 @@
...
+            List<Long> commitTimestamps = new ArrayList<>();
+            while (commitTimestamps.size() < count) {
+                GetCommitTimestampsResponse response = leaseService.getCommitTimestamps(cache.lastKnownVersion(),
+                        count - commitTimestamps.size());
+                commitTimestamps.addAll(process(batch.subList(commitTimestamps.size(), count), response, cache));
+            }
+
+            for (int i = 0; i < count; i++) {
+                batch.get(i).result().set(commitTimestamps.get(i));
+            }
```
jeremyk-91, line 75: non-actionable: This is a little tricky, but makes sense (the cache updates itself so in the next batch you only ask for the updates from that point onwards).

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1101,8 +1100,12 @@ private static LockAndTimestampServices getLockAndTimestampServices(
                 .timestamp(new TimelockTimestampServiceAdapter(remoteTimelockServiceAdapter))
                 .timestampManagement(timestampManagementService)
                 .timelock(remoteTimelockServiceAdapter)
-                .lockWatcher(lockWatcher)
-                .close(remoteTimelockServiceAdapter::close)
+                .lockWatcher(lockWatchManager)
+                .eventCache(lockWatchEventCache)
+                .close(() -> {
+                    remoteTimelockServiceAdapter.close();
+                    lockWatchManager.close();
```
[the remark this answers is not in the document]
jkozlowski: I thought about that, but not many places in the codebase really do this very well, and in general this is a little bit all over the place. e.g. see TransactionManagers#withRefreshingLockService does the exact same thing.

I fixed it now, but I think this whole class needs a bit of a cleanup wrt. resource handling potentially.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void close() {
+        refreshTask.cancel(false);
+        executorService.shutdown();
+    }
+
+    @Override
+    public synchronized void registerWatches(Set<LockWatchReferences.LockWatchReference> newLockwatches) {
```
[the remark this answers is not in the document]
jkozlowski: Fixed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -0,0 +1,88 @@
...
+        lockWatchReferences.addAll(newLockwatches);
+    }
+
+    private synchronized void reregisterWatches() {
```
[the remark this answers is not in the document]
jkozlowski: ah yes, good catch. I didn't consider that this will block calls where we add it, while we're calling timelock.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -164,38 +161,35 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
...
 
         List<StartIdentifiedAtlasDbTransactionResponse> responses =
-                timelockService.startIdentifiedAtlasDbTransactionBatch(requests.size());
-        Preconditions.checkState(requests.size() == responses.size(), "Different number of responses and conditions");
+                timelockService.startIdentifiedAtlasDbTransactionBatch(conditions.size());
+        Preconditions.checkState(conditions.size() == responses.size(), "Different number of responses and conditions");
         try {
             long immutableTs = Collections.max(responses.stream()
                     .map(response -> response.immutableTimestamp().getImmutableTimestamp())
                     .collect(Collectors.toList()));
```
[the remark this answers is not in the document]
jkozlowski: Fixed

lock-api-objects/src/main/java/com/palantir/lock/watch/TransactionUpdate.java
```diff
@@ -14,11 +14,15 @@
...
+package com.palantir.lock.watch;
 
 import org.immutables.value.Value;
 
+import com.palantir.lock.v2.LockToken;
+
 @Value.Immutable
-public interface StartTransactionRequest {
-    PreCommitCondition preCommitCondition();
+public interface TransactionUpdate {
```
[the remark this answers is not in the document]
jkozlowski: Really, this is just an internal type for the cache to hold onto (and Jolyon might adjust to fit whatever he needs). But I'm happy to add the docs in the meantime.

lock-api/src/main/java/com/palantir/lock/client/CommitTimestampGetter.java
jkozlowski: Yeah, this stitching is annoying, but I believe it's correct.

lock-api/src/main/java/com/palantir/lock/client/NamespacedConjureTimelockService.java
```diff
@@ -49,6 +51,10 @@ public ConjureGetFreshTimestampsResponse getFreshTimestamps(ConjureGetFreshTimes
         return conjureTimelockService.getFreshTimestamps(AUTH_HEADER, namespace, request);
     }
 
+    public GetCommitTimestampsResponse getCommitTimestamps(GetCommitTimestampsRequest request) {
+        return conjureTimelockService.getCommitTimestamps(AUTH_HEADER, namespace, request);
+    }
```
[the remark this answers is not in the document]
jkozlowski: Found it! good catch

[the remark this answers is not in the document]
jkozlowski: I got the version as earliest that had this commit: https://github.com/palantir/atlasdb/commit/69610e0ccb808a0c19706071f002a09098b3d4d4

PR #4812, opened 2020-05-28
[Dialogue] Part 11: Lock V1 Client Uses >99% Pure Dialogue

jeremyk-91:
Note to the reviewer: This PR unfortunately depends on both Part 8 and Part 9 which were separate branches off develop. If you're @gmaretic, though, you've already reviewed both parts so just ignore the Part 9 bits 😄 

**Goals (and why)**:
- V1 lock is still about 20% of RPCs AtlasDB makes to TimeLock, >99% is 3 endpoints
- V1 lock has an expansive API, I don't want to provide bridges for ALL endpoints.
- Previous parts implement timelock side support, this adds client support.

- Client side impl of pure Dialogue for the relevant lock endpoints.

- I changed the ETE resource so it uses the pure Dialogue endpoints to talk to the lock service.

- The zipWith API is maybe too clever?

**Where should we start reviewing?**: AtlasDbDialogueServiceProvider.java


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/timelock/ShortAndLongTimeoutServicesTest.java
```diff
@@ -57,4 +58,38 @@ public void mapOperatesOnCorrespondingElements() {
         assertThat(strings.shortTimeout()).isEqualTo("1");
         assertThat(strings.longTimeout()).isEqualTo("3");
     }
+
+    @Test
+    public void zipWithZipsCorrectElements() {
```
gmaretic, line 63: 🎉

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/lock/SimpleLockResource.java
```diff
@@ -60,9 +61,9 @@ public boolean lockUsingLegacyLockApi(int numLocks, int descriptorSize) {
                 .builder(generateDescriptorMap(numLocks, descriptorSize))
                 .build();
         try {
-            LockRefreshToken response = transactionManager.getLockService().lock("test", request);
+            HeldLocksToken response = transactionManager.getLockService().lockAndGetHeldLocks("test", request);
```
gmaretic, line 64: This is a behaviour change, but it should be fine for existing tests. I guess this was done to reuse existing tests to test the pure dialogue methods?

jeremyk-91: Yep, this now goes through the Dialogue methods as you say. This is only in the ETE server, as well

PR #4817, opened 2020-06-02
Fix incorrect getRangeSlices trace span names

jeremyk-91:
**Goals (and why)**:
- Fix a misleading trace we ran into (together with @azarum) when debugging a perf issue

- Correctly assign arguments: the count on the range is a batch hint, not number of columns; the count on the slice predicate is the number of columns, not number of rows.

- Is this the right way round?

**Where should we start reviewing?**: Small PR

**Priority (whenever / two weeks / yesterday)**: this week? It's small


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
```diff
@@ -99,13 +99,16 @@ public CassandraClient delegate() {
             KeyRange range,
             ConsistencyLevel consistency_level)
             throws InvalidRequestException, UnavailableException, TimedOutException, TException {
-        int numberOfKeys = predicate.slice_range.count;
-        int numberOfColumns = range.count;
+        int numberOfColumns = predicate.slice_range.count;
+        int batchHint = range.count;
```
sudiksha27, line 104: Nice!

PR #4819, opened 2020-06-02
[WIP] Remove LockWatchStateUpdate#Failed.

jkozlowski:
**Goals (and why)**:

This is a naive fix for this issue, that just uses already available primitives to synchronize snapshot computation, by prohibiting adding new events. I am starting this PR to start a conversation about the implementation here though.

From my read of the code, it is the way it is to support a **mostly** non-blocking implementation of the whole lock watch event log in timelock.

I have reviewed all the code and I am quite unconvinced this is entirely worth it. We have made the adding of events blocking already (i.e. there is a synchronized block for a bit of fast code that increments a pointer and writes to a buffer in LockEventLogImpl, so in that sense all lock/unlock calls will already be ordered wrt. to each other, therefore we likely already take some sort of performance hit on descriptors matching watches). But the machinery to support not-blocking lock/unlock codepaths for too long by anyone potentially asking for a snapshot or adding a lockwatchcreated event is in my opinion just not worth the complexity.

We also made getting timestamps (as part of startTransactions call) block writes to the LockEventLogImpl.

My argument is that:

**In steady state (i.e. client is not behind)**: for normal startTransactions calls adding just a bit more work under lock to copy a few pointers to immutable events from a buffer (effectively a System.arrayCopy of a generally small amount of events) is not be very costly.

Then: 

Clients needing a snapshot computed will occur:
1. Client is genuinely behind. *sigh* but should really only be happening when a new client comes up, so relatively infrequently (and since the log is per namespace, it generally affects one client, except for locking up a few timelock request threads for slightly longer).
2. We just had a leader election: this is slightly more annoying since everyone will need a snapshot, but good news is that the log and held locks will generally be empty at this point!

### DIFF
--- a/lock-api-objects/src/main/java/com/palantir/lock/watch/LockWatchStateUpdate.java
+++ b/lock-api-objects/src/main/java/com/palantir/lock/watch/LockWatchStateUpdate.java
@@ -32,16 +32,11 @@ import com.palantir.lock.watch.LockWatchReferences.LockWatchReference;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonSubTypes({
-        @JsonSubTypes.Type(value = LockWatchStateUpdate.Failed.class, name = LockWatchStateUpdate.Failed.TYPE),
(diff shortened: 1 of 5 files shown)


jkozlowski:
Closing in favour of https://github.com/palantir/atlasdb/pull/4820/files

PR #4820, opened 2020-06-02
Remove LockWatchStateUpdate#Failed.

jkozlowski:
**Goals (and why)**:

Try to simplify the code by removing a response that we should be able to not have.

If the client is too far behind, calculate a snapshot. Operations that touch the log, always lock it exclusively (whether for reading or writing), therefore we can always calculate a cohesive snapshot: all mutations will wait while we are calculating the snapshot, therefore the client will then later be able to get all the events. They might get the same "event" multiple times (because HeldLocks iteration is not blocking), but the ordering of these events will always be causal (e.g. unlock always comes after lock), but they should be able to deal with that

Existing unit tests

Whether there is any problems with this approach.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockEventLogImpl.java
```diff
@@ -50,58 +50,67 @@
...
 
-    private SnapshotEventReplayer getSnapshotEstimateAndCreateReplayer() {
+    private LockWatchStateUpdate calculateSnapshot() {
+        long lastVersion = slidingWindow.lastVersion();
         LockWatches currentWatches = watchesSupplier.get();
-        Set<LockWatchReference> watched = new HashSet<>(currentWatches.references());
-        Set<LockDescriptor> estimatedLocks = calculateOpenLocks(currentWatches.ranges());
-        return new SnapshotEventReplayer(watched, estimatedLocks);
+        Set<LockWatchReference> watches = new HashSet<>(currentWatches.references());
+        Set<LockDescriptor> openLocks = calculateOpenLocks(currentWatches.ranges());
```
jkozlowski, line 103: The thing to understand here is what does this snapshot actually mean: it's a best effort thing to indicate if there are any open locks and return them. This snapshot isn't really a *consistent* snapshot: while we're locked here in the log, the HeldLocksCollection is still *granting* locks, which, as soon as we unlock the log, will add their events. But as per the RFC, this doesn't matter for any transactions that end up working with this snapshot: what matters for conflict checking is that there were no events for cached rows in between the transaction starting and trying to commit.

After these changes (and possibly before them), if we compute a snapshot, the node might later receive events it already knows about from the snapshot again. And that should be fine.

jkozlowski:
To be fair, I think I might do another PR where we'll actually keep the still open locks in the log, instead of a fixed size. Then snapshot calculation is just a matter of copying the whole log. They are already in memory, so shouldn't be a big deal.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -44,92 +38,43 @@ long lastVersion() {
...
     }
 
     private int incrementAndMod(int num) {
         num++;
         return num >= maxSize ? num % maxSize : num;
     }
 
-    private boolean versionInTheFuture(long lastVersion, long lastWrittenSequence) {
-        return lastVersion > lastWrittenSequence;
+    private boolean validateVersion(long version) {
```
jeremyk-91, line 69: This returns true if the version is unloadable, which I find strange. Maybe either flip this, or `isVersionUnknown` or something along those lines.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -44,92 +38,43 @@ long lastVersion() {
...
     }
 
-    synchronized void finalizeAndAddSnapshot(long startVersion, LockWatchCreatedEventReplayer eventReplayer) {
-        Optional<List<LockWatchEvent>> remaining = getFromVersion(startVersion);
-        if (remaining.isPresent()) {
-            remaining.get().forEach(eventReplayer::replay);
-            add(LockWatchCreatedEvent.builder(eventReplayer.getReferences(), eventReplayer.getLockedDescriptors()));
-        }
+    boolean hasNextEvents(long version) {
+        return !validateVersion(version);
```
jeremyk-91, line 48: sanity check: Is hasNextEvents() queried _at_ the last version intentionally true?

non-actionable: this is basically `hasEventsFromHereToLastKnown` though that is kind of verbose so just `hasNextEvents` is fine.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
```diff
@@ -18,22 +18,16 @@
...
 
 import com.google.common.math.LongMath;
 import com.google.common.primitives.Ints;
-import com.palantir.atlasdb.timelock.lock.watch.LockEventLogImpl.LockWatchCreatedEventReplayer;
-import com.palantir.lock.watch.LockWatchCreatedEvent;
 import com.palantir.lock.watch.LockWatchEvent;
+import com.palantir.logsafe.Preconditions;
 
-@ThreadSafe
 public class ArrayLockEventSlidingWindow {
```
jeremyk-91, line 27: nit: Might be worth calling out `@NotThreadSafe` (it's really obvious, but worth noting)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
jkozlowski: Yes, if you present the last version, that means you're up to date, and we're return you last version and empty list of events. This method is used to calculate if the query to get the client up to date can be satisfied from the log, or else we need to calculate a snapshot.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
jkozlowski: Done

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/ArrayLockEventSlidingWindow.java
jkozlowski: Yeah, I tried not to rewrite this because that's how Grgur wrote these. But you're right, this reads weird.

jkozlowski: Actually, correction: I rewrote this to be weird, so just gonna fix that.

jkozlowski: Done

PR #4821, opened 2020-06-03
[GCV Blocker] Don't use org.json.simple

jeremyk-91:
**Goals (and why)**:
- Unblock #4769, gradle consistent versions is good.
- This PR is nontrivial enough that I feel it needs a review nonetheless.

- Swap out the implementation of the JSON parser library in `NameMetadataDescription`, from json.simple to Jackson (which we use everywhere anyway)
- Validate that the behaviour of parsing is unchanged.
- Note that while it might not look like this is used anywhere, actually *it is* internally!

- I wrote a heavy set of tests, mainly because I need to understand the nuances of this codepath myself.

There are some minor behaviour breaks:
- Passing in JSON of the wrong type (e.g. `"bleh"`) used to give you a `ClassCastException`, while now it gives you an `IllegalStateException`. I think this is fine.
- Causes of PalantirRuntimeExceptions owing to parser failures are going to be different, especially for the 'gobbledygook' case.

Is there anything else I missed out on?

**Where should we start reviewing?**: NameMetadataDescriptionTest

@j-baker for SA

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
@@ -20,10 +20,10 @@ import java.util.List;
 
 import javax.annotation.concurrent.Immutable;
 
-import org.json.simple.JSONObject;
-import org.json.simple.parser.JSONParser;
-import org.json.simple.parser.ParseException;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
@@ -40,6 +40,8 @@ import com.palantir.util.Pair;
 @Immutable
 public class NameMetadataDescription {
     public static final String HASH_ROW_COMPONENT_NAME = "hashOfRowComponents";
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private final List<NameComponentDescription> rowParts;
     private final int numberOfComponentsHashed;
@@ -183,8 +185,13 @@ public class NameMetadataDescription {
 
     public byte[] parseFromJson(String json, boolean allowPrefix) {
         try {
-            JSONObject obj = (JSONObject) new JSONParser().parse(json);
-            int numDefinedFields = countNumDefinedFields(obj);
+            JsonNode jsonNode = OBJECT_MAPPER.readTree(json);
+            Preconditions.checkState(jsonNode.isObject(),
+                    "Only JSON objects can be deserialized into parsed byte arrays.  Passed json was: %s",
+                    json);
+            ObjectNode objectNode = (ObjectNode) jsonNode;
+
+            int numDefinedFields = countNumDefinedFields(objectNode);
             byte[][] bytes = new byte[numDefinedFields][];
 
             Preconditions.checkArgument(numDefinedFields > 0,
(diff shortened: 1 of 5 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
```diff
@@ -183,8 +185,13 @@ public String renderToJson(byte[] name) {
 
     public byte[] parseFromJson(String json, boolean allowPrefix) {
         try {
-            JSONObject obj = (JSONObject) new JSONParser().parse(json);
-            int numDefinedFields = countNumDefinedFields(obj);
+            JsonNode jsonNode = OBJECT_MAPPER.readTree(json);
+            Preconditions.checkState(jsonNode.isObject(),
+                    "Only JSON objects can be deserialized into parsed byte arrays.  Passed json was: %s",
```
Jolyon-S, line 190: nit: I think there were two spaces after the full stop

```suggestion
                    "Only JSON objects can be deserialized into parsed byte arrays. Passed json was: %s",
```

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ValueType.java
```diff
@@ -464,9 +463,7 @@ public String convertToJava(byte[] value, int offset) {
 
         @Override
         public byte[] convertFromJson(String jsonValue) {
-            Object s = JSONValue.parse(jsonValue);
-            Preconditions.checkArgument(s instanceof String, "%s must be a json string", jsonValue);
-            return convertFromString((String) s);
+            return convertFromString(ValueType.readJson(jsonValue, String.class));
```
Jolyon-S, line 466: To make sure I understand: before, you would get `SafeIllegalArgumentException`s if this was not a string, but now you just get a plain `RuntimeException` that wraps an `IOException`. Is my understanding correct, and is this behaviour OK?

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/NameMetadataDescriptionTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    // TODO (jkong): This looks ridiculous, but I don't wish to change existing behaviour.
+    public void duplicateFieldsAreTolerated() {
+        String invalidJson = "{\"string\": \"tom\", \"string\": \"robert\"}";
+        byte[] result = SIMPLE_NAME_METADATA_DESCRIPTION.parseFromJson(invalidJson, false);
+        assertThat(result).satisfiesAnyOf(
+                (bytes) -> assertThat(bytes).containsExactly(PtBytes.toBytes("tom")),
+                (bytes) -> assertThat(bytes).containsExactly(PtBytes.toBytes("robert")));
```
Jolyon-S, line 78: I can't quite understand this test - it implies to me that you could get either result non-deterministically. Is this really the case currently?

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/NameMetadataDescription.java
jeremyk-91: I wouldn't normally write this, but this is intentional given the other error messages also have this.

Jolyon-S: Ah well, whatever.

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ValueType.java
jeremyk-91: That's mostly correct (strictly speaking: it used Google, not Palantir Preconditions, so you'd just get a regular `IllegalArgumentException`). That's a good call: I _think_ that is fine, but I'll change it back.

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/NameMetadataDescriptionTest.java
jeremyk-91: Currently you'll always get `robert` (in general, the last key-value mapping). This is dependent on the JSON library, though - while we could write just that assertion I don't think it's something we can guarantee.

I think my comment was probably a bit misleading in that "don't wish to change" referred to "tolerate duplicate fields", not specifically which key gets returned.

Jolyon-S: I think the comments you have provided make this a lot clearer. Thanks!

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/NameMetadataDescriptionTest.java
```diff
@@ -0,0 +1,167 @@
...
+        assertThat(bytes).containsExactly(SAMPLE_ROW_PREFIX);
+    }
+
+    @Test
+    public void missingNonSuffixFieldsAreNotToleratedInPrefixMode() throws JsonProcessingException {
+        JsonNode jsonNode = OBJECT_MAPPER.readTree(MULTIPART_NAME_METADATA_DESCRIPTION.renderToJson(SAMPLE_ROW));
+        ((ObjectNode) jsonNode).remove(ImmutableList.of("alpha", "omega"));
+
+        assertThatThrownBy(() -> MULTIPART_NAME_METADATA_DESCRIPTION.parseFromJson(jsonNode.toString(), false))
+                .isInstanceOf(IllegalArgumentException.class)
```
sudiksha27, line 151: I think the flag should be true for it to be in prefix mode?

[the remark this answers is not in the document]
jeremyk-91: Well spotted! I'll fix this

PR #4826, opened 2020-06-09
Override blocking thrift threadpool.

jkozlowski:
**Goals (and why)**:

Allow overriding blocking Cassandra threadpool. atlas-proxy ends up creating a lot of TransactionManagers and these threads add up to thousands of threads.

The large number of threads along with the fact that they are configured to terminate after 1 minute of inactivity, leads to a lot of thread churn. Internally we have discovered either a JVM or Kernel bug which non-deterministically leads to slowdowns around the thread creation, slowing the service down to a crawl.

Optional ExecutorService is passed to CassandraKeyValueServiceImpl. I looked at whether this could be an Executor, but, rightly, under normal operation CassandraKeyValueServiceImpl should be responsible for closing the threadpool, so it needs those methods from ExecutorService.

I have tested this on the affected stacks, and run into CassandraReloadableKvsConfig gotcha, where my config was not being applied.

Now

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
@@ -19,6 +19,7 @@ import java.net.InetSocketAddress;
 import java.net.SocketTimeoutException;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutorService;
 
 import org.immutables.value.Value;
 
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -190,6 +191,12 @@ default CassandraAsyncKeyValueServiceFactory asyncKeyValueServiceFactory() {
         return DefaultCassandraAsyncKeyValueServiceFactory.DEFAULT;
     }
 
+    @Value.Default
+    @JsonIgnore
+    default Optional<ExecutorService> blockingExecutor() {
+        return Optional.empty();
```
jkozlowski, line 197: This is a glorious gotcha.

PR #4806, opened 2020-05-27
[LW] Client side event cache

Jolyon-S:
~This PR depends on https://github.com/palantir/atlasdb/pull/4811 and should not be merged before.~

**Goals (and why)**:
Provide a means for the client to get a mapping from start timestamp to version for a batch of transactions. Contains all the underlying implementation associated with this - client side 
mapping from timestamp to known version for that timestamp, client side mapping from version to event that occurred at that version, and client side snapshot updater.

**Implementation Description (bullets)**:
* Provided a concrete implementation for `LockWatchEventCache` interface; this uses two main data structures: a `TreeMap` to store mapping from start timestamp to version, commit timestamp, commit-time version, and commit lock token (the last three provided when the commit update is processed), and a `TreeMultimap` which contains a mapping from version to all timestamps with that version
* The multimap allows us to easily keep track of the oldest still in-flight version; this is passed down to the `ClientLockWatchEventLog` to tell it to age out old entries (which are used to update the snapshot).
* `ClientLockWatchEventLog` contains the mapping from version to event, using a `TreeMap` to allow for easy range lookups. Old entries are passed to the snapshotter to update, then are deleted.
* `TransactionsLockWatchEvents` has been reworked - no longer requires a visitor, as a snapshot just sets the first element to a `LockWatchCreatedEvent` which is sufficient to re-seed the cache on the atlasdb-proxy side.

**Testing (What was existing testing like?  What have you done to improve it?)**:
Unit tests added for each concrete implementation.

**Concerns (what feedback would you like?)**:
* Do we leave ourselves open to any race conditions? Everything is synchronised, so I do not think so.
* Are there any ways that we can have a memory leak?
* Does this expose the necessary operations that we will need? Are there methods that are missing?
* Will this be performant enough?
* `LockWatchEventCacheImpl` is pretty beefy; any thoughts as to how or if we should reduce its size?

**Where should we start reviewing?**:
* `LockWatchEventCacheImpl`
* `ClientLockWatchEventLogImpl`
* `ClientLockWatchSnapshotUpdaterImpl`

**Priority (whenever / two weeks / yesterday)**:
This week (beginning 01/06).

PR #4827, opened 2020-06-09
[LW] Change TransactionLockWatchEvents interface

Jolyon-S:
**Goals (and why)**:
* This change is introduced in https://github.com/palantir/atlasdb/pull/4806, but that is taking a long time to go through and changes in atlasdb-proxy depends on this.

* Removed visitor pattern from `TransactionLockWatchEvents`; the intent is to always provide an update, even on snapshot.

N/A

* Is this considered a breaking change? Do we need to do a `revapi` anywhere as a result of this change?

* `TransactionLockWatchEvents`
* `NoOpLockWatchEventCache`

Soon

### DIFF
--- a/lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
+++ b/lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
@@ -20,14 +20,11 @@ import java.util.Collection;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
-
-import com.google.common.collect.ImmutableSet;
+import java.util.stream.Collectors;
 
 @SuppressWarnings("FinalClass") // mocks
 public class NoOpLockWatchEventCache implements LockWatchEventCache {
     public static final LockWatchEventCache INSTANCE = new NoOpLockWatchEventCache();
-    private static final TransactionsLockWatchEvents NONE = TransactionsLockWatchEvents.failure(
-            LockWatchStateUpdate.snapshot(UUID.randomUUID(), -1L, ImmutableSet.of(), ImmutableSet.of()));
 
     private NoOpLockWatchEventCache() {
         // singleton
@@ -55,10 +52,19 @@ public class NoOpLockWatchEventCache implements LockWatchEventCache {
     @Override
     public TransactionsLockWatchEvents getEventsForTransactions(Set<Long> startTimestamps,
             Optional<IdentifiedVersion> version) {
-        return NONE;
+        IdentifiedVersion fakeVersion = generateFakeVersion();
+        return ImmutableTransactionsLockWatchEvents.builder()
+                .clearCache(true)
+                .startTsToSequence(
+                        startTimestamps.stream().collect(Collectors.toMap(startTs -> startTs, $ -> fakeVersion)))
+                .build();
     }
 
     @Override
     public void removeTransactionStateFromCache(long startTimestamp) {
+    }
+
+    private IdentifiedVersion generateFakeVersion() {
+        return IdentifiedVersion.of(UUID.randomUUID(), -1L);
     }
 }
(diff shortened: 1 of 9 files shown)


lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
```diff
@@ -55,10 +52,19 @@ public CommitUpdate getCommitUpdate(long startTs) {
     @Override
     public TransactionsLockWatchEvents getEventsForTransactions(Set<Long> startTimestamps,
             Optional<IdentifiedVersion> version) {
-        return NONE;
+        IdentifiedVersion fakeVersion = generateFakeVersion();
```
jeremyk-91, line 55: This now needs to generate a new random UUID on each call to `getEventsForTransactions` - is that intentional? I thought `-1` as a sequence number was enough to indicate that the Events value was meaningless.

lock-api-objects/src/main/java/com/palantir/lock/watch/TransactionsLockWatchEvents.java
```diff
@@ -25,53 +25,9 @@
...
-        LockWatchStateUpdate.Snapshot snapshot();
-
-        @Override
-        default <T> T accept(Visitor<T> visitor) {
-            return visitor.visit(this);
-        }
-    }
+    List<LockWatchEvent> events();
+    Map<Long, IdentifiedVersion> startTsToSequence();
+    boolean clearCache();
```
jeremyk-91, line 32: I see why we'd want a boolean here, though this class feels like it should be renamed - it's quite a bit more than just the events now. Maybe `TransactionsLockWatchUpdate`?

lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
```diff
@@ -55,10 +52,19 @@ public CommitUpdate getCommitUpdate(long startTs) {
...
+                .startTsToSequence(
+                        startTimestamps.stream().collect(Collectors.toMap(startTs -> startTs, $ -> fakeVersion)))
+                .build();
     }
 
     @Override
     public void removeTransactionStateFromCache(long startTimestamp) {
     }
+
+    private IdentifiedVersion generateFakeVersion() {
```
jeremyk-91, line 67: nit: If we want to keep this, this should be static. Though I think having a constant for the UUID would be preferable

lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
Jolyon-S: This was a suggestion from the other PR by Jakub (ensuring that it 100% is treated as a snapshot each time) but in honesty this is unnecessary - things go off the snapshot flag anyway.

lock-api-objects/src/main/java/com/palantir/lock/watch/TransactionsLockWatchEvents.java
Jolyon-S: I think that is reasonable

lock-api-objects/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
Jolyon-S: As per the other, I agree.

lock-api-objects/src/main/java/com/palantir/lock/watch/TransactionsLockWatchEvents.java
Jolyon-S: Is that renaming considered a breaking change?

jeremyk-91: Technically yes (though this is internal API so I wouldn't be too concerned).

PR #4831, opened 2020-06-10
[PDS-119588] Relax Cassandra Schema Version Check

jeremyk-91:
**Goals (and why)**:
- Give Atlas a way of being resilient to migrations that change the cluster size. This blocked startup in PDS-119588.

- After discussion with @leonz and @tpetracca, it appears that the check was written in this way because there was a concern over performance (if more schema operations are kicked off while the cluster is not in agreement, this could make reaching said agreement take longer).
- Recent changes to the fork of Cassandra we use should have addressed these concerns.
- Separately: I audited all other uses of the server config, which appear to be either delegating, part of the startup process, or are seeding the Datastax driver (which has its own discovery mechanism).

- Updated unit tests to reflect the new behaviour.

- Do the tests provide good enough coverage?
- I believe this is safe for the gated describe keyspace call as well, because concurrent schema changes may or may not be reflected. Does this make sense?

**Where should we start reviewing?**: CKVSConsensusTest

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
@@ -48,6 +48,13 @@ public interface CassandraKeyValueServiceConfig extends KeyValueServiceConfig {
 
     String TYPE = "cassandra";
 
+    /**
+     * These are only the initial 'contact points' that will be used in connecting with the cluster. AtlasDB will
+     * subsequently discover additional hosts in the cluster. (This is true for both Thrift and CQL endpoints.)
+     *
+     * This value, or values derived from it (e.g. the number of Thrift hosts) MUST NOT be used to ascertain the
+     * current live state of the cluster.
+     */
     CassandraServersConfigs.CassandraServersConfig servers();
 
     @Value.Default
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
@@ -52,8 +52,8 @@ public interface CassandraKeyValueServiceConfig extends KeyValueServiceConfig {
      * These are only the initial 'contact points' that will be used in connecting with the cluster. AtlasDB will
      * subsequently discover additional hosts in the cluster. (This is true for both Thrift and CQL endpoints.)
      *
-     * This value, or values derived from it (e.g. the number of Thrift hosts) MUST NOT be used to ascertain the
-     * current live state of the cluster.
+     * This value, or values derived from it (e.g. the number of Thrift hosts) must ONLY be used on KVS initialization
+     * to generate the initial connection(s) to the cluster, or as part of startup checks.
      */
     CassandraServersConfigs.CassandraServersConfig servers();
 
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
@@ -75,14 +75,14 @@ public final class CassandraKeyValueServices {
     /**
      * Attempt to wait until nodes' schema versions match.
      *
-     * @param config the KVS configuration.
+     * @param schemaMutationTimeMillis time to wait for nodes' schema versions to match.
      * @param client Cassandra client.
      * @param unsafeSchemaChangeDescription description of the schema change that was performed prior to this check.
      *
      * @throws IllegalStateException if we wait for more than schemaMutationTimeoutMillis specified in config.
      */
     static void waitForSchemaVersions(
-            CassandraKeyValueServiceConfig config,
+            int schemaMutationTimeMillis,
             CassandraClient client,
             String unsafeSchemaChangeDescription)
             throws TException {
@@ -97,7 +97,7 @@ public final class CassandraKeyValueServices {
                 return;
             }
             sleepTime = sleepAndGetNextBackoffTime(sleepTime);
-        } while (System.currentTimeMillis() < start + config.schemaMutationTimeoutMillis());
+        } while (System.currentTimeMillis() < start + schemaMutationTimeMillis);
 
         StringBuilder schemaVersions = new StringBuilder();
         for (Entry<String, List<String>> version : versions.entrySet()) {
@@ -131,9 +131,9 @@ public final class CassandraKeyValueServices {
             CassandraClient client,
             String unsafeSchemaChangeDescription)
             throws TException {
-        waitForSchemaVersions(config, client, "before " + unsafeSchemaChangeDescription);
+        waitForSchemaVersions(config.schemaMutationTimeoutMillis(), client, "before " + unsafeSchemaChangeDescription);
         task.run();
-        waitForSchemaVersions(config, client, "after " + unsafeSchemaChangeDescription);
+        waitForSchemaVersions(config.schemaMutationTimeoutMillis(), client, "after " + unsafeSchemaChangeDescription);
     }
 
     private static boolean majorityOfClusterHasConsistentSchemaVersionAndNoDivergentNodes(
@@ -183,7 +183,7 @@ public final class CassandraKeyValueServices {
             CassandraKeyValueServiceConfig config) {
         try {
             clientPool.run(client -> {
-                waitForSchemaVersions(config, client, " during an initialization check");
+                waitForSchemaVersions(config.schemaMutationTimeoutMillis(), client, " during an initialization check");
                 return null;
             });
         } catch (Exception e) {
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
@@ -91,7 +91,7 @@ public final class CassandraKeyValueServices {
             // This may only include some of the nodes if the coordinator hasn't shaken hands with someone; however,
             // this existed largely as a defense against performance issues with concurrent schema modifications.
             versions = client.describe_schema_versions();
-            if (!clusterKnownToHaveSchemaDivergence(versions)) {
+            if (reachablePartOfClusterHasConsistentSchemaVersions(versions)) {
                 return;
             }
             sleepTime = sleepAndGetNextBackoffTime(sleepTime);
@@ -136,8 +136,8 @@ public final class CassandraKeyValueServices {
         waitForSchemaVersions(config, client, "after " + unsafeSchemaChangeDescription);
     }
 
-    static boolean clusterKnownToHaveSchemaDivergence(Map<String, List<String>> versions) {
-        return getDistinctReachableSchemas(versions).size() > 1;
+    static boolean reachablePartOfClusterHasConsistentSchemaVersions(Map<String, List<String>> versions) {
+        return getDistinctReachableSchemas(versions).size() == 1;
     }
 
     private static List<String> getDistinctReachableSchemas(Map<String, List<String>> versions) {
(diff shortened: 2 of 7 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -137,18 +136,8 @@ static void runWithWaitingForSchemas(
...
-        if (reachableSchemas.size() > 1) {
-            return false;
-        }
-
-        int numberOfServers = config.servers().numberOfThriftHosts();
-        int numberOfVisibleNodes = getNumberOfReachableNodes(versions);
-
-        return numberOfVisibleNodes >= ((numberOfServers / 2) + 1);
+    static boolean reachablePartOfClusterHasConsistentSchemaVersions(Map<String, List<String>> versions) {
+        return getDistinctReachableSchemas(versions).size() == 1;
```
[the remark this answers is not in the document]
tpetracca: it probably doesn't matter what we return in that case? expectation being your schema mutation will fail because there is no online node for it to succeed against.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -48,6 +48,13 @@
 
     String TYPE = "cassandra";
 
+    /**
+     * These are only the initial 'contact points' that will be used in connecting with the cluster. AtlasDB will
+     * subsequently discover additional hosts in the cluster. (This is true for both Thrift and CQL endpoints.)
+     *
+     * This value, or values derived from it (e.g. the number of Thrift hosts) MUST NOT be used to ascertain the
```
tpetracca, line 55: maybe change the wording to "must only be used on KVS initialization to generate the initial connection(s) to the cluster"?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -88,19 +88,18 @@ static void waitForSchemaVersions(
         long sleepTime = INITIAL_SLEEP_TIME;
         Map<String, List<String>> versions;
         do {
-            // This only returns the schema versions of nodes that the client knows exist. In particular, if a node we
-            // shook hands with goes down, it will have schema version UNREACHABLE; however, if we never shook hands
-            // with a node, there will simply be no entry for it in the map. Hence the check for the number of nodes.
+            // This may only include some of the nodes if the coordinator hasn't shaken hands with someone; however,
+            // this existed largely as a defense against performance issues with concurrent schema modifications.
             versions = client.describe_schema_versions();
-            if (uniqueSchemaWithQuorumAgreementAndOtherNodesUnreachable(config, versions)) {
```
tpetracca, line 95: we should delete the `CassandraKeyValueServiceConfig config` from this method's signature and remove its usage across the entirety fo the method

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -88,19 +88,18 @@ static void waitForSchemaVersions(
         long sleepTime = INITIAL_SLEEP_TIME;
         Map<String, List<String>> versions;
         do {
-            // This only returns the schema versions of nodes that the client knows exist. In particular, if a node we
-            // shook hands with goes down, it will have schema version UNREACHABLE; however, if we never shook hands
-            // with a node, there will simply be no entry for it in the map. Hence the check for the number of nodes.
+            // This may only include some of the nodes if the coordinator hasn't shaken hands with someone; however,
```
tpetracca, line 91: As far as I know this state is impossible to be in provided you have a quorum nodes, and therefore I don't believe is a legitimate concern.

The primary goal of this method in my opinion (and we should directly document this in the javadoc is):
- backoff doing schema mutations if the cluster is known to be in disagreement currently
  - I think this is just sensible and a natural thing to do, but admit given recent (read: when we did all the rescue restore optimization work) schema optimization work in our fork that this is less of a concern
- don't do a schema mutation if it can potentially result in "split-brain" (from the schema's point of view) in the future
  - The code as written in this PR allows for the case that we 1 out of 3 nodes are online, we push a schema mutation to that node. That node goes down, two other nodes come online, a schema mutation is pushed to those two nodes (whether by us or by another atlasdb client). The original node comes online, the nodes are in disagreement. While cassandra in theory should handle this case fine, the more divergent the schemas get the harder it will be for it to rectify them. Additionally there is the nuance/complexity around each different version of the schema having matching new pieces (like same keyspace/table name), but with a different cf_id. While I believe we've resolved this via deterministic cf_ids its worth pointing out that for that case, we would treat the situation as an outage and require manual intervention. We can prevent all the risks of the "split-brain" case here, by ensuring we only apply changes when we believe a majority of nodes are going to receive. From a maintain "HA" perspective this should be fine: see my next bullet.
- schema mutations are HA
  - We want to make sure critical path operations in AtlasDb are HA. We believe modifying a schema to be one of those things and thus it should be HA. That being said the rest of the critical path operations in AtlasDb, even when HA, have limitations to their availability. The most obvious of which are basic reads/writes to cassandra: they require quorum. Given the way we deploy cassandra (network topology aware RF=3), at any given moment [in the best case] if we have more than a third of the nodes of the cluster offline we have lost availability and can no longer read/write. Because of this I do not view being able to do a schema mutation where less than 2/3 of the nodes of the cluster are online as necessary.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
tpetracca: Tf we haven't already I think we should audit use of this method across the codebase and ensure it aligns with the expectations of this javadoc. The method below we've altered in this PR, even after these changes actually calls this method when it should not (even if it is only for logging purposes, it is still incorrect).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
```diff
@@ -88,19 +88,18 @@ static void waitForSchemaVersions(
...
             sleepTime = sleepAndGetNextBackoffTime(sleepTime);
         } while (System.currentTimeMillis() < start + config.schemaMutationTimeoutMillis());
 
         StringBuilder schemaVersions = new StringBuilder();
         for (Entry<String, List<String>> version : versions.entrySet()) {
-            schemaVersions = addNodeInformation(schemaVersions,
+            addNodeInformation(schemaVersions,
                     String.format("%nAt schema version %s:", version.getKey()),
                     version.getValue());
         }
```
tpetracca, line 105: all of the error message stuff below should talk about the cluster as being in terms of node returned in the describe_schema_versions() call above (regardless of if they're unreachable or not). none of it should use `config.servers()` as that isn't correct and also just isn't even how the checks of the rest of the method are done anymore.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
tpetracca: Given my statements above, even if there was a fear that describe_schema_versions() might not return a node that is freshly bootstrapping/about-to-bootstrap (which we only have at most 1 of at a time), we could set the requirement to simple-majority i.e. `floor(N/2)+1` rather than quorum (which is actually hard to accurately calculate in terms of raw numbers but is a stricter requirement to meet than simple-majority).

jeremyk-91: Yep, I'll change this to require a simple-majority. I'll push that first, and then look at the docs/reorganisation of the code here.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServicesSchemaConsensusTest.java
```diff
@@ -85,16 +85,16 @@ public void waitThrowsForAllUnreachableSchemaVersion() throws TException {
...
-    public void waitThrowsForFewerThanQuorumOnSameVersion() throws TException {
+    public void waitSucceedsOnMinorityOnSameVersion() throws TException {
         when(client.describe_schema_versions()).thenReturn(ImmutableMap.of(VERSION_1, REST_OF_NODES));
-        assertWaitForSchemaVersionsThrowsAndContainsConfigNodesInformation();
+        assertWaitForSchemaVersionsDoesNotThrow();
     }
 
     @Test
-    public void waitThrowsForQuorumOfUnreachableNodes() throws TException {
+    public void waitSucceedsOnMinorityOnSameVersionAndRestUnreachable() throws TException {
```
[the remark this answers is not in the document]
jeremyk-91: (Note: Changed to fails now, going along with the discussion on the later comment.)

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServicesSchemaConsensusTest.java
```diff
@@ -85,16 +85,16 @@ public void waitThrowsForAllUnreachableSchemaVersion() throws TException {
     }
 
     @Test
-    public void waitThrowsForFewerThanQuorumOnSameVersion() throws TException {
+    public void waitSucceedsOnMinorityOnSameVersion() throws TException {
```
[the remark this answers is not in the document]
jeremyk-91: I think it was feared that bootstrapping might mean we don't know about some of the nodes in the cluster (if the coordinator for some reason doesn't know about these yet).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
jeremyk-91: Yep, changed. I did do this audit actually, though agree that the log message makes more sense when it actually uses the servers discovered in the cluster.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
jeremyk-91: Yep, now updated to refer to the output returned from the describe_schema_versions call

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
jeremyk-91: It's still used for the blocking interval, though I've changed that to passing in an int.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServices.java
jeremyk-91: (Added the docs now.)

changelog/@unreleased/pr-4831.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: AtlasDB is now more resilient to migrations involving changes in the
+    cluster size relative to the original install config. Previously, if a cluster
+    was downsized at runtime, schema operations would consistently fail until the
+    service was bounced.
```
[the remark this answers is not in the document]
jeremyk-91: Yep, fair enough - good call, the description should mention Cassandra somewhere :)

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServicesSchemaConsensusTest.java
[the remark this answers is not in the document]
jeremyk-91: I've renamed the test to `waitSucceedsWithClusterHavingDownsizedAtRuntime`, since now that's what the test should be testing in practice

PR #4833, opened 2020-06-12
Feedback 1

sudiksha27:
**Goals (and why)**:
Collect client feedback for timeLock adjudication.

Have a background task that collects and reports feedback every 30secs.

TimeLockFeedbackBackgroundTask.java

EOD


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -538,6 +545,19 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
+        String serviceName = "UNKNOWN";
+        if (config().namespace().isPresent()) {
+            serviceName = config().namespace().get();
+        } else if (config().timelock().isPresent()) {
+            TimeLockClientConfig timeLockClientConfig = config().timelock().get();
+            serviceName = timeLockClientConfig.getClientOrThrow();
+        } else if (config().keyValueService().namespace().isPresent()) {
+            serviceName = config().keyValueService().namespace().get();
+        }
+        return serviceName;
```
jeremyk-91, line 558: Consider using a stream-of-optionals impl? It's probably going to be neater (provided it works)

lock-api/src/main/java/com/palantir/lock/client/metrics/TimeLockFeedbackBackgroundTask.java
```diff
@@ -0,0 +1,57 @@
...
+        // no op
+    }
+
+    public static void create(TaggedMetricRegistry taggedMetricRegistry, Supplier<String> versionSupplier,
+            String serviceName) {
+        UUID nodeId = UUID.randomUUID();
+
+        ScheduledExecutorService scheduledExecutorService =
+                Executors.newScheduledThreadPool(5);
+        scheduledExecutorService.scheduleWithFixedDelay(() -> {
```
jeremyk-91, line 39: Fine for now. Watch out for exceptions though!

lock-api/src/main/java/com/palantir/lock/client/metrics/ClientFeedback.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableClientFeedback.class)
+@JsonDeserialize(as = ImmutableClientFeedback.class)
+public interface ClientFeedback {
```
jeremyk-91, line 29: This is OK, but I'd prefer we use the opportunity to define it as a Conjure object

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -440,6 +441,65 @@ public void keyValueServiceMetricsDoNotContainUserAgent() {
...
+        TransactionManagers transactionManagers = TransactionManagers.builder()
+                .config(atlasDbConfig)
+                .userAgent(USER_AGENT)
+                .globalMetricsRegistry(metrics)
+                .globalTaggedMetricRegistry(DefaultTaggedMetricRegistry.getDefault())
+                .registrar(environment)
+                .build();
+
+        assertThat(transactionManagers.getServiceName()).isEqualTo("namespace");
+    }
```
jeremyk-91, line 483: We might want a timelock one for completeness?

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
```diff
@@ -76,6 +76,7 @@ static TransactionManagerContext createTransactionManager(
                                 .servers(serverUris)
                                 .sslConfiguration(SslConfiguration.of(Paths.get("var/security/trustStore.jks")))
                                 .build())
+                        .client(Optional.of(agent))
```
jeremyk-91, line 79: Did we need this? I think it would be a break if it turned out that we _must_ specify a client

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -359,6 +361,11 @@ public TransactionManager serializable() {
     private TransactionManager serializableInternal(@Output List<AutoCloseable> closeables) {
         MetricsManager metricsManager = MetricsManagers.of(globalMetricsRegistry(), globalTaggedMetricRegistry());
 
+        TimeLockFeedbackBackgroundTask.create(
```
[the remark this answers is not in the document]
jeremyk-91: (I don't think we fixed this yet - we've made the task _AutoCloseable_ but it is not actually closed when the TM shuts down!)

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TimeLockTestUtils.java
sudiksha27: So, the problem is highlighted by this line - `config().timelock().map(TimeLockClientConfig::getClientOrThrow)`. If timeLockConfig is present then client must be present otherwise, if we try to fetch the client, we throw this SafeIllegalStateEx - `"Tried to read a client from a TimeLockClientConfig, but it hadn't been initialised."`

jeremyk-91: Ah, this is a break then. Can you see a way around this?

lock-api/src/main/java/com/palantir/lock/client/metrics/Constants.java
```diff
@@ -0,0 +1,26 @@
...
+package com.palantir.lock.client.metrics;
+
+public class Constants {
+    private Constants () {
+        // no op
+    }
+
+    public static final String TIMELOCK_FEEDBACK_THREAD_PREFIX = "TimeLockFeedbackBackgroundTask";
+    public static final long TIMELOCK_CLIENT_FEEDBACK_INTERVAL_SECONDS = 30;
+}
```
sudiksha27, line 26: move to bckTask

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
sudiksha27: This is hard to write because the migrator tries to reach to timeLock servers.

PR #4835, opened 2020-06-15
PG-151833: Apply the default cancel-remaining-calls value globally

carterkozak:
**Goals (and why)**:

Prevent excessive handshaking due to lack of connection reuse when request are interrupted.

Use the same default across all paths

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
@@ -188,7 +188,7 @@ public final class Leaders {
                 createExecutorsForService(metricsManager, otherLeaders, "leader-ping"),
                 config.leaderPingResponseWait(),
                 leaderUuid,
-                true);
+                PaxosConstants.CANCEL_REMAINING_CALLS);
 
         LeaderElectionService uninstrumentedLeaderElectionService = new LeaderElectionServiceBuilder()
                 .leaderUuid(leaderUuid)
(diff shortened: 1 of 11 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.paxos;
+
+public final class PaxosConstants {
+
+    public static final boolean CANCEL_REMAINING_CALLS = false;
```
j-baker, line 20: if we're just makign this be a constant, please can we just remove the optionality?

carterkozak: This simply expands the existing constant. I agree that the optionality should be removed, but this is causing serious problems at the moment and must be addressed quickly.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/Leaders.java
```diff
@@ -175,7 +176,7 @@ public static LocalPaxosServices createInstrumentedLocalServices(
                 acceptors,
                 config.quorumSize(),
                 createExecutorsForService(metricsManager, acceptors, "latest-round-verifier"),
-                true);
+                PaxosConstants.CANCEL_REMAINING_CALLS);
```
Jolyon-S, line 179: It looks like you missed changing this for `leaderPinger` defined just below - was that intentional?

carterkozak: Nope, that should be changed as well, thanks!

PR #4837, opened 2020-06-15
[PD$-110002] Part 2: Economising Snapshot Transaction commitWrites() Timers

jeremyk-91:
**Goals (and why)**:
- Save money on metrics
- But still get as much signal as possible

- In internal atlasdb-proxy, these are a fairly heavily hit set of timers (one per constituent database). Some are useful, but not all.
- I think the least contentious ones are the in-memory operations (tryUnlock, asyncPuncher). There is a middling group that is a bit more open for discussion: I chose them based on how valuable the signal is compared to an already existing metric (e.g. putting the commit timestamp into the database is largely reflected in TransactionService.putUnlessExists and I can believe the latter is representative of the former; but I can't say the same for transaction locks).

- Consider each decision on whether the timing metric was kept or not - do you agree with each one?

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -1703,9 +1703,12 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 // We check the pre-commit conditions first since they may operate similarly to read write conflict
                 // handling - we should check lock validity last to ensure that sweep hasn't affected the checks.
                 timedAndTraced("userPreCommitCondition", () -> throwIfPreCommitConditionInvalid(commitTimestamp));
+
+                // Not timed, because this just calls ConjureTimelockServiceBlocking.refreshLockLeases, and that is
+                // timed.
                 traced("preCommitLockCheck", () -> throwIfImmutableTsOrCommitLocksExpired(commitLocksToken));
 
-                // TransactionService.putUnlessExists: actually commits the transaction
+                // Not timed, because this just calls TransactionService.putUnlessExists, and that is timed.
                 traced("commitPutCommitTs",
                         () -> putCommitTimestamp(commitTimestamp, commitLocksToken, transactionService));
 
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1700,28 +1703,33 @@ private void commitWrites(TransactionService transactionService) {
                 // We check the pre-commit conditions first since they may operate similarly to read write conflict
                 // handling - we should check lock validity last to ensure that sweep hasn't affected the checks.
                 timedAndTraced("userPreCommitCondition", () -> throwIfPreCommitConditionInvalid(commitTimestamp));
-                timedAndTraced("preCommitLockCheck", () -> throwIfImmutableTsOrCommitLocksExpired(commitLocksToken));
+                traced("preCommitLockCheck", () -> throwIfImmutableTsOrCommitLocksExpired(commitLocksToken));
 
-                timedAndTraced("commitPutCommitTs",
+                // TransactionService.putUnlessExists: actually commits the transaction
```
Jolyon-S, line 1708: I think this comment, in isolation, is not that useful: I assume it is explaining why this is not timed, but outside of the context of this PR it doesn't really imply. Perhaps:

```// Not timed because TransactionService.putUnlessExists actually commits the transaction, and is timed```

File: atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java [hunk 2]
```diff
@@ -1700,28 +1703,33 @@ private void commitWrites(TransactionService transactionService) {
                 // We check the pre-commit conditions first since they may operate similarly to read write conflict
                 // handling - we should check lock validity last to ensure that sweep hasn't affected the checks.
                 timedAndTraced("userPreCommitCondition", () -> throwIfPreCommitConditionInvalid(commitTimestamp));
-                timedAndTraced("preCommitLockCheck", () -> throwIfImmutableTsOrCommitLocksExpired(commitLocksToken));
+                traced("preCommitLockCheck", () -> throwIfImmutableTsOrCommitLocksExpired(commitLocksToken));
```
Jolyon-S, line 1706: You've mentioned:
"refreshing commit locks: consider ConjureTimelockServiceBlocking.refreshLockLeases"
in the changelog, but I don't see a comment here in the code (where you have for others that are no longer timed). Is it perhaps adding a comment?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: That's a fair point. Makes sense

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: Yep, I'll add one. I think I left this out as I thought it was obvious, but probably worth doing for the sake of completeness

PR #4840, opened 2020-06-16
override service name

sudiksha27:
**Goals (and why)**:
Override service name so services with multiple keyspaces can be accounted as one.  

Have serviceName property with default value.

Added a new test.

TransactionManagers.java 

Today

### DIFF
--- a/atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
+++ b/atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
@@ -442,6 +442,26 @@ public class TransactionManagersTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -441,6 +441,26 @@ public void keyValueServiceMetricsDoNotContainUserAgent() {
                 .anyMatch(metricName -> metricName.contains(USER_AGENT_NAME)), is(false));
     }
 
+    @Test
+    public void overridenServiceNameIsReturned() {
```
jeremyk-91, line 445: nit: `overridden` (2 ds)

PR #4843, opened 2020-06-17
Add serviceId

sudiksha27:
**Goals (and why)**:
The goad is to have override-able serviceIdentifier property in TransactionManagers.

Add optional serviceIdentifier property.

Modified existing test.

Today

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -552,12 +552,12 @@ public abstract class TransactionManagers {
         return transactionManager;
     }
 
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -552,16 +552,18 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
         return transactionManager;
     }
 
+    abstract Optional<String> serviceIdentifier();
```
jeremyk-91, line 555: nit: can we name this `serviceIdentifierOverride` or `explicitServiceIdentifier` or something like that?

PR #4846, opened 2020-06-17
[PD$-110002] CONTENTIOUS - Part 4: Redirect All Metrics To TaggedMetricRegistries

jeremyk-91:
**Reviewable, but DO NOT MERGE: I want to do a quick smoke-test internally.**

**Goals (and why)**:
- Collect AtlasDB's metrics neatly into a `TaggedMetricSet` so that filtering and pre-processing can be performed before they are released into the wild.

- Implement DisjointUnionTaggedMetricSet which combines metrics.
- Reorganise TransactionManagers to publish everything to the tagged metric registry.

**Testing (What was existing testing like?  What have you done to improve it?)**: DisjointUnionTaggedMetricSetTest. I will manually smoke-test with a service internally as well.

- Adding the origin tags is not really what I was going for (see https://github.com/palantir/metric-schema/pull/288), but this change, and even the combination of this change and the metric-schema change will not add metric cardinality (both tags are applied to all metric-schema metrics, and this tag is applied to all existing AtlasDB metrics). I opted to avoid conflict with metric-schema's intended tag. Is that correct?
- If users expect their metrics to be in specifically the untagged metric registry and try and read from it directly, that might cause pain. With that in mind, is this change too heavy-handed? Is that a reasonable use case?
- Should I just have added the Dropwizard metric set and the tagged metric set as _two_ metrics sets, maybe atlasdb-tagged and atlasdb-untagged? I don't know if internal metrics platforms are good at performing disjunctive aggregations across tag values though.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/DisjointUnionTaggedMetricSet.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/DisjointUnionTaggedMetricSet.java
@@ -1,0 +1,53 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.metrics;
+
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+import com.codahale.metrics.Metric;
+import com.google.common.collect.ImmutableMap;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricSet;
+
+/**
+ * Combines two {@link TaggedMetricSet}s. It is expected that the metric names present from the two sets are disjoint.
+ */
+public class DisjointUnionTaggedMetricSet implements TaggedMetricSet {
+    private final TaggedMetricSet first;
+    private final TaggedMetricSet second;
+
+    public DisjointUnionTaggedMetricSet(TaggedMetricSet first, TaggedMetricSet second) {
+        this.first = first;
+        this.second = second;
+    }
+
+    @Override
+    public Map<MetricName, Metric> getMetrics() {
+        return ImmutableMap.<MetricName, Metric>builder()
+                .putAll(first.getMetrics())
+                .putAll(second.getMetrics())
+                .build();
+    }
+
+    @Override
+    public void forEachMetric(BiConsumer<MetricName, Metric> consumer) {
+        first.getMetrics().forEach(consumer);
+        second.getMetrics().forEach(consumer);
+    }
+
+}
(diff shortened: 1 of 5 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/DisjointUnionTaggedMetricSet.java
```diff
@@ -0,0 +1,53 @@
...
+        return ImmutableMap.<MetricName, Metric>builder()
+                .putAll(first.getMetrics())
+                .putAll(second.getMetrics())
+                .build();
+    }
+
+    @Override
+    public void forEachMetric(BiConsumer<MetricName, Metric> consumer) {
+        first.getMetrics().forEach(consumer);
+        second.getMetrics().forEach(consumer);
```
iamdanfox, line 50: prefer `first.forEachMetric` as it's more performant

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
```diff
@@ -33,6 +33,9 @@ private CellFilterMetrics() {
         public static final String EMPTY_VALUE = "emptyValuesCellFilterCount";
     }
 
+    public static final String LIBRARY_ORIGIN_TAG = "libraryOrigin";
```
iamdanfox, line 36: If we're going to ask datadog to make a new report itemizing costs, we need to make sure we use a consistent tag across all libraries. In the metric-schema PR, we're intending to use `libraryName`, so I don't think it will help much if you pick a different one

atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/DisjointUnionTaggedMetricSet.java
jeremyk-91: Right, yeah that makes sense (I think I was trying to do something else)

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
jeremyk-91: My main worry here would be how metric registries handle duplicate tags - we are in the process of rolling out metric-schema as we touch new code. Do we know if that's okay? If we do then I'm happy to change this

jeremyk-91: A read of `ExtraEntrySortedMap` in tritium suggests this might become a problem if the underlying metrics already have a `libraryName`.

j-baker: Ok - so it was just flagged to me that this is non-ideal for our metrics concerns - for dan's original comment. @iamdanfox is there any chance you could work with jeremy to figure out an agreeable solution? I guess in true Highlander style There Can Only Be One which makes me hackily think of an extra layer of indirection (you have a libraryOrigin -> atlasdb tag, and then that wraps anotehr metric set which _eitehr_ adds library name atlasdb or if already set, lets it still be set.

iamdanfox: I think a reasonable solution here could be to just make ExtraEntrySortedMap not freak out if you're asking to add a `libraryName:atlasdb` but the map already contains `libraryName:atlasdb`. Collisions should only a problem with same key and different value, but we currently barf on _all_ collisions...

jeremyk-91: The intention I had when doing this was a bit different: when I wrote this I didn't really care about `libraryOrigin: atlasdb`, I just need a tag that won't be shared to be added here so that I can process the TaggedMetricSets nicely and then register them to the world outside AtlasDB.

The cost tracking was intended to be a separate thing. Can we be sure that the Atlas metric registries will never register any metrics that already have metricName? (e.g. dialogue, tritium?) - if so then I'm happy to change this to be libraryName.

PR #4847, opened 2020-06-18
[PDS$-110002] Part 5 | PoC?: Threshold-Based Metric Filtering and Metric-Schema for Targeted Sweep Metrics

jeremyk-91:
Warning: A bit on the heavier side, could possibly be split up.

**Goals (and why)**:
- Save $$$
- Allow metrics to be _filtered for publication_ based on a user-implemented filter that interprets their values (or, in practice, runs an arbitrary predicate)

- Implement FilteredTaggedMetricsSet that allows filtering based on user-registered filters
- Use the set in a MetricManager and allow users to register filters
- Implement a kill switch: runtime config that forces all filters to be ignored
- Rewire targeted sweep metrics to use metric-schema
- Implement a filter on targeted sweep metrics to publish if the number of reads/writes is >1000, or if the head of the queue is more than 6 hours old.

- Added some unit tests.
- I want to do a smoke test.

- Does it actually work?
- Does the wiring broadly make sense?
- Is the choice of constants for Targeted Sweep reasonable?
- I don't deregister metrics until the next bounce once they leave the thresholds because it probably won't save that much money, and it is probably valuable to see how a system recovers (see discussion in internal chat channel), does this make sense?

**Where should we start reviewing?**: FilteredTaggedMetricsSet


atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/FilteredTaggedMetricSet.java
```diff
@@ -0,0 +1,74 @@
...
+
+import com.codahale.metrics.Metric;
+import com.palantir.common.streams.KeyedStream;
+import com.palantir.refreshable.Refreshable;
+import com.palantir.tritium.metrics.registry.MetricName;
+import com.palantir.tritium.metrics.registry.TaggedMetricSet;
+
+public class FilteredTaggedMetricSet implements TaggedMetricSet {
+    private final TaggedMetricSet unfiltered;
+    private final Map<MetricName, List<MetricPublicationFilter>> singleMetricFilters;
```
jkozlowski, line 32: Would it make sense for this class to just take a Predicate<MetricName>, and implement the filtering rules somewhere else?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetricPublicationFilter.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        boolean conditionsAchieved = testConditions();
+        if (conditionsAchieved) {
+            publicationLatch.compareAndSet(false, true);
+        }
+        return conditionsAchieved;
+    }
+
+    private boolean testConditions() {
+        return enqueuedWrites.getAsLong() >= MINIMUM_READS_WRITES_TO_BE_CONSIDERED_ACTIVE
```
jkozlowski, line 68: I wonder if metric-schema should provide all of this for us? Basically the ability to specify that a metrics should be published only once a certain threshold is breached? I suspect we'll need this in a bunch of projects that are hoping to cut on spend. @ferozco

atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/FilteredTaggedMetricSet.java
jeremyk-91: As in this class just takes `Predicate<MetricName>` (and then have some kind of filter registry that handles this map)? Yep, we could - I don't feel strongly either way (what you propose does split responsibilities better, though it does add another piece).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetricPublicationFilter.java
jeremyk-91: Yep, it would be cool if metric-schema would do that (and subject to team prioritisation I would be very interested in helping with such an effort, if we think it should live there). 

Note that here we register these filters even for the other TS metrics so having 'publish A only if B is above a certain value' might be a little bit much to ask (though if we want to support that in metric schema I'm not complaining!).

atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/FilteredTaggedMetricSet.java
jeremyk-91: (for clarity: yeah I plan to change this)

jeremyk-91:
@sudiksha27 and I have gone through the first two bullet points in the impl.

Also I've smoke tested this on internal stack and we've verified that it works functionally (i.e. the targeted sweep stuff gets correctly filtered)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetricPublicationFilter.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * The criteria for publishing metrics for a given strategy, is if the number of enqueued writes or reads is in
+ * excess of {@link #MINIMUM_READS_WRITES_TO_BE_CONSIDERED_ACTIVE}, or it is believed that targeted sweep is behind by
+ * more than {@link #MINIMUM_STALE_DURATION}. If any of these conditions are true, then all targeted sweep metrics
+ * will be published.
+ */
+public class TargetedSweepMetricPublicationFilter implements MetricPublicationFilter {
+    @VisibleForTesting
+    static final long MINIMUM_READS_WRITES_TO_BE_CONSIDERED_ACTIVE = 1_000;
+    @VisibleForTesting
```
sudiksha27, line 39: Should we make this user-configurable?

jkozlowski: Dunno in this particular case, but in general, less config is always better. We should aim to build systems that either self-tune or have good enough defaults that work in most cases.

jeremyk-91: Good question. I think I'll hang on to this for now: the kill switch allows us to recover our metrics if there's a P0 (or even a P1), and otherwise I doubt there'll be a situation where we need to change these so urgently that it can't wait for a quick PR that adds configurability.

jeremyk-91: (In general I think I'd slant more towards leaving uncertain things configurable in AtlasDB, while less so in AtlasDB-Proxy mainly because of the relative difficulty/friction of recalling bad versions of AtlasDB.)

PR #4850, opened 2020-06-22
feedback analysis | poc 1

sudiksha27:
**Goals (and why)**:
TimeLock needs to accept and aggregate metrics from clients for self adjudication.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/Utils.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class Utils {
+    static <T> List<T> getMajority(Stream<T> streamOfStuff) {
```
jeremyk-91, line 27: nit: `getModalElements` or something? Majority would usually imply strictly more than half

PR #4853, opened 2020-06-22
Error rates

sudiksha27:
**Goals (and why)**:
Incorporate error in TimeLock feedback health report.

- Update metrics schema to record startTransaction and leaderTime error rates

- Update feedback background task in Atlas so it builds the health report with error rate

Added some new tests

DialogueAdaptingConjureTimelockService.java

EOD 23 June

### DIFF
--- a/timelock-api/src/main/conjure/timelock-feedback.yml
+++ b/timelock-api/src/main/conjure/timelock-feedback.yml
@@ -12,7 +12,8 @@ types:
       EndpointStatistics:
         fields:
           p99: double
-          oneMin: double
+          minRate: double
+          errorRate: double
 
 services:
   TimeLockClientFeedbackService:
--- a/timelock-api/src/main/conjure/timelock-feedback.yml
+++ b/timelock-api/src/main/conjure/timelock-feedback.yml
@@ -12,8 +12,8 @@ types:
       EndpointStatistics:
         fields:
           p99: double
-          minRate: double
-          errorRate: double
+          oneMin: double
+          errorRate: optional<double>
 
 services:
   TimeLockClientFeedbackService:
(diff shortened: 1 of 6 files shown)


timelock-api/src/main/conjure/timelock-feedback.yml
```diff
@@ -12,7 +12,8 @@ types:
       EndpointStatistics:
         fields:
           p99: double
-          oneMin: double
+          minRate: double
+          errorRate: double
```
jeremyk-91, line 16: Unfortunately, this breaks compatibility.
- For backward compatibility (old clients can talk to new servers), the minRate and errorRate need to be optional
- For forward compatibility (new clients can talk to old servers), that isn't readily possible in Conjure so you'll want to add a product dependency. Servers are required to reject unknown fields.

It's not ideal but I'd suggest leaving `oneMin` as it was named

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/FeedbackMetricsTest.java
```diff
@@ -0,0 +1,100 @@
...
+        assertThat(metrics.leaderTimeErrors().getCount()).isEqualTo(0);
+    }
+
+    @Test
+    public void leaderTimeErrorMetricsAreRecordedOnException() {
+        when(conjureTimelockServiceBlocking.leaderTime(AUTH_HEADER, NAMESPACE)).thenThrow(new RuntimeException());
+        try {
+            service.leaderTime(AUTH_HEADER, NAMESPACE);
+        } catch (RuntimeException e) {
+            // no op
```
jeremyk-91, line 69: nit - Missed this in round 1: I'd suggest verifying that the exception thrown by the timelock service _is_ the exception that's also thrown out of the Dialogue adapting service.

PR #4855, opened 2020-06-22
[Dialogue] Part 13: Bonfire of Configuration

jeremyk-91:
**Goals (and why)**:
- Dialogue is already forced everywhere, avoid giving the false impression that it is configurable.
- Avoid running an unnecessary test.
- Make steps toward removing AtlasDbHttpClients altogether. Still used in leader mode.

- Remove config option for enableDialogue
- Replace switching proxies with the pure Dialogue proxy
- Remove flavors of ETE tests that used Dialogue, back when it was not default.

- Not much additional here. Existing ETEs should fail if the simplifications were wrong.

- This removes the ability to prefer the CJR Dialogue shim over pure Dialogue for most TimeLock communication. Is that a bad idea? We have not configured enableDialogue to false _anywhere_.


jeremyk-91:
Probably saves us a minute on the build or so as well?

jkozlowski:
> Probably saves us a minute on the build or so as well?

yes please

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1083,46 +1078,23 @@ private static LockAndTimestampServices getLockAndTimestampServices(
         ServiceCreator creator = ServiceCreator.withPayloadLimiter(
```
jkozlowski, line 1078: Is this thing used still here?

jeremyk-91: Yeah, by `ConjureLockWatchingService`. I'll clear that out in a quick follow up to this one.

PR #4865, opened 2020-06-24
Name key count logged safe parameter

schlosna:
**Goals (and why)**:
Avoid incorrect duplicate `rowCount` safe logging parameter

Define new `keyCount` logging argument

Unit tests

Whenever


jeremyk-91:
(I edited the changelog slightly, we don't normally put all of the PR metadata in)

PR #4866, opened 2020-06-24
Adjust AssertUtils to be safelogging compatible

dxiao:
Format string and args will be passed to slf4j directly,
instead of doing our own manual formatting.

Helps with whitelisting logging on internal repo.


atlasdb-commons/src/main/java/com/palantir/util/AssertUtils.java
```diff
@@ -119,7 +120,11 @@ public static void assertAndLogWithException(boolean cheapTest, String msg, Thro
     public static void assertAndLogWithException(Logger log, boolean cheapTest, String format, Throwable t,
             Object... args) {
         if (!cheapTest) {
-            assertAndLogWithException(log, false, String.format(format, args), t);
+            Object[] newArgs = new Object[args.length + 1];
+            newArgs[0] = t;
```
carterkozak, line 124: Have you verified this works as expected? I think the throwable needs to be the last parameter.

atlasdb-commons/src/main/java/com/palantir/util/AssertUtils.java
```diff
@@ -119,7 +121,10 @@ public static void assertAndLogWithException(boolean cheapTest, String msg, Thro
     public static void assertAndLogWithException(Logger log, boolean cheapTest, String format, Throwable t,
             Object... args) {
         if (!cheapTest) {
-            assertAndLogWithException(log, false, String.format(format, args), t);
+            Object[] newArgs = Arrays.copyOf(args, args.length + 1);
+            newArgs[args.length] = t;
+            log.error("Assertion " + format + " with exception ", newArgs);
```
jeremyk-91, line 126: Unfortunately this generally isn't allowed in gradle-baseline. I'm not sure what a good solution here is: separately doing a string substitution, maybe?

PR #4867, opened 2020-06-25
[PD$-110002] Part 6: One-Shots and Legacies

jeremyk-91:
**Goals (and why)**:
- $$$

This PR contains four small metrics cuts.
- Only log the background sweep batch multiplier if <1. This should save |namespaces| series in atlasdb-proxy.
- Don't instrument `TimestampManagementService` endpoints (used only in timelock migration). This should save 10 * |namespaces| series in atlasdb-proxy.
- Don't instrument `CoordinationService` endpoints (used only in transactions2 migration). This should save 10 * |namespaces| series in atlasdb-proxy.
- Don't publish `timelockSuccessfulRequest` and `timelockFailedRequest` - I'm not aware of anyone who uses these outside of the health check that we have. This should save 4 * |namespaces| series in atlasdb-proxy.

(Note that meters create two series, and timers five.)

Nothing beyond existing tests.

- Are any of these metrics more valuable than what I give them credit for?

**Where should we start reviewing?**: Commit by commit?

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/AdjustableSweepBatchConfigSource.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/AdjustableSweepBatchConfigSource.java
@@ -24,10 +24,8 @@ import org.slf4j.LoggerFactory;
 
 import com.codahale.metrics.Gauge;
 import com.google.common.collect.ImmutableMap;
-import com.palantir.atlasdb.metrics.MetricPublicationFilter;
 import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.logsafe.SafeArg;
-import com.palantir.tritium.metrics.registry.MetricName;
 
 public final class AdjustableSweepBatchConfigSource {
     private static final Logger log = LoggerFactory.getLogger(BackgroundSweeperImpl.class);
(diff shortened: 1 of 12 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/InstrumentedTimelockService.java
```diff
@@ -31,16 +33,35 @@
...
+    }
+
+    public static TimelockService create(TimelockService timelockService, MetricsManager metricsManager) {
+        // The instrumentation here is used primarily for the health check, not for external viewing.
+        metricsManager.addMetricFilter(
+                MetricName.builder().safeName(AtlasDbMetricNames.TIMELOCK_SUCCESSFUL_REQUEST).build(),
+                () -> false);
+        metricsManager.addMetricFilter(
+                MetricName.builder().safeName(AtlasDbMetricNames.TIMELOCK_FAILED_REQUEST).build(),
+                () -> false);
```
sudiksha27: Maybe we can extract out the always blocking filter; would be more readable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/AdjustableSweepBatchConfigSource.java
```diff
@@ -42,8 +46,13 @@ public static AdjustableSweepBatchConfigSource create(
             Supplier<SweepBatchConfig> rawSweepBatchConfig) {
         AdjustableSweepBatchConfigSource configSource = new AdjustableSweepBatchConfigSource(rawSweepBatchConfig);
 
+        Gauge<Double> gauge = AdjustableSweepBatchConfigSource::getBatchSizeMultiplier;
+        metricsManager.addMetricFilter(AdjustableSweepBatchConfigSource.class,
+                "batchSizeMultiplier",
+                ImmutableMap.of(),
+                () -> gauge.getValue() != 1.0);
```
sudiksha27, line 53: I am confused b/w the description and implementation of filter here.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/InstrumentedTimelockService.java
jeremyk-91: Will do

PR #4868, opened 2020-06-26
Increase gradle build memory to 2 GB

blakehurlburt:
Full builds of atlasdb sometimes fail due to gradle running out of memory. The gradle wrapper does not allow passing JVM arguments for the build VM on the command-line, so it must be done in gradle.properties instead. Increasing the maximum amount of memory that can be allocated from the default of 512M to 2G.


jeremyk-91:
build passes in #4872

PR #4869, opened 2020-06-26
Namespace in TM

sudiksha27:
**Goals (and why)**:
Client feedback report for TimeLock should include both the original namespace and the overridden service name.

- Optional namespace field is added to ConjureTimeLockClientFeedback

- No new tests added

Today

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -618,7 +618,8 @@ public abstract class TransactionManagers {
         return serviceIdentifierOverride().orElseGet(this::namespace);
     }
 
-    private String namespace() {
+    @Value.Derived
+    String namespace() {
         return Stream.of(config().namespace(),
                 config().timelock().flatMap(TimeLockClientConfig::client),
                 config().keyValueService().namespace())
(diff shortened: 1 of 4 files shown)


lock-api/src/main/java/com/palantir/lock/client/metrics/TimeLockFeedbackBackgroundTask.java
```diff
@@ -88,6 +93,7 @@ public void scheduleWithFixedDelay() {
                         .atlasVersion(versionSupplier.get())
                         .nodeId(nodeId)
                         .serviceName(serviceName)
+                        .namespace(namespace)
                         .build();
```
sudiksha27, line 97: remove it for now

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -614,13 +615,17 @@ private MetricsManager setUpMetricsAndGetMetricsManager() {
     @VisibleForTesting
     @Value.Derived
     String serviceName() {
-        return serviceIdentifierOverride().orElseGet(() -> Stream.of(config().namespace(),
+        return serviceIdentifierOverride().orElseGet(this::namespace);
+    }
+
+    private String namespace() {
```
jeremyk-91, line 621: nit: could we make this `Value.Derived`, and/or have the service name take a `Suppliers.memoize(() -> namespace())`? I'm worried we're doing a bunch of unnecessary stream reads every time we look up the namespace.

PR #4873, opened 2020-06-29
[PD$-110002] Part 7: Cassandra Client Pool

jeremyk-91:
**Goals (and why)**:
- $$$

- Remove four gauge metrics from the CassandraClientPoolingContainer. These are registered once for each Cassandra node, meaning that we would save 4 * |Cassandra cluster| * |namespaces| time series per period on AtlasDb-Proxy.
- Three of these are derived metrics: the aggregation should be done using tools rather than also reporting.
- One more has always been zero **everywhere** in prod for the past **year** (I'm not sure that we do any meaningful validation on borrowing).

- Are the other metrics actually useful for debugging? I could see edge cases where they might be useful, but not wholly convinced.

**Where should we start reviewing?**: CassandraClientPoolingContainer.java

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -311,7 +311,6 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
         registerPoolMetric("numActive", pool::getNumActive);
         registerPoolMetric("created", pool::getCreatedCount);
         registerPoolMetric("destroyedByEvictor", pool::getDestroyedByEvictorCount);
-        registerPoolMetric("destroyedByBorrower", pool::getDestroyedByBorrowValidationCount);
     }
 
     private void registerPoolMetric(String metricName, Gauge gauge) {
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -309,14 +309,9 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
         registerPoolMetric("meanBorrowWaitTimeMillis", pool::getMeanBorrowWaitTimeMillis);
         registerPoolMetric("numIdle", pool::getNumIdle);
         registerPoolMetric("numActive", pool::getNumActive);
-        registerPoolMetric("approximatePoolSize", () -> pool.getNumIdle() + pool.getNumActive());
         registerPoolMetric("created", pool::getCreatedCount);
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -309,14 +309,8 @@ private void registerMetrics(GenericObjectPool<CassandraClient> pool) {
         registerPoolMetric("meanBorrowWaitTimeMillis", pool::getMeanBorrowWaitTimeMillis);
         registerPoolMetric("numIdle", pool::getNumIdle);
         registerPoolMetric("numActive", pool::getNumActive);
```
jkozlowski, line 311: I think this is a good change. Just wondering if we could basically keep all metrics in memory and then use your filtering infra to report outliers?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -309,14 +309,8 @@ private void registerMetrics(GenericObjectPool<CassandraClient> pool) {
         registerPoolMetric("meanBorrowWaitTimeMillis", pool::getMeanBorrowWaitTimeMillis);
```
jkozlowski, line 309: I'm curious about this metric as well, I thought we don't wait at all, but waiting is done in the CassandraRetrier. So in effect this might be timing a memory operation of checking out a connection from the pool

jkozlowski, line 309: meanIdleTimeMillis is interesting: I guess it might tell you if your min connection pool size is too large, but I don't think anyone ever cares? Also the idle timeout is 10 minutes, so I'd vote for canning this one too, unless you can think of a purpose for it?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
jeremyk-91: Yep, we could. This was intended to be the 'pain free' or 'risk free' version of this PR, that's definitely something that we could do (just more complex)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
jeremyk-91: Yeah, didn't really have a purpose for this: the main one was just if it was very close to 10m that could be interesting, but agree that it's not particularly valuable. Let's kill it

jeremyk-91: Hmm. There are places where this does run into 40+ ms, though I'd be fine with removing it: not sure what this would be used for, and you only have one data point in a long while. So while I don't think it's a memory op it doesn't seem very valuable

PR #4874, opened 2020-06-30
Consolidate PTExexutors fixed executor usage into string-based factories

carterkozak:
This new API will allow us to optimize thread utilization in a
future change without impacting callers.

### DIFF
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
@@ -207,17 +207,16 @@ public final class PTExecutors {
 
     /**
      * Creates a thread pool that reuses a fixed number of threads operating off a shared unbounded
-     * queue, using the provided ThreadFactory to create new threads when needed.  At any point, at
-     * most <tt>numThreads</tt> threads will be active processing tasks.  If additional tasks are
-     * submitted when all threads are active, they will wait in the queue until a thread is
-     * available.  If any thread terminates due to a failure during execution prior to shutdown, a
-     * new one will take its place if needed to execute subsequent tasks.  The threads in the pool
-     * will exist until it is explicitly {@link ExecutorService#shutdown}.
+     * queue.  At any point, at most <tt>numThreads</tt> threads will be active processing tasks.  If
+     * additional tasks are submitted when all threads are active, they will wait in the queue until
+     * a thread is available.  If any thread terminates due to a failure during execution prior to
+     * shutdown, a new one will take its place if needed to execute subsequent tasks.  The threads
+     * in the pool will exist until it is explicitly {@link
+     * ExecutorService#shutdown shutdown}.
      *
      * @param numThreads the number of threads in the pool
      * @param name Executor name used for thread naming and instrumentation
      * @return the newly created thread pool
-     * @throws NullPointerException if threadFactory is null
      * @throws IllegalArgumentException if <tt>numThreads &lt;= 0</tt>
      */
     public static ThreadPoolExecutor newFixedThreadPool(int numThreads, String name) {
--- a/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
+++ b/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
@@ -196,11 +196,32 @@ public final class PTExecutors {
      * @return the newly created thread pool
      * @throws NullPointerException if threadFactory is null
      * @throws IllegalArgumentException if <tt>numThreads &lt;= 0</tt>
+     * @deprecated Prefer {@link #newFixedThreadPool(int, String)}.
      */
+    @Deprecated
     public static ThreadPoolExecutor newFixedThreadPool(int numThreads, ThreadFactory threadFactory) {
         return newThreadPoolExecutor(numThreads, numThreads,
                 DEFAULT_THREAD_POOL_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS,
                 new LinkedBlockingQueue<Runnable>(), threadFactory);
+    }
+
+    /**
+     * Creates a thread pool that reuses a fixed number of threads operating off a shared unbounded
+     * queue, using the provided ThreadFactory to create new threads when needed.  At any point, at
+     * most <tt>numThreads</tt> threads will be active processing tasks.  If additional tasks are
+     * submitted when all threads are active, they will wait in the queue until a thread is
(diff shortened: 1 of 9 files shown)


commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
```diff
@@ -196,13 +196,34 @@ public static ThreadPoolExecutor newFixedThreadPool(int numThreads) {
...
+    @Deprecated
     public static ThreadPoolExecutor newFixedThreadPool(int numThreads, ThreadFactory threadFactory) {
         return newThreadPoolExecutor(numThreads, numThreads,
                 DEFAULT_THREAD_POOL_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS,
                 new LinkedBlockingQueue<Runnable>(), threadFactory);
     }
 
+    /**
+     * Creates a thread pool that reuses a fixed number of threads operating off a shared unbounded
+     * queue, using the provided ThreadFactory to create new threads when needed.  At any point, at
```
jkozlowski, line 210: There's no ThreadFactory passed in.

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
```diff
@@ -107,7 +107,7 @@ private AwaitingLeadershipProxy(
                 "Unable to create an AwaitingLeadershipProxy with no supplier");
         this.delegateSupplier = delegateSupplier;
         this.leaderElectionService = leaderElectionService;
-        this.executor = PTExecutors.newSingleThreadExecutor(PTExecutors.newNamedThreadFactory(true));
+        this.executor = PTExecutors.newSingleThreadExecutor();
```
jkozlowski, line 110: Feels like this should have a name?

jkozlowski: So this will loose daemon threading? @jeremyk-91

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
```diff
@@ -40,7 +39,7 @@ public PaxosStateLogBatchReader(PaxosStateLog<V> delegate, Persistable.Hydrator<
         this.delegate = delegate;
         this.hydrator = hydrator;
         this.executor = MoreExecutors.listeningDecorator(
-                PTExecutors.newFixedThreadPool(numThreads, new NamedThreadFactory("psl-reader", true)));
+                PTExecutors.newFixedThreadPool(numThreads, "psl-reader"));
```
jkozlowski, line 42: Again, no daemons?

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
carterkozak: This uses the same logic as before, discovering a name based on the call site.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
carterkozak: updated docs.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
carterkozak: Yes. this is a change to avoid daemon threads. Do you think non-daemon threads are required here?

leader-election-impl/src/main/java/com/palantir/leader/proxy/AwaitingLeadershipProxy.java
carterkozak: This used daemon threads before, and the default is using daemon threads. This one hasn't changed, only the psl threads are changing to daemons.

commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
```diff
@@ -196,13 +196,33 @@ public static ThreadPoolExecutor newFixedThreadPool(int numThreads) {
...
+     * in the pool will exist until it is explicitly {@link
+     * ExecutorService#shutdown shutdown}.
+     *
+     * @param numThreads the number of threads in the pool
+     * @param name Executor name used for thread naming and instrumentation
+     * @return the newly created thread pool
+     * @throws IllegalArgumentException if <tt>numThreads &lt;= 0</tt>
+     */
+    public static ThreadPoolExecutor newFixedThreadPool(int numThreads, String name) {
+        return newFixedThreadPool(numThreads, new NamedThreadFactory(name));
```
jeremyk-91, line 223: Might want to call out (method name? or at least docs) that this is non-daemon since a lot of the other methods in this class seem to use daemon threads

timelock-agent/src/main/java/com/palantir/timelock/paxos/LeaderPingHealthCheck.java
```diff
@@ -51,7 +50,7 @@ public LeaderPingHealthCheck(
         this.pingers = pingers;
         this.executorService = PTExecutors.newFixedThreadPool(
                 pingers.size(),
-                new NamedThreadFactory("leader-ping-healthcheck", true));
+                "leader-ping-healthcheck");
```
jeremyk-91, line 53: This one seems a bit suspicious - we're changing it to be non-daemon, and there doesn't seem to be anything that shuts down the executor service anywhere. Do we know/have we tested if timelock will shut down gracefully?

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogBatchReader.java
jeremyk-91: This one could be either way I think, it's part of the SQLite migration so it blocks startup for timelock, and there'll always be other main threads going around.

timelock-agent/src/main/java/com/palantir/timelock/paxos/LeaderPingHealthCheck.java
carterkozak: Ah, I had though the default NamedThreadFactory ctor used daemon threads, but that was incorrect. I definitely want this to use daemon threads.

I don't think we actually want any ptexecutors to use non-daemon threads since they've historically been a source of pain.

carterkozak:
policy bot?

PR #4875, opened 2020-06-30
Remove duplicate metrics from TimelockAgent.createSharedExecutor

carterkozak:
This removes the untagged variants of instrumentation, but leaves
the tagged executor and thread-factory instrumentation provided
by PTExecutors.
The tagged metrics are useful more broadly because they can be
used with general purpose dashboards and alerts.

I'm unsure if y'all alert on these, so I defer to your judgement whether or not to merge this change.

### DIFF
--- a/changelog/@unreleased/pr-4875.v2.yml
+++ b/changelog/@unreleased/pr-4875.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |-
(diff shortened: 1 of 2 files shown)


carterkozak:
ci failure looks like a flake downloading artifacts. Would you mind retriggering the build, or giving me perms to do so?

jeremyk-91:
Retriggered. Probably a separate discussion, but @sudiksha27 do you remember how we got you retrigger perms? I don't see it in circleci

PR #4879, opened 2020-07-02
add refresh failure callbacks

berler:
**Goals (and why)**:

Adds an option to register callback functions that get called whenever a
LockRefreshingLockService stops refreshing a lock because it failed to
refresh.

This helps with situations where a client intentionally will be holding
locks for a long time, and they want to be nofied immediately as soon as
refreshing the lock fails. Without this callback, clients would need to
implement their own background thread to check if the locks were still
valid.

We have an internal product that would already like to use this feature, and this will save us an additional background task that would currently be used to repeatedly check that locks are still valid.

Adds methods to register and remove callbacks from LockRefreshingLockService, and calls the callback when lock tokens fail to refresh. To avoid problems where a callback is very slow, the callbacks are only called on a new thread that is submitted to the same executor that runs refreshes. The executor service already has an unbounded max number of threads so it will create new threads and for these callbacks and then dispose of the threads when not needed.

Added a simple test.

N/A (simple change)

simple change

normal

### DIFF
--- a/changelog/@unreleased/pr-4879.v2.yml
+++ b/changelog/@unreleased/pr-4879.v2.yml
@@ -1,40 +1,8 @@
 type: improvement
 improvement:
   description: |-
-    add refresh failure callbacks
-
-    **Goals (and why)**:
-
     Adds an option to register callback functions that get called whenever a
     LockRefreshingLockService stops refreshing a lock because it failed to
     refresh.
-
-    This helps with situations where a client intentionally will be holding
-    locks for a long time, and they want to be nofied immediately as soon as
-    refreshing the lock fails. Without this callback, clients would need to
-    implement their own background thread to check if the locks were still
-    valid.
-
-    We have an internal product that would already like to use this feature, and this will save us an additional background task that would currently be used to repeatedly check that locks are still valid.
-
-    **Implementation Description (bullets)**:
-
-    Adds methods to register and remove callbacks from LockRefreshingLockService, and calls the callback when lock tokens fail to refresh. To avoid problems where a callback is very slow, the callbacks are only called on a new thread that is submitted to the same executor that runs refreshes. The executor service already has an unbounded max number of threads so it will create new threads and for these callbacks and then dispose of the threads when not needed.
-
-    **Testing (What was existing testing like?  What have you done to improve it?)**:
-
-    Added a simple test.
-
-    **Concerns (what feedback would you like?)**:
-
-    N/A (simple change)
-
-    **Where should we start reviewing?**:
-
-    simple change
-
-    **Priority (whenever / two weeks / yesterday)**:
-
-    normal
   links:
   - https://github.com/palantir/atlasdb/pull/4879
--- a/changelog/@unreleased/pr-4879.v2.yml
+++ b/changelog/@unreleased/pr-4879.v2.yml
@@ -1,0 +1,40 @@
+type: improvement
+improvement:
+  description: |-
+    add refresh failure callbacks
+
+    **Goals (and why)**:
+
+    Adds an option to register callback functions that get called whenever a
+    LockRefreshingLockService stops refreshing a lock because it failed to
+    refresh.
+
+    This helps with situations where a client intentionally will be holding
+    locks for a long time, and they want to be nofied immediately as soon as
+    refreshing the lock fails. Without this callback, clients would need to
+    implement their own background thread to check if the locks were still
+    valid.
+
+    We have an internal product that would already like to use this feature, and this will save us an additional background task that would currently be used to repeatedly check that locks are still valid.
+
+    **Implementation Description (bullets)**:
+
+    Adds methods to register and remove callbacks from LockRefreshingLockService, and calls the callback when lock tokens fail to refresh. To avoid problems where a callback is very slow, the callbacks are only called on a new thread that is submitted to the same executor that runs refreshes. The executor service already has an unbounded max number of threads so it will create new threads and for these callbacks and then dispose of the threads when not needed.
+
+    **Testing (What was existing testing like?  What have you done to improve it?)**:
+
+    Added a simple test.
+
+    **Concerns (what feedback would you like?)**:
+
+    N/A (simple change)
+
+    **Where should we start reviewing?**:
+
+    simple change
+
+    **Priority (whenever / two weeks / yesterday)**:
+
+    normal
+  links:
+  - https://github.com/palantir/atlasdb/pull/4879
--- a/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
@@ -21,9 +21,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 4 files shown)


changelog/@unreleased/pr-4879.v2.yml
```diff
@@ -0,0 +1,40 @@
...
+    locks for a long time, and they want to be nofied immediately as soon as
+    refreshing the lock fails. Without this callback, clients would need to
+    implement their own background thread to check if the locks were still
+    valid.
+
+    We have an internal product that would already like to use this feature, and this will save us an additional background task that would currently be used to repeatedly check that locks are still valid.
+
+    **Implementation Description (bullets)**:
+
+    Adds methods to register and remove callbacks from LockRefreshingLockService, and calls the callback when lock tokens fail to refresh. To avoid problems where a callback is very slow, the callbacks are only called on a new thread that is submitted to the same executor that runs refreshes. The executor service already has an unbounded max number of threads so it will create new threads and for these callbacks and then dispose of the threads when not needed.
```
carterkozak, line 22: > The executor service already has an unbounded max number of threads

I think it's a single threaded scheduled executor not an unbounded executor, right?

changelog/@unreleased/pr-4879.v2.yml
```diff
@@ -0,0 +1,40 @@
+type: improvement
+improvement:
+  description: |-
+    add refresh failure callbacks
```
carterkozak, line 4: changelog description should be between a sentence and a paragraph, not the PR template.

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -135,11 +138,25 @@ private void refreshLocks() {
...
+                    failedTokens = new HashSet<>();
+                }
+                failedTokens.add(token);
+            }
+        }
+        if (failedTokens != null && !failedTokens.isEmpty()) {
+            Set<LockRefreshToken> unmodifiableFailedTokens = Collections.unmodifiableSet(failedTokens);
+            // submit callbacks to the executor so a slow callback doesn't prevent future refreshes
+            for (FailedRefreshCallback callback : failedRefreshCallbacks) {
+                exec.submit(() -> callback.onFailedRefresh(unmodifiableFailedTokens));
```
carterkozak, line 159: This must use execute, otherwise exceptions will not be recorded.
```suggestion
                exec.execute(() -> callback.onFailedRefresh(unmodifiableFailedTokens));
```

It's unfortunate we don't run baseline errorprone pre-merge, it would have caught this:
https://github.com/palantir/gradle-baseline/blob/7039925b163c647a4e738444a400e67a9fcbec8e/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/ExecutorSubmitRunnableFutureIgnored.java#L33-L42

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -183,4 +200,23 @@ public void close() {
...
+     * Register a callback function to get called any time tokens fail to be refreshed.
+     */
+    public void registerRefreshFailedCallback(FailedRefreshCallback callback) {
+        failedRefreshCallbacks.add(callback);
+    }
+
+    /**
+     * Remove a callback function that was previously registered to be called when tokens fail to be refreshed.
+     */
+    public void removeRefreshFailedCallback(FailedRefreshCallback callback) {
```
carterkozak, line 219: Might be cleaner to return a `Disposable` from the registration function to unregister callbacks, imo the code is easier to follow and debug that way as opposed to reference-equality register/unregister functions.

lock-impl/src/test/java/com/palantir/lock/client/LockRefreshingLockServiceTest.java
```diff
@@ -74,4 +79,19 @@ public void testClosed() {
...
+    public void testFailedRefreshCallback() throws InterruptedException {
+        AtomicReference<Set<LockRefreshToken>> failedTokens = new AtomicReference<>();
+        server.registerRefreshFailedCallback(failedTokens::set);
+        LockRequest request = LockRequest.builder(ImmutableSortedMap.of(lock1, LockMode.WRITE))
+                .timeoutAfter(SimpleTimeDuration.of(5, TimeUnit.SECONDS))
+                .build();
+        LockResponse lock = server.lockWithFullLockResponse(LockClient.ANONYMOUS, request);
+        // use the internal server to unlock (refreshing service will still try to refresh but fail)
+        internalServer.unlock(lock.getToken());
+        Thread.sleep(10000);
```
carterkozak, line 93: Can we use awaitility instead of sleeping?

j-baker: beyond this, please can we figure out a way to have tests that don't just take 10 seconds? seems pretty unreasonable

changelog/@unreleased/pr-4879.v2.yml
berler: It's a ScheduledThreadPoolExecutor with a corePoolSize of 1. If you follow through to when it actually constructs it the maxPoolSize is Integer.MAX_VALUE.

carterkozak: Executors are subtle and dangerous creatures. Can you write a test that shows tasks executing in parallel on this executor?

lock-impl/src/test/java/com/palantir/lock/client/LockRefreshingLockServiceTest.java
berler: Unfortunately the 5-second refresh time is hard-coded in LockRefreshingLockService and all of the refresh logic is not exposed in any way that tests can interact with it. We'd need to refactor that to be able to make tests shorter. I think refactoring LockRefreshingLockService is out of scope for the change I'm trying to make.

changelog/@unreleased/pr-4879.v2.yml
berler: Done

changelog/@unreleased/pr-4879.v2.yml
berler: You're right, this doesn't appear to work as I had expected right now. I think I'll need to use a separate ExecutorService for the callbacks and use a SynchronousQueue so that it actually creates new threads on demand instead of just queuing up tasks.

berler: Added a separate cached executor service to use for the callbacks so this should work as expected now (and a test that asserts it actually does run the callbacks in parallel).

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
berler: Done

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
berler: Done

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -75,7 +82,9 @@ public static LockRefreshingLockService create(LockService delegate) {
     private LockRefreshingLockService(LockService delegate) {
         this.delegate = delegate;
         toRefresh = ConcurrentHashMap.newKeySet();
+        failedRefreshCallbacks = ConcurrentHashMap.newKeySet();
         exec = PTExecutors.newScheduledThreadPool(1, PTExecutors.newNamedThreadFactory(true));
+        callbackExec = PTExecutors.newCachedThreadPool("LockRefreshingLockService callbacks");
```
jeremyk-91, line 87: Unbounded seems a bit scary. Do we know what kind of tasks users plan to schedule? My main worry is if they are _too_ long running we get threading issues.

lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -183,4 +207,26 @@ public void close() {
...
+    public interface FailedRefreshCallback {
+        void onFailedRefresh(Set<LockRefreshToken> tokens);
+    }
+
+    /**
+     * Register a callback function to get called any time tokens fail to be refreshed.
+     *
+     * <p>Call {@link Disposable#dispose} on the returned value to remove the callback.
+     *
+     * @throws SafeIllegalArgumentException if the callback was already registered.
```
jeremyk-91, line 221: Can we add a warning here that this API should not be used outside of large internal product unless you have the AtlasDB dev team's approval? I wouldn't want users in the metal product to end up using this.

PR #4885, opened 2020-07-07
Add original client namespace while sending adjudication feedback

sudiksha27:
**Goals (and why)**:
To have original client namespace while sending adjudication feedback. This is required to determine the leader for the client.

**Concerns**:
- Is it safe suppress the case where namespace is not present and fall back quietly to service name? 

Today

### DIFF
--- a/lock-api-objects/build.gradle
+++ b/lock-api-objects/build.gradle
@@ -23,7 +23,6 @@ dependencies {
     compileOnly project(":atlasdb-processors")
 }
 
-
 recommendedProductDependencies {
     productDependency {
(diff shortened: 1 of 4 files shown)


lock-api-objects/build.gradle
```diff
@@ -27,7 +27,7 @@ recommendedProductDependencies {
     productDependency {
         productGroup = 'com.palantir.timelock'
         productName = 'timelock-server'
-        minimumVersion = '0.200.0'
+        minimumVersion = '0.218.0'
```
jeremyk-91, line 30: verified this exists

PR #4889, opened 2020-07-08
Tweak acceptable request rate

sudiksha27:
**Goals (and why)**:
This PR lowers the minimum threshold for request rate (defensive estimate considering Autobatchers). There is also an additional check to flag badness when p99 is very high, irrespective of request rate.

Test case added

None

FeedbackHandler.java

Tomorrow

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/Constants.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/Constants.java
@@ -33,18 +33,18 @@ public final class Constants {
 
     public static final ServiceLevelObjectiveSpecification LEADER_TIME_SERVICE_LEVEL_OBJECTIVES
             = ServiceLevelObjectiveSpecification.builder()
-            .maximumPermittedP99(Duration.ofMillis(200))
-            .minimumRequestRateForConsideration(0.1)
+            .maximumPermittedSteadyStateP99(Duration.ofMillis(200))
+            .minimumRequestRateForConsideration(0.025)
             .maximumPermittedErrorProportion(0.5)
-            .p99Multiplier(100)
+            .maximumPermittedQuietP99(Duration.ofSeconds(40))
             .build();
 
     public static final ServiceLevelObjectiveSpecification START_TRANSACTION_SERVICE_LEVEL_OBJECTIVES
             = ServiceLevelObjectiveSpecification.builder()
-            .maximumPermittedP99(Duration.ofMillis(500))
-            .minimumRequestRateForConsideration(0.1)
+            .maximumPermittedSteadyStateP99(Duration.ofMillis(500))
+            .minimumRequestRateForConsideration(0.02)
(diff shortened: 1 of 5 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/FeedbackHandler.java
```diff
@@ -149,7 +148,13 @@ HealthStatus pointFeedbackHealthStatus(ConjureTimeLockClientFeedback healthRepor
     private HealthStatus getHealthStatusForService(EndpointStatistics endpointStatistics,
             double rateThreshold,
             long p99Limit,
-            double errorRateProportion) {
+            double errorRateProportion,
+            double p99Multiplier) {
+
+        // Outliers indicate badness, regardless of request rate
+        if (endpointStatistics.getP99() > p99Limit * p99Multiplier) {
```
jeremyk-91, line 155: On one hand I'd prefer not to add complexity - on the other I think we should special case a flat 0 for the request rate, because this is backed by an exponential metric registry and so an old, slow request might end up hanging for more than we would like.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/Constants.java
```diff
@@ -34,15 +34,17 @@ private Constants() {
...
+            .minimumRequestRateForConsideration(0.1)
             .maximumPermittedErrorProportion(0.5)
+            .p99Multiplier(100)
             .build();
 
     public static final ServiceLevelObjectiveSpecification START_TRANSACTION_SERVICE_LEVEL_OBJECTIVES
             = ServiceLevelObjectiveSpecification.builder()
             .maximumPermittedP99(Duration.ofMillis(500))
-            .minimumRequestRateForConsideration(1)
+            .minimumRequestRateForConsideration(0.1)
```
jeremyk-91, line 45: Could we make this `0.02`?

The reason is because these endpoints on the AtlasDB side are autobatched. So the current p99 multiplier there means we'll say unhealthy on 50 seconds or above, and it would be good to catch a server that is queueing up 40 second autobatcher calls.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/ServiceLevelObjectiveSpecification.java
```diff
@@ -27,6 +27,7 @@
     Duration maximumPermittedP99();
     double maximumPermittedErrorProportion();
     double minimumRequestRateForConsideration();
+    double p99Multiplier();
```
jeremyk-91, line 30: I'd prefer a design where we have something like

```
  Duration maximumPermittedSteadyStateP99();
  Duration maximumPermittedQuietP99(); // currently this is maximumPermittedP99 * multiplier
```

with comments explaining the difference. The multiplier doesn't really say what it does

PR #4891, opened 2020-07-08
[PoC][PD$-110002] Part 10-12: Top N for SnapshotTransaction / Table-Level Metrics

jeremyk-91:
**This PR exists to show how I plan on fitting the pieces together. I will open individual PRs for components for merging.**

**Goals (and why)**:
- $$$
- With many table-level metrics, we are interested in being able to identify which tables are offending. This can be useful when dealing with high-priority issues - we normally want to see if a given table has an unexpectedly large number of reads or writes.
- We are also interested in load spikes.

- Add a controller that publishes the top-N series that are registered to it.
- Instead of publishing the number of cells read / other counters, have a gauge that publishes the number of entries processed in the last 30 seconds.
- Publish counter metrics for the top ten tables, in terms of their change in the last 30 second period, inside SnapshotTransaction.

- Does the idea of looking for _deltas_ and otherwise publishing the busiest tables in the steady state make sense?
- Is the zero based delta gauge a problem?

**Where should we start reviewing?**: TopNMetricPublicationController

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManagers.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/MetricsManagers.java
@@ -38,4 +38,12 @@ public final class MetricsManagers {
     public static MetricsManager createForTests() {
         return MetricsManagers.of(new MetricRegistry(), new DefaultTaggedMetricRegistry());
     }
+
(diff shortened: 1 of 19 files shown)


jeremyk-91:
All parts have merged! Closing out.

PR #4892, opened 2020-07-08
[PD$-110002] Part 10: Top-N Tables | Primitives

jeremyk-91:
**Goals (and why)**:
- See #4891 

- TopNMetricPublicationController sets up metrics filters so only the top-N things get published.
- ZeroBasedDeltaGauge reports the delta between measurements (the eventual goal is handling load spikes)

Unit tests for the new things

Nothing much

**Where should we start reviewing?**: TopNMetricPublicationController

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TopNMetricPublicationController.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/TopNMetricPublicationController.java
@@ -18,11 +18,13 @@ package com.palantir.atlasdb.util;
 
 import java.time.Duration;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import com.codahale.metrics.Gauge;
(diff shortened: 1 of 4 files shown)


atlasdb-client/src/test/java/com/palantir/atlasdb/util/TopNMetricPublicationControllerTest.java
```diff
@@ -0,0 +1,153 @@
...
+        Gauge<Long> gauge2 = value2::get;
+
+        TopNMetricPublicationController<Long> controller = TopNMetricPublicationController.create(50);
+        MetricPublicationFilter filter1 = controller.registerAndCreateFilter(gauge1);
+        MetricPublicationFilter filter2 = controller.registerAndCreateFilter(gauge2);
+
+        assertThat(filter1.shouldPublish()).isTrue();
+        assertThat(filter2.shouldPublish()).isTrue();
+    }
+
```
sudiksha27, line 143: can we also have a test that checks lower bound is refreshed?

atlasdb-client/src/main/java/com/palantir/atlasdb/util/TopNMetricPublicationController.java
```diff
@@ -0,0 +1,99 @@
...
+        gauges.add(gauge);
+        return () -> shouldPublishIndividualGaugeMetric(gauge);
+    }
+
+    private boolean shouldPublishIndividualGaugeMetric(Gauge<T> constituentGauge) {
+        T value = constituentGauge.getValue();
+        return lowerBoundSupplier.get()
+                .map(lowerBound -> isAtLeastLowerBound(value, lowerBound))
+                .orElse(true); // This means there aren't maxPermittedRank series, so we should publish the metric.
+    }
```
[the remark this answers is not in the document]
jeremyk-91: That's a good flag. I think originally I thought this was fine, though you're right that maybe we should just arbitrarily pick if there are ties.

atlasdb-client/src/test/java/com/palantir/atlasdb/util/TopNMetricPublicationControllerTest.java
jeremyk-91: Will do!

PR #4894, opened 2020-07-09
[LW] Test for lock watch updates that are too far behind

Jolyon-S:
**Goals (and why)**:
Catch a case in the code where the successful update has a version before the snapshot.
Since the start transaction flow is synchronous and batched, it should not be possible to have this case - it would mean that the cache has received some updates, and the transactions that carried those updates have committed, all before the second set of updates comes in. This defends against it anyway.

* Added a test for successful updates that have a version before the snapshot (which essentially represents the earliest version for which we have events).

* Added another test to `LockWatchEventCacheIntegrationTests`.

* Is this change necessary? Note that if the above case were to happen with the current code, the main point of worry is the commit update, and that would end up invalidating all (i.e. causing a conflict, and retrying the transaction anyway).

* LockWatchEventLog

Whenever.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Iterables;
+import com.palantir.atlasdb.transaction.api.TransactionLockWatchFailedException;
 import com.palantir.lock.watch.IdentifiedVersion;
 import com.palantir.lock.watch.LockWatchCreatedEvent;
 import com.palantir.lock.watch.LockWatchEvent;
@@ -120,6 +121,13 @@ final class LockWatchEventLog {
 
     private void processSuccess(LockWatchStateUpdate.Success success) {
         Preconditions.checkState(latestVersion.isPresent(), "Must have a known version to process successful updates");
+        Optional<IdentifiedVersion> snapshotVersion = snapshot.getSnapshotVersion();
+        Preconditions.checkState(snapshotVersion.isPresent(),
+                "Must have a snapshot before processing successful updates");
+
+        if (success.lastKnownVersion() < snapshotVersion.get().version()) {
+            throw new TransactionLockWatchFailedException("Cannot process events before the oldest event");
+        }
 
         if (success.lastKnownVersion() > latestVersion.get().version()) {
             assertEventsAreContiguousAndNoEventsMissing(success.events());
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -120,6 +121,13 @@ private IdentifiedVersion getLatestVersionAndVerify(IdentifiedVersion endVersion
 
     private void processSuccess(LockWatchStateUpdate.Success success) {
         Preconditions.checkState(latestVersion.isPresent(), "Must have a known version to process successful updates");
+        Optional<IdentifiedVersion> snapshotVersion = snapshot.getSnapshotVersion();
+        Preconditions.checkState(snapshotVersion.isPresent(),
+                "Must have a snapshot before processing successful updates");
```
Jolyon-S, line 126: Already enforced elsewhere (i.e. must have a snapshot before processing successes), but this is just because we have to handle optionals everywhere.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -77,7 +78,7 @@
             LockEvent.builder(ImmutableSet.of(DESCRIPTOR), LockToken.of(EVENT2_UUID)).build(7L);
 
     private static final UUID LEADER = UUID.fromString("470c855e-f77b-44df-b56a-14d3df085dbc");
-    public static final LockWatchStateUpdate.Success SUCCESS_2 = LockWatchStateUpdate.success(LEADER, 7L,
+    private static final LockWatchStateUpdate.Success SUCCESS_2 = LockWatchStateUpdate.success(LEADER, 7L,
```
Jolyon-S, line 81: Drive by fix.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -120,6 +121,13 @@ private IdentifiedVersion getLatestVersionAndVerify(IdentifiedVersion endVersion
...
+
+        if (success.lastKnownVersion() < snapshotVersion.get().version()) {
+            throw new TransactionLockWatchFailedException("Cannot process events before the oldest event");
```
jkozlowski, line 129: Checked that this should get retried both by atlasdb-proxy and atlasdb internally.

jkozlowski: You might wanna add a bit more around the "this will be retried and should happen very rarely".

PR #4898, opened 2020-07-13
CassandraClientPoolHostLevelMetric fields final

blakehurlburt:
Public fields on enums should be marked with final in order to pass internal testing.


jeremyk-91:
merging: see internal build 33966.

PR #4899, opened 2020-07-13
Stream store batch loads blocks.

jkozlowski:
**Goals (and why)**:

Enable loading multiple blocks at the same time (with default setting 4)


jkozlowski:
@jkong the build is just some weirdness with building RCs again

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractGenericStreamStore.java
```diff
@@ -225,7 +214,8 @@ private void tryWriteStreamToFile(Transaction transaction, T id, StreamMetadata
 
     protected abstract File createTempFile(T id) throws IOException;
 
-    protected abstract void loadSingleBlockToOutputStream(Transaction tx, T streamId, long blockId, OutputStream os);
```
jeremyk-91, line 228: This is fine. I don't know if the large internal product has this lined up appropriately to auto-regenerate their schemas, so we might need to do something if it doesn't

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
```diff
@@ -321,27 +326,56 @@ private void createTempFile() {
...
-                    } line("} catch (RuntimeException e) {"); {
-                        line("log.error(\"Error storing block {} for stream id {}\", row.getBlockId(), row.getId(), e);");
+                line("protected void loadBlocksToOutputStream(Transaction t, ", StreamId,
+                        " streamId, long firstBlock, long numBlocks, OutputStream os) {");
+                {
+                    line("try {");
+                    {
+                        line("BiMap<", StreamValueRow, ", Long> blockRows");
+                        line("        = KeyedStream.of(");
+                        line("        LongStream.rangeClosed(firstBlock, firstBlock + numBlocks).boxed())");
```
jeremyk-91, line 338: This should just be `range()` I think (open interval, not closed)

jkozlowski: How come tests were passing I wonder?

atlasdb-ete-test-utils/src/main/java/com/palantir/atlasdb/todo/generated/SnapshotsStreamStore.java
```diff
@@ -179,22 +184,35 @@ protected File createTempFile(Long id) throws IOException {
...
         try {
-            os.write(getBlock(t, row));
+            BiMap<SnapshotsStreamValueTable.SnapshotsStreamValueRow, Long> blockRows
+                    = KeyedStream.of(
+                    LongStream.rangeClosed(firstBlock, firstBlock + numBlocks).boxed())
+                    .mapKeys(blockId -> SnapshotsStreamValueTable.SnapshotsStreamValueRow.of(streamId, blockId))
+                    .collectTo(HashBiMap::create);
+            Map<SnapshotsStreamValueTable.SnapshotsStreamValueRow, byte[]> blocks = getBlocks(t, blockRows.keySet());
+            for (long i = 0; i < numBlocks; i++) {
+                os.write(blocks.get(blockRows.inverse().get(firstBlock + 1)));
```
Jolyon-S, line 196: Uhh, this looks very much wrong - should this not be `firstBlock + i` like the rest?!

atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/render/StreamStoreRenderer.java
Jolyon-S: Fixed.

atlasdb-ete-test-utils/src/main/java/com/palantir/atlasdb/todo/generated/SnapshotsStreamStore.java
Jolyon-S: Fixed.

Jolyon-S:
@jkozlowski  / @jeremyk-91 I just want a sanity check here:
`loadBlocksToOutputStream` is called when the input stream created in `AbstractGenericStreamStore#makeStream` is consumed, right?
This inputStream is made from a `BlockConsumingInputStream` which calls `blockGetter.get(..., numBlocksToRead, ...)`. If I'm not mistaken, `numBlocksToGet` is essentially the number of blocks in the stream to consume. So won't this just try to load the entire thing at once? Previously, it called `loadNblocksToOutputStream` which essentially did it sequentially, but since we're trying to load multiple at a time, we're just going to load _all_ at a time, right?

The PR description hints that we want 4 at a time, but it appears to me that that was never done - I don't see the number 4 anywhere in this PR. Does that sound about right for why this caused issues with kvs-slow-log-2?

I expect the way to fix is to re-introduce `loadNBlocksToOutputStream`, but delegate to the new method instead, with number set to 4.

Jolyon-S:
Right, I've done what I said above. @jeremyk-91 would you be able to have a quick review? It's not the most elegant so if you have a better suggestion, let me know.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractGenericStreamStore.java
```diff
@@ -204,6 +204,20 @@ private void writeStreamToFile(Transaction transaction, T id, StreamMetadata met
...
+    private void loadNBlocksToOutputStream(
+            Transaction tx,
+            T streamId,
+            long firstBlock,
+            long numBlocks,
+            OutputStream os) {
+        long position = 0;
+        for (long i = 0; i < numBlocks; i += 4) {
+            loadBlocksToOutputStream(tx, streamId, firstBlock + i, 4, os);
+            position += i;
```
jeremyk-91, line 216:
```suggestion
            position = i;
```

You want the last value where you left off. If there are 9 blocks, this would end up being 12, but you probably want it to be 8.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractGenericStreamStore.java
```diff
@@ -204,6 +204,20 @@ private void writeStreamToFile(Transaction transaction, T id, StreamMetadata met
...
+        }
+        loadBlocksToOutputStream(tx, streamId, firstBlock + position, numBlocks % 4, os);
```
jeremyk-91, line 218: Confirmed this doesn't explode if you try to read 0 blocks.

atlasdb-client/src/main/java/com/palantir/atlasdb/stream/AbstractGenericStreamStore.java
```diff
@@ -210,9 +211,11 @@ private void loadNBlocksToOutputStream(
             long firstBlock,
             long numBlocks,
             OutputStream os) {
-        for (long i = 0; i < numBlocks; i++) {
-            loadSingleBlockToOutputStream(tx, streamId, firstBlock + i, os);
+        long position = 0;
+        for (; position + BLOCK_BATCH <= numBlocks; position += BLOCK_BATCH) {
```
jkozlowski, line 215: Are you sure this is right? The batching should be controlled outside of this method. It is actually a parameter to the codegen, so you should not invent a new constant here.

Jolyon-S: I know it's parameterised in the code-gen too, but I don't think you want to pass for example 100 blocks to that method, as it will get you all 100 blocks at once. This is the closest thing to what the old gen did, but for 2 blocks at once (although it's entirely possible that I'm misunderstanding something along the way).

jkozlowski: My point is that this method should be doing exactly what it's being asked to do (so if it's 100 blocks, it's 100 blocks) and the users control the batching somewhere else. There's efficient batching below this method too, so it should not be making batching decisions.

Jolyon-S: the first pass, and then even with this at 4, caused SLOs to fail (although not at this level) while upgrading atlasdb-proxy.

The "user" that consumes this is `BlockGetter` if I'm not mistaken - so maybe the point is that the methods there need to change it, but by the time the user gets the result from this, I'm pretty certain that it's just an `InputStream` that they don't know or care about how many blocks are being loaded - that's the responsibility of code in this area (whether this class or `BlockGetter` or one of the other classes on that path), I'm not sure, but it should surely amount to roughly the same.

PR #4900, opened 2020-07-14
[PDS-125544] Configurable, Instrumented Feedback Sink

jeremyk-91:
**Goals (and why)**:
- Attempt to debug inexplicable performance regressions, _even with the same server side code_ on internal timelock.
- Understand the memory footprint of adjudication better.

- Add runtime config to make adjudication configurable

FeedbackHandlerTest should cover most of it.

Nothing in particular. Is there anything actually broken?

**Where should we start reviewing?**: TLAC


timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockAdjudicationConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+package com.palantir.timelock.config;
+
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(as = ImmutableTimeLockAdjudicationConfiguration.class)
+@Value.Immutable
+public interface TimeLockAdjudicationConfiguration {
+    default boolean enabled() {
```
jeremyk-91, line 26: This won't deserialize

PR #4901, opened 2020-07-16
[PDS-126257] Turning Tables: Checked Rejection Executor Services

jeremyk-91:
**Goals (and why)**:
- See PDS-126257
- We can't trust developers (including myself) to correctly remember to catch RejectedExecutionException: it's a random case that only happens occasionally at our biggest deployments.

- A {@link CheckedRejectionExecutorService} is an executor where submitters of tasks MUST be prepared to handle execution being rejected - and we thus throw a checked exception when submitting tasks. (Thanks to @j-baker for the idea!)
- This should make things fail at compile time.

- Added tests for the new classes. Otherwise the compiler should be as good a test as any that wiring is correct.

- Did I fudge the wiring up nonetheless?

**Where should we start reviewing?**: CheckedRejectionExecutorService.java

**Priority (whenever / two weeks / yesterday)**: next week?

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
@@ -27,7 +27,7 @@ import com.palantir.common.concurrent.PTExecutors;
 
 final class TimeLockPaxosExecutors {
     @VisibleForTesting
-    static final int MAXIMUM_POOL_SIZE = 256;
+    static final int MAXIMUM_POOL_SIZE = 384;
 
     private TimeLockPaxosExecutors() {
         // no
--- a/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
+++ b/timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
@@ -23,6 +23,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.palantir.common.concurrent.CheckedRejectionExecutorService;
 import com.palantir.common.concurrent.PTExecutors;
 
 final class TimeLockPaxosExecutors {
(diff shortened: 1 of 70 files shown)


timelock-agent/src/main/java/com/palantir/atlasdb/timelock/paxos/TimeLockPaxosExecutors.java
```diff
@@ -39,15 +40,17 @@ private TimeLockPaxosExecutors() {
      *
      * It is assumed that tasks run on the local node will return quickly (hence the use of the direct executor).
      */
-    static <T> Map<T, ExecutorService> createBoundedExecutors(
-            MetricRegistry metricRegistry, LocalAndRemotes<T> localAndRemotes, String useCase) {
+    static <T> Map<T, CheckedRejectionExecutorService> createBoundedExecutors(
+            LocalAndRemotes<T> localAndRemotes, String useCase) {
         int numRemotes = localAndRemotes.remotes().size();
```
sudiksha27, line 45: 👍

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/AutobatchingPaxosAcceptorNetworkClientFactory.java
```diff
@@ -28,6 +28,7 @@
 import com.google.common.util.concurrent.ListenableFuture;
```
sudiksha27, line 28: nit: there are unused imports in some other places as well where ExecutorService is replaced by CheckedRejectionExecutorService

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/PaxosQuorumCheckingCoalescingFunction.java
```diff
@@ -111,7 +112,7 @@ private PaxosQuorumCheckingCoalescingFunction(
     public static <REQ, RESP extends PaxosResponse, SERVICE, F extends CoalescingRequestFunction<REQ, RESP>>
     PaxosQuorumCheckingCoalescingFunction<REQ, RESP, F> wrapWithRemotes(
             List<SERVICE> services,
-            Map<SERVICE, ExecutorService> executors,
+            Map<SERVICE, CheckedRejectionExecutorService> executors,
             int quorumSize,
             Function<SERVICE, F> functionFactory) {
         List<FunctionAndExecutor<F>> functionsAndExecutors = KeyedStream.of(services)
```
sudiksha27, line 118: Where is this used?

jeremyk-91: It's used inside `wrap()`. I think you might be referencing the private constructor on line 68 though, which is unused? I've removed that.

jeremyk-91: Ah, I guess it's not. Well spotted. I'll remove this.

PR #4903, opened 2020-07-17
POC | TimeLock B/G

sudiksha27:
**Goals (and why)**:


leader-election-impl/src/main/java/com/palantir/leader/PaxosLeaderElectionService.java
```diff
@@ -356,6 +362,33 @@ private static long getNextSequenceNumber(Optional<PaxosValue> paxosValue) {
         return paxosValue.map(PaxosValue::getRound).orElse(PaxosAcceptor.NO_LOG_ENTRY) + 1;
     }
 
+    // todo sudiksha
+    @Override
+    public void forcefullyTakeoverLeadershipIfAdjudicating() {
+        if (leaderPinger.isCurrentAdjudicating()) {
```
sudiksha27, line 368: async?

PR #4910, opened 2020-07-23
Copy batch events

sudiksha27:
**Goals (and why)**:


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -99,10 +99,19 @@ public void close() {
         }
     }
 
-    private static final class DefaultBatchElement<T, R> implements BatchElement<T, R> {
+    private static final class DefaultBatchElement<T extends DeepCopy, R> implements BatchElement<T, R> {
         private T argument;
```
sudiksha27, line 103: copy constructor

jkozlowski:
Can you explain what bug do we actually think is here?

jkozlowski:
Did a bit of a code scan and some reading, and I'm not convinced that as long as we're only flushing using the endOfBatch argument pushed to us that this would be a bug?

Also this: https://stackoverflow.com/questions/18299903/lmax-disruptor-must-eventhandler-clone-object-received-from-eventhandleroneven

@j-baker

PR #4914, opened 2020-07-23
[PDS-127937] Distinct Executors for Pinging as Opposed to Paxos

jeremyk-91:
**Goals (and why)**:
- Still allow pinging a node that has a lot of ongoing Paxos computations.

- Use a separate thread pool.

- Added a test that verifies the creation paths are distinct.

- Potential load of 8*2 more threads when a node is really bad. I think it is actually just 1*2 because of autobatching, but went with 8 in case I misunderstood the code.

**Where should we start reviewing?**: PaxosRemoteClients


jeremyk-91:
Changelogs exist so pushing forward with merge

PR #4916, opened 2020-07-27
[PDS-128523] Don't validate immutable locks on every read.

felixdesouza:
**Goals (and why)**:
We validate locks and precommit conditions on every read. It should be sufficient to check at commit time (for effectively read only transactions). But we want to avoid wasted work where we do a ton of reads but our condition is invalid. A good balance would be having a rate limit.

See internal ticket PDS-128523. But crux is, *each* iterator returned by `getRowsColumnRangeIterator` will validate on every batch read from the KVS. If you want to combine those iterators by any chance, you will then have at most `n` sequential calls to validate pre commit requirements.

* Always validate on commit (if thoroughly swept) regardless of `validateOnReads`
* Rate limit number of calls

Somewhat hard to test, but open to ideas. Existing tests are broken, but happy to change once we've established a good way to test this.

Preferably quite soon, but understand other priorities.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -17,7 +17,6 @@ package com.palantir.atlasdb.transaction.impl;
 
 import java.nio.ByteBuffer;
 import java.time.Duration;
-import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -81,6 +80,7 @@ import com.google.common.primitives.UnsignedBytes;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.RateLimiter;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.AtlasDbPerformanceConstants;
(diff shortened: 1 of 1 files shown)


jeremyk-91:
We're uncomfortable with this because it changes the semantics of Atlas in some cases (in particular around exceptions and side effects). This would have that happen in rarely seen cases, leading to bugs that would be rarely seen. 

e.g. while I know that today this is seen as bad practice, we do have a number of important legacy consumers that have transactions that call out to external services on each row read within a transaction task.

Agree that the general behaviour, that validation on _every_ iterator is needed, is not ideal. There are different Atlas-side fixes we can and will implement to help address the original issue.

felixdesouza:
Yeah, we had a similar discussion when it was being implemented. Although I find it hard to believe that side-effecting transactions are correct (in terms of the consumer) *now* anyway i.e. transactions can be retried. Unless of course the side effecting transaction calls external consumers that are transactional..

Also, correct me if wrong, but the lock check seems to check that if you read back an empty value and that if you hold the lock you know it wasn't swept away by sweep. So really all a successful call back from a `get*` says that your empty read is not swept away, but it may not be the right value. So unless the external service is transactional, I don't really understand how their calls can be correct, even worse if the service is stateful.

In any case, it feels weird to establish defined behaviour for side effects in transactions, but it's up to you! Let me know what you decide on!

jeremyk-91:
Yeah, it's fair to say that the expectations you can make based on the things you read are fairly complicated. But given that that pattern is not all that rare in large internal product, difficult to track down and kind of works in the current state, I think it's a bit risky to change it.

@jkozlowski I believe is looking at a way of handling the range request use case safely (by recognising that when you've done the first batch, it was fine at that time), modulo support loads

felixdesouza:
Fair enough, I have a bit of bandwidth, is this something I can contribute? If I've understood correctly, don't check on the first batch since it's eagerly loaded, but fallback to checking on every batch call?

jeremyk-91:
Yep! I think @jkozlowski might have got in touch yesterday, but that's exactly the fix I had in mind

felixdesouza:
Closing in favour of #4929.

PR #4917, opened 2020-07-27
[PDS-127121] Log refreshed Cassandra hosts at info level

Jolyon-S:
**Goals (and why)**:
Expose more information about the state of the Cassandra pool in logs.

Changed `DEBUG` -> `INFO` on one log line (that details which hosts that have been added and removed).

N/A.

I don't think this introduces much spam - it is unlikely that anyone refreshes their token ranges / servers more than once per 30 seconds (default is once per 120 seconds), so one log line per refresh is minimal.

As for whether this is enough - I believe so, as it would tell us what hosts that are being added/removed at each point so that we can have a reasonable picture of what the pool is. Naturally this doesn't tell us the _actual_ current state of the pool, only the diff - but `debugLogStateOfPool` is a lot more heavy-handed, and possibly not necessary here.

`CassandraClientPoolImpl`

Whenever

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
@@ -315,7 +315,7 @@ public class CassandraClientPoolImpl implements CassandraClientPool {
         logRefreshedHosts(serversToAdd, serversToRemove);
     }
 
-    private void logRefreshedHosts(Set<InetSocketAddress> serversToAdd, Set<InetSocketAddress> serversToRemove) {
+    private static void logRefreshedHosts(Set<InetSocketAddress> serversToAdd, Set<InetSocketAddress> serversToRemove) {
         if (serversToRemove.isEmpty() && serversToAdd.isEmpty()) {
             log.debug("No hosts added or removed during Cassandra pool refresh");
         } else {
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -312,7 +312,7 @@ private void setServersInPoolTo(Set<InetSocketAddress> desiredServers) {
             cassandra.refreshTokenRangesAndGetServers();
         }
 
-        log.debug("Cassandra pool refresh added hosts {}, removed hosts {}.",
+        log.info("Cassandra pool refresh added hosts {}, removed hosts {}.",
```
jeremyk-91, line 315: nit: How valuable is an empty refresh line? Maybe we can have a refresh that did something at INFO and a refresh that did nothing at DEBUG.

Jolyon-S: I like it.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolImpl.java
```diff
@@ -312,9 +312,17 @@ private void setServersInPoolTo(Set<InetSocketAddress> desiredServers) {
             cassandra.refreshTokenRangesAndGetServers();
         }
 
-        log.debug("Cassandra pool refresh added hosts {}, removed hosts {}.",
-                SafeArg.of("serversToAdd", CassandraLogHelper.collectionOfHosts(serversToAdd)),
-                SafeArg.of("serversToRemove", CassandraLogHelper.collectionOfHosts(serversToRemove)));
+        logRefreshedHosts(serversToAdd, serversToRemove);
+    }
+
+    private void logRefreshedHosts(Set<InetSocketAddress> serversToAdd, Set<InetSocketAddress> serversToRemove) {
```
jeremyk-91, line 318: nit: static?

PR #4921, opened 2020-07-29
[PD$-110002] Part 13: Filter Resiliency

jeremyk-91:
**Goals (and why)**:
- Metrics filters shouldn't cause metrics to broadly not be published. This may cause issues when bootstrapping a cluster of leader-block users, and also if timelock is inaccessible (though it may not be the primary concern in that case...)

- If testing a filter throws an exception, treat that as returning true.

- Added tests for handling of exceptions

- Will there be log spam?
- Doing an RPC inside getMetrics() is kind of crap, but since it only happens every 5 minutes in practice (see cached composed supplier) I don't see it as an issue.

**Where should we start reviewing?**: MetricPublicationArbiter, or its test

**Priority (whenever / two weeks / yesterday)**: today - internal security product is I think unblocked, but this is a genuine bug.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/MetricPublicationArbiter.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/MetricPublicationArbiter.java
@@ -61,7 +61,7 @@ public class MetricPublicationArbiter implements Predicate<MetricName> {
     private static boolean safeShouldPublish(MetricName metricName, MetricPublicationFilter filter) {
         try {
             return filter.shouldPublish();
-        } catch (Exception e) {
+        } catch (RuntimeException e) {
             log.warn("Exception thrown when attempting to determine whether a metric {} should be published."
                             + " In this case we don't filter out the metric.",
                     SafeArg.of("metricName", metricName),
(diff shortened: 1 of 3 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/metrics/MetricPublicationArbiter.java
```diff
@@ -48,7 +54,19 @@ public void registerMetricsFilter(MetricName metricName, MetricPublicationFilter
...
-    private static boolean allFiltersMatch(List<MetricPublicationFilter> relevantFilters) {
-        return relevantFilters.stream().allMatch(MetricPublicationFilter::shouldPublish);
+    private static boolean allFiltersMatch(MetricName metricName, List<MetricPublicationFilter> relevantFilters) {
+        return relevantFilters.stream().allMatch(filter -> safeShouldPublish(metricName, filter));
+    }
+
+    private static boolean safeShouldPublish(MetricName metricName, MetricPublicationFilter filter) {
+        try {
+            return filter.shouldPublish();
+        } catch (Exception e) {
```
carterkozak, line 64: nit: Prefer catching RuntimeException over Exception to allow the type system to help if we ever add a checked exception to `shouldPublish`.
```suggestion
        } catch (RuntimeException e) {
```

jeremyk-91: you're right! good call

PR #4922, opened 2020-07-29
Remove local vs remote host selected metric

Jolyon-S:
**Goals (and why)**:
Our internal fleet has long since moved to using the local weighting of 1.0, so this metric generally just shows a straight line all of the time.

Removed counters recording when local or remote host was selected.

N/A.

Does anyone actually look at this metric, and/or is it ever useful?

`CassandraService`

Whenever

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -82,9 +82,6 @@ public class CassandraService implements AutoCloseable {
     private volatile Set<InetSocketAddress> localHosts = ImmutableSet.of();
     private final Supplier<Optional<HostLocation>> myLocationSupplier;
 
-    private final Counter randomHostsSelected;
-    private final Counter localHostsSelected;
-
     private final Random random = new Random();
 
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -93,10 +87,6 @@ public CassandraService(
             Blacklist blacklist,
             CassandraClientPoolMetrics poolMetrics) {
         this.metricsManager = metricsManager;
```
jkozlowski, line 89: I'm wondering if instead you could do @jeremyk-91 's filtering? So filter when this is 1, but publish when it's not?

Jolyon-S: I'm removing it more because I don't think I have ever used it after the initial rollout. So even if we filter it, would it be used in debugging?

jkozlowski: I've seen it go down and slowly recover before. But I think you might be right, there might not be any in keeping the complexity here

PR #4923, opened 2020-07-29
Bg gain leadership

sudiksha27:
**Goals (and why)**:

- For B/G upgrade to be effective, node which is on newest version of TimeLock should try to gain leadership

- LeaderPingResult accommodates response - remote is the leader but on an older version of TimeLock

- The local server treats this response as `pingReturnedFalse` and tries to gain leadership 

Tests modified

Should backoff before gaining leadership reduced for this case to be more aggressive?

- SingleLeaderPinger

- Tomorrow

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
@@ -90,6 +90,7 @@ class LeadershipEvents {
     }
 
     public void leaderOnOlderTimeLockVersion(OrderableSlsVersion version) {
+        // TODO(snanda): Kill log after few successful runs of blue-green deployment.
         leaderLog.info("We contacted the leader and it reported that it is on an older version of TimeLock - {}",
                 withContextArgs(SafeArg.of("version", version)));
         leaderElectionServiceMetrics.leaderOnOlderTimeLockVersion().mark();
--- a/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
@@ -23,6 +23,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.util.concurrent.RateLimiter;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.paxos.PaxosRoundFailureException;
 import com.palantir.paxos.PaxosValue;
@@ -37,6 +38,7 @@ class LeadershipEvents {
     private static final Logger leaderLog = LoggerFactory.getLogger(LEADER_LOG_NAME);
     private final Object[] contextArgs;
     private final LeaderElectionServiceMetrics leaderElectionServiceMetrics;
+    private final RateLimiter leaderOnOlderVersionLoggingRateLimiter = RateLimiter.create(0.03);
 
     LeadershipEvents(TaggedMetricRegistry metrics, List<SafeArg<String>> safeLoggingArgs) {
         leaderElectionServiceMetrics = LeaderElectionServiceMetrics.of(metrics);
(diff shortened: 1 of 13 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -84,14 +91,14 @@ public LeaderPingResult pingLeaderWithUuid(UUID uuid) {
...
+        MultiplexingCompletionService<LeaderPingerContext<PingableLeader>, PingResult> multiplexingCompletionService
                 = MultiplexingCompletionService.createFromCheckedExecutors(leaderPingExecutors);
 
         try {
-            multiplexingCompletionService.submit(leader, () -> leader.pinger().ping());
-            Future<Map.Entry<LeaderPingerContext<PingableLeader>, Boolean>> pingFuture = multiplexingCompletionService
-                    .poll(leaderPingResponseWait.toMillis(), TimeUnit.MILLISECONDS);
-            return getLeaderPingResult(uuid, pingFuture);
+            multiplexingCompletionService.submit(leader, () -> leader.pinger().pingV2());
+            Future<Map.Entry<LeaderPingerContext<PingableLeader>, PingResult>> pingFuture
```
sudiksha27, line 99: TimeLock dependency

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -88,6 +89,12 @@ void proposalFailure(PaxosRoundFailureException paxosException) {
         leaderElectionServiceMetrics.proposalFailure().mark();
     }
 
+    public void leaderOnOlderTimeLockVersion(OrderableSlsVersion version) {
+        leaderLog.info("We contacted the leader and it reported that it is on an older version of TimeLock - {}",
+                withContextArgs(SafeArg.of("version", version)));
```
sudiksha27, line 94: Comment - Kill log after _some_ successful runs of B/G

leader-election-impl/src/main/java/com/palantir/leader/LeadershipEvents.java
```diff
@@ -88,6 +89,13 @@ void proposalFailure(PaxosRoundFailureException paxosException) {
         leaderElectionServiceMetrics.proposalFailure().mark();
     }
 
+    public void leaderOnOlderTimeLockVersion(OrderableSlsVersion version) {
+        // TODO(snanda): Kill log after few successful runs of blue-green deployment.
+        leaderLog.info("We contacted the leader and it reported that it is on an older version of TimeLock - {}",
+                withContextArgs(SafeArg.of("version", version)));
+        leaderElectionServiceMetrics.leaderOnOlderTimeLockVersion().mark();
```
jeremyk-91, line 96: I'd suggest we rate limit the log message to maybe once every 10 or 30 seconds - timelocks otherwise ping each other 10 times per second meaning it'll be really noisy!

lock-api-objects/build.gradle
```diff
@@ -27,7 +27,7 @@ recommendedProductDependencies {
     productDependency {
         productGroup = 'com.palantir.timelock'
         productName = 'timelock-server'
-        minimumVersion = '0.218.0'
+        minimumVersion = '0.227.1'
```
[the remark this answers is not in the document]
jeremyk-91: I think we might need to use Schema Versions for this and/or have a fallback mechanism that hits pingv1 actually

jeremyk-91:
(Rate limiter looks good. We discussed the migration mechanisms offline.)

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -77,7 +77,7 @@
 @SuppressWarnings("checkstyle:FinalClass") // This is mocked internally
 public class TimeLockAgent {
     // Schema version from 2 onwards are on SQLite
-    private static final Long SCHEMA_VERSION = 3L;
+    private static final Long SCHEMA_VERSION = 4L;
```
jeremyk-91, line 80: This is not required.

PR #4928, opened 2020-08-04
Bufix/pds 129897

dxiao:
**Goals (and why)**:


dxiao:
Agreed, I'm also worried about the amount of metrics filters being instantiated, but I figure this is necessary to do regardless.

PR #4929, opened 2020-08-05
[PDS-128523] Fewer leader time calls.

felixdesouza:
**Goals (and why)**:
A different take on #4916. Instead of rate limiting calls everywhere, we're just going to ensure that the first batch is not validated.

Some further thought, I'll make a different issue. But why bother validating at each kvs batch read? Shouldn't we instead validate after every batch that's *actually* returned to the user i.e. after post filtering and filtering deletes? I don't have any metrics on this, but I imagine this will only affect places where TS is slow and deletes aren't removed, or (guessing here) there are a ton of concurrent writes from different transactions that get returned? Rationale being, if we're validating based on when the user read, we're reading from the kvs in multiple batches in one go to make one batch for a single user read, which feels like we should only validate for that single read. Thoughts?

* Extract the existing iterator into its own class so we can test it
* On the first batch (saved to a field) we skip validation
* We also make `BatchSizeIncreasingIterator` return whether or not a batch is the last.
   * Reason being, the previous implementation would take another iteration and another validation step to decide whether or not to terminate. i.e. if you had `n * BATCH_SIZE + k` columns where `k < n`, you would make `n + 1` calls as expected, and then you would make a further call to figure out that it's empty and that would incur a validation step. Rather than try to special case it, just make the page feel more like a batch iterator by giving some sort of `hasNext` functionality.

I've *just* tested this wrapped iterator, do you want something more?

Is coverage sufficient.

Anywhere. Half of it is tests however and there is a bit of bloat there.

ASAP, blocking internal product, and also violating SLOs of internal skiing product.

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/GetRowsColumnRangeIteratorTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/GetRowsColumnRangeIteratorTest.java
@@ -1,0 +1,156 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.transaction.impl;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/transaction/impl/GetRowsColumnRangeIteratorTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .putAll(firstBatchBarOne)
+                .put(lastInFirstBatch)
+                .put(firstInSecondBatch)
+                .putAll(restOfSecondBatch)
+                .build();
+
+
+        assertThat(returnedEntries.keySet())
+                .as("we can read all that we wrote")
+                .isEqualTo(puts);
```
jeremyk-91, line 115: Might be good to check the ordering here (we can make a `SortedSet` copy of this - `Cell` compare is based on the byte order of rows and columns - and then build a List rather than a Map of returnedEntries and check the ordering)

jeremyk-91: maybe something like
```
        List<Cell> cellsReadInOrder = Lists.newArrayList();
        cellsReadInOrder.addAll(firstBatchBarOne.stream().map(Map.Entry::getKey).collect(Collectors.toList()));
        cellsReadInOrder.add(lastInFirstBatch.getKey());
        cellsReadInOrder.add(firstInSecondBatch.getKey());
        cellsReadInOrder.addAll(restOfSecondBatch.stream().map(Map.Entry::getKey).collect(Collectors.toList()));

        assertThat(cellsReadInOrder)
                .as("we can read all that we wrote")
                .hasSameElementsAs(puts);
        assertThat(cellsReadInOrder)
                .as("iterator returns cells back in sorted order")
                .isSorted();
```

changelog/@unreleased/pr-4929.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: There are now fewer `getLeaderTime` calls when calling `getRowsColumnRangeIterator`.
```
jeremyk-91, line 3: nit: when reading the iterator that is returned by `getRowsColumnRangeIterator`.

PR #4931, opened 2020-08-05
[PDS-129897] Straight Stopgap Revert

jeremyk-91:
**Goals (and why)**:
- I don't want to set the world on fire

- Revert the table level top-N filtering of metrics

- If this is not the leak, then we will still run into problems

**Where should we start reviewing?**: +1-1


jeremyk-91:
We decided we would go for the full solution and verify it #4930

PR #4933, opened 2020-08-07
Bound lock acquire timeout

sudiksha27:
**Goals (and why)**:
Bound lock acquire timeout by client read timeout because 1. the async request expires before server cancels the request thus avoiding `java.util.concurrent.CancellationException` and 2. this ensures locks are not given out to phantom client.

BELS limits lock acquire timeout to client read timeout via duration limiter.

TBD

- If extended timeout is not used, the client read timeout will be - `ClientOptionsConstants.SHORT_READ_TIMEOUT` and phantom client problem would not be solved.

- Maintaining two constants is not a good idea. 

BlockEnforcingLockService

Next week

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
@@ -46,6 +46,13 @@ import com.palantir.logsafe.exceptions.SafeIllegalStateException;
  * Fairness is admittedly compromised, but this is a closer approximation than the previous behaviour.
  */
 final class BlockEnforcingLockService {
+    /**
+     * Bound lock acquire timeout by client read timeout because 1. the async request expires before server cancels
+     * the request thus avoiding {@link java.util.concurrent.CancellationException} and
+     * 2. this ensures locks are not given out to phantom client.
+     */
+    private static final Duration MAX_PERMISSIBLE_LOCK_ACQUIRE_TIMEOUT = Duration.ofSeconds(65);
+
     private final NamespacedConjureTimelockService namespacedConjureTimelockService;
     private final RemoteTimeoutRetryer timeoutRetryer;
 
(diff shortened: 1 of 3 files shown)


lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -46,6 +46,13 @@
  * Fairness is admittedly compromised, but this is a closer approximation than the previous behaviour.
  */
 final class BlockEnforcingLockService {
+    /**
+     * Bound lock acquire timeout by client read timeout because 1. the async request expires before server cancels
+     * the request thus avoiding {@link java.util.concurrent.CancellationException} and
+     * 2. this ensures locks are not given out to phantom client.
+     */
+    private static final Duration MAX_PERMISSIBLE_LOCK_ACQUIRE_TIMEOUT = Duration.ofSeconds(65);
```
jeremyk-91, line 54: This needs to be slightly less - a client's connect timeout also covers the time timelock takes to deserialize the request, check it is the leader etc.

lock-api/src/main/java/com/palantir/lock/client/BlockEnforcingLockService.java
```diff
@@ -46,6 +46,13 @@
...
+
```
sudiksha27, line 55: From Jeremy - exactly 65 won’t quite work, because a client’s connect timeout will also cover the time timelock takes to deserialise the request, check it is the leader

lock-api/src/test/java/com/palantir/lock/client/LockLeaseServiceTest.java
```diff
@@ -141,6 +148,33 @@ public void leasedTokenShouldHaveValidLeaseForTheLeasePeriod() {
...
+    public void lockAcquireTimeoutIsBoundedAndRequestRetried() {
+        when(lockRequest.getAcquireTimeoutMs()).thenReturn(TIMEOUT_GREATER_THAN_MAX_PERMISSIBLE_TIMEOUT.toMillis());
+        when(timelock.lock(any()))
+                .thenThrow(TIMEOUT_EXCEPTION)
+                .thenReturn(ConjureLockResponse.successful(SuccessfulLockResponse.of(LOCK_TOKEN, getLease())));
+
+        LockResponse lockResponse = lockLeaseService.lock(lockRequest);
+        assertValid(lockResponse.getToken());
+        verify(timelock, times(2)).lock(any());
+    }
```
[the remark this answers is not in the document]
jeremyk-91: Discussed offline: That's hard because it requires you to inject a clock.

Issue #4927, opened 2020-08-03
Duplicate TimestampService metrics


pkoenig10:
Atlas creates duplicate `TimestampService` metrics which causes problems with the `DisjointUnionTaggedMetricSet`. This error occurs on 0.239.10, even after https://github.com/palantir/atlasdb/pull/4921.
```
java.lang.IllegalArgumentException: Multiple entries with same key: MetricName{safeName=com.palantir.timestamp.TimestampService.getFreshTimestamp, safeTags={}}=com.codahale.metrics.Timer@460c003c and MetricName{safeName=com.palantir.timestamp.TimestampService.getFreshTimestamp, safeTags={}}=com.codahale.metrics.Timer@110af4b7
	at com.google.common.collect.ImmutableMap.conflictException(ImmutableMap.java:215)
	at com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:209)
	at com.google.common.collect.RegularImmutableMap.checkNoConflictInKeyBucket(RegularImmutableMap.java:146)
	at com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:109)
	at com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:394)
	at com.palantir.atlasdb.metrics.DisjointUnionTaggedMetricSet.getMetrics(DisjointUnionTaggedMetricSet.java:44)
	at com.palantir.atlasdb.metrics.FilteredTaggedMetricSet.getMetrics(FilteredTaggedMetricSet.java:47)
	at com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry.lambda$getMetrics$5(AbstractTaggedMetricRegistry.java:181)
	at java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1603)
	at com.palantir.tritium.metrics.registry.AbstractTaggedMetricRegistry.getMetrics(AbstractTaggedMetricRegistry.java:181)
```

I presume these duplicate metrics are because we are instrument `TimestampService` in multiple places. It appears that we instrument `TimestampService` at:
https://github.com/palantir/atlasdb/blob/3290b9f6b2ae720f1d2733bb65c6d72c025d103d/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L1198-L1201
[...]

PR #4938, opened 2020-08-11
Fix duplicate service instrumentation

pkoenig10:
**Goals (and why)**:
Follow up to https://github.com/palantir/atlasdb/pull/4935, fixing the root cause described in https://github.com/palantir/atlasdb/issues/4927.

Currently some service interfaces are instrumented twice. This can then cause problems due to duplicate metrics in the `MetricRegistry` and `TaggedMetricRegistry`.

This change removes the duplicate instrumentation and switches to using the `TaggedMetricRegistry` in `ServiceCreator#instrumentService`.


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceCreator.java
```diff
@@ -100,12 +101,11 @@ public static ServiceCreator withPayloadLimiter(
                 parameters);
     }
 
-    public static <T> T instrumentService(MetricRegistry metricRegistry, T service, Class<T> serviceClass) {
-        return AtlasDbMetrics.instrument(
-                metricRegistry,
+    public static <T> T instrumentService(TaggedMetricRegistry taggedMetrics, T service, Class<T> serviceClass) {
```
pkoenig10, line 104: Is this an acceptable break?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1187,26 +1187,13 @@ private static LockAndTimestampServices createRawLeaderServices(
...
-                TimestampService.class);
-
-        TimestampManagementService localManagement = ServiceCreator.instrumentService(
-                metricsManager.getRegistry(),
-                managedTimestampProxy,
-                TimestampManagementService.class);
         env.accept(localLock);
-        env.accept(localTime);
-        env.accept(localManagement);
+        env.accept(localManagedTimestamp);
```
jeremyk-91, line 1196: I'll take this from here. We can't really do this because the JAX-RS semantics for annotation inheritance are different.

pkoenig10: Ah that explains it, thanks!

PR #4940, opened 2020-08-12
Merge some per-txn-manager thread pools into shared thread pools

j-baker:
As of now, the thread pools I merge are responsible for around 480
threads on production stacks I've looked at. Each of them is doing some
menial task that probably doesn't need to be done in parallel, but needs
scheduling.

Shared scheduled thread pools are hard to get right, so instead we just
share between tasks with the same type. We therefore take 8 thread pools
(480 threads) and replace with 8 threads.

Given the durations of tasks, this shouldn't affect top end performance
a whole lot.


lock-api/src/main/java/com/palantir/lock/client/LockRefreshingLockService.java
```diff
@@ -42,16 +43,18 @@
...
     final Set<LockRefreshToken> toRefresh;
-    final ScheduledExecutorService exec;
     final long refreshFrequencyMillis = 5000;
+    ScheduledFuture<?> task;
     volatile boolean isClosed = false;
 
     public static LockRefreshingLockService create(LockService delegate) {
         final LockRefreshingLockService ret = new LockRefreshingLockService(delegate);
-        ret.exec.scheduleWithFixedDelay(() -> {
+        task = executor.scheduleWithFixedDelay(() -> {
```
carterkozak, line 57:
```suggestion
        ret.task = executor.scheduleWithFixedDelay(() -> {
```

gmaretic (approved):
Looks good!

jeremyk-91 (approved):
Grgur's changes were only a develop merge.

PR #4941, opened 2020-08-12
Much less ram used by disruptors

j-baker:
Disruptors at the moment hold their arguments and results for potentially far longer
than their expected lifetime. We use large buffers and so likely 16k copies of the arg
and result are held, per autobatcher (internally we might have 300, or 5M copies of the results, spans, etc).

### DIFF
--- a/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
+++ b/atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
@@ -63,14 +63,14 @@ public final class DisruptorAutobatcher<T, R>
         return new NamedThreadFactory("autobatcher." + safeLoggablePurpose, true);
     }
 
-    private final Disruptor<DefaultBatchElement<T, R>> disruptor;
-    private final RingBuffer<DefaultBatchElement<T, R>> buffer;
+    private final Disruptor<DisruptorBatchElement<T, R>> disruptor;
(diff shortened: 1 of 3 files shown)


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/DisruptorAutobatcher.java
```diff
@@ -159,9 +160,9 @@ public boolean setFuture(ListenableFuture<? extends R> future) {
             EventHandler<BatchElement<T, R>> eventHandler,
             int bufferSize,
             String safeLoggablePurpose) {
-        Disruptor<DefaultBatchElement<T, R>> disruptor =
-                new Disruptor<>(DefaultBatchElement::new, bufferSize, threadFactory(safeLoggablePurpose));
-        disruptor.handleEventsWith(eventHandler);
+        Disruptor<DisruptorBatchElement<T, R>> disruptor =
+                new Disruptor<>(DisruptorBatchElement::new, bufferSize, threadFactory(safeLoggablePurpose));
+        disruptor.handleEventsWith((event, sequence, endOfBatch) -> eventHandler.onEvent(event.consume(), sequence, endOfBatch));
```
jeremyk-91, line 165: nit: this has 129 chars - quick refactor please

jeremyk-91: (I'm going to do this, so don't do this one if you see it!)

PR #4948, opened 2020-08-14
[PDS-130107] Enable Divergent KVS Namespaces and TimeLock Clients

jeremyk-91:
**Goals (and why)**:
- Used in some DBKVS use cases, where through the use of table mapping it is legitimate to have multiple AtlasDB users interface with the same KVS in the same way.
- Minimise the risk that anyone who does not need this use case actually tries to (this is very error prone).

- Allow the config to accept different KVS namespaces and timelock clients, provided this is explicitly enabled AND the user is not using Cassandra.

- Added a few new tests for new cases in the configuration

- Are there correctness issues?

paired

**Priority (whenever / two weeks / yesterday)**: this week or early next

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
@@ -271,11 +271,25 @@ public abstract class AtlasDbConfig {
      */
     public abstract Optional<TimestampCache> timestampCache();
 
+    /**
+     * If set, allows the timelock and key-value-service namespaces to be different.
+     *
+     * Note that this only works on DBKvs. If using Cassandra, this behaviour is strictly not allowed. Please contact
+     * the AtlasDB team if this is an issue.
+     *
+     * You should only be using this if you know what you are doing. Using this flag where it is not necessary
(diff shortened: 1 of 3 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/config/AtlasDbConfig.java
```diff
@@ -319,69 +333,56 @@ private static void checkServersListHasAtLeastOneServerIfPresent(Optional<Server
...
-            if (timelock().isPresent()) {
-                TimeLockClientConfig timeLockConfig = timelock().get();
+        if (timelock().isPresent()) {
+            TimeLockClientConfig timeLockConfig = timelock().get();
 
-                com.palantir.logsafe.Preconditions.checkState(timeLockConfig.client().isPresent(),
-                        "Either the atlas root-level namespace config or the TimeLock client config"
-                                + " should be present.");
+            com.palantir.logsafe.Preconditions.checkState(timeLockConfig.client().isPresent(),
+                    "Either the atlas root-level namespace config or the TimeLock client config should be present.");
```
sudiksha27, line 373: nit: remove `com.palantir.logsafe`?

PR #4949, opened 2020-08-14
Relieve Flakes - Make Stress Test Run Slightly Longer

jeremyk-91:
**Goals (and why)**:
- The stress test started flaking a lot more, probably because our containers might be a bit weaker here and the theoretical permitted bound of thread blow-up with a bad node (770 if I'm not mistaken) is very close to our 800 limit.

- Increase the allowance to 1,000.
- Add a few more requests on the tail end of the paxos test, so that an implementation that spins up unbounded threads talking to the bad node will still fail.

- This PR is about tests.

- Please validate the test still retains discriminatory power.

- the test

**Priority (whenever / two weeks / yesterday)**: yesterday 🔥  rate of build flakes now is wild

### DIFF
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -455,9 +455,7 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
     }
 
     private static void assertNumberOfThreadsReasonable(int startingThreads, int threadCount, boolean nonLeaderDown) {
-        // TODO (jkong): Lower the amount over the threshold. This needs to be slightly higher for now because of the
-        // current threading model in batch mode, where separate threads may be spun up on the autobatcher.
-        int threadLimit = startingThreads + 800;
+        int threadLimit = startingThreads + 1000;
         if (nonLeaderDown) {
             if (threadCount > threadLimit) {
                 System.out.println("hello");
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -437,7 +437,7 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
         int startingNumThreads = ManagementFactory.getThreadMXBean().getThreadCount();
         boolean isNonLeaderTakenOut = false;
         try {
-            for (int i = 0; i < 1_500; i++) { // Needed as it takes a while for the thread buildup to occur
+            for (int i = 0; i < 1_800; i++) { // Needed as it takes a while for the thread buildup to occur
                 assertNumberOfThreadsReasonable(
                         startingNumThreads,
                         ManagementFactory.getThreadMXBean().getThreadCount(),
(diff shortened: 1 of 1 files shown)


timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -455,9 +455,7 @@ public void stressTestForPaxosEndpoints() {
     }
 
     private static void assertNumberOfThreadsReasonable(int startingThreads, int threadCount, boolean nonLeaderDown) {
-        // TODO (jkong): Lower the amount over the threshold. This needs to be slightly higher for now because of the
-        // current threading model in batch mode, where separate threads may be spun up on the autobatcher.
-        int threadLimit = startingThreads + 800;
+        int threadLimit = startingThreads + 1000;
         if (nonLeaderDown) {
             if (threadCount > threadLimit) {
```
Jolyon-S, line 460: Drive by comment here, but do we really need this if statement?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -437,7 +437,7 @@ public void stressTestForPaxosEndpoints() {
         int startingNumThreads = ManagementFactory.getThreadMXBean().getThreadCount();
         boolean isNonLeaderTakenOut = false;
         try {
-            for (int i = 0; i < 1_500; i++) { // Needed as it takes a while for the thread buildup to occur
+            for (int i = 0; i < 1_800; i++) { // Needed as it takes a while for the thread buildup to occur
```
Jolyon-S, line 440: If I understand this test correctly:
1. This number (`1_800`) represents how long we are running for - the longer, the more accurate the test is (i.e. if we were to run for 100k and still never exceeded the thread limit we'd know it is unlikely to ever go over that amount).
2. The number that you increased below (`+ 800` -> `+ 1000`) is the tolerance - by increasing that you make the tests more tolerant of bad thread buildup, although only by a small amount.

So presumably the concern is that this PR relaxes the test a bit (to make it pass on Circle, sure, but even so). I'm happy with this as it is - 1) should help counteract 2) - but if you wanted to err on the side of caution, increasing 1) further would improve the test (at the cost of time, naturally).

Thoughts?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: We do not. Look at line 463 😅  This must have been an artifact of when I did some previous testing

Jolyon-S: yep, that's what I was looking at ^^. I'm guessing this was for a breakpoint.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: Yep, your understanding is correct. Here, `1_800` is the number of requests, and `1_000` is the number of threads over the amount at startup that are allowed.

We do expect `(384 + 1) * 2 = 770` threads to be blocked on talking to the dying timelock node. However, there is probably enough variation in other threads that tolerating noise of `800 - 770 = 30` is insufficient (this includes any threads that may be spun up by the JVM or by our application code for whatever reason).

What's important here is that we don't want the number of threads blocked to scale with the number of requests to the dead node. Thus, strictly speaking as a minimum bound you want the number of requests made whilst one node is unresponsive to be at least _half_ of the number of spare threads allowed (the reason for half is because Dialogue spins up a thread as well). So in theory we should have at least `800 / 2 = 400` and now `1000 / 2 = 500` as a minimum, and once you add the offset from requests made before a node crashes we're looking at `900` and `1_100` as theoretical minimum bounds. We want more though to ensure that if 'noise' from other threads being shut down actually makes a _negative_ contribution, the test still doesn't really have a chance of passing.

Jolyon-S: > What's important here is that we don't want the number of threads blocked to scale with the number of requests to the dead node.

Yes, that's the main thing, and my main original concern was that increasing the tolerance by 200 but increasing the requests by 300 may or may not be enough, but I think as you have reasoned out above, it is sufficient.

Looks good to me then, merge away!

PR #4953, opened 2020-08-18
Pingv2 fallback

sudiksha27:
**Goals (and why)**:
Fallback to ping if pingV2 throws, this is being done to support services where there is no concept of schema.

- Hit ping if pingV2 throws.

- Added new tests

- Hitting ping when pingV2 throws with status other than 404.

Small

By end of week

### DIFF
--- a/atlasdb-config/src/test/java/com/palantir/atlasdb/SinglePingableLeaderTest.java
+++ b/atlasdb-config/src/test/java/com/palantir/atlasdb/SinglePingableLeaderTest.java
@@ -91,8 +91,8 @@ public class SinglePingableLeaderTest {
         availableServer.stubFor(PING_V2_MAPPING.willReturn(WireMock.aResponse().withStatus(200)
                 .withBody("{\"isLeader\":false}")));
         pingerWithVersion(OrderableSlsVersion.valueOf("1.1.1")).pingLeaderWithUuid(REMOTE_UUID);
-        assertLegacyPingRequestsMade(1);
-        assertPingV2RequestsMadeAreAtMost(0);
+        assertPingV2RequestsMadeAreExactly(1);
+        assertLegacyPingRequestsMade(0);
     }
 
 
@@ -117,19 +117,22 @@ public class SinglePingableLeaderTest {
     public void verifyPingRequests(int totalNumOfPings, int verifyLegacyPingRequests) {
         LeaderPinger pinger = pingerWithVersion(OrderableSlsVersion.valueOf("1.1.1"));
         IntStream.range(0, totalNumOfPings).forEach(idx -> pinger.pingLeaderWithUuid(REMOTE_UUID));
-
-        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
-        assertThat(requests.size()).isLessThan(totalNumOfPings);
-        assertPingV2RequestsMadeAreAtMost(verifyLegacyPingRequests);
+        assertPingV2RequestsMadeAreAtMost(totalNumOfPings);
+        assertLegacyPingRequestsMade(verifyLegacyPingRequests);
     }
 
     public void assertPingV2RequestsMadeAreAtMost(int count) {
-        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING)));
+        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
+        assertThat(requests.size()).isLessThan(count);
+    }
+
+    public void assertPingV2RequestsMadeAreExactly(int count) {
+        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
         assertThat(requests.size()).isEqualTo(count);
     }
 
     public void assertLegacyPingRequestsMade(int count) {
-        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
+        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING)));
         assertThat(requests.size()).isEqualTo(count);
     }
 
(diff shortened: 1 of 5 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -88,20 +89,48 @@ public LeaderPingResult pingLeaderWithUuid(UUID uuid) {
...
+            LeaderPingerContext<PingableLeader> leader,
+            Callable<PingResult> pingEndpoint) throws ExecutionException {
         try {
-            multiplexingCompletionService.submit(leader, () -> leader.pinger().pingV2());
+            multiplexingCompletionService.submit(leader, pingEndpoint);
             Future<Map.Entry<LeaderPingerContext<PingableLeader>, PingResult>> pingFuture
                     = multiplexingCompletionService.poll(leaderPingResponseWait.toMillis(), TimeUnit.MILLISECONDS);
             return getLeaderPingResult(uuid, pingFuture, timeLockVersion);
-        } catch (InterruptedException e) {
+        } catch (InterruptedException exc) {
```
jeremyk-91, line 131: nit: is there a reason we can't use `e`?

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -88,20 +89,48 @@ public LeaderPingResult pingLeaderWithUuid(UUID uuid) {
...
+        }
+
+        try {
+            return actuallyPingLeaderWithUuid(multiplexingCompletionService,
+                    uuid,
+                    leader,
+                    () -> getPingResultFromLegacyEndpoint(leader));
+        } catch (ExecutionException ex) {
+            return LeaderPingResults.pingCallFailure(ex.getCause());
+        }
```
jeremyk-91, line 112: My main concern with this is that for users who may not have access to ping v2, this is going to make performance of ping worse by a full RPC. It's not really the hottest codepath so this may not necessarily be terrible, but is there something we can do to avoid repeatedly trying if there is no ping v2?

jeremyk-91 (commented):
This broadly looks correct, but I'd be a bit worried for the performance of certain legacy use cases.

sudiksha27:
Validated on internal test stack.

atlasdb-config/src/test/java/com/palantir/atlasdb/SinglePingableLeaderTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Test
+    public void doesNotFallbackOnPingWhenPingV2Responds() {
+        availableServer.stubFor(PING_V2_MAPPING.willReturn(WireMock.aResponse().withStatus(200)
+                .withBody("{\"isLeader\":false}")));
+        pingerWithVersion(OrderableSlsVersion.valueOf("1.1.1")).pingLeaderWithUuid(REMOTE_UUID);
+        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
+        assertThat(requests.size()).isEqualTo(1);
+        requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING)));
```
jeremyk-91, line 96: Let's not reuse the set, for readability.

atlasdb-config/src/test/java/com/palantir/atlasdb/SinglePingableLeaderTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public void verifyPingRequests(int totalNumOfPings, int verifyLegacyPingRequests) {
+        LeaderPinger pinger = pingerWithVersion(OrderableSlsVersion.valueOf("1.1.1"));
+        IntStream.range(0, totalNumOfPings).forEach(idx -> pinger.pingLeaderWithUuid(REMOTE_UUID));
+
+        List<LoggedRequest> requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING_V2)));
+        assertThat(requests.size()).isLessThan(totalNumOfPings);
+        requests = WireMock.findAll(WireMock.getRequestedFor(WireMock.urlMatching(PING)));
+        assertThat(requests.size()).isEqualTo(verifyLegacyPingRequests);
```
jeremyk-91, line 125: I would suggest having this be two methods: one for `assertLegacyPingRequestsMade(int)` and one for `assertPingV2RequestsMadeAreAtMost(int)`. The 2 numbers is a bit messy to read.

atlasdb-config/src/test/java/com/palantir/atlasdb/SinglePingableLeaderTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Test
+    public void fallbackOnPingWhenPingV2DoesNotExist() {
+        availableServer.stubFor(PING_V2_MAPPING.willReturn(WireMock.aResponse().withStatus(404)));
+        verifyPingRequests(5, 5);
+    }
+
+    @Test
+    public void fallbackOnPingWhenPingV2Throws() {
```
jeremyk-91, line 107: Unclear if this is something we want to specify in the tests. I guess it's not wrong.

leader-election-impl/src/main/java/com/palantir/paxos/SingleLeaderPinger.java
```diff
@@ -46,15 +49,17 @@
 import com.palantir.sls.versions.VersionComparator;
 
 public class SingleLeaderPinger implements LeaderPinger {
-
     private static final Logger log = LoggerFactory.getLogger(SingleLeaderPinger.class);
+    private static final RateLimiter pingV2RateLimiter = RateLimiter.create(1.0 / (5 * 60));
```
jeremyk-91, line 53: Let's not use statics - this should likely be injected into each SingleLeaderPinger.

This is "once every 5 minutes" - could you include a comment to that effect, and/or extract a constant explaining what this value is?

jeremyk-91 (approved):
👍  This looks reasonable.

sudiksha27:
Validated on internal test stack after final changes.

PR #4954, opened 2020-08-18
Don't Register Duplicate Metrics

jeremyk-91:
Based on @pkoenig10's #4938 but fixed to handle annoying JAX-RS semantics correctly.

**Goals (and why)**:
- See #4938

- Don't wire proxies through ServiceCreator for embedded leader block
- But open TimestampService and TimestampManagementService facades so that registration works properly.

- Existing tests should suffice for correctness.
- Removal of metrics is verified by removing the test involved

Nothing in particular

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -1265,10 +1265,12 @@ public abstract class TransactionManagers {
             // remote services are pointed at them anyway.
             LockService dynamicLockService = LocalOrRemoteProxy.newProxyInstance(
                     LockService.class, localLock, remoteLock, useLocalServicesFuture);
+
+            // Use managedTimestampProxy here to avoid local calls going through indirection.
             TimestampService dynamicTimeService = LocalOrRemoteProxy.newProxyInstance(
-                    TimestampService.class, localTime, remoteTime, useLocalServicesFuture);
+                    TimestampService.class, managedTimestampProxy, remoteTime, useLocalServicesFuture);
             TimestampManagementService dynamicManagementService = LocalOrRemoteProxy.newProxyInstance(
-                    TimestampManagementService.class, localManagement, remoteManagement, useLocalServicesFuture);
+                    TimestampManagementService.class, managedTimestampProxy, remoteManagement, useLocalServicesFuture);
             return ImmutableLockAndTimestampServices.builder()
                     .lock(dynamicLockService)
                     .timestamp(dynamicTimeService)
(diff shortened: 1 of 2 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1187,23 +1188,16 @@ private static LockAndTimestampServices createRawLeaderServices(
...
 
         ManagedTimestampService managedTimestampProxy =
                 AwaitingLeadershipProxy.newProxyInstance(ManagedTimestampService.class, time::get, leader);
 
-        TimestampService localTime = ServiceCreator.instrumentService(
-                metricsManager.getRegistry(),
-                managedTimestampProxy,
-                TimestampService.class);
+        // These facades are necessary because of the semantics of the JAX-RS algorithm (in particular, accepting
+        // just the managed timestamp service will *not* work.
```
Jolyon-S, line 1197: nit: close bracket (that was opened on the line above)

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
```diff
@@ -612,15 +612,6 @@ private TransactionManagers withLockImmutableTsOnReadOnlyTransaction(boolean opt
                 .build();
     }
 
-    @Test
-    public void metricsAreReportedExactlyOnceWhenUsingLocalService() throws IOException {
```
Jolyon-S, line 616: You've removed two tests, but added none - are there any tests we can add to have confidence that this fix is correct?

Jolyon-S: I see you've said
>Existing tests should suffice for correctness.
Removal of metrics is verified by removing the test involved

but I'm not 100% on the second point - removing the tests just removes the check that there are metrics, but surely it doesn't actively check that _no_ metrics are registered? That's what we want here, right?

Jolyon-S (commented):
Just a couple of comments on things here or there.

I haven't dug in deep into the semantics of the JAX-RS algorithm or the facade change - I can do that in a second pass, or maybe @sudiksha27 you want to have a look at that part?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: good spot!

atlasdb-config/src/test/java/com/palantir/atlasdb/factory/TransactionManagersTest.java
jeremyk-91: Fair enough, yep I'll replace that with a check that they are not registered.

sudiksha27 (approved):
LGTM!

PR #4955, opened 2020-08-20
memoize ssl socket factory creation

j-baker:
each ssl socket factory holds a presized 16k entry ssl session cache.
At present AtlasDB-proxy creates a new ssl socket factory per-host,
per-namespace. These really should all be the same ssl socket factory,
since the ssl configuration is the same.

We now memoize these. The config point is not the right place to do
this, since internally we mutate the configs to add extra options in a
few places - can't work via config equality.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
@@ -18,10 +18,8 @@ package com.palantir.atlasdb.keyvalue.cassandra;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.SocketException;
-import java.security.NoSuchAlgorithmException;
 import java.util.Map;
 
-import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocket;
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
```diff
@@ -41,18 +39,24 @@
...
 import com.palantir.common.exception.AtlasDbDependencyException;
+import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
 import com.palantir.conjure.java.config.ssl.SslSocketFactories;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 
 public class CassandraClientFactory extends BasePooledObjectFactory<CassandraClient> {
     private static final Logger log = LoggerFactory.getLogger(CassandraClientFactory.class);
+    private static final LoadingCache<SslConfiguration, SSLSocketFactory> sslSocketFactoryCache = Caffeine.newBuilder()
+            .weakValues()
```
jeremyk-91, line 58: Nice. I checked that we retain a reference to the SSL socket factory if you're using a TM with Cassandra. So we do actually want to clear this out quickly once unreferenced

jeremyk-91 (approved):

PR #4956, opened 2020-08-20
Intern the token map

j-baker:
This is a very large object and it's duplicated (and computed
separately) per Cassandra keyspace, for a total of
(256 * say 9 hosts * num datastores = O(100k entries) * (size of token = 32 + range map overhead = 8 + 16 + 16 + 16 =
~100) = 16MB heap which is proportional to the number of datastores. Since it's a simple fix, we might as well.


jeremyk-91 (approved):
Yep, this seems reasonable. Thanks!

PR #4959, opened 2020-08-21
[LW] Fixes, 3: Additional logging towards fixing obscure bug

Jolyon-S:
**Goals (and why)**:
In production, there were gaps here, but no logging as to why. Ideally, I'll work this out without having to keep trying on production, but may as well have this logging too.

Add safearg logging

Tests exist for this, but probably not at a full integration level.

These args are safe, right?

`LockWatchEventLog`

Soon, please!

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
@@ -27,6 +27,7 @@ import com.palantir.lock.watch.LockWatchCreatedEvent;
 import com.palantir.lock.watch.LockWatchEvent;
 import com.palantir.lock.watch.LockWatchStateUpdate;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 
 final class LockWatchEventLog {
     private final ClientLockWatchSnapshot snapshot;
@@ -152,7 +153,9 @@ final class LockWatchEventLog {
             Preconditions.checkNotNull(firstEvent, "First element not preset in list of events");
             Preconditions.checkArgument(firstEvent.sequence() < latestVersion.get().version()
                             || latestVersion.get().version() + 1 == firstEvent.sequence(),
-                    "Events missing between last snapshot and this batch of events");
+                    "Events missing between last snapshot and this batch of events",
+                    SafeArg.of("latestVersionSequence", latestVersion.get().version()),
+                    SafeArg.of("firstNewVersionSequence", firstEvent.sequence()));
         }
     }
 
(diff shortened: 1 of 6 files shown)


lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -163,12 +172,36 @@ public void close() {
...
+                Optional<IdentifiedVersion> requestedVersion = lockWatchEventCache.lastKnownVersion();
                 ConjureStartTransactionsResponse response = lockLeaseService.startTransactionsWithWatches(
-                        lockWatchEventCache.lastKnownVersion(), numberOfTransactions - result.size());
+                        requestedVersion, numberOfTransactions - result.size());
                 lockWatchEventCache.processStartTransactionsUpdate(
                         response.getTimestamps().stream().boxed().collect(Collectors.toSet()),
                         response.getLockWatchUpdate());
                 result.addAll(split(response));
+
+                if (log.isDebugEnabled()) {
```
jkozlowski, line 183: I think you need the same change in CommitTimestampGetter?

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -163,12 +172,36 @@ public void close() {
...
+
+                                @Override
+                                public Optional<LockWatchStateUpdate.Success> visit(
+                                        LockWatchStateUpdate.Snapshot snapshot) {
+                                    return Optional.empty();
+                                }
+                            });
+
+                    successfulUpdate.ifPresent(success ->
+                            log.debug("Start transaction batch request and response",
```
jkozlowski, line 200: Same here, I think you want unsafe arg for the events?

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
jkozlowski: And probs extract a small method?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -152,7 +153,9 @@ private void assertEventsAreContiguousAndNoEventsMissing(List<LockWatchEvent> ev
             Preconditions.checkNotNull(firstEvent, "First element not preset in list of events");
             Preconditions.checkArgument(firstEvent.sequence() < latestVersion.get().version()
                             || latestVersion.get().version() + 1 == firstEvent.sequence(),
-                    "Events missing between last snapshot and this batch of events");
+                    "Events missing between last snapshot and this batch of events",
+                    SafeArg.of("latestVersionSequence", latestVersion.get().version()),
+                    SafeArg.of("firstNewVersionSequence", firstEvent.sequence()));
```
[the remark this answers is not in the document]
jkozlowski: Actually, scrap that, dunno if we want actual data lying around in the logs.

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
jkozlowski: Same as above.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -298,7 +298,7 @@ public void missedEventThrows() {
         assertThatThrownBy(() -> eventCache.processStartTransactionsUpdate(TIMESTAMPS_2,
```
jkozlowski, line 298: This should be:

```
assertThatLoggableExceptionThrownBy(() -> eventCache.processStartTransactionsUpdate(TIMESTAMPS_2,
                LockWatchStateUpdate.success(LEADER, 5L, ImmutableList.of(UNLOCK_EVENT))))
                .isExactlyInstanceOf(SafeIllegalArgumentException.class)
                .hasLogMessage("Events missing between last snapshot and this batch of events");
```

jkozlowski (approved):

PR #4961, opened 2020-08-24
[LW] Fixes, 4: Register lockwatches eagerly

Jolyon-S:
**Goals (and why)**:
* Register locks immediately, _as well as_ periodically. This not only allows the benefits to come through faster, but also makes testing easier.

* Calls the internal register method when the client registers.

N/A

* Any hidden issues here?

`LockWatchManagerImpl`

This week, please


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -72,6 +72,7 @@ public void close() {
     @Override
     public void registerWatches(Set<LockWatchReferences.LockWatchReference> newLockWatches) {
         lockWatchReferences.addAll(newLockWatches);
+        registerWatchesWithTimelock();
```
jkozlowski, line 75: I think this works, everything is nice and threadsafe anyway.

jkozlowski (approved):

PR #4962, opened 2020-08-24
[LW] Fixes, 5: Throw underlying exception if already on fallback cache

Jolyon-S:
**Goals (and why)**:
* We don't want to throw retryable exceptions if the fallback cache is throwing.

* Add check that fallback is equal to delegate.

* Added new test

* This shouldn't really happen since the fallback is no op, but I think this is correct anyway (saw this on a test in atlasdb-proxy on that one and wanted to replicate here).

`ResilientLockWatchEventCache`

This week

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
@@ -74,7 +74,7 @@ final class ResilientLockWatchEventCache extends AbstractInvocationHandler {
         } catch (TransactionLockWatchFailedException e) {
             throw e;
         } catch (Throwable t) {
-            if (delegate.equals(fallbackCache)) {
+            if (delegate == fallbackCache) {
                 throw new RuntimeException(t);
             } else {
                 log.warn("Unexpected failure occurred when trying to use the default cache. "
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
@@ -74,11 +74,15 @@ final class ResilientLockWatchEventCache extends AbstractInvocationHandler {
         } catch (TransactionLockWatchFailedException e) {
             throw e;
         } catch (Throwable t) {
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
```diff
@@ -74,11 +74,15 @@ synchronized RuntimeException handleException(InvocationTargetException rethrow)
         } catch (TransactionLockWatchFailedException e) {
             throw e;
         } catch (Throwable t) {
-            log.warn("Unexpected failure occurred when trying to use the default cache. Switching to the fallback "
-                    + "implementation", t);
-            fallbackCacheSelectedCounter.inc();
-            delegate = fallbackCache;
-            throw new TransactionLockWatchFailedException("Unexpected failure in the lock watch cache", t);
+            if (delegate.equals(fallbackCache)) {
```
jeremyk-91, line 77: are you sure you want `equals` here and not `==`?

jeremyk-91 (approved):
Fair enough - not sure how you caught this, but looks reasonable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
Jolyon-S: yeah, good catch. `equals` here would mean we never bother actually falling back, but we probably want to do that even if this was configured with the same (but not same in memory) for each for some reason

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
```diff
@@ -74,11 +74,15 @@ synchronized RuntimeException handleException(InvocationTargetException rethrow)
         } catch (TransactionLockWatchFailedException e) {
             throw e;
         } catch (Throwable t) {
-            log.warn("Unexpected failure occurred when trying to use the default cache. Switching to the fallback "
-                    + "implementation", t);
-            fallbackCacheSelectedCounter.inc();
-            delegate = fallbackCache;
-            throw new TransactionLockWatchFailedException("Unexpected failure in the lock watch cache", t);
+            if (delegate == fallbackCache) {
+                throw new RuntimeException(t);
```
jkozlowski, line 78: Should've been SafeRuntimeException

PR #4968, opened 2020-08-28
Transaction#markTableInvolved.

jkozlowski:
**Goals (and why)**:

Sometimes tables will be read out of bounds (LW with tracking of reads external to atlasdb). In that situation we need to be able to not skip certain checks that are ordinarily done on reads.

Transaction#markTableInvolved effectively does first part of a read, without doing immutable timestamp lock checks. Those are only done IFF the transaction did no reads at all, so this feature is effectively free for transactions that did any reads.

Only in AbstractSerializableTransactionTest.java, because that's the test I have most experience with. I know there's SnapshotTransactionTest, which feels somewhat similar, so would be interested in feedback.

Should I add some more testing somewhere? I did not add a test for transaction timeouts simply because transactions do System.currentTimeMillis, and I didn't feel like refactoring that.


jkozlowski:
@jeremyk-91 so we're also be skipping SnapshotTransaction#getWithPostFilteringAsync in the case of LW caching (and specifically I'm wondering about SnapshotTransaction#getCommitTimestamps). I never spent a significant amount of time with this codepath, so would be interested in your thoughts.

jeremyk-91:
Alright. So for something like this we basically need to answer the question "what happens on a write transaction that only does reads: but these reads are not visible to the AtlasDB library because they are cached on the client side".

- When a read is performed, get preconditions are checked. Apart from the transaction lifecycle ones, the main effect of this is to note that we need to validate the immutable timestamp lock at the end. This would be addressed by a markTableRead() type of call. Also, we note that there are reads, which means we need to check the pre-commit condition.
- Then as you note there is the getting of commit timestamps, which blocks on inflight transactions to finish and possibly rolls them back if they haven't committed for long enough. However, if you're seeing one of these entries, they must already have acquired their locks (they need to acquire the lock to write to the DB). Also this lock is watched (otherwise you can't skip the read), so the commit update you get will necessarily have a lock acquisition or leader election (I think; @gmaretic to keep me honest?)
  - This could theoretically read to starvation of readers by a constant stream of writers (but given the use case of lock watches should be fine)

jkozlowski:
@jeremyk-91 yep; it's just that read-only transaction will not get to the commit stage, so that doesn't matter. So even if there are any concurrently executing transactions, that read-only transaction will commit. And I think that's exactly how normal atlas transactions that only do reads work.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1629,13 +1635,13 @@ private void checkConstraints() {
 
     private void commitWrites(TransactionService transactionService) {
         if (!hasWrites()) {
-            if (hasReads()) {
+            if (hasReads() || hasAnyInvolvedTables()) {
```
jkozlowski, line 1638: Actually this should take into account if they're serializable or not, gimme a sec...

jkozlowski: Fixed

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2384,13 +2386,22 @@ public void useTable(TableReference tableRef, ConstraintCheckable table) {
...
+    private boolean hasAnyInvolvedTables() {
+        return !involvedTables.isEmpty();
+    }
+
     private boolean validationNecessaryForInvolvedTablesOnCommit() {
-        return involvedTables.stream().anyMatch(this::isValidationNecessaryOnCommit);
+        boolean anyTableRequiresImmutableTimestampLocking = involvedTables.stream().anyMatch(
+                this::requiresImmutableTimestampLocking);
+        boolean needsToValidate = !validateLocksOnReads || !hasReads();
+        return anyTableRequiresImmutableTimestampLocking && needsToValidate;
```
jeremyk-91, line 2404: So we validate if there's an interesting table AND either there were no reads (because lock watches), or validate locks on reads was false.

There are potentially cases where this might behave differently from the original:

```
user reads table A, checks lock
<loses immutable timestamp lock>
user reads table B via notifications, marked B as used
txn.commit()
```

This would previously throw, but now reads the value from the cache, which is fine - the goal of the immutable timestamp lock is resilience against a delete clearing the value we want out from under us by Thorough Sweep.

Is the above understanding correct? May be good to have a quick check with @gmaretic here

changelog/@unreleased/pr-4968.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Transaction#markTableInvolved.
```
jeremyk-91, line 3: Might be good to have a bit more detail here (in particular, don't use this unless you really know what you're doing type of warnings)

jeremyk-91 (commented):
I think the semantics of this will work, though want to make sure my understanding is correct (see comment). Wiring and tests look reasonable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2384,13 +2386,22 @@ public void useTable(TableReference tableRef, ConstraintCheckable table) {
...
         involvedTables.add(tableRef);
     }
 
-    private boolean validationNecessaryForInvolvedTablesOnCommit() {
-        return involvedTables.stream().anyMatch(this::isValidationNecessaryOnCommit);
+    private boolean hasAnyInvolvedTables() {
+        return !involvedTables.isEmpty();
+    }
+
+    private boolean validationNecessaryForInvolvedTablesOnCommit(boolean hasReads) {
```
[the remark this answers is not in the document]
jeremyk-91: ah, you fixed this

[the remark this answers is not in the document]
jkozlowski: Yep, sorry checkstyle was complaining too...

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jkozlowski: Where would it throw? Let me clarify my understanding: I understand your timeline as "table A read checks lock and still has it, but lock expires just afterwards". With that:

Before my change the behavior would be to not throw: when A does the read, it checks the lock, then user does some reads from a cache (AtlasDB is not aware of those reads), and then user calls commit. Then ``#hasReads`` is true (because of the A read) so we recheck PreCommitCondition, but #validationNecessaryForInvolvedTablesOnCommit returns false, because #isValidationNecessaryOnCommit checks ``!validateLocksOnReads``. And ``validateLocksOnReads`` is a preeety deep rabbit hole, but if you follow it it ultimately is set from ``TransactionManagers#validateLocksOnReads`` where the default is true.

After my change, hasReads is true, so again we check PreCommitCondition, and then anyTableRequiresImmutableTimestampLocking is true, because I'm assuming that's what your example does, ``!validateLocksOnReads`` is false, but ``!hasReads()`` is true, so we do it. I guess I could add a recheck for ``!hasReads() && hasAnyInvolvedTables()`` just to be nice?

Ultimately, I think what I care about most is the PreCommitCondition check: the immutable timestamp lock, as you say, is just for TargetedSweep to not nuke the cells under us, but we kinda don't care (because we have them in-memory!). So if you think this is just too complicated I can ignore, but I think I'd rather have this behave roughly the same as the case of only cache reads (the only difference is that the immutable lock check will be checked at commit time).

changelog/@unreleased/pr-4968.v2.yml
jkozlowski: Updated, but without the warning. I don't think anyone really reads these, and if someone were to use this, I'm sure they'd reach out first, or if not, I doubt a warning would make them do it...

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: In my post when I meant "original" I meant "before lock watches even exist", so the operation is reading A and then reading B - i.e. my concern is about the difference between

```
user reads table A, checks lock
<loses immutable timestamp lock>
user reads table B without notifications (because they don't exist), checks lock, fails
txn.commit()
```
and
```
user reads table A, checks lock
<loses immutable timestamp lock>
user reads table B via notifications, marked B as used
txn.commit()
```

Yep, I think the thing I wanted to verify was that this is _only_ for sweep protection, which I believe is correct - if that's the case then I'm pretty sure this is fine

jkozlowski: Yep, discussed offline. I think we don't need to make it this strict, as long as there's a check during the transaction's lifetime, I think that's fine by me.

jeremyk-91: Just as a note for future awareness, we had a number of bad legacy users which did goofy things like read data read by a transaction (that was not the thing returned) outside of the scope of that transaction. Among other things, this is why we had to have the whole `checkLocksOnReads` true by default thing (and so "check during the transaction's lifetime" might actually not be sufficient _in general_ - though I don't have concrete examples to hand), but within the parameters of our usage is definitely fine.

jeremyk-91 (approved):

PR #4973, opened 2020-09-03
[TS] Auto Tune TS Pause Between Iterations

gmaretic:
**Goals (and why)**:
Allow TS to change the pause between iterations depending on how much work was done. This serves primarily to slow things down when there are not many entries to sweep and conserve resources, but also to reduce the pause when we detect we are falling behind.

The auto tuning is enabled through a runtime configuration. When true, after each iteration, we will determine the next pause based on the last one and the number of entries swept. Furthermore, we will continue to read from the queue until either a full batch is read or we reach the sweep timestamp.

The `batchShardIterations` runtime config has deprecated and is ignored now. This is because it is better to use `maximumPartitionsToBatchInSingleRead` or the new auto tuning for the purpose, and it was adding unnecessary complexity. 

Added tests for `SweepDelay`, modified some tests in `TargetedSweeperTest`. Do you think we need more?

Generally, I don't think this is too contentious. Mostly interested if there are disagreements on the numbers in `SweepDelay`, and if I missed something.
Is there a better way of doing the recurring task in `TargetedSweeper`, and do we need to be defensive around `executorService.schedule` throwing?

`TargetedSweepRuntimeConfig.java`

Ideally today

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
@@ -19,26 +19,42 @@ package com.palantir.atlasdb.sweep.queue;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicLong;
 
-public class SweepDelay {
-    public static final int BATCH_CELLS_LOW_THRESHOLD = 100;
-    public static final long MIN_PAUSE_MILLIS = 1;
-    public static final long MAX_PAUSE_MILLIS = 5000;
-    public static final long BACKOFF = Duration.ofMinutes(2).toMillis();
+class SweepDelay {
+    static final int BATCH_CELLS_LOW_THRESHOLD = 100;
+    static final long MIN_PAUSE_MILLIS = 1;
+    static final long DEFAULT_MAX_PAUSE_MILLIS = 5000;
+    static final long BACKOFF = Duration.ofMinutes(2).toMillis();
 
     private final long initialPause;
+    private final long maxPauseMillis;
     private final AtomicLong currentPause;
 
-    public SweepDelay(long configPause) {
-        this.initialPause = Math.min(Math.max(MIN_PAUSE_MILLIS, configPause), MAX_PAUSE_MILLIS);
+    /**
+     * This class calculates the delay for the next iteration of targeted sweep from the current delay and the outcome
+     * of the last iteration of TS. If the sweep iteration was successful, the next delay will gravitate towards the
+     * target delay using the formula 0.2 * target + 0.8 * current. The target delay is as follows:
+     *
+     *  1. if the sweep iteration processed fewer than {@link #BATCH_CELLS_LOW_THRESHOLD} cells, the target pause is
+     *  {@link #maxPauseMillis} milliseconds.
+     *  2. if the sweep iteration processed a full batch of {@link SweepQueueUtils#SWEEP_BATCH_SIZE} or more cells, the
+     *  target pause is {@link #MIN_PAUSE_MILLIS} milliseconds.
+     *  3. otherwise, the target pause is {@link #initialPause} milliseconds.
+     *
+     *  In case of an unsuccessful iteration, the pause is temporarily set to a constant as determined in
+     *  {@link #getNextPause(SweepIterationResult)}.
+     */
+    SweepDelay(long configPause) {
+        this.maxPauseMillis = Math.max(DEFAULT_MAX_PAUSE_MILLIS, configPause);
+        this.initialPause = Math.max(MIN_PAUSE_MILLIS, configPause);
         this.currentPause = new AtomicLong(initialPause);
     }
 
-    public long getNextPause(SweepIterationResult result) {
+    long getNextPause(SweepIterationResult result) {
         return SweepIterationResults.caseOf(result)
                 .success(this::updateCurrentPauseAndGet)
-                .unableToAcquireShard_(MAX_PAUSE_MILLIS)
+                .unableToAcquireShard_(maxPauseMillis)
                 .insufficientConsistency_(BACKOFF)
-                .otherError_(MAX_PAUSE_MILLIS)
+                .otherError_(maxPauseMillis)
                 .disabled_(BACKOFF);
     }
 
@@ -49,7 +65,7 @@ public class SweepDelay {
 
     private long pauseTarget(long numSwept) {
         if (numSwept <= BATCH_CELLS_LOW_THRESHOLD) {
-            return MAX_PAUSE_MILLIS;
+            return maxPauseMillis;
         } else if (numSwept >= SweepQueueUtils.SWEEP_BATCH_SIZE) {
             return MIN_PAUSE_MILLIS;
         }
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
@@ -19,6 +19,20 @@ package com.palantir.atlasdb.sweep.queue;
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicLong;
 
+/**
+ * This class calculates the delay for the next iteration of targeted sweep from the current delay and the outcome
+ * of the last iteration of TS. If the sweep iteration was successful, the next delay will gravitate towards the
+ * target delay using the formula 0.2 * target + 0.8 * current. The target delay is as follows:
+ *
+ *  1. if the sweep iteration processed fewer than {@link #BATCH_CELLS_LOW_THRESHOLD} cells, the target pause is
+ *  {@link #maxPauseMillis} milliseconds.
+ *  2. if the sweep iteration processed a full batch of {@link SweepQueueUtils#SWEEP_BATCH_SIZE} or more cells, the
+ *  target pause is {@link #MIN_PAUSE_MILLIS} milliseconds.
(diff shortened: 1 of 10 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
```diff
@@ -0,0 +1,58 @@
...
+    private final long initialPause;
+    private final AtomicLong currentPause;
+
+    public SweepDelay(long configPause) {
+        this.initialPause = Math.min(Math.max(MIN_PAUSE_MILLIS, configPause), MAX_PAUSE_MILLIS);
+        this.currentPause = new AtomicLong(initialPause);
+    }
+
+    public long getNextPause(SweepIterationResult result) {
+        return SweepIterationResults.caseOf(result)
```
jeremyk-91, line 37: This is pretty nice.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -229,44 +217,62 @@ private void assertInitialized() {
...
         private BackgroundSweepScheduler(int numThreads, SweeperStrategy sweepStrategy) {
             this.numThreads = numThreads;
             this.sweepStrategy = sweepStrategy;
+            this.delay = new SweepDelay(runtime.get().pauseMillis());
         }
 
         private void scheduleBackgroundThreads() {
             if (numThreads > 0 && executorService == null) {
                 executorService = PTExecutors
                         .newScheduledThreadPoolExecutor(numThreads, new NamedThreadFactory("Targeted Sweep", true));
```
jeremyk-91, line 243: Something to be aware of (I guess it's fine for now but will affect the next stage of changes): this will spin up `numThreads` many core threads and core thread timeout is not allowed by the executor returned here, so some changes will be needed when we extend this to cover that.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/config/TargetedSweepRuntimeConfig.java
```diff
@@ -46,9 +46,10 @@ public int shards() {
...
 
     /**
-     * If true, we batch many iterations on each shard and strategy upon obtaining the lock. This should lead to
-     * higher throughput in targeted sweep at the expense of more uneven sweeping across different shards.
+     * @deprecated This configuration is ignored, use {@link #enableAutoTuning()} or
+     * {@link #maximumPartitionsToBatchInSingleRead()} instead.
      */
+    @Deprecated
     @Value.Default
     public boolean batchShardIterations() {
```
jeremyk-91, line 54: I think you can delete it and put in a JsonIgnoreProperties annotation at the top.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.sweep.queue;
+
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SweepDelay {
```
jeremyk-91, line 22: would be good to add some documentation on how the self-tuning mechanism works

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
```diff
@@ -0,0 +1,58 @@
...
+    public static final int BATCH_CELLS_LOW_THRESHOLD = 100;
+    public static final long MIN_PAUSE_MILLIS = 1;
+    public static final long MAX_PAUSE_MILLIS = 5000;
+    public static final long BACKOFF = Duration.ofMinutes(2).toMillis();
+
+    private final long initialPause;
+    private final AtomicLong currentPause;
+
+    public SweepDelay(long configPause) {
+        this.initialPause = Math.min(Math.max(MIN_PAUSE_MILLIS, configPause), MAX_PAUSE_MILLIS);
```
jeremyk-91, line 32: I'd question whether we want to place a cap on MAX_PAUSE_MILLIS: it seems a bit unexpected that if the user configures the gap to be 10 seconds for whatever reason we just throw it away. Even if we think that's ok I think it makes sense to fail more aggressively.

In other words I'd go for 
```
initial = max(min, config)
min = 1
max = max(5000, config)
```

Note that if you go this route you may need some special integer mathematics in line 47...

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        assertThat(delay.getNextPause(SUCCESS_TOO_FAST)).isGreaterThanOrEqualTo((long) (MAX_PAUSE_MILLIS * 0.95));
+    }
+
+    @Test
+    public void consistentFullBatchesGravitatesTowardsMinimumPause() {
+        for (int i = 0; i < 20; i++) {
+            delay.getNextPause(SUCCESS_TOO_SLOW);
+        }
+        assertThat(delay.getNextPause(SUCCESS_TOO_FAST)).isGreaterThanOrEqualTo((long) (MIN_PAUSE_MILLIS * 1.05));
```
jeremyk-91, line 95: If I'm understanding this correctly this one should read

```suggestion
        assertThat(delay.getNextPause(SUCCESS_TOO_SLOW)).isLessThanOrEqualTo((long) (MIN_PAUSE_MILLIS * 1.05));
```

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    public void consistentNormalBatchesAfterFullBatchesGravitatesTowardsInitialPause() {
+        for (int i = 0; i < 20; i++) {
+            delay.getNextPause(SUCCESS_TOO_SLOW);
+        }
+        for (int i = 0; i < 20; i++) {
+            delay.getNextPause(SUCCESS);
+        }
```
jeremyk-91, line 105: at this point it's probably worth having a `sweepTwentyBatchesWithOutcome(long)`?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
```diff
@@ -0,0 +1,111 @@
...
+        long nextPause = delay.getNextPause(SUCCESS);
+        assertThat(nextPause).isGreaterThanOrEqualTo((long) (INITIAL_DELAY * 0.95));
+        assertThat(nextPause).isLessThanOrEqualTo((long) (INITIAL_DELAY * 1.05));
```
jeremyk-91, line 108:
```suggestion
        assertThat(nextPause).isBetween((long) (INITIAL_DELAY * 0.95), (long) (INITIAL_DELAY * 1.05));
```

changelog/@unreleased/pr-4973.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    Targeted sweep runtime configuration now has a `enableAutoTuning` parameter. While true, targeted sweep will dynamically adjust the pause between sweep iterations based on the amount of work that was done. It will, furthermore, attempt to read a full batch (100k) of entries to sweep whenever possible.
```
jeremyk-91, line 4: Probably worth calling out this is not yet enabled by default and please talk to the AtlasDB team if you want to use it.

jeremyk-91 (commented):
Broadly looks good! I have a question on the clamping - others are mostly small details

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
gmaretic: 🥳

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
gmaretic: Yeah, once we do task scaling, this is set to 1 instead

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
gmaretic: Makes sense, I now just use it as both initial and max. Do we want to do something like max should always be at least 2 * initial instead?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
gmaretic: 🎖️

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
gmaretic: Nice catch!

jeremyk-91 (approved):
👍  looks good!

PR #4974, opened 2020-09-04
Leader election slos

sudiksha27:
**Goals (and why)**:
Fix bug with leader election rate health check. The health check should be analyzing the metrics per client.

- Register client wise metrics to leader election rate health check.

- Leader election health check performs analysis on the sum of rates of all clients

Tested on internal testing environment

Does the wiring make sense?

LeaderElectionHealthCheck.java, TimeLockAgent.java

As soon as possible 🔥

### DIFF
--- a/changelog/@unreleased/pr-4974.v2.yml
+++ b/changelog/@unreleased/pr-4974.v2.yml
@@ -1,5 +1,5 @@
-type: improvement
-improvement:
+type: fix
+fix:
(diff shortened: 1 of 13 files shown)


gmaretic (approved):
Looks fine, a few nits and would like a test for multiple clients though

gmaretic:
LGTM!

PR #4975, opened 2020-09-07
[TS] Auto tune number of parallel background tasks

gmaretic:
**Goals (and why)**:
Changing the pause between iterations can only get us so far. Here we allow TS to slowly increase or decrease the parallelism over time as necessary.

We keep track of the number of running tasks and use the number of entries swept in an iteration to decide if we want to reduce, maintain, or increase the number of tasks. The number of tasks is always between 1 and 128.

Normally, we only reduce/increase the number of tasks by 1, and only once every 5 minutes. Furthermore, if we wanted to increase the number in the last 5 minutes (but couldn't because of cooldown), we will not reduce it as this implies some shards have fallen behind.

In the exceptional cases where we cannot obtain a shard to sweep, or sweep is disabled, we aggressively reduce the number of tasks to avoid using up resources.

There are tests for the scheduler. Let me know if there is something that was ommitted.

Tests seem to be flaking, on completely unrelated ete tests (`LockWithoutTimelockEteTest`)
 
Tests?

Ideally today


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ScalingSweepTaskSchedulerTest.java
```diff
@@ -0,0 +1,310 @@
...
+        when(delay.getInitialPause()).thenReturn(INITIAL_PAUSE);
+        when(delay.getNextPause(any(SweepIterationResult.class))).thenReturn(DELAY);
+    }
+
+    @Test
+    public void whenExpectedNumberOfEntriesIsSweptKeepReschedulingAfterDelay() throws Exception {
+        when(sweepIteration.call()).thenReturn(SUCCESS_MEDIUM);
+        scheduler.start(2);
+
+        deterministicScheduler.tick(INITIAL_DELAY + 8 * DELAY, TimeUnit.MILLISECONDS);
```
jeremyk-91, line 75: nit: probably just factor out `runSweepIterations(iterations)` for ticking `INITIAL_DELAY + (iterations - 1) * DELAY`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ScalingSweepTaskSchedulerTest.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    @Test
+    public void whenUnableToAcquireShardOnLastTaskRescheduleAfterMaxPause() throws Exception {
+        ScalingSweepTaskScheduler schedulerWithRealDelay = createScheduler(new SweepDelay(1L), Duration.ZERO);
+        when(sweepIteration.call()).thenReturn(SweepIterationResults.unableToAcquireShard(), SUCCESS_MEDIUM);
+
+        schedulerWithRealDelay.start(1);
+        deterministicScheduler.tick(INITIAL_DELAY + SweepDelay.DEFAULT_MAX_PAUSE_MILLIS + 2, TimeUnit.MILLISECONDS);
+        verify(sweepIteration, times(1 + 3)).call();
```
jeremyk-91, line 170: In general for tests like this one, you might want to tick the ticker a bit less, check that it hadn't retriggered yet, and then tick it to the required value

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ScalingSweepTaskSchedulerTest.java
```diff
@@ -0,0 +1,310 @@
...
+        when(sweepIteration.call()).thenReturn(
+                SUCCESS_LARGE,
+                SUCCESS_LARGE,
+                SUCCESS_SMALL);
+        schedulerWithCoolDown.start(2);
+
+        Uninterruptibles.sleepUninterruptibly(coolDown.toMillis() / 2 + 1, TimeUnit.MILLISECONDS);
+        deterministicScheduler.tick(INITIAL_DELAY, TimeUnit.MILLISECONDS);
+
+        Uninterruptibles.sleepUninterruptibly(coolDown.toMillis() / 2 + 1, TimeUnit.MILLISECONDS);
```
jeremyk-91, line 152: We should be able to avoid this: you can pass in a clock or rather `Supplier<Instant>` to the scheduler

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ScalingSweepTaskScheduler.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private Void determineAction(long numThreads, long pause) {
+        if (numThreads <= BATCH_CELLS_LOW_THRESHOLD) {
+            maybeDecreaseNumberOfTasks(pause);
+        } else if (numThreads >= BATCH_CELLS_HIGH_THRESHOLD) {
+            maybeIncreaseNumberOfTasks(pause);
+        } else {
+            scheduleAfterDelay(pause);
+        }
+        return null;
```
jeremyk-91, line 158: I would throw a `SafeIllegalStateException` in this case (unless there's some vagary of the type system I'm missing).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/ScalingSweepTaskScheduler.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private void retryingTask(Callable<SweepIterationResult> task) {
+        try {
+            SweepIterationResult sweepResult = task.call();
+            if (!scalingEnabled.getAsBoolean()) {
+                scheduleAfterDelay(delay.getInitialPause());
+            } else {
+                long pause = delay.getNextPause(sweepResult);
+                SweepIterationResults.caseOf(sweepResult)
+                        .success(numThreads -> determineAction(numThreads, pause))
```
jeremyk-91, line 139: nit: `runningTasks`, for consistency with elsewhere in the class.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/LockWithoutTimelockEteTest.java
```diff
@@ -1,37 +0,0 @@
...
-import org.junit.Test;
-
-import com.palantir.atlasdb.lock.LockResource;
-
-public class LockWithoutTimelockEteTest {
-    private LockResource lockResource = EteSetup.createClientToSingleNode(LockResource.class);
-
-    @Test
-    public void hugeV1LockSucceeds() throws InterruptedException {
-        assertThat(lockResource.lockUsingLegacyLockApi(100, 500_000)).isTrue();
```
jeremyk-91, line 30: Let's talk about this tomorrow. I don't see how this change would affect behaviour here, but because of large internal product's lock service implementation, we do need to know if the huge requests were broken.

jeremyk-91 (commented):
The PR has good structure. My main questions are around the removal of the LockWithoutTimelockEteTests. Otherwise there are some incremental improvements to the impl and tests that I'd suggest, but broadly everything looks right!

gmaretic (commented):
Addressed CR

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/LockWithoutTimelockEteTest.java
gmaretic: Reverted the change as discussed

jeremyk-91 (approved):
🚀

PR #4985, opened 2020-09-15
[TS] Increase max delay

gmaretic:
**Goals (and why)**:
5 seconds was good, but we can increase it further without falling behind

The default number of shards is 8, so on a service with a single node we should not fall behind by 2 minutes, which is reasonable. In practice, I expect at least 2 nodes, so this number goes down to 1 minute. Do you think we should increase it even further?


jeremyk-91 (approved):
👍

PR #4988, opened 2020-09-18
remove unnecessary oracle config restrictions

berler:
**Goals (and why)**:

Forbidding configuring the keystore/truststore when the oracle protocol
is not TCPS makes it a lot harder to automate conifgs. It is much easier
to allow our config automation to always fill in these values even when
they won't be used.

This does not remove the validation that checks they are present when
they really are required.

* remove precondition checks for keystore/truststore being absent when the protocol isn't TCPS

* unit tests have been added. no existing tests were validating the current behavior

**Priority (whenever / two weeks / yesterday)**: high -- we'd like to make so that it is easier to switch between TCP/TCPS on oracle in our internal products


atlasdb-dbkvs-hikari/src/test/java/com/palantir/nexus/db/pool/config/OracleConnectionConfigTest.java
```diff
@@ -85,6 +90,71 @@ public void namespaceIsNamespaceOverrideIfServiceNameConfigurationSpecified() {
...
+                    .containsEntry("javax.net.ssl.trustStore", truststoreFile.getAbsolutePath())
+                    .containsEntry("javax.net.ssl.trustStorePassword", "password")
+                    .doesNotContainKeys("javax.net.ssl.keyStore", "javax.net.ssl.keyStorePassword");
+        } finally {
+            // clean up the truststore file if we created it
+            if (truststoreFileCreated) {
+                truststoreFile.delete();
+            }
+        }
+    }
```
jeremyk-91, line 156: thanks for the detailed tests!

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/OracleConnectionConfig.java
```diff
@@ -196,16 +196,6 @@ protected final void check() {
...
-            Preconditions.checkArgument(!getServerDn().isPresent(),
-                    "a server dn cannot be given without enabling ConnectionProtocol.TCPS");
-            Preconditions.checkArgument(!getTruststorePath().isPresent(),
-                    "a truststore path cannot be given without enabling ConnectionProtocol.TCPS");
-            Preconditions.checkArgument(!getTruststorePassword().isPresent(),
-                    "a truststore password cannot be given without enabling ConnectionProtocol.TCPS");
-            Preconditions.checkArgument(!getKeystorePath().isPresent(),
-                    "a keystore file cannot be given without enabling ConnectionProtocol.TCPS");
-            Preconditions.checkArgument(!getKeystorePassword().isPresent(),
-                    "a keystore password without enabling ConnectionProtocol.TCPS");
```
jeremyk-91, line 208: I'd prefer if we put a `WARN` log here if any of them are present? Basically it's easy to make a config error and then be really puzzled as to why the settings aren't taking effect.

jeremyk-91 (approved):
👍  looks good, I think we can improve the observability a bit, but this makes sense

jeremyk-91:
whoops, didn't see MWR+AR. Will open that PR in a bit

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/OracleConnectionConfig.java
berler: In this case if you set the protocol incorrectly you wouldn't be able to connect to the DB. Warning on here would just be unnecessary noise if the rest of the prefs were actually correct.

PR #4990, opened 2020-09-21
Log WARN if some security settings were ignored in an Oracle conn config

jeremyk-91:
**Goals (and why)**:
#4988 allows Oracle configs to be created with TCP (not TCPS) and security settings. This is easier for large internal product's automation. I had a suggestion and didn't realise merge-when-ready was there when I approved.

Add a WARN log line if both plain TCP is used and security settings are present, but allow the system to proceed.

None; only logging

Nothing in particular

**Priority (whenever / two weeks / yesterday)**: today please, it's small


gmaretic (approved):
LGTM

jeremyk-91:
policybot check actually passes 👍

PR #4991, opened 2020-09-22
Remote history 3

sudiksha27:
**Goals (and why)**:
Integrate local and remote histories for corruption analysis

- Basically data transforms on local and remote histories.

- Added tests 

- The transforms are hard to read, are there better ways to implement this?

- PaxosLogHistoryProvider

tomorrow would be ideal.


timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -24,24 +24,43 @@
...
 import com.palantir.paxos.ImmutableNamespaceAndUseCase;
 import com.palantir.paxos.NamespaceAndUseCase;
+import com.palantir.timelock.history.HistoryQuery;
+import com.palantir.timelock.history.LogsForNamespaceAndUseCase;
+import com.palantir.timelock.history.PaxosLogWithAcceptedAndLearnedValues;
+import com.palantir.timelock.history.TimeLockPaxosHistoryProvider;
+import com.palantir.tokens.auth.AuthHeader;
 
 public class PaxosLogHistoryProvider {
+    private final Logger log = LoggerFactory.getLogger(PaxosLogHistoryProvider.class);
```
jeremyk-91, line 51:
```suggestion
    private static final Logger log = LoggerFactory.getLogger(PaxosLogHistoryProvider.class);
```

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
...
         Map<NamespaceAndUseCase, Long> lastVerifiedSequences = KeyedStream
                 .of(getNamespaceAndUseCaseTuples().stream())
                 .map(namespaceAndUseCase -> verificationProgressStateCache.computeIfAbsent(namespaceAndUseCase,
                         this::getOrInsertVerificationState))
                 .collectToMap();
+
+        /**
+         * The history queries are built from the lastVerifiedSequences map above,
+         * required for conjure endpoints to load remote history
+         */
```
jeremyk-91, line 101: _Generally_, and I think this also does apply here: when functions reach this kind of complexity, you should consider extracting some of the individual calls to methods with descriptive names. (In this way you usually won't need comments as well.)

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
                 namespaceAndUseCase.namespace(), namespaceAndUseCase.useCase());
     }
 
-    public List<CompletePaxosHistoryForNamespaceAndUseCase> getHistory() {
+    /**
+     * Gets history from all the nodes for all unique namespace, useCase tuples since last verified sequence numbers
```
jeremyk-91, line 84: nit: `(namespace, useCase) tuples` (with parentheses)

It's also useful to explain what `last verified sequence numbers` are here (specifically, the highest sequence number that was verified since the last time the bounds were reset?)

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
...
+     *
+     * @throws  RuntimeException
+     *          if fails to fetch history from all remote servers
+     */
+    public List<CompletePaxosHistoryForNamespaceAndUseCase> getHistory() throws RuntimeException {
```
jeremyk-91, line 89: nit: we don't generally declare unchecked exceptions
```suggestion
    public List<CompletePaxosHistoryForNamespaceAndUseCase> getHistory() {
```

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
...
+         * List of logs from all remotes
+         */
+        List<List<LogsForNamespaceAndUseCase>> rawHistoryFromAllRemotes = remoteHistoryProviders.stream().map(
+                remote -> fetchHistoryFromRemote(historyQueries, remote)).collect(Collectors.toList());
+
+        /**
+         * List of history from all remotes. Each history is map of namespaceAndUseCase to Paxos logs. The Paxos logs
+         * are mapped against sequence numbers.
+         */
+        List<Map<NamespaceAndUseCase, Map<Long, LearnedAndAcceptedValue>>> historyLogsFromRemotes
```
jeremyk-91, line 120: I think we should have some wrapper types here, the primitives get tricky to read after a while.

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
...
+                        log -> ImmutableLearnedAndAcceptedValue.of(log.getPaxosValue(), log.getAcceptedState())));
+    }
+
+    private List<LogsForNamespaceAndUseCase> fetchHistoryFromRemote(List<HistoryQuery> historyQueries,
+            TimeLockPaxosHistoryProvider remote) {
+        try {
+            return remote.getPaxosHistory(AUTH_HEADER, historyQueries);
+        } catch (Exception e) {
+            log.warn("The remote failed to provide the history,"
+                    + "we cannot perform corruption checks without history from all nodes.");
```
jeremyk-91, line 201: nit: Please include `e`

timelock-corruption-detection/src/main/java/com/palantir/history/PaxosLogHistoryProvider.java
```diff
@@ -61,16 +80,131 @@ private Long getOrInsertVerificationState(NamespaceAndUseCase namespaceAndUseCas
...
+            throw new RuntimeException(e.getCause());
```
jeremyk-91, line 202: How do we know that `e` is (1) not a runtime exception, (2) has a cause?

timelock-corruption-detection/src/test/java/com/palantir/history/utils/Utils.java
```diff
@@ -0,0 +1,61 @@
...
+import java.nio.ByteBuffer;
+import java.util.UUID;
+import java.util.stream.IntStream;
+
+import com.palantir.paxos.PaxosAcceptorState;
+import com.palantir.paxos.PaxosProposalId;
+import com.palantir.paxos.PaxosStateLog;
+import com.palantir.paxos.PaxosValue;
+
+public final class Utils {
```
jeremyk-91, line 28: I'd suggest naming this `PaxosSerializationTestUtils` or something like that, since it's public

timelock-corruption-detection/src/test/java/com/palantir/history/utils/Utils.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static PaxosAcceptorState getAcceptorStateForRound(long round) {
+        return PaxosAcceptorState.newState(new PaxosProposalId(
+                round, UUID.randomUUID().toString()));
+    }
+
+
+    public static void writeToLogs(PaxosStateLog<PaxosAcceptorState> acceptorLog,
+            PaxosStateLog<PaxosValue> learnerLog, int range) {
+        IntStream.range(0, range).forEach(i -> {
```
jeremyk-91, line 50: `IntStream.rangeClosed(1, range)` probably does what you're intending here

timelock-corruption-detection/src/main/java/com/palantir/history/models/PaxosHistoryOnSingleNode.java
```diff
@@ -17,14 +17,38 @@
...
+import com.google.common.collect.ImmutableMap;
 import com.palantir.paxos.NamespaceAndUseCase;
 
 @Value.Immutable
 public interface PaxosHistoryOnSingleNode {
 
     @Value.Parameter
     Map<NamespaceAndUseCase, LearnerAndAcceptorRecords> history();
+
+    default Map<Long, LearnedAndAcceptedValue> getConsolidatedLocalAndRemoteRecord(
```
jeremyk-91, line 35: If we keep this model, it may be worth doing some kind of caching. I'm wondering if a derived or lazy `Map<NamespaceAndUseCase, Map<Long, LearnedAndAcceptedValue>` is a good idea, even

timelock-corruption-detection/src/main/java/com/palantir/history/models/LearnerAndAcceptorRecords.java
```diff
@@ -31,4 +33,34 @@
 
     @Value.Parameter
     Map<Long, PaxosAcceptorState> acceptorRecords();
+
+    default Optional<PaxosValue> getLearnedValueAtSeqIfExists(long seq) {
+        return learnerRecords().containsKey(seq) ? Optional.of(learnerRecords().get(seq)) : Optional.empty();
```
jeremyk-91, line 38: isn't this `Optional.ofNullable()`?

timelock-corruption-detection/src/main/java/com/palantir/history/TimeLockPaxosHistoryProviderResource.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @VisibleForTesting
+    TimeLockPaxosHistoryProviderResource(LocalHistoryLoader localHistoryLoader) {
+        this.localHistoryLoader = localHistoryLoader;
+    }
+
+    @Override
+    public ListenableFuture<List<LogsForNamespaceAndUseCase>> getPaxosHistory(AuthHeader authHeader,
+            List<HistoryQuery> historyQueries) {
+        Map<NamespaceAndUseCase, Long> lastVerifiedSequences = historyQueries.stream().collect(
```
jeremyk-91, line 53: Ideally let's separate the business logic from the controller-y work that this layer is doing

timelock-api/build.gradle
```diff
@@ -10,6 +10,8 @@ conjure {
 }
 
 dependencies {
+    compile project(":leader-election-impl")
```
jeremyk-91, line 13: Let's try to avoid dependencies on `leader-election-impl` in `timelock-api`. For example, this means internal backup service will be able to create a Paxos leader election service impl, which feels off.

timelock-api/src/main/conjure/timelock-history.yml
```diff
@@ -0,0 +1,49 @@
+types:
+  imports:
+    PaxosValue:
+      base-type: any
+      external:
+        java: com.palantir.paxos.PaxosValue
+    PaxosAcceptorState:
```
jeremyk-91, line 7: This is the class we got from `leader-election-impl`, so we'll probably need an adapter. It looks like that wraps proposal IDs and Values, which are from `leader-election-api`, and that seems OK.

Maybe we can look at whether this fits in the timelock corruption project?

jeremyk-91 (commented):
This review doesn't cover the tests. I think the rough approach seems correct, though I think there's some improvements we can make around how it's laid out.

PR #4992, opened 2020-09-22
[WIP] LW client event log takes immutable timestamp into account when reten…

jkozlowski:
…tioning events.

**Goals (and why)**:


jkozlowski:
Going to go with max size for now, then do this as an additional policy.

PR #4993, opened 2020-09-22
Remote history resource

sudiksha27:
**Goals (and why)**:
Implements the remote history provider piece

- Add conjure endpoint that provides history

Added new tests

- Do the Conjure definitions seem reasonable?

timelock-history.yml 

ASAP


timelock-corruption-detection/build.gradle
```diff
@@ -1,7 +1,21 @@
...
+}
 
 dependencies {
-    compile project(":leader-election-impl")
+    conjureCompiler 'com.palantir.conjure:conjure'
+    conjureJava 'com.palantir.conjure.java:conjure-java'
+
+    compile project(":timelock-corruption-detection:timelock-corruption-detection-undertow")
+    compile project(":timelock-corruption-detection:timelock-corruption-detection-jersey")
+    compile project(":timelock-corruption-detection:timelock-corruption-detection-objects")
```
jeremyk-91, line 18: nit: no need to depend on `objects` project directly

timelock-corruption-detection/src/main/conjure/timelock-history.yml
```diff
@@ -0,0 +1,52 @@
+types:
+  imports:
+    PaxosValue:
+      base-type: any
+      external:
+        java: com.palantir.paxos.PaxosValue
```
jeremyk-91, line 6: I'm afraid this won't actually work in its current form, because `PaxosValue` is not JSON serializable.

timelock-corruption-detection/src/main/conjure/timelock-history.yml
```diff
@@ -0,0 +1,52 @@
...
+    PaxosAcceptorState:
+      base-type: any
+      external:
+        java: com.palantir.paxos.PaxosAcceptorState
```
jeremyk-91, line 10: I think for this one we should probably still have a separate layer, because this class is used internally within the Paxos implementation. (Unlike PaxosValue, this one isn't a part of TimeLock's internode API.)

timelock-corruption-detection/src/main/conjure/timelock-history.yml
```diff
@@ -0,0 +1,52 @@
...
+          logs: list<PaxosLogWithAcceptedAndLearnedValues>
+      PaxosHistoryOnRemote:
+        fields:
+          logs: list<LogsForNamespaceAndUseCase>
+
+services:
+  TimeLockPaxosHistoryProvider:
+    name: TimeLock Paxos History Provider Service
+    default-auth: header
+    package: com.palantir.timelock.history
```
jeremyk-91, line 43: We should standardise on the package to be used (I think, unless I'm missing something - you've used both `com.palantir.timelock.history` and `com.palantir.history`). I think the former is the correct one (i.e. no changes here, but we might want to move the others around).

timelock-corruption-detection/src/main/java/com/palantir/history/models/LearnerAndAcceptorRecords.java
```diff
@@ -31,4 +33,34 @@
 
     @Value.Parameter
     Map<Long, PaxosAcceptorState> acceptorRecords();
+
+    default Optional<PaxosValue> getLearnedValueAtSeqIfExists(long seq) {
+        return Optional.ofNullable(learnerRecords().getOrDefault(seq, null));
```
jeremyk-91, line 38: nit: This is fine, but have a look at the documentation of `get()`.

timelock-corruption-detection/src/main/java/com/palantir/history/models/LearnerAndAcceptorRecords.java
```diff
@@ -31,4 +33,34 @@
...
+
+    default long getMinSequence() {
+        long minSeq = Long.MAX_VALUE;
+        if (!learnerRecords().isEmpty()) {
+            minSeq = Math.min(minSeq, Collections.min(learnerRecords().keySet()));
+        }
+        if (!acceptorRecords().isEmpty()) {
+            minSeq = Math.min(minSeq, Collections.min(acceptorRecords().keySet()));
+        }
+        return minSeq;
```
jeremyk-91, line 53: Yeah, I think this function is correct, though this is maybe a bit neater?

```suggestion
        return Stream.of(learnerRecords(), acceptorRecords())
                .map(Map::keySet)
                .flatMap(Set::stream)
                .min(Comparator.naturalOrder())
                .orElse(Long.MAX_VALUE);
```

timelock-corruption-detection/src/main/java/com/palantir/history/models/LearnerAndAcceptorRecords.java
```diff
@@ -31,4 +33,34 @@
...
+
+    default Optional<PaxosValue> getLearnedValueAtSeqIfExists(long seq) {
+        return Optional.ofNullable(learnerRecords().getOrDefault(seq, null));
+    }
+
+    default Optional<PaxosAcceptorState> getAcceptedValueAtSeqIfExists(long seq) {
+        return Optional.ofNullable(acceptorRecords().getOrDefault(seq, null));
+    }
+
+    default long getMinSequence() {
```
jeremyk-91, line 45: Looks like this could be `@Value.Lazy`

timelock-corruption-detection/src/main/java/com/palantir/history/models/LearnerAndAcceptorRecords.java
```diff
@@ -31,4 +33,34 @@
...
+
+    default long getMaxSequence() {
+        long maxSeq = Long.MIN_VALUE;
+        if (!learnerRecords().isEmpty()) {
+            maxSeq = Math.max(maxSeq, Collections.max(learnerRecords().keySet()));
+        }
+        if (!acceptorRecords().isEmpty()) {
+            maxSeq = Math.max(maxSeq, Collections.max(acceptorRecords().keySet()));
+        }
+        return maxSeq;
```
jeremyk-91, line 64: as above, I don't know if we want to cache the value for recomputations, and I think streams give an easier pipeline.

timelock-corruption-detection/src/main/java/com/palantir/history/remote/HistoryLoaderAndTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+
+public final class HistoryLoaderAndTransformer {
+    private HistoryLoaderAndTransformer() {
+        //no op
+    }
+
+    public static List<LogsForNamespaceAndUseCase> getLogsForHistoryQueries(LocalHistoryLoader localHistoryLoader,
+            List<HistoryQuery> historyQueries) {
+        Map<NamespaceAndUseCase, Long> lastVerifiedSequences = historyQueries.stream().collect(
+                Collectors.toMap(HistoryQuery::getNamespaceAndUseCase, HistoryQuery::getSeq, Math::min));
```
jeremyk-91, line 42: nice!

timelock-corruption-detection/src/main/java/com/palantir/history/remote/HistoryLoaderAndTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+
+        long minSeq = records.getMinSequence();
+        long maxSeq = records.getMaxSequence();
+
+        List<PaxosLogWithAcceptedAndLearnedValues> logs = LongStream.rangeClosed(minSeq, maxSeq).boxed().map(
+                sequence -> PaxosLogWithAcceptedAndLearnedValues.builder()
+                        .paxosValue(records.getLearnedValueAtSeqIfExists(sequence))
+                        .acceptedState(records.getAcceptedValueAtSeqIfExists(sequence))
+                        .seq(sequence)
+                        .build()).collect(Collectors.toList());
```
jeremyk-91, line 63: Can we iterate over the map entries, as opposed to over this range? I'm a bit worried about this at deployments like big internal stack, where we truncated a good number of these entries so we have -1 and then a big gap, and then four million.

timelock-corruption-detection/src/test/java/com/palantir/history/HistoryLoaderAndTransformerTest.java
```diff
@@ -0,0 +1,183 @@
...
+public class HistoryLoaderAndTransformerTest {
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
+    private static final Client CLIENT = Client.of("client");
+    private static final String USE_CASE = "useCase";
+    private static final String USE_CASE_LEARNER = "useCase!learner";
+    private static final String USE_CASE_ACCEPTOR = "useCase!acceptor";
+
+    private DataSource dataSource;
```
jeremyk-91, line 58: This probably doesn't need to be exposed 😅

timelock-corruption-detection/src/test/java/com/palantir/history/HistoryLoaderAndTransformerTest.java
```diff
@@ -0,0 +1,183 @@
...
+        assertThat(logsForNamespaceAndUseCase.getNamespaceAndUseCase().useCase()).isEqualTo(USE_CASE);
+        assertThat(logsForNamespaceAndUseCase.getLogs().size()).isEqualTo(0);
+    }
+
+    private void writeToLogs(int range) {
+        IntStream.range(0, range).forEach(i -> {
+            writeAcceptorStateForLogAndRound(acceptorLog, i + 1);
+            writeValueForLogAndRound(learnerLog, i + 1);
+        });
+    }
```
jeremyk-91, line 182: This is pretty solid. I think you might want to add a test for a discontinuity (i.e. logs only exist from 45 - 100, and lastVerified was 3 or something like that), but other than that there's good coverage here.

jeremyk-91 (commented):
I think the structure you've got here is broadly correct. I've got a couple of concerns on the API, and there are a number of suggestions I have around the algorithmic part of this as well, but nothing that would require major rework.

This PR is pretty challenging. In the future, it could actually be worth just having a first PR for the conjure yml and projects, and then a second for the impl and resource. (Some might even say a second for the impl and third for the resource, but I think two is enough here.)

timelock-corruption-detection/src/test/java/com/palantir/history/HistoryLoaderAndTransformerTest.java
```diff
@@ -0,0 +1,183 @@
...
+        assertThat(logsForNamespaceAndUseCase.getLogs().size()).isEqualTo(100 - lastVerified);
+    }
+
+    @Test
+    public void canHandleDuplicateQueries() {
+        writeToLogs(100);
+        int minLastVerified = 27;
+
+        List<HistoryQuery> queries = IntStream.range(0, 10).boxed().map(
+                idx -> HistoryQuery.of(ImmutableNamespaceAndUseCase.of(CLIENT, USE_CASE), minLastVerified + idx))
```
jeremyk-91, line 96: nit: This object's used quite a bit, you can just extract it as a constant and then use it. Also for the assertions you can then use `assertThat(bleh.getNamespaceAndUseCase()).isEqualTo(NAMESPACE_AND_USE_CASE)`

timelock-corruption-detection/src/main/conjure/timelock-history.yml
sudiksha27: @Jkong verified that PaxosValue is JSON serializable.

jeremyk-91 (approved):
I think we should be good here.

PR #5000, opened 2020-09-24
[TimeLock] Fix Antiquated cURL Instructions From TimeLock Documentation

jeremyk-91:
**Goals (and why)**:
- TimeLock docs should allow developers to independently get some understanding of how to talk to timelock. 
- The current docs have a lot of references to paths like `bleh/api/timestamp/fresh-timestamp` when it should be `/api/tl/ts/bleh`, which are outdated and end up requiring us to assist people in `help-atlasdb`.

- Replace uses of TimeLock endpoints _on TimeLock_ with the new Conjure paths. Similarly update related guidance on required headers and wire format changes (e.g. for getFreshTimestamp).

- N/A, docs change.

- Is the Conjure path correctly constructed?
- There are a few places in the docs where I have *deliberately* kept the old notation, generally when we are referring to an embedded service. Have I confused anything (i.e. treated an embedded thing as timelock, or vice versa)?

**Where should we start reviewing?**: wherever

### DIFF
--- a/docs/source/services/timelock_service/manual-migration.rst
+++ b/docs/source/services/timelock_service/manual-migration.rst
@@ -53,13 +53,19 @@ from here on out, we'll refer to it as ``TS``.
 Step 3: Fast-Forwarding the Timelock Server
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. tip::
+
+   Throughout this document, ``curl`` commands that make requests to TimeLock may receive a 308 indicating the location
+   of the leader, or a 503 indicating that a node is not the leader (and doesn't know this). Commands should be re-run
+   until they succeed on the leader.
+
 The Timelock Server exposes an administrative interface, which features a ``fast-forward`` endpoint. Note that this is
 not typically exposed to AtlasDB clients. One can use it to advance the timestamp on the Timelock Server to ``TS``, as
 follows:
 
    .. code:: bash
 
-      curl -XPOST <protocol>://<host>:<port>/timelock/api/<namespace>/timestamp-management/fast-forward?currentTimestamp=TS \
+      curl -iXPOST <protocol>://<host>:<port>/timelock/api/<namespace>/timestamp-management/fast-forward?currentTimestamp=TS \
         -H "Authorization: Bearer q"
 
 .. danger::
@@ -80,7 +86,7 @@ To verify that this step was completed correctly, you may curl the Timelock Serv
 
    .. code:: bash
 
-      curl -XPOST <protocol>://<host>:<port>/timelock/api/tl/ts/<namespace> \
+      curl -iXPOST <protocol>://<host>:<port>/timelock/api/tl/ts/<namespace> \
         --data '{"numTimestamps": 1}' -H "Authorization: Bearer q" -H "Content-Type: application/json"
 
 The value returned should be (assuming no one else is using the Timelock Server) 1 higher than ``TS``.
--- a/docs/source/services/timelock_service/manual-migration.rst
+++ b/docs/source/services/timelock_service/manual-migration.rst
@@ -55,11 +55,12 @@ Step 3: Fast-Forwarding the Timelock Server
 
 The Timelock Server exposes an administrative interface, which features a ``fast-forward`` endpoint. Note that this is
 not typically exposed to AtlasDB clients. One can use it to advance the timestamp on the Timelock Server to ``TS``, as
-follows (where ``test`` is the namespace you want your client to use).
+follows:
 
    .. code:: bash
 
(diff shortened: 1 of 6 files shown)


docs/source/services/timelock_service/manual-migration.rst
```diff
@@ -79,7 +80,8 @@ To verify that this step was completed correctly, you may curl the Timelock Serv
 
    .. code:: bash
 
-      curl -XPOST localhost:8080/test/timestamp/fresh-timestamp
+      curl -XPOST <protocol>://<host>:<port>/timelock/api/tl/ts/<namespace> \
+        --data '{"numTimestamps": 1}' -H "Authorization: Bearer q" -H "Content-Type: application/json"
```
lycarter, line 84: It'd probably be nice to have a callout/reminder that only the leader will respond, and other nodes will return a 308. Adding a `-i` to the curl was the only thing that tipped me off to the 308, otherwise there's just no output from non-leaders.

docs/source/services/timelock_service/reverse-migration.rst
```diff
@@ -44,10 +44,10 @@ of timestamps across TimeLock and the key-value service backed services is monot
 This can be obtained using ``curl`` as follows. Note that if you obtain an HTTP 503 error, this is because you curled a
 node that was not the leader; please try the other nodes.
 
-.. code-block:: none
+   .. code:: bash
 
-   $ curl -XPOST https://<timelock-host>:8421/timelock/api/client/timestamp/fresh-timestamp
-   123456789
+      curl -XPOST <protocol>://<host>:<port>/timelock/api/tl/ts/<namespace> \
+        --data '{"numTimestamps": 1}' -H "Authorization: Bearer q" -H "Content-Type: application/json"
```
lycarter, line 50: possibly also a reminder about the leader here

docs/source/services/timelock_service/service-management.rst
```diff
@@ -33,7 +33,8 @@ less than the fresh timestamp. This may be obtained through the fresh timestamp
 
 .. code:: bash
 
-      curl -XPOST localhost:8080/timelock/api/old_namespace/timelock/fresh-timestamp
+    curl -XPOST https://localhost:8421/timelock/api/tl/ts/old_namespace \
+      --data '{"numTimestamps": 1}' -H "Authorization: Bearer q" -H "Content-Type: application/json"
 
 A cluster of TimeLock servers elects a single leader, so if you contact a node that is not the leader you'll receive a
 ``NotCurrentLeaderException`` and an HTTP 503. In this case, try the same request on other nodes until you find the
```
lycarter, line 40: For this endpoint I received a 308, not 503

docs/source/services/timelock_service/service-management.rst
```diff
@@ -49,7 +50,7 @@ on TimeLock.
 
 .. code:: bash
 
-      curl -XPOST localhost:8080/timelock/api/new_namespace/timestamp-management/fast-forward?currentTimestamp=TS
+      curl -XPOST https://localhost:8421/timelock/api/new_namespace/timestamp-management/fast-forward?currentTimestamp=TS
```
lycarter, line 53: here you refer directly to `https` and `localhost` while in other files you refer to `<protocol>` and `<host>` - probably best to be consistent?

lycarter: As an unrelated comment, it is a bit surprising to me that this method is unauth'd - wouldn't this cause an issue if the endpoint was called with `MAX_LONG`?

lycarter (commented):
Thanks for the docs update!

docs/source/services/timelock_service/service-management.rst
jeremyk-91: Ah, good catch. Yeah, I'll edit this: it can be either

docs/source/services/timelock_service/service-management.rst
jeremyk-91: Yeah, will switch to host/port. 

You're right that there is an issue with long max. However, internally we configure our proxies so that this method is never exposed in a way where it can be called from non-Palantir applications, so that's why this hasn't been prioritised.

gmaretic (approved):
Looks good overall, make sure to address the CR but otherwise should be good to ship

docs/source/services/timelock_service/manual-migration.rst
jeremyk-91: Yep. I added the `-i` to display headers to the commands too.

docs/source/services/timelock_service/reverse-migration.rst
jeremyk-91: I think Grgur means below, but yep

PR #5001, opened 2020-09-25
[TS] Enable TS autotuning

gmaretic:
**Goals (and why)**:
Since this was performing quite well on altas-proxy, I don't see a compelling reason to put off on enabling it everywhere


jeremyk-91 (approved):
🚀 release the kraken!

PR #5006, opened 2020-09-29
[LW] Rename IdentifiedVersion

gmaretic:
**Goals (and why)**:
Rename `IdentifiedVersion` to something more descriptive

Straightforward

Existing tests pass

I have checked that this object is never serialised outside of our tests, but would appreciate a sanity check as this is definitely not worth a wire format break. In particular, `StartTransactionRequestV5` is not serialised anywhere and will be removed in #5005 

It's small

Let's get his in today pls

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchTestRuntimeConfig.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchTestRuntimeConfig.java
@@ -31,11 +31,8 @@ import com.palantir.atlasdb.keyvalue.api.TableReference;
 @JsonDeserialize(as = ImmutableLockWatchTestRuntimeConfig.class)
 @Value.Immutable
 public abstract class LockWatchTestRuntimeConfig {
-    @Value.Default
     @JsonProperty("namespace")
-    Optional<String> namespaceToWatch() {
-        return Optional.empty();
-    }
+    abstract Optional<String> namespaceToWatch();
 
     @Value.Default
     @JsonProperty("tables")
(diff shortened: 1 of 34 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchTestRuntimeConfig.java
```diff
@@ -31,11 +31,8 @@
 @JsonDeserialize(as = ImmutableLockWatchTestRuntimeConfig.class)
 @Value.Immutable
 public abstract class LockWatchTestRuntimeConfig {
-    @Value.Default
     @JsonProperty("namespace")
-    Optional<String> namespaceToWatch() {
-        return Optional.empty();
-    }
+    abstract Optional<String> namespaceToWatch();
```
Jolyon-S, line 35: Looks like a separate drive by?

Jolyon-S (commented):
For the most part, this looks like a simple rename (and the odd replacement of the Immutable `.of()` method with the one in the interface), and that looks fine. Just want to sanity check the other drive-by before approving.

Also, do be aware that atlasdb-proxy uses IdentifiedVersion too, so this is technically breaking and we'll need to fix that over there too.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/lock/watch/LockWatchTestRuntimeConfig.java
gmaretic: I got a warning while running checkstyle, so cleaned it up :P

gmaretic:
Yup, I'm aware of the proxy break, but it isn't being serialised there either right?

Jolyon-S (approved):
Just checked, and we don't serialize this anywhere in proxy either => good to go here

PR #5011, opened 2020-09-30
[DB TimeLock] 1B, 1C: DB TimeLock for Multiple Series

jeremyk-91:
**Goals (and why)**:
- Allow DB Timelock to support multiple namespaces and behave correctly in these circumstances.
- Don't actually turn it on.

- Change the timestamp bound store to accept a _strategy_ that in its legacy form reads/writes from a simple table
- And in the new format has a `namespace -> last_allocated` mapping.

- Existing ETEs/integrations should valid nothing broke here: we are still using the old impl.
- I added some tests specific to the new bound store impl

- The SQL handling code is laden with technical debt, but given timelines I don't really see us cleaning this up right now. I'd consider dedicating two days post-workstream to clean up this path.
- There may be some duplication when we create the Oracle version of the Postgres multi series test, though an abstract test class hierarchy here felt like more complexity than I would like.

**Where should we start reviewing?**: MultiSeriesDbTimestampBoundStore

**Priority (whenever / two weeks / yesterday)**: today would be nice, but tomorrow is fine too

### DIFF
--- a/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
+++ b/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
@@ -70,6 +70,16 @@ public class InDbTimestampBoundStore implements TimestampBoundStore {
         return inDbTimestampBoundStore;
     }
 
+    public static InDbTimestampBoundStore createMultiTableForSeries(
+            ConnectionManager connManager,
+            TableReference timestampTable,
+            String series) {
+        InDbTimestampBoundStore inDbTimestampBoundStore = new InDbTimestampBoundStore(connManager,
+                new MultiSequencePhysicalBoundStoreStrategy(timestampTable, series));
+        inDbTimestampBoundStore.init();
+        return inDbTimestampBoundStore;
+    }
+
     private InDbTimestampBoundStore(ConnectionManager connManager,
             PhysicalBoundStoreStrategy physicalBoundStoreStrategy) {
         this.connManager = Preconditions.checkNotNull(connManager, "connectionManager is required");
@@ -157,7 +167,7 @@ public class InDbTimestampBoundStore implements TimestampBoundStore {
     @Override
     public synchronized void storeUpperLimit(final long limit) {
         runOperation((connection, oldLimit) -> {
-            if (oldLimit != null) {
+            if (oldLimit.isPresent()) {
                 physicalBoundStoreStrategy.writeLimit(connection, limit);
             } else {
                 physicalBoundStoreStrategy.createLimit(connection, limit);
--- a/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
+++ b/atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
@@ -16,19 +16,12 @@
 package com.palantir.atlasdb.keyvalue.dbkvs.timestamp;
 
 import java.sql.Connection;
-import java.sql.PreparedStatement;
 import java.sql.SQLException;
-import java.sql.Statement;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 
-import org.apache.commons.dbutils.QueryRunner;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.palantir.atlasdb.keyvalue.api.TableReference;
-import com.palantir.atlasdb.keyvalue.dbkvs.OracleErrorConstants;
 import com.palantir.common.base.Throwables;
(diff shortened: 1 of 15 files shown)


atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
```diff
@@ -72,32 +62,39 @@ public static InDbTimestampBoundStore create(
...
+        return createWithStrategy(connManager, new LegacyPhysicalBoundStoreStrategy(timestampTable, tablePrefixString));
+    }
 
-        inDbTimestampBoundStore.init();
+    public static InDbTimestampBoundStore createForMultiSeries(
+            ConnectionManager connManager,
+            TableReference timestampTable,
+            String series) {
+        return createWithStrategy(connManager, new MultiSequencePhysicalBoundStoreStrategy(timestampTable, series));
+    }
```
sudiksha27, line 73: I see we're removing the option of providing tablePrefix for multi series, which should be fine. Maybe we should should use Client rather than String for type checking?

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/MultiSequencePhysicalBoundStoreStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.commons.dbutils.QueryRunner;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.nexus.db.DBType;
+
+public class MultiSequencePhysicalBoundStoreStrategy implements PhysicalBoundStoreStrategy {
+    private static final Logger log = LoggerFactory.getLogger(MultiSequencePhysicalBoundStoreStrategy.class);
+
```
sudiksha27, line 34: Same as above.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/PhysicalBoundStoreDatabaseUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    private static void createTimestampTableIgnoringAlreadyExistsError(
+            Statement statement, String oracleQuery) throws SQLException {
+        try {
+            statement.execute(oracleQuery);
+        } catch (SQLException e) {
+            if (!e.getMessage().contains(OracleErrorConstants.ORACLE_ALREADY_EXISTS_ERROR)) {
+                log.error("Error occurred creating the Oracle timestamp table", e);
```
sudiksha27, line 58: Maybe add a comment that pg query to create table already accommodates this check?

sudiksha27 (approved):
Minor comments, looks good to go otherwise!

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/LegacyPhysicalBoundStoreStrategy.java
```diff
@@ -0,0 +1,85 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.logsafe.Preconditions;
+import com.palantir.nexus.db.DBType;
+
+public class LegacyPhysicalBoundStoreStrategy implements PhysicalBoundStoreStrategy {
+    private static final Logger log = LoggerFactory.getLogger(LegacyPhysicalBoundStoreStrategy.class);
+
```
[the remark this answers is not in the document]
jeremyk-91: Doh! Yeah, this went away since I extracted this before factoring out the `createTimestampTable()` stuff. Removed

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/PhysicalBoundStoreDatabaseUtils.java
jeremyk-91: Alright, added a comment below this if-block. I think it makes sense actually, since it looks a bit like a bug at first that you don't throw the exception in that case.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/timestamp/InDbTimestampBoundStore.java
jeremyk-91: Ah. We can't use Client, because that's a part of leader-election-api and this is in atlasdb-dbkvs. However, I agree that having a string wrapper type is a good suggestion, I'll add one.

PR #5013, opened 2020-10-01

Jolyon-S:
**Goals (and why)**:

* WIP

Let's get this rolling by EOW.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -51,31 +52,36 @@ CacheUpdate processUpdate(LockWatchStateUpdate update) {
...
-            LockWatchVersion endVersion) {
-        Optional<LockWatchVersion> startVersion = lastKnownVersion.map(this::createStartVersion);
-        LockWatchVersion currentVersion = getLatestVersionAndVerify(endVersion);
+    public ClientLogEvents getEventsBetweenVersions(VersionBounds versionBounds) {
+        Optional<LockWatchVersion> startVersion = versionBounds.startVersion().map(this::createStartVersion);
+        LockWatchVersion currentVersion = getLatestVersionAndVerify(versionBounds.endVersion());
 
         if (!startVersion.isPresent() || differentLeaderOrTooFarBehind(currentVersion, startVersion.get())) {
+            long snapshotVersion = versionBounds.snapshotVersion();
+            Collection<LockWatchEvent> afterSnapshotEvents = eventStore.getEventsBetweenVersionsInclusive(
```
Jolyon-S, line 69: Let's assume that the user is on version `U`, the cache's earliest is `V` (so its internal snapshot is on that version), and the timestamp mapping contains versions from `W` to `X`.

Case 1: `snapshotVersion < endVersion` (i.e. `W < X`):
The condensed snapshot will contain the information from `V`, the base snapshot, up to and including `W` - therefore transactions on this earliest version will contain exactly the info they need, and any on >`W` will have the additional events that they need (which is computed here, events `W + 1` -> `X`).

Case 2: `snapshotVersion == endVersion` (or `W == X`):
In this case, there are no further events we need to retrieve. The extra change in the `eventStore` makes asking for `X+1, X` a valid request (where before it would throw); I felt it just made sense to push that down so we don't have to worry about it here.

That should be all cases covered.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -238,24 +238,22 @@ public void cacheClearedOnSnapshotUpdate() {
     }
 
     @Test
-    public void getEventsForTransactionsReturnsSnapshotWithOldEvents() {
+    public void getEventsForTransactionsReturnsSnapshotWithCondensedEvents() {
```
Jolyon-S, line 241: this test keeps having its behaviour changed, but it didn't make a huge amount of sense with the new scheme to leave as-is.

Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/5015

PR #5014, opened 2020-10-02
[DB TimeLock] 1D - Agent Wiring

jeremyk-91:
**Goals (and why)**:
- DB Timelock needs to support multiple series

- Set db timelock up to use the new multi series bound store.
- Pass the series in as a client.

The ETEs validate that nothing has truly imploded. There are some small unit tests. More involved testing should be handled in task 3 of this workstream.

- `DbTimestampCreationParameters` would really better be served by a Visitor pattern and subtypes, though I think that's a bit too much work and not particularly valuable. What do you think?
- Some duplication in the checks for Cassandra and JDBC. I will separately check if we can kill the latter, this is frustrating. Impl-Shared is strangely not pulled in by JDBC and there aren't many choices on where
- The testing setup for this PR is admittedly limited, and I do not like it. That said, we will not roll this out until it's further improved, and this code path is not used in production anywhere, and I think this component is reviewable by itself.

**Where should we start reviewing?**: DbBoundTimestampCreator is the entrypoint and probably where you should drill down from.

**Priority (whenever / two weeks / yesterday)**: this week for a first pass (though not the highest priority as I'll be back in on Monday)


atlasdb-api/src/main/java/com/palantir/atlasdb/config/DbTimestampCreationSetting.java
```diff
@@ -0,0 +1,41 @@
...
+@Data
+public abstract class DbTimestampCreationSetting {
+    public interface Cases<R> {
+        R multipleSeries(Optional<TableReference> tableReference, TimestampSeries series);
+        R singleSeries(Optional<TableReference> tableReference);
+    }
+
+    public abstract <R> R match(Cases<R> cases);
+
+    @Override
```
gmaretic, line 35: Note: these indicate that the methods should be generated

timelock-agent/src/main/java/com/palantir/timelock/paxos/DbBoundTimestampCreator.java
```diff
@@ -49,15 +52,18 @@ public DbBoundTimestampCreator(KeyValueServiceConfig kvsConfig) {
...
-        Preconditions.checkArgument(timestampService instanceof TimestampManagementService,
-                "The timestamp service is not a managed timestamp service.");
 
         return () -> new DelegatingManagedTimestampService(timestampService,
                 (TimestampManagementService) timestampService);
     }
+
+    @VisibleForTesting
+    static DbTimestampCreationSetting getTimestampCreationParameters(Client client) {
+        return DbTimestampCreationSettings.multipleSeries(Optional.empty(), TimestampSeries.of(client.value()));
```
gmaretic, line 67: Is it safe to just switch here?

gmaretic (approved):
LGTM, just have one question that is really relevant. Otherwise double check that you agree with my changes and ship

atlasdb-api/src/main/java/com/palantir/atlasdb/config/DbTimestampCreationSetting.java
jeremyk-91: interesting - I guess you wouldn't want them if you aren't testing or putting them in sets, though it does feel a bit awkward that these aren't by default

timelock-agent/src/main/java/com/palantir/timelock/paxos/DbBoundTimestampCreator.java
jeremyk-91: Yes. This isn't currently used in production anywhere.

jeremyk-91:
👍

PR #5015, opened 2020-10-02

Jolyon-S:
**Goals (and why)**:

Current tests covered most of this. However, one of the tests did have to change.

Let's get this rolling by EOW.


Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/5021

PR #5021, opened 2020-10-12

Jolyon-S:
**Goals (and why)**:

Current tests covered most of this. However, one of the tests did have to change.

This week.


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -236,25 +236,23 @@ public void cacheClearedOnSnapshotUpdate() {
     }
 
     @Test
-    public void getEventsForTransactionsReturnsSnapshotWithOldEvents() {
+    public void getEventsForTransactionsReturnsSnapshotWithCondensedEvents() {
```
jkozlowski, line 239: Can you also add tests here? I think there's a bunch of edge cases to cover around the snapshot computation piece? Or add unit tests, whatever works for you.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLockWatchSnapshot.java
```diff
@@ -61,6 +61,13 @@ private ClientLockWatchSnapshot() {
                 ImmutableSet.copyOf(watches));
     }
 
+    LockWatchStateUpdate.Snapshot getSnapshotWithEvents(LockWatchEvents events, UUID versionId) {
```
jkozlowski, line 64: Check continuity?

jkozlowski (commented):
Releasing comments

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -100,7 +105,13 @@ public TransactionsLockWatchUpdate getUpdateForTransactions(
         lastKnownVersion.ifPresent(version -> assertTrue(version.version() <= timestampMapping.lastVersion().version(),
```
jkozlowski, line 105: For completeness, can you also remove the "Preconditions.checkArgument(!startTimestamps.isEmpty(), "Cannot get events for empty set of transactions");" here; it's checked in the TimestampMapping

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -49,31 +51,36 @@ CacheUpdate processUpdate(LockWatchStateUpdate update) {
     }
 
     /**
-     * @param lastKnownVersion latest version that the client knows about; should be before timestamps in the mapping.
-     * @param endVersion       mapping from timestamp to identified version from client-side event cache.
-     * @return lock watch events that occurred from (exclusive) the provided version, up to the end version (inclusive)
+     * @param versionBounds contains:
```
jkozlowski, line 54: This comment should just live on VersionBounds, if you really care about it.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -77,7 +77,10 @@ VersionedEventStoreState getStateForTesting() {
                 .build();
     }
 
-    private Collection<LockWatchEvent> getValuesBetweenInclusive(long endVersion, Long startVersion) {
+    private Collection<LockWatchEvent> getValuesBetweenInclusive(long endVersion, long startVersion) {
+        if (endVersion < startVersion) {
```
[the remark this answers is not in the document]
Jolyon-S: Fair enough

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLockWatchSnapshot.java
Jolyon-S: Factored out a check so that this is always the case (in `ClientLockWatchSnapshot` normally as well as in this case).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
Jolyon-S: Good point - removed

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
Jolyon-S: we now have tests for:
1. no condensing
2. some condensing
3. max condensing

I think that covers all the main cases.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -49,35 +51,41 @@ CacheUpdate processUpdate(LockWatchStateUpdate update) {
...
-        Optional<LockWatchVersion> startVersion = lastKnownVersion.map(this::createStartVersion);
-        LockWatchVersion currentVersion = getLatestVersionAndVerify(endVersion);
+    public ClientLogEvents getEventsBetweenVersions(VersionBounds versionBounds) {
+        Optional<LockWatchVersion> startVersion = versionBounds.startVersion().map(this::createStartVersion);
+        LockWatchVersion currentVersion = getLatestVersionAndVerify(versionBounds.endVersion());
 
         if (!startVersion.isPresent() || differentLeaderOrTooFarBehind(currentVersion, startVersion.get())) {
+            long snapshotVersion = versionBounds.snapshotVersion() + 1;
+            Collection<LockWatchEvent> afterSnapshotEvents;
+            if (snapshotVersion > versionBounds.endVersion().version()) {
```
Jolyon-S, line 65: heh, this is `long > Long` which I BELIEVE IS FINE.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -86,18 +85,28 @@ public CommitUpdate getCommitUpdate(long startTs) {
...
+        VersionBounds versionBounds = new VersionBounds.Builder()
+                .startVersion(startVersion)
+                .endVersion(commitInfo.commitVersion())
+                .build();
+
+        return eventLog.getEventsBetweenVersions(versionBounds).toCommitUpdate(startVersion.get(), commitInfo);
     }
 
     @Override
     public TransactionsLockWatchUpdate getUpdateForTransactions(
```
jkozlowski, line 98: Would we get anything in terms of simplification of code if this was taking Long startTimestamp, instead of Set<Long>? atlas-proxy just needs one transaction pushed here.

Jolyon-S: Yes, that would be simpler, but I don't think by leaps and bounds - it just simplifies what we can compress a little.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -86,18 +85,28 @@ public CommitUpdate getCommitUpdate(long startTs) {
...
             Set<Long> startTimestamps,
             Optional<LockWatchVersion> lastKnownVersion) {
-        Preconditions.checkArgument(!startTimestamps.isEmpty(), "Cannot get events for empty set of transactions");
```
Jolyon-S, line 97: This check lives in `TimestampMapping`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
```diff
@@ -49,35 +52,44 @@ CacheUpdate processUpdate(LockWatchStateUpdate update) {
...
         if (!startVersion.isPresent()
                 || differentLeaderOrTooFarBehind(currentVersion, lastKnownVersion.get(), startVersion.get())) {
+            long snapshotVersion = versionBounds.snapshotVersion() + 1;
+            Collection<LockWatchEvent> afterSnapshotEvents;
+            if (snapshotVersion > versionBounds.endVersion().version()) {
+                afterSnapshotEvents = ImmutableList.of();
+            } else {
+                afterSnapshotEvents = eventStore.getEventsBetweenVersionsInclusive(
+                        Optional.of(snapshotVersion), versionBounds.endVersion().version());
+            }
```
Jolyon-S, line 72: We want to get the events _after_ the snapshot only

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -296,30 +296,72 @@ public void cacheClearedOnSnapshotUpdate() {
...
-
+        eventCache.processStartTransactionsUpdate(ImmutableSet.of(), SUCCESS);
         eventCache.processStartTransactionsUpdate(TIMESTAMPS_2, SUCCESS_2);
         verifyStage();
 
-        TransactionsLockWatchUpdate results = eventCache.getUpdateForTransactions(TIMESTAMPS_2, Optional.empty());
+        // Client is behind and needs a snapshot, but no compression can be done as the first transaction is at the
+        // first version
+        TransactionsLockWatchUpdate results =
+                eventCache.getUpdateForTransactions(Sets.union(TIMESTAMPS, TIMESTAMPS_2), Optional.empty());
```
gmaretic, line 308: Can we make these tests more legible by not using seemingly magic numbers?
I.e.,
```suggestion
               eventCache.getUpdateForTransactions(ImmutableSet.of(START_TS, TS_16), Optional.empty());
```
and then using those in the asserts, and similarly below not using the magic `LockWatchVersion` but call `logId` and `lastKnownVersion` on the relevant updates?

gmaretic (changes requested):
Overall makes sense, have some nits and absolutely blocking `Builder` definitions

gmaretic:
In general, maybe at some time in the future we can consider optimising this by caching the partial snapshots or something similar, but this is probably fine for now

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEvents.java
```diff
@@ -62,5 +64,22 @@ default void rangeOnlyPresentIffEventsAre() {
         }
     }
 
+    default void assertNoEventsAreMissing(Optional<LockWatchVersion> latestVersion) {
```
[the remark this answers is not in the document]
Jolyon-S: Ha! Good point, actually.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEvents.java
```diff
@@ -62,5 +64,22 @@ default void rangeOnlyPresentIffEventsAre() {
...
+        if (events().isEmpty()) {
+            return;
+        }
+
+        if (latestVersion.isPresent()) {
+            Preconditions.checkArgument(versionRange().isPresent(), "First element not preset in list of events");
```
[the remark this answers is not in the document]
Jolyon-S: Not only because I only make a range if the events are non-empty, but also because of the other check. Removed.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEvents.java
```diff
@@ -62,5 +64,22 @@ default void rangeOnlyPresentIffEventsAre() {
...
+            long firstVersion = versionRange().get().lowerEndpoint();
+            Preconditions.checkArgument(
+                    firstVersion <= latestVersion.get().version()
+                            || latestVersion.get().version() + 1 == firstVersion,
+                    "Events missing between last snapshot and this batch of events",
+                    SafeArg.of("latestVersionSequence", latestVersion.get().version()),
+                    SafeArg.of("firstNewVersionSequence", firstVersion));
+        }
+    }
+
```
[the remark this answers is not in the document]
Jolyon-S: Don't worry, I've come to see the sins of my ways.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEvents.java
```diff
@@ -62,5 +64,22 @@ default void rangeOnlyPresentIffEventsAre() {
...
+    default void assertNoEventsAreMissing(Optional<LockWatchVersion> latestVersion) {
+        if (events().isEmpty()) {
+            return;
+        }
+
+        if (latestVersion.isPresent()) {
+            Preconditions.checkArgument(versionRange().isPresent(), "First element not preset in list of events");
+            long firstVersion = versionRange().get().lowerEndpoint();
+            Preconditions.checkArgument(
+                    firstVersion <= latestVersion.get().version()
```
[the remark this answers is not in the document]
Jolyon-S: That's not the same check - what if latest version is > the upper bound of the range?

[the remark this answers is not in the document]
Jolyon-S: Although, on review of the code, I don't _think_ that can happen.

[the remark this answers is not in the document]
Jolyon-S: but to be safe, I'd prefer to do this.

atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/getEventsForTransactionsNoCondensing/event-cache-1.json
```diff
@@ -0,0 +1,78 @@
...
+      "1" : {
+        "version" : {
+          "id" : "470c855e-f77b-44df-b56a-14d3df085dbc",
+          "version" : 3
+        },
+        "commitInfo" : null
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
Jolyon-S: No - these are generated, so it doesn't make sense to add a newline.

atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/getEventsForTransactionsSomeCondensing/event-cache-1.json
```diff
@@ -48,6 +48,21 @@
...
+      "1" : {
+        "version" : {
+          "id" : "470c855e-f77b-44df-b56a-14d3df085dbc",
+          "version" : 3
+        },
+        "commitInfo" : null
+      }
+    }
   }
 }
```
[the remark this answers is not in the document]
Jolyon-S: Same as above

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
Jolyon-S: Done

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEvents.java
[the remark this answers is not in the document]
gmaretic: I know it's not equivalent, but the method name `assertNoEventsAreMissing` implies they are connected. So then change the method name to `assertNoEventsAreMissingAfterLatestVersion`?

gmaretic (approved):
One nit remaining, but doesn't have to be actioned

PR #5024, opened 2020-10-12
[DB TimeLock] 3B.1 - DB TimeLock Bugfixes and Preparation for MNPTLSIT

jeremyk-91:
**Goals (and why)**:
- DB timelock needs to be a mostly full fledged timelock server. In particular, MNPTLSIT which is our gold standard for correct behaviour of the standard timelock needs to pass with a DB-backed timelock.

- Fix bugs in DB TimeLock. Specifically:
  - we used to create one instance of the KVS (and hence one instance of the Hikari connection pool) per client; now we only create one for the entire life of TimeLock.
  - we used to not close the KVS when timelock stops; we now do.
  - we would silently fail and give the user fewer namespaces if they decided to call the getAllNamespaces() endpoint. We now throw, which is a safer solution.

This PR doesn't add tests, though MNPTLSIT does work when I wired it up (changing the config.ftl and hand-jamming the dbkvs rule in the testable cluster's rule chain). I plan to add the actual testing with MNPTLSIT in a later PR but that's also blocked on #5014 for us to do anything meaningful.

- This PR is not optimal in that the Postgres test rule infra should really be its own PR, but I needed it to verify that MNPTLSIT works as I expect it to.
- Does the choice of more opinionated defaults in `ServiceDiscoveringDatabaseTimeLockSupplier` make sense?
- We now fail fast if user calls getAllNamespaces(). Are there unintended consequences I'm missing here?

**Where should we start reviewing?**: DbBoundTimestampCreator.java

**Priority (whenever / two weeks / yesterday)**: yesterday 🔥  current SOTW is really buggy, and this blocks almost all my further work here


timelock-agent/src/main/java/com/palantir/timelock/paxos/DbBoundTimestampCreator.java
```diff
@@ -18,46 +18,45 @@
...
+    private DbBoundTimestampCreator(
+            ServiceDiscoveringDatabaseTimeLockSupplier serviceDiscoveringDatabaseTimeLockSupplier) {
+        this.serviceDiscoveringDatabaseTimeLockSupplier = serviceDiscoveringDatabaseTimeLockSupplier;
+    }
+
+    public static TimestampCreator create(
+            KeyValueServiceConfig kvsConfig,
+            MetricsManager metricsManager,
+            Supplier<Optional<KeyValueServiceRuntimeConfig>> kvsRuntimeConfig,
+            LeaderConfig leaderConfig) {
```
sudiksha27, line 43: When do we have runtimeConfig?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/management/TimeLockManagementResource.java
```diff
@@ -101,6 +100,9 @@ public static TimeLockManagementService jersey(
 
     private static Set<PersistentNamespaceLoader> createNamespaceLoaders(
             PersistentNamespaceContext persistentNamespaceContext) {
+        if (persistentNamespaceContext.databasePersistence().isPresent()) {
+            return ImmutableSet.of(new DatabaseNamespaceLoader());
+        }
```
sudiksha27, line 105: Are we using the kvsConfig OR plan on using it? If so, It might be worth adding a comment otherwise, I am skeptical of passing around the config.

sudiksha27 (commented):
The restructure + refactor looks okay at a high level. I do have some small questions.

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimestampCreator.java
```diff
@@ -21,6 +21,11 @@
...
 import com.palantir.timestamp.ManagedTimestampService;
 
-public interface TimestampCreator {
+public interface TimestampCreator extends AutoCloseable {
     Supplier<ManagedTimestampService> createTimestampService(Client client, LeaderConfig leaderConfig);
+
+    @Override
+    default void close() {
+        // Do nothing
+    }
```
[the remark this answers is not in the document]
sudiksha27: I meant the latter, apologies it was not clear; I thought it would be good to force overriding close() so it can not be ignored.

sudiksha27 (approved):
Broadly makes sense, should be good to go if we have verified this refactor with MNPTLSIT.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
```diff
@@ -66,7 +63,7 @@ public ServiceDiscoveringAtlasSupplier(
             LongSupplier timestampSupplier) {
```
gmaretic, line 63: Not in scope (?) but the inversion of parameters here and in `atlasFactory.createRawKeyValueService` is infuriating

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/management/DatabaseNamespaceLoader.java
```diff
@@ -0,0 +1,28 @@
...
+package com.palantir.atlasdb.timelock.management;
+
+import java.util.Set;
+
+import com.palantir.paxos.Client;
+
+public class DatabaseNamespaceLoader implements PersistentNamespaceLoader {
+    @Override
+    public Set<Client> getAllPersistedNamespaces() {
+        throw new UnsupportedOperationException("Loading all namespaces from the database is not supported yet!");
```
gmaretic, line 26: This will eventually be implemented?

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimestampCreator.java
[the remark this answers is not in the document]
gmaretic: I agree with no default here

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/DbKvsRule.java
```diff
@@ -0,0 +1,128 @@
...
+                        || ex.getMessage().contains("the database system is starting up")) {
+                    return false;
+                } else {
+                    throw ex;
+                }
+            } finally {
+                if (kvs != null) {
+                    kvs.close();
+                }
+            }
```
gmaretic, line 121:
```suggestion
            try (ConnectionManagerAwareDbKvs kvs = createKvs()) {
                try (Connection connection = kvs.getConnectionManager().getConnection()) {
                    return connection.isValid(FIVE_SECONDS);
                }
            } catch (Exception ex) {
                if (ex.getMessage().contains("The connection attempt failed.")
                        || ex.getMessage().contains("the database system is starting up")) {
                    return false;
                } else {
                    throw ex;
                }
            }
```

gmaretic (approved):
LGTM, a couple of nits but not blocking

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimestampCreator.java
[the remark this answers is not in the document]
jeremyk-91: Cool, added no-op close to PaxosTimestampCreator and removed the default impl. Makes sense

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
jeremyk-91: Yeah, now's not a good time to touch this class's API (and above)

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/management/DatabaseNamespaceLoader.java
jeremyk-91: Yep! I probably should have given a bit more context. Internally this part of the workstream has two phases: 3B (which is ensuring db timelock runs single client namespace integration tests correctly) and 3C (which is giving full compatibility with MNPTLSIT, including the getAllNamespaces tests). Eventually in 3C.3 this will be implemented :)

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/DbKvsRule.java
jeremyk-91: Good call, yeah I copypasted this one

PR #5025, opened 2020-10-12
[DB TimeLock] 3B.2 - DbTimeLockSingleLeaderPaxosSuite

jeremyk-91:
**Goals (and why)**:
- Gather confidence that db timelock works well.

- Wire up DB TimeLock to MNPTLSIT which we consider to be our benchmark for basic timelock correctness.

- This PR is about tests!

- This reruns a bunch of tests that are quite pointless because they test leadership code. I excluded the most egregious exceptions.
- I assumeFalse()d the namespace loader since it really doesn't make any sense until #5014 

**Where should we start reviewing?**: DbTimelockSingleLeaderPaxosSuite

### DIFF
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -78,7 +78,7 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
 
     @ClassRule
     public static ParameterInjector<TestableTimelockCluster> injector =
-            ParameterInjector.withFallBackConfiguration(() -> DbTimeLockSingleLeaderPaxosSuite.DB_TIMELOCK_CLUSTER);
+            ParameterInjector.withFallBackConfiguration(() -> SingleLeaderPaxosSuite.BATCHED_TIMESTAMP_PAXOS);
 
     @Parameterized.Parameter
     public TestableTimelockCluster cluster;
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -50,7 +50,6 @@ import com.palantir.atlasdb.timelock.api.ConjureLockToken;
 import com.palantir.atlasdb.timelock.api.ConjureUnlockRequest;
 import com.palantir.atlasdb.timelock.api.SuccessfulLockResponse;
 import com.palantir.atlasdb.timelock.api.UnsuccessfulLockResponse;
-import com.palantir.atlasdb.timelock.suite.DbTimeLockSingleLeaderPaxosSuite;
 import com.palantir.atlasdb.timelock.suite.SingleLeaderPaxosSuite;
 import com.palantir.atlasdb.timelock.util.ExceptionMatchers;
 import com.palantir.atlasdb.timelock.util.ParameterInjector;
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -50,6 +50,7 @@ import com.palantir.atlasdb.timelock.api.ConjureLockToken;
 import com.palantir.atlasdb.timelock.api.ConjureUnlockRequest;
 import com.palantir.atlasdb.timelock.api.SuccessfulLockResponse;
 import com.palantir.atlasdb.timelock.api.UnsuccessfulLockResponse;
+import com.palantir.atlasdb.timelock.suite.DbTimeLockSingleLeaderPaxosSuite;
 import com.palantir.atlasdb.timelock.suite.SingleLeaderPaxosSuite;
 import com.palantir.atlasdb.timelock.util.ExceptionMatchers;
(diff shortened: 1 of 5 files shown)


timelock-server/src/testCommon/resources/dbTimeLockPaxosMultiServer.ftl
```diff
@@ -0,0 +1,76 @@
...
+      ddl:
+        type: "postgres"
+      connection:
+        type: "postgres"
+        host: "localhost"
+        port: 5432
+        dbName: "atlas"
+        dbLogin: "palantir"
+        dbPassword: "palantir"
+
```
sudiksha27, line 34: Not a big fan of having these constants in 2 places but it may be fine for this test.

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TestableTimelockServerConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Value.Immutable
+public interface TestableTimelockServerConfiguration {
+    TemplateVariables templateVariables();
+
+    @Value.Default
+    default boolean needsPostgresDatabase() {
+        return false;
+    }
+}
```
sudiksha27, line 29: I wonder if this the best we can do design wise, I feel like having these layers makes code a bit harder to read.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -469,6 +470,7 @@ private static void assertNumberOfThreadsReasonable(int startingThreads, int thr
 
     private void abandonLeadershipPaxosModeAgnosticTestIfRunElsewhere() {
         Assume.assumeTrue(cluster == FIRST_CLUSTER);
+        Assume.assumeFalse(cluster.isDbTimelock()); // We will never test only DB timelock when releasing.
     }
```
sudiksha27, line 474: I did not understand the comment, could you please elaborate on that.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/suite/DbTimeLockSingleLeaderPaxosSuite.java
```diff
@@ -0,0 +1,70 @@
...
+                    .isUseBatchPaxosTimestamp(false)
+                    .isBatchSingleLeader(true))
+                    .leaderMode(PaxosInstallConfiguration.PaxosLeaderMode.SINGLE_LEADER));
+    private static final List<TestableTimelockServerConfiguration> TESTABLE_CONFIGURATIONS = StreamSupport.stream(
+            TEMPLATE_VARIABLES.spliterator(), false)
+            .map(variables -> ImmutableTestableTimelockServerConfiguration.builder()
+                    .templateVariables(variables)
+                    .needsPostgresDatabase(true)
+                    .build())
+            .collect(Collectors.toList());
```
sudiksha27, line 55: It may be worth extracting out a utility method that takes templateVariables and needsPG as parameters? There is similar code in TestableTimelockCluster.

sudiksha27 (commented):
Looks right, some questions for my better understanding.

timelock-server/src/testCommon/resources/dbTimeLockPaxosMultiServer.ftl
jeremyk-91: The docker-compose files are different, so it's in theory possible to change these independently of what you see in the DbKVS tests. It could make sense to extract a common DbKVS TestRule for both the DbKVS and DB-TimeLock tests though I don't think we have the bandwidth for that now.

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TestableTimelockServerConfiguration.java
jeremyk-91: What alternatives do you have in mind? The main ones I'm aware of are:
- passing a boolean and/or context directly into the testable timelock cluster (also OK with that, though it seems the property of whether you need a postgres is dependent on an individual server's config)
- having a template variable that is the yml-serialized form of the database bound persistence (don't like this; the freemarker templates don't natively support structured formats other than xml and html I think)

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: There are some slow tests that aren't dependent on timestamp bound persistence (check where this is called), and so this skips them when running in a db timelock environment. This is only safe if we are sure it is tested elsewhere (which it is, because we will never test only db timelock when releasing).

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/suite/DbTimeLockSingleLeaderPaxosSuite.java
jeremyk-91: Sure, can do

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TestableTimelockServerConfiguration.java
sudiksha27: I see, what you have now is fine.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
sudiksha27: Makes sense

sudiksha27 (approved):
LGTM!

PR #5026, opened 2020-10-13
Double check that the SQLite migration is consistent with current state

gmaretic:
**Goals (and why)**:
If we upgraded to version with SQLite paxos log, downgraded to a version before, used timelock for a bit, and then upgraded again, migration would not run because it had already run. Normally, the downgrade itself would have already cause data corruption if allowed, but it is possible to restore from backup to a previous version, and then upgrade after running on the restored version at which point the migration would silently noop and we would start using the pre-restore SQLite paxos state log. This PR verifies the consistency whenever we detect that the migration has already run.

We have imperfect information and do not want to effectively redo the migration on every startup, so we do the following:

1. Verify the cutoff that was persisted matches a newly calculated one. This guarantees the learner has not advanced except in the special case where the greatest log entry is was less than 50 at migration time and then advanced but is still less than 50.
2. Verify that the greatest log entry in the legacy log is not larger than the one in SQLite, and also verify that the entry is present in SQLite and equal. This covers both the special case above and the acceptor as acceptor uses the cutoff from the learner.

Also needed to implement equality for `PaxosAcceptorState.java`, which has to not include the version which is not persisted. Not in love with this, but beats the alternatives.

Unit tests for the migrator and for the PAS equality.

Note that (2) above is not foolproof as it is possible that the legacy log wrote some new entries and then happened to write the latest one to exactly mach one that SQLite wrote after the migration while there are some in between that do not match. I am not aware of a way to do this with more certainty while not also effectively redoing the migration on every startup, which is prohibitively expensive. An alternative approach would be to have a new migration state table where we persist the greatest migrated sequence as well as cutoff, and would then have to migrate all already migrated deployments, but that seems excessive since the chance of (2) failing while also (1) passing is very low.

PaxosStateLogMigrator.java

ASAP

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorState.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorState.java
@@ -16,6 +16,7 @@
 package com.palantir.paxos;
 
 import com.google.common.base.Defaults;
+import com.google.common.base.Objects;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.palantir.common.annotation.Immutable;
 import com.palantir.common.base.Throwables;
@@ -116,5 +117,39 @@ public final class PaxosAcceptorState implements Persistable, Versionable {
 
     public PaxosValue getLastAcceptedValue() {
         return lastAcceptedValue;
+    }
+
+    /**
+     * Standard equals implementation, note that {@link #version} is not persisted and as such is not used in the
+     * calculation.
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        PaxosAcceptorState other = (PaxosAcceptorState) obj;
+        if (!Objects.equal(lastPromisedId, other.lastPromisedId)) {
+            return false;
+        }
+        if (!Objects.equal(lastAcceptedId, other.lastAcceptedId)) {
+            return false;
+        }
+        return Objects.equal(lastAcceptedValue, other.lastAcceptedValue);
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((lastPromisedId == null) ? 0 : lastPromisedId.hashCode());
+        result = prime * result + ((lastAcceptedId == null) ? 0 : lastAcceptedId.hashCode());
+        return prime * result + ((lastAcceptedValue == null) ? 0 : lastAcceptedValue.hashCode());
     }
 }
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorState.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorState.java
@@ -119,12 +119,8 @@ public final class PaxosAcceptorState implements Persistable, Versionable {
         return lastAcceptedValue;
     }
 
-    /**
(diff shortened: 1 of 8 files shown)


jkozlowski:
> Note that (2) above is not foolproof as it is possible that the legacy log wrote some new entries and then happened to write the latest one to exactly mach one that SQLite wrote after the migration while there are some in between that do not match:

How is that possible? Wouldn't it have to advance the promise ids and whatnot. Either way I think this is good enough.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosAcceptorState.java
```diff
@@ -117,4 +118,38 @@ public PaxosProposalId getLastAcceptedId() {
     public PaxosValue getLastAcceptedValue() {
         return lastAcceptedValue;
     }
+
+    /**
+     * Standard equals implementation, note that {@link #version} is not persisted and as such is not used in the
+     * calculation.
+     */
+    @Override
+    public boolean equals(Object obj) {
```
jkozlowski, line 127: Can we figure out a way to introduce a special method for this instead of relying on #equals? There's no harm in that, and relying on equals feels wrong, especially since you're ignoring some fields + if you have equals that forces you to have hashCode, which again we don't need.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/paxos/PaxosStateLogMigrationIntegrationTest.java
```diff
@@ -160,25 +161,17 @@ public void learningValuesBeforeCutoffPersistsToLegacyLog() throws IOException {
...
-        assertValuePresent(LATEST_ROUND_BEFORE_MIGRATING, sqliteLog);
-        assertValueAbsent(newRound, sqliteLog);
-
-        PaxosLearner learner = paxosComponents.learner(CLIENT);
-        assertValueNotLearned(newRound, learner);
-
-        assertThat(paxosComponents.getWriteCounter(PaxosLearner.class).getCount()).isEqualTo(0L);
-        assertThat(paxosComponents.getReadCounter(PaxosLearner.class).getCount()).isEqualTo(0L);
+        assertThatThrownBy(this::createPaxosComponents)
+                .as("Written to file based log at greater sequence after migration alredy ran")
```
jkozlowski, line 173: nit: already.

jkozlowski (commented):
Looks good to me, ideally I'd like us not to use #equals here. Not approving until @Jolyon-S does a quick pass.

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogMigrator.java
```diff
@@ -117,6 +120,39 @@ private void writeBatchRetryingUpToFiveTimes(PaxosStateLog<V> target, List<Paxos
...
+                    "The migration to the destination state log was already performed in the past, but the "
+                            + "entry with the greatest sequence in source log does not match the entry in the "
+                            + "destination log. This indicates the source log has advanced since the migration was "
+                            + "performed which could lead to data corruption if allowed to continue.",
+                    SafeArg.of("source entry", source),
+                    SafeArg.of("destination entry", dest));
+        } catch (IOException e) {
+            Preconditions.checkState(false,
+                    "Unable to verify consistency between source and destination paxos logs because the "
+                            + "source log entry could not be read.");
```
Jolyon-S, line 152: That's just weird - just throw a `SafeIllegalStateException`, surely?

Jolyon-S (commented):
Just did a quick pass, and picked up a nit. Didn't review the logic in depth - if you want me to, I'll need to spend some time working out this general flow.

jkozlowski (approved):

PR #5032, opened 2020-10-14
[TS] Parallelism is nice and should be respected

gmaretic:
**Goals (and why)**:
Fix a bug that caused TS to always run with a single thread only

The scheduled executor now uses a core pool size equal to the install configs. The logic for scaling parallelism has been removed because `ScheduledExecutorService` does not create threads beyond core pool size.

Existing tests, added metrics for `pauseMillis` so we can ensure it works correctly.

Should be pretty straightforward

you know it


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -137,6 +137,10 @@ public void registerEntriesReadInBatch(ShardAndStrategy shardStrategy, long batc
         updateMetricsIfPresent(shardStrategy, metrics -> metrics.registerEntriesReadInBatch(batchSize));
     }
 
+    public void updateSweepDelay(ShardAndStrategy shardStrategy, long delay) {
+        updateMetricsIfPresent(shardStrategy, metrics -> metrics.updateSweepDelay(delay));
+    }
```
Jolyon-S: If I understand correctly, this _just_ updates a metric? Can you rename to `updateSweepDelayMetric` to make sure it is unambiguous?

Also, will this be an expensive metric? Just making sure we're not going back to our old $$$ ways.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -172,6 +176,7 @@ private static String getTagForStrategy(SweeperStrategy strategy) {
         private final Gauge<Long> millisSinceLastSwept;
         private final SweepOutcomeMetrics outcomeMetrics;
         private final SlidingWindowMeanGauge batchSizeMean;
+        private final CurrentValueMetric<Long> sweepDelay;
```
Jolyon-S, line 179: nit: as above, rename to `sweepDelayMetric`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -278,13 +285,17 @@ private void updateProgressForShard(int shard, long sweptTs) {
...
         }
 
-        public void registerEntriesReadInBatch(long batchSize) {
+        private void registerEntriesReadInBatch(long batchSize) {
             batchSizeMean.update(batchSize);
         }
+
+        private void updateSweepDelay(long delay) {
+            sweepDelay.setValue(delay);
+        }
```
Jolyon-S, line 298: nit: overloaded methods should be next to each other

atlasdb-impl-shared/src/main/metrics/targeted-sweep.yml
```diff
@@ -49,3 +49,8 @@ namespaces:
         tags:
           - strategy
         docs: Mean number of entries read in a batch from the targeted sweep queue.
+      sweepDelay:
+        type: gauge
+        tags:
+          - strategy
+        docs: The latest delay between iterations of targeted sweep.
```
Jolyon-S, line 56: Can we get a numThreads metric? Does that still scale?

Jolyon-S (commented):
Just some quick comments, will have to dive into the code to ensure that the correctness has not been violated.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
gmaretic: ah now I understand, will do 👍

atlasdb-impl-shared/src/main/metrics/targeted-sweep.yml
gmaretic: We removed the thread scaling part because we can't do it without implementing some nontrivial scheduler, so it is back to using the install config

Jolyon-S: that's a bit of a pain - presumably the change of sweep delay will mean that if someone has configured more threads than they should, they will mostly be idle (presumably desirable, as that's still less work being done)? Does this mean we should make the default more threads, or still go through a bump-when-we-need workflow?

gmaretic: Yeah, they will be idle, but tbh no one really configures this unnecessarily. Bumping manually is sad, but implementing a scheduler is probably not something we want to get into right now.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
Jolyon-S: you didn't rename this

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
```diff
@@ -33,69 +34,81 @@
...
         sweepTwentyIterationsWithResult(SUCCESS_TOO_FAST);
         assertThat(delay.getNextPause(SUCCESS_TOO_FAST))
                 .isGreaterThanOrEqualTo((long) (DEFAULT_MAX_PAUSE_MILLIS * 0.95));
+        assertThat(metrics).hasValueGreaterThanOrEqualTo((long) (DEFAULT_MAX_PAUSE_MILLIS * 0.95));
     }
 
     @Test
     public void consistentFullBatchesGravitatesTowardsMinimumPause() {
         sweepTwentyIterationsWithResult(SUCCESS_TOO_SLOW);
-        assertThat(delay.getNextPause(SUCCESS_TOO_SLOW)).isGreaterThanOrEqualTo((long) (MIN_PAUSE_MILLIS * 1.05));
```
gmaretic, line 98: This was passing because it was 1 😓

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
```diff
@@ -73,7 +74,8 @@ private CellFilterMetrics() {
             SWEEP_TS,
             LAST_SWEPT_TS,
             LAG_MILLIS,
-            BATCH_SIZE_MEAN);
+            BATCH_SIZE_MEAN,
+            SWEEP_DELAY);
```
jeremyk-91, line 78: 😢

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepDelay.java
```diff
@@ -41,11 +42,13 @@
 
     private final long initialPause;
     private final long maxPauseMillis;
+    private final LongConsumer updateMetrics;
```
jeremyk-91, line 45: nit: this name is a bit meh, I would normally name this like `sweepDelayMetricsUpdater` or something. Not too concerned as it's in the SweepDelay class.

jeremyk-91 (approved):
👍  Looks correct, only minor comments.

Sigh. It's a bit of a shame that we may need to go back to manually bumping people where they need it, so this is a piece of work that we could look at picking up again if there's bandwidth (subject to suitable testing).

PR #5036, opened 2020-10-15
[TL] Persist Schema Version in Sqlite

gmaretic:
**Goals (and why)**:
This allows us to explode in the future if we go down to an unsupported version

Straightforward. Left the schema version at 3 because there is no real benefit to changing it at this time.

Unit tests

Note that we upgrade the schema version  as the last task before the `TimelockAgent` is created. This means that it is done after paxos sqlite migration. In the future it may be necessary to upgrade the schema version only after some asynchronous migration has run. I added a comment about this, is this a reasonable amount of caution?

Small

This week


jkozlowski (approved):

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -124,6 +126,13 @@ public static TimeLockAgent create(MetricsManager metricsManager,
                 metricsManager,
                 Suppliers.compose(TimeLockRuntimeConfiguration::paxos, runtime::get));
 
+        // Upgrading the schema version should generally happen AFTER any migration has finished running. Keep this in
+        // mind for any potential live migrations
+        PersistedSchemaVersion persistedSchemaVersion = PersistedSchemaVersion
+                .create(installationContext.sqliteDataSource());
+        persistedSchemaVersion.upgradeVersion(SCHEMA_VERSION);
+        verifySchemaVersion(persistedSchemaVersion);
```
jkozlowski, line 134: Doesn't this need to be before all the setting up of PaxosResources etc.? Isn't this currently where the migration happens?

PR #5038, opened 2020-10-15
make oracle protocol case insensitive

berler:
**Goals (and why)**:

Makes the protocol (tcp/tcps) case insensitive in oracle configs.
Previously oracle connection configs required the protocol to be
upper-case. Most of our other internal products allow any case
on the oracle protocol, so the requirement here for it to be
upper-case in config results in some errors when people try to
configure oracle.

Adds jackson annotations to the ConnectionProtocol enum.

Added a unit test.

**Priority (whenever / two weeks / yesterday)**: high - allowing this to be case insensitive avoids configuration errors and helps with interoperability with other products that don't enforce case sensitivity on the oracle protocol.

### DIFF
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionProtocol.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionProtocol.java
@@ -15,6 +15,8 @@
  */
 package com.palantir.nexus.db.pool.config;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+
 public enum ConnectionProtocol {
     TCP("tcp"), TCPS("tcps");
 
@@ -28,9 +30,10 @@ public enum ConnectionProtocol {
         return urlStr;
     }
 
+    @JsonCreator
     public static ConnectionProtocol fromUrlString(String val) {
         for (ConnectionProtocol cp : ConnectionProtocol.values()) {
-            if (cp.getUrlString().toLowerCase().equals(val.toLowerCase())) {
+            if (cp.getUrlString().equalsIgnoreCase(val)) {
                 return cp;
             }
         }
--- a/atlasdb-dbkvs-hikari/src/test/java/com/palantir/nexus/db/pool/config/OracleConnectionConfigTest.java
+++ b/atlasdb-dbkvs-hikari/src/test/java/com/palantir/nexus/db/pool/config/OracleConnectionConfigTest.java
@@ -21,11 +21,14 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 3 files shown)


atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionProtocol.java
```diff
@@ -28,9 +30,10 @@ public String getUrlString() {
         return urlStr;
     }
 
+    @JsonCreator
     public static ConnectionProtocol fromUrlString(String val) {
         for (ConnectionProtocol cp : ConnectionProtocol.values()) {
-            if (cp.getUrlString().toLowerCase().equals(val.toLowerCase())) {
+            if (cp.getUrlString().equalsIgnoreCase(val)) {
```
Jolyon-S, line 36: While I agree that `equalsIgnoreCase` is better than what we were doing before, it _should_ be semantically the same - and indeed, reverting this change resulted in the tests passing.

I think the fix here is actually `@JsonCreator` if I am not mistaken - before, it was trying to map directly to one of the enums, whereas now it actually goes through this method - is that right?

atlasdb-dbkvs-hikari/src/test/java/com/palantir/nexus/db/pool/config/OracleConnectionConfigTest.java
```diff
@@ -163,4 +166,23 @@ public void tcpsSetsTrustStore() throws IOException {
                 .port(PORT);
     }
 
+    @Test
+    public void protocolCaseInsensitiveTest() throws IOException {
```
Jolyon-S, line 170: Would you mind writing a test that verifies that this does throw if an invalid protocol is passed?

Jolyon-S (commented):
I think this is pretty much there - would just like one more test before we merge.

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionProtocol.java
berler: yes, this is just a slight optimization, as well as making the code more readable

atlasdb-dbkvs-hikari/src/test/java/com/palantir/nexus/db/pool/config/OracleConnectionConfigTest.java
berler: Done.

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionProtocol.java
berler: it also avoids an extra string allocation for the case that val is not already lowercase

Jolyon-S (approved):
Looks good. Thanks!

PR #5040, opened 2020-10-16
LockWatchManager#isEnabled.

jkozlowski:
**Goals (and why)**:

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
@@ -34,6 +34,10 @@ public abstract class LockWatchManager {
 
     // These methods are hidden on purpose as they should not be generally available, only for brave souls!
 
+    boolean isEnabled() {
+        return true;
+    }
+
     /**
      * Gets the {@link CommitUpdate} taking into account all changes to lock watch state since the start of the
      * transaction, excluding the transaction's own commit locks.
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
@@ -34,9 +34,7 @@ public abstract class LockWatchManager {
 
     // These methods are hidden on purpose as they should not be generally available, only for brave souls!
 
-    boolean isEnabled() {
-        return true;
-    }
+    abstract boolean isEnabled();
 
     /**
      * Gets the {@link CommitUpdate} taking into account all changes to lock watch state since the start of the
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
@@ -34,15 +34,6 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 final class ResilientLockWatchEventCache extends AbstractInvocationHandler {
 
     private static final Logger log = LoggerFactory.getLogger(ResilientLockWatchEventCache.class);
-    private static final Method IS_ENABLED;
-
-    static {
-        try {
-            IS_ENABLED = LockWatchEventCache.class.getMethod("isEnabled");
-        } catch (NoSuchMethodException e) {
-            throw new RuntimeException("Did someone rename this method?", e);
-        }
-    }
 
     static LockWatchEventCache newProxyInstance(
             LockWatchEventCache defaultCache,
@@ -72,11 +63,7 @@ final class ResilientLockWatchEventCache extends AbstractInvocationHandler {
     protected synchronized Object handleInvocation(Object proxy, Method method, Object[] args)
             throws IllegalAccessException {
         try {
-            if (method.equals(IS_ENABLED)) {
-                return delegate != fallbackCache;
-            } else {
-                return method.invoke(delegate, args);
-            }
+            return method.invoke(delegate, args);
         } catch (InvocationTargetException e) {
             throw handleException(e);
         }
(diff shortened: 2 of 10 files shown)


changelog/@unreleased/pr-5040.v2.yml
```diff
@@ -0,0 +1,24 @@
+type: improvement
+improvement:
+  description: |-
```
Jolyon-S, line 3: Fix this changelog, please!

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
```diff
@@ -21,6 +21,11 @@
 import java.util.Set;
 
 public interface LockWatchEventCache {
+
+    default boolean isEnabled() {
+        return true;
+    }
```
Jolyon-S, line 27: This feels wrong - this should call down to `LockWatchEventCacheImpl` or `NoOpLockWatchEventCache`.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCacheTest.java
```diff
@@ -54,6 +56,21 @@ public void before() {
...
     }
 
+    @Test
+    public void testCanDelegateIsEnabled() {
+        assertThat(proxyCache.isEnabled()).isTrue();
+
+        RuntimeException runtimeException = new RuntimeException();
+        when(defaultCache.getCommitUpdate(anyLong())).thenThrow(runtimeException);
+        assertThatThrownBy(() -> proxyCache.getCommitUpdate(0L)).hasCause(runtimeException)
+                .isExactlyInstanceOf(TransactionLockWatchFailedException.class);
```
Jolyon-S, line 66: do you mind expanding this test with `hasMessage` as there are a number of exceptions under that class (I know I am guilty of this too, but in an effort to make it clearer what we are testing, this should be done).

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
```diff
@@ -34,6 +34,10 @@
 
     // These methods are hidden on purpose as they should not be generally available, only for brave souls!
 
+    boolean isEnabled() {
+        return true;
+    }
```
Jolyon-S, line 39: Why is this necessary? Also, if we're going to default to anything, wouldn't it be better to be false? I still think this should just be:
`boolean isEnabled;`

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
Jolyon-S: Oh I see - you went with an approach entirely through the proxy. I'm not sure I like that really - surely the logic is 100x cleaner just by doing what you originally said, i.e. having `LWECI` return true and `NOLWEC` return false?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
```diff
@@ -34,6 +34,15 @@
...
     private static final Logger log = LoggerFactory.getLogger(ResilientLockWatchEventCache.class);
+    private static final Method IS_ENABLED;
+
+    static {
+        try {
+            IS_ENABLED = LockWatchEventCache.class.getMethod("isEnabled");
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException("Did someone rename this method?", e);
+        }
+    }
```
Jolyon-S, line 45: I don't really like this - I much prefer just implementing in LWECI and NOLWEC as I feel it is less error prone, and doing this kind of proxy-wrangling is abusing the whole point of it IMO.

Jolyon-S (changes requested):
I'm not really sure I like the approach of abusing the proxy like this - can we not just go with the first thing you suggested, which is just have the classes implement `isEnabled` to true or false?

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
jkozlowski: Discussed offline

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
jkozlowski: I don’t think it matters, I just went with this approach.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCache.java
jkozlowski: Discussed offline

changelog/@unreleased/pr-5040.v2.yml
jkozlowski: Done

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCacheTest.java
jkozlowski: Nope, ResilientLockWatchEventCache has just one from what I can see?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
Jolyon-S: I still think we should not set this - I'd rather never, ever have to debug another bug like that auto-delegate, and while I know this is very different, it's just less error-prone to not set a default.

Jolyon-S: (I know you have a test for it, but still).

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchEventCacheTest.java
Jolyon-S: Ah yea, I see you're just throwing a generic exception instead of actually causing a legit exception inside LWECI.

Jolyon-S (approved):
Just one thing, then I think good to merge.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManager.java
jkozlowski: Done

PR #5043, opened 2020-10-16
[Draft] Just experimenting with merging the baseline PR

Jolyon-S:
**Goals (and why)**:

**Implementation Description (bullets)**:

**Testing (What was existing testing like?  What have you done to improve it?)**:

**Concerns (what feedback would you like?)**:

**Where should we start reviewing?**:

**Priority (whenever / two weeks / yesterday)**:

PR #5044, opened 2020-10-18
Upgrade gradle-baseline

schlosna:
**Goals (and why)**:
AtlasDB's gradle-baseline version is significantly out of date and is not being kept up-to-date by excavators. Similar to #5043 , this upgrades the gradle-baseline version used so that atlasdb is enforcing (most of) the checks baseline provides

Steps below can be seen in individual commits on this PR

- Apply gradle-palantir-java-format plugin
- Bump gradle-baseline plugin version
- Update baseline configs
- Bump error-prone and immutables versions for compatibility with error-prone
- Apply automated gradle-baseline refactorings (`./gradlew compileJava compileTestJava -PrefasterApply -PerrorProneApply`)
- Reformat (`./gradlew format`)
- Temporarily relax some error-prone checks from error to warning (to be fixed iteratively in follow up PRs after merge)
- Workaround error-prone fatal error with lambda method reference check
- Unify checkstyle configuration to gradle-baseline version
- Standardized IntelliJ Idea configuration to gradle-baseline version
- Suppress some checkstyle rules until they can be remediated 
- Fix several missing Apache license headers
- Upgrade to gradle 6.7
- Standardize baseline configuration in build.gradle

Local compilation

The reformat generated a ton of diffs, we may want to split that out separately, but I wanted to use the automated refactorings which also generate a bunch of diffs.

`build.gradle` has most of the changes and temporary exclusions.

Whenever


.baseline/checkstyle/checkstyle-suppressions.xml
```diff
@@ -21,17 +25,5 @@
...
-
-    <!-- Certain AtlasDB projects are ignored. -->
-    <suppress files="atlasdb-client/*" checks="." />
-    <suppress files="atlasdb-commons/*" checks="." />
-    <suppress files="atlasdb-console/*" checks="." />
-    <suppress files="atlasdb-jdbc/*" checks="." />
-    <suppress files="atlasdb-jdbc-tests/*" checks="." />
-    <suppress files="cassandra-partitioner/*" checks="." />
-    <suppress files="commons-db/*" checks="." />
-    <suppress files="lock-impl/*" checks="." />
```
schlosna, line 36: moved to custom-suppressions.xml per gradle-baseline standard

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/TableSplittingKeyValueService.java
```diff
@@ -377,8 +378,7 @@ public void compactInternally(TableReference tableRef, boolean inMaintenanceMode
     public ClusterAvailabilityStatus getClusterAvailabilityStatus() {
         return delegates.stream()
                 .map(kvs -> kvs.getClusterAvailabilityStatus())
-                .sorted()
-                .findFirst()
+                .min(Comparator.naturalOrder())
```
carterkozak, line 381: nice

.baseline/checkstyle/custom-suppressions.xml
```diff
@@ -0,0 +1,40 @@
...
+    <!-- Skip some explicit checks until they can be remediated -->
+    <suppress files=".*" checks="AnnotationUseStyle"/>
+    <suppress files=".*" checks="AvoidStaticImport"/>
+    <suppress files=".*" checks="CyclomaticComplexity"/>
+    <suppress files=".*" checks="DesignForExtension"/>
+    <suppress files=".*" checks="RegexpSinglelineJava"/>
+    <suppress files=".*" checks="RedundantModifier"/>
+    <suppress files=".*" checks="SummaryJavadocCheck"/>
+    <suppress files=".*" checks="VisibilityModifier"/>
+    <suppress files=".*" id="BanGuavaCaches"/>
```
schlosna, line 29: I opted to skip these checks for now so that we have the vast majority of improved baseline checks, and we can burn these down in smaller, independent PRs that are easier to review

build.gradle
```diff
@@ -12,18 +12,19 @@ buildscript {
...
-        classpath 'gradle.plugin.com.palantir.gradle.gitversion:gradle-git-version:0.11.0'
+        classpath 'com.palantir.javaformat:gradle-palantir-java-format:1.0.1'
+        classpath 'com.palantir.metricschema:gradle-metric-schema:0.5.5'
         classpath 'com.palantir.sls-packaging:gradle-sls-packaging:4.10.1'
-        classpath 'com.netflix.nebula:nebula-publishing-plugin:17.2.1'
         classpath 'gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:0.15.0'
+        classpath 'gradle.plugin.com.palantir.gradle.docker:gradle-docker:0.22.1'
+        classpath 'gradle.plugin.com.palantir.gradle.gitversion:gradle-git-version:0.11.0'
         classpath 'gradle.plugin.org.inferred:gradle-processors:3.3.0'
+        classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:4.9.8'
```
schlosna, line 27: sorted these to simplify future updates.

changelog/@unreleased/pr-5044.v2.yml
```diff
@@ -0,0 +1,34 @@
...
+    **Concerns (what feedback would you like?)**:
+    The reformat generated a ton of diffs, we may want to split that out separately, but I wanted to use the automated refactorings which also generate a bunch of diffs.
+
+    **Where should we start reviewing?**:
+    `build.gradle` has most of the changes and temporary exclusions.
+
+    **Priority (whenever / two weeks / yesterday)**:
+    Whenever
+  links:
+  - https://github.com/palantir/atlasdb/pull/5044
```
carterkozak, line 34: I don't think we need a changelog for these unless there are specific fixes we want to call out.

If we keep the changelog we should remove the template components and write a a short description of the pieces consumers may be interested in.

build.gradle
```diff
@@ -35,8 +36,7 @@ plugins {
     id 'com.google.protobuf' version '0.8.10'
 }
 
-apply plugin: 'com.palantir.baseline-config'
-apply plugin: 'com.palantir.baseline-circleci'
```
schlosna, line 39: applied by default now https://github.com/palantir/gradle-baseline/blob/32b092e918cb3f46380273fd7957a26b09ee0d6a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java#L36

build.gradle
```diff
@@ -35,8 +36,7 @@ plugins {
     id 'com.google.protobuf' version '0.8.10'
 }
 
-apply plugin: 'com.palantir.baseline-config'
```
schlosna, line 38: applied by default now https://github.com/palantir/gradle-baseline/blob/32b092e918cb3f46380273fd7957a26b09ee0d6a/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/Baseline.java#L35

build.gradle
```diff
@@ -65,6 +65,7 @@ allprojects {
 
     apply plugin: 'java-library'
     apply plugin: 'org.inferred.processors'  // installs the "processor" configuration needed for baseline-error-prone
+    apply plugin: 'com.palantir.java-format'
```
schlosna, line 68: use standard formatting to streamline the baseline refactorings

build.gradle
```diff
@@ -104,6 +105,44 @@ allprojects {
...
+                'ReturnValueIgnored',
+                'SafeLoggingExceptionMessageFormat',
+                'Slf4jLevelCheck',
+                'StreamOfEmpty',
+                'StrictUnusedVariable',
+                'ValidateConstantMessage',
+            ]
+            for (String checkName : errorsToRelax) {
+                options.errorprone.warn(checkName)
+            }
```
schlosna, line 142: relaxing from error to warning for now, will burn down in separate PRs

gradle/baseline.gradle
```diff
@@ -1,42 +0,0 @@
-apply plugin: 'com.palantir.baseline-checkstyle'
-
-// ./gradlew compileJava compileTestJava -PenableErrorProne=true
-if (enableErrorProne.toBoolean()) {
```
schlosna, line 4: moved all of this to `build.gradle` and this is no longer opt-in, its default on everywhere

gradle/idea.gradle
```diff
@@ -1,215 +0,0 @@
...
-assert targetCompatibility instanceof JavaVersion;
-assert sourceCompatibility == targetCompatibility;
-
-project.tasks.withType(JavaCompile) { JavaCompile task ->
-  // This actually runs javax.tools.JavaCompiler in-process, but we can still pass arguments
-  // as though it were running the javac executable.
-  task.options.compilerArgs << "-XDignore.symbol.file";
-  task.options.compilerArgs << "-Xlint:none";
-  task.options.compilerArgs << '-encoding' << 'utf-8';
-  task.options.encoding = 'UTF-8'
```
schlosna, line 14: defaulting to gradle-baseline's IntelliJ Idea integration configs. Checkstyle now works properly and compiles fine for me in IDEA 2020.3 CE EAP.

gradle/shared.gradle
```diff
@@ -27,7 +26,6 @@ dependencies {
     // Dont add Java8 compile/runtime dependencies here as commons-executors needs to be compatible with Java6 clients
     compile group: 'com.google.code.findbugs', name: 'jsr305'
     compile group: 'com.google.code.findbugs', name: 'findbugs-annotations'
-    checkstyle group: 'com.puppycrawl.tools', name: 'checkstyle', version: libVersions.checkstyle
     testCompile group: 'junit', name: 'junit'
     testCompile group: 'org.assertj', name: 'assertj-core'
```
schlosna, line 30: We can probably clean up more of these in a follow up PR

gradle/versions.gradle
```diff
@@ -28,7 +28,6 @@ ext.libVersions =
     protoc: '3.5.1',
     findbugs: '1.3.9',
     hikariCP: '3.4.1',
-    checkstyle: '6.18',
```
schlosna, line 31: AtlasDB has had JDK8 as minimum dependency for a while, and this old version of checkstyle isn't compatible with gradle-baseline's checkstyle rules, and this old version would breaks Idea Checkstyle integration, but now this all works properly.

versions.props
```diff
@@ -45,7 +45,7 @@ org.clojure:clojure = 1.8.0
 org.codehaus.groovy:* = 2.4.4
 org.derive4j:* = 1.1.1
 org.freemarker:freemarker = 2.3.29
-org.immutables:* = 2.8.2
+org.immutables:* = 2.8.8
```
schlosna, line 48: bumping this was necessary to fix gradle-baseline & error-prone compatibility

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/impl/TableSplittingKeyValueService.java
schlosna: This was an automated gradle-baseline refactor

carterkozak:
I think the CI failures are related to the scale of error-prone compiler warnings, they seem to go away when there are fewer failures.

schlosna:
@carterkozak should we go the route of https://github.com/palantir/atlasdb/pull/5048 to disable those specific checks if they're causing overhead and CI flakes?

carterkozak:
I've updated that PR to set most of them to WARN, disable a couple, and fix the noisy ones with safe automatic fixes. It seems to be passing.

carterkozak:
It targets this PR, so we can pull in any subset of the changes we want

gradle/baseline.gradle
jkozlowski: So reading through it looks like this was doing some per main/test package differences, but I think it's fine. We should quickly drive all the turned off warnings.

jkozlowski:
Ok, so I went through the chain of commits, and individually they make sense. I scrolled through the large automated format, and read through errorprone auto-refactorings.

I think this makes sense and needs to be merged quickly, before it gets stale (it already started accumulating merge conflicts).

@schlosna are we now running with the equivalent of -WError for all of compiler|checkstyle|errorprone?

jkozlowski (approved):

.baseline/idea/intellij-java-palantir-style.xml
```diff
@@ -100,30 +87,44 @@
...
           <option name="KEEP_SIMPLE_CLASSES_IN_ONE_LINE" value="true" />
+          <option name="KEEP_SIMPLE_LAMBDAS_IN_ONE_LINE" value="true" />
           <option name="KEEP_SIMPLE_METHODS_IN_ONE_LINE" value="true" />
-          <option name="METHOD_PARAMETERS_WRAP" value="1" />
+          <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+          <option name="LINE_COMMENT_ADD_SPACE" value="true" />
+          <option name="METHOD_CALL_CHAIN_WRAP" value="5" />
+          <option name="METHOD_PARAMETERS_LPAREN_ON_NEXT_LINE" value="true" />
+          <option name="METHOD_PARAMETERS_WRAP" value="5" />
+          <option name="OPTIMIZE_IMPORTS_ON_THE_FLY" value="true" />
```
jeremyk-91, line 117: 💯

atlasdb-console/src/main/groovy/com/palantir/atlasdb/console/AtlasConsoleBinder.java
```diff
@@ -99,7 +99,8 @@ public Object call() {
             System.out.println("\nTopics with help messages:\n"); // (authorized)
             System.out.println(help.keySet().toString()); // (authorized)
             System.out.println("\nCall help(\'topic\') to view information on a specific topic.\n"); // (authorized)
-            System.out.println("If you'd like to enable database mutations (not recommended), re-run console with the 'mutations_enabled' flag\n"); // (authorized)
+            System.out.println("If you'd like to enable database mutations (not recommended), re-run console with the"
+                    + " 'mutations_enabled' flag\n"); // (authorized)
```
jeremyk-91, line 103: Is this OK? Lines 114-115 suggests that `(authorized)` should be indicated on both lines

jeremyk-91 (commented):
Thanks a lot for doing this! I think there's one thing to validate that I'm not sure is correct, since I remember this was one of the initial sticking points we had when refaster was first going out. @carterkozak I'm tagging you as I think you might know as well, if this is safe?

I put a "do not merge" just so this doesn't accidentally merge without us making a decision on this: once we get confirmation we should be fine to proceed.

-Werror question: given description I'd say probably not (since things were moved to warn to avoid erroring)?

atlasdb-console/src/main/groovy/com/palantir/atlasdb/console/AtlasConsoleBinder.java
jeremyk-91: Had a look internally - this is probably OK (see e.g. the large internal product's class that inspects atlas's lock service)

jeremyk-91 (approved):
validated my earlier concern. Thanks a lot for doing this!

jkozlowski:
ok so @schlosna is it possible to move the ones we have disabled to just be suppressed and then -Werror everything? that's certainly how we run atlas-proxy internally

schlosna:
@jkozlowski right now this PR just upgrades atlasdb to build against a modern version of gradle-baseline so that we can keep baseline up to date with excavators. We will need separate PRs to fix up the remaining checks we had to reduce from error to warning to treat those as errors, and then we'll also need to promote warnings to error (equivalent to `-Wall -Werror`). As you've seen, this PR is already too large and quickly accrues conflicts, so would prefer we land this first then work those down so we're at least progressing in the right direction.

schlosna:
After this goes green and merges, I think we can bump https://errorprone.info/bugpattern/BoxedPrimitiveEquality and https://errorprone.info/bugpattern/ReferenceEquality to error and fix the findings.

PR #5046, opened 2020-10-19
reduce log level.

anjor:
**Goals (and why)**:
Internal PDS-138463

This can happen during a timelock leader election, and retrying transaction should work.
The log level can be misleading, hence reducing the level to warn.

fairly trivial

none, but again, fairly trivial

none


jeremyk-91 (approved):
Looks good! I think we'll need to get the changelog check to pass as well (either add one, or decide not needed - I'd choose to add one in this case) and we should be good to merge/autorelease

PR #5048, opened 2020-10-19
Update error-prone configuration

carterkozak:
Disable relaxed errors rather than switching to WARN


build.gradle
```diff
@@ -106,17 +106,9 @@ allprojects {
...
-                '-XepDisableWarningsInGeneratedCode',
-                '-XepExcludedPaths:.*(generated).*',
-            ])
-
-            // temporarily relax constraints until we can fix all violations
-            Set<String> errorsToRelax = [
-                'CatchBlockLogException',
+    tasks.withType(JavaCompile) {
+        // temporarily relax constraints until we can fix all violations
+        options.errorprone.disable 'CatchBlockLogException',
```
schlosna, line 111: Do we actually want to disable? I tend to grep the `(error|warning):` from build log to fix these up

carterkozak: We can set them to warn instead, no strong preference. I've normally added things to a list in order to burn through it later since it's easier to promote a single check and fix all occurrences.

Mind if I push a few automated check fixes to cut down on the warnings along with this?

build.gradle
```diff
@@ -124,8 +124,8 @@ allprojects {
                 'ReturnValueIgnored',
                 'SafeLoggingExceptionMessageFormat',
                 'StreamOfEmpty',
-                'StrictUnusedVariable',
                 'ValidateConstantMessage'
+        options.errorprone.disable 'StrictUnusedVariable'
```
carterkozak, line 128: @schlosna I've disabled `StrictUnusedVariable` since it's responsible for the vast majority of compiler warning spam and we can attack it separately.

schlosna (approved):
👍

carterkozak:
Feel free to merge this into your branch when you're happy with it

PR #5057, opened 2020-10-20
[TimeLock corruption detection] | Timestamp corruption checks

sudiksha27:
**Goals (and why)**:
Adds Timestamp corruption indicating checks. The checks analyze history from local and all remotes servers and determine if there are signs of corruption.

 Implements following (per round) corruption checks 
- Learned values for any round are same across all nodes

- Learned value is accepted by quorum of nodes

- In a round, greatest accepted value <= learned value

Also, refactors utils for tests.

Added tests

- Do the checks seem reasonable? 

- Are there more checks that can be added?

- Did I miss something that would cause false alerts?

HistoryAnalyzer
- The wiring is not there, so please only review the logical checks

ASAP 🏃‍♀️


timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -0,0 +1,143 @@
...
+import com.google.common.collect.Iterables;
+import com.palantir.atlasdb.encoding.PtBytes;
+import com.palantir.paxos.PaxosValue;
+import com.palantir.timelock.history.PaxosAcceptorData;
+import com.palantir.timelock.history.models.CompletePaxosHistoryForNamespaceAndUseCase;
+import com.palantir.timelock.history.models.ConsolidatedLearnerAndAcceptorRecord;
+
+public final class HistoryAnalyzer {
+
+    public static boolean runCorruptionCheckOnHistory(CompletePaxosHistoryForNamespaceAndUseCase history) {
```
jeremyk-91, line 36: Let's not return `boolean`: would be good to have some kind of wrapper type

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -0,0 +1,143 @@
...
+                .stream()
+                .allMatch(seq -> {
+                    Optional<PaxosValue> optionalLearnedValue = getLearnedValue(records, seq);
+                    if(!optionalLearnedValue.isPresent()) {
+                        return true;
+                    }
+
+                    PaxosValue learnedValue = optionalLearnedValue.get();
+                    List<PaxosValue> acceptedValues = getAcceptedValues(records, seq, learnedValue);
+                    return acceptedValues.size() >= quorum;
```
jeremyk-91, line 68: I'd extract a method for this. Depending on taste, you could also process the stream more aggressively and/or use keyed-streams

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -0,0 +1,143 @@
...
+            List<ConsolidatedLearnerAndAcceptorRecord> records, Long seq) {
+        byte[] learnedValueData = getPaxosValueData(getLearnedValue(records, seq));
+        if (learnedValueData == null) {
+            return true;
+        }
+
+        byte[] greatestAcceptedValueData = getPaxosValueData(getGreatestAcceptedValueAtSequence(records, seq));
+        if (greatestAcceptedValueData == null) {
+            // should not reach here
+            return false;
```
jeremyk-91, line 90: if this is a bug in atlas code we should probably throw a SafeIllegalStateException

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -0,0 +1,143 @@
...
+                        .map(PaxosAcceptorData::getLastAcceptedValue)
+                        .orElseGet(Optional::empty))
+                .filter(paxosValue -> getPaxosValueData(paxosValue) != null)
+                .map(Optional::get)
+                .max(Comparator.comparingLong(paxosValue ->  PtBytes.toLong(paxosValue.getData())));
+    }
+
+
+    private static byte[] getPaxosValueData(Optional<PaxosValue> learnedValue) {
+        return learnedValue.map(PaxosValue::getData).orElse(null);
```
jeremyk-91, line 141: Let's have a comment here that this should only be used for timestamp bound checks: this would behave strangely if we did it on leader election based checks, because there is a semantic difference between `null` and no agreement because paxos phase two didn't happen

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/HistoryAnalyzerTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final String USE_CASE_LEARNER = LearnerUseCase.createLearnerUseCase(USE_CASE).value();
+    private static final String USE_CASE_ACCEPTOR = AcceptorUseCase.createAcceptorUseCase(USE_CASE).value();
+
+    private StateLogComponents localStateLogComponents;
+    private List<StateLogComponents> remoteStateLogComponents;
+    PaxosLogHistoryProvider paxosLogHistoryProvider;
+
+    @Before
+    public void setup() throws IOException {
+        localStateLogComponents = createShitForServer("randomFile1");
```
jeremyk-91, line 69: 😅

jeremyk-91 (commented):
Broadly looks good. Probably small bits of refactoring to do (mainly around wrapper object + separate method for the accepted-by-quorum check), but the logic of the checks is correct and the tests look broadly good.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -16,65 +16,74 @@
...
-        return verifyLearnersHaveLearnedSameValues(history)
-        && verifyLearnedValueWasAcceptedByQuorum(history)
-        && verifyLearnedValueIsGreatestAcceptedValue(history);
+    public static List<CorruptionStatus> corruptionStateForNamespaceAndUseCase(
+            CompletePaxosHistoryForNamespaceAndUseCase history) {
+        return Stream.of(learnersHaveLearnedSameValues(history),
+                learnedValueWasAcceptedByQuorum(history),
+                learnedValueIsGreatestAcceptedValue(history))
+                .filter(status -> status != CorruptionStatus.HEALTHY)
+                .collect(Collectors.toList());
```
jeremyk-91, line 43: This is empty if everyone is healthy. Is that by design? (If so that's fine, just make sure it is!)

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionStatus.java
```diff
@@ -17,16 +17,26 @@
...
 
 public enum CorruptionStatus {
-    HEALTHY(false),
-    CORRUPTION(true);
+    HEALTHY(false, false),
+    DIVERGED_LEARNERS(true, false), // this is false for now
+    VALUE_LEARNED_WITHOUT_QUORUM(true, false),
+    ACCEPTED_VALUE_GREATER_THAN_LEARNED(true, false),
+    DEFINITIVE_CORRUPTION_DETECTED_BY_LOCAL(true, true),
+    DEFINITIVE_CORRUPTION_DETECTED_BY_REMOTE(true, true);
```
jeremyk-91, line 25: The first 4 are one layer and the bottom 2 are another. I'd suggest having maybe a `CorruptionCheckStatus` that is one of `HEALTHY, DIVERGED_LEARNERS ...` and then a `CorruptionStatus` (this class) just has `HEALTHY` and the two `DEFINITIVE_CORRUPTION` ones

jeremyk-91 (approved):
Should be fine, though make sure you agree with the comment on an empty list being returned if all checks are healthy

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
sudiksha27: Yes, as of now, the plan is to only react to health check violations. So, the list for healthy namespaceAndUse will be empty.

PR #5060, opened 2020-10-20
Match dialogue read/write timeout

leonz:
**Goals (and why)**:

Read and write timeouts for Dialogue clients must be the same.

Currently logs:
```
Read and write timeouts do not match, The value of the readTimeout {} will be used and write timeout {} will be ignored.
readTimeout
"PT1M5S"
writeTimeout
"PT5M"
```


jeremyk-91 (approved):
Thanks. Sorry for the inconvenience!

PR #5062, opened 2020-10-21
Leader Election Metrics 1a: Client Side Wiring

Jolyon-S:
**Goals (and why)**:
* Add metrics to the client to monitor the perceived impact of leader elections. Note that further metrics on the actual length of leader elections will be introduced in https://github.com/palantir/atlasdb/pull/5069.

* Extracted an interface for `NamespacedConjureTimelockService`, and created another class that delegates to the implementation.
* Monitor startTransactions and getCommitTimestamp requests, and report metrics when these perceive a leader change.

* Added a unit test; intend on adding ETE tests later (in a followup PR).

* Confirm that this won't affect performance

* `LeaderElectionReportingTimelockService`

This week

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/NamespacedConjureTimelockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/NamespacedConjureTimelockService.java
@@ -24,54 +24,27 @@ import com.palantir.atlasdb.timelock.api.ConjureRefreshLocksRequest;
 import com.palantir.atlasdb.timelock.api.ConjureRefreshLocksResponse;
 import com.palantir.atlasdb.timelock.api.ConjureStartTransactionsRequest;
 import com.palantir.atlasdb.timelock.api.ConjureStartTransactionsResponse;
-import com.palantir.atlasdb.timelock.api.ConjureTimelockService;
 import com.palantir.atlasdb.timelock.api.ConjureUnlockRequest;
 import com.palantir.atlasdb.timelock.api.ConjureUnlockResponse;
 import com.palantir.atlasdb.timelock.api.ConjureWaitForLocksResponse;
 import com.palantir.atlasdb.timelock.api.GetCommitTimestampsRequest;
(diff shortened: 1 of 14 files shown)


lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -59,17 +57,6 @@ public static RemoteTimelockServiceAdapter create(
         return new RemoteTimelockServiceAdapter(rpcClient, conjureClient, lockWatchEventCache);
     }
 
-    public static RemoteTimelockServiceAdapter create(
```
Jolyon-S, line 62: drive-by

gmaretic (approved):
We paired on this so I'm good with it, but going to leave here so others have a chance to look through if they like.

This should be completely not contentious except if someone has a better idea how to wire it into `TimeLockFeedbackBackgroundTask`

lock-api/src/main/java/com/palantir/lock/client/NamespacedConjureTimelockService.java
```diff
@@ -24,54 +24,27 @@
...
-import com.palantir.tokens.auth.AuthHeader;
 
-public class NamespacedConjureTimelockService {
-    private static final AuthHeader AUTH_HEADER = AuthHeader.valueOf("Bearer omitted");
-    private final String namespace;
-    private final ConjureTimelockService conjureTimelockService;
+public interface NamespacedConjureTimelockService {
+    ConjureUnlockResponse unlock(ConjureUnlockRequest request);
 
-    public NamespacedConjureTimelockService(ConjureTimelockService conjureTimelockService, String namespace) {
```
jeremyk-91, line 41: This will have a number of downstream breaks in the internal backup product and internal timelock migration CLI tools. Overall impact is limited, but we need to ensure these get fixed.

jeremyk-91 (approved):
Broadly looks good. Please make sure to track fixing the services that will be broken by this change

PR #5063, opened 2020-10-21
[RC only] | Suppress 5026

sudiksha27:
**Goals (and why)**:
This is temporary - Suppress invalid log state from throwing, instead log error.

- Do not throw but log error if migration logs point to a state of corruption

- Also, log namespace and useCase

N/A

Is logging namespace and useCase a prroblem?

PaxosStateLogMigrator.java

🔥

### DIFF
--- a/changelog/0.254.4-rc1/pr-5063.v2.yml
+++ b/changelog/0.254.4-rc1/pr-5063.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: |2
+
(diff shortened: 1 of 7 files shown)


gmaretic (approved):
Looks good, but probably need to rebase on develop because checks are going crazy, and you'll have failing tests for the throwing

sudiksha27:
Fix in https://github.com/palantir/atlasdb/pull/5074

PR #5064, opened 2020-10-21
[LE] Implement SlidingWindowWeightedMeanGauge

gmaretic:
**Goals (and why)**:
We need a gauge that will accumulate the reported leader election metrics into a sensible approximation of what is going on, which is probably the weighted mean for each of the non-count metrics (we get p99, p95, mean, and count).

Using caffeine cache as a sliding window.

Unit tests

We explicitly do not protect against overflows in calculations since that will not happen for our use case. Should we implement it anyway, and just return null/throw if encountered?

Small

Today

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGauge.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGauge.java
@@ -16,18 +16,16 @@
 
 package com.palantir.atlasdb.util;
 
-import java.time.Duration;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicLong;
-
-import org.immutables.value.Value;
-
 import com.codahale.metrics.Gauge;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.collect.ImmutableList;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import org.immutables.value.Value;
 
 /**
  * A gauge that calculates the weighted mean of updates during a sliding window of time. Correctness is only guaranteed
@@ -39,9 +37,8 @@ public class SlidingWindowWeightedMeanGauge implements Gauge<Double> {
     private final AtomicLong counter = new AtomicLong();
 
     public SlidingWindowWeightedMeanGauge(Duration expirationDuration) {
-        this.updates = Caffeine.newBuilder()
-                .expireAfterWrite(expirationDuration)
-                .build();
+        this.updates =
+                Caffeine.newBuilder().expireAfterWrite(expirationDuration).build();
     }
 
     public static SlidingWindowWeightedMeanGauge create() {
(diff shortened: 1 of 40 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGauge.java
```diff
@@ -0,0 +1,83 @@
...
+import org.immutables.value.Value;
+
+/**
+ * A gauge that calculates the weighted mean of updates during a sliding window of time. Correctness is only guaranteed
+ * as long as the sum of recorded weights and sum of weighted values do not exceed {@link Long#MAX_VALUE} and
+ * {@link Double#MAX_VALUE}, respectively.
+ */
+public class SlidingWindowWeightedMeanGauge implements Gauge<Double> {
+    private final Cache<Long, WeightedEntry> updates;
+    private final AtomicLong counter = new AtomicLong();
```
Jolyon-S, line 37: Rename to something just so that we know the purpose here is literally just as arbitrary keys for the cache.

atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGauge.java
```diff
@@ -0,0 +1,83 @@
...
+        Preconditions.checkArgument(weight >= 0, "Weight cannot be negative.", SafeArg.of("weight", weight));
+        if (weight == 0) {
+            return;
+        }
+        updates.put(counter.getAndIncrement(), ImmutableWeightedEntry.of(value, weight));
+    }
+
+    private double summarize(List<WeightedEntry> snapshot) {
+        long totalWeight =
+                snapshot.stream().map(WeightedEntry::weight).mapToLong(x -> x).sum();
```
Jolyon-S, line 64: Use just `mapToLong`

atlasdb-client/src/main/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGauge.java
```diff
@@ -0,0 +1,83 @@
...
+        if (totalWeight == 0) {
+            return 0.0;
+        }
+        double valueSum = snapshot.stream()
+                .map(entry -> entry.value() * entry.weight())
+                .mapToDouble(x -> x)
```
Jolyon-S, line 70: same as above

atlasdb-client/src/test/java/com/palantir/atlasdb/util/SlidingWindowWeightedMeanGaugeTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertThatThrownBy(() -> gauge.update(2345.0, -1)).isInstanceOf(IllegalArgumentException.class);
+        assertValueWithinOnePercentOf(5.0);
+    }
+
+    @Test
+    public void entriesExpire() {
+        SlidingWindowWeightedMeanGauge expiringGauge = new SlidingWindowWeightedMeanGauge(Duration.ZERO);
+        expiringGauge.update(100.0, 50);
+        expiringGauge.update(50.0, 4);
+        assertThat(expiringGauge.getValue()).isCloseTo(0.0, withinPercentage(1));
```
Jolyon-S, line 79: can we reduce to <1%? Just to make this a bit tighter

Jolyon-S: Also, use other method

Jolyon-S (commented):
Mostly fine, couple of nits

Jolyon-S (approved):
Paired offline, and looks good to me. :+1:

Jolyon-S (approved):
:+1:

edit: this didn't cause it to go through, ugh

gmaretic:
:+1:

PR #5068, opened 2020-10-21
[DB TimeLock] 2D - Documentation of DB-TimeLock Considerations

jeremyk-91:
**Goals (and why)**:
- Don't confuse readers of the atlasdb docs

- Replace references to migrations only existing for Cassandra to add DbKVS
- Place warnings around sections of the TimeLock docs that only apply to TimeLocks that run with Paxos persistence.

- Docs change only

- Did I miss any references?

**Where should we start reviewing?**: migration.rst?

**Priority (whenever / two weeks / yesterday)**: I'd say this blocks announcement, so Thursday would be good.

### DIFF
--- a/docs/source/services/timelock_service/manual-migration.rst
+++ b/docs/source/services/timelock_service/manual-migration.rst
@@ -5,8 +5,11 @@ Manual Migration to Timelock Server
 
 .. warning::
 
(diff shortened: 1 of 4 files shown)


sudiksha27 (approved):
LGTM!

PR #5069, opened 2020-10-21
Leader Election Metrics 1b: Estimate Leader Election Duration

gmaretic:
**Goals (and why)**:
In addition to existing metrics, calculate the estimated leader duration

Javadoc explains the approach in detail. Other things to note:

- The ordering of updates for the maps and creating the snapshots is selected so that the keys of upperbounds are a superset of lowerbounds
- We do some simple trivial cleanup from the maps, but generally there shouldn't be so many leader elections to warrant too much effort here
- Still need to wire this through, but that is not relevant for the review

Some initial unit tests are there

Probably needs some polishing, but overall I think the approach is reasonable. Would appreciate a detailed review and suggestions for additional tests.

Read the Javadoc

Soon


lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -122,6 +140,129 @@ public ConjureStartTransactionsResponse startTransactions(ConjureStartTransactio
...
+    /**
+     * Estimating the duration of leader election:
+     *
+     * If a request is sent out at time T_1, and a response arrives at time T_2 with the id of leader A, we can
+     * deduce three things:
+     *   1. A could not have lost leadership before T_1 if it was the leader before it (upper bound)
+     *   2. A must have gained leadership before T_2 (lower bound)
+     *   3. A was the leader at least in some point in the interval [T_1, T_2]
+     * Although from just one request, the lower bound is greater than the upper bound, after multiple responses that
+     * is generally not going to be the case anymore.
```
Jolyon-S, line 156: IMO this is pretty hard to take in. My understanding of it is as follows:
0. Lower and upper bound are referring to the temporal bounds of when A is the leader (I don't think that's really mentioned explicitly, and it's not obvious just from the text here).
1. Lower bound is the first successful response from leader A;
2. Upper bound is the last successful request to leader A.

While what you've said is true for a single request/response pair, it's quite unintuitive reasoning on that scale.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -122,6 +140,129 @@ public ConjureStartTransactionsResponse startTransactions(ConjureStartTransactio
...
+     *   1. (L_A < U_A && L_B < U_B): for both leaders there is an interval of guaranteed leadership, which by
+     *   definitions cannot overlap. It is therefore trivial to order leadership of A and B.
+     *   2. (w.l.o.g. L_A < U_A && L_B >= U_B):
+     *      a) if U_B > U_A, A was the leader before B
+     *      b) if L_B < L_A, B was the leader before A
+     *      c) otherwise, we cannot determine the ordering
+     *   3. (L_A >= U_A && L_B >= U_B):
+     *      a) if U_B > L_A, A was the leader before B
+     *      b) if U_A > L_B, B was the leader before A
+     *      c) otherwise, we cannot determine the ordering
```
Jolyon-S, line 170: I _think_ this is right, but was also pretty hard to reason through - I had to write it out as above to verify that this works.

We're definitely going to want some solid ETE tests here to make sure that this is right. Fortunately, the cost of failure here is just dodgy metrics, so an ETE test in a separate PR is fine, but we'll still want to verify this logic.

Jolyon-S (commented):
Just going to release comments; probably best to sync up to review the logic here.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
gmaretic: It's literally said in 1 and 2:
A could not have lost leadership before T_1
A must have gained leadership before T_2

But yeah, maybe upper bound is confusing as it is not an upper bound on A's leadership, it's the best underapproximation of the leadership end time we can make with known information. Same for the lower bound, it's the best overaproximation of the leadership start time that we can make with known information. That is however, quite wordy...

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
gmaretic: That's not how mathematical proofs work

Jolyon-S: I know, but it would be nice to verify that your code actually works as intended here (unless you want to write a proof that your code is the same here too?)

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
Jolyon-S: Yes, I can read that, but reasoning over a single request/response pair doesn't really make a lot of sense mentally (at least, not to me), even if what you have written is correct. The other thing is that you didn't really say what the bound if of - it could refer to bounding when the time when A was the leader, or refer to the moment when A lost leadership. It might sound obvious, but in my opinion it is best to be as explicit as possible here so that someone without all context can read this with minimal frustration.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -122,6 +144,120 @@ public ConjureStartTransactionsResponse startTransactions(ConjureStartTransactio
...
+        }
+
+        UUID lastLongTermLeader = sortedLongTermLeaders.get(sortedLongTermLeaders.size() - 1);
+
+        Optional<Duration> result = durationToNextLeader(lowerBounds, upperBounds, leaders, lastLongTermLeader);
+        if (result.isPresent()) {
+            return result;
+        }
+
+        if (sortedLongTermLeaders.size() == 1) {
```
Jolyon-S, line 203: Can you just pair this check with the `isEmpty` one? e.g.
```
if (sortedLongTermLeaders.isEmpty() || sortedLongTermLeaders.size()` == 1) {
```

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+     *             [ B ]
+     *       <=========>
+     */
+    @Test
+    public void detectLeaderElectionWithFreshLeader() {
+        executeCalls(
+                ImmutableSingleCall.of(0L, 1L, LEADER_1),
+                ImmutableSingleCall.of(3L, 6L, LEADER_1),
+                ImmutableSingleCall.of(10L, 15L, LEADER_2));
+        assertExpectedDuration(Instant.ofEpochMilli(3L), Instant.ofEpochMilli(15L));
```
Jolyon-S, line 142: hm - given that the second call returns at 6L, why is the expected duration not 6L -> 15L? I realise that I'm probably misunderstanding your massive and lengthy explanation above, but when laid out this way, it just seems weird

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+    }
+
+    /**
+     * [ A ]
+     *           [ A ]
+     *       [     A     ]
+     *                     [ B ]
+     *           <=============>
```
Jolyon-S, line 150: start of B is after that last A call, so you should make the ASCII overlap

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+        assertExpectedDuration(Instant.ofEpochMilli(3L), Instant.ofEpochMilli(15L));
+    }
+
+    /**
+     * [ A ]
+     *       [ A ]
+     *             [ B ]
+     *                   [ B ]
+     *                         [ C ]
+     *             <===============>
```
Jolyon-S, line 185: 21 is the start of the second B call, so presumably this arrow should represent that?

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+     */
+    @Test
+    public void detectFreshLeaderElectionWithTwoLongTurnLeaders() {
```
Jolyon-S, line 188: nit: `LongTerm`?

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+        assertExpectedDuration(Instant.ofEpochMilli(3L), Instant.ofEpochMilli(15L));
+    }
+
+    /**
+     * [ A ]
+     *       [ A ]
+     *             [       B       ]
+     *                [ C ]
+     *                      [ C ]
+     *       <===========>
```
Jolyon-S, line 240: I think the arrow is one short

Jolyon-S (commented):
A few nits, and a couple comments on semantics

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
jeremyk-91: I think the theory here is that we want to have an upper bound, and what could have happened is (where the epsilon_i are small values ordered parallel to i)

T = 3, A makes request
T = 3 + epsilon, the leadership check on timelock passes, and timelock starts running the actual code that was protected by a leader election service
T = 3 + epsilon + mu, timelock leader elects
T = 6, the response finally comes back (note: this looks spooky but is allowed)

In general, if you have a request running 3 to 6 and this is the last request you know about that had a leader A, you know that the election didn't begin before 3, but you don't know that the election hadn't begun at any point after 3.

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+                ImmutableSingleCall.of(21L, 28L, LEADER_3));
+        assertThat(timelockService.calculateLastLeaderElectionDuration()).isEmpty();
+    }
+
+    /**
+     * [ A ]
+     *       [        A       ]
+     *          [ C ]
+     *                 [ C ]
+     *       <======>
```
jeremyk-91, line 274: this is clever. Thanks a lot for the ASCII art, really helps convey what's going on

jeremyk-91 (commented):
Thanks a lot for the PR, I'm happy with the algorithm/explanation, and the tests are well explained with the ASCII art as well.

@Jolyon-S do you want to have a final look before we go ahead?

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+                ImmutableSingleCall.of(10L, 15L, LEADER_2));
+        assertExpectedDuration(Instant.ofEpochMilli(3L), Instant.ofEpochMilli(15L));
+    }
+
+    /**
+     * [ A ]
+     *           [ A ]
+     *       [     A     ]
+     *                  [ B ]
+     *           <=============>
```
jeremyk-91, line 150: This should end at the end of B, surely

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -101,4 +115,207 @@ public void leaderIsInitialisedForAllMethods() {
...
+     */
+    @Test
+    public void detectLeaderElectionWithFreshLeaderOverlappingRequests() {
```
jeremyk-91, line 153: Could we have one for overlapping but non-consuming?

i.e.
```
[         A         ]
               [         A        ]
                          [         B     ]
               <==========================>
```

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
Jolyon-S: When you put it that way, it's obvious - the upper bound here being the last successfully sent request. Thanks!

Jolyon-S:
> Thanks a lot for the PR, I'm happy with the algorithm/explanation, and the tests are well explained with the ASCII art as well.
> 
> @Jolyon-S do you want to have a final look before we go ahead?

If all of your comments are actioned, it's good from me too. (I'm not going to bother approving, seeing as I have committed on this branch...)

gmaretic:
Someone does actually need to 👍

jeremyk-91 (approved):
👍

PR #5071, opened 2020-10-22
[Timelock Corruption]: Final Wiring part 1

sudiksha27:
**Goals (and why)**:
Wire corruption components

- Basically connects all pieces together

Modified existing tests

- Is it correct to initialize the PaxosHistoryProvider in PaxosResourceFactory?

- Did I miss something?

- PaxosResourceFactory

ASAP 🏃‍♀️

### DIFF
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionDetector.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionDetector.java
@@ -17,5 +17,5 @@
 package com.palantir.timelock.corruption.detection;
 
 public interface CorruptionDetector {
-    boolean shootTimeLock();
+    boolean shouldRejectRequests();
 }
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionDetector.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionDetector.java
@@ -17,5 +17,5 @@
 package com.palantir.timelock.corruption.detection;
 
 public interface CorruptionDetector {
-    boolean hasDetectedCorruption();
+    boolean shootTimeLock();
 }
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionHealthCheck.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionHealthCheck.java
@@ -16,21 +16,18 @@
 
 package com.palantir.timelock.corruption.detection;
 
-import com.google.common.collect.ImmutableList;
-import java.util.List;
-
 public class CorruptionHealthCheck {
     private final LocalCorruptionDetector localCorruptionDetector;
-    private final List<CorruptionDetector> localAndRemoteCorruptionDetectors;
+    private final RemoteCorruptionDetector remoteCorruptionDetector;
 
     public CorruptionHealthCheck(
             LocalCorruptionDetector localCorruptionDetector, RemoteCorruptionDetector remoteCorruptionDetector) {
         this.localCorruptionDetector = localCorruptionDetector;
-        this.localAndRemoteCorruptionDetectors = ImmutableList.of(localCorruptionDetector, remoteCorruptionDetector);
+        this.remoteCorruptionDetector = remoteCorruptionDetector;
     }
 
-    public boolean shootTimeLock() {
-        return localAndRemoteCorruptionDetectors.stream().anyMatch(detector -> detector.shootTimeLock());
+    public boolean shouldRejectRequests() {
+        return localCorruptionDetector.shouldRejectRequests() || remoteCorruptionDetector.shouldRejectRequests();
     }
 
     public CorruptionHealthReport localCorruptionReport() {
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionHealthCheck.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionHealthCheck.java
@@ -16,16 +16,25 @@
 
 package com.palantir.timelock.corruption.detection;
 
+import com.google.common.collect.ImmutableList;
 import java.util.List;
 
 public class CorruptionHealthCheck {
-    private final List<CorruptionDetector> corruptionDetectors;
+    private final LocalCorruptionDetector localCorruptionDetector;
+    private final List<CorruptionDetector> localAndRemoteCorruptionDetectors;
 
-    public CorruptionHealthCheck(List<CorruptionDetector> corruptionDetectors) {
-        this.corruptionDetectors = corruptionDetectors;
+    public CorruptionHealthCheck(LocalCorruptionDetector localCorruptionDetector,
+            RemoteCorruptionDetector remoteCorruptionDetector) {
+        this.localCorruptionDetector = localCorruptionDetector;
+        this.localAndRemoteCorruptionDetectors = ImmutableList.of(localCorruptionDetector, remoteCorruptionDetector);
     }
 
-    public boolean isHealthy() {
-        return corruptionDetectors.stream().noneMatch(CorruptionDetector::hasDetectedCorruption);
+    public boolean shootTimeLock() {
+        return localAndRemoteCorruptionDetectors.stream()
+                .anyMatch(detector -> detector.shootTimeLock());
+    }
+
+    public CorruptionHealthReport localCorruptionReport() {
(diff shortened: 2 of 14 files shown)


Jolyon-S (commented):
Just picked up some nits in a drive-by.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionDetector.java
```diff
@@ -17,5 +17,5 @@
 package com.palantir.timelock.corruption.detection;
 
 public interface CorruptionDetector {
-    boolean hasDetectedCorruption();
+    boolean shootTimeLock();
```
jeremyk-91, line 20: maybe `shouldRejectRequests()`? The impression I get if I come across a method like this is that it actually makes the server stop requests and returns true if successful/false if not

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/CorruptionHealthCheck.java
```diff
@@ -16,16 +16,24 @@
 
 package com.palantir.timelock.corruption.detection;
 
+import com.google.common.collect.ImmutableList;
 import java.util.List;
 
 public class CorruptionHealthCheck {
-    private final List<CorruptionDetector> corruptionDetectors;
+    private final LocalCorruptionDetector localCorruptionDetector;
+    private final List<CorruptionDetector> localAndRemoteCorruptionDetectors;
```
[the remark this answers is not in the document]
jeremyk-91: Have a look at `LocalAndRemotes`? It's a bit more powerful than what you need here, but has a nice way of accessing all or specifically the local.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -36,15 +43,49 @@ private HistoryAnalyzer() {
...
+    public static CorruptionHealthReport corruptionHealthReportForHistory(
+            List<CompletePaxosHistoryForNamespaceAndUseCase> history) {
+
+        Map<NamespaceAndUseCase, CorruptionCheckViolation> namespaceAndUseCaseCorruptionCheckViolationMap =
+                history.stream()
+                        .collect(Collectors.toMap(
+                                HistoryAnalyzer::extractNamespaceAndUseCase,
+                                HistoryAnalyzer::violatedCorruptionChecksForNamespaceAndUseCase));
+
+        SetMultimap<CorruptionCheckViolation, NamespaceAndUseCase> entryEntrySetMultimap = KeyedStream.stream(
```
jeremyk-91, line 55: Let's name this based on what the map actually does: `namespacesExhibitingViolations` or something like this.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -36,15 +43,49 @@ private HistoryAnalyzer() {
...
+            CompletePaxosHistoryForNamespaceAndUseCase historyForNamespaceAndUseCase) {
+        return ImmutableNamespaceAndUseCase.builder()
+                .namespace(historyForNamespaceAndUseCase.namespace())
+                .useCase(historyForNamespaceAndUseCase.useCase())
+                .build();
+    }
+
     @VisibleForTesting
-    static List<CorruptionCheckViolation> violatedCorruptionChecksForNamespaceAndUseCase(
+    static CorruptionCheckViolation violatedCorruptionChecksForNamespaceAndUseCase(
```
jeremyk-91, line 75: I think this should return List or some Collection given the name?

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -36,15 +43,49 @@ private HistoryAnalyzer() {
...
-                .filter(CorruptionCheckViolation::raiseErrorAlert)
-                .collect(Collectors.toList());
+        Function<CompletePaxosHistoryForNamespaceAndUseCase, CorruptionCheckViolation> divergedLearnedCheck =
+                HistoryAnalyzer::divergedLearners;
+        Function<CompletePaxosHistoryForNamespaceAndUseCase, CorruptionCheckViolation> learnedValueWithoutQuorum =
+                HistoryAnalyzer::learnedValueWithoutQuorum;
+        Function<CompletePaxosHistoryForNamespaceAndUseCase, CorruptionCheckViolation> greatestAcceptedValueNotLearned =
+                HistoryAnalyzer::greatestAcceptedValueNotLearned;
+
+        return Stream.of(divergedLearnedCheck, learnedValueWithoutQuorum, greatestAcceptedValueNotLearned)
```
jeremyk-91, line 84: Might be worth extracting a constant for the set of checks

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/LocalCorruptionDetector.java
```diff
@@ -31,46 +32,63 @@
...
-    private void killTimeLock() {
-        localCorruptionState = CorruptionStatus.DEFINITIVE_CORRUPTION_DETECTED_BY_LOCAL;
-        corruptionHandler.notifyRemoteServersOfCorruption();
+    private CorruptionHealthReport analyzeHistoryAndBuildCorruptionHealthReport() {
+        return HistoryAnalyzer.corruptionHealthReportForHistory(historyProvider.getHistory());
+    }
+
+    private void processLocalHealthReport() {
+        localCorruptionState = getLocalCorruptionState(localCorruptionReport);
+        if (localCorruptionState.shootTimeLock()) {
```
jeremyk-91, line 75: as flagged elsewhere, this gets confusing, I think because it looks like a command, but it's not

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/HistoryAnalyzerTest.java
```diff
@@ -81,9 +83,15 @@ public void correctlyPassesIfThereIsNotCorruption() {
...
+
         assertThat(HistoryAnalyzer.violatedCorruptionChecksForNamespaceAndUseCase(
                         Iterables.getOnlyElement(historyForAll)))
-                .hasSize(0);
+                .isEqualTo(CorruptionCheckViolation.NONE);
+
+        assertThat(HistoryAnalyzer.corruptionHealthReportForHistory(historyForAll)
+                        .statusesToNamespaceAndUseCase()
+                        .size())
+                .isEqualTo(0);
```
jeremyk-91, line 94: nit: Prefer isEmpty()

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/HistoryAnalyzerTest.java
```diff
@@ -81,9 +83,15 @@ public void correctlyPassesIfThereIsNotCorruption() {
...
 
         List<CompletePaxosHistoryForNamespaceAndUseCase> historyForAll = paxosLogHistoryProvider.getHistory();
+
         assertThat(HistoryAnalyzer.violatedCorruptionChecksForNamespaceAndUseCase(
                         Iterables.getOnlyElement(historyForAll)))
-                .hasSize(0);
+                .isEqualTo(CorruptionCheckViolation.NONE);
+
+        assertThat(HistoryAnalyzer.corruptionHealthReportForHistory(historyForAll)
+                        .statusesToNamespaceAndUseCase()
```
jeremyk-91, line 92: could we rename this `violatingStatuses...` or something like that?

jeremyk-91 (commented):
Looks correct. Main things:

- `shootTimeLock()` looks like a command, I'd suggest making it clearer that it is a predicate (e.g. `shouldRejectUserRequests()`)
- I don't really like the findFirst() when identifying rogue clients, it seems we should just report all failing checks when multiple are unhappy.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
sudiksha27: Right, the issue is that (and I missed this while writing #5057) these checks are to be executed in an order, e.g. the check for acceptor quorum for a learned value does not hold value if the learners have diverged as we anyway do not have a learner value to compare accepted values against.

jeremyk-91 (approved):
discussed: yeah ordered checks make a lot of sense!

PR #5073, opened 2020-10-22
Unify LeaderClock's LeadershipId with Lock Watch Log's Leader ID

Jolyon-S:
**Goals (and why)**:
* There is a concept of a Timelock Leader ID which is essentially a random UUID generated for each leader;
* This concept is used both in calls to `leaderTime` and lock watch relevant calls (`startTransactions`, `getCommitTimestamps`).
* However, these UUIDs are generated separately, and so cannot be used interchangeably (as they will appear to be different leaders when actually they are not).

* Pass the leadership ID generated in the `LeaderClock` to the `LockWatchingServiceImpl`'s constructor

* Unit test in `AsyncLockServiceEteTest` to verify UUIDs are equal.

* Double check that there are no breaking changes (I really don't anticipate any)
* Check that these are the same concepts - as far as I can tell, the UUID is generated at the same time as the one for LWSI was, but another pair of eyes here would be good.

`AsyncLockService`

Today / tomorrow


gmaretic (approved):
Thanks!

PR #5074, opened 2020-10-22
Do not kill timelock for no reason

gmaretic:
**Goals (and why)**:
Timelock should fail to start if SQLite migration has run, but the state is inconsistent. The way we are checking this is:

1. Verify that the migration cutoff would be the same if we ran the migration now
2. Verify that the greatest entry in the old log is present in the log

These checks are correct for paxos learner, because 1) will fail if old log has advanced and 2) will catch the special case where old learner advanced but is still below 50 so the cutoff matches. However, the calculation of the cutoff for the acceptor is based on the greatest entry observed by the **current learner**. Therefore, after the migration ran and the SQLite learner learned something, the cutoff for the learner is no longer going to match the persisted cutoff.

Ignore the first check when `migrateFrom` is specified, the second check should realistically be good enough to catch any bad cases.


jeremyk-91 (approved):
Good spot! 🚀

PR #5076, opened 2020-10-22
[TS] Tune sweepdelay metrics

gmaretic:
**Goals (and why)**:
Currently the metric reflects the actual delay of each sweep iteration. The problem with this is that if sweep is set up with more threads * nodes than shards, or we run into errors, this introduces noise into the metric. It was interesting to have this to see that it works, but given that these cases can be detected with the sweep outcomes, there is no need to pollute sweep delay.

Sweep delay now reports the on success delay only

Fixed unit tests

None

Tiny

Not high, but small

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
@@ -60,28 +60,32 @@ public class SweepDelayTest {
 
     @Test
     public void unableToAcquireShardReturnsMaxPause() {
+        delay.getNextPause(SUCCESS);
         assertThat(delay.getNextPause(SweepIterationResults.unableToAcquireShard()))
                 .isEqualTo(DEFAULT_MAX_PAUSE_MILLIS);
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/SweepDelayTest.java
```diff
@@ -60,28 +60,32 @@ public void configurationAboveDefaultMaximumIsRespected() {
 
     @Test
     public void unableToAcquireShardReturnsMaxPause() {
+        delay.getNextPause(SUCCESS);
         assertThat(delay.getNextPause(SweepIterationResults.unableToAcquireShard()))
                 .isEqualTo(DEFAULT_MAX_PAUSE_MILLIS);
-        assertThat(metrics).hasValue(DEFAULT_MAX_PAUSE_MILLIS);
+        assertThat(metrics).hasValue(INITIAL_DELAY);
     }
```
sudiksha27, line 67: Do we have tests for the case where sweepDelayMetricsUpdater is updated to not INITIAL_DELAY?

sudiksha27 (approved):
LGTM! Minor query.

PR #5028, opened 2020-10-13
[DB TimeLock] 3C.2 - Timestamp Storage and Series Extraction Preliminary Wiring

jeremyk-91:
**Goals (and why)**:
- Support getAllNamespaces(), which is used as part of the SQLite and CentOS migrations. While it's debatable whether these will be needed in the short run for users of DB TimeLock, I would strongly prefer for us to not have DB TimeLock behave differently from a standard SQLite TimeLock as far as possible. (e.g. when we want to perform VM migrations, the CLI needs to work, and I wouldn't want it to suddenly implode or fail silently). As of #5027, there is still the test of getAllNamespaces() that is explicitly ignored by DB TimeLock.
- This PR doesn't achieve that yet, but performs a fairly involved refactor that will be needed in the next part of this.

**Implementation Description (bullets)**:
- Migrate `PersistentNamespaceContext` to derive4j which works a lot better IMO
- Define an SPI interface `DbTimeLockFactory` that is an extended version of `AtlasDbFactory` including the ability to extract known timestamp series.

**Testing (What was existing testing like?  What have you done to improve it?)**:
This is atop the branch that supports #5025 so we verify that MNPTLSIT still passes even for db-timelock + postgres.

**Concerns (what feedback would you like?)**:
- Did I break anything badly?

**Where should we start reviewing?**: TimeLockAgent

**Priority (whenever / two weeks / yesterday)**: this week please

PR #5077, opened 2020-10-22
[DB TimeLock] 3Q - Simplify DbTimestampCreationSetting To Only Apply To DB TimeLock

jeremyk-91:
**Goals (and why)**:
- Make DB timelock code easier to reason about

- We had DbTimestampCreationSetting be an algebraic datatype with single and multi series options. 
- But we don't actually need to use the cases at runtime, because they are known at compile time: the path through AtlasDbFactory must be single series (embedded) while the path through DbTimeLockFactory must be multi-series.
- DbTimeLockFactory was only introduced in #5028 which I guess is why the original cases thing was needed.

Nothing new, existing tests passing should be fine

Did I accidentally break something?

**Where should we start reviewing?**: DbTimestampCreationSetting


sudiksha27 (commented):
Looks much cleaner. 

I think these changes do not break anything but would be great to have @gmaretic to take a look as well.

atlasdb-dbkvs-tests/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/DbTimestampStoreInvalidatorCreationTest.java
```diff
@@ -75,35 +71,15 @@ public void setUp() {
...
-        TimestampStoreInvalidator timestampStoreInvalidator = storeUpperLimitAndGetTimestampStoreInvalidator(
-                Optional.of(DbTimestampCreationSettings.singleSeries(Optional.of(otherTable))));
+        TimestampStoreInvalidator timestampStoreInvalidator =
+                storeUpperLimitAndGetTimestampStoreInvalidator(Optional.of(otherTable));
         assertThat(timestampStoreInvalidator.backupAndInvalidate()).isEqualTo(TIMESTAMP_1);
 
         assertBoundNotReadableAfterBeingPoisoned(otherStore);
     }
 
-    @Test
```
gmaretic, line 96: Why did we remove this test?

gmaretic (approved):
LGTM

atlasdb-dbkvs-tests/src/test/java/com/palantir/atlasdb/keyvalue/dbkvs/DbTimestampStoreInvalidatorCreationTest.java
jeremyk-91: Since the changes you can't physically create a config where single series is not implied. So this test becomes equivalent to `invalidatesDefaultTableForEmptyParameters()` below it

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/DbAtlasDbFactory.java
```diff
@@ -84,7 +80,7 @@ public KeyValueService createRawKeyValueService(
 
     @Override
     public ManagedTimestampService createManagedTimestampService(
-            KeyValueService rawKvs, Optional<DbTimestampCreationSetting> creationParameters, boolean initializeAsync) {
+            KeyValueService rawKvs, Optional<TableReference> tableReferenceOverride, boolean initializeAsync) {
         if (initializeAsync) {
```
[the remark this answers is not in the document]
jeremyk-91: Ah, that's a decent call. I think that makes sense

PR #5078, opened 2020-10-22
[TimeLock Corruption] | Implement progress tracker

sudiksha27:
**Goals (and why)**:
Adds a ProgressTracker to keep track of last verified sequence numbers + update last verified one history till that point is loaded from all servers.

- Add a ProgressTracker to maintain local cache, get lastVerified seq numbers and update lastVerified

- Maintain local cache of (lastVerifiedSeq, greatestSeqNumberToBeVerified) for each (namespace, useCase) pair

- Init state => lastVerifiedSeq = -1, greatestSeqNumberToBeVerified = latest learned value for respective (namespace, useCase) pair

- Each iteration will do two things - 

     - get seq bounds [lowerInclusive, upperInclusive] to load logs for corruption verification. If the lastVerifiedSeq > greatestSeqNumberToBeVerified, reset to init state

    - update the lastVerified with  upperInclusive after logs have been loaded

Added tests

Did I miss any edge cases in the ProgressTracker?

- No batching while making DB calls
 
- No mechanism for detecting inactive series. In the worst case, inactive series can get their progressState refreshed on each iteration.

PaxosLogHistoryProgressTracker.java

ASAP 🏃‍♀️


timelock-corruption-detection/src/main/java/com/palantir/timelock/history/PaxosLogHistoryProvider.java
```diff
@@ -205,7 +219,16 @@ private PaxosHistoryOnRemote fetchHistoryFromRemote(
...
+        return Maps.immutableEntry(
+                namespaceAndUseCase, HistoryQuery.of(namespaceAndUseCase, bounds.lower(), bounds.upper()));
+    }
+
+    private void updateProgressState(Map<NamespaceAndUseCase, SequenceBounds> namespaceAndUseCaseSequenceBoundsMap) {
+        namespaceAndUseCaseSequenceBoundsMap.forEach((key, value) -> {
+            long lastVerifiedSequence = value.upper();
+            verificationProgressStateCache.put(key, lastVerifiedSequence);
+            logVerificationProgressState.updateProgress(key.namespace(), key.useCase(), lastVerifiedSequence);
+        });
```
sudiksha27, line 232: Do this in a different thread along with reset progress? 

Logic for resetting  - fetch latest seq number for the namespace useCase tuple and see if it is less than lastVerified?

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/PaxosLogHistoryProgressTracker.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import javax.sql.DataSource;
+
+public class PaxosLogHistoryProgressTracker {
+    public static final int MAX_ROWS_ALLOWED = 500;
+
+    private final LogVerificationProgressState logVerificationProgressState;
+    private final SqlitePaxosStateLogHistory sqlitePaxosStateLogHistory;
+
+    private Map<NamespaceAndUseCase, ProgressComponents> verificationProgressStateCache = new ConcurrentHashMap<>();
```
Jolyon-S, line 37: As far as I can tell, you only ever put or get from this cache (i.e. never delete). I don't have full context on this piece of work, but I wonder if it is possible to OOM yourself? How many entries do you expect to be present in this cache?

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/models/ProgressComponents.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.timelock.history.models;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ProgressComponents {
```
Jolyon-S, line 22: I think these immutables are fine, but if you really want to avoid crossed wires, you can try a staged builder: https://immutables.github.io/immutable.html#staged-builder

(I appreciate not everyone likes them, but I do!)

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/remote/HistoryLoaderAndTransformer.java
```diff
@@ -36,8 +38,11 @@ private HistoryLoaderAndTransformer() {
 
     public static List<LogsForNamespaceAndUseCase> getLogsForHistoryQueries(
             LocalHistoryLoader localHistoryLoader, List<HistoryQuery> historyQueries) {
-        Map<NamespaceAndUseCase, Long> lastVerifiedSequences = historyQueries.stream()
-                .collect(Collectors.toMap(HistoryQuery::getNamespaceAndUseCase, HistoryQuery::getSeq, Math::min));
+        Map<NamespaceAndUseCase, SequenceBounds> lastVerifiedSequences = historyQueries.stream()
+                .collect(Collectors.toMap(
```
Jolyon-S, line 42: I'm not a huge fan - `KeyedStream.of(historyQueries).mapKeys(...).map(...)` gets you most of the way, although I'll confess that I'm not sure how to do the merging resolution in `KeyedStream` (although you can always collect to multimap).

timelock-corruption-detection/src/test/java/com/palantir/timelock/history/HistoryLoaderAndTransformerTest.java
```diff
@@ -73,11 +73,12 @@ public void setup() {
...
         writeToLogs(1, 100);
         int lastVerified = 27;
-        List<HistoryQuery> historyQueries = ImmutableList.of(HistoryQuery.of(NAMESPACE_AND_USE_CASE, lastVerified));
+        List<HistoryQuery> historyQueries =
+                ImmutableList.of(HistoryQuery.of(NAMESPACE_AND_USE_CASE, lastVerified, lastVerified + 500));
         List<LogsForNamespaceAndUseCase> paxosHistory =
                 HistoryLoaderAndTransformer.getLogsForHistoryQueries(history, historyQueries);
 
-        sanityCheckLoadedHistory(paxosHistory, 100 - lastVerified);
+        sanityCheckLoadedHistory(paxosHistory, 100 - lastVerified + 1);
```
Jolyon-S, line 81: Speaking of sanity checks, is this:

`(100 - lastVerified) + 1`
OR
`100 - (lastVerified + 1)`?

I assume the latter, although I'm never certain on how it determines this - so maybe bracket it?

Jolyon-S (commented):
Drive-by of a few things; didn't review in-depth semantics as I don't have full context on what's going on here.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/models/SequenceBounds.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface SequenceBounds {
+    @Value.Parameter
+    long lower();
+
+    @Value.Parameter
+    long upper();
```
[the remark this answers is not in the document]
jeremyk-91: Actually, I would strongly prefer for us to use inclusive ranges. A lot of the `lastVerified` stuff is misleading because `upper()` is not actually verified.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/sqlite/SqlitePaxosStateLogHistory.java
```diff
@@ -82,16 +96,20 @@ public LearnerAndAcceptorRecords getLearnerAndAcceptorLogsSince(
...
+                @Bind("lowerBound") long lowerBound,
+                @Bind("upperBound") long upperBound);
 
-        @SqlQuery("SELECT seq, val FROM paxosLog "
-                + "WHERE namespace = :namespace.value AND useCase = :useCase AND seq > :seq "
-                + "ORDER BY seq ASC LIMIT 500")
-        Map<Long, PaxosAcceptorData> getAcceptorLogsSince(
-                @BindPojo("namespace") Client namespace, @Bind("useCase") String useCase, @Bind("seq") long seq);
+        @SqlQuery("SELECT seq, val FROM paxosLog WHERE namespace = :namespace.value AND useCase = :useCase AND seq >="
+                + " :lowerBound AND seq < :upperBound")
```
jeremyk-91, line 108: See discussion elsewhere: I think I'd have a strong preference for having inclusive ranges here, because we've used terminology like `latestVerified` through many other parts of the code, which are not strictly correct when the upper bound has actually not been verified.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/PaxosLogHistoryProgressTracker.java
```diff
@@ -0,0 +1,112 @@
...
+
+        return logVerificationProgressState
+                .getProgressComponents(client, useCase)
+                .orElseGet(() -> logVerificationProgressState.resetProgressState(
+                        client, useCase, getLatestSequenceForNamespaceAndUseCase(namespaceAndUseCase)));
+    }
+
+    @VisibleForTesting
+    void updateProgressStateForNamespaceAndUseCase(NamespaceAndUseCase key, SequenceBounds value) {
+        long lastVerifiedSequence = value.upper();
```
jeremyk-91, line 75: This is an exclusive bound, so it is _not_ the last verified sequence (see comment on my thoughts on SequenceBound).

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/PaxosLogHistoryProgressTracker.java
```diff
@@ -0,0 +1,112 @@
...
+        verificationProgressStateCache.put(key, progressState);
+        logVerificationProgressState.updateProgress(key.namespace(), key.useCase(), lastVerifiedSequence);
+        return progressState;
+    }
+
+    private Optional<ProgressComponents> resetIfRequired(
+            NamespaceAndUseCase key, SequenceBounds value, ProgressComponents currentProgressState) {
+        if (value.upper() <= currentProgressState.progressLimit()) {
+            return Optional.empty();
+        }
```
jeremyk-91, line 101: This doesn't behave the way I think is intended: your seeding of the bound in `getLatestSequence` is inclusive, however your use of `value` here is exclusive. This means that even if there are no new entries, the validation task will repeatedly cycle over the entries (e.g. of an inactive client), which I doubt is intended.

jeremyk-91 (commented):
The structure makes sense, but I have some concerns around readability. Specifically, you use `lastVerifiedSequence` throughout most of the code, but associate it with an exclusive upper bound in `SequenceBounds` which means that the "last verified sequence" isn't actually verified. I'd suggest using inclusive bounds throughout.

I think this also leads to a strange edge case in `PaxosLogHistoryProgressTracker`: as written I think the tracker spins infinitely on an inactive series, because you're comparing an inclusive bound with an exclusive bound.

jeremyk-91:
I'd suggest having as integration tests (maybe not in this PR, but before the workstream is complete):
- verifying a sequence with over 500 entries (so that it needs to be verified in batches) works correctly
- an invalid sequence at entry 499 and 500 are caught
- inactive sequence does not get verified more than once.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/remote/HistoryLoaderAndTransformer.java
sudiksha27: I think what we have now is more readable

sudiksha27 (commented):
I agree, I'll add some ETE tests in a separate PR.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/models/SequenceBounds.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.timelock.history.models;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface SequenceBounds {
+    int MAX_ROWS_ALLOWED = 500;
```
jeremyk-91, line 23: It feels kind of dubious that this is here; it seems this is just a history query without a namespace and use case. If possible I'd urge you to try and kill this class altogether, but if that's not possible I would strongly prefer naming this class so that it's obvious that it is specific to history queries (e.g. `HistoryQuerySequenceRange`).

timelock-corruption-detection/src/test/java/com/palantir/timelock/history/utils/HistoryQueryUtils.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.palantir.timelock.history.utils;
+
+import com.palantir.paxos.Client;
+import com.palantir.paxos.ImmutableNamespaceAndUseCase;
+import com.palantir.paxos.NamespaceAndUseCase;
+import com.palantir.timelock.history.HistoryQuery;
+
+public class HistoryQueryUtils {
```
jeremyk-91, line 24: nit: the standard form for factory classes of this kind is the plural of the class name i.e. `HistoryQueries`

jeremyk-91 (commented):
🕐  1:00

Thanks for adding more comprehensive tests and changing the interval queries to operate on closed intervals; I think the PR looks correct now.

I'm afraid I think there are still some issues, largely around readability:
- I'm uncertain of the decomp: I don't understand the point of `SequenceBounds` when `HistoryQuery` already exists. I'd kill it if possible, and if not, given that you have a constant in the class and a factory that strongly suggest it is intended only for history queries, I would suggest naming it accordingly.
- Some variable names, in particular around the form of `Map<T, SequenceBounds> lastVerifiedSequences` need to be changed - in particular, they often refer to the bounds on queries that need to be made (while previously there was only a lower bound, so `lastVerifiedSequences` _was_ correct).
- I would urge you to use method names indicating the _intention_ behind your methods as opposed to just what the methods do. For example, `getBoundsSinceLastVerified(long lastVerified)` doesn't say much, and I'd argue the most natural interpretation of that is `[lastVerified + 1, inf)` - while if you describe it as `getBoundsForNextHistoryQuery(long lastVerified)` a reader will understand what it's for, and is more likely to have assumptions in line with what you're intending.

timelock-corruption-detection/src/main/java/com/palantir/timelock/history/sqlite/LogVerificationProgressState.java
```diff
@@ -49,38 +50,52 @@ private void initialize() {
...
 
     private <T> T execute(Function<LogVerificationProgressState.Queries, T> call) {
         return jdbi.withExtension(LogVerificationProgressState.Queries.class, call::apply);
     }
 
     public interface Queries {
-        @SqlUpdate("CREATE TABLE IF NOT EXISTS logVerificationProgress (namespace TEXT, useCase TEXT, seq BIGINT,"
-                + "PRIMARY KEY(namespace, useCase))")
+        @SqlUpdate("CREATE TABLE IF NOT EXISTS logVerificationProgress (namespace TEXT, useCase TEXT, lastVerifiedSeq"
+                + " BIGINT, greatestSeqNumberToBeVerified BIGINT, PRIMARY KEY(namespace, useCase))")
```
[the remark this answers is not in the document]
sudiksha27: This will break since https://github.com/palantir/atlasdb/pull/5071 merged, my bad, I will fix this.

jeremyk-91 (approved):
👍  Checkpointing of multiple values can be messy, but discussed offline - we think it's not unreasonable to remove it and use in memory state for remembering when we should reset. 

There is a degenerate condition where you might never verify old entries if the server is restarted extremely frequently, so that there is always one paxos round in between the verification task reading the bound and processing a batch, but I don't think we need to worry about that.

PR #5079, opened 2020-10-23
Leader Election Metrics 2a: Server side wiring 

Jolyon-S:
**Goals (and why)**:
* Add a new endpoint to the Timelock feedback service to allow for reporting of leader election metrics.
* Add wiring to aggregate received metrics

* Added a new endpoint
* Added a class to aggregate received metrics

* WIP

* I'm only adding a new endpoint - no requests will be sent yet. I assume this is fine?

* `timelock-api.yml` is probably a good start.

After https://github.com/palantir/atlasdb/pull/5069 merges.


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionMetricAggregator.java
```diff
@@ -0,0 +1,55 @@
...
+import com.palantir.atlasdb.util.MetricsManager;
+import com.palantir.atlasdb.util.SlidingWindowWeightedMeanGauge;
+import com.palantir.conjure.java.lib.SafeLong;
+import com.palantir.timelock.feedback.LeaderElectionStatistics;
+
+public final class LeaderElectionMetricAggregator {
+    private final SlidingWindowWeightedMeanGauge weightedGaugeP99;
+    private final SlidingWindowWeightedMeanGauge weightedGaugeP95;
+    private final SlidingWindowWeightedMeanGauge weightedGaugeMean;
+    private volatile long lastElectionTime = 0L;
```
Jolyon-S, line 29: I'm not a huge fan, but not sure what the best way to use a gauge here is.

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -197,13 +230,17 @@ private void assertExpectedDuration(Instant instant, Instant instant2) {
...
+        @Value.Parameter
         long responseMillis();
 
-        @Parameter
+        @Value.Parameter
         UUID responseLeader();
+
+        static SingleCall of(long requestMillis, long responseMillis, UUID responseLeader) {
+            return ImmutableSingleCall.of(requestMillis, responseMillis, responseLeader);
+        }
```
Jolyon-S, line 244: Driveby

gmaretic: Have fun rebasing :P

timelock-api/src/main/conjure/timelock-feedback.yml
```diff
@@ -17,6 +17,13 @@ types:
           p99: double
           oneMin: double
           errorRate: optional<double>
+      LeaderElectionStatistics:
+        fields:
+          p99: double
+          p95: double
+          mean: double
+          count: safelong
+          perceivedTime: optional<safelong>
```
gmaretic, line 26: durationEstimate?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionMetricAggregator.java
```diff
@@ -0,0 +1,53 @@
...
+                LeaderElectionMetricAggregator.class, "leaderElectionImpactP99", weightedGaugeP99);
+        metricsManager.registerMetric(LeaderElectionMetricAggregator.class, "leaderElectionDuration", lastElectionTime);
+    }
+
+    void report(LeaderElectionStatistics statistics) {
+        long count = statistics.getCount().longValue();
+        weightedGaugeMean.update(statistics.getMean(), count);
+        weightedGaugeP95.update(statistics.getP95(), count);
+        weightedGaugeP99.update(statistics.getP99(), count);
+        lastElectionTime.setValue(
```
gmaretic, line 50: Might be better to just not set when empty

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -115,6 +120,24 @@ public ConjureStartTransactionsResponse startTransactions(ConjureStartTransactio
...
+    }
+
+    @VisibleForTesting
+    LeaderElectionStatistics statisticsWithRegistry(TaggedMetricRegistry metricRegistry) {
+        Snapshot metricsSnapshot = metrics.observedDuration().getSnapshot();
+        LeaderElectionStatistics electionStatistics = LeaderElectionStatistics.builder()
+                .p99(metricsSnapshot.get99thPercentile())
+                .p95(metricsSnapshot.get95thPercentile())
+                .mean(metricsSnapshot.getMean())
+                .count(SafeLong.of(metrics.observedDuration().getCount()))
```
gmaretic, line 134: There is of course a small race condition here, but it's fine

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -115,6 +120,24 @@ public ConjureStartTransactionsResponse startTransactions(ConjureStartTransactio
                 .logId());
     }
 
+    public LeaderElectionStatistics statistics() {
+        return statisticsWithRegistry(new DefaultTaggedMetricRegistry());
+    }
+
+    @VisibleForTesting
+    LeaderElectionStatistics statisticsWithRegistry(TaggedMetricRegistry metricRegistry) {
```
gmaretic, line 128: This is a bit janky, but probably fine for tests. Rename to `getStatisticsAndSetRegistryTo` or something such?

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
Jolyon-S: I will, don't worry

gmaretic (commented):
LGTM, a few nits and needs rebasing on the base branch, but other than that 👍

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionMetricAggregator.java
Jolyon-S: Reasoning here being that, when no election is going on, we might want to set to 0 (else the metric will sit on the previous observed value between elections, which might give the impression that we're having a bad time)

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
Jolyon-S: Yea, I hate this but it does make it much easier to test.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
Jolyon-S: Did some digging and I believe that `metricSnapshot.size()` is equivalent, just without the race condition, so replacing with that.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionMetricAggregator.java
gmaretic: Yeah, still not buying it as that's exactly what the metric is supposed to do: report the last known leader election, and 0 will skew the picture

gmaretic (approved):

Jolyon-S:
:+1: seems legit

PR #5080, opened 2020-10-26
Fix hashCode related error-prone checks

carterkozak:
**Goals (and why)**:
follow baseline best practices

commits with automated code changes include the commands used to generate them.

n/a

n/a

Anywhere is fine, all code changes outside of build.gradle are automated.

ideally before more failures are introduced

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
@@ -78,7 +78,7 @@ public final class Namespace {
 
     @Override
     public int hashCode() {
-        return Objects.hash(name);
+        return Objects.hashCode(name);
     }
 
     @Override
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
@@ -78,7 +78,7 @@ public final class Namespace {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(name);
+        return 31 + Objects.hashCode(name);
     }
 
     @Override
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
@@ -18,8 +18,8 @@ package com.palantir.atlasdb.keyvalue.api;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.Objects;
 import com.palantir.logsafe.Preconditions;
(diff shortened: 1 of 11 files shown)


carterkozak:
Unfortunately we can't enforce these at error because there are failures in generated protobuf code. The error-prone plugin flag only disables warnings in generated code, errors still fail compilation.

lock-api-objects/src/main/java/com/palantir/lock/LockRequest.java
```diff
@@ -200,15 +200,14 @@ public boolean equals(@Nullable Object obj) {
...
+                && Objects.equals(blockingDuration, other.blockingDuration)
+                && Objects.equals(versionId, other.versionId);
     }
 
     @Override
     public int hashCode() {
         if (hashCode == null) {
-            hashCode = Objects.hashCode(
-                    lockMap, lockTimeout, lockGroupBehavior, blockingMode, blockingDuration, versionId);
+            hashCode = Objects.hash(lockMap, lockTimeout, lockGroupBehavior, blockingMode, blockingDuration, versionId);
```
jeremyk-91, line 210: I was a bit worried about this one because it could lead to fun times (if you deserialize something from the old version that has a different way of calculating the hash code and shove it into a set you have fun times), but think it's safe given that you can only create this through the serialization proxy, meaning that hash codes are always computed with the current JVM. I'd appreciate a double-check though.

jeremyk-91 (commented):
Not approving because of [merge when ready], but I'm basically happy if my explanation of a suspicious-looking line is correct.

lock-api-objects/src/main/java/com/palantir/lock/LockRequest.java
carterkozak: I don’t think we support java serialization across versions

carterkozak: That said, the hashCode field on serializable objects should be transient, and there's no reason for it to be volatile.

carterkozak: Not sure why we're doing all this boxing/unboxing either. We should use an `int` with a sentinel value (zero) for the uninitialized case. Some objects may be unlucky, however the hashcode is relatively inexpensive to recompute.

carterkozak:
@jeremyk-91 Any idea what's happening with CassandraTimestampsEteTest?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/Namespace.java
```diff
@@ -78,7 +78,7 @@ public String getName() {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(name);
+        return 31 + Objects.hashCode(name);
```
carterkozak, line 81: This produces the same result as the previous implementation, which was effectively `return Arrays.hashCode(new Object[] { name });`

jeremyk-91: sigh. I don't know where we rely on this, of all things, but it looks like it actually makes the tests pass.

jeremyk-91 (approved):
👍  thanks!

PR #5081, opened 2020-10-26
Fail on compiler warnings

carterkozak:
**Goals (and why)**:

Catch bugs at compile time rather than runtime.

Enable `-Werror`, and disable error-prone on generated projects (those suffixed `-protobufs`).
Warnings that do not currently pass are disabled for now, allowing current and new checks to be automatically validated.

These are tests.

gradle files.

The sooner it merges, the less likely we are to regress.


build.gradle
```diff
@@ -107,23 +107,98 @@ allprojects {
...
     tasks.withType(JavaCompile) {
+        options.compilerArgs += ['-Werror']
         // temporarily relax constraints until we can fix all violations
-        options.errorprone.warn 'CatchBlockLogException',
+        options.errorprone.disable 'AlmostJavadoc', 
+                'AnnotateFormatMethod',
+                'ArrayAsKeyOfSetOrMap',
+                'AssertEqualsArgumentOrderChecker',
+                'AssertionFailureIgnored',
+                'ByteBufferBackingArray',
```
carterkozak, line 117: Potential data correctness issues, we should fix the underlying code and enforce this asap.

build.gradle
```diff
@@ -107,23 +107,98 @@ allprojects {
...
+                'CacheLoaderNull',
+                'CatchBlockLogException',
+                'CatchFail',
                 'CheckReturnValue',
+                'ClassCanBeStatic',
```
carterkozak, line 122: Trivial to fix (as many of these are), should be done in a follow-up.

jeremyk-91 (approved):
Yeah, the list of exceptions is pretty gory but in principle this is the right thing to do, and should get us on a path to avoid further regressions.

build.gradle
```diff
@@ -31,7 +31,7 @@ buildscript {
 plugins {
     id 'com.github.johnrengelman.shadow' version '5.1.0'
     id 'com.palantir.configuration-resolver' version '0.4.0'
-    id 'com.palantir.consistent-versions' version '1.17.3'
+    id 'com.palantir.consistent-versions' version '1.26.1'
```
carterkozak, line 34: Note this upgrade reduced the gradle `CONFIGURING` time from ~50 seconds to ~4 seconds on my machine.

PR #5082, opened 2020-10-26
Enforce more error-prone checks

carterkozak:
This enforces checks with automated fixes, there are a few that should be trivial to fix by hand, but I haven't done that yet.


jkozlowski (approved):

carterkozak:
Missed `compileSuiteTestJava` and `compileIntegTestJava` when I validated/ran fixes locally. Updated.

PR #5083, opened 2020-10-27
Fix AsyncInitializerTest message assertion

carterkozak:
Previously it set the assertion failure message for a nonexistant
next assert rather than making an assertion about the input.


jeremyk-91 (approved):
nice spot!

PR #5084, opened 2020-10-27
[PSL] Better Logging and Option to Skip Validation + Truncate for PSL Migration

gmaretic:
**Goals (and why)**:
This PR does two things: improves the logging on the migration and validation tasks, and adds a timelock install config that allows us to skip validation and truncate old logs when we need this to move forward if safe with failing validation

- Added `toString()` to `PaxosAcceptorState`, basically no reason not to do it this way
- Wired in and added client and use case to the logs
- Added a `truncateAllRounds` method to PSL interface
- If the config is set to true, skip validation if migrated, and in any case truncate the file based log
Funnily enough found two bugs in PaxosStateLogImpl so fixed them along the way.

Basically added some new tests that are modified existing tests and verify that the changed behaviour makes sense.

Not the biggest fan of adding more dangerous stuff to the interface, but the alternative was to change the behaviour of `PaxosStateLogImpl` to truncate everything. The JavaDoc implicitly allows this, but in the vein of being more defensive we just don't need this kind of implementation detail change to somehow kick is in the teeth.

Tests

🔥

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
@@ -224,9 +224,7 @@ public class PaxosStateLogImpl<V extends Persistable & Versionable> implements P
             for (File file : files) {
                 long fileSeq = getSeqFromFilename(file);
                 if (fileSeq <= toDeleteInclusive) {
-                    if (!file.delete()) {
-                        log.warn("failed to delete log file {}", file.getAbsolutePath());
-                    }
+                    deleteLogFile(file);
                 } else {
                     break;
                 }
@@ -241,12 +239,15 @@ public class PaxosStateLogImpl<V extends Persistable & Versionable> implements P
         lock.writeLock().lock();
         try {
             File dir = new File(path);
-            boolean success = getLogEntries(dir).stream().map(File::delete).reduce(true, (x, y) -> x && y);
-            if (!success) {
-                log.warn("failed to delete all log files.");
-            }
+            getLogEntries(dir).forEach(PaxosStateLogImpl::deleteLogFile);
         } finally {
             lock.writeLock().unlock();
+        }
+    }
+
+    private static void deleteLogFile(File file) {
+        if (!file.delete()) {
+            log.warn("Failed to delete log file {}", SafeArg.of("path", file.getAbsolutePath()));
         }
     }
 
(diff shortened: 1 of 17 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
```diff
@@ -223,7 +224,7 @@ public void truncate(long toDeleteInclusive) {
             for (File file : files) {
                 long fileSeq = getSeqFromFilename(file);
                 if (fileSeq <= toDeleteInclusive) {
-                    if (file.delete()) {
+                    if (!file.delete()) {
                         log.warn("failed to delete log file {}", file.getAbsolutePath());
```
jeremyk-91, line 228: wat

jeremyk-91: good spot though!

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
```diff
@@ -235,10 +236,24 @@ public void truncate(long toDeleteInclusive) {
...
     }
 
+    @Override
+    public void truncateAllRounds() {
+        lock.writeLock().lock();
+        try {
+            File dir = new File(path);
+            boolean success = getLogEntries(dir).stream().map(File::delete).reduce(true, (x, y) -> x && y);
+            if (!success) {
+                log.warn("failed to delete all log files.");
```
jeremyk-91, line 246: nit: could we include, at least at `Unsafe` level, the directory involved?

leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
```diff
@@ -235,10 +236,24 @@ public void truncate(long toDeleteInclusive) {
...
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
     private List<File> getLogEntries(File dir) {
         File[] files = dir.listFiles();
         if (files == null) {
-            return null;
+            return ImmutableList.of();
```
jeremyk-91, line 256: I realise this case is handled already above in `getExtremeLogEntry`, but we should clean it up.

jeremyk-91 (approved):
🕐 0:32, 🏁 2:29

Implementation looks good, apart from a little thing in PaxosStateLogImpl that should be cleaned up. Thanks for very detailed tests!

My main question here is whether we really want to couple both "skip the check" and "truncate the old log". I imagine we might want to do the first to unblock a stack which we know to not appear corrupted but not the second if we want to try and do an RCA. But I think the name is clear enough, and someone on storage support should know that truncating the log means, well, truncating the log.

PR #5086, opened 2020-10-27
Fix and enforce ByteBufferBackingArray

carterkozak:
**Goals (and why)**:
ByteBufferBackingArray failures can result in data corruption. There's no guarantee that an arbitrary heap buffer uses offset zero.


jeremyk-91 (approved):
🕐 0:07, 🏁 1d.

Cool, makes sense. Is there anything that needs to be done / thing we need to check retroactively? I took a cursory pass of the host-partitioner/sweep (which use `allocate()` and `wrap()`) and poked into the thrift code for receiving buffers from Cassandra and it looks like those also use `TBaseHelper.copyBinary` which calls to `wrap()`, so I think we're actually safe (but we should nonetheless be defensive of course!)

PR #5087, opened 2020-10-27
Simplify protobuf configuration

pkoenig10:
**Goals (and why)**:
Simplify Gradle configuration

Use the default generated protobuf directory. The protobuf plugin [integrates with the `idea` plugin](https://github.com/google/protobuf-gradle-plugin#intellij-idea) and marks the generated directories as a source directories.

Screen Shot 2020-10-27 at 3 08 04 PM

This change means we no longer need to disable error-prone in these projects.


carterkozak (approved):
Nice!

jeremyk-91 (approved):
❤️ Thanks!

PR #5088, opened 2020-10-27
[LE] Server Side Metrics Accumulator For Leadership Duration

gmaretic:
**Goals (and why)**:
Currently, each client will send to timelock its view of how long the leader election took. These will initially be imprecise, but should stabilise pretty quickly as the shortest duration is the most precise estimate here. Also, for each client the leader IDs will be unique. The idea is to accumulate these durations in a histogram so that exactly one duration is reported per client once we have achieved confidence.

Whenever we get an update, we check if we have already reported the duration for this leader id pair, if yes we ignore. Otherwise, we keep the minimum seen so far for this pair and increase the update count for it. One we hit 10 updates we report the current minimum and remember this pair so it is never reported again.

Added quite extensive unit tests, with some pretty aggressive multithreaded behaviour.

I'm not concerned about correctness, but does this type of accumulation where we basically have a histogram using a sliding window over 5 minutes (should it be longer?) that accumulates each client's view of that leader election give useful information?
 
Tests?

Before Wednesday


lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -352,10 +354,13 @@ public void statisticsCausesMetricRegistryToBeReset() {
         verifyNoMoreInteractions(mockedTimer);
     }
 
-    private void assertExpectedDuration(Instant instant, Instant instant2) {
-        Optional<Duration> estimatedDuration = timelockService.calculateLastLeaderElectionDuration();
-        assertThat(estimatedDuration).isPresent();
-        assertThat(estimatedDuration.get()).isEqualTo(Duration.between(instant, instant2));
+    private void assertExpectedDurationAndLeaders(Instant instant, Instant instant2, UUID old, UUID next) {
```
Jolyon-S, line 357: nit:
```suggestion
    private void assertExpectedDurationAndLeaders(Instant instant, Instant instant2, UUID oldLeader, UUID newLeader) {
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionMetricAggregator.java
```diff
@@ -16,39 +16,41 @@
...
                 LeaderElectionMetricAggregator.class, "leaderElectionImpactP95", weightedGaugeP95);
         metricsManager.registerMetric(
                 LeaderElectionMetricAggregator.class, "leaderElectionImpactP99", weightedGaugeP99);
-        metricsManager.registerMetric(
-                LeaderElectionMetricAggregator.class, "leaderElectionEstimate", leaderElectionEstimate);
+        Histogram leaderElectionHistogram = metricsManager.registerOrGetHistogram(
+                LeaderElectionMetricAggregator.class,
+                "leaderElectionDurationEstimate",
+                () -> new Histogram(new SlidingTimeWindowArrayReservoir(5, TimeUnit.MINUTES)));
+        leaderElectionDurationAccumulator = new LeaderElectionDurationAccumulator(leaderElectionHistogram::update, 10);
```
Jolyon-S, line 46: nit: can we extract a constant for this number? Possibly above too although for some reason I don't mind as much for the reservoir.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.LongConsumer;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Parameter;
+
+public class LeaderElectionDurationAccumulator {
+    private Set<LeadersContext> alreadyReportedLeaderElections = ConcurrentHashMap.newKeySet();
+    private Map<LeadersContext, ModifiableSoakingDuration> currentlySoaking = new ConcurrentHashMap<>();
+
+    private final LongConsumer consumer;
```
Jolyon-S, line 34: nit: rename to something informative, like `metricsConsumer` or `updateConsumer` etc.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+        this.consumer = consumer;
+        this.updatesToAchieveConfidence = updatesToAchieveConfidence;
+        Preconditions.checkArgument(
+                updatesToAchieveConfidence > 1,
+                "Number of required updates must be greater than 1.",
+                SafeArg.of("updatesToAchieveConfidence", updatesToAchieveConfidence));
+    }
+
+    public void add(LeaderElectionDuration duration) {
+        LeadersContext leadersContext = LeadersContext.of(duration.getOldLeader(), duration.getNewLeader());
```
Jolyon-S, line 55: I'm not a fan of using `of` here, given that they have the same type - this is one of the places where I would strongly prefer a builder despite there being only two params.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+        if (alreadyReportedLeaderElections.contains(leadersContext)) {
+            return;
+        }
+        currentlySoaking.compute(
+                leadersContext,
+                (context, previous) -> increaseConfidence(
+                        context, previous, duration.getDuration().longValue()));
```
Jolyon-S, line 62: I think this is cleaner (with proper indentation, of course)
```suggestion
        if (!alreadyReportedLeaderElections.contains(leadersContext)) {
                   currentlySoaking.compute(
                            leadersContext,
                           (context, previous) -> increaseConfidence(
                            context, previous, duration.getDuration().longValue()));
        }
        
```

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private ModifiableSoakingDuration increaseConfidence(
+            LeadersContext context, ModifiableSoakingDuration accumulatedSoakingDuration, long duration) {
+        if (alreadyReportedLeaderElections.contains(context)) {
```
Jolyon-S, line 67: nit: can you add a newline between these if statements? they feel very crowded

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+            return null;
+        }
+        if (accumulatedSoakingDuration == null) {
+            return ModifiableSoakingDuration.create(duration, 1);
```
Jolyon-S, line 71: nit: would it be cleaner to add a method on `SoakingDuration` where you just pass a duration and it creates it with 1 set?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        long accumulatedDuration = accumulatedSoakingDuration.value();
```
Jolyon-S, line 73: nit: inline?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+        if (accumulatedDuration > duration) {
+            accumulatedSoakingDuration.setValue(duration);
+        }
+        int currentCount = accumulatedSoakingDuration.count() + 1;
+        if (currentCount >= updatesToAchieveConfidence) {
+            alreadyReportedLeaderElections.add(context);
+            consumer.accept(accumulatedSoakingDuration.value());
+            return null;
+        }
+        accumulatedSoakingDuration.setCount(currentCount);
```
Jolyon-S, line 83: I feel like you should increase the count and duration together

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+    public void consumeOnlyOnceUsingMinFromFirstFiveUpdates() {
+        leaderElectionResultsWithDurationInRandomOrder(LEADER_1, LEADER_2, 5, 15);
+        leaderElectionResultsWithDurationInRandomOrder(LEADER_1, LEADER_2, 10, 1);
+
+        verify(mockConsumer).accept(15L);
+        verifyNoMoreInteractions(mockConsumer);
+    }
+
+    @Test
+    public void consumeOnceForEachLeaderPair() {
```
Jolyon-S, line 73: I think it would be best if the range of values for each pair (other than the repeated ones) should be non-intersecting - i.e. for leader (2,3) you should set the min to say 200 to make 100% certain that you aren't getting a spurious result.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        verify(mockConsumer).accept(7L);
+        verifyNoMoreInteractions(mockConsumer);
+    }
+
+    @Test
+    @SuppressWarnings("ExecutorSubmitRunnableFutureIgnored")
+    public void testInterleavingNoPause() throws InterruptedException {
+        ExecutorService executorService = PTExecutors.newFixedThreadPool(3);
+        accumulator = new LeaderElectionDurationAccumulator(mockConsumer, 300);
+        executorService.submit(() -> leaderElectionResultsWithDurationInRandomOrder(LEADER_1, LEADER_2, 300, 1));
```
Jolyon-S, line 90: The comment above is especially important here, IMO.

Jolyon-S (commented):
A load of nits but I think that this makes sense semantically - and I do think that it will provide some better data. If the data doesn't change, I'm pretty certain we send it (or we will) anyway - so if we have already reached a steady-state, this will just report that value after some time.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
jeremyk-91: I think it's still technically okay (assuming each election generates one thing) because the 1L could only have come from the first call and so on, though it is probably clearer in terms of readability to have non-overlapping ranges.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+                IntStream.range(0, 100).map(x -> x * 100).boxed().collect(Collectors.toList());
+        Collections.shuffle(durationBuckets);
+
+        durationBuckets.forEach(duration ->
+                executorService.submit(() -> leaderElectionResultsWithPause(LEADER_1, LEADER_2, 50, duration)));
+
+        executorService.shutdown();
+        executorService.awaitTermination(1, TimeUnit.MINUTES);
+
+        verify(mockConsumer).accept(0L);
```
jeremyk-91, line 134: ~I'm possibly confused: how do we guarantee zero is processed in the first five~

Ah, this test has an accumulator with 5000. What you have is fine, don't really like that it relies on 5000 being 50 * 100 (maybe introduce some local variables so that it's obvious?) / if you change line 129 to 60 the test breaks, so maybe we could change that slightly

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.LongConsumer;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Parameter;
+
+public class LeaderElectionDurationAccumulator {
+    private Set<LeadersContext> alreadyReportedLeaderElections = ConcurrentHashMap.newKeySet();
```
jeremyk-91, line 31: Looks like this grows unbounded. Is this a problem / should we consider using some kind of cache with an ExpiryAfterWrite/Access?

jeremyk-91 (commented):
🕐 0:27 🏁 4:15 (since Jolyon ping)

I think I broadly agree with the architecture of this - my main flag would be the unbounded set in `LeaderElectionDurationAccumulator` and perhaps an opportunity to not use `Modifiables`. Other than that, comments are largely style-based and I'm in sync with most of @Jolyon-S's review.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
Jolyon-S: Only L1->L3 _should_ have generated the 7, but if you have a min of 5 and 10 requests, then there exists a request of time 7 there, even if they should report 5. By separating this, you _guarantee_ that it is coming from the right place.

jeremyk-91: discussed offline: subject to the one-from-one-election assumption, that should be fine as if, e.g. the [7,11] range gave a 10 instead, there's no way for the [10,14] or [15,19] ranges to push a 7. Though let's just discard the need for that assumption and use disjoint ranges.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
Jolyon-S: Agreed - removed use of modifiables.

timelock-api/src/main/conjure/timelock-feedback.yml
```diff
@@ -17,13 +17,18 @@ types:
           p99: double
           oneMin: double
           errorRate: optional<double>
+      LeaderElectionDuration:
+        fields:
+          oldLeader: uuid
+          newLeader: uuid
+          duration: safelong
```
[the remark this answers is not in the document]
Jolyon-S: agreed - done

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
```diff
@@ -0,0 +1,108 @@
...
+        @Parameter
+        UUID next();
+
+        static LeadersContext of(UUID previous, UUID next) {
+            return ImmutableLeadersContext.of(previous, next);
+        }
+    }
+
+    @Value.Modifiable
+    interface SoakingDuration {
```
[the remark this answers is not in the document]
Jolyon-S: `updatesToAchieveConfidence * clients` per election, but still fine to be Immutable IMO

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
Jolyon-S: Factored out two variables here, should make it clearer.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
Jolyon-S: Let's say that you have 50 clients, so each leader election you store 50 * 2 UUIDs, which are 16 bytes IIRC. So that's 160 bytes per leader election. You'd need say >6500 leader elections to occur in the lifetime of this accumulator before you'd have 1MB of space allocated.

I'm not by any means against making this an expiry cache, but probably won't be necessary.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulator.java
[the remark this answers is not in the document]
jeremyk-91: Ah, okay. Not the best then, but it's at most a few thousand small objects so probably fine

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Test
+    @SuppressWarnings("ExecutorSubmitRunnableFutureIgnored")
+    public void testManyUpdatesForSameLeaders() throws InterruptedException {
+        ExecutorService executorService = PTExecutors.newFixedThreadPool(50);
+        int numBuckets = 100;
+        int requestsPerBucket = 50;
+        accumulator = new LeaderElectionDurationAccumulator(mockConsumer, numBuckets * requestsPerBucket);
+        List<Integer> durationBuckets =
+                IntStream.range(0, numBuckets).map(x -> x * numBuckets).boxed().collect(Collectors.toList());
```
jeremyk-91, line 127: nit: I realise it's the same as the original test, though really the second multiplicand seems like it shouldn't be tied to `numBuckets` but `requestsPerBucket` (and I guess if you want a bit of additional space that's fine).

jeremyk-91 (approved):
👍  One very small thing. Otherwise looks good.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/adjudicate/LeaderElectionDurationAccumulatorTest.java
Jolyon-S: Yes, I actually realised this when I wrote this. I suppose this is wrong if requestsPerBucket > numBuckets (which we could do), but I was thinking of having additional space (i.e. making sure to test with space). I think I'll just make this requestsPerBucket though, as it makes the logic easier.

PR #5091, opened 2020-10-30
Fix build fail

Jolyon-S:
**Goals (and why)**:
Failing builds

now 🔥


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/memory/InMemoryDbTimeLockFactory.java
```diff
@@ -51,7 +51,7 @@ public KeyValueService createRawKeyValueService(
     public ManagedTimestampService createManagedTimestampService(
             KeyValueService rawKvs, DbTimestampCreationSetting dbTimestampCreationSetting, boolean initializeAsync) {
         return services.computeIfAbsent(
-                dbTimestampCreationSetting.timestampSeries(), (unused) -> new InMemoryTimestampService());
+                dbTimestampCreationSetting.timestampSeries(), unused -> new InMemoryTimestampService());
     }
```
sudiksha27, line 55: may have to underscore it? but if you've run the build  this is fine.

sudiksha27 (approved):
👍  Merge when green.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/memory/InMemoryDbTimeLockFactory.java
Jolyon-S: I _think_ that if you call it "unused" it doesn't mind, but we'll see!

jeremyk-91 (approved):
👍 DPR changes were written in a pairing session.

PR #5092, opened 2020-10-31
Prototype of getSortedColumns

j-baker:
Internally we do a fair amount of merge sorting between rows, and this
is quite inefficient within Atlas. How it goes wrong is related to the
batching - basically each batch is lock checked, timestamp checked
independently, and so you get the nice waterfall in the gantt chart.

Here we add a specific method that does the sorting and merging, giving
the user an iterator which broadly does the right thing.

Skip all the rebatching stuff because sweep works nicely now.

Obviously I have no tests, and there are quite a few that you'd need to
write - especially since I've caught three more bugs in the original
getColumnsRange implementation (unsafe usages of transformEntries).

The other question I have is: most use cases (although not all) for this
actually read a fixed number of things. We could probably simplify the
contract a bit (although the implementation is already very clean) by
returning a Map of the first n or whatever.

Lastly, there is some downside of this. The present approach means that
if you e.g. read _all_ the columns of 1000 rows using distinct getColumnRange
calls, they'll be batched together at conflict checking. This approach
cannot extend in this way, so if you make 1000 calls, you'll do 1000
waterfalls here. That said, based on use cases this is probably fine -
the fanouts we see _should_ be on gets.

It does mean that if one row is requested, we simply delegate to the
present method of read only conflict checking since this will have
more opportunities to be optimized in conflict checking.

I think long term there's some fairly strong argument here that we're
suffering due to our KVS interface. IRL, the mgms implementation if
exposed on KVS would make the conflict checking here be awesome.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
@@ -131,6 +131,21 @@ public interface Transaction {
             TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection);
 
     /**
+     * Returns an iterator over cell-value pairs within {@code tableRef} for the specified {@code rows}, where the
+     * columns fall within the provided {@link BatchColumnRangeSelection}. The single provided {@link BatchColumnRangeSelection}
+     * applies to all of the rows.
+     *
+     * The returned iterator is guaranteed contains cells sorted first in order of column, then in order of row.
+     *
+     * @param tableRef table to load values from
+     * @param rows unique rows to apply column range selection to
+     * @param columnRangeSelection range of columns and batch size to load for all rows provided.
+     * @return iterator of cells matching predicate in rows, following ordering outlined above.
+     */
+    Iterator<Map.Entry<Cell, byte[]>> getSortedColumns(
+            TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection);
+
+    /**
      * Gets the values associated for each cell in {@code cells} from table specified by {@code tableRef}.
      *
      * @param tableRef the table from which to get the values
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -15,6 +15,8 @@
  */
 package com.palantir.atlasdb.transaction.impl;
 
+import static java.util.stream.Collectors.toList;
+
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Timer;
@@ -39,6 +41,7 @@ import com.google.common.collect.Multimaps;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.PeekingIterator;
 import com.google.common.collect.Sets;
+import com.google.common.collect.Streams;
 import com.google.common.primitives.UnsignedBytes;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -106,6 +109,7 @@ import com.palantir.common.base.ClosableIterators;
 import com.palantir.common.base.ForwardingClosableIterator;
 import com.palantir.common.collect.IteratorUtils;
 import com.palantir.common.collect.MapEntries;
+import com.palantir.common.streams.KeyedStream;
 import com.palantir.common.streams.MoreStreams;
 import com.palantir.lock.AtlasCellLockDescriptor;
 import com.palantir.lock.AtlasRowLockDescriptor;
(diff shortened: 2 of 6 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -431,6 +435,58 @@ public void markTableInvolved(TableReference tableRef) {
...
+        int batchSize = Math.max(
+                Math.min(100, columnRangeSelection.getBatchHint()), columnRangeSelection.getBatchHint() / size);
+        BatchColumnRangeSelection perBatchSelection = BatchColumnRangeSelection.create(
+                columnRangeSelection.getStartCol(), columnRangeSelection.getEndCol(), batchSize);
+        Map<byte[], RowColumnRangeIterator> rawResults =
+                keyValueService.getRowsColumnRange(tableRef, rows, perBatchSelection, getStartTimestamp());
+
+        Iterator<Map.Entry<Cell, Value>> cells = mergeColumnFirst(rawResults.values());
+        Iterator<Map.Entry<Cell, Value>> postFilterIterator =
+                getRowColumnRangePostFiltered(tableRef, cells, columnRangeSelection.getBatchHint());
```
jeremyk-91, line 461: As written, this method performs a re-sort on the entries so that they will not be in the column-first order but instead in row query order.

I wrote the following test, which fails:

```
    @Test
    public void getSortedColumns() {
        byte[] row1 = "foo".getBytes();
        byte[] row2 = "bar".getBytes();
        byte[] col1 = "a".getBytes();
        byte[] col2 = "b".getBytes();

        Cell rowOneColumnOneCell = Cell.create(row1, "a".getBytes());
        Cell rowOneColumnTwoCell = Cell.create(row1, "b".getBytes());
        Cell rowTwoColumnOneCell = Cell.create(row2, "a".getBytes());
        Cell rowTwoColumnTwoCell = Cell.create(row2, "b".getBytes());
        byte[] value = new byte[1];

        serializableTxManager.runTaskWithRetry(tx -> {
            tx.put(TABLE, ImmutableMap.of(rowOneColumnOneCell, value, rowOneColumnTwoCell, value,
                    rowTwoColumnOneCell, value, rowTwoColumnTwoCell, value));
            return null;
        });

        List<Cell> entries = serializableTxManager.runTaskWithRetry(tx -> {
            Iterator<Entry<Cell, byte[]>> sortedColumns =
                    tx.getSortedColumns(TABLE, ImmutableList.of(row1, row2), BatchColumnRangeSelection.create(
                            PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 1000));
            return Streams.stream(sortedColumns).map(Entry::getKey).collect(Collectors.toList());
        });
        org.assertj.core.api.Assertions.assertThat(entries).containsExactly(
                rowTwoColumnOneCell,
                rowOneColumnOneCell,
                rowTwoColumnTwoCell,
                rowOneColumnTwoCell); // Nope, order is R2C1, R2C2, R1C1, R1C2
    }
```

Basically I think there's some additional processing around this that means we don't want to just grab the same impl from getRowsColumnRangeIterator. I get that a lot of the logic _is_ similar, but there's a fair bit of postprocessing that was written for that method that's kind of specific there.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -130,6 +130,21 @@
     Map<byte[], Iterator<Map.Entry<Cell, byte[]>>> getRowsColumnRangeIterator(
             TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection);
 
+    /**
+     * Returns an iterator over cell-value pairs within {@code tableRef} for the specified {@code rows}, where the
+     * columns fall within the provided {@link BatchColumnRangeSelection}. The single provided {@link BatchColumnRangeSelection}
+     * applies to all of the rows.
+     *
+     * The returned iterator is guaranteed contains cells sorted first in order of column, then in order of row.
```
jeremyk-91, line 138: Probably worth clarifying that order of row here is the lexicographic ordering of rows (which is crucially different from that in e.g. `getRowsColumnRangeIterator`).

jeremyk-91 (commented):
🕐 0:40

I still need to look at `SerializableTransaction` impl/refactoring. Was focusing mostly on the `Snapshot` part of this. I think Snapshot makes sense and is very close, though needs a little bit of refactoring to avoid a probably unexpected side effect of getRowsColumnRangeIterator's merge (which is part of fixing the P0 that cropped up a while back with the ordering).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: The inconsistency with the other endpoints is a little uncomfortable, so it might be better to adhere to that unless it's too costly to do so.

jeremyk-91:
grr, should have marked this as `long-running`

PR #5094, opened 2020-11-03
[LE] Add timing to additional endpoints

Jolyon-S:
**Goals (and why)**:
There are two other responses that already have a leader ID as part of the response. I figured that we may as well get metrics on these endpoints too.

* Extracted the leader ID from `leaderTime` and `refreshLocks`.

* Added tests to confirm that all methods that we intend are being timed.

* This should not affect performance - right?

* `LeaderElectionReportingTimelockService`

Today


lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -49,13 +55,17 @@
 import org.mockito.stubbing.OngoingStubbing;
 
 public class LeaderElectionReportingTimelockServiceTest {
-    private static final UUID LEADER_1 = UUID.randomUUID();
+    private static final LeaderTime LEADER_1_TIME = LeaderTime.of(LeadershipId.random(), NanoTime.createForTests(999L));
```
Jolyon-S, line 58: the method to create from a UUID is not public.

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
```diff
@@ -302,6 +335,11 @@ public void detectWhenOldLeaderBecomesLongTerm() {
         assertExpectedDuration(Instant.ofEpochMilli(3L), Instant.ofEpochMilli(15L));
     }
 
+    private void verifyEndpointRunsWithTiming(Runnable endpointCall) {
+        endpointCall.run();
+        verify(mockedClock, times(2)).instant();
```
jeremyk-91, line 340: a bit leaky, but I don't really have a better suggestion.

jeremyk-91 (approved):
🕐 0:03

👍

lock-api/src/test/java/com/palantir/lock/client/LeaderElectionReportingTimelockServiceTest.java
Jolyon-S: I totally agree with both sentiments here :)

PR #5096, opened 2020-11-04
[Cross Client Batching - 1] | Refactor AtlasDB

sudiksha27:
**Goals (and why)**:
Extract out interfaces for batching endpoints. The final goal here is to be able to inject batcher and in absence of that fallback to current implementation.

Extracts interfaces for CommitTimestampGetter and LeaderTimeGetter

Existing tests should suffice

- Is the refactor in the right place?

- Did I break something horribly?

CommitTimestampGetter, LeaderTimeGetter

Today 🏃‍♀️


lock-api/src/main/java/com/palantir/lock/client/CommitTimestampGetter.java
```diff
@@ -16,96 +16,11 @@
...
-        autobatcher.close();
-    }
-
-    @Value.Immutable
-    interface Request {
-        long startTs();
-
-        LockToken commitLocksToken();
-    }
+    void close();
```
Jolyon-S, line 25: Remove this method - you extend `AutoCloseable` anyway so any class that implements this interface will have to implement close.

lock-api/src/main/java/com/palantir/lock/client/CoalescingLeaderTimeGetter.java
```diff
@@ -0,0 +1,37 @@
...
+
+public final class CoalescingLeaderTimeGetter implements LeaderTimeGetter {
+    private final CoalescingSupplier<LeaderTime> time;
+
+    private CoalescingLeaderTimeGetter(CoalescingSupplier<LeaderTime> time) {
+        this.time = time;
+    }
+
+    public static LeaderTimeGetter create(NamespacedConjureTimelockService delegate) {
+        return new CoalescingLeaderTimeGetter(new CoalescingSupplier<>(delegate::leaderTime));
```
Jolyon-S, line 30: I wonder if it would be better if you just had the create pass in `Supplier<LeaderTime>`, so that this class does not need to know about `NamespacedConjureTimelockService` in any way.

lock-api/src/main/java/com/palantir/lock/client/CoalescingLeaderTimeGetter.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.common.concurrent.CoalescingSupplier;
+import com.palantir.lock.v2.LeaderTime;
+
+public final class CoalescingLeaderTimeGetter implements LeaderTimeGetter {
+    private final CoalescingSupplier<LeaderTime> time;
```
Jolyon-S, line 23: nit: while the field in `LockLeaseService` was indeed called `time` before, I think this is a good opportunity to rename; maybe `leaderTimeSupplier` or `timeSupplier`?

lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
```diff
@@ -0,0 +1,112 @@
...
+import com.palantir.lock.watch.TransactionUpdate;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import org.immutables.value.Value;
+
+final class BatchingCommitTimestampGetter implements CommitTimestampGetter {
```
Jolyon-S, line 39: As far as I can tell, you've copied and pasted this from the old one (which you made an interface). Did you make any changes? If yes, let me know and I'll review critically - otherwise I'll just assume that this code is good because it's the same as before.

Jolyon-S (commented):
🕐 00:10, 🏁 00:30

A couple of nits and a sanity check, but otherwise looks good from my end.

lock-api/src/main/java/com/palantir/lock/client/CommitTimestampGetter.java
sudiksha27: I wanted to kill exception thrown by AutoCloseable#close (which was the case before my refactor), here, the batcher only calls DisrupterAutobatcher#close which does the exception handling.

lock-api/src/main/java/com/palantir/lock/client/CoalescingLeaderTimeGetter.java
sudiksha27: Makes sense

lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
sudiksha27: That's right, I have not made changes to the logic, only extracted out the interface.

lock-api/src/main/java/com/palantir/lock/client/CommitTimestampGetter.java
Jolyon-S: You can anyway - if you just implement `close` in `BatchingCommitTimestampGetter` without `throws Exception` it just works.

Jolyon-S: Discussed offline - we do this subtly differently internally, so was thinking of that. This works, or can equally extract a micro-interface that extends `AutoCloseable` but without the `throws Exception` on the close method.

lock-api/src/main/java/com/palantir/lock/client/CoalescingLeaderTimeGetter.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.common.concurrent.CoalescingSupplier;
+import com.palantir.lock.v2.LeaderTime;
+import java.util.function.Supplier;
+
+public final class CoalescingLeaderTimeGetter implements LeaderTimeGetter {
```
jeremyk-91, line 23: non-actionable: This is fine. I'm not super happy with the name just because the other intended implementation is also functionally coalescing, but don't have a better solution.

jeremyk-91 (commented):
🕐 0:09 🏁 0:11

Broad structure seems fine. I think I have some disagreements about `ExceptionHandlingAutoCloseable`: the general principle is fine if you want, but the specific implementation/placement is a bit weird.

lock-api/src/main/java/com/palantir/lock/client/ExceptionHandlingAutoCloseable.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+public interface ExceptionHandlingAutoCloseable extends AutoCloseable {
```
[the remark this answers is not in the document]
sudiksha27: Going to revert this change for now. We can live with overriding close if needed.

Jolyon-S:
Before we merge, would like to discuss some things about the design of CCB in general, so putting `do not merge` for now.

PR #5098, opened 2020-11-04
[LE] Wire in the metrics from client to server

Jolyon-S:
**Goals (and why)**:
We have implemented the calculation and aggregating of leader election time, but not yet wired it so that the client sends to the server.

Tacked the sending of metrics on to the current timelock feedback background task. 

Didn't add any testing, there was none existing.

I think this was what we intended.

`TimeLockFeedbackBackgroundTask`

Want to get this through quickly.

### DIFF
--- a/lock-api-objects/build.gradle
+++ b/lock-api-objects/build.gradle
@@ -27,7 +27,7 @@ recommendedProductDependencies {
     productDependency {
         productGroup = 'com.palantir.timelock'
         productName = 'timelock-server'
-        minimumVersion = '0.218.0'
+        minimumVersion = '0.313.0'
         maximumVersion = '0.x.x'
     }
 }
--- a/lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
@@ -63,7 +63,8 @@ public class LeaderElectionReportingTimelockService implements NamespacedConjure
     private Map<UUID, Instant> leadershipUpperBound = new ConcurrentHashMap<>();
     private Map<UUID, Instant> leadershipLowerBound = new ConcurrentHashMap<>();
 
-    public LeaderElectionReportingTimelockService(
+    @VisibleForTesting
+    LeaderElectionReportingTimelockService(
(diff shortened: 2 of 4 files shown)


lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -63,7 +63,8 @@
     private Map<UUID, Instant> leadershipUpperBound = new ConcurrentHashMap<>();
     private Map<UUID, Instant> leadershipLowerBound = new ConcurrentHashMap<>();
 
-    public LeaderElectionReportingTimelockService(
+    @VisibleForTesting
+    LeaderElectionReportingTimelockService(
```
Jolyon-S, line 67: Changes to this class are just cleanup (adding static, moving methods around).

lock-api-objects/build.gradle
```diff
@@ -27,7 +27,7 @@ recommendedProductDependencies {
     productDependency {
         productGroup = 'com.palantir.timelock'
         productName = 'timelock-server'
-        minimumVersion = '0.218.0'
+        minimumVersion = '0.313.0'
```
jeremyk-91, line 30: This has AtlasDB 0.265.0, which is necessary.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -230,26 +232,30 @@ LeaderElectionStatistics getStatisticsAndSetRegistryTo(TaggedMetricRegistry metr
         return durationToNextLeader(lowerBounds, upperBounds, leaders, secondToLastLongTermLeader);
     }
 
-    private Set<UUID> leadersWithBothBounds(Map<UUID, Instant> lowerBounds, Map<UUID, Instant> upperBounds) {
+    private void clearOldLongTermLeaders(List<UUID> sortedLongTermLeaders) {
+        for (int i = 0; i < sortedLongTermLeaders.size() - 2; i++) {
```
jeremyk-91, line 236: this is a bit trippy, but correct

jeremyk-91 (approved):
🕐  0:08 🏁  0:16

Looks good!

Jolyon-S (commented):
:+1:

PR #5099, opened 2020-11-04
[TS] Allow configuration of lower batch size

gmaretic:
**Goals (and why)**:
With autotuning we attempt to construct batches of up to 100k entries. This is generally what we want, as if there are so many entries we want to catch up as soon as possible. However it can take ~30min to process such an iteration, while sometimes we prefer to be slightly slower but able to be more uniformly up to date.

Mostly wiring

Some tests to verify it works as expected, and that sweep delay treats it sanely

Did I miss something?

Small

asap


jeremyk-91 (commented):
I'm broadly happy with the implementation, though let's get a bit of time tomorrow to discuss some of the tradeoffs/implications.

gmaretic:
On the fence if adding this config is that useful as opposed to just modifying the batch size to a fixed 10k. Given how long it takes to run one iteration of 100k entries, seems the overhead is low enough to do it.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/SweepQueueReader.java
```diff
@@ -50,4 +50,16 @@ SweepBatchWithPartitionInfo getNextBatchToSweep(ShardAndStrategy shardStrategy,
         }
         return accumulator.toSweepBatch();
     }
+
+    @Immutable
+    public interface ReadBatchingRuntimeConfig {
```
jeremyk-91, line 55: minor nits: config objects typically have just values. I'd suggest we either pass `Supplier<RBRC>` and use some `Suppliers.transform(...)` higher up, or rename this `ReadBatchingRuntimeContext` or something like that

jeremyk-91 (approved):
Looks broadly correct, and we spoke externally so I have more context on why this PR is useful. One comment on naming which I found to be a little counterintuitive

gmaretic:
👍

PR #5100, opened 2020-11-05
[LW] Filter out locks using the serverToken, not the fake token in LeasedLockToken

Jolyon-S:
**Goals (and why)**:
* Using the `requestId` from a lock token is not guaranteed to be the right request id - particularly for the `LeasedLockToken` that we actually receive in the `commitWrites` flow. Therefore we need to cast and filter correctly.

* Check if the lock token is the right type, then cast and retrieve the correct request ID.

* Added a test to the integration tests. However, for best coverage we need ETE tests - these are coming in a subsequent PR.

* Just to make sure that this won't break anything.
* Is there a better way to test this without creating a really hacky `LeasedLockTokenCreator` test class?

* `ClientLogEvents`

ASAP

### DIFF
--- a/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
+++ b/atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
@@ -28,9 +28,15 @@ import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.palantir.atlasdb.timelock.api.ConjureLockToken;
 import com.palantir.atlasdb.transaction.api.TransactionLockWatchFailedException;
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -391,6 +398,28 @@ public void newEventsCauseOldEventsToBeDeleted() {
...
+        eventCache.processGetCommitTimestampsUpdate(
+                ImmutableSet.of(ImmutableTransactionUpdate.builder()
+                        .commitTs(5L)
+                        .startTs(START_TS)
+                        .writesToken(commitToken)
+                        .build()),
+                SUCCESS_2);
+
+        assertThat(eventCache.getCommitUpdate(START_TS).accept(new CommitUpdateVisitor()))
+                .containsExactlyInAnyOrder(DESCRIPTOR);
```
Jolyon-S, line 420: this shows that `LOCK_EVENT` is filtered out, which contains `DESCRIPTOR_3`, whereas `LOCK_EVENT_2` is present, with `DESCRIPTOR` present.

gmaretic (approved):
LGTM

PR #5101, opened 2020-11-05
[Cross Client Batching - 1] | TimeLock Batch endpoints

sudiksha27:
**Goals (and why)**:
Implements cross client endpoints for leaderTime and getCommitTimestamps. 
Does not wire the endpoints through TimeLock.

- Adds conjure api definition of cross client batch endpoints for leaderTime and getCommitTimestamps

- Resource implementation for endpoints

- The getCommitTimestamps endpoint does not have additional defense against multiple queries for a namespace in one request.

Added sanity tests

- In the current impl, we fail the entire request if any of the future tasks fail.

- Exception handling

- Code repetition

Small

Today 🏃‍♀️ 🚀 🧨


timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
```diff
@@ -0,0 +1,63 @@
...
+      external:
+        java: java.lang.Long
+    LockWatchStateUpdate:
+      base-type: any
+      external:
+        java: com.palantir.lock.watch.LockWatchStateUpdate
+    ConjureIdentifiedVersion:
+      base-type: any
+      external:
+        java: com.palantir.atlasdb.timelock.api.ConjureIdentifiedVersion
```
Jolyon-S, line 18: I don't know what opinions are on just adding these endpoints / requests to the current timelock api, or creating a new one (as you have done here) - but either way, I think it's really weird to import in definitions from the other API.

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
```diff
@@ -0,0 +1,63 @@
...
+    name: Cross Client Batched Timelock Service
+    default-auth: header
+    package: com.palantir.atlasdb.timelock.batch.api
+    base-path: /tl/batch
+    endpoints:
+      leaderTimes:
+        http: POST /lts
+        args:
+          namespaces: list<string>
+        returns: list<NamespacedLeaderTime>
```
Jolyon-S, line 54: I'm not convinced this is the right approach. I'd have thought you would want a request type like `NamespacedLeaderTimes`, then call with a list of those. The point of that would be if you had incoming requests for leader time such as:

namespace1
namespace2
namespace3
namespace1

say they all get batched together - you'd likely want to further gather together all requests of the same namespace into their own requests, so that you can do clever things (coalesce / reduce overheads etc). Maybe I'm missing something - I've just looked at this briefly - but that seems like the right thing to do.

Jolyon-S (commented):
Something feels a bit off, but it's hard to put a finger on it exactly. Happy to chat if that'd be helpful

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
sudiksha27: The server should be defensive against multiple queries for same namespace.

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
jeremyk-91: I'd prefer we just use the existing API file. In my mind the model was that they would be placed on a different _service_, but in the same API.

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
jeremyk-91: This is LeaderTime, though, which is a _coalescing_ endpoint, and so for this one Sudiksha's API is correct (notice that `namespaces` is a list of strings, and any futures for those namespaces will be answered with the correct answer from the list of `NamespacedLeaderTime`s).

sudiksha27: For leaderTime, I think having a set of namespaces should be okay. I am unsure if we need more than one computation for any client (even if they are making more than one query).

Resolving collision for NamespacedGetCommitTimestampsRequest might be trickier because of LockWatchVersion but, I can also not see the case where we manage to make two queries in one request with different lockWatchVersion for same client.

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
```diff
@@ -0,0 +1,63 @@
...
+        http: POST /lts
+        args:
+          namespaces: set<string>
+        returns: list<NamespacedLeaderTime>
+        docs: |
+          Cross client batched version of ConjureTimelockService#leaderTime endpoint for acquiring leaderTimes.
+      getCommitTimestamps:
+        http: POST /gcts
+        args:
+          requests: list<NamespacedGetCommitTimestampsRequest>
```
jeremyk-91, line 60: This one on the other hand is a bit messier - there's some nontrivial logic in how to reduce all of the legacy `GetCommitTimestampRequest`s into a single one (basically you can take the oldest easily, maybe the newest if there's some clever hackery on the client side).

jeremyk-91 (commented):
🕐 0:18 🏁 1:55

Only reviewed the conjure API

timelock-api/src/main/conjure/cross-client-batched-timelock-api.yml
Jolyon-S: I can:
Imagine we have one service using atlasdb-proxy that has multiple nodes. They might all talk to the same node of atlasdb-proxy but with different internal lockWatchVersions, and would expect different responses (and yet would all be the same namespace in Atlas).

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -181,4 +193,24 @@ services:
         http: POST /sw/{namespace}
         args:
           namespace: string
-          request: LockWatchRequest
\ No newline at end of file
+          request: LockWatchRequest
+  CrossClientBatchedConjureTimelockService:
```
jeremyk-91, line 197: missed on pass 1: I don't think CrossClientBatched should be a part of the service name as it's more of an implementation detail, it should really be something like MultiClientConjureTimelockService?

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -181,4 +193,24 @@ services:
...
+    package: com.palantir.atlasdb.timelock.batch.api
+    base-path: /tl/batch
+    endpoints:
+      leaderTimes:
+        http: POST /lts
+        args:
+          namespaces: set<string>
+        returns: list<NamespacedLeaderTime>
+        docs: |
+          Cross client batched version of ConjureTimelockService#leaderTime endpoint for acquiring leaderTimes.
```
jeremyk-91, line 209: as above, the notion of batching should not exist at this level.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -181,4 +193,24 @@ services:
         http: POST /sw/{namespace}
         args:
           namespace: string
-          request: LockWatchRequest
\ No newline at end of file
+          request: LockWatchRequest
+  CrossClientBatchedConjureTimelockService:
+    name: Cross Client Batched Timelock Service
+    default-auth: header
+    package: com.palantir.atlasdb.timelock.batch.api
```
jeremyk-91, line 200: as above: don't think batching is correct. multiclient maybe - or can this just go in the same package?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/batch/CrossClientBatchedConjureTimeLockResource.java
```diff
@@ -0,0 +1,106 @@
...
+                commitTimestampsResponse -> NamespacedGetCommitTimestampsResponse.builder()
+                        .namespace(request.getNamespace())
+                        .inclusiveLower(commitTimestampsResponse.getInclusiveLower())
+                        .inclusiveUpper(commitTimestampsResponse.getInclusiveUpper())
+                        .lockWatchUpdate(commitTimestampsResponse.getLockWatchUpdate())
+                        .build(),
+                MoreExecutors.directExecutor());
+    }
+
+    private AsyncTimelockService forNamespace(String namespace) {
```
jeremyk-91, line 95: can we name this `getServiceForNamespace` or something like that?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/batch/CrossClientBatchedConjureTimeLockResource.java
```diff
@@ -0,0 +1,106 @@
...
+                        request.getLastKnownVersion().map(this::toIdentifiedVersion));
+        return Futures.transform(
+                commitTimestamps,
+                commitTimestampsResponse -> NamespacedGetCommitTimestampsResponse.builder()
+                        .namespace(request.getNamespace())
+                        .inclusiveLower(commitTimestampsResponse.getInclusiveLower())
+                        .inclusiveUpper(commitTimestampsResponse.getInclusiveUpper())
+                        .lockWatchUpdate(commitTimestampsResponse.getLockWatchUpdate())
+                        .build(),
+                MoreExecutors.directExecutor());
```
jeremyk-91, line 92: I think this is safe and correct; can you explain why this is the case?

jeremyk-91 (commented):
🕐 0:16 🏁 3:45 (sigh)

I missed some things in part 1. I think the execution flow in part 2 is correct; most comments here are stylistic.

jeremyk-91:
You probably haven't done it yet because it's a draft: but a reminder that we need tests before merge.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -181,4 +193,24 @@ services:
...
           namespace: string
-          request: LockWatchRequest
\ No newline at end of file
+          request: LockWatchRequest
+  MultiClientConjureTimelockService:
+    name: Multi Client Timelock Service
+    default-auth: header
+    package: com.palantir.atlasdb.timelock.api
+    base-path: /tl/batch
+    endpoints:
```
sudiksha27, line 202: change base path

jeremyk-91 (approved):
🕐 0:07, 🏁 ???

Seems fine. An integration test would be nice, though hard at this point (because you don't have the Jersey resource wired yet). I'd suggest you do that next

PR #5102, opened 2020-11-05
[LW] ETE tests

Jolyon-S:
**Goals (and why)**:
* Test lock watches ETE. This has already discovered one bug!

* Implement using the same pattern as other Atlas ETEs.

* This is a testing PR.

* There are some annoying constraints - I'd like to cause a leader election but I don't think that is possible.

* `LockWatchEteTest`

Probably within a week, but not urgent.


atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/lock/EteLockWatchResource.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("lock-watch")
+public interface EteLockWatchResource {
+    @POST
+    @Path("start-transaction")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
```
Jolyon-S, line 32: I know I don't need so many annotations - will remove when I want to get this in.

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ExposedLockWatchManager.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.palantir.atlasdb.keyvalue.api.watch;
+
+import com.palantir.lock.watch.CommitUpdate;
+import com.palantir.lock.watch.LockWatchVersion;
+import com.palantir.lock.watch.TransactionsLockWatchUpdate;
+import java.util.Optional;
+import java.util.Set;
+
+public class ExposedLockWatchManager {
```
Jolyon-S, line 25: This class is necessary as we have hidden some of the methods on `LockWatchManager` to stop the average consumer from using it.

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/lock/SimpleEteLockWatchResource.java
```diff
@@ -0,0 +1,160 @@
...
+    public LockWatchVersion getVersion(TransactionId transactionId) {
+        return lockWatchManager
+                .getUpdateForTransactions(ImmutableSet.of(transactionId.startTs()), Optional.empty())
+                .startTsToSequence()
+                .get(transactionId.startTs());
+    }
+
+    @Override
+    public void setTable(String tableName) {
+        table = tableName.replace("\"", "");
```
Jolyon-S, line 96: I don't know why, but I kept getting issues where the input table name was `"blah"` instead of `blah`.

atlasdb-ete-tests/src/main/java/com/palantir/atlasdb/lock/SimpleEteLockWatchResource.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        CommitUpdate getCommitUpdate() {
+            return commitUpdate.get();
+        }
+
+        @Override
+        public void throwIfConditionInvalid(long timestamp) {
+            if (startTs.filter(ts -> ts != timestamp).isPresent()) {
+                commitUpdate.set(lockWatchManager.getCommitUpdate(startTs.get()));
```
Jolyon-S, line 147: This is necessary as we only want the commit update when we are passed the commit timestamp, not the start.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/LockWatchEteTest.java
```diff
@@ -0,0 +1,265 @@
...
+    private static final String ROW_2 = row(2);
+    private static final String ROW_3 = row(3);
+
+    private final EteLockWatchResource lockWatcher = EteSetup.createClientToSingleNode(EteLockWatchResource.class);
+
+    private TableReference tableReference;
+
+    @Before
+    public void before() {
+        String tableName = UUID.randomUUID().toString().substring(0, 16).replace("-", "X");
```
jeremyk-91, line 61: nit: standard in other tests is to replace with `_`, not `X`.
```suggestion
        String tableName = UUID.randomUUID().toString().substring(0, 16).replace("-", "_");
```

jeremyk-91 (commented):
🕐  0:35 🏁  too long

It's great we have some tests! Most of the comments here are on minor semantic details or stylistic.

Probably not in this PR: I think there are a couple of more things we should add to lock-watch testing in general (in particular, I'd probably validate the ordering of the events too, and I don't know if there's an easy way to validate watch registration events). But this is a massive improvement over what we previously had

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/TimelockNamespaces.java
```diff
@@ -42,8 +42,10 @@
     @VisibleForTesting
     static final String MAX_CLIENTS = "maxClients";
 
-    private static final Predicate<String> isValidName =
-            Pattern.compile("^(?!tl$)[a-zA-Z0-9_-]+$").asPredicate();
+    @VisibleForTesting
+    static final Predicate<String> isValidName =
+            Pattern.compile("^(?!((tl|lw)$))[a-zA-Z0-9_-]+$").asPredicate();
```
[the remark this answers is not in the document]
Jolyon-S: Fixed in this one

Jolyon-S:
If we can force leader elections then I think it'd be much more straightforward to be able to test watch registrations, and it would also allow us to test a lot of the tricky behaviour around snapshotting. However, until then, I think this covers most of what we can do.

jeremyk-91 (approved):
👍

PR #5105, opened 2020-11-06
add cache warming option to ssm

berler:
Adds a CacheWarming option that can control how the SweepStrategyManager
cache is warmed on initialization. The possible options are FULL and NONE.
FULL remains the default (so default behavior is not changing). NONE skips
warming the cache.

**Goals (and why)**:
The existing SweepStrategyManager cache warming code causes problems for internal products when there is a very large number of tables.

Adds a CacheWarming enum that can control how the SweepStrategyManager warms the cache on init.

Parameterized the postgres sweep integration tests to test all possible values for CacheWarming.

**Priority (whenever / two weeks / yesterday)**: very high -- current open support issues internally

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -525,6 +525,14 @@ public interface KeyValueService extends AutoCloseable, AsyncKeyValueService {
     Set<TableReference> getAllTableNames();
 
     /**
+     * Returns a list of up to {@code maxResults} tables stored in this key value service.
+     *
+     * This should be preferred over {@link #getAllTableNames()} if possible because the number of tables may be
+     * unbounded.
+     */
+    Iterable<TableReference> getLimitedTableNames(int maxResults);
+
+    /**
      * Gets the metadata for a given table. Also useful for checking to see if a table exists.
      *
      * @return a byte array representing the metadata for the table. Array is empty if no table
@@ -542,6 +550,14 @@ public interface KeyValueService extends AutoCloseable, AsyncKeyValueService {
      */
     @Idempotent
     Map<TableReference, byte[]> getMetadataForTables();
+
+    /**
+     * Gets the metadata for up to {@code maxResults} tables.
+     *
+     * This should be preferred over {@link #getMetadataForTables()} if possible because the number of tables may be
+     * unbounded.
+     */
+    Map<TableReference, byte[]> getLimitedMetadataForTables(int maxResults);
 
     @Idempotent
     void putMetadataForTable(TableReference tableRef, byte[] metadata);
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
@@ -525,14 +525,6 @@ public interface KeyValueService extends AutoCloseable, AsyncKeyValueService {
     Set<TableReference> getAllTableNames();
 
     /**
-     * Returns a list of up to {@code maxResults} tables stored in this key value service.
-     *
-     * This should be preferred over {@link #getAllTableNames()} if possible because the number of tables may be
-     * unbounded.
-     */
-    Iterable<TableReference> getLimitedTableNames(int maxResults);
-
-    /**
      * Gets the metadata for a given table. Also useful for checking to see if a table exists.
      *
      * @return a byte array representing the metadata for the table. Array is empty if no table
@@ -550,14 +542,6 @@ public interface KeyValueService extends AutoCloseable, AsyncKeyValueService {
      */
     @Idempotent
     Map<TableReference, byte[]> getMetadataForTables();
-
-    /**
-     * Gets the metadata for up to {@code maxResults} tables.
-     *
-     * This should be preferred over {@link #getMetadataForTables()} if possible because the number of tables may be
-     * unbounded.
-     */
-    Map<TableReference, byte[]> getLimitedMetadataForTables(int maxResults);
 
     @Idempotent
     void putMetadataForTable(TableReference tableRef, byte[] metadata);
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
@@ -1481,16 +1481,6 @@ public class CassandraKeyValueServiceImpl extends AbstractKeyValueService implem
                 .collect(Collectors.toSet());
     }
 
-    @Override
-    public Iterable<TableReference> getLimitedTableNames(int maxResults) {
-        // Note: This is not optimal since getTableReferencesWithoutFiltering() loads *all* names, even though it
-        // returns a Stream. Getting a partial list of table names might not be possible with cassandra thrift.
-        return cassandraTables
-                .getTableReferencesWithoutFiltering()
-                .filter(tr -> !HiddenTables.isHidden(tr))
-                .limit(maxResults)::iterator;
-    }
-
     /**
      * Gets the metadata for a given table. Also useful for checking to see if a table exists. Requires a quorum of
      * Cassandra nodes to be reachable.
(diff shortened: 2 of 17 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -45,9 +56,8 @@ public static SweepStrategyManager createDefault(KeyValueService kvs) {
                             .expireAfterAccess(1, TimeUnit.DAYS)
```
schlosna, line 56: do we want a max size on this cache to limit memory footprint?

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -84,9 +94,18 @@ public static SweepStrategyManager completelyConservative() {
...
-        return ImmutableMap.copyOf(
-                Maps.transformValues(kvs.getMetadataForTables(), SweepStrategyManagers::getSweepStrategy));
+    private static Map<TableReference, SweepStrategy> getSweepStrategiesForWarmingCache(
+            KeyValueService kvs, CacheWarming cacheWarming) {
+        switch (cacheWarming) {
+            case FULL:
+                return Maps.transformValues(kvs.getMetadataForTables(), SweepStrategyManagers::getSweepStrategy);
+            case LIMITED:
+                return Maps.transformValues(
+                        kvs.getLimitedMetadataForTables(50), SweepStrategyManagers::getSweepStrategy);
```
schlosna, line 104: How much value is there in having a separate `LIMITED` mode between `FULL` and `NONE`? If we set to `NONE` and just let the cache warm up on its own based on requests, what behavior would we see? How did we pick `50`? Are the first `50` tables actually the ones we want cached?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -524,6 +524,14 @@ void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete
     @Idempotent
     Set<TableReference> getAllTableNames();
 
+    /**
+     * Returns a list of up to {@code maxResults} tables stored in this key value service.
+     *
+     * This should be preferred over {@link #getAllTableNames()} if possible because the number of tables may be
+     * unbounded.
+     */
+    Iterable<TableReference> getLimitedTableNames(int maxResults);
```
schlosna, line 533: Should this be a paged result? What is the epoch for concurrent consistency (e.g. tables created while this is executing)? What happens if one specifies various values for `maxResults`, say `0`, `-1`, `Integer.MAX_VALUE`? I'm a little skeptical we even need these methods, but wondering if we actually do should instead consider something like `Stream<TableReference> getTableNames()` and let consumers filter/limit/transform/collect as desired.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
```diff
@@ -543,6 +551,14 @@ void deleteAllTimestamps(TableReference tableRef, Map<Cell, TimestampRangeDelete
     @Idempotent
     Map<TableReference, byte[]> getMetadataForTables();
 
+    /**
+     * Gets the metadata for up to {@code maxResults} tables.
+     *
+     * This should be preferred over {@link #getMetadataForTables()} if possible because the number of tables may be
+     * unbounded.
+     */
+    Map<TableReference, byte[]> getLimitedMetadataForTables(int maxResults);
```
schlosna, line 560: same comment as above

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceImpl.java
```diff
@@ -1481,6 +1481,16 @@ public void createTables(final Map<TableReference, byte[]> tablesToMetadata) {
                 .collect(Collectors.toSet());
     }
 
+    @Override
+    public Iterable<TableReference> getLimitedTableNames(int maxResults) {
+        // Note: This is not optimal since getTableReferencesWithoutFiltering() loads *all* names, even though it
+        // returns a Stream. Getting a partial list of table names might not be possible with cassandra thrift.
+        return cassandraTables
+                .getTableReferencesWithoutFiltering()
+                .filter(tr -> !HiddenTables.isHidden(tr))
```
schlosna, line 1490: Similar to above, should these lines (same as in `getAllTableNames()`)  be pulled out to a separate method that returns `Stream<TableReference>`?

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/KeyValueService.java
berler: I was thinking about that, but it is significantly more complicated to implement correctly. Additionally the caller would need to call close() on the Stream, which is not something that is usually done in general (calling .collect() etc doesn't do it), so it would be really easy for people to leak resources if we made this a Stream.

CloseableIterator or something similar might make more sense, but it would also be more complicated to implement.

In general I think we should be discouraging anything that would try to enumerate all tables in the system. Clients can only interact with tables if they know about some Schema, and clients should be using the Schema to know what tables it needs to care about.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
berler: I don't think so, considering the default is to pre-warm the cache with all tables. The intended use case is that there won't be very many tables. If we had a max size here it seems like you could get very bad sweep perf once you have more than that amount of tables.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
berler: 50 was recommended I believe by @jeremyk-91 when this was discussed a while ago internally

gmaretic (commented):
Largely looks good. I'm fine with getting this in, but should probably add a limit to the dbkvs query first.

atlasdb-dbkvs/src/main/java/com/palantir/atlasdb/keyvalue/dbkvs/impl/DbKvs.java
```diff
@@ -1137,6 +1138,26 @@ public void createTable(TableReference tableRef, byte[] tableMetadata) {
...
     }
 
+    @Override
+    public Iterable<TableReference> getLimitedTableNames(int maxResults) {
+        if (maxResults <= 0) {
+            return Collections.emptyList();
+        }
+        return run(conn -> {
+            try (AgnosticLightResultSet rows = conn.selectLightResultSetUnregisteredQuery(
+                    "SELECT table_name FROM " + config.metadataTable().getQualifiedName())) {
```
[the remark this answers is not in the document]
berler: The limit syntax is completely different in postgres and oracle. We cannot just add a limit clause here unless we do something completely different for postgres and oracle.

From a perf standpoint there shouldn't be very much difference between using a limit and only iterating through the ResultSet part way unless the configured prefetch is huge -- the AgnosticIterator impl in AgnosticLightResultSetImpl always sets the initial prefetch to 10, and doubles it (up to a max) after each batch is fetched.

atlasdb-client/src/main/java/com/palantir/atlasdb/transaction/impl/SweepStrategyManagers.java
```diff
@@ -45,9 +56,8 @@ public static SweepStrategyManager createDefault(KeyValueService kvs) {
                             .expireAfterAccess(1, TimeUnit.DAYS)
                             .build(tableRef -> getSweepStrategy(kvs.getMetadataForTable(tableRef)));
 
-                    // On async initialization, add all tables to optimize for cases when using mostly non-dynamic
-                    // tables.
-                    cache.putAll(getSweepStrategies(kvs));
+                    // On async initialization, add a possibly limited number of tables to the cache.
```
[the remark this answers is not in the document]
berler: things do init async -- this happens the first time it attempts to be used because this is inside a RecomputingSupplier

gmaretic (approved):
Thanks for clarifying, looks good!

PR #5106, opened 2020-11-09
[LW] Check the correct client versions range

Jolyon-S:
**Goals (and why)**:
* Turns out that we were checking the wrong version range for validation. We should check that the events returned enclose (userVersion, latestTimestampVersion], but we were actually checking [earliestTimestampVersion, latestTimestampVersion].

* Fixed the bug as listed above.

* Added yet another test. This was caught by the ETE test PR, so that will also help.

* Is this finally right?

* `ClientLogEvents`

This week.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
@@ -54,8 +54,9 @@ interface ClientLogEvents {
     default TransactionsLockWatchUpdate toTransactionsLockWatchUpdate(
             TimestampMapping timestampMapping, Optional<LockWatchVersion> lastKnownVersion) {
         /*
-         Case 1: client is behind earliest transaction. Therefore we want to ensure that there are events from
-                 after the client version up to the latest transaction version.
+         Case 1: client is behind earliest transaction. Therefore we want to ensure that there are events present for
+                 each version starting with the client version (exclusive) and ending with latest transaction version
+                 (inclusive).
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
```diff
@@ -73,28 +79,23 @@ default CommitUpdate toCommitUpdate(LockWatchVersion startVersion, CommitInfo co
...
         }
 
-        Range<Long> rangeToTest;
-        if (offsetStartVersion) {
-            rangeToTest = Range.closed(versionRange.lowerEndpoint() + 1, versionRange.upperEndpoint());
-        } else {
-            rangeToTest = versionRange;
-        }
-
+        Range<Long> rangeToTest = Range.closed(lowerBound, upperBound);
```
gmaretic, line 98: `Range.openClosed` so you don't need to do +1 on the lower bound everywhere

gmaretic: Though I guess you need a -1 :D

gmaretic (approved):
A couple nits, but otherwise looks good.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
Jolyon-S: Yeah nope, because there is one case where you need something different - probably easier at this point to just add the one.

PR #5108, opened 2020-11-09
Fix MaskedValue Jackson 2.12 compatibility

carterkozak:
**Goals (and why)**:

Previously we relied on Jackson magically deserializing a string
value into a MaskedValue. This no longer works in Jackson 2.12
snapshots, and I'm not convinced it ever should have worked. Other
proejcts represent MaskedValue as a string in configuration, so we
must add explicit handling and test coverage.

Explicit `@JsonCreator` annotations.


jeremyk-91 (approved):
🕐0:06, 🏁0:10

👍

PR #5112, opened 2020-11-12
[Tl Corruption Checks] | Bump down the limit for rows verified per iteration

sudiksha27:
**Goals (and why)**:
Bump down the limit on number of rows to be read per iteration for verification. The high number of rows read per iteration for verification leads to performance degradation in TimeLock.
The rate at which rows are read for corruption checks is still higher than the max rate at which rows are written on internal stack.


jeremyk-91 (approved):
🕐 0:03 🏁 0:03

Discussed offline - @sudiksha27 ran perf experiments, and we confirmed also that this rate is enough for us to not fall behind on powerful internal stack's shopping product.

PR #5113, opened 2020-11-13
Better metrics

gmaretic:
**Goals (and why)**:
Current histogram is not very useful because it gives us the worst estimates for leader duration of all the clients. We know that the actual time leader election took is in fact less than the minimum for all the clients, so we now report minimum instead. Added median for good measure, to give us an idea of how accurate the estimate is (the closer the median to the min, the more accurate the estimate should be).


Jolyon-S (approved):
Are there any tests that need updating? Otherwise, LGTM.

PR #5114, opened 2020-11-13
remove moreutils dependency in build

berler:
**Goals (and why)**:
Make builds more reliable, and remove one of the blockers for re-running much older builds (to allow easier backporting of fixes). It is not sustainable to require regularly updating a url just to keep builds working.

Remove the moreutils dependency completely from builds.

Note that I tried using`sudo apt-get update && sudo apt-get install moreutils` to install moreutils instead (which should be reliable), except that seemed to install a version that is not compatible (it did not have support for the options being used).

If people really worry about the value of timestamps in tests, we should configure slf4j (or whatever we use) logging in tests to output timestamps. I would prefer if this is done in a separate change so that if needed removing the moreutils dependency can be done easily in older versions.

**Priority (whenever / two weeks / yesterday)**: High if reliable builds are important.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,7 @@ jobs:
           key: gradle-cache-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}
 
       - run: git tag -l | xargs git tag -d && git fetch -t # ensure all tags are fetched and up-to-date
-      - run: ./scripts/circle-ci/download-moreutils.sh
-      - run: ./scripts/time-cmd.sh ./gradlew --profile --parallel --stacktrace classes testClasses
+      - run: ./gradlew --profile --parallel --stacktrace classes testClasses
 
       - save_cache:
           key: gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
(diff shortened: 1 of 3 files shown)


jeremyk-91 (commented):
It's worth noting that configuring a test logger in Java as an alternative wouldn't be sufficient, because this wraps more than the log output of tests: it also adds timestamps to the output of commands etc. (so it can identify things like "I took two minutes to talk to Gradle").

I'm fine with this, I don't think anyone has used the timestamps for debugging. I'll check with the team, and will approve this PR by EOD tomorrow if there are no further comments.

An alternative would probably be an implementation of `ts` that maybe wraps the output of a shell with `date` in some way (which I assume is supported without moreutils), but I don't really think the timestamps are valuable here.

berler:
> It's worth noting that configuring a test logger in Java as an alternative wouldn't be sufficient, because this wraps more than the log output of tests: it also adds timestamps to the output of commands etc. (so it can identify things like "I took two minutes to talk to Gradle").
I'd be pretty suspicious if there was any value in this part of the logging. I also highly doubt anyone is actually looking at these logs.

Regarding getting output of how long gradle commands take, gradle supports this with publishing build scans. You should use that instead if you're trying to debug how long gradle tasks take.

jeremyk-91 (approved):
Sure, I think the point was that this includes timestamps for every line of output (including whatever else the scripts do, which I don't know offhand) - as you point out, there are other mechanisms that would work in most cases, but I don't know what needs to be done offhand.

I don't think anyone of the team has objected to this, so we can go ahead.

PR #5115, opened 2020-11-16
[Cross Client Batching - 1b] | Wiring TimeLock Batch endpoints

sudiksha27:
**Goals (and why)**:
Wiring the multi client endpoints through TimeLockAgent

Added integration sanity test

MultiClientConjureTimelockResource.java

Would be good to do before tomorrow EOD

### DIFF
--- a/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
+++ b/timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
@@ -31,6 +31,8 @@ import com.palantir.atlasdb.timelock.api.ConjureLockRequest;
 import com.palantir.atlasdb.timelock.api.ConjureLockResponse;
 import com.palantir.atlasdb.timelock.api.ConjureLockToken;
 import com.palantir.atlasdb.timelock.api.ConjureUnlockRequest;
+import com.palantir.atlasdb.timelock.api.GetCommitTimestampsRequest;
+import com.palantir.atlasdb.timelock.api.GetCommitTimestampsResponse;
 import com.palantir.atlasdb.timelock.api.MultiClientConjureTimelockService;
 import com.palantir.atlasdb.timelock.api.NamespacedGetCommitTimestampsRequest;
 import com.palantir.atlasdb.timelock.api.NamespacedGetCommitTimestampsResponse;
@@ -58,6 +60,7 @@ import com.palantir.lock.v2.LockResponse;
 import com.palantir.lock.v2.LockToken;
 import com.palantir.lock.v2.WaitForLocksRequest;
 import com.palantir.lock.v2.WaitForLocksResponse;
+import com.palantir.lock.watch.LockWatchStateUpdate;
 import com.palantir.tokens.auth.AuthHeader;
 import java.lang.management.ManagementFactory;
 import java.time.Duration;
@@ -433,7 +436,8 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
             for (int i = 0; i < 10_000; i++) { // Needed as it takes a while for the thread buildup to occur
                 client.getFreshTimestamp();
                 assertNumberOfThreadsReasonable(
-                        startingNumThreads, ManagementFactory.getThreadMXBean().getThreadCount(), isNonLeaderTakenOut);
+                        startingNumThreads, ManagementFactory.getThreadMXBean().getThreadCount(),
+    isNonLeaderTakenOut);
                 if (i == 1_000) {
                     makeServerWaitTwoSecondsAndThenReturn503s(nonLeader);
                     isNonLeaderTakenOut = true;
@@ -477,7 +481,8 @@ public class MultiNodePaxosTimeLockServerIntegrationTest {
         try {
             for (int i = 0; i < 1_800; i++) { // Needed as it takes a while for the thread buildup to occur
                 assertNumberOfThreadsReasonable(
-                        startingNumThreads, ManagementFactory.getThreadMXBean().getThreadCount(), isNonLeaderTakenOut);
+                        startingNumThreads, ManagementFactory.getThreadMXBean().getThreadCount(),
+    isNonLeaderTakenOut);
                 cluster.currentLeaderFor(client.namespace())
                         .timeLockManagementService()
                         .achieveConsensus(AuthHeader.valueOf("Bearer pqrstuv"), ImmutableSet.of(client.namespace()));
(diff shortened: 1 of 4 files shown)


timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -484,6 +487,18 @@ public void stressTestForPaxosEndpoints() {
...
+    @Test
+    public void leaderRespondsToMultiClientRequests() {
+        MultiClientConjureTimelockService multiClientConjureTimelockService =
+                cluster.currentLeaderFor(client.namespace()).multiClientService();
+        Set<String> expectedNamespaces = ImmutableSet.of("alpha", "beta");
+        List<NamespacedLeaderTime> leaderTimes =
+                multiClientConjureTimelockService.leaderTimes(AuthHeader.valueOf("Bearer omitted"), expectedNamespaces);
+        Set<String> namespaces =
+                leaderTimes.stream().map(NamespacedLeaderTime::getNamespace).collect(Collectors.toSet());
+        assertThat(namespaces).hasSameElementsAs(expectedNamespaces);
```
Jolyon-S, line 499: I don't know leader time (and/or the internals of Timelock) all that well - would leader times from different namespaces have different leadership IDs, or the same? If you expect one or the other, can you verify that in this test?

Jolyon-S (commented):
Just one comment on the test. I haven't been keeping up with the CCB stuff, so while it looks fine, would defer to @jeremyk-91 for the approval.

jeremyk-91:
Looks fine, but would be good to add the test that @Jolyon-S suggested. It may be a bit early, but could be worth taking a stab at adding a sanity check for the getCommitTimestamp one.

jeremyk-91:
guess this counts, so 🕐 0:05 🏁 0:39

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
sudiksha27: Leadership ids should be different as each client will have its own instance of AsyncTimelockService -> AsyncLockService -> LeaderClock. I've added the check now.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -484,6 +491,37 @@ public void stressTestForPaxosEndpoints() {
...
+    public void leaderCanProcessMultiClientGetCommitTimestampRequest() {
+        MultiClientConjureTimelockService service =
+                cluster.currentLeaderFor(client.namespace()).multiClientService();
+        Set<String> expectedNamespaces = ImmutableSet.of("alpha", "beta");
+        List<NamespacedGetCommitTimestampsResponse> commitTimestamps =
+                service.getCommitTimestamps(AUTH_HEADER, getGetCommitTimestampsRequests(expectedNamespaces));
+        Set<String> namespaces = commitTimestamps.stream()
+                .map(NamespacedGetCommitTimestampsResponse::getNamespace)
+                .collect(Collectors.toSet());
+        assertThat(namespaces).hasSameElementsAs(expectedNamespaces);
```
Jolyon-S, line 522: This is perhaps being a bit pedantic, but get commit timestamps have a leader ID hidden inside (it's the `logId` inside `LockWatchStateUpdate`). Perhaps worth a similar check to the one above?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -484,6 +491,37 @@ public void stressTestForPaxosEndpoints() {
...
+                multiClientConjureTimelockService.leaderTimes(AUTH_HEADER, expectedNamespaces);
+        Set<String> namespaces =
+                leaderTimes.stream().map(NamespacedLeaderTime::getNamespace).collect(Collectors.toSet());
+        assertThat(namespaces).hasSameElementsAs(expectedNamespaces);
+
+        Set<LeadershipId> leadershipIds = leaderTimes.stream()
+                .map(NamespacedLeaderTime::getLeaderTime)
+                .map(LeaderTime::id)
+                .collect(Collectors.toSet());
+        assertThat(leadershipIds).hasSameSizeAs(expectedNamespaces);
```
jeremyk-91, line 509: maybe a bit overkill: if we run these against the regular conjuretimelockservice.getLeaderTime we get the same answer right?

jeremyk-91 (commented):
Broadly looks good. Afraid I'd like one more test, though I wouldn't mind that being in a separate PR if it's hard to write

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
sudiksha27: Same leader will serve the request, but leaderTime will be different (time elapsed b/w two requests). I will add this test for commitTs request as well, where I can add stricter assertions on the response.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
sudiksha27: Done

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
jeremyk-91: Yeah, makes sense. I think I intended an ID comparison, which you've added below.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -514,12 +520,70 @@ public void leaderCanProcessMultiClientGetCommitTimestampRequest() {
...
+                    .namespacedConjureTimelockService()
+                    .getCommitTimestamps(defaultCommitTimestampRequest());
+            assertThat(namespacedGetCommitTimestampsResponse
+                            .getLockWatchUpdate()
+                            .logId())
+                    .isEqualTo(conjureTimelockServiceGetCommitTimestampResponse
+                            .getLockWatchUpdate()
+                            .logId());
+            assertThat(namespacedGetCommitTimestampsResponse.getInclusiveUpper())
+                    .isEqualTo(conjureTimelockServiceGetCommitTimestampResponse.getInclusiveLower() - 1);
```
jeremyk-91, line 583: nit/suggestion: Usually these assertions are `assertThat(actual).isEqualTo(expected)`, so I'd probably flip these two. Also, you can use `as(String)` to say in English what the assertion is doing, which I think is useful for this later one, which verifies `timestamps are contiguously increasing per namespace if there are no elections`.

jeremyk-91 (approved):
Looks good - one nit left on my end.

PR #5117, opened 2020-11-17
LockServiceImpl should ThreadInterrupt all Blocked threads

nartmal:
**Goals (and why)**:

I'm trying to figure out why leader election changes tend to line up to ~40 seconds of no progress on an internal product that loves to make WRITE_LOCK LockRequests with `TxManager#runTaskWithLocksWithRetry`

I believe that there are still requests that are parked while trying to `tryAcquireNanos` on the old node which sit there and causes things to fail slowly. I would like recovery from these situations to be a lot quicker to drive down our P99's.

Instead of tracking just `indefinitelyBlockingThreads`, track all threads that are blocking. Upon `close()` which I think also executes on Leadership changes per `AwaitingLeadershipProxy`, then we should interrupt **all** blocking threads.

~This should trigger `LockServerSync#tryAcquireNanos` to return `false` quickly and subsequently `return new LockResponse(null, failedLocks);` quickly back to the client when it's `LOCK_ALL_OR_NONE` or if all locks failed (what I want)~. The existing behavior is LockServiceImpl throws InterruptedException for callers to handle. Of which it looks like `AwaitingLeadershipProxy#handleDelegateThrewException` will re-throw as `NotCurrentLeaderException`

I couldn't find any tests for the behavior of `close()`, mind pointing them to me if I missed them?

I read the code wrong and this doesn't do what I think it does.

LockServiceImpl.java

Whenever but would appreciate feedback sooner rather than later.

### DIFF
--- a/changelog/@unreleased/pr-5117.v2.yml
+++ b/changelog/@unreleased/pr-5117.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: LockServiceImpl now interrupts all blocked threads upon closing.
+  links:
+  - https://github.com/palantir/atlasdb/pull/5117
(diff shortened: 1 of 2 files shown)


gmaretic (approved):
Looks reasonable, don't really have any concerns about this

PR #5118, opened 2020-11-17
[Cross client Batching - 1c] |  Implement multi-client startTransactions endpoint

sudiksha27:
**Goals (and why)**:
Implement multi-client startTransactions endpoint

- Adds conjure api definition of cross client batch endpoint for startTransactions

- The endpoints returns successfully if startTransactions futures for **_all_** queries are successful

- There is additional defense mechanism against more than one requestors querying for same namespace.

Added unit and integration tests

1. Failure mechanism - fails the request for all queries without retry in case there is sanity check failure
2. Should the sanity check be moved to the client side? 
3. Should optimize the sanity check to avoid multiple iterations on input query list?

timelock-api.yml, MultiClientConjureTimelockResource.java

Today would be great

### DIFF
--- a/timelock-api/src/main/conjure/timelock-api.yml
+++ b/timelock-api/src/main/conjure/timelock-api.yml
@@ -125,6 +125,20 @@ types:
           inclusiveLower: Long
           inclusiveUpper: Long
           lockWatchUpdate: LockWatchStateUpdate
+      NamespacedStartTransactionsRequest:
+        fields:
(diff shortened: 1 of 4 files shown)


jeremyk-91 (commented):
I think we should use Maps for the API: the List of request/response objects was a special case for getCommitTimestamps, where there is no obvious per-namespace way of reducing requests.

PR #5120, opened 2020-11-19
[Cross Client Batching LeaderTimes - 2a] | Refactor coalescing leaderTimeGetter

sudiksha27:
**Goals (and why)**:
Refactor LockLeaseService so we can inject leaderTimeGetter. As of now, we are falling back on the original (coalescing) leaderTimeGetter.

- Refactor out LeaderTimeGetter

- Implement CoalescingLeaderTimeGetter (coalescing requests for one namespace)

- Implement NamespacedLeaderTimeGetter -> it maintains the context of namespace for a client and directs requests to 
coalescingBatcher that batches requests across clients

- Allow AutoBatcherBuilder to accept bufferSize

Existing tests should suffice for now

- Does the refactor seem reasonable?

LeaderTimeGetter.java, 

Today please


lock-api/src/main/java/com/palantir/lock/client/LeaderTimeCoalescingBatcher.java
```diff
@@ -0,0 +1,63 @@
...
+import com.palantir.atlasdb.futures.AtlasFutures;
+import com.palantir.atlasdb.timelock.api.MultiClientConjureTimelockService;
+import com.palantir.atlasdb.timelock.api.Namespace;
+import com.palantir.lock.v2.LeaderTime;
+import com.palantir.tokens.auth.AuthHeader;
+import java.util.Map;
+import java.util.Set;
+
+public class LeaderTimeCoalescingBatcher implements AutoCloseable {
+    private static final int BUFFER_SIZE = 16384;
```
jeremyk-91, line 31: I think this should still be configurable - internal atlasdb-proxy probably wants more than 16384.

lock-api/src/main/java/com/palantir/lock/client/CoalescingLeaderTimeGetter.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.common.concurrent.CoalescingSupplier;
+import com.palantir.lock.v2.LeaderTime;
+
+public class CoalescingLeaderTimeGetter implements LeaderTimeGetter {
```
jeremyk-91, line 22: I'd name this differently - the problem here is that we have two implementations that are both logically coalescing, so this name isn't very helpful. I'd name this `SingleClientLeaderTimeGetter`? Not even opposed to `LegacyLeaderTimeGetter` at this point

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -47,19 +46,21 @@
...
         this.clientId = clientId;
-        this.time = new CoalescingSupplier<>(delegate::leaderTime);
+        this.leaderTimeGetter = leaderTimeGetter.orElseGet(() -> new CoalescingLeaderTimeGetter(delegate));
         this.lockService = BlockEnforcingLockService.create(delegate);
     }
 
-    static LockLeaseService create(NamespacedConjureTimelockService conjureTimelock) {
-        return new LockLeaseService(conjureTimelock, UUID.randomUUID());
+    static LockLeaseService create(
+            NamespacedConjureTimelockService conjureTimelock, Optional<LeaderTimeGetter> leaderTimeGetter) {
```
jeremyk-91, line 62: I would make this non-Optional actually: the setting of a default should happen at the TransactionManagers builder level, and then all usages within Atlas code should be non-optional (e.g. see how `lockWatchingCache` is set)

lock-api/src/main/java/com/palantir/lock/client/NamespacedCoalescingLeaderTimeGetter.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.atlasdb.timelock.api.Namespace;
+import com.palantir.lock.v2.LeaderTime;
+
+public class NamespacedCoalescingLeaderTimeGetter implements LeaderTimeGetter {
```
jeremyk-91, line 22: I'd add some javadocs here, explaining more precisely how this works, as the difference between this and the "obvious" implementation is important and subtle.

jeremyk-91 (commented):
🕐  0:17 🏁  0:34

The wiring looks fine, but I'd want to minimise the amount we pass Optional around

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -88,6 +89,7 @@ private Autobatchers() {}
         private final ImmutableMap.Builder<String, String> safeTags = ImmutableMap.builder();
 
         private Observability observability = Observability.UNDECIDED;
+        private OptionalInt optionalBufferSize = OptionalInt.empty();
```
jeremyk-91, line 92: nit: I'd just name this `bufferSize`

jeremyk-91 (approved):
🕐 0:07 🏁 2:34

👍 Looks good.

PR #5122, opened 2020-11-19
[Cross Client Batching LeaderTimes - 2b] | MultiClientConjureTimelockService Dialogue Clients

sudiksha27:
**Goals (and why)**:
Add Dialogue client for MultiClientConjureTimelockService

Is short timeout a reasonable choice here?

AtlasDbDialogueServiceProvider.java

Tomorrow EOD


lock-api/src/main/java/com/palantir/lock/client/DialogueAdaptingMultiClientConjureTimelockService.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.palantir.lock.client;
+
+import com.palantir.atlasdb.timelock.api.LeaderTimes;
+import com.palantir.atlasdb.timelock.api.MultiClientConjureTimelockServiceBlocking;
+import com.palantir.atlasdb.timelock.api.Namespace;
+import com.palantir.tokens.auth.AuthHeader;
+import java.util.Set;
+
+public class DialogueAdaptingMultiClientConjureTimelockService {
```
jeremyk-91, line 25: This should implement `MultiClientConjureTimelockService`?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/AtlasDbDialogueServiceProvider.java
```diff
@@ -113,6 +115,17 @@ ConjureTimelockService getConjureTimelockService() {
         return new TimeoutSensitiveConjureTimelockService(shortAndLongTimeoutServices);
     }
 
+    DialogueAdaptingMultiClientConjureTimelockService getMultiClientConjureTimelockService() {
```
jeremyk-91, line 118: This shouldn't depend on the concrete type

jeremyk-91 (commented):
🕐 0:04 🏁 2:39

The construction makes sense, but in general you should depend on abstractions as opposed to concrete implementations (and so, I'd claim that DAMCCTS should implement MCCTS and you should only depend on MCCTS elsewhere). There's no reason why the leader time batcher needs to know that the delegate has anything to do with Dialogue.

lock-api/src/main/java/com/palantir/lock/client/LeaderTimeCoalescingBatcher.java
```diff
@@ -49,15 +47,15 @@ public void close() {
     }
 
     static class LeaderTimeCoalescingConsumer implements CoalescingRequestFunction<Namespace, LeaderTime> {
-        private final MultiClientConjureTimelockService delegate;
+        private final AuthenticatingMultiClientConjureTimelockService delegate;
 
-        public LeaderTimeCoalescingConsumer(MultiClientConjureTimelockService delegate) {
+        public LeaderTimeCoalescingConsumer(AuthenticatingMultiClientConjureTimelockService delegate) {
```
Jolyon-S, line 52: IntelliJ says this can be package-private, so please do so.

lock-api/src/main/java/com/palantir/lock/client/AuthenticatingMultiClientConjureTimelockService.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.atlasdb.timelock.api.LeaderTimes;
+import com.palantir.atlasdb.timelock.api.Namespace;
+import java.util.Set;
+
+public interface AuthenticatingMultiClientConjureTimelockService {
```
Jolyon-S, line 23: I _think_ I understand why we need the impl, but not the interface - I feel like this is a very specific interface (where I'd rather there is a more generic interface and AMCCTS is one of the implementations). Regardless, this is also possibly a misleading name: this always passes a token that is "Bearer omitted", which while technically is on the theme of authentication, feels a bit odd to call this an authenticating service.

Jolyon-S (commented):
A couple of comments - it feels a bit strange personally, but that's possibly just the naming. I am aware that a raw `MultiClientConjureTimelockService` interface does exist, but it's generated in Jersey to probably not what we want to implement (even if that is the kind of interface I'd imagine we want).

lock-api/src/main/java/com/palantir/lock/client/AuthenticatingMultiClientConjureTimelockService.java
jeremyk-91: I wouldn't name this in this way: users elsewhere in the AtlasDB client code shouldn't need to know that authentication (or the lack thereof!) is a thing that's happening beyond this layer, which is an implementation detail. Conceptually this interface should be something like `InternalMultiClientConjureTimelockService` (possibly not a great name) and then your `AuthMCCTSImpl` should be `AuthIMCCTS` (or even `UnauthedIMCCTS`).

jeremyk-91 (commented):
🕐 0:06 🏁 2:28

The components look correct

lock-api/src/main/java/com/palantir/lock/client/AuthenticatingMultiClientConjureTimelockService.java
sudiksha27: I hit a mental block in trying to come up with a name for this interface, ```InternalMultiClientConjureTimelockService``` is reasonable.

lock-api/src/main/java/com/palantir/lock/client/AuthInternalMultiClientConjureTimelockService.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.palantir.lock.client;
+
+import com.palantir.atlasdb.timelock.api.LeaderTimes;
+import com.palantir.atlasdb.timelock.api.MultiClientConjureTimelockServiceBlocking;
+import com.palantir.atlasdb.timelock.api.Namespace;
+import com.palantir.tokens.auth.AuthHeader;
+import java.util.Set;
+
+public class AuthInternalMultiClientConjureTimelockService implements InternalMultiClientConjureTimelockService {
```
jeremyk-91, line 25: nit: sorry for confusion, `AuthIMCCTS` was just used as a short-hand, I'd prefer we use either `Authenticated` or `Unauthenticated` in full rather than `Auth`, which doesn't say much

jeremyk-91 (approved):

PR #5131, opened 2020-11-25
Clock went backwards corruption check

sudiksha27:
**Goals (and why)**:
TimeLock should validates the expected invariant for its timestamp bounds: they must increase with increasing sequence numbers.

- LocalTimestampInvariantsVerifier - verifies  timestamp bounds increase with increasing sequence numbers

- The validation is done batch wise, e.g. [1, n], [n, 2 * n - 1] and so on. Two consecutive batches share
 boundaries, this is done to catch inversion at batch end.

- The verifier uses the existing corruption handling infra i.e. invariant violation will show up in the health check and (when enabled) is capable of killing TimeLock in the case of violation.

Added unit tests

- Does the design seem reasonable?

- Can we optimize the implementation?

- Did I miss any edge cases?

HistoryAnalyzer.java

End of week.

### DIFF
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
@@ -17,7 +17,6 @@
 package com.palantir.timelock.corruption.detection;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
@@ -38,7 +37,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import org.immutables.value.Value;
 
 public final class HistoryAnalyzer {
     private HistoryAnalyzer() {
@@ -79,7 +77,6 @@ public final class HistoryAnalyzer {
         List<Function<CompletePaxosHistoryForNamespaceAndUseCase, CorruptionCheckViolation>> violationChecks =
                 ImmutableList.of(
                         HistoryAnalyzer::divergedLearners,
-                        HistoryAnalyzer::clockWentBackwards,
                         HistoryAnalyzer::learnedValueWithoutQuorum,
                         HistoryAnalyzer::greatestAcceptedValueNotLearned);
         return violationChecks.stream()
@@ -98,14 +95,6 @@ public final class HistoryAnalyzer {
                 })
                 ? CorruptionCheckViolation.NONE
                 : CorruptionCheckViolation.DIVERGED_LEARNERS;
-    }
-
-    @VisibleForTesting
-    static CorruptionCheckViolation clockWentBackwards(CompletePaxosHistoryForNamespaceAndUseCase history) {
-        return history.localAndRemoteLearnerAndAcceptorRecords().stream()
-                        .anyMatch(HistoryAnalyzer::clockWentBackwardsOnNode)
-                ? CorruptionCheckViolation.CLOCK_WENT_BACKWARDS
-                : CorruptionCheckViolation.NONE;
     }
 
     @VisibleForTesting
@@ -205,36 +194,5 @@ public final class HistoryAnalyzer {
     // This method is only used for paxos rounds run for timestamp consensus (leader paxos rounds are strictly ignored).
     private static byte[] getPaxosValueData(Optional<PaxosValue> learnedValue) {
         return learnedValue.map(PaxosValue::getData).orElse(null);
-    }
-
-    private static boolean clockWentBackwardsOnNode(ConsolidatedLearnerAndAcceptorRecord record) {
-        List<SequenceAndTimestampPair> seqAndTimestampPairs = KeyedStream.stream(record.record())
-                .map(val -> getPaxosValueData(val.learnedValue()))
-                .filter(Predicates.notNull())
-                .mapEntries((sequence, timestamp) -> Maps.immutableEntry(
-                        sequence, ImmutableSequenceAndTimestampPair.of(sequence, PtBytes.toLong(timestamp))))
-                .values()
-                .sorted(Comparator.comparingLong(SequenceAndTimestampPair::sequence))
-                .collect(Collectors.toList());
-
-        long greatestTimestampSoFar = -1L;
-        long currentTimestamp;
-        for (SequenceAndTimestampPair sequenceAndTimestampPair : seqAndTimestampPairs) {
-            currentTimestamp = sequenceAndTimestampPair.timestamp();
-            if (currentTimestamp <= greatestTimestampSoFar) {
-                return true;
-            }
-            greatestTimestampSoFar = currentTimestamp;
-        }
-        return false;
-    }
-
-    @Value.Immutable
-    interface SequenceAndTimestampPair {
-        @Value.Parameter
-        long sequence();
-
-        @Value.Parameter
-        long timestamp();
     }
 }
--- a/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
+++ b/timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
@@ -17,6 +17,7 @@
 package com.palantir.timelock.corruption.detection;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
@@ -37,6 +38,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.immutables.value.Value;
 
 public final class HistoryAnalyzer {
     private HistoryAnalyzer() {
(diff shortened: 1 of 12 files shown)


timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/HistoryAnalyzer.java
```diff
@@ -195,4 +206,35 @@ private static int getQuorumSize(List<ConsolidatedLearnerAndAcceptorRecord> reco
...
+    }
+
+    @Value.Immutable
+    interface SequenceAndTimestampPair {
+        @Value.Parameter
+        long sequence();
+
+        @Value.Parameter
+        long timestamp();
+    }
```
gmaretic, line 239: If you want to be a bit fancy:
```suggestion
        Stream<Long> expectedSortedTimestamps = KeyedStream.stream(record.record())
                .map(val -> getPaxosValueData(val.learnedValue()))
                .filter(Predicates.notNull())
                .mapEntries((sequence, timestamp) -> Maps.immutableEntry(sequence, PtBytes.toLong(timestamp)))
                .entries()
                .sorted(Comparator.comparingLong(Entry::getKey))
                .map(Entry::getValue);
        return StreamEx.of(expectedSortedTimestamps)
                .pairMap((first, second) -> first >= second)
                .anyMatch(x -> x);
```
If you don't want to use `StreamEx`, you can still avoid the immutable by checking `expectedSortedTimestamps` in a for loop, but just compare neighbours instead of maintaining a global maximum

gmaretic (commented):
Makes sense, though the analyzer itself need some polish

sudiksha27:
This will need redesigning. Will be moving the check out of History analyzer and adding a separate thread to verify timestamp invariants.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifier.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.sql.DataSource;
+import one.util.streamex.StreamEx;
+
+public class LocalTimestampInvariantsVerifier {
+    @VisibleForTesting
+    static final int DELTA = 5;
+
+    private final SqlitePaxosStateLogHistory sqlitePaxosStateLogHistory;
```
sudiksha27, line 45: This won't work, need to fetch the next seq number > upper and use range [lower, next_seq_after_upper] for analysis.

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/LocalCorruptionDetector.java
```diff
@@ -34,30 +34,44 @@
...
     private void scheduleWithFixedDelay() {
         executor.scheduleWithFixedDelay(
                 () -> {
-                    localCorruptionReport = analyzeHistoryAndBuildCorruptionHealthReport();
+                    CorruptionHealthReport paxosRoundCorruptionReport = analyzeHistoryAndBuildCorruptionHealthReport();
+                    CorruptionHealthReport timestampInvariantsReport =
+                            timestampInvariantsVerifier.timestampInvariantsHealthReport();
+                    localCorruptionReport = ImmutableCorruptionHealthReport.builder()
+                            .putAllViolatingStatusesToNamespaceAndUseCase(
+                                    paxosRoundCorruptionReport.violatingStatusesToNamespaceAndUseCase())
```
gmaretic, line 71:
```suggestion
                                                        .from(paxosRoundCorruptionReport)
```

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifier.java
```diff
@@ -0,0 +1,129 @@
...
+                .collectToSetMultimap();
+        return ImmutableCorruptionHealthReport.builder()
+                .violatingStatusesToNamespaceAndUseCase(namespacesExhibitingViolations)
+                .build();
+    }
+
+    private CorruptionCheckViolation timestampInvariantsViolationLevel(NamespaceAndUseCase namespaceAndUseCase) {
+        Stream<Long> expectedSortedTimestamps = KeyedStream.stream(getLearnerLogs(namespaceAndUseCase))
+                .map(PaxosValue::getData)
+                .filter(Predicates.notNull())
```
gmaretic, line 75:
```suggestion
                .filter(Objects::nonNull)
```

timelock-corruption-detection/src/main/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifier.java
```diff
@@ -0,0 +1,129 @@
...
+ * This class validates that timestamp bounds increase with increasing sequence numbers.
+ *
+ * The validation is done batch wise, e.g. [1, n], [n, 2 * n - 1] and so on. Two consecutive batches share
+ * boundaries, this is expected and is done to catch inversion at batch end.
+ * */
+public class LocalTimestampInvariantsVerifier {
+    @VisibleForTesting
+    public static final int LEARNER_LOG_BATCH_SIZE_LIMIT = 250;
+
+    public static final long MIN_SEQUENCE_TO_BE_VERIFIED = -1L;
```
gmaretic, line 51:
```suggestion
    public static final long MIN_SEQUENCE_TO_BE_VERIFIED = Long.MIN_VALUE;
```

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Test;
+
+public class LocalTimestampInvariantsVerifierTest {
+    @Rule
+    public TimeLockCorruptionDetectionHelper helper = new TimeLockCorruptionDetectionHelper();
+
+    @Test
+    public void detectCorruptionIfClockWentBackwardsOnNode() {
+        helper.writeLogsOnDefaultLocalServer(1, LEARNER_LOG_BATCH_SIZE_LIMIT - 1);
+        helper.forceTimestampToGoBackwards(5);
```
gmaretic, line 31: This method name is kinda misleading, it's actually forcing the timestamp to go forward / sequence to go backwards. It's only making the timestamp go backwards for the sequence after, right?

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public void detectIfClockWentBackwardsAtBatchStart() {
+        helper.writeLogsOnDefaultLocalServer(1, LEARNER_LOG_BATCH_SIZE_LIMIT);
+        helper.forceTimestampToGoBackwards(1);
+        helper.assertClockWentBackwards();
+    }
+
+    @Test
+    public void detectIfClockWentBackwardsAtBatchEnd() {
+        helper.writeLogsOnDefaultLocalServer(1, LEARNER_LOG_BATCH_SIZE_LIMIT);
+        helper.forceTimestampToGoBackwards(LEARNER_LOG_BATCH_SIZE_LIMIT / 2);
```
gmaretic, line 45: Shouldn't this then be `helper.forceTimestampToGoBackwards(LEARNER_LOG_BATCH_SIZE_LIMIT - 1);`?

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Test
+    public void resetsProgressIfNotEnoughLogsForVerification() {
+        helper.writeLogsOnDefaultLocalServer(1, 1);
+        // No signs of corruption
+        helper.assertLocalTimestampInvariantsStand();
+
+        helper.writeLogsOnDefaultLocalServer(2, LEARNER_LOG_BATCH_SIZE_LIMIT);
+        helper.forceTimestampToGoBackwards(LEARNER_LOG_BATCH_SIZE_LIMIT / 2);
```
gmaretic, line 80:
```suggestion
        helper.forceTimestampToGoBackwards(1);
```
This way we test that we don't skip the one entry from first check. Also I guess do this before the first check to be consistent

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/TimeLockCorruptionDetectionHelper.java
```diff
@@ -105,6 +105,31 @@ void assertViolationDetectedForNamespaceAndUseCases(
         return timeLockCorruptionTestSetup.createStatLogForNamespaceAndUseCase(namespaceAndUseCase);
     }
 
+    void assertClockWentBackwards() {
```
gmaretic, line 108: Maybe rename to `assertClockWentBackwardsInNextBatch` to be clearer? Same for the method below

gmaretic (commented):
Mostly looks good, have some nits

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
sudiksha27: Ah yes, that makes sense. Basically what I mean is - create inversion b/w this sequence and the next. I have changed the name to createTimestampInversion now.

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
sudiksha27: Good catch! yes, this is a copy pasting mistake.

timelock-corruption-detection/src/test/java/com/palantir/timelock/corruption/detection/LocalTimestampInvariantsVerifierTest.java
sudiksha27: Done.

gmaretic (approved):
LGTM

PR #5137, opened 2020-11-27
[Productionizing GetSortedColumns] Part 1: Row Column Range Traps

jeremyk-91:
**Goals (and why)**:
- Fix long running bugs in AtlasDB around the use of multiple iterators in getRowsColumnRange.
- See @j-baker's #5092 

- Replace offending usages of Maps.transformEntries(...) that generate distinct objects.
- Add tests.

All new code is tested

Nothing much

**Where should we start reviewing?**: SnapTransTest and SerTransTest

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -214,7 +214,7 @@ public class SerializableTransaction extends SnapshotTransaction {
                 super.getRowsColumnRangeIterator(tableRef, rows, columnRangeSelection);
         return KeyedStream.stream(ret)
                 .map((row, iterator) -> wrapIteratorWithBoundsChecking(tableRef, columnRangeSelection, row, iterator))
-                .collectToMap();
+                .collectTo(() -> new TreeMap<>(UnsignedBytes.lexicographicalComparator()));
     }
 
     @Override
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -827,15 +827,15 @@ public class SerializableTransaction extends SnapshotTransaction {
                 tableRef, getTimestamp(), System.currentTimeMillis() - timeCreated);
     }
 
-    private BatchingVisitable<Entry<Cell, byte[]>> wrapWithColumnRangeChecking(
+    private BatchingVisitable<Map.Entry<Cell, byte[]>> wrapWithColumnRangeChecking(
             TableReference tableRef,
             BatchColumnRangeSelection columnRangeSelection,
             byte[] row,
-            BatchingVisitable<Entry<Cell, byte[]>> visitable) {
-        return new BatchingVisitable<Entry<Cell, byte[]>>() {
+            BatchingVisitable<Map.Entry<Cell, byte[]>> visitable) {
+        return new BatchingVisitable<Map.Entry<Cell, byte[]>>() {
             @Override
             public <K extends Exception> boolean batchAccept(
-                    int batchSize, AbortingVisitor<? super List<Entry<Cell, byte[]>>, K> visitor) throws K {
+                    int batchSize, AbortingVisitor<? super List<Map.Entry<Cell, byte[]>>, K> visitor) throws K {
                 boolean hitEnd = visitable.batchAccept(batchSize, items -> {
                     if (items.size() < batchSize) {
                         reachedEndOfColumnRange(tableRef, row, columnRangeSelection);
@@ -851,13 +851,13 @@ public class SerializableTransaction extends SnapshotTransaction {
         };
     }
 
-    public Iterator<Entry<Cell, byte[]>> wrapIteratorWithBoundsChecking(
+    public Iterator<Map.Entry<Cell, byte[]>> wrapIteratorWithBoundsChecking(
             TableReference tableRef,
             BatchColumnRangeSelection columnRangeSelection,
             byte[] row,
-            Iterator<Entry<Cell, byte[]>> iterator) {
-        return new Iterator<Entry<Cell, byte[]>>() {
-            Entry<Cell, byte[]> next = null;
+            Iterator<Map.Entry<Cell, byte[]>> iterator) {
+        return new Iterator<Map.Entry<Cell, byte[]>>() {
+            Map.Entry<Cell, byte[]> next = null;
 
             @Override
             public boolean hasNext() {
@@ -876,11 +876,11 @@ public class SerializableTransaction extends SnapshotTransaction {
             }
 
             @Override
-            public Entry<Cell, byte[]> next() {
+            public Map.Entry<Cell, byte[]> next() {
                 if (!hasNext()) {
                     throw new NoSuchElementException();
                 }
-                Entry<Cell, byte[]> result = next;
+                Map.Entry<Cell, byte[]> result = next;
                 next = null;
                 return result;
             }
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -70,6 +70,7 @@ import com.palantir.common.base.BatchingVisitableView;
 import com.palantir.common.base.Throwables;
 import com.palantir.common.collect.IterableUtils;
 import com.palantir.common.collect.Maps2;
+import com.palantir.common.streams.KeyedStream;
 import com.palantir.lock.v2.LockToken;
 import com.palantir.lock.v2.TimelockService;
 import com.palantir.logsafe.Preconditions;
@@ -81,6 +82,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.NoSuchElementException;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -191,23 +193,9 @@ public SerializableTransaction(
...
-                });
-                if (hitEnd) {
-                    reachedEndOfColumnRange(tableRef, row, columnRangeSelection);
-                }
-                return hitEnd;
-            }
-        });
+        return KeyedStream.stream(ret)
+                .map((row, visitable) -> wrapWithColumnRangeChecking(tableRef, columnRangeSelection, row, visitable))
+                .collectTo(() -> new TreeMap<>(UnsignedBytes.lexicographicalComparator()));
```
Jolyon-S, line 198: [Sanity check]: This collection is consistent with the one in SnapshotTransaction after this PR. I need to validate that this is equivalent before, too.

Is there a way that we can force this coupling? I.e. what if someone in the future changes one of the collectors and not the other?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -865,6 +827,66 @@ private void handleTransactionConflict(TableReference tableRef) {
                 tableRef, getTimestamp(), System.currentTimeMillis() - timeCreated);
     }
 
+    private BatchingVisitable<Map.Entry<Cell, byte[]>> wrapWithColumnRangeChecking(
```
Jolyon-S, line 830: [Sanity check]: This code is semantically the same as the previous code, right? As far as I can tell, the main change in this class is the use of KeyedStream over Maps.transformEntries?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -381,9 +383,9 @@ public void markTableInvolved(TableReference tableRef) {
     @Override
     public Map<byte[], BatchingVisitable<Map.Entry<Cell, byte[]>>> getRowsColumnRange(
             TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection) {
-        return Maps.transformEntries(
-                getRowsColumnRangeIterator(tableRef, rows, columnRangeSelection),
-                (row, iterator) -> BatchingVisitableFromIterable.create(iterator));
+        return KeyedStream.stream(getRowsColumnRangeIterator(tableRef, rows, columnRangeSelection))
+                .map((row, iterator) -> BatchingVisitableFromIterable.create(iterator))
+                .collectTo(() -> new TreeMap<>(UnsignedBytes.lexicographicalComparator()));
```
Jolyon-S, line 388: Ah, so the ordering actually comes from `getRowsColumnRangeIterator`. As mentioned elsewhere, is there any way we can be defensive and make sure that we always use a sorted collector? I guess tests would fail if you just used any random collector?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1091,6 +1095,68 @@ public void testDisableReadWriteConflictChecking() {
...
+                BatchColumnRangeSelection.create(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 1000));
+
+        Iterator<Map.Entry<Cell, byte[]>> iterator1 = iterators.get(row);
+        Iterator<Map.Entry<Cell, byte[]>> iterator2 = iterators.get(row);
+        assertThat(iterator1.hasNext()).isTrue();
+        assertThat(iterator2.hasNext()).isTrue();
+
+        Map.Entry<Cell, byte[]> entry = iterator1.next();
+        assertThat(entry.getKey()).isEqualTo(cell);
+        assertThat(Arrays.equals(entry.getValue(), value)).isTrue();
```
Jolyon-S, line 1157: What happens if you call `iterator2.next()`?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1091,6 +1095,68 @@ public void testDisableReadWriteConflictChecking() {
         t2.commit();
     }
 
+    @Test
+    public void testGetRowsColumnRangeMultipleIteratorsWorkSafely() {
```
Jolyon-S, line 1099: Is there a test somewhere in this class that checks the output is correctly ordered? If not, can we add one such test? Even if there is a test in `SnapshotTransaction` there should be one here too, to confirm that the ordering stays the same.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1220,6 +1221,41 @@ public void cleanup() {}
         assertEquals(ImmutableList.of(firstCell, secondCell), cells);
     }
 
+    @Test
+    public void testGetRowsColumnRangeMultipleIteratorsWorkSafely() {
```
Jolyon-S, line 1225: You added two tests for the `SerializableTransaction`, why only one here?

Jolyon-S: Also, similar comment as above - do we test for the right ordering of values read?

Jolyon-S (commented):
The changelog seems to imply that this is potentially a semantic change for some - is that right? Would that be just that they are depending on bad behaviour?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
jeremyk-91: Yep, that's right.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
jeremyk-91: This is equivalent before (see javadoc of `Maps#transformValues`: it takes on properties of the previous map). 

That said, in general, it is not _wrong_ for these to be different types, as long as they satisfy the properties we need (here, mainly that a map can be looked up by value of `byte[]`) so they don't need to be coupled. The tests should fail pretty quickly if that property is broken (since they do perform lookups).

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
jeremyk-91: Yes (though it's a bit non-obvious to find): `AbstractTransactionTest#testColumnRangePagingTransaction_batchingVisitable()`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
Jolyon-S: I'm convinced this is the same as before, but my concern is that someone depends on this being ordered (i.e. not just a lookup but also an ordering), and that this could be broken in a later PR.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91: I changed two methods in `SerializableTransaction` but only one in `SnapshotTransaction`: `getRowsColumnRangeIterator` was already safely implemented in `SnapshotTransaction` but not in `SerializableTransaction`. That said I think I'll move the two in the AbSTT to the general AbTT, so this test shouldn't be needed any more.

In terms of value testing: yes - this is in the abstract transaction test.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
jeremyk-91: NoSuchElementException - I've added a pair of false hasNext() calls after the value is read.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: Yep, as discussed on the test comments - there are tests that verify that elements are returned in order. 

The interface demands a Map (but not a SortedMap and we can't change that super easily), so the best we can do is maybe something like 

```
    @Override
    public Map<byte[], BatchingVisitable<Map.Entry<Cell, byte[]>>> getRowsColumnRange(
            TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection) {
        SortedMap<byte[], BatchingVisitable<Entry<Cell, byte[]>>> sortedVisitables =
                KeyedStream.stream(getRowsColumnRangeIterator(tableRef, rows, columnRangeSelection))
                        .map((row, iterator) -> BatchingVisitableFromIterable.create(iterator))
                        .collectTo(() -> new TreeMap<>(UnsignedBytes.lexicographicalComparator()));
        return sortedVisitables;
    }
```

though given the tests I think that's unnecessary.

Jolyon-S: Reasonable - I suppose the fact that it is not a SortedMap makes anyone's reliance on the ordering a bit suspicious. And either way, so long as the tests catch this we should catch any regressions.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
jeremyk-91: I'd be fine with breaking someone on that - the API documentation (see `Transaction`) does not guarantee that behaviour.

jeremyk-91: If we wanted to guarantee that we'd probably return `SortedMap` as interface type as well

Jolyon-S (approved):
LGTM.

PR #5140, opened 2020-12-02
[LW] No Op cache tracks version

Jolyon-S:
**Goals (and why)**:
If the no op cache was actually used after fallback, it would request snapshots from Timelock on every single start and end transaction. That would be badTM.

* `NoOpLockWatchEventCache` now stores and tracks its own version.

* All the existing integration tests now validate that the version is the same between the two caches.

* Ensure this is right.

* `NoOpLockWatchEventCache`

This week, really.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransactionManager.java
```diff
@@ -472,8 +472,8 @@ public static SerializableTransactionManager createForTest(
                 metricsManager,
                 keyValueService,
                 new LegacyTimelockService(timestampService, lockService, lockClient),
-                NoOpLockWatchManager.INSTANCE,
-                NoOpLockWatchEventCache.INSTANCE,
+                NoOpLockWatchManager.create(),
+                NoOpLockWatchEventCache.create(),
```
Jolyon-S, line 476: Strictly speaking, this is a bit weird (the LWM has a different cache to the one passed directly below), but it's for tests. No strong pref if you think we should resolve this.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/DuplicatingLockWatchEventCache.java
```diff
@@ -0,0 +1,84 @@
...
+            Set<Long> startTimestamps, Optional<LockWatchVersion> version) {
+        return mainCache.getUpdateForTransactions(startTimestamps, version);
+    }
+
+    @Override
+    public void removeTransactionStateFromCache(long startTimestamp) {
+        mainCache.removeTransactionStateFromCache(startTimestamp);
+    }
+
+    private void validateVersionParity() {
```
jeremyk-91, line 81: discussed in internal PR: I know you intend parity as in equality, but would suggest using equality as "validating parity" often means checking oddness/evenness which is distinctly not what we're doing here

changelog/@unreleased/pr-5140.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: fix
+fix:
+  description: The No op lock watch event cache now correctly tracks the latest known
```
jeremyk-91, line 3: nit: `no-op` or `no op` (or even `No Op`?) - something seems strange with the capitalisation

lock-api/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
```diff
@@ -58,11 +67,34 @@ public TransactionsLockWatchUpdate getUpdateForTransactions(
...
     @Override
     public void removeTransactionStateFromCache(long startTimestamp) {}
+
+    private void updateVersion(Optional<LockWatchVersion> maybeNewVersion) {
+        currentVersion = maybeNewVersion.map(newVersion -> currentVersion
+                .filter(current -> current.id().equals(newVersion.id()) && current.version() > newVersion.version())
+                .orElse(newVersion));
+    }
+
+    private Optional<LockWatchVersion> extractVersionFromUpdate(LockWatchStateUpdate update) {
```
jeremyk-91, line 84: Unclear if this has been a convention throughout Lock Watches, but the decomp - in particular, rejecting a success from post-leader election without a snapshot _here_ - seems a bit unexpected. I'd prefer this just return the version, and updateVersion take care of that filtering. It looks like you do that anyway in updateVersion!

jeremyk-91 (commented):
🕐 0:05 🏁 !

Seems broadly right, some minor comments.

lock-api/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
Jolyon-S: So in the real implementation, a success after a leader election _will_ fail (as in, the version will be empty). `updateVersion` doesn't quite do the same thing - if there has been a leader election without a snapshot, `updateVersion` will set it to empty, whereas if we permitted this, `currentVersion` would have a value.

Jolyon-S: The test `leaderChangeClearsCache` in `LWECIT` demonstrates this behaviour - accepting a success from a post-leader election causes the versions to diverge.

Jolyon-S: There's a good reason for this - we can't have a valid version unless we've had a snapshot telling us:
a) what descriptors are being watched;
b) what the currently held watches are.

Otherwise we could tell the client that there are only X watches taken out, when that may not be accurate.

lock-api/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
```diff
@@ -58,11 +67,34 @@ public TransactionsLockWatchUpdate getUpdateForTransactions(
...
                 .build();
     }
 
     @Override
     public void removeTransactionStateFromCache(long startTimestamp) {}
+
+    private void updateVersion(Optional<LockWatchVersion> maybeNewVersion) {
+        currentVersion = maybeNewVersion.map(newVersion -> currentVersion
+                .filter(current -> current.id().equals(newVersion.id()) && current.version() > newVersion.version())
+                .orElse(newVersion));
```
jeremyk-91, line 81: from discussion below: can we use imperative statements here? I find this difficult to parse (I read the filter as a top-level thing that rejects versions that were unfriendly, when that's not the case).

lock-api/src/main/java/com/palantir/lock/watch/NoOpLockWatchEventCache.java
jeremyk-91: Ah, I misread the updateVersion code - I've left a comment there.

I still find `extractVersionFromUpdate` to return an Optional weird: the update clearly _has_ a log ID and a version, just it may not be one we're interested in. I think we could

1. add a brief comment indicating the properties of when this returns the version
2. name this `extractRelevantVersion...`?

I don't have a good name for 2 so suggest 1.

jeremyk-91 (approved):
👍  thanks!

jeremyk-91:
If one looks at the test build, one gets 
> GitHub returned status 500 adding commit status of success

Thus merging.

PR #5145, opened 2020-12-04
[LW] Smile for the snapshot

Jolyon-S:
**Goals (and why)**:
There was yet another edge case where this check was bad. Let's fix it.

Fix the check.

Added yet another test.

Make sure we don't mess this up.

ClientLogEvents

Yesterday :fire:

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
@@ -65,9 +65,12 @@ interface ClientLogEvents {
          Case 3: client is completely up-to-date. Here, we don't need to check for any versions.
          Case 4: client has no version. Then we expect that the events returned at least enclose the versions of
                  the transactions.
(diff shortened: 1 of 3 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
```diff
@@ -65,9 +65,12 @@ each version starting with the client version (exclusive) and ending with latest
          Case 3: client is completely up-to-date. Here, we don't need to check for any versions.
          Case 4: client has no version. Then we expect that the events returned at least enclose the versions of
                  the transactions.
+         Case 5: the client is very far behind, but still has a version. In this case, we should not check based on the
+                 client version, but rather the lower bound of the timestamp mapping.
         */
         verifyReturnedEventsEnclosesTransactionVersions(
                 lastKnownVersion
+                        .filter(_unused -> !clearCache())
```
jeremyk-91, line 73: doh

jeremyk-91 (approved):
🕐 0:04 🏁 0:05

Makes sense. Good spot!

PR #5146, opened 2020-12-04
[LW] [PDS-148088] Close your eyes

Jolyon-S:
**Goals (and why)**:
Now only watch the tables we request, as opposed to a union of all tables requested to watch

Clear, then add all.

Test this case. The test fails if I remove my change.

Race condition? Should only be called within the txn manager, and since atlasdb-proxy only does that in a B/G task in a single-threaded way, should not be a concern.

LockWatchManagerImpl.

yesterday :fire:


changelog/@unreleased/pr-5146.v2.yml
```diff
@@ -0,0 +1,9 @@
+type: fix
+fix:
+  description: Calling registerWatches now tells the lockWatchManager to _only_ watch
+    those watches, instead of adding it to the set of all registered watches so far.
+    Note that this does not affect Timelock (which will still contain all watches
+    registered so far), but rolling Timelock will mean that the LockWatchManager now
+    only re-registers the latest requested watches.
```
jeremyk-91, line 7: I'd probably re-phrase this in terms of what the client sees (i.e. the impl details are not needed, I think the last part is the key here - `The LockWatchManager now only re-registers the latest set of requested watches when TimeLock is bounced`).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -70,6 +70,7 @@ public void close() {
 
     @Override
     public void registerWatches(Set<LockWatchReferences.LockWatchReference> newLockWatches) {
+        lockWatchReferences.clear();
```
jeremyk-91, line 73: the way this works in combination with the javadoc is strange (it seems like register should definitely be commutative and distributive over union based on its javadoc; that it's not is a bit weird and so needs calling out I think, at least internally). I'd suggest we either
- keep this endpoint the way it is, and have a new endpoint called `registerPreciselyWatches` or something like that that makes it apparent that the client will no longer request previously registered things to be watched.
- or just rename this to `registerPreciselyWatches`, if the previous behaviour is not used anywhere

jeremyk-91 (commented):
🕐 0:10 🏁 0:13

I think I agree with this change in principle, though we should make sure that people don't accidentally violate the assumptions internal to this class later on.

jeremyk-91 (approved):
👍

PR #5149, opened 2020-12-08
Make EC2 Location supplier memoize on failures

ashrayjain:
**Goals (and why)**:
This stops the constant warn logging from the HostLocationSupplier in environments where services don't get to access the Instance metadata endpoint by default (example, when running on a locked down K8s cluster).

We will now log an warning once if we fail to get location info from the metadata service and then assume we don't have any location info for the lifetime of the process.

### DIFF
--- a/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
+++ b/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
@@ -44,7 +44,7 @@ public class HostLocationSupplierTest {
                 new HostLocationSupplier(ec2SnitchSupplier, ec2LocationSupplier, Optional.empty());
 
         assertThat(hostLocationSupplier.get())
-                .hasValue(ec2LocationSupplier.get().get());
+                .isEqualTo(ec2LocationSupplier.get());
     }
 
     @Test
--- a/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
+++ b/atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
@@ -43,8 +43,7 @@ public class HostLocationSupplierTest {
         Supplier<Optional<HostLocation>> hostLocationSupplier =
                 new HostLocationSupplier(ec2SnitchSupplier, ec2LocationSupplier, Optional.empty());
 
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
```diff
@@ -42,7 +43,8 @@ public void shouldReturnEc2Location() {
         Supplier<Optional<HostLocation>> hostLocationSupplier =
                 new HostLocationSupplier(ec2SnitchSupplier, ec2LocationSupplier, Optional.empty());
 
-        assertThat(hostLocationSupplier.get()).hasValue(ec2LocationSupplier.get());
+        assertThat(hostLocationSupplier.get())
+                .hasValue(ec2LocationSupplier.get().get());
```
Jolyon-S, line 47: Does this work?
```suggestion
        assertThat(hostLocationSupplier.get())
                .isEqualTo(ec2LocationSupplier.get());
```

Jolyon-S (commented):
How was the exception handled before? Was it just thrown?

Other than that, I think this looks fine. You might find it interesting to know that this piece was the first code I wrote for the company :)

I don't know if @sudiksha27 or @jeremyk-91 want to sanity check before this goes through or not.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
ashrayjain: Giving it a shot, but i don't think it will work. The two suppliers are not "equal", the values they produce are.

ashrayjain:
The exception was caught in HostLocationSupplier previously (and still is) and logged there as a warning. The difference is that Ec2HostLocationSupplier is memoized while HostLocationSupplier is not.

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/pool/HostLocationSupplierTest.java
ashrayjain: I was wrong, we are comparing Optionals here, not Suppliers :). It works.

Jolyon-S: yep, that was my thought! Agreed that comparing suppliers would have been wrong though.

Jolyon-S (approved):

PR #5154, opened 2020-12-11
Make TransactionLockAcquisitionTimeoutException retryable

Jolyon-S:
**Goals (and why)**:
* Make TLATE retryable - this would likely reduce most of the 500s caused by leader elections.

* Change the exception it extends.

N/A

* If there is a genuine issue with locks timing out, we're now going to retry several times before actually failing. This could be an issue.


jeremyk-91 (approved):
See internal discussion thread for more detail, but at a high level:

- we've confirmed internal large product team is OK with this change
- if you only use TransactionManagers this doesn't really affect any guarantees or non-guarantees we used to offer (e.g. user transaction task could always have been retried)
- locks in question are v2, so retrying should be generally harmless because locking is idempotent (this is not the case in v1)
- as @Jolyon-S notes on the PR requests may end up retrying in cases where lock acquisition timing out isn't a _client-side_ transient issue (e.g. timelock is down) but the tradeoff here makes sense; the worst case I can see is that we fail a bit more slowly in cases where we were going to fail anyway.

PR #5161, opened 2020-12-16
[WIP] Publish github pages correctly.

jeremyk-91:
**Goals (and why)**:
- Don't explode the same way the output of #5159 did

- Still only push the docs if the Circle project name is `palantir` but the git user name should remain as `atlasdb`

- No automated testing. I will manually test the script as it runs on this branch.

- Does this actually work? I will test.

**Where should we start reviewing?**: +2 -2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,9 @@ jobs:
       CIRCLE_ARTIFACTS: /tmp/artifacts
     steps:
       - attach_workspace: { at: . }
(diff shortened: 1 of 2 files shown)


jeremyk-91:
The original PR is actually fine, I think there's an issue with the token. I'm checking with the relevant internal teams.

gmaretic (approved):
A bit magical, but as long as it's tested and works lgtm

PR #5162, opened 2020-12-17
bump commons

anjor:
**Goals (and why)**:


jeremyk-91:
This looks right, but you'll need to rerun the dependency locks task: `./gradlew --write-locks`

Also, might be missing something, but we could just remove the version you've specified in the gradle file and add the commons-pool coordinates to `versions.props` (assuming a different version isn't used elsewhere).

jeremyk-91 (approved):

PR #5164, opened 2020-12-17
Don't try to publish internal docs from external CircleCI!

jeremyk-91:
**Goals (and why)**:
- Have a green develop

**Implementation Description (bullets)**: Don't attempt an internal publish

**Testing (What was existing testing like?  What have you done to improve it?)**: Not much testing. Tests on the previous PR only checked that the publish script worked.

- Is there anything else broken?
- This means internal docs will drift out of date. I've added a task on internal task-tracking service for early 2021.

**Where should we start reviewing?**: +2 -1


gmaretic (approved):
🥇

PR #5167, opened 2020-12-18
[LW] Handle Snapshots into Empty Successes Correctly

jeremyk-91:
**Goals (and why)**:
- Fix a bug in lock watches that caused unnecessary snapshots and, worse, monitor false-positives.

- Suppose a Lock Watch Event Cache gets a `Snapshot [logID = L1, sequence = 10000, data = {...}]` state update. This can be because TimeLock had a leader election, or just the node running the cache fell behind. The node then flushes its cache and updates itself to be in sync. However, if no further events happen, the cache will receive multiple state updates of the form `Success [logID = L1, lastKnownVersion = 10000, events = []]`. 
- The check that happens when a user calls `getUpdateForTransactions` and wants to get a `ClientLogEvents` for purposes of atlasdb-proxy says that something is a snapshot if there is no previously known version, the leader has changed or the log is too far behind (see `LockWatchEventLog#getEventsBetweenVersions`). The implementation of the last of these is `!eventStore.containsEntryLessThanOrEqualTo(startVersion.version())` where the `startVersion` is `(l1, 10001)`. In any case, the event store is _empty_ because it was reset when the snapshot was taken so no event prior to 10001 exists in the log, so we repeatedly return updates that have clearCache as true, i.e. we see many snapshots downstream.
- A single success after sequence 10000 that has events will make everything work again, but we really should be resilient to this case. If the version provided by the user equals the latest version on TimeLock, then they are not behind and should not need to clear their cache. Given the point of lock-watches is often to be able to cache values that are not read frequently, having to wait for a write for this to unblock itself seems unacceptable.
- At the same time this is relatively benign compared to the problems in e.g. PDS-148088, because although we are returning a lot of snapshots, these snapshots are generally very small (realistically, the snapshots can be big if there are a lot of watched locks being taken out: but in that case usually the rate of events should be high enough to quickly break you out of this state anyway)

- Integration Test added for this specific case.

- Is there some edge case I've missed out? This change did not change the state of any of the lock watch caches other than the added test, so I think we should be fine.

- LockWatchEventLog.java

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventLog.java
@@ -29,7 +29,6 @@ final class LockWatchEventLog {
     private final ClientLockWatchSnapshot snapshot;
     private final VersionedEventStore eventStore;
     private Optional<LockWatchVersion> latestVersion = Optional.empty();
-    private Optional<LockWatchVersion> lastSuccessfulVersion = Optional.empty();
 
     static LockWatchEventLog create(int maxEvents) {
         return new LockWatchEventLog(ClientLockWatchSnapshot.create(), maxEvents);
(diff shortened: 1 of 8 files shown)


gmaretic (approved):
This was a fun way to spend the Friday :)

PR #5174, opened 2021-01-05
[PDS-146088] Additional logging around Cassandra client pool issues

jeremyk-91:
**Goals (and why)**:
- Understand what is happening in PDS-146088. With this PR in, enabling debug logging on the CassandraClientPoolingContainer origin should yield some interesting information.

- Add logging to creation of a Cassandra client pool
- Add more costly debug logging when eviction decisions are made, in particular negatively

Not much :( existing testing was bad too

- Are any of the SafeArg things actually unsafe?

**Where should we start reviewing?**: it's small


gmaretic (approved):
LGTM, should be good with safe args

PR #5178, opened 2021-01-06
[LE] Add logging to debug metrics + fix bug

gmaretic:
**Goals (and why)**:
The leader election duration estimates can be drastically longer than the observed impact. Furthermore, sometimes there is a leader election duration estimate when impact is 0 and the leader has not apparently changed. These logs should help us gain more insight into what's happening.

Also fix a bug with reporting the metrics.

Is this enough? I know it's a bit aggressive, but hopefully it's not excessive.

New rules say the clock is ticking

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
@@ -170,8 +170,9 @@ public class LeaderElectionReportingTimelockService implements NamespacedConjure
         boolean election = !newLeader.equals(currentLeader);
 
         if (election) {
-            if (leaderId != newLeader) {
-                log.info("Apparent leader change from {} to {}.",
+            if (!leaderId.equals(newLeader)) {
+                log.info(
+                        "Apparent leader change from {} to {}.",
                         SafeArg.of("old leader", leaderId),
                         SafeArg.of("new leader", newLeader));
             }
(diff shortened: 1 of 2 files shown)


lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
```diff
@@ -164,6 +170,12 @@ LeaderElectionStatistics getStatisticsAndSetRegistryTo(TaggedMetricRegistry metr
         boolean election = !newLeader.equals(currentLeader);
 
         if (election) {
+            if (!newLeader.equals(leaderId)) {
```
Jolyon-S, line 173: When would you expect this to be false when election is true? If my read is correct:
`election = !newLeader.equals(currentLeader)`and `currentLeader = leaderId`.
Is this because of the potential for a race condition?

lock-api/src/main/java/com/palantir/lock/client/metrics/TimeLockFeedbackBackgroundTask.java
```diff
@@ -96,11 +96,11 @@ private void scheduleWithFixedDelay() {
                                 .serviceName(serviceName)
                                 .namespace(namespace)
                                 .build();
+                        Optional<LeaderElectionStatistics> maybeStatistics =
```
Jolyon-S, line 99: Is this the bug that you found? If so, nice.

Jolyon-S (commented):
🏁 00:25 | ⌛ 00:4
I think this is useful - I just have one reservation about the check for one of the log lines.

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
gmaretic: Yup, this is because all the requests that were sent during the leader election will come here, so this is just to filter out excessive logging, as once the leaderId gets updated, the rest will not be logged. Note that we can still race condition into getting more than one log line, but that's fine

lock-api/src/main/java/com/palantir/lock/client/metrics/TimeLockFeedbackBackgroundTask.java
gmaretic: Yeah, the method is not idempotent, so two nodes would get virtually no impact data

lock-api/src/main/java/com/palantir/lock/client/LeaderElectionReportingTimelockService.java
Jolyon-S: That's reasonable.

Jolyon-S (approved):

PR #5180, opened 2021-01-06
[LW] Unify the NoOpLockWatchManager and its underlying cache

jeremyk-91:
**Goals (and why)**:
- The NoOpLockWatchManager should still track updates from the timelock server in terms of its leadership.

- We now create the NoOpLockWatchManager from an already-defined Event Cache, as opposed to creating it with a fresh (and different!) cache. Previously there were 2 event caches when using no-op, one which updates were read from and _another one_ where responses from timelock were processed.

- Nothing in particular.

- The implementation in TestTransactionManagerImpl is janky, though I didn't see a good way to remove it without refactoring the whole class to static factories :(

**Where should we start reviewing?**: NoOpLockWatchManager


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
```diff
@@ -103,7 +103,7 @@ public TestTransactionManagerImpl(
                 metricsManager,
                 createAssertKeyValue(keyValueService, lockService),
                 new LegacyTimelockService(timestampService, lockService, lockClient),
-                NoOpLockWatchManager.create(),
+                NoOpLockWatchManager.create(NoOpLockWatchEventCache.create()),
```
Jolyon-S, line 106: Could you not just factor this out to be a private final field that is instantiated there and then? i.e.
`private final LockWatchEventCache eventCache = NoOpLockWatchEventCache.create();`?

Jolyon-S (commented):
Looks good - just a comment on how to maybe make the janky bit a bit less janky.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/TestTransactionManagerImpl.java
jeremyk-91: In general, you can't reference even private fields before calling a supertype constructor (which is what is happening here).

Jolyon-S: I didn't know that. Fair enough, then - I'll approve.

Jolyon-S (approved):

PR #5181, opened 2021-01-07
Revert "Clock went backwards corruption check (#5131)"

sudiksha27:
**Goals (and why)**:
This reverts commit 259908aa0f1034881f975199f07806b29fce3095.
The check fired on 2 internal stacks. We suspect that these were false positives so we are reverting the check while we investigate the root cause.

🔥


Jolyon-S (approved):
🚀

PR #5182, opened 2021-01-07
Revert revert 5131

sudiksha27:
**Goals (and why)**:
#5131 was reverted in #5181 as there were false positives. in #5131, it was (incorrectly) assumed the timestamps are strictly increasing with increasing sequence numbers. There is a possibility where consecutive seq numbers have the same timestamp. This is possible when a node gains leadership and does not know the value for latest sequence number - n . In this case, it forces agreement for nth sequence on the same value as that of (n-1)th or (n-2)th round (if value for n-1 is not known).
See PaxosTimestampBoundStore#getAgreedState for reference. 

The invariant is now modified to - timestamps should not decrease with increasing sequence numbers.

Added unit tests.
Also, I have validated the invariant on paxos logs from internal dev stack

Corner cases I might have missed?

All of the code is same as in #5131 except the inequality check in LocalTimestampInvariantsVerifier#timestampInvariantsViolationLevel

EOD tomorrow

### DIFF
--- a/changelog/@unreleased/pr-5182.v2.yml
+++ b/changelog/@unreleased/pr-5182.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: TimeLock now validates the expected invariant for its timestamp bounds
+    i.e. the timestamp bound must not decrease with increasing sequence numbers.
+  links:
+  - https://github.com/palantir/atlasdb/pull/5182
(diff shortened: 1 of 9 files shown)


gmaretic (approved):
Looks good!

sudiksha27:
Merge #5183 before getting this in.

PR #5197, opened 2021-01-14
[Cross Client Batching] Autobatchers support configurable wait strats

jeremyk-91:
**Goals (and why)**:
- Minimise the performance overhead imposed by the use of an Autobatcher.

- Expose a way of specifying wait strategy on the builder.

Not much, sadly.

- There's an API break though we explicitly don't guarantee that that API will be preserved
- Are the defaults chosen correctly?

**Where should we start reviewing?**: eh


sudiksha27 (approved):
💯

PR #5198, opened 2021-01-14
[PDS-156719] Filter learner namespace from namespaces

Jolyon-S:
**Goals (and why)**:
Filter the learner namespace out of returned namespaces.

Filter the namespace out.

I haven't tested it, but I should. Suggestions on the correct modification to the test? I'm not all that familiar with how the structure should appear.

Is this correct?!

TimeLockManagementResource

Whenever


jeremyk-91 (commented):
Learner _is_ a valid client name though. I've commented on the ticket - not sure we want to do this actually

Jolyon-S:
Let's not.

PR #5202, opened 2021-01-18
[Cross Client Batching] Broadside LeaderTimeGetter

jeremyk-91:
**Goals (and why)**:
- Make requests to timelock efficiently

- Makes requests to TimeLock for multiple leader times, based on the entire history of namespaces known to this instance of this class. The idea here is that on a stable, busy stack, the overhead of doing this is near zero (since all namespaces have requests), and we can simply re-use the existing set that was used to send the request.

**Testing (What was existing testing like?  What have you done to improve it?)**: Added a couple of tests, including simulations. Also ran the integration tests using Broadside instead of the Legacy LeaderTimeGetter and tests passed (see `arraylist` commit)

- Is the overhead going to be a problem?

**Where should we start reviewing?**: BLPTests

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -155,10 +155,9 @@ import com.palantir.lock.client.BroadsideLeaderPoller;
 import com.palantir.lock.client.BroadsideLeaderTimeGetter;
 import com.palantir.lock.client.InternalMultiClientConjureTimelockService;
 import com.palantir.lock.client.LeaderElectionReportingTimelockService;
-import com.palantir.lock.client.LeaderTimeCoalescingBatcher;
 import com.palantir.lock.client.LeaderTimeGetter;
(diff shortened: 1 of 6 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1206,7 +1206,7 @@ private static LockAndTimestampServices getLockAndTimestampServices(
         LockWatchManagerImpl lockWatchManager = new LockWatchManagerImpl(lockWatchEventCache, lockWatchingService);
 
         LeaderTimeGetter leaderTimeGetter = getLeaderTimeGetter(
-                timelockNamespace, timelockRequestBatcherProviders, serviceProvider, namespacedConjureTimelockService);
+                namespacedConjureTimelockService, timelockNamespace, timelockRequestBatcherProviders, serviceProvider);
```
sudiksha27, line 1210: nit: unnecessary diff?

sudiksha27 (commented):
LGTM! Waiting for prod validation.

jeremyk-91:
At least on internal herby stack, this solution was not good enough.

PR #5224, opened 2021-01-26
Avoid creating thousands of get-ranges threads

carterkozak:
Our metrics show services with thousands of threads in the
`serializabletransactionmanager-get-ranges` pool, however the
executor was not instrumented with Tritium, so it's not clear
how saturated it is, or if it is used at all. Threads are incredibly
expensive, it's generally a sign of failure when a service reaches
1000 total threads.

Using PTExecutors factories we get tracing and execution metrics
for free, as well as resource utilization improvements to share
a slice of an underlying cached executor so threads are only
used as needed. The provided `numThreads` is still an upper limit
for the ExecutorService instance, however idle threads can be
used elsewhere.

The existing queue size warning logic is preserved using a counter
rather than instrumenting the queue itself in much the same way
tritium estimates ExecutorService queue size.

**Goals (and why)**:

Vastly reduce memory overhead for several services.

Use the standard ptexecutors factory with a wrapper to support queue size warnings. Ideally this would move to Hyperion instead, but that's out of scope here.

No behavior change, only a reduction in resource utilization.


gmaretic (approved):
LGTM, thanks!

PR #5225, opened 2021-01-26
[Fix develop] Fix a potential deadlock in SweepQueueReader

carterkozak:
Unsure how this passed baseline originally, but it's blocking
commits on develop.

I've searched internal sources and don't see any references
to the modified member.


gmaretic (approved):
Thanks, this was merged before the baseline PR which was not rebased on develop!

carterkozak:
Thanks @gmaretic! This test failure looks like a flake and I don't have perms to retrigger it, any chance you could hit the button?

Once this goes through I'll rebase #5224

PR #5229, opened 2021-01-27
Suppress deprecation warnings in generated code

pkoenig10:
https://github.com/palantir/atlasdb/pull/5169 deprecated some constants that were used in generated classes. While these generated classes do include `@SuppressWarnings("all")`, that does not suppress javac lints such as `-Xlint:deprecation`. Those warnings need to be suppressed explicitly. For example, see https://github.com/jOOQ/jOOQ/issues/2413.

I've verified locally that this resolves the compilation issues with generated code.


sudiksha27:
We should fix these constants instead of removing because we do intend to remove the deprecated code in future releases. @gmaretic thoughts?

gmaretic (approved):
Thanks for the contribution. I think this is fine for now, and we should just make sure to remove it at the same time as we remove the deprecated constants

PR #5231, opened 2021-02-02
[PDS-158510] Improved Detection of Orphaned Sentinels

jeremyk-91:
**Goals (and why)**:
- PDS-158510: some users were running into issues with orphaned sentinel skip not working.

- In #3610 detection and skipping of sweep sentinels was added. However, there is an edge case: if a transaction T1 reads an orphaned sentinel S, for which an _uncommitted_ T2 has written to the KVS, then T1 doesn't recognise S as orphaned even though the value T2 covered it with is uncommitted.

Added a new group of tests.

Not much. Is perf a concern?

**Where should we start reviewing?**: Probably start with the tests

### DIFF
--- a/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
+++ b/atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
@@ -31,48 +31,6 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import java.math.BigInteger;
-import java.nio.ByteBuffer;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Random;
-import java.util.SortedMap;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CompletionService;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorCompletionService;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
-
-import org.apache.commons.lang3.mutable.MutableInt;
-import org.apache.commons.lang3.mutable.MutableLong;
-import org.apache.commons.lang3.tuple.Pair;
-import org.assertj.core.api.Assertions;
-import org.assertj.core.api.HamcrestCondition;
-import org.hamcrest.Matchers;
-import org.jmock.Expectations;
-import org.jmock.Mockery;
-import org.jmock.Sequence;
-import org.jmock.lib.concurrent.DeterministicScheduler;
-import org.jmock.lib.concurrent.Synchroniser;
-import org.junit.Before;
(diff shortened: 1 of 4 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1438,6 +1443,190 @@ public void checkImmutableTsLockAfterReadsForConservativeIfFlagIsSet() {
...
+                .isInstanceOf(TransactionFailedRetriableException.class)
+                .hasMessageContaining("Tried to read a value that has been deleted.");
+
+        Transaction retryTxn = txManager.createNewTransaction();
+        assertThatCode(() -> retryTxn.get(TABLE_SWEPT_CONSERVATIVE, ImmutableSet.of(TEST_CELL)))
+                .doesNotThrowAnyException();
+    }
+
+    @Test
+    public void getSweepSentinelUnderEarlyUncommittedWriteDoesNotThrow() {
```
gmaretic, line 1473: should swap the names of this and the following test

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1438,6 +1443,190 @@ public void checkImmutableTsLockAfterReadsForConservativeIfFlagIsSet() {
...
+        t2.commit();
+
+        assertThatThrownBy(() -> t1.get(TABLE_SWEPT_CONSERVATIVE, ImmutableSet.of(TEST_CELL, TEST_CELL_2)))
+                .isInstanceOf(TransactionFailedRetriableException.class)
+                .hasMessageContaining("Tried to read a value that has been deleted.");
+    }
+
+    @SuppressWarnings("unchecked") // ArgumentCaptor
+    @Test
+    public void getSentinelValuesStressTest() {
```
gmaretic, line 1565: nice 👍

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1438,6 +1443,190 @@ public void checkImmutableTsLockAfterReadsForConservativeIfFlagIsSet() {
...
+
+        List<Iterable<Long>> stressTestRequests = captor.getAllValues();
+        assertThat(stressTestRequests).hasSize(100);
+        for (int index = 1; index < stressTestRequests.size(); index++) {
+            List<Long> previousTimestamps = StreamSupport.stream(
+                            stressTestRequests.get(index - 1).spliterator(), false)
+                    .collect(Collectors.toList());
+            assertThat(stressTestRequests.get(index)).hasSizeLessThan(previousTimestamps.size());
+        }
+    }
```
gmaretic, line 1616: assert that retry succeeds?

gmaretic (approved):
Paired on most of this.

A few nits. I think perf is fine, took a look at the metrics for encountering naked sentinels and it's not very common. The reads are going to dominate the perf of this either way and there is no way around that.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91: ah, you're right!

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91: Yep, makes sense - added.

PR #5243, opened 2021-02-08
[TS] Calculate exact TS lag and allow configuring per shard lag metrics

gmaretic:
**Goals (and why)**:
Currently, targeted sweep lag metrics are capped at 7 days because calculating the exact value requires an expensive range scan. This is generally fine, but if we fall behind by a large amount, and it takes a long time to catch up, it is rather difficult to gauge progress. This is further complicated by the fact that remediation usually involves increasing the number of shards, which further obscures the situation (are we up to date on the new shards, is there an imbalance and we are waiting for a single shard to catch back up?). This PR addresses both issues:

- Implemented a new method on `KeyValueServicePuncherStore` that efficiently calculates the exact delay
- Added a configuration parameter that lets a user specify strategies for which the lag metrics should be reported per shard (defaults to none)

- The `getMillisForTimestampWithinBounds` implementation is rather involved. Generally, it first finds some lower bound for the result by reading from the puncher store at exponentially increasing offsets from current time. Then, it uses a modified binary search to reduce the range to range scan through to less than 4 hours in real time. The search assumes that the real time millis and timestamps have a linear dependency to estimate the next test value instead of the standard midpoint.
- The metrics implementation for the min metric is similar to before, except now the metric is tagged by a shard of -1 (no good way around it with generated metrics). We now expose a way to get the last swept timestamp per shard, and track the last `MillisAndMaybeTimestamp` for each shard. This allows us to just recalculate the lag metric every time per shard as the punch table entry will either be the same, where we skip kvs operations, or very close -- which is a small range scan.
- The rest is tests, fixing some sketchy tests from before, and the filtering logic

Tested out the new methods on `KeyValueServicePuncherStore`, though some of the tests could not have reasonable unit tests without leaking too many implementation details, so opted fora more fuzzy approach. Ran the fuzz tests a few hundred times locally to make sure they are not flaking.
Furthermore, added some metrics tests to verify we can indeed track per shard, or continue using the default metrics.

Unfortunately, this is not a trivial PR, so I might have missed something, though the test coverage is pretty good.

- The new method can return odd results in the presence of significant clock skew in the punch table, but arguably in that case all bets are off, and it should still generally return the same result as the old method if the skew is transient.
- The default metric will now be tagged by shard -1, which is hacky, but didn't really see a good way around it. Maybe we can create a new metric for the per shard case and then use the filter to filter out the one that is not in use?
- Number of shards for the per shard metrics is taken from the runtime config and fixed at metric creation time; this will not pick up newly created shards and might be incorrect if somehow the config gets out of sync with the persisted value. Neither should be a big concern, but flagging.

`getMillisForTimestampWithinBounds`, then `TargetedSweepMetrics`

I appreciate there are a lot of fires right now, and this PR is nontrivial, but it would be great to be able to roll it out before the weekend.


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStoreTest.java
```diff
@@ -120,66 +138,217 @@ public void returnsTimestampAssociatedWithGreatestPunchedTimeEvenIfItIsNotGreate
...
+        long maxOffset = KeyValueServicePuncherStore.MAX_RANGE_SCAN_SIZE * LongMath.pow(2, maxReads - 1);
+
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        for (int i = 0; i < 100; i++) {
+            puncherStore.put(random.nextLong(10_000L), random.nextLong(maxOffset));
+        }
+
+        MillisAndMaybeTimestamp result = KeyValueServicePuncherStore.findOlder(kvs, 1_000L, maxOffset);
+        assertThat(result.millis()).isLessThanOrEqualTo(maxOffset);
+        assertThat(result).satisfies(res -> res.timestampSatisfies(ts -> ts <= 1_000L));
```
jeremyk-91, line 227: this could fail maybe around one in 10^10 times? not sure, if your random.nextLong(...) generates numbers such that you don't see anything smaller than 1000. In any case we should allow timestamp being absent anyway.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStore.java
```diff
@@ -160,20 +162,223 @@ public static long getMillisForTimestamp(KeyValueService kvs, long timestamp) {
...
+            MillisAndMaybeTimestamp betweenBounds = getOlder(kvs, testValue);
+
+            if (betweenBounds.timestampSatisfies(ts -> ts == timestamp)) {
+                return noRangeScan(betweenBounds);
+            } else if (betweenBounds.timestampSatisfies(ts -> ts > timestamp)) {
+                upperBound = betweenBounds;
+            } else {
+                millisLowEstimate = Optional.of(betweenBounds);
+                lowerBound = ImmutableMillisAndMaybeTimestamp.builder()
+                        .from(betweenBounds)
```
jeremyk-91, line 321: This doesn't really save us any code 😅  just get the timestamp from betweenBounds?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStoreTest.java
```diff
@@ -120,66 +138,217 @@ public void returnsTimestampAssociatedWithGreatestPunchedTimeEvenIfItIsNotGreate
...
+        tsToMillis.forEach(puncherStore::put);
+
+        Collection<Long> millisCandidates = tsToMillis.get(queryTimestamp);
+
+        Optional<MillisAndMaybeTimestamp> actual = KeyValueServicePuncherStore.getMillisForTimestampWithinBounds(
+                kvs, queryTimestamp, null, MAX_OFFSET_FOR_BOUNDS_TESTS);
+
+        assertThat(actual).isPresent();
+        assertThat(millisCandidates).contains(actual.get().millis());
+        assertThat(actual.get().timestamp()).hasValue(queryTimestamp);
```
jeremyk-91, line 285: what happens if we have more than one match for the timestamp?

jeremyk-91 (approved):
🕐 0:55 🏁 ????

Pretty close, we paired on this - thanks for the explanation! Almost ready to go

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/metrics/TargetedSweepMetrics.java
```diff
@@ -208,49 +218,96 @@ private void registerProgressMetrics(String strategy, TargetedSweepMetricPublica
...
+                        .<Gauge<Long>>map(shard -> () -> estimateMillisSinceTs(
+                                lastSweptTsWithShard.getLastValueForKey(shard), wallClock, tsToMillis))
+                        .collectToMap();
+            } else {
+                Supplier<Long> millisSinceLastSweptTs = new CachedComposedSupplier<>(
+                        shardAndTs -> estimateMillisSinceTs(shardAndTs, wallClock, tsToMillis),
+                        lastSweptTsWithShard::getVersionedValue,
+                        metricsConfiguration.millisBetweenRecomputingMetrics(),
+                        wallClock);
+                return ImmutableMap.of(-1, millisSinceLastSweptTs::get);
```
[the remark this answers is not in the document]
gmaretic: Oh, that's an excellent point!

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/cleaner/KeyValueServicePuncherStoreTest.java
gmaretic: fixed

PR #5245, opened 2021-02-10
[WIP] Circle speedup

anjor:
**Goals (and why)**:
To speed up CI.

Splits up the stress tests for timelock server into a separate class, and give them their own container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           paths: [ . ]
 
   test:
-    parallelism: 8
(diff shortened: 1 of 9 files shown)


jeremyk-91 (approved):
🕐 0:15 🏁 5:21 🔥 

This PR looks correct in that it preserves test coverage. I've removed Merge when Ready because the changelog is in a weird state: please fix and then merge.

Build times look down to around 8 minutes for the test phase; further improvements may be more difficult since we don't have as clear of a bottleneck any more.

PR #5247, opened 2021-02-11
[PDS-146088] Log and metrics for a few more edge cases

Jolyon-S:
**Goals (and why)**:
* We log when the evictor is considering but _not_ evicting, but it could actually be trying but failing to evict.
* `EvictionTimer` has a static, singleton executor service. If this dies somehow, then no eviction tasks will be run.

* Log even more on all the edge cases
* Added a metric for the number of tasks in the eviction executor.

N/A

* Is this metric expensive?
* Is there a better way of exposing the method from `EvictionTimer`?
* Does this approach of exposing the method require us to include a license / documentation?

`CassandraClientPoolingContainer`

Sooner is better.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -43,6 +43,7 @@ import org.apache.commons.pool2.PooledObject;
 import org.apache.commons.pool2.impl.DefaultEvictionPolicy;
 import org.apache.commons.pool2.impl.EvictionConfig;
 import org.apache.commons.pool2.impl.EvictionPolicy;
+import org.apache.commons.pool2.impl.ExposedEvictionTimer;
 import org.apache.commons.pool2.impl.GenericObjectPool;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import org.apache.thrift.transport.TFramedTransport;
@@ -324,6 +325,7 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
         registerPoolMetric(CassandraClientPoolHostLevelMetric.NUM_ACTIVE, () -> (long) pool.getNumActive());
         registerPoolMetric(CassandraClientPoolHostLevelMetric.CREATED, pool::getCreatedCount);
         registerPoolMetric(CassandraClientPoolHostLevelMetric.DESTROYED_BY_EVICTOR, pool::getDestroyedByEvictorCount);
+        registerPoolMetric(CassandraClientPoolHostLevelMetric.EVICTOR_TASK_SIZE, ExposedEvictionTimer::getNumTasks);
     }
 
     private void registerPoolMetric(CassandraClientPoolHostLevelMetric metric, Gauge<Long> gauge) {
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -341,13 +343,22 @@ private NonEvictionLoggingEvictionPolicy(EvictionPolicy<T> delegate) {
...
-                log.debug(
-                        "Considered an object to be evicted from the Cassandra client pool, but did not evict it",
-                        SafeArg.of("underTestState", underTest.getState()),
-                        SafeArg.of("idleState", underTest.getIdleTimeMillis()),
-                        SafeArg.of("idleCount", idleCount),
-                        SafeArg.of("evictionConfig", config));
+            if (log.isDebugEnabled()) {
+                if (delegateResult) {
+                    log.debug(
+                            "Attempting to evict an object from the Cassandra client pool",
```
Jolyon-S, line 349: This could be happening on every run, but it fails to evict, so I thought it was worth logging (we've actually validated that the other log line is _not_ logged).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolHostLevelMetric.java
```diff
@@ -21,7 +21,8 @@
     NUM_IDLE("numIdle", 0.1, 2.0),
     NUM_ACTIVE("numActive", 0.1, 2.0),
     CREATED("created", 0.01, 2.0),
-    DESTROYED_BY_EVICTOR("destroyedByEvictor", 0.01, 2.0);
+    DESTROYED_BY_EVICTOR("destroyedByEvictor", 0.01, 2.0),
+    EVICTOR_TASK_SIZE("evictorTaskSize", 0.0, 2.0);
```
Jolyon-S, line 25: I don't actually know how to configure this properly - the code was really dense. Would welcome some insight into the best way to approach this (and if we can avoid registering this on all deployments).

atlasdb-cassandra/src/main/java/org/apache/commons/pool2/impl/ExposedEvictionTimer.java
```diff
@@ -0,0 +1,31 @@
...
+ * the internal investigation into the missing eviction problem is resolved.
+ */
+public final class ExposedEvictionTimer {
+    private ExposedEvictionTimer() {
+        // no-op
+    }
+
+    public static long getNumTasks() {
+        return EvictionTimer.getNumTasks();
+    }
```
Jolyon-S, line 30: This is pretty sneaky.

Since we're in the apache package, do we need a license?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
jeremyk-91: I'm ok with stacks that are turning into a fire pit to turn into _more_ of a fire pit here anyway.

jeremyk-91 (commented):
🕐  0:14 🏁  1:52

Broadly looks right, some questions before we merge

atlasdb-cassandra/src/main/java/org/apache/commons/pool2/impl/ExposedEvictionTimer.java
Jolyon-S: Perfect.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraClientPoolHostLevelMetric.java
Jolyon-S: ah right, I'll do that.

jeremyk-91 (approved):
🚀

PR #5249, opened 2021-02-11
allow defining cassandra server hosts in runtime config

pnepywoda:
**Goals (and why)**:
Because cassandra server hosts can dynamically change, we should allow defining them in runtime config.

* If there are more than 1 host defined in the install config, we prefer that over runtime (to not break backcompat)

* unit tested

today (many recent tickets/paging)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
```diff
@@ -45,6 +45,11 @@ public String type() {
         return TYPE;
     }
 
+    @Value.Default
+    public CassandraServersConfigs.CassandraServersConfig servers() {
+        return ImmutableDefaultConfig.builder().build();
```
pkoenig10, line 50: nit: consider adding `singleton = true` to the `@Value.Immutable` annotation on `DefaultConfig`
```suggestion
        return ImmutableDefaultConfig.of();
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -39,7 +39,10 @@ public CassandraReloadableKvsConfig(
 
     @Override
     public CassandraServersConfigs.CassandraServersConfig servers() {
-        return config.servers();
+        // get servers from install config (for backcompat), otherwise get servers from runtime config
+        return config.servers().numberOfThriftHosts() > 0
+                ? config.servers()
+                : chooseConfig(CassandraKeyValueServiceRuntimeConfig::servers, null);
```
pkoenig10, line 45: nit: Can we just match other uses of `chooseConfig` to make this less confusing?
```suggestion
        if (config.servers().numberOfThriftHosts() > 0) {
            return config.servers();
        }
        return chooseConfig(CassandraKeyValueServiceRuntimeConfig::servers, config.servers());
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
pnepywoda: `of` doesn't exist

pkoenig10:
I'd suggest we also provide a default value for `servers` in the install configuration so services can just remove it entirely and don't have to provide a `servers: []`.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
pnepywoda: no - the other uses of `chooseConfig` prefer runtime over install

pkoenig10: Right, but we know `config.servers()` is empty at this point

pkoenig10 (approved):

changelog/@unreleased/pr-5249.v2.yml
```diff
@@ -0,0 +1,22 @@
+type: improvement
+improvement:
+  description: |-
+    allow defining cassandra server hosts in runtime config
```
Jolyon-S, line 4: you should fix this changelog

jeremyk-91 (commented):
🕐  0:35 🏁  🔥 

This makes sense in principle. Couple of small things to fix; also, would be good to switch one of the ETE tests to use the runtime config just to verify that it is actually able to work

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -39,7 +39,11 @@ public CassandraReloadableKvsConfig(
...
     public CassandraServersConfigs.CassandraServersConfig servers() {
-        return config.servers();
+        // get servers from install config (for backcompat), otherwise get servers from runtime config
+        if (config.servers().numberOfThriftHosts() > 0) {
+            return config.servers();
+        }
+        return chooseConfig(CassandraKeyValueServiceRuntimeConfig::servers, config.servers());
     }
 
     @Override
```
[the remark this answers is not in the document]
pnepywoda: `CassandraKeyValueServiceRuntimeConfig` doesn't have `concurrentGetRangesThreadPoolSize`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
```diff
@@ -55,7 +55,10 @@
      * This value, or values derived from it (e.g. the number of Thrift hosts) must ONLY be used on KVS initialization
      * to generate the initial connection(s) to the cluster, or as part of startup checks.
      */
-    CassandraServersConfigs.CassandraServersConfig servers();
+    @Value.Default
+    default CassandraServersConfigs.CassandraServersConfig servers() {
+        return ImmutableDefaultConfig.of();
```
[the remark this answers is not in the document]
pkoenig10: Can we update the validation instead? Ultimately we need the install configuration servers to be empty for this to work.

pnepywoda:
Do you think this warrants an ETE test? If so, where's an example of different-config ete tests we already do?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java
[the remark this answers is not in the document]
pnepywoda: I updated the reloadable merging config to assert there's >0 servers

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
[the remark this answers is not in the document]
pnepywoda: The point of this PR is to allow defining "discoverable" things in runtime config. Only the hosts can automatically change on us so this is a targeted fix for `servers`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -256,8 +258,9 @@ public boolean usingSsl() {
...
     public void check() {
-        //         The config instance passed here gets checked at the time its of construction
-        //         (default implementation in the parent class). Hence, it is okay to do no operation here.
+        Preconditions.checkState(
```
pkoenig10, line 261: This method is never actually called. It is called by `ImmutableCassandraKeyValueServiceConfig` and we have to override it here, but we don't call it when creating a `CassandraReloadableKvsConfig` because it's not an immutable class.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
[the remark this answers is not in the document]
pkoenig10: Yes but the correct value of `concurrentGetRangesThreadPoolSize` depends on the number of servers.
https://github.com/palantir/atlasdb/blob/fbab4fc3d9e55e0fb7a651c52a15314a96766f35/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceConfig.java#L349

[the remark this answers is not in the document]
pnepywoda: I'd also have to copy `poolSize`. Isn't this getting too complex? I just want `servers` to be reloadable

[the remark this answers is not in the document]
pkoenig10: I don't think this is optional. If you don't copy it, `concurrentGetRangesThreadPoolSize` will be 0 which is almost certainly incorrect.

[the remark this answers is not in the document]
pkoenig10: A potentially better alternative is creating an abstract `DelegatingCassandraKeyValueServiceConfig` that delegates to the install configuration and then only override the methods for which there is a runtime configuration. That way you can pick up the default implementations in `CassandraKeyValueServiceConfig` for free.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
```diff
@@ -84,7 +84,10 @@ static CassandraKeyValueServiceConfig preprocessKvsConfig(
         CassandraKeyValueServiceConfig configWithNamespace =
                 CassandraKeyValueServiceConfigs.copyWithKeyspace(cassandraConfig, desiredKeyspace);
 
-        return new CassandraReloadableKvsConfig(configWithNamespace, runtimeConfig);
+        CassandraReloadableKvsConfig reloadableKvsConfig =
+                new CassandraReloadableKvsConfig(configWithNamespace, runtimeConfig);
+        reloadableKvsConfig.check();
```
pkoenig10, line 89: This method should not be called in callers - it should not be their respsonsibility. If you want to keep it, then it should be called in the constructor.

However I'd strongly prefer to just remove it. It's doesn't make sense to have a precondition method on this non-immutable class since updates to the underlying values won't be checked (unlike runtime configuration classes which are instantiated every time new config is loaded).

If we want to have a check method, it belongs on the `CassandraKeyValueServiceRuntimeConfig` class.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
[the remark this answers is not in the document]
jeremyk-91: Yeah, afraid we have to copy it or wire things around - users of `concurrentGetRangesThreadPoolSize` will be sad as @pkoenig10 notes.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java
jeremyk-91: We can't really check Runtime either though right? Because people could still have it on Install.

I think when I suggested moving the check what I had in mind was doing it in the constructor, though you have a good point that updates are not checked.

Fine with removing the checks at creation time. We might want `servers()` to throw if it's going to return an empty, as I don't imagine users would expect that case.

pkoenig10: Are we not worried about runtime errors in the places that call `servers()`?

pkoenig10: Maybe it's worth adding an `InvalidConfig` variant to `CassandraServersConfig` to force callers to handle this case.

Most uses are `ThriftHostsExtractingVisitor`. We can either:
- Change the return type to `Optional<Set<InetSocketAddress>>` and let callers handle an empty optional
- Just throw in `ThriftHostsExtractingVisitor`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -39,7 +42,19 @@ public CassandraReloadableKvsConfig(
 
     @Override
     public CassandraServersConfigs.CassandraServersConfig servers() {
-        return config.servers();
+        CassandraServersConfig servers = chooseServers();
+        Preconditions.checkState(
```
pnepywoda, line 46: @jeremyk-91 is it ok to throw at runtime here (and for `concurrentGetRangesThreadPoolSize` and `poolsize`) ?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
```diff
@@ -94,6 +99,11 @@ public Integer sweepReadThreads() {
         return AtlasDbConstants.DEFAULT_SWEEP_CASSANDRA_READ_THREADS;
     }
 
+    @Value.Default
+    public int poolSize() {
```
pkoenig10, line 103: I don't think we want this to be live-reloadable?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -39,7 +42,19 @@ public CassandraReloadableKvsConfig(
...
+                !servers.accept(new ThriftHostsExtractingVisitor()).isEmpty(),
+                "'servers' must have at least one defined host");
+        return servers;
+    }
+
+    private CassandraServersConfig chooseServers() {
```
pkoenig10, line 52: This method is only called in one place, just inline it

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -252,8 +267,7 @@ public boolean usingSsl() {
 
     @Override
     public void check() {
-        //         The config instance passed here gets checked at the time its of construction
-        //         (default implementation in the parent class). Hence, it is okay to do no operation here.
+        // this class is not an immutables so this method is never called
```
pkoenig10, line 270: Can we just remove this override here?

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceConfigTest.java
```diff
@@ -73,4 +73,13 @@ public void usingSslIfSslParamNotPresentAndSslConfigurationPresent() {
     public void notUsingSslIfSslParamNotPresentAndSslConfigurationNotPresent() {
         assertThat(CASSANDRA_CONFIG.usingSsl()).isFalse();
     }
+
+    @Test
+    public void checkMinimalConfig() {
```
pkoenig10, line 78: What's the purpose of this test?

pnepywoda: we now don't have to provide `servers`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
pnepywoda: `concurrentGetRangesThreadPoolSize` reads it in `poolSize() * servers().numberOfThriftHosts()`. 

should `servers` be reloadable but not `poolSize` then? (I dunno how weird that is)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -42,21 +42,19 @@ public CassandraReloadableKvsConfig(
...
     @Override
     public CassandraServersConfigs.CassandraServersConfig servers() {
-        CassandraServersConfig servers = chooseServers();
+        CassandraServersConfig servers;
+        // get servers from install config (for backcompat), otherwise get servers from runtime config
+        if (config.servers().numberOfThriftHosts() > 0) {
+            servers = config.servers();
+        } else {
+            servers = chooseConfig(CassandraKeyValueServiceRuntimeConfig::servers, config.servers());
+        }
```
pkoenig10, line 51:
```suggestion
        // get servers from install config (for backcompat), otherwise get servers from runtime config
        if (config.servers().numberOfThriftHosts() > 0) {
            return config.servers();
        }

        CassandraServersConfig servers = chooseConfig(CassandraKeyValueServiceRuntimeConfig::servers, config.servers());
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -237,7 +252,7 @@ public String type() {
 
     @Override
     public int concurrentGetRangesThreadPoolSize() {
-        return config.concurrentGetRangesThreadPoolSize();
+        return poolSize() * servers().numberOfThriftHosts();
```
pkoenig10, line 253: This is going to drop overrides of `concurrentGetRangesThreadPoolSize` in install configuration.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -237,7 +244,14 @@ public String type() {
 
     @Override
     public int concurrentGetRangesThreadPoolSize() {
-        return config.concurrentGetRangesThreadPoolSize();
+        return chooseConfig(
```
pkoenig10, line 247: This still isn't quite right. We may use install `servers` but runtime `concurrentGetRangesThreadPoolSize`/`pool`. 

For example, if an existing user took this upgrade they would have a `concurrentGetRangesThreadPoolSize` value of 0.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -237,7 +244,20 @@ public String type() {
 
     @Override
     public int concurrentGetRangesThreadPoolSize() {
-        return config.concurrentGetRangesThreadPoolSize();
+        // get from install config (to allow overriding in install when runtime is left as default),
+        // otherwise get from runtime config
+        if (config.concurrentGetRangesThreadPoolSize() > 0) {
```
pkoenig10, line 249: This relies on the fact that `concurrentGetRangesThreadPoolSize` will be 0 if `servers` is empty. We should probably be checking the same condition in `servers` and here.

pkoenig10: Although maybe this is okay?

pnepywoda: 0 is the "default" value. granted it's computed from other defaults, but it's still the default thing I'm comparing to to see if it's an override

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
jeremyk-91: yeah, afraid this one should probably stay in install: we don't actually allow the size of the connection pools to be adjusted at runtime I think, so having this here is a bit misleading.

ugh, the multiplication logic should live in the Reloadable one then.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
jeremyk-91: Hmm, yeah it's not great. Looking at this I think the ideal behaviour is to do what we normally do for runtime configs internally:

- on startup, get a runtime config. If it means the state of the world is bad, explode (throw). Otherwise save the last seen good runtime config in a field here.
- thereafter, when this is called, get a runtime config; if it's consistent then update the last seen one, and if it means the state of the world is inconsistent, log a WARN, reject the config and use the previous value of the runtime config.

The main alternatives would involve
- throw (as here): probably not the worst, but I don't know if this is fully safe.
- just return empty: this will break a lot of the assumptions the c* code makes
- @pkoenig10's suggestion of returning a failed type: this is good, though with a lot of the places where this is used I don't know what to do except re-throw, and I don't know if that is safe.

Apologies this has turned out to be a fair bit more complex than initially scoped - if you'd prefer someone on the team writes some of the code once we've got a solution we like, that can be arranged.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
jeremyk-91: in principle we should be defensive to overrides that _are_ zero, but that isn't really meaningful for this anyway so I think I'm ok with this.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
pkoenig10: @jeremyk-91 Thoughts on just using `Refreshable` here? Then we get all this behavior "for free" without having to reimplement the logic to apply updates, save the last good value, etc.

This would also allow us to keep the server size check, because we would construct a new `ImmutableCassandraKeyValueServiceConfig` on each runtime config update.

We would just need to change the interface of `AtlasDbFactory` to accept a `Refreshable` instead of a `Supplier`.
https://github.com/palantir/atlasdb/blob/fd73863df0d8d65eeaa6ed5801a1b6ef8578199c/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java#L58

Fortunately, this should be fairly easy because we're already using `Refreshable` as an argument here.
https://github.com/palantir/atlasdb/blob/fd73863df0d8d65eeaa6ed5801a1b6ef8578199c/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java#L402

jeremyk-91: sourcegraph search shows usages in AtlasDB Proxy and Backup product only, so I'm fine with this change.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
pnepywoda: then we also have to remove `concurrentGetRangesThreadPoolSize` from runtimeconfig since it needs `poolSize` to compute

pnepywoda: unless we set poolsize to a default for the multiplication.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/DelegatingCassandraKeyValueServiceConfig.java
```diff
@@ -15,259 +15,231 @@
...
-    private final Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfigSupplier;
+public final class DelegatingCassandraKeyValueServiceConfig implements CassandraKeyValueServiceConfig {
 
-    public CassandraReloadableKvsConfig(
-            CassandraKeyValueServiceConfig config, Supplier<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig) {
-        this.config = config;
-        this.runtimeConfigSupplier = runtimeConfig;
+    private final Supplier<CassandraKeyValueServiceConfig> delegate;
+
+    public DelegatingCassandraKeyValueServiceConfig(Supplier<CassandraKeyValueServiceConfig> delegate) {
```
pkoenig10, line 32: Typical delegating patterns (at least in AtlasDB) are an abstract class with a `delegate` method

pnepywoda: like what? all I'm seeing is similar to `DelegatingManagedTimestampService` which isn't abstract

pkoenig10: We should just use `@AutoDelegate` to generate the delegate implementation for us.

pnepywoda: `@AutoDelegate` requires us to put `@DoDelegate` on all 30 methods

Looks like atlasdb calls classes "Forwarding*" and yes uses `abstract delegate()`

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraKeyValueServiceRuntimeConfig.java
pkoenig10: > we don't actually allow the size of the connection pools to be adjusted at runtime I think

We also don't let `concurrentGetRangesThreadPoolSize` be adjusted as runtime. It's fetched once when creating the `TransactionManager`.

jeremyk-91 (commented):
Thanks for the detailed implementation! I think this is close, just one last thing - I'm uncomfortable with having user-configurable runtime config that doesn't actually respond to runtime changes...

pkoenig10:
I took a stab at this in https://github.com/palantir/atlasdb/pull/5272. I tried to keep the implementation as simple as possible while addressing all of the concerns mentioned here.

pnepywoda:
prefer https://github.com/palantir/atlasdb/pull/5272

PR #5250, opened 2021-02-12
Stop spamming logs when local host filtering fails and the weight is 0

Jolyon-S:
**Goals (and why)**:
* We attempt to find local hosts even if the weighting is 0; this is a waste of time and spams logs.

* No longer attempt to find local hosts if the local host weighting is 0.

* Had to expose a few things in CassandraService to let me test this via some mocks, but now we can verify that the supplier is or is not called.

* This should not affect the majority of uses that are using this feature. A sanity check is always good here.

`CassandraService`

Sooner rather than later.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -87,6 +87,23 @@ public class CassandraService implements AutoCloseable {
 
     private final Random random = new Random();
 
+    @VisibleForTesting
+    public CassandraService(
+            MetricsManager metricsManager,
+            CassandraKeyValueServiceConfig config,
+            Blacklist blacklist,
+            CassandraClientPoolMetrics poolMetrics,
+            Supplier<Optional<HostLocation>> myLocationSupplier) {
+        this.metricsManager = metricsManager;
+        this.config = config;
+        this.myLocationSupplier = myLocationSupplier;
+        this.blacklist = blacklist;
+        this.poolMetrics = poolMetrics;
+
+        Supplier<Map<String, String>> hostnamesByIpSupplier = new HostnamesByIpSupplier(this::getRandomGoodHost);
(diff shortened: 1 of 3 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -87,6 +87,23 @@
 
     private final Random random = new Random();
 
+    @VisibleForTesting
+    public CassandraService(
```
Jolyon-S, line 91: I can't make the other constructor call this one because it uses `this::getSnitch` in the construction of the location supplier.

j-baker:
So, I don't think we should need to config override whenever we deploy a service in K8s. Atlas should be able to detect that it can't hit the AWS metadata endpoint and stop trying.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
jeremyk-91: This isn't used internally (I checked internal code search), so I'd suggest breaking the API, this is kind of weird.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -179,7 +196,12 @@ private String getSnitch() {
         }
     }
 
-    private Set<InetSocketAddress> refreshLocalHosts(List<TokenRange> tokenRanges) {
+    @VisibleForTesting
+    Set<InetSocketAddress> refreshLocalHosts(List<TokenRange> tokenRanges) {
+        if (config.localHostWeighting() == 0) {
+            return ImmutableSet.of();
+        }
```
jeremyk-91, line 203: can/should we automatically switch this off if we have repeated failures?

jeremyk-91 (commented):
🕐 0:18 🏁 a lot

Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/5252

PR #5252, opened 2021-02-12
Disable local host weighting after failing 5 times

Jolyon-S:
An improved version of https://github.com/palantir/atlasdb/pull/5250.

**Goals (and why)**:
* Stop spamming logs when attempting but failing to get the local hosts.

* Implemented a wrapper around supplier that stops calling the delegate after receiving N exceptions in a row.

* Added some tests to `HostLocationSupplierTest`, added some new tests for the new class.

* I assume that 5 logs being logged is acceptable?
* Perhaps I was trigger happy with using `create` over a constructor - not sure on thoughts here.

ExceptionHandlingSupplier

Sooner is better, but Monday is probably acceptable too.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/ExceptionHandlingSupplier.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/ExceptionHandlingSupplier.java
@@ -40,7 +40,7 @@ final class ExceptionHandlingSupplier<T> implements Supplier<Optional<T>> {
 
     @Override
     public Optional<T> get() {
-        if (numFailures.get() >= numRetries) {
+        if (numFailures.get() > numRetries) {
             return Optional.empty();
         }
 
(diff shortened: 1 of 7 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/ExceptionHandlingSupplier.java
```diff
@@ -0,0 +1,56 @@
...
+        numFailures = new AtomicInteger(0);
+    }
+
+    public static <T> ExceptionHandlingSupplier<T> create(Supplier<T> delegate, int numRetries) {
+        return new ExceptionHandlingSupplier<>(delegate, numRetries);
+    }
+
+    @Override
+    public Optional<T> get() {
+        if (numFailures.get() >= numRetries) {
```
jeremyk-91, line 43: nit: strictly speaking this should be `>` given you've named the param `numRetries`

jeremyk-91 (approved):
🕐  0:13 🏁 0:36

One thing to fix, but broadly looks good.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/ExceptionHandlingSupplier.java
Jolyon-S: ha, good catch

PR #5261, opened 2021-02-16
[Cross Client Batching?] Liveness Checkers for Jepsen

jeremyk-91:
**Goals (and why)**:
- returning 0 successful events is by definition not an inconsistent analysis, but is pretty useless in giving confidence that our jepsen tests are verifying anything.

I added Jepsen verifiers that introduce the following rules:
- Timestamps: a timestamp test needs at least 1 timestamp to be returned in order to pass.
- Locks: a lock test needs at least 1 lock request to have a lock token returned in order to pass.

This PR is about testing.

- Is this too heavy-handed?
- Some irritating duplication, though I tried to factor most of it out into a common LivenessChecker.

**Where should we start reviewing?**: LivenessChecker

**Priority (whenever / two weeks / yesterday)**: eh, this week. Not the highest.

### DIFF
--- a/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/events/RequestType.java
+++ b/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/events/RequestType.java
@@ -16,7 +16,9 @@
 package com.palantir.atlasdb.jepsen.events;
 
 public abstract class RequestType {
-    public static final String TIMESTAMP = "timestamp";
+    // This is the actual type used in Jepsen for timestamp requests.
+    public static final String TIMESTAMP = "read-operation";
+
     public static final String LOCK = "lock";
     public static final String REFRESH = "refresh";
     public static final String UNLOCK = "unlock";
--- a/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/events/RequestType.java
+++ b/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/events/RequestType.java
@@ -17,7 +17,7 @@ package com.palantir.atlasdb.jepsen.events;
 
 public abstract class RequestType {
     // This is the actual type used in Jepsen for timestamp requests.
-    public static final String TIMESTAMP = "read-operation";
+    public static final String TIMESTAMP = "timestamp";
 
     public static final String LOCK = "lock";
     public static final String REFRESH = "refresh";
--- a/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/lock/LockAcquisitionLivenessChecker.java
+++ b/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/lock/LockAcquisitionLivenessChecker.java
@@ -19,6 +19,7 @@ package com.palantir.atlasdb.jepsen.lock;
 import com.palantir.atlasdb.jepsen.CheckerResult;
 import com.palantir.atlasdb.jepsen.events.Checker;
 import com.palantir.atlasdb.jepsen.events.Event;
+import com.palantir.atlasdb.jepsen.events.RequestType;
 import com.palantir.atlasdb.jepsen.utils.LivenessChecker;
 import java.util.List;
 import java.util.Objects;
@@ -27,7 +28,7 @@ public class LockAcquisitionLivenessChecker implements Checker {
     private final LivenessChecker delegate;
 
     public LockAcquisitionLivenessChecker() {
-        delegate = new LivenessChecker(okEvent -> okEvent.function().equals("lock")
+        delegate = new LivenessChecker(okEvent -> okEvent.function().equals(RequestType.LOCK)
                 && !Objects.requireNonNull(okEvent.value()).isEmpty());
     }
 
--- a/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/lock/LockAcquisitionLivenessChecker.java
+++ b/atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/lock/LockAcquisitionLivenessChecker.java
@@ -27,9 +27,8 @@ public class LockAcquisitionLivenessChecker implements Checker {
     private final LivenessChecker delegate;
 
     public LockAcquisitionLivenessChecker() {
-        delegate =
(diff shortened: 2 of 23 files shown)


atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/utils/LivenessChecker.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        public List<Event> errors() {
+            return valid()
+                    ? ImmutableList.of()
+                    : ImmutableList.of(ImmutableInfoEvent.builder()
+                            .time(lastSeenEventTimestamp)
+                            .value("No live requests were actually observed up to this time, which is worrying as to "
+                                    + "the "
+                                    + "validity of this test.")
```
jeremyk-91, line 97: spotless didn't fix this! will fix

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/utils/LivenessChecker.java
```diff
@@ -0,0 +1,107 @@
...
+    private final Predicate<OkEvent> livenessJudgmentHandler;
+
+    public LivenessChecker(Predicate<OkEvent> livenessJudgmentHandler) {
+        this.livenessJudgmentHandler = livenessJudgmentHandler;
+    }
+
+    @Override
+    public CheckerResult check(List<Event> events) {
+        Visitor visitor = new Visitor(livenessJudgmentHandler);
+        events.forEach(event -> event.accept(visitor));
```
gmaretic, line 43: If you're already doing this, why don't we make the visitor generic ("corrected visitor pattern") so you can return true when you want to stop and not have to go through the entire history when you probably know very quickly liveness is satisfied. Could also choose not to process entire history when we find a correctness case, but probably don't want to do that

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/events/RequestType.java
```diff
@@ -16,7 +16,9 @@
 package com.palantir.atlasdb.jepsen.events;
 
 public abstract class RequestType {
-    public static final String TIMESTAMP = "timestamp";
+    // This is the actual type used in Jepsen for timestamp requests.
```
gmaretic, line 19: Why don't we just change the method name in `timestamp.clj` instead?

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/lock/LockAcquisitionLivenessChecker.java
```diff
@@ -0,0 +1,38 @@
...
+import com.palantir.atlasdb.jepsen.events.Event;
+import com.palantir.atlasdb.jepsen.utils.LivenessChecker;
+import java.util.List;
+import java.util.Objects;
+
+public class LockAcquisitionLivenessChecker implements Checker {
+    private final LivenessChecker delegate;
+
+    public LockAcquisitionLivenessChecker() {
+        delegate = new LivenessChecker(okEvent -> okEvent.function().equals("lock")
```
gmaretic, line 30: Use constant from `RequestType.java` which should by the way really be an enum, but I'm fine not making that change

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/timestamp/TimestampLivenessChecker.java
```diff
@@ -0,0 +1,41 @@
...
+import com.palantir.atlasdb.jepsen.events.Checker;
+import com.palantir.atlasdb.jepsen.events.Event;
+import com.palantir.atlasdb.jepsen.utils.LivenessChecker;
+import java.util.List;
+
+/**
+ * Verifies that the number of timestamp requests that succeeded was at least 1.
+ */
+public class TimestampLivenessChecker implements Checker {
+    private static final String READ_OPERATION = "read-operation";
```
gmaretic, line 29: Use constant from `RequestType.java `

gmaretic (approved):
A few remaining nits, but gtg otherwise

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/utils/LivenessChecker.java
sudiksha27: This is much better!

atlasdb-jepsen-tests/src/test/java/com/palantir/atlasdb/jepsen/timestamp/TimestampLivenessCheckerTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static void assertResultValidAndErrorFree(CheckerResult result) {
+        assertThat(result.valid()).isTrue();
+        assertThat(result.errors()).isEmpty();
+    }
+
+    private static CheckerResult runTimestampLivenessChecker(Event... events) {
+        TimestampLivenessChecker livenessChecker = new TimestampLivenessChecker();
+        return livenessChecker.check(ImmutableList.copyOf(events));
+    }
```
sudiksha27, line 97: Good test coverage! 🏅

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/utils/LivenessChecker.java
```diff
@@ -0,0 +1,115 @@
...
+        @Override
+        public LivenessCheckResult visit(OkEvent event) {
+            return ImmutableLivenessCheckResult.of(livenessJudgmentHandler.test(event), event);
+        }
+
+        @Override
+        public LivenessCheckResult visit(FailEvent event) {
+            return ImmutableLivenessCheckResult.of(false, event);
+        }
+    }
```
sudiksha27, line 105: This is indeed better!

sudiksha27 (commented):
Neat!

atlasdb-jepsen-tests/src/main/java/com/palantir/atlasdb/jepsen/timestamp/TimestampLivenessChecker.java
jeremyk-91: yep, changed!

PR #5262, opened 2021-02-16
TL product dependency bump

sudiksha27:
**Goals (and why)**:
Required to enable batching in internal AtlasDb proxy.


Jolyon-S (approved):
Looks good, but we should hold off merging until this is stable everywhere.

sudiksha27:
0.438.0 is stable now

PR #5264, opened 2021-02-16
[CCB] | Multi-client startTransactions endpoint

sudiksha27:
**Goals (and why)**:
Implement multi-client startTransactions endpoint

- Adds conjure api definition of cross client batch endpoint for startTransactions

- The endpoints returns successfully if startTransactions futures for **_all_** queries are successful

Added unit and integration tests

Failure mechanism - fails the requests for all namespaces without retry if any request fails  

timelock-api.yml, MultiClientConjureTimelockResource.java

Before EOD tomorrow


timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -197,3 +197,10 @@ services:
         returns: LeaderTimes
         docs: |
           Version of ConjureTimelockService#leaderTime endpoint for acquiring leaderTimes for a set of namespaces.
+      startTransactions:
```
Jolyon-S, line 200: [Possibly non-actionable]: you went from `leaderTime` -> `leaderTimes` when batching, but we already had `startTransactions` as it is batched in some capacity already. No idea if or how you could differentiate (or if it's even worth doing!), but thought I'd call it out anyway.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/batch/MultiClientConjureTimelockResource.java
```diff
@@ -74,6 +77,29 @@ public static MultiClientConjureTimelockService jersey(
...
+                        requests)
+                .map(this::getStartTransactionsResponseListenableFutures)
+                .values()
+                .collect(Collectors.toList());
+        return handleExceptions(() ->
+                Futures.transform(Futures.allAsList(futures), ImmutableMap::copyOf, MoreExecutors.directExecutor()));
+    }
+
+    private ListenableFuture<Map.Entry<Namespace, ConjureStartTransactionsResponse>>
+            getStartTransactionsResponseListenableFutures(
```
Jolyon-S, line 93: Method looks right as far as I can tell, but I think the name is a bit off. Maybe just `startTransactionsForSingleNamespace` or something?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/batch/MultiClientConjureTimelockResourceTest.java
```diff
@@ -91,6 +105,41 @@ public void requestThrowsIfAnyQueryFails() {
...
+                .map(namespace -> ConjureStartTransactionsRequest.builder()
+                        .numTransactions(5)
+                        .requestId(UUID.randomUUID())
+                        .requestorId(UUID.randomUUID())
+                        .build())
+                .mapKeys(Namespace::of)
+                .collectToMap();
+    }
+
+    private ConjureStartTransactionsResponse mockedStartTxnResponse() {
```
Jolyon-S, line 134: Hmm, I get that all the fields here are mocked, but the object itself is not. Maybe worth a rename? Just `fakeStartTxnResponse` or even just `startTransactionResponse`? I don't know this one, but does this need to be a method, or does it work if you make it a field? Maybe the mocking doesn't work that way, not 100% sure there.

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/batch/MultiClientConjureTimelockResourceTest.java
```diff
@@ -91,6 +105,41 @@ public void requestThrowsIfAnyQueryFails() {
                 .isInstanceOf(BlockingTimeoutException.class);
     }
 
+    @Test
+    public void canStartTransactionsForMultipleClients() {
```
Jolyon-S, line 109: Nice!

However, it's probably also worth having a test where the requests and responses are different for each namespace (even if as simple as different number of requests). Then you can verify that different requests and responses are actually being sent / received.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -471,6 +477,70 @@ public void sanityCheckMultiClientLeaderTimeAgainstConjureTimelockService() {
...
+            TestableTimelockServer leader, List<String> expectedNamespaces) {
+        MultiClientConjureTimelockService multiClientConjureTimelockService = leader.multiClientService();
+        Map<Namespace, ConjureStartTransactionsRequest> namespaceToRequestMap =
+                defaultNamespacedStartTransactionsRequests(expectedNamespaces);
+
+        Map<Namespace, ConjureStartTransactionsResponse> startTransactions =
+                multiClientConjureTimelockService.startTransactions(AUTH_HEADER, namespaceToRequestMap);
+
+        Set<String> namespaces =
+                startTransactions.keySet().stream().map(Namespace::get).collect(Collectors.toSet());
```
Jolyon-S, line 539: I don't suppose you could also check the number of transactions started?

Jolyon-S (commented):
This looks good. Have a couple of comments on the tests if you want to be really defensive, but the actual logic looks fine.

Not going to approve because it's probably worth a once-over from @jeremyk-91 as he's generally been tracking this much more closely.

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -471,6 +477,81 @@ public void sanityCheckMultiClientLeaderTimeAgainstConjureTimelockService() {
...
+                multiClientConjureTimelockService.startTransactions(AUTH_HEADER, namespaceToRequestMap);
+
+        // Whether we hit the multi client endpoint or conjureTimelockService endpoint, for a namespace, the underlying
+        // service to process the request is the same
+        startTransactions.forEach((namespace, responseFromBatchedEndpoint) -> {
+            ConjureStartTransactionsResponse startTransactionsResponse = leader.client(namespace.get())
+                    .namespacedConjureTimelockService()
+                    .startTransactions(namespaceToRequestMap.get(namespace));
+            assertThat(startTransactionsResponse.getLockWatchUpdate().logId())
+                    .isEqualTo(responseFromBatchedEndpoint.getLockWatchUpdate().logId());
```
jeremyk-91, line 515: Should we add an assertion on their start timestamps as well?

timelock-server/src/suiteTest/java/com/palantir/atlasdb/timelock/MultiNodePaxosTimeLockServerIntegrationTest.java
```diff
@@ -471,6 +477,81 @@ public void sanityCheckMultiClientLeaderTimeAgainstConjureTimelockService() {
...
+
+    private Map<Namespace, ConjureStartTransactionsResponse> assertSanityAndStartTransactions(
+            TestableTimelockServer leader, List<String> expectedNamespaces) {
+        MultiClientConjureTimelockService multiClientConjureTimelockService = leader.multiClientService();
+        int numTransactions = 5;
+
+        Map<Namespace, ConjureStartTransactionsRequest> namespaceToRequestMap =
+                defaultNamespacedStartTransactionsRequests(expectedNamespaces, numTransactions);
+
+        Map<Namespace, ConjureStartTransactionsResponse> startTransactions =
```
jeremyk-91, line 539: nit: `startedTransactions` or something to that effect.

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/batch/MultiClientConjureTimelockResource.java
```diff
@@ -74,6 +77,28 @@ public static MultiClientConjureTimelockService jersey(
...
+    @Override
+    public ListenableFuture<Map<Namespace, ConjureStartTransactionsResponse>> startTransactions(
+            AuthHeader authHeader, Map<Namespace, ConjureStartTransactionsRequest> requests) {
+        List<ListenableFuture<Map.Entry<Namespace, ConjureStartTransactionsResponse>>> futures = KeyedStream.stream(
+                        requests)
+                .map(this::startTransactionsForSingleNamespace)
+                .values()
+                .collect(Collectors.toList());
+        return handleExceptions(() ->
+                Futures.transform(Futures.allAsList(futures), ImmutableMap::copyOf, MoreExecutors.directExecutor()));
```
jeremyk-91, line 89: [not actionable now]: I'm fine with this, though maybe we should explore at some point whether the futures are costing us anything. I don't think it will, but just in case...

jeremyk-91 (commented):
🕐 0:21 🏁 🔥 

Broadly in agreement. I think one of the tests could be beefed up a little if I'm not mistaken. Otherwise looks good.

jeremyk-91 (approved):
👍

PR #5265, opened 2021-02-16
[TS] Metrics logging for debugging

gmaretic:
**Goals (and why)**:
Seems that the new lag metrics sometimes just disappear. Since the rest of the metrics is fine, this probably means that the gauge is returning null. Added logging to debug this issue.


jeremyk-91 (approved):
🕐 0:21 🏁 🔥

PR #5266, opened 2021-02-17
Fix WeightedHosts concurrency bug

jeremyk-91:
**Goals (and why)**:
- @gsheasby noted that he caught a `NoSuchElementException` in `WeightedHosts`. This should not happen.
- > Jeremy Kong  12 minutes ago
I think there’s a fun race condition here…
so buildHostsWeightedByActiveConnections returns an empty map, and we know pools is nonempty since there’s an assert just before that’s called
the map passed in as an input is a view of a mutable ConcurrentHashMap (CassandraService:316)
this map is updated in the background when we see the cassandra ring change
so if there were some matching pools to the query that were all taken out when we refreshed our view of the cluster, then you could get a NoSuchElementException in WeightedHosts :fire:

- This PR takes a snapshot of the state of the cluster before doing the WeightedHosts computation.

Unfortunately none, we don't have good infra for this right now.

- Is this safe? I argue it is: this method could always already have returned an `Optional.of(x)` where `x` had been removed from the host map.
- Is this too costly? I argue no: a host-map should not have more than O(100) elements even in extreme cases.

**Where should we start reviewing?**: It's really small

**Priority (whenever / two weeks / yesterday)**: Today please

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -314,8 +314,8 @@ public class CassandraService implements AutoCloseable {
 
         Set<InetSocketAddress> localFilteredHosts = maybeFilterLocalHosts(desiredHosts);
 
-        Map<InetSocketAddress, CassandraClientPoolingContainer> matchingPools =
-                KeyedStream.stream(ImmutableMap.copyOf(currentPools))
+        Map<InetSocketAddress, CassandraClientPoolingContainer> matchingPools = KeyedStream.stream(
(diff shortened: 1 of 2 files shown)


Jolyon-S:
>this map is updated in the background when we see the cassandra ring change
so if there were some matching pools to the query that were all taken out when we refreshed our view of the cluster, then you could get a NoSuchElementException in WeightedHosts 
I'm curious to know what behaviour you want or expect instead. Yes, this will fix the NSEE, but if the view of the C* ring changes such that there are no hosts now, do we still want to be able to run a query on a host that is no longer in that map? Would it be better to check again and have the option to return empty elsewhere? Will this throw elsewhere because we're trying to run a query that is no longer in the map?

Jolyon-S (approved):
Discussed offline - my concern was already true anyway.

PR #5268, opened 2021-02-18
ALP Hotfix

sudiksha27:
**Goals (and why)**:
Fixes NullPointerExceptions in LeadershipState.
It is legal to have delegate set to null while the leadershipTokenRef still exists i.e. when the proxy is closed.

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/leader/proxy/LeadershipStateManager.java
+++ b/leader-election-impl/src/main/java/com/palantir/leader/proxy/LeadershipStateManager.java
@@ -160,6 +160,9 @@ public class LeadershipStateManager<T> {
     interface LeadershipState<T> {
         LeadershipToken leadershipToken();
 
+        // Making this nullable instead of optional as this object will be called by 1000s of threads and we want to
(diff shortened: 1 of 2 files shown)


leader-election-impl/src/main/java/com/palantir/leader/proxy/LeadershipStateManager.java
```diff
@@ -160,6 +160,12 @@ void handleNotLeading(final LeadershipToken leadershipToken, @Nullable Throwable
     interface LeadershipState<T> {
         LeadershipToken leadershipToken();
 
+        // It is legal to have delegate set to null while the leadershipTokenRef has not been cleared i.e. when the
+        // proxy is closed.
+        //
+        // Making this nullable instead of optional as this object will be called by 1000s of threads and we want to
+        // avoid the cost of setting an getting the optional.
```
jeremyk-91, line 167:
```suggestion
        // avoid the cost of setting and getting the optional.
```

jeremyk-91 (approved):
🕐  0:03 🏁  0:04

One thing

PR #5272, opened 2021-02-23
Allow configuring Cassandra servers in runtime configuration 

pkoenig10:
**Goals (and why)**:
Another (hopefully simpler) attempt at https://github.com/palantir/atlasdb/pull/5249, that I believe addresses all of the mentioned concerns.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -19,11 +19,11 @@ import com.palantir.atlasdb.config.LeaderConfig;
 import com.palantir.atlasdb.keyvalue.api.KeyValueService;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
 import com.palantir.atlasdb.util.MetricsManager;
+import com.palantir.refreshable.Refreshable;
 import com.palantir.timestamp.ManagedTimestampService;
 import com.palantir.timestamp.TimestampStoreInvalidator;
 import java.util.Optional;
 import java.util.function.LongSupplier;
-import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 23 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/DelegatingCassandraKeyValueServiceConfig.java
```diff
@@ -0,0 +1,241 @@
...
+import com.palantir.atlasdb.keyvalue.cassandra.async.CassandraAsyncKeyValueServiceFactory;
+import com.palantir.atlasdb.keyvalue.cassandra.pool.HostLocation;
+import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
+import java.net.InetSocketAddress;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Supplier;
+
+interface DelegatingCassandraKeyValueServiceConfig extends CassandraKeyValueServiceConfig {
```
pnepywoda, line 28: nit: looks like atlasdb calles these `Forwarding*`

pnepywoda (commented):
worth adding some tests? you can just take mine from https://github.com/palantir/atlasdb/pull/5249/files#diff-b9ef4433996a0706ed1d72c26d6f2334a45f0fd15425454bbedaed2d471b7d4dR170 if you wish

pnepywoda (approved):

jeremyk-91 (approved):
Thanks for the solid implementation! This looks good to me as well.

It's important to note exactly what this change does. AtlasDB has a mechanism by which it discovers the Cassandra cluster. However, to bootstrap this and/or in the event of network failures, the config provides a list of initial contact points between AtlasDB and Cassandra. This change unlocks having that list of contact points update at runtime. Is this what you expect / will this be sufficient for what you're expecting more broadly?

pnepywoda:
Specifically for us, we just need the list of servers to be in runtime instead of install config. This PR more than satisfies that

PR #5275, opened 2021-02-23
[CCB] | MultiClientTransactionStarter refactor - part 2

sudiksha27:
**Goals (and why)**:
Atlas should be able to accept a common batcher for startTransactions requests for multiple clients.

- Implements AutoBatcher capable of handling startTransactions requests for multiple clients
- The model is very close to the origin model - for each namespace, while all requests cannot be served, we batch all the requests together, try to startTransactions in one call to the server, update the cache after receiving the response and split the response into list of StartIdentifiedAtlasDbTransactionResponse.
- The LockWatchEventCache#lastKnownVersion is fetched before making request to TimeLock and LockWatchEventCache#processStartTransactionsUpdate is invoked as response is received. 
- We try to serve requests as responses become available instead of waiting for all responses to serve all requests in one go.
- In case of exception, all pending requests are cancelled.
- I haven't done the wiring yet as 1. it is subject to changes if the data structures change as per reviews and 2. the PR is very chunky already.

- Added unit and integ tests.

- Does the design look okay? Should we look more into optimizing the perf of batch processing?
- Did I miss/ break something horribly?

MultiClientBatchingIdentifiedAtlasDbTransactionStarter.java

Before EOD tomorrow is ideal.

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
@@ -19,6 +19,10 @@ package com.palantir.lock.client;
 import com.palantir.lock.v2.LockToken;
 import java.util.Set;
 
+/**
+ * This is a facade of {@link LockLeaseService} meant to expose only the functionality required for starting a
+ * batch of transactions, see {@link MultiClientTransactionStarter}.
+ * */
 public final class LockCleanupService {
     private final LockLeaseService delegate;
 
--- a/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
@@ -30,11 +30,11 @@ public final class LockCleanupService {
         this.delegate = delegate;
     }
 
-    Set<LockToken> refreshLockLeases(Set<LockToken> uncastedTokens) {
-        return delegate.refreshLockLeases(uncastedTokens);
+    Set<LockToken> refreshLockLeases(Set<LockToken> userTokens) {
+        return delegate.refreshLockLeases(userTokens);
     }
 
-    Set<LockToken> unlock(Set<LockToken> tokens) {
-        return delegate.unlock(tokens);
+    Set<LockToken> unlock(Set<LockToken> userTokens) {
+        return delegate.unlock(userTokens);
     }
 }
--- a/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
@@ -1,0 +1,36 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.lock.v2.LockToken;
+import java.util.Set;
+
+public final class LockCleanupService {
+    private final LockLeaseService delegate;
+
+    LockCleanupService(LockLeaseService delegate) {
+        this.delegate = delegate;
+    }
+
+    Set<LockToken> refreshLockLeases(Set<LockToken> uncastedTokens) {
+        return delegate.refreshLockLeases(uncastedTokens);
+    }
+
+    Set<LockToken> unlock(Set<LockToken> tokens) {
+        return delegate.unlock(tokens);
+    }
+}
--- a/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
@@ -211,7 +211,7 @@ class LockLeaseService {
     public static final class LockCleanupService {
         private final LockLeaseService delegate;
 
-        public LockCleanupService(LockLeaseService delegate) {
+        private LockCleanupService(LockLeaseService delegate) {
             this.delegate = delegate;
         }
 
--- a/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
@@ -194,20 +194,4 @@ class LockLeaseService {
     private static Set<ConjureLockToken> serverTokens(Set<LeasedLockToken> leasedTokens) {
         return leasedTokens.stream().map(LeasedLockToken::serverToken).collect(Collectors.toSet());
     }
-
-    public static final class LockCleanupService {
-        private final LockLeaseService delegate;
-
-        private LockCleanupService(LockLeaseService delegate) {
-            this.delegate = delegate;
-        }
-
-        Set<LockToken> refreshLockLeases(Set<LockToken> uncastedTokens) {
-            return delegate.refreshLockLeases(uncastedTokens);
-        }
-
-        Set<LockToken> unlock(Set<LockToken> tokens) {
-            return delegate.unlock(tokens);
-        }
-    }
 }
--- a/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
@@ -97,10 +97,10 @@ class LockLeaseService {
                 .numTransactions(batchSize)
                 .lastKnownVersion(ConjureLockRequests.toConjure(maybeVersion))
                 .build();
-        return getMassagedConjureStartTransactionsResponse(delegate.startTransactions(request));
+        return assignLeasedLockTokenToImmutableTimestampLock(delegate.startTransactions(request));
     }
 
-    public static ConjureStartTransactionsResponse getMassagedConjureStartTransactionsResponse(
+    public static ConjureStartTransactionsResponse assignLeasedLockTokenToImmutableTimestampLock(
             ConjureStartTransactionsResponse response) {
         Lease lease = response.getLease();
         LeasedLockToken leasedLockToken = LeasedLockToken.of(
--- a/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
@@ -94,7 +94,11 @@ class LockLeaseService {
                 .numTransactions(batchSize)
                 .lastKnownVersion(toConjure(maybeVersion))
                 .build();
-        ConjureStartTransactionsResponse response = delegate.startTransactions(request);
+        return getMassagedConjureStartTransactionsResponse(delegate.startTransactions(request));
+    }
+
+    public static ConjureStartTransactionsResponse getMassagedConjureStartTransactionsResponse(
+            ConjureStartTransactionsResponse response) {
         Lease lease = response.getLease();
         LeasedLockToken leasedLockToken = LeasedLockToken.of(
                 ConjureLockToken.of(response.getImmutableTimestamp().getLock().getRequestId()), lease);
@@ -154,6 +158,10 @@ class LockLeaseService {
                 .collect(Collectors.toSet());
     }
 
+    LockCleanupService lockCleanupService() {
+        return new LockCleanupService(this);
+    }
+
     private Set<LeasedLockToken> refreshTokens(Set<LeasedLockToken> leasedTokens) {
         if (leasedTokens.isEmpty()) {
             return leasedTokens;
@@ -184,10 +192,33 @@ class LockLeaseService {
         return leasedTokens.stream().map(LeasedLockToken::serverToken).collect(Collectors.toSet());
     }
 
-    private Optional<ConjureIdentifiedVersion> toConjure(Optional<LockWatchVersion> maybeVersion) {
+    // Todo snanda
+    public static Optional<ConjureIdentifiedVersion> toConjure(Optional<LockWatchVersion> maybeVersion) {
         return maybeVersion.map(identifiedVersion -> ConjureIdentifiedVersion.builder()
                 .id(identifiedVersion.id())
                 .version(identifiedVersion.version())
                 .build());
+    }
+
+    public static Optional<LockWatchVersion> fromConjure(Optional<ConjureIdentifiedVersion> maybeVersion) {
+        return maybeVersion.map(
+                identifiedVersion -> LockWatchVersion.of(identifiedVersion.getId(), identifiedVersion.getVersion()));
+    }
+
+    // Todo snanda
+    public static final class LockCleanupService {
+        private final LockLeaseService delegate;
+
+        public LockCleanupService(LockLeaseService delegate) {
+            this.delegate = delegate;
+        }
+
+        Set<LockToken> refreshLockLeases(Set<LockToken> uncastedTokens) {
+            return delegate.refreshLockLeases(uncastedTokens);
+        }
+
+        Set<LockToken> unlock(Set<LockToken> tokens) {
+            return delegate.unlock(tokens);
+        }
     }
 }
--- a/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
@@ -19,7 +19,6 @@ package com.palantir.lock.client;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.Sets;
-import com.palantir.atlasdb.timelock.api.ConjureIdentifiedVersion;
 import com.palantir.atlasdb.timelock.api.ConjureLockToken;
 import com.palantir.atlasdb.timelock.api.ConjureRefreshLocksRequest;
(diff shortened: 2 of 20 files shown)


lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -41,20 +41,23 @@
...
 import java.util.UUID;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 class LockLeaseService {
     private final NamespacedConjureTimelockService delegate;
     private final UUID clientId;
     private final LeaderTimeGetter leaderTimeGetter;
     private final BlockEnforcingLockService lockService;
+    private final Supplier<LockCleanupService> lockCleanupServiceSupplier;
```
jeremyk-91, line 52: How important is is that there is only one for each LockLeaseService? I _think_ it's fine for factories to create their own view each time they need a LockCleanupService

lock-api/src/main/java/com/palantir/lock/client/MultiClientBatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -0,0 +1,319 @@
...
+
+    private static void clearResources(
+            MultiClientRequestManager multiClientRequestManager,
+            Map<Namespace, List<StartIdentifiedAtlasDbTransactionResponse>> startTransactionResponses) {
+        startTransactionResponses.forEach((namespace, responseList) -> TransactionStarterHelper.unlock(
+                responseList.stream()
+                        .map(response -> response.immutableTimestamp().getLock())
+                        .collect(Collectors.toSet()),
+                multiClientRequestManager.getLockCleanupService(namespace)));
+    }
```
jeremyk-91, line 118: I think we also need to clear pending responses in the multi-client request manager?

jeremyk-91: It may be easier to have it be (auto)closeable and have a finally block around the while loop in `processBatch`.

lock-api/src/main/java/com/palantir/lock/client/MultiClientBatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -0,0 +1,319 @@
...
+
+    private static Map<Namespace, ResponseHandler> getResponseHandlers(
+            List<BatchElement<NamespaceAndRequestParams, List<StartIdentifiedAtlasDbTransactionResponse>>> batch) {
+        Map<Namespace, Queue<SettableResponse>> namespaceWisePendingFutures = new HashMap<>();
+
+        for (BatchElement<NamespaceAndRequestParams, List<StartIdentifiedAtlasDbTransactionResponse>> element : batch) {
+            Namespace namespace = element.argument().namespace();
+            namespaceWisePendingFutures
+                    .computeIfAbsent(namespace, _u -> new LinkedList<>())
```
jeremyk-91, line 127: nit: `_unused`?

lock-api/src/main/java/com/palantir/lock/client/MultiClientBatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -0,0 +1,319 @@
...
+                    originalRequestMap
+                            .get(namespace)
+                            .cache()
+                            .processStartTransactionsUpdate(
+                                    response.getTimestamps().stream().boxed().collect(Collectors.toSet()),
+                                    response.getLockWatchUpdate());
+                    LockWatchLogUtility.logTransactionEvents(
+                            fromConjure(namespaceWiseRequests.get(namespace).getLastKnownVersion()),
+                            response.getLockWatchUpdate());
+                    return Maps.immutableEntry(namespace, TransactionStarterHelper.split(response));
```
jeremyk-91, line 168: Hmm, there may be some deduplication we can do with `BatchingIdentifiedAtlasDbTransactionStarter`, at least once we've got the cache. Give this a try, though with this one I'm aware there may be unexpected difficulties 🔥

lock-api/src/main/java/com/palantir/lock/client/MultiClientBatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Queue;
+import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import org.immutables.value.Value;
+
+public final class MultiClientBatchingIdentifiedAtlasDbTransactionStarter implements AutoCloseable {
```
jeremyk-91, line 49: thought: would `MultiClientTransactionStarter` be enough for the name?

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -184,10 +195,19 @@ WaitForLocksResponse waitForLocks(WaitForLocksRequest request) {
         return leasedTokens.stream().map(LeasedLockToken::serverToken).collect(Collectors.toSet());
     }
 
-    private Optional<ConjureIdentifiedVersion> toConjure(Optional<LockWatchVersion> maybeVersion) {
-        return maybeVersion.map(identifiedVersion -> ConjureIdentifiedVersion.builder()
-                .id(identifiedVersion.id())
-                .version(identifiedVersion.version())
-                .build());
+    public static final class LockCleanupService {
```
jeremyk-91, line 198: discussed offline: not sure this should be an inner class

lock-api/src/main/java/com/palantir/lock/watch/StartTransactionsLockWatchEventCache.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.watch;
+
+import java.util.Optional;
+import java.util.Set;
+
+public final class StartTransactionsLockWatchEventCache {
```
jeremyk-91, line 22: nit: might be worth a brief javadoc saying this is a facade of LWEC.

jeremyk-91 (commented):
🕐 1:19 (including pairing time on #5283) 🏁 🔥 

This review covers the production code, I have not looked at the tests yet. I think the algorithms in this PR make sense, though there are elements of the factoring I'd change.

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+        return responseHandlers;
+    }
+
+    @VisibleForTesting
+    static Map<Namespace, RequestParams> getNamespaceWiseRequestParams(
+            List<BatchElement<NamespaceAndRequestParams, List<StartIdentifiedAtlasDbTransactionResponse>>> batch) {
+        return batch.stream()
+                .map(BatchElement::argument)
+                .collect(Collectors.toMap(
+                        NamespaceAndRequestParams::namespace, NamespaceAndRequestParams::params, RequestParams::merge));
```
gmaretic, line 138: non-actionable: I would probably spend too much time figuring out how to do this with groupingBy and not creating intermediate objects, but this is fine :)

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+
+        static NamespaceAndRequestParams of(Namespace namespace, RequestParams params) {
+            return ImmutableNamespaceAndRequestParams.of(namespace, params);
+        }
+    }
+
+    @Value.Immutable
+    interface RequestParams {
+        @Value.Parameter
+        Integer numTransactions();
```
gmaretic, line 207: This class is basically always used in a way where this number changes and the other two params are fixed. Given we don't really use the immutability or the generated code anywhere, maybe this should just be a normal concrete class with a method to change `numTransactions`? That said, we are not creating millions of objects, so it'a also fine to leave as is

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+        public void updatePendingStartTransactionsCount(Namespace namespace, int startedTransactionsCount) {
+            Optional<RequestParams> updatedParams = paramsAfterResponse(namespace, startedTransactionsCount);
+            if (updatedParams.isPresent()) {
+                requestMap.put(namespace, updatedParams.get());
+            } else {
+                requestMap.remove(namespace);
+            }
+        }
+
+        private Optional<RequestParams> paramsAfterResponse(Namespace namespace, int startedTransactionsCount) {
```
gmaretic, line 262: Maybe rename to something like `findRemainingRequests`?

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+            RequestParams params = requestMap.get(namespace);
+            int numTransactions = params.numTransactions();
+            return startedTransactionsCount < numTransactions
+                    ? Optional.of(RequestParams.of(
+                            numTransactions - startedTransactionsCount, params.cache(), params.lockCleanupService()))
+                    : Optional.empty();
+        }
```
gmaretic, line 269: Also, I think this is kinda nicer, but double check that it makes sense :)

```suggestion
        public void updatePendingStartTransactionsCount(Namespace namespace, int startedTransactionsCount) {
            requestMap.compute(namespace, (_unused, params) -> paramsAfterResponse(params, startedTransactionsCount));
        }

        private RequestParams paramsAfterResponse(RequestParams params, int startedTransactionsCount) {
            int numTransactions = params.numTransactions();
            if (startedTransactionsCount < numTransactions) {
                return ImmutableRequestParams
                        .builder()
                        .from(params)
                        .numTransactions(numTransactions - startedTransactionsCount)
                        .build();
            }
            return null;
        }
```

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+    static void processBatch(
+            InternalMultiClientConjureTimelockService delegate,
+            UUID requestorId,
+            List<BatchElement<NamespaceAndRequestParams, List<StartIdentifiedAtlasDbTransactionResponse>>> batch) {
+
+        Map<Namespace, ResponseHandler> namespaceWiseResponseHandler = getResponseHandlers(batch);
+        MultiClientRequestManager multiClientRequestManager =
+                new MultiClientRequestManager(getNamespaceWiseRequestParams(batch));
+
+        Map<Namespace, List<StartIdentifiedAtlasDbTransactionResponse>> startTransactionResponses = new HashMap<>();
```
gmaretic, line 91: initializer redundant

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -0,0 +1,323 @@
...
+            Map<Namespace, RequestParams> originalRequestMap,
+            InternalMultiClientConjureTimelockService delegate,
+            UUID requestorId) {
+
+        Map<Namespace, ConjureStartTransactionsRequest> namespaceWiseRequests =
+                getNamespaceWiseRequests(originalRequestMap, requestorId);
+        Map<Namespace, ConjureStartTransactionsResponse> responseMap = getResponseMap(delegate, namespaceWiseRequests);
+        return KeyedStream.stream(responseMap)
+                .mapEntries((namespace, response) -> {
+                    TransactionStarterHelper.updateCacheWithStartTransactionResponse(
```
gmaretic, line 151: This is a side-effect inside a stream, which is not really good practice. I suggest you just rewrite this using a for loop

gmaretic: Also the `getStartTransactionResponses` should indicate it's updating the cache

lock-api/src/main/java/com/palantir/lock/client/NamespacedBatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package com.palantir.lock.client;
+
+import com.palantir.atlasdb.timelock.api.Namespace;
+import com.palantir.lock.v2.StartIdentifiedAtlasDbTransactionResponse;
+import com.palantir.lock.watch.StartTransactionsLockWatchEventCache;
+import java.util.List;
+
+public class NamespacedBatchingIdentifiedAtlasDbTransactionStarter implements IdentifiedAtlasDbTransactionStarter {
```
gmaretic, line 24: Maybe a more concise name?

gmaretic (commented):
Reviewed non-test classes. Mostly looks good, but some style suggestions

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private void assertSanityOfResponse(
+            List<BatchElement<NamespaceAndRequestParams, List<StartIdentifiedAtlasDbTransactionResponse>>>
+                    requestsForClients,
+            boolean assertValues) {
+
+        UUID requestorId = UUID.randomUUID();
+        Map<Namespace, ConjureStartTransactionsResponse> responseMap =
+                getMultiClientStartTransactionsResponse(requestsForClients, requestorId);
+        when(timelockService.startTransactions(any())).thenReturn(responseMap);
```
jeremyk-91, line 147: Ah, I see. This probably should use `thenAnswer(...)` so that you can also validate the values on the `canServiceOneRequestWithMultipleServerRequests` test

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+        SafeIllegalStateException exception = new SafeIllegalStateException("Something went wrong!");
+        when(timelockService.startTransactions(any())).thenReturn(responseMap).thenThrow(exception);
+
+        assertThatThrownBy(() -> processBatch(timelockService, requestorId, requests))
+                .isEqualTo(exception);
+        verify(lockCleanupService).refreshLockLeases(any());
+        verify(lockCleanupService).unlock(any());
+    }
+
```
jeremyk-91, line 138: I think we need one more test, which unfortunately may not be that easy to write (I'm fine whether this is separate or rolled into the above one): the lock tokens corresponding to requests that were already served are _not_ automatically unlocked.

You will likely need to use `ArgumentCaptor` or some mechanism for inspecting what exactly was unlocked in this case.

jeremyk-91 (commented):
🕐 2:00 🏁 ????

Reviewed the tests. They broadly make sense, though there's an important case we should check, and I do think in general we could have some stronger assertions.

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
sudiksha27: Right, but we will need need to do one iteration of cloning as the params in the original BatchElement should not be updated, which will become one more thing to worry about in this class.

lock-api/src/main/java/com/palantir/lock/client/LockCleanupService.java
```diff
@@ -0,0 +1,40 @@
...
+ * batch of transactions, see {@link MultiClientTransactionStarter}.
+ * */
+public final class LockCleanupService {
+    private final LockLeaseService delegate;
+
+    LockCleanupService(LockLeaseService delegate) {
+        this.delegate = delegate;
+    }
+
+    Set<LockToken> refreshLockLeases(Set<LockToken> uncastedTokens) {
```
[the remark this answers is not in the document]
sudiksha27: Apologies, I have not looked deeply into the actual mechanics of locking/ unlocking/ refreshing. This probably refers to non-leased lockTokens.

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
jeremyk-91: we paired on this! turns out the argumentcaptor can do some pretty cool stuff here

jeremyk-91 (approved):
👍

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
```diff
@@ -161,6 +166,47 @@ public void shouldNotFreeResourcesIfRequestIsServed() {
...
+        assertThatThrownBy(() -> processBatch(timelockService, requestorId, requests))
+                .isEqualTo(EXCEPTION);
+
+        // assert requests made by client omega are served
+        assertSanityOfRequestBatch(
+                ImmutableList.of(requestForOmega), ImmutableMap.of(omega, ImmutableList.of(responseMap.get(omega))));
+
+        @SuppressWarnings({"unchecked", "rawtypes"}) // :internally-screaming:
+        ArgumentCaptor<Set<LockToken>> refreshArgumentCaptor =
+                (ArgumentCaptor<Set<LockToken>>) ArgumentCaptor.forClass((Class) Set.class);
```
sudiksha27, line 196: let's edit the first comment and let's remove all the comments here

sudiksha27:
🎖️ 👍

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
sudiksha27: Note - We actually do use the immutability of this class to ensure that the request batch cannot have its state mutated.

PR #5276, opened 2021-02-23
[TS] Hashes, now even better

gmaretic:
**Goals (and why)**:
The previous impl was actually only generating even values for equal row and column

Without going into details, this should give us good enough has distribution while still being reasonably performant.

Unit tests

🔥

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
@@ -28,6 +28,12 @@ public abstract class CellReference {
 
     public abstract Cell cell();
 
+
+    /**
+     * {@link Cell#hashCode()} implementation has a rather unfortunate case where it is always 0 if the row name and
+     * the column name match. We did not want to change it to keep backwards compatibility, but we need a uniform
+     * distribution here for all reasonable patterns.
+     */
     public int goodHash() {
         int hash = 5381;
         hash = hash * 31 + tableRef().hashCode();
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
@@ -15,27 +15,35 @@
  */
 package com.palantir.atlasdb.keyvalue.api;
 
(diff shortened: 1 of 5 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/AbstractSweepQueueTest.java
```diff
@@ -165,14 +166,12 @@ void putTimestampIntoTransactionTable(long ts, long commitTs) {
...
-        int counter = 0;
-        CellReference cellReference;
-        do {
-            byte[] colName = PtBytes.toBytes(counter);
-            cellReference = CellReference.of(tableRef, Cell.create(rowName, colName));
-            counter++;
-        } while (IntMath.mod(cellReference.hashCode(), shards) != 0);
-        return cellReference;
+
+        return IntStream.iterate(0, i -> i + 1)
```
berler, line 170: Is there any concern here that this implementation becomes more expensive as the number of shards increases?

berler: Just realized this seems to only be used for tests

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
```diff
@@ -15,31 +15,35 @@
...
+
     public abstract TableReference tableRef();
 
     public abstract Cell cell();
 
-    /**
-     * {@link Cell#hashCode()} implementation has a rather unfortunate case where it is always 0 if the row name and
-     * the column name match. We did not want to change it to keep backwards compatibility, but we need a uniform
-     * distribution here for all reasonable patterns.
-     */
```
jeremyk-91, line 31: I think this context is still useful?

jeremyk-91 (approved):
🕐 0:03 🏁  🔥 

👍  removed MWR as I don't know if the comment is still useful, defer that to you

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/CellReference.java
gmaretic: makes sense

PR #5277, opened 2021-02-23
[TS | LIP-164000] Reset Sweep Progress

jeremyk-91:
**Goals (and why)**:
- Have the ability to reset sweep progress. This may be useful to deal with circumstances where entries are written to the targeted sweep queue after the sweep timestamp has progressed past it - while the transaction in question will necessarily fail, there may still be cruft in the targeted sweep queue.

- Add a config flag. If set, on startup does a CAS to 0 and doesn't perform targeted sweep. This is needed because a service has multiple nodes, but once the last node rolls, assuming it starts successfully, no one is running sweep and we CAS the progress to 0 -> we have 0 for all shards and strategies.

**Testing (What was existing testing like?  What have you done to improve it?)**: Added some unit tests.

- What happens if the number of shards in the config is not what's actually persisted in the DB? I believe we do this correctly, but worth a check.
- Have I thought through the concurrency model (especially across multiple nodes) correctly? and/or, is this too heavy-handed?

**Where should we start reviewing?**: ShardProgress

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
@@ -52,6 +52,7 @@ import org.slf4j.LoggerFactory;
 public class TargetedSweeper implements MultiTableSweepQueueWriter, BackgroundSweeper {
     private static final Logger log = LoggerFactory.getLogger(TargetedSweeper.class);
 
+    private final boolean shouldResetAndStopSweep;
     private final Supplier<TargetedSweepRuntimeConfig> runtime;
     private final List<Follower> followers;
     private final MetricsManager metricsManager;
@@ -76,6 +77,7 @@ public class TargetedSweeper implements MultiTableSweepQueueWriter, BackgroundSw
         this.conservativeScheduler =
                 new BackgroundSweepScheduler(install.conservativeThreads(), SweeperStrategy.CONSERVATIVE);
         this.thoroughScheduler = new BackgroundSweepScheduler(install.thoroughThreads(), SweeperStrategy.THOROUGH);
+        this.shouldResetAndStopSweep = install.resetTargetedSweepQueueProgressAndStopSweep();
         this.followers = followers;
         this.metricsConfiguration = install.metricsConfiguration();
     }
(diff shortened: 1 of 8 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ShardProgressTest.java
```diff
@@ -186,6 +188,46 @@ public void repeatedlyFailingCasThrows() {
...
+    public void repeatedlyFailingCasThrowsForReset() {
+        KeyValueService mockKvs = mock(KeyValueService.class);
+        when(mockKvs.get(any(), anyMap()))
+                .thenReturn(ImmutableMap.of(DUMMY, createValue(8L)))
+                .thenReturn(ImmutableMap.of(DUMMY, createValue(9L)))
+                .thenReturn(ImmutableMap.of(DUMMY, createValue(10L)));
+        doThrow(new CheckAndSetException("sadness")).when(mockKvs).checkAndSet(any());
+        ShardProgress instrumentedProgress = new ShardProgress(mockKvs);
+
+        assertThatCode(() -> instrumentedProgress.resetProgressForShard(CONSERVATIVE_TEN))
```
gmaretic, line 227: Probably want to verify that it only throws once the value actually repeats

gmaretic (approved):
It's fine, but I'm not really sure why we want to stop sweep in addition to resetting. Though it's a good forcing function to make people remove the flag so I guess that's the reason.

gmaretic:
Discussed offline: document why we need to stop running sweep and wait for all nodes to reset

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ShardProgressTest.java
jeremyk-91: Hmm, I think this is tested by `stopsTryingToResetIfSomeoneElseDid()` above - in practice you could check the CheckAndSetException's actual values but we don't get that here because of mocks, so I don't see another way of easily doing this?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/queue/TargetedSweeper.java
```diff
@@ -175,8 +177,15 @@ public void initializeWithoutRunning(
     @Override
     public void runInBackground() {
         assertInitialized();
-        conservativeScheduler.scheduleBackgroundThreads();
-        thoroughScheduler.scheduleBackgroundThreads();
+        if (shouldResetAndStopSweep) {
+            log.warn("This AtlasDB node is operating in a mode where it is attempting to reset the progress of "
```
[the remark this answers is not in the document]
jeremyk-91: Discussed offline: this centers around the behaviour of targeted sweep where nodes CAS the bound from (thing I read -> my progress) repeatedly. We need to wait for *all* nodes to report that they're done with this.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/sweep/queue/ShardProgressTest.java
gmaretic: OH I meant just verify that the method was called 4 times

jeremyk-91: added!

PR #5279, opened 2021-02-24
[CCB] | MultiClientTransactionStarter refactor - part 3

sudiksha27:
See #5275 for reference
**Goals (and why)**:
AtlasDb accepts external request batcher for startTransactions requests.

- Wired startTransactionsBatcherProvider through TransactionManagers

Tested locally

Does the design seem reasonable?

TransactionManagers.java

Before Monday EOD please.


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1206,11 +1212,23 @@ private static LockAndTimestampServices getLockAndTimestampServices(
                 serviceProvider.getConjureLockWatchingService(), timelockNamespace);
         LockWatchManagerImpl lockWatchManager = new LockWatchManagerImpl(lockWatchEventCache, lockWatchingService);
 
-        LeaderTimeGetter leaderTimeGetter = getLeaderTimeGetter(
-                timelockNamespace, timelockRequestBatcherProviders, serviceProvider, namespacedConjureTimelockService);
```
Jolyon-S, line 1210: [Sanity check]: drive by inline?

Jolyon-S (commented):
The wiring seems pretty sane tbh, but I don't have super good context on the underlying PR.

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -44,11 +45,12 @@ private RemoteTimelockServiceAdapter(
...
             LockWatchEventCache lockWatchEventCache,
-            LeaderTimeGetter leaderTimeGetter) {
+            LeaderTimeGetter leaderTimeGetter,
+            Function<LockLeaseService, IdentifiedAtlasDbTransactionStarter> batchingTransactionStarterFactory) {
         this.rpcClient = rpcClient;
         this.lockLeaseService = LockLeaseService.create(conjureTimelockService, leaderTimeGetter);
-        this.transactionStarter = TransactionStarter.create(
-                lockLeaseService, StartTransactionsLockWatchEventCache.create(lockWatchEventCache));
+        this.transactionStarter =
+                TransactionStarter.create(lockLeaseService, batchingTransactionStarterFactory.apply(lockLeaseService));
```
jeremyk-91, line 53: nit: Unsure if this is preferable, but it might be better to just pass in the Factory here, to ensure that the lockLeaseService used in the TransactionStarter (whether directly or indirectly) is the same one.

jeremyk-91 (commented):
🕐 0:07 🏁 2:35

Broadly makes sense

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
sudiksha27: Yep

lock-api/src/main/java/com/palantir/lock/client/TransactionStarter.java
```diff
@@ -37,15 +37,15 @@
...
-        this.batchingTransactionStarter =
-                BatchingIdentifiedAtlasDbTransactionStarter.create(lockLeaseService, lockWatchEventCache);
+        this.batchingTransactionStarter = batchingTransactionStarter;
     }
 
     static TransactionStarter create(
-            LockLeaseService lockLeaseService, StartTransactionsLockWatchEventCache lockWatchEventCache) {
-        return new TransactionStarter(lockLeaseService, lockWatchEventCache);
+            LockLeaseService lockLeaseService,
+            Function<LockLeaseService, IdentifiedAtlasDbTransactionStarter> batchingTransactionStarter) {
```
jeremyk-91, line 47: nit: the function should probably be `batchingTransactionStarterFactory`

lock-api/src/main/java/com/palantir/lock/client/BatchingIdentifiedAtlasDbTransactionStarter.java
```diff
@@ -42,7 +42,7 @@ private BatchingIdentifiedAtlasDbTransactionStarter(
         this.autobatcher = autobatcher;
     }
 
-    static BatchingIdentifiedAtlasDbTransactionStarter create(
+    public static BatchingIdentifiedAtlasDbTransactionStarter create(
             LockLeaseService lockLeaseService, StartTransactionsLockWatchEventCache lockWatchEventCache) {
```
jeremyk-91, line 46: in hindsight: if STLWEC is only used here, does it make sense to have the factory take a full LWEC?

lock-api/src/main/java/com/palantir/lock/client/LockLeaseService.java
```diff
@@ -42,7 +42,7 @@
 import java.util.UUID;
 import java.util.stream.Collectors;
 
-class LockLeaseService {
+public class LockLeaseService {
```
jeremyk-91, line 45: It seems off that you had to do this visibility change. I think it's probably an issue of changing where these are created (i.e. you want to create them inside factory methods - these are in many cases implementation details of the lock client)

atlasdb-config/src/main/java/com/palantir/atlasdb/config/TimeLockRequestBatcherProviders.java
```diff
@@ -17,9 +17,12 @@
...
 
 import com.palantir.lock.client.LeaderTimeCoalescingBatcher;
+import com.palantir.lock.client.MultiClientTransactionStarter;
 import org.immutables.value.Value;
 
 @Value.Immutable
 public interface TimeLockRequestBatcherProviders {
     TimeLockRequestBatcherProvider<LeaderTimeCoalescingBatcher> leaderTimeBatcherProvider();
+
+    TimeLockRequestBatcherProvider<MultiClientTransactionStarter> startTransactionsBatcherProvider();
```
jeremyk-91, line 27: nit: given the class is already `...BatcherProviders`, I'd just name this `leaderTime()` or `startTransactions()` respectively.

jeremyk-91 (commented):
🕐 0:15 🏁 0:17

Makes sense, I think there's some factoring that might need to be changed

jeremyk-91 (approved):
👍

sudiksha27 (commented):
👍

PR #5282, opened 2021-02-25
install and runtime keyValueService config must be merged

pnepywoda:
followup fix to https://github.com/palantir/atlasdb/pull/5272

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -27,7 +27,7 @@ import java.util.function.LongSupplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public interface AtlasDbFactory {
+public interface AtlasDbFactory<MERGED_CONFIG extends KeyValueServiceConfig> {
     Logger log = LoggerFactory.getLogger(AtlasDbFactory.class);
 
     long NO_OP_FAST_FORWARD_TIMESTAMP = Long.MIN_VALUE + 1; // Note: Long.MIN_VALUE itself is not allowed.
@@ -38,11 +38,11 @@ public interface AtlasDbFactory {
 
     String getType();
 
-    default KeyValueServiceConfig createMergedKeyValueServiceConfig(
+    default MERGED_CONFIG createMergedKeyValueServiceConfig(
             KeyValueServiceConfig config,
             Refreshable<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig,
             Optional<String> namespace) {
-        return config;
+        return (MERGED_CONFIG) config;
     }
 
     /**
@@ -61,7 +61,7 @@ public interface AtlasDbFactory {
      */
     KeyValueService createRawKeyValueService(
             MetricsManager metricsManager,
-            KeyValueServiceConfig config,
+            MERGED_CONFIG config,
             Refreshable<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig,
             Optional<LeaderConfig> leaderConfig,
             Optional<String> namespace,
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -38,6 +38,13 @@ public interface AtlasDbFactory {
 
     String getType();
 
+    default KeyValueServiceConfig createMergedKeyValueServiceConfig(
+            KeyValueServiceConfig config,
+            Refreshable<Optional<KeyValueServiceRuntimeConfig>> runtimeConfig,
+            Optional<String> namespace) {
+        return config;
+    }
+
     /**
      * Creates a KeyValueService instance of type according to the config parameter.
      *
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
@@ -38,6 +38,13 @@ public interface AtlasDbFactory<MERGED_CONFIG extends KeyValueServiceConfig> {
 
     String getType();
 
+    /**
+     * Create the config (a merging of the given install and runtime config) that will be passed to
+     * {@link #createRawKeyValueService}.
+     * {@link KeyValueServiceConfig#defaultGetRangesConcurrency()} and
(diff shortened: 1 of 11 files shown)


atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -533,8 +535,8 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                         () -> areTransactionManagerInitializationPrerequisitesSatisfied(
                                 initializer, lockAndTimestampServices),
                         allowHiddenTableAccess(),
-                        config().keyValueService().concurrentGetRangesThreadPoolSize(),
-                        config().keyValueService().defaultGetRangesConcurrency(),
+                        mergedKeyValueServiceConfig.concurrentGetRangesThreadPoolSize(),
```
pnepywoda, line 538: this value will be 0 without this fix

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
```diff
@@ -75,6 +77,10 @@ public KeyValueService getKeyValueService() {
         return keyValueService.get();
     }
 
+    public KeyValueServiceConfig getMergedKeyValueServiceConfig() {
```
pkoenig10, line 80: `getKeyValueService` already returns the one true key value service. Having a duplicative method here is unnecessary and error-prone.

pkoenig10 (dismissed):

pnepywoda:
This is the config

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/ServiceDiscoveringAtlasSupplier.java
pnepywoda: this is the config, not the service

pkoenig10: Oh I see, sorry.

Can we expose this through the existing value? It feels wrong to build this config object twice.

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
```diff
@@ -38,6 +38,13 @@
 
     String getType();
 
+    default KeyValueServiceConfig createMergedKeyValueServiceConfig(
```
pkoenig10, line 41:
```suggestion
    default KeyValueServiceConfig createKeyValueServiceConfig(
```

pkoenig10 (commented):
If we're building the merged config outside of `createRawKeyValueService`, let's just drop the runtime config parameter from `createRawKeyValueService`.

It looks like we only use it in two places in `CassandraKeyValueServiceImpl` and those probably should be using the reloadable `CassandraKeyValueServiceConfig` anyways.

pkoenig10:
Err, dropping the runtime config looks non-trivial. This seems fine.

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
pkoenig10: I'd also suggest we change the field name, etc. That fact that it's merged isn't really relevant to consumers - there should be a single config that is the source of truth.

pkoenig10, line 41: If we really wanted to make this nice, we could parameterize `AtlasDbFactory` by the config class so we wouldn't need to cast.

pnepywoda: yeah but look how they catch class cast in https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraAtlasDbFactory.java#L92

pnepywoda: I'd rather differentiate the names

pkoenig10: How is that related to my suggestion?

pnepywoda: you'd have to move the class cast catching outside of the factory then

pkoenig10: I'm just talking about the install config here (the type returned by `createMergedKeyValueServiceConfig`.

you're link above is for the runtime config, that can stay as is.

pkoenig10: This might be more trouble than it's worth. But it would help force callers to use `createMergedKeyValueServiceConfig` instead of accidentally passing in the un-merged install config.

pnepywoda: `atlasFactory.createRawKeyValueService` doesn't know what type it is so it'll be a raw type....but maybe it'll help implementations understand what's going on

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
jeremyk-91: good spot!

atlasdb-api/src/main/java/com/palantir/atlasdb/spi/AtlasDbFactory.java
```diff
@@ -38,6 +38,13 @@
 
     String getType();
 
+    default MERGED_CONFIG createMergedKeyValueServiceConfig(
```
jeremyk-91, line 41: Could we add a bit of docs saying how this function is expected to be used? In particular, values read off this merged config, even if the values originate from the runtime config, may not necessarily be respected for live reloading.

jeremyk-91 (commented):
🕐 0:18 🏁  🔥 

Broadly makes sense. I think there are some things we need to be explicit about what users of the MERGED_CONFIG can or cannot expect, and we probably want to add a bit more detail to the changelog

gmaretic (approved):
Makes sense 👍 

@jeremyk-91 , feel free to merge if you are fine with it

jeremyk-91 (approved):

PR #5289, opened 2021-03-01
Add migration mode for RETRY_ON_WRITE_WRITE conflict handlers

Jolyon-S:
**Goals (and why)**:
* Enable migration between `RETRY_ON_WRITE_WRITE` and `RETRY_ON_WRITE_WRITE_CELL` conflict handler modes.

* Added new mode to enum;
* Updated `.proto` file.

* Existing tests tend to parameterise on the enum, so should cover the new value automatically.

* Am I missing any important tests or things to update?

`ConflictHandler`

ASAP.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
@@ -114,7 +114,8 @@ public enum ConflictHandler {
      * 1. Upgrade product to V1, supporting schema version SV1 (not supporting downgrades to SV0), and now knows
      * about ROWWM (but does *not* use it).
      * 2. Upgrade product to V2, supporting SV2 (supporting downgrades to SV1), and uses ROWWM.
-     * 3. Upgrade product to V3, supporting SV3 (supporting downgrades to SV2), and uses the desired new strategy
+     * 3. Upgrade product to V3, supporting SV3 (supporting downgrades to SV2 but not SV1), and uses the desired new
+     * strategy
      * (for example, now using ROWWC where it was using ROWW in version v0).
      * <p>
      * Failure to adhere to this flow may cause arbitrary data corruption.
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
@@ -102,7 +102,13 @@ public enum ConflictHandler {
      * cell then correctness issues may occur as there is no way for t1 to be aware of t2. Locking on both cell and row
      * level prevents this issue.
      */
-    SERIALIZABLE_LOCK_LEVEL_MIGRATION(true, true, true, true);
+    SERIALIZABLE_LOCK_LEVEL_MIGRATION(true, true, true, true),
+
+    /**
+     * This conflict handler is analogous to SERIALIZABLE_LOCK_LEVEL_MIGRATION, but for migrating between
+     * RETRY_ON_WRITE_WRITE and RETRY_ON_WRITE_WRITE_CELL.
+     */
+    RETRY_ON_WRITE_WRITE_MIGRATION(true, true, true, false);
 
     private final boolean lockCellsForConflicts;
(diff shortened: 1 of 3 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -102,7 +102,13 @@
...
      * level prevents this issue.
      */
-    SERIALIZABLE_LOCK_LEVEL_MIGRATION(true, true, true, true);
+    SERIALIZABLE_LOCK_LEVEL_MIGRATION(true, true, true, true),
+
+    /**
+     * This conflict handler is analogous to SERIALIZABLE_LOCK_LEVEL_MIGRATION, but for migrating between
+     * RETRY_ON_WRITE_WRITE and RETRY_ON_WRITE_WRITE_CELL.
+     */
+    RETRY_ON_WRITE_WRITE_MIGRATION(true, true, true, false);
```
jeremyk-91, line 111: Probably worth adding some docs around how to actually do this migration, given the version is pretty recent.

atlasdb-client-protobufs/src/main/proto/TableMetadataPersistence.proto
```diff
@@ -108,6 +108,7 @@ enum TableConflictHandler {
     SERIALIZABLE_CELL = 6;
     SERIALIZABLE_INDEX = 7;
     SERIALIZABLE_LOCK_LEVEL_MIGRATION = 8;
+    RETRY_ON_WRITE_WRITE_MIGRATION = 9;
```
jeremyk-91, line 111: The main danger here is that another node reading `RETRY_ON_WRITE_WRITE_MIGRATION` that doesn't know this enum value exists will read `IGNORE_ALL`, which would be bad.

> Notably, unrecognized enum values are discarded when the message is deserialized, which makes the field's `has..` accessor return false and its getter return the first value listed in the enum definition, or the default value if one is specified.

This is fine: we don't have a way of retrofitting changes to old versions though. This means that users of this level (which I understand to largely be atlasdb-proxy) need to follow a careful schema upgrade process. Supposing the current product version is P1 and schema version is S1, you need to

- release (P2, S2) with this version of AtlasDB, where rollback from S2 to S1 is not allowed.
- release (P3, S3) with the table set to RETRY_ON_WRITE_WRITE_MIGRATION, where rollback from S3 to S2 is allowed.
- release (P4, S4) with the table set to RETRY_ON_WRITE_WRITE, where rollback from S4 to S3 is allowed.

This is complex and can go wrong, so we need docs.

I also think we should make the protobuf deserialization code more robust to this (we just hit the getter directly, while if we see it doesn't exist we should explode) but we can't retrofit old code anyway so the above process is required.

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/ConflictHandler.java
```diff
@@ -102,7 +102,25 @@
...
+     * This conflict handler is analogous to SERIALIZABLE_LOCK_LEVEL_MIGRATION, but for migrating between
+     * RETRY_ON_WRITE_WRITE and RETRY_ON_WRITE_WRITE_CELL.
+     * <p>
+     * Migrations between RETRY_ON_WRITE_WRITE (ROWW) and RETRY_ON_WRITE_WRITE_CELL (ROWWC) using
+     * RETRY_ON_WRITE_WRITE_MIGRATION (ROWWM) takes three steps:
+     * 0. Product is on version V0, and schema version SV0, and does not even know about ROWWM.
+     * 1. Upgrade product to V1, supporting schema version SV1 (not supporting downgrades to SV0), and now knows
+     * about ROWWM (but does *not* use it).
+     * 2. Upgrade product to V2, supporting SV2 (supporting downgrades to SV1), and uses ROWWM.
+     * 3. Upgrade product to V3, supporting SV3 (supporting downgrades to SV2), and uses the desired new strategy
```
jeremyk-91, line 117: it's implied, but I'd prefer to be explicit here: supporting downgrades to SV2, but not SV1

jeremyk-91 (commented):
Looks good! One last thing

jeremyk-91 (approved):
👍

PR #5290, opened 2021-03-01
[CCB] | MultiClient-CommitTimestampGetter refactor - part 1

sudiksha27:
**Goals (and why)**:
Refactor CommitTimestampGetter to accommodate the flow for batching getCommitTimestamps requests across multiple clients.

- Extract interface for CommitTimestampGetter - the only implementation right now is the BatchingCommitTimestampGetter which was the default request batching mode per client.
- Does not actually change anything in the current flow.

Existing tests should suffice

Does this refactor break something?

CommitTimestampGetter

EOD today would be great

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
@@ -1,0 +1,112 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
```diff
@@ -0,0 +1,112 @@
...
+import com.palantir.lock.watch.TransactionUpdate;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import org.immutables.value.Value;
+
+final class BatchingCommitTimestampGetter implements CommitTimestampGetter {
```
Jolyon-S, line 39: [Sanity check]: This is the single namespace batching commit timestamp getter, right? In other words, this does _not_ batch between namespaces?

Jolyon-S (commented):
Just a sanity check (because I can never keep up with naming conventions...) but I think this looks pretty much fine otherwise.

lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
sudiksha27: That is correct. Added docs on this class to say the same.

jeremyk-91 (approved):
🕐  0:02 🏁  -:-- Jolyon reviewed

👍  looks good!

PR #5294, opened 2021-03-02
[CCB] | MultiClientCommitTimestampGetter refactor - part 2

sudiksha27:
**Goals (and why)**:
Atlas should be able to accept a common batcher for getCommitTimestamps requests for multiple clients.

- Implements AutoBatcher capable of handling getCommitTimestamps requests for multiple clients
- The model is very close to the origin model - for each namespace, while all requests cannot be served, we batch all the requests together, try to do getCommitTimestamps in one call to the server & update the cache after receiving the response.
- The LockWatchEventCache#lastKnownVersion is fetched before making request to TimeLock and LockWatchEventCache#processGetCommitTimestampsUpdate is invoked as response is received. 
- We try to serve requests as responses become available instead of waiting for all responses to serve all requests in one go.
- In case of exception, all pending requests are cancelled.
- This PR actually does not change the current flow
- I haven't done the wiring yet as the PR is pretty big already

Added tests

- There is code duplication b/w NamespacedBatchStateManager#process and BatchingCommitTimestampGetter#process but I could not think of a way of refactoring without 1. adding another iteration over request list 2. creating new instances of a suitable DS
- Did I miss/ break something horribly?

MultiClientCommitTimestampGetter

Before EOD tomorrow is the dream

### DIFF
--- a/lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
+++ b/lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
@@ -31,19 +31,20 @@ import com.palantir.common.streams.KeyedStream;
 import com.palantir.lock.v2.LockToken;
 import com.palantir.lock.watch.ImmutableTransactionUpdate;
 import com.palantir.lock.watch.LockWatchEventCache;
+import com.palantir.lock.watch.LockWatchVersion;
 import com.palantir.lock.watch.TransactionUpdate;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.LongStream;
 import org.immutables.value.Value;
 
-public class MultiClientCommitTimestampGetter implements AutoCloseable {
+public final class MultiClientCommitTimestampGetter implements AutoCloseable {
     private final DisruptorAutobatcher<NamespacedRequest, Long> autobatcher;
 
     private MultiClientCommitTimestampGetter(DisruptorAutobatcher<NamespacedRequest, Long> autobatcher) {
@@ -68,15 +69,12 @@ public class MultiClientCommitTimestampGetter implements AutoCloseable {
     @VisibleForTesting
     static Consumer<List<BatchElement<NamespacedRequest, Long>>> consumer(
             InternalMultiClientConjureTimelockService delegate) {
-        return batch -> processBatch(delegate, batch);
-    }
-
-    private static void processBatch(
-            InternalMultiClientConjureTimelockService delegate, List<BatchElement<NamespacedRequest, Long>> batch) {
-        BatchStateManager batchStateManager = new BatchStateManager(getNamespaceWiseBatchStateManager(batch));
-        while (batchStateManager.hasPendingRequests()) {
-            batchStateManager.processResponse(delegate.getCommitTimestamps(batchStateManager.getRequests()));
-        }
+        return batch -> {
+            BatchStateManager batchStateManager = new BatchStateManager(getNamespaceWiseBatchStateManager(batch));
+            while (batchStateManager.hasPendingRequests()) {
+                batchStateManager.processResponse(delegate.getCommitTimestamps(batchStateManager.getRequests()));
+            }
+        };
     }
 
     private static Map<Namespace, NamespacedBatchStateManager> getNamespaceWiseBatchStateManager(
@@ -87,7 +85,7 @@ public class MultiClientCommitTimestampGetter implements AutoCloseable {
             NamespacedRequest argument = elem.argument();
             Namespace namespace = argument.namespace();
             NamespacedBatchStateManager namespacedBatchStateManager =
-                    requestMap.computeIfAbsent(namespace, _n -> new NamespacedBatchStateManager(argument.cache()));
+                    requestMap.computeIfAbsent(namespace, _unused -> new NamespacedBatchStateManager(argument.cache()));
             namespacedBatchStateManager.addRequest(elem);
         }
 
@@ -101,18 +99,18 @@ public class MultiClientCommitTimestampGetter implements AutoCloseable {
             this.requestMap = requestMap;
         }
 
-        boolean hasPendingRequests() {
+        private boolean hasPendingRequests() {
             return requestMap.values().stream().anyMatch(NamespacedBatchStateManager::hasPendingRequests);
         }
 
-        Map<Namespace, GetCommitTimestampsRequest> getRequests() {
+        private Map<Namespace, GetCommitTimestampsRequest> getRequests() {
             return KeyedStream.stream(requestMap)
                     .filter(NamespacedBatchStateManager::hasPendingRequests)
                     .map(NamespacedBatchStateManager::getPendingRequest)
                     .collectToMap();
         }
 
-        void processResponse(Map<Namespace, GetCommitTimestampsResponse> namespaceWiseResponse) {
+        private void processResponse(Map<Namespace, GetCommitTimestampsResponse> namespaceWiseResponse) {
             KeyedStream.stream(namespaceWiseResponse)
                     .forEach((namespace, getCommitTimestampsResponse) ->
                             requestMap.get(namespace).serviceRequests(getCommitTimestampsResponse));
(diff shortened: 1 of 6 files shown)


lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+            List<BatchElement<NamespacedRequest, Long>> batch) {
+        Map<Namespace, NamespacedBatchStateManager> requestMap = new HashMap<>();
+
+        for (BatchElement<NamespacedRequest, Long> elem : batch) {
+            NamespacedRequest argument = elem.argument();
+            Namespace namespace = argument.namespace();
+            NamespacedBatchStateManager namespacedBatchStateManager =
+                    requestMap.computeIfAbsent(namespace, _unused -> new NamespacedBatchStateManager(argument.cache()));
+            namespacedBatchStateManager.addRequest(elem);
+        }
```
jeremyk-91, line 90: This function makes sense, but I'd expect to live within the BatchStateManager as opposed to here.

lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    private static final class NamespacedBatchStateManager {
+        private final Queue<BatchElement<NamespacedRequest, Long>> pendingRequestQueue;
+        private final LockWatchEventCache cache;
+        private Optional<LockWatchVersion> lastKnownVersion;
+
+        private NamespacedBatchStateManager(LockWatchEventCache cache) {
+            this.pendingRequestQueue = new LinkedList<>();
```
jeremyk-91, line 126: Might have missed this in the previous review, but `ArrayDeque` is probably a better choice: https://docs.oracle.com/javase/tutorial/collections/implementations/deque.html

lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+
+        private boolean hasPendingRequests() {
+            return !pendingRequestQueue.isEmpty();
+        }
+
+        private void addRequest(BatchElement<NamespacedRequest, Long> elem) {
+            pendingRequestQueue.add(elem);
+        }
+
+        private GetCommitTimestampsRequest getPendingRequest() {
```
jeremyk-91, line 139: nit: `getRequestForServer`?

lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        private Map<Namespace, GetCommitTimestampsRequest> getRequests() {
+            return KeyedStream.stream(requestMap)
+                    .filter(NamespacedBatchStateManager::hasPendingRequests)
+                    .map(NamespacedBatchStateManager::getPendingRequest)
+                    .collectToMap();
+        }
+
+        private void processResponse(Map<Namespace, GetCommitTimestampsResponse> namespaceWiseResponse) {
```
jeremyk-91, line 113: as before: the type indicates this is namespace-keyed.

lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+            KeyedStream.stream(namespaceWiseResponse)
+                    .forEach((namespace, getCommitTimestampsResponse) ->
```
jeremyk-91, line 115: nit: `Map` has a direct `forEach`, so I don't think you need to copy it

lock-api/src/main/java/com/palantir/lock/client/MultiClientCommitTimestampGetter.java
```diff
@@ -0,0 +1,193 @@
...
+                            requestMap.get(namespace).serviceRequests(getCommitTimestampsResponse));
+        }
+    }
+
+    private static final class NamespacedBatchStateManager {
```
jeremyk-91, line 120: Confirmed core logic is correct. Not too satisfied with the name, but don't have a better solution right now.

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import one.util.streamex.StreamEx;
+import org.junit.Test;
+
+public class MultiClientCommitTimestampGetterTest {
+    private static final Map<Namespace, Long> LOWEST_START_TS_MAP = new HashMap<>();
+    private static final int COMMIT_TS_LIMIT_PER_REQUEST = 5;
+    private static final SafeIllegalStateException EXCEPTION = new SafeIllegalStateException("Something went wrong!");
+    private static final Map<Namespace, LockWatchEventCache> LOCK_WATCH_EVENT_CACHE_MAP_MAP = new HashMap<>();
```
jeremyk-91, line 61: private **static** final for this and the `LOWEST_START_TS_MAP` seems a bit odd, this means that `LockWatchEventCache`s will be reused across test methods. Do you need that?

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+        LockWatchEventCache alphaCache = LOCK_WATCH_EVENT_CACHE_MAP_MAP.get(alpha);
+        verify(alphaCache).lastKnownVersion();
+        verify(alphaCache).processGetCommitTimestampsUpdate(any(), any());
+
+        // assert unsuccessful requests were made by client beta
+        assertThat(requestForBeta.result().isDone()).isFalse();
+
+        LockWatchEventCache betaCache = LOCK_WATCH_EVENT_CACHE_MAP_MAP.get(beta);
+        verify(betaCache, times(2)).lastKnownVersion();
+        verify(betaCache, never()).processGetCommitTimestampsUpdate(any(), any());
```
jeremyk-91, line 136: 👍

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+            List<BatchElement<NamespacedRequest, Long>> batch,
+            Map<Namespace, List<GetCommitTimestampsResponse>> expectedResponseMap) {
+        // all requests should be served
+        assertThat(batch.stream().filter(elem -> !elem.result().isDone()).collect(Collectors.toSet()))
+                .isEmpty();
+
+        Map<Namespace, List<Long>> partitionedResponseMap = batch.stream()
+                .collect(groupingBy(
+                        elem -> elem.argument().namespace(),
+                        Collectors.mapping(elem -> Futures.getUnchecked(elem.result()), toList())));
```
jeremyk-91, line 167: nice!

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+        consumer.accept(batch);
+        assertSanityOfResponse(batch, expectedResponseMap);
+    }
+
+    private void assertSanityOfResponse(
+            List<BatchElement<NamespacedRequest, Long>> batch,
+            Map<Namespace, List<GetCommitTimestampsResponse>> expectedResponseMap) {
+        // all requests should be served
+        assertThat(batch.stream().filter(elem -> !elem.result().isDone()).collect(Collectors.toSet()))
+                .isEmpty();
```
jeremyk-91, line 162: It's usually worth having an `as(...)` clause as opposed to a comment here, to make errors more readable when a test fails

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    private static void assertCorrectnessOfCompletedRequests(
+            Map<Namespace, List<GetCommitTimestampsResponse>> expectedResponseMap,
+            Map<Namespace, List<Long>> partitionedResponseMap) {
+        KeyedStream.stream(partitionedResponseMap)
+                .forEach((namespace, commitTsList) ->
+                        assertSanityOfServedTimestamps(commitTsList, expectedResponseMap.get(namespace)));
+    }
+
+    private static void assertSanityOfServedTimestamps(
```
jeremyk-91, line 181: nit: can we use `correctness` instead of `sanity`?

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+            List<Long> commitTsList, List<GetCommitTimestampsResponse> commitTimestampsResponses) {
```
jeremyk-91, line 182: the 2nd argument should be `expectedCommitTimestampsResponses`. I'd suggest switching them around for consistency with `assertCorrectnessOfCompletedRequests`

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+        long requestedCommitTsCount = commitTimestampsResponses.stream()
+                .mapToLong(resp -> resp.getInclusiveUpper() - resp.getInclusiveLower() + 1)
+                .sum();
+        assertThat(requestedCommitTsCount).isEqualTo(commitTsList.size());
+        assertThat(ImmutableSet.copyOf(commitTsList)).hasSameSizeAs(commitTsList);
+        assertThat(StreamEx.of(commitTsList)
+                        .pairMap((first, second) -> first > second)
+                        .anyMatch(x -> x))
+                .isFalse();
```
jeremyk-91, line 191: Would suggest using `as(...)` clauses to outline what each assertion means. They are, in order, if I'm not mistaken

1. we got as many commit timestamps as we asked for
2. commit timestamps we got have no duplicates
3. responses are in increasing order of timestamp (technically this isn't required given this is a single timestamp batch I think, but it's fine)

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+        when(timelockService.getCommitTimestamps(any())).thenReturn(responseMap).thenThrow(EXCEPTION);
+
+        assertThatThrownBy(() -> consumer.accept(allRequests)).isEqualTo(EXCEPTION);
+
+        // assert requests made by client alpha are served
+        assertSanityOfResponse(alphaRequestList, ImmutableMap.of(alpha, ImmutableList.of(responseMap.get(alpha))));
+
+        LockWatchEventCache alphaCache = LOCK_WATCH_EVENT_CACHE_MAP_MAP.get(alpha);
+        verify(alphaCache).lastKnownVersion();
```
jeremyk-91, line 128: nit: do we care that this is exactly 1? Feels like a bit of an implementation detail

lock-api/src/test/java/com/palantir/lock/client/MultiClientCommitTimestampGetterTest.java
```diff
@@ -0,0 +1,251 @@
...
+        verify(alphaCache).processGetCommitTimestampsUpdate(any(), any());
+
+        // assert unsuccessful requests were made by client beta
+        assertThat(requestForBeta.result().isDone()).isFalse();
+
+        LockWatchEventCache betaCache = LOCK_WATCH_EVENT_CACHE_MAP_MAP.get(beta);
+        verify(betaCache, times(2)).lastKnownVersion();
```
jeremyk-91, line 135: as above, does it matter that this is 2?

jeremyk-91 (commented):
🕐 0:45 🏁 0:58

The core logic does look equivalent to a valid parallel execution of the current BatchingCommitTimetsampGetters. My comments here are largely around factoring.

lock-api/src/main/java/com/palantir/lock/client/MultiClientTransactionStarter.java
```diff
@@ -278,8 +278,8 @@ public void close() {
         private final LockCleanupService lockCleanupService;
 
         ResponseHandler(LockCleanupService lockCleanupService) {
-            this.pendingFutures = new LinkedList<>();
-            this.transientResponseList = new LinkedList<>();
+            this.pendingFutures = new ArrayDeque<>();
+            this.transientResponseList = new ArrayDeque<>();
```
jeremyk-91, line 282: ❤️

jeremyk-91 (approved):
👍  looks good, thanks for fixing all the comments!

PR #5296, opened 2021-03-03
[CCB] | MultiClient CommitTimestampGetter refactor - part 3

sudiksha27:
**Goals (and why)**:
AtlasDb accepts external request batcher for getCommitTimestamp requests.

- Have a common factory - RequestBatchers for getting TransactionStarter and CommitTimestampGetter

Tested locally

Does the design look okay?

RequestBatchers

Before EOD tomorrow please


lock-api/src/main/java/com/palantir/lock/client/RequestBatchers.java
```diff
@@ -0,0 +1,76 @@
...
+import com.palantir.lock.watch.NoOpLockWatchEventCache;
+import com.palantir.lock.watch.StartTransactionsLockWatchEventCache;
+import java.util.Optional;
+
+public final class RequestBatchers {
+    private final LockWatchEventCache lockWatchEventCache;
+    private final StartTransactionsLockWatchEventCache startTransactionsLockWatchEventCache;
+    private final Optional<Namespace> namespace;
+    private final Optional<MultiClientTransactionStarter> maybeTransactionStarter;
+    private final Optional<MultiClientCommitTimestampGetter> maybeCommitTimestampGetter;
```
jeremyk-91, line 30: Is there a way of avoiding this? / would we expect correlations? It feels like either all 3 will be present or absent - if that's the case it's probably better to have an `Optional<T>` where `T` is some type that covers all of these three objects.

lock-api/src/main/java/com/palantir/lock/client/RequestBatchers.java
```diff
@@ -0,0 +1,76 @@
...
+                new LockCleanupService(lockLeaseService));
+    }
+
+    public CommitTimestampGetter getBatchingCommitTimestampGetter(LockLeaseService lockLeaseService) {
+        if (!namespace.isPresent() || !maybeCommitTimestampGetter.isPresent()) {
+            return BatchingCommitTimestampGetter.create(lockLeaseService, lockWatchEventCache);
+        }
+        return new NamespacedCommitTimestampGetter(
+                lockWatchEventCache, namespace.get(), maybeCommitTimestampGetter.get());
+    }
```
jeremyk-91, line 75: Hmm, I think we need to clarify this a bit. Either

1. name these methods `create` rather than `get`, or
2. change the methods so that once an instance has been created for a given lock-lease service it always returns the same one.

`get` suggests that this is idempotent, but it's not: if you call `getBatchingTransactionStarter` twice with the same lock-lease service, you will get _two_ autobatchers!

lock-api/src/main/java/com/palantir/lock/client/RequestBatchers.java
jeremyk-91: Looking at the usage it seems you want a non-optional namespace (just create a fake one for the tests) and an optional of the 2 multi-client objects.

jeremyk-91 (commented):
🕐  0:07 🏁  0:00 (remember that this clock only counts business hours)

Broadly looks good! Couple of suggestions

lock-api/src/main/java/com/palantir/lock/client/MultiClientRequestBatchers.java
```diff
@@ -0,0 +1,36 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.client;
+
+public final class MultiClientRequestBatchers {
```
jeremyk-91, line 19: this is fine, but can we just use immutables?

jeremyk-91 (commented):
Almost there! One question

lock-api/src/main/java/com/palantir/lock/client/MultiClientRequestBatchers.java
sudiksha27: Done

jeremyk-91 (approved):
👍  looks good!

PR #5297, opened 2021-03-03
fix ranges concurrency when specifying cassandra servers in runtime conf

pnepywoda:


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfig.java
```diff
@@ -115,4 +115,17 @@ public int concurrentGetRangesThreadPoolSize() {
...
+    @Override
+    public int defaultGetRangesConcurrency() {
+        if (config.servers().numberOfThriftHosts() > 0) {
+            return config.defaultGetRangesConcurrency();
+        }
+
+        // Use the initial number of thrift hosts as a best guess
+        return runtimeConfigSupplier
+                .get()
+                .map(_runtime -> poolSize() * servers().numberOfThriftHosts())
```
pkoenig10, line 128: I think this needs to be
```suggestion
                .map(_runtime -> Math.min(8, concurrentGetRangesThreadPoolSize() / 2))
```

atlasdb-cassandra/src/test/java/com/palantir/atlasdb/cassandra/CassandraReloadableKvsConfigTest.java
```diff
@@ -121,6 +121,22 @@ public void ifInstallServersEmpty_resolvesToRuntimeConfig() {
         assertThat(reloadableConfig.concurrentGetRangesThreadPoolSize()).isEqualTo(84);
     }
 
+    @Test
+    public void ifInstallServersEmpty_resolvesToRuntimeConfigForRangesConcurrency() {
```
pkoenig10, line 125: This needs to be tested for all the cases.

Instead of new tests, just add assertions to the existing tests.

jeremyk-91 (approved):

PR #5299, opened 2021-03-04
TimeLock Product dependency upgrade

sudiksha27:
**Goals (and why)**:
Required for internal Atlas Proxy to enable use of multi-client getCommitTimestamps

🏃‍♂️


Jolyon-S (approved):
I have **not** verified that this is necessarily the right version, but the syntax is definitely fine!

PR #5308, opened 2021-03-09
allow hikari password changes at runtime

berler:
**Goals (and why)**: Many customers have password policies that require regular password rotation for Oracle database users (often every 60 days). Currently since the password is in static config, this requires restarting services that use AtlasDB or rely on the Hikari pool. Ideally we should support being able to change user passwords without requiring services restart.

* Hikari internally already supports changing the password of a pool at runtime without needing to shutdown or reinitialize the pool (it also supports changing the username, but we won't allow that for now).
* Adds a setPassword method on the ConnectionManager interface which can be used to change the password of the pool. Note that the implementation here needs to do something different depending on if the pool is already initialized or not. Also note that the way hikari implements this, changing the password (or username) means that it would now copy the JDBC properties object every time a new connection is called. This hurts performance very slightly, but only for new connections (reusing an existing connection is not affected). The way I've implemented it here, you only pay this performance cost if you *actually* call setPassword at runtime (existing consumers will not be affected).
* (NOT YET IMPLEMENTED) A future change can add a runtime config for postgres and oracle so that the hikari pool setPassword method gets called whenever the password changes in runtime config. This is not necessary for many of our internal products though, because internal products often do not use the normal atlas config (they have their own completely independent install/runtime configs).

**Testing (What was existing testing like?  What have you done to improve it?)**: Added tests against postgres. Also fixed the tests (HikariCpConnectionManagerTest was never being run because it was not in a suite, and the setup for that class was broken because it didn't actually wait for the database to be available which meant that even running it manually would have failed).

### DIFF
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
@@ -23,6 +23,7 @@ import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import com.palantir.nexus.db.DBType;
 import com.palantir.nexus.db.pool.config.ConnectionConfig;
 import com.palantir.nexus.db.sql.ExceptionCheck;
+import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import com.zaxxer.hikari.HikariPoolMXBean;
 import com.zaxxer.hikari.pool.HikariPool.PoolInitializationException;
@@ -52,7 +53,8 @@ import org.slf4j.LoggerFactory;
 public class HikariCPConnectionManager extends BaseConnectionManager {
     private static final Logger log = LoggerFactory.getLogger(HikariCPConnectionManager.class);
 
-    private ConnectionConfig connConfig;
+    private final ConnectionConfig connConfig;
+    private final HikariConfig hikariConfig;
 
     private enum StateType {
         // Base state at construction.  Nothing is set.
@@ -83,6 +85,7 @@ public class HikariCPConnectionManager extends BaseConnectionManager {
 
     public HikariCPConnectionManager(ConnectionConfig connConfig) {
         this.connConfig = Preconditions.checkNotNull(connConfig, "ConnectionConfig must not be null");
+        this.hikariConfig = connConfig.getHikariConfig();
     }
 
     @Override
@@ -293,17 +296,15 @@ public class HikariCPConnectionManager extends BaseConnectionManager {
 
         try {
             try {
-                dataSourcePool = new HikariDataSource(connConfig.getHikariConfig());
+                dataSourcePool = new HikariDataSource(hikariConfig);
             } catch (IllegalArgumentException e) {
                 // allow multiple pools on same JVM (they need unique names / endpoints)
                 if (e.getMessage().contains("A metric named")) {
                     String poolName = connConfig.getConnectionPoolName();
 
-                    connConfig
-                            .getHikariConfig()
-                            .setPoolName(
-                                    poolName + "-" + ThreadLocalRandom.current().nextInt());
-                    dataSourcePool = new HikariDataSource(connConfig.getHikariConfig());
+                    hikariConfig.setPoolName(
+                            poolName + "-" + ThreadLocalRandom.current().nextInt());
+                    dataSourcePool = new HikariDataSource(hikariConfig);
                 } else {
                     throw e;
                 }
@@ -362,6 +363,50 @@ public class HikariCPConnectionManager extends BaseConnectionManager {
     @Override
     public DBType getDbType() {
         return connConfig.getDbType();
+    }
+
+    @Override
+    public void setPassword(String newPassword) {
+        Preconditions.checkNotNull(newPassword, "password cannot be null");
+        State localState = state;
+        if (localState.type == StateType.ZERO) {
+            synchronized (this) {
+                // need to check again in case init just happened
+                localState = state;
+                if (localState.type == StateType.ZERO) {
+                    // the pool is not initialized, need to set password on the config object
+                    internalSetPassword(newPassword, hikariConfig);
+                    return;
+                }
+            }
+        }
+        if (localState.type == StateType.NORMAL) {
+            // pool is initialized, need to set password directly on the pool
+            internalSetPassword(newPassword, localState.dataSourcePool);
+        } else if (localState.type == StateType.CLOSED) {
+            throw new SafeRuntimeException("Hikari pool is closed", localState.closeTrace);
+        }
+    }
+
+    /**
+     * Sets the password (and username) on the HikariConfig (or HikariDataSource pool) if it is necessary.
+     *
+     * <p>Note that when the password matches static config, performance is better when setting the username and
+     * password to null so that hikari uses the original JDBC parameters and avoids making new copies for every
+     * new connection. See {@link com.zaxxer.hikari.pool.PoolBase#newConnection}. Also see
+     * {@link com.palantir.nexus.db.pool.config.OracleConnectionConfig#getHikariProperties} for the JDBC properties.
+     *
+     * <p>Hikari checks the username field for being null to determine if it needs to copy the JDBC parameters, so
+     * the username *must* be set for it to respect the password change.
+     */
+    private void internalSetPassword(String newPassword, HikariConfig configOrPool) {
+        if (newPassword.equals(connConfig.getDbPassword().unmasked())) {
+            configOrPool.setUsername(null);
+            configOrPool.setPassword(null);
+        } else {
+            configOrPool.setPassword(newPassword);
+            configOrPool.setUsername(connConfig.getDbLogin());
+        }
     }
 
     private final class ConnectionAcquisitionProfiler {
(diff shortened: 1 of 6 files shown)


atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
```diff
@@ -364,6 +365,50 @@ public DBType getDbType() {
...
 
+    @Override
+    public void setPassword(String newPassword) {
+        Preconditions.checkNotNull(newPassword, "password cannot be null");
+        State localState = state;
+        if (localState.type == StateType.ZERO) {
+            synchronized (this) {
+                // need to check again in case init just happened
+                localState = state;
+                if (localState.type == StateType.ZERO) {
```
Jolyon-S, line 376: While I believe that this is fine (since the only `State` with type set to `ZERO` has all nulls, and so checking the type is equivalent to checking the whole object), the method `checkAndGetDataSourcePool` performs a similar operation in the sense that it switches against the possible states that we can be in, and notably it compares the whole object vs just the type.

To err on the side of caution, can we follow the structure used there? In other words:
1. use a switch statement
2. compare `state` against `localState` directly

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
```diff
@@ -364,6 +365,50 @@ public DBType getDbType() {
...
+                    // the pool is not initialized, need to set password on the config object
+                    internalSetPassword(newPassword, hikariConfig);
+                    return;
+                }
+            }
+        }
+        if (localState.type == StateType.NORMAL) {
+            // pool is initialized, need to set password directly on the pool
+            internalSetPassword(newPassword, localState.dataSourcePool);
```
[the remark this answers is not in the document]
Jolyon-S: Equally, what happens (and can it happen?) if there is a race to change the password - is it possible that the username gets set to `null` but the password is set, or vice versa?

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
```diff
@@ -364,6 +365,50 @@ public DBType getDbType() {
...
+     *
+     * <p>Note that when the password matches static config, performance is better when setting the username and
+     * password to null so that hikari uses the original JDBC parameters and avoids making new copies for every
+     * new connection. See {@link com.zaxxer.hikari.pool.PoolBase#newConnection}. Also see
+     * {@link com.palantir.nexus.db.pool.config.OracleConnectionConfig#getHikariProperties} for the JDBC properties.
+     *
+     * <p>Hikari checks the username field for being null to determine if it needs to copy the JDBC parameters, so
+     * the username *must* be set for it to respect the password change.
+     */
+    private void internalSetPassword(String newPassword, HikariConfig configOrPool) {
```
Jolyon-S, line 402: As above, I'm a little concerned about concurrency (but I might be being paranoid here).

atlasdb-dbkvs-tests/src/test/java/com/palantir/nexus/db/pool/HikariCpConnectionManagerTest.java
```diff
@@ -141,6 +130,103 @@ public void testCloseIdempotent() throws SQLException {
         manager.close(); // shouldn't throw
     }
 
+    @Test
+    public void testRuntimePasswordChange() throws SQLException {
```
Jolyon-S, line 134: From looking at the base class and the tests, it seems rather difficult to write a test that confirms that attempts to concurrently edit the password don't cause an issue; still, if at all possible, it'd be quite interesting to see.

Jolyon-S (commented):
I've done a preliminary review; @jeremyk-91 will give his own pass here (I'm trying to familiarise myself with parts of the codebase I don't know, so apologies if I've misunderstood this codepath at all).

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
berler: I thought about that, but that is actually suboptimal (and would result in needing to write duplicate code). When you examine the state again (while being synchronized) if the state happens to be the normal state, you would actually want to have the same behavior as what is outside of the synchronized block). It would also mean that you potentially need to add a third volatile read instead of what is written here where it is guaranteed a max of two volatile memory reads.

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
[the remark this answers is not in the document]
berler: This doesn't need to be synchronized because the only other possible state change is closing the pool. That is completely fine to happen concurrently because all this does is update a volatile field on the pool (it is fine to update that concurrently/after the pool is closed).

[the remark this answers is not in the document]
berler: Also it is completely fine for the username to be null while the password is not null (the password field would just be ignored in that case which means it will use the original configured password). This would only happen with a specific execution order of concurrent calls where one setPassword passes in the original password and another passes in a different password.

So consider the following two **concurrent** calls where the static config has the password set as "originalPassword":

1. setPassword("originalPassword")
2. setPassword("newPassword")

The possible outcomes of this are as follows for the username/password override fields:

1. username = null; password = null; => "originalPassword" used for connections
2. username = null; password = "newPassword"; => "originalPassword" used for connections
3. username = "username"; password = "newPassword"; => "newPassword" used for connections (JDBC parameters copied for all new connections)
4. username = "username"; password = null; => "originalPassword" used for connections (although in this case the JDBC parameters are copied for all new connections)

Regardless of the outcome, the behavior is arguably correct (you can say one of the calls "wins" and that is what the end behavior is). There is also no point in time where any password would be used other than one of the passwords from the concurrent calls.

In practice there should not be concurrent calls to this. The only caller should be a subscription on the witchcraft Runtime config object. Witchcraft uses a single task on a schedule using scheduleWithFixedDelay to check the runtime config and calls the subscribers on the same thread (so no Runtime config subscription calls will ever overlap).

The only thing we need to really worry about is other threads that may call init() or close(), but that is handled correctly by the current code which checks the state.

atlasdb-dbkvs-tests/src/test/java/com/palantir/nexus/db/pool/HikariCpConnectionManagerTest.java
berler: I'm not sure that a test would be able to confirm correctness unless you somehow can control the execution order on different threads and iterate through all possible execution orders.

I think it's fine to have undefined behavior as the only caller to this should be a witchcraft Runtime subscription which is guaranteed to never have multiple concurrent calls.

If someone does end up having concurrent calls and ends up in an unexpected state, either the password is correct or not. If the password is not correct, all connections would fail and they'd probably need to restart the service (or have something which triggers calling setPassword again later). For this case I would argue the bug is in the service which made concurrent calls to the setPassword method.

Jolyon-S:
Seems reasonable in light of what you've said - thanks for providing additional context around these methods and changes.

I'll leave it to @jeremyk-91 for the +2 before we merge.

Thanks!

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
```diff
@@ -364,6 +365,50 @@ public DBType getDbType() {
...
+        if (newPassword.equals(connConfig.getDbPassword().unmasked())) {
+            configOrPool.setUsername(null);
+            configOrPool.setPassword(null);
```
jeremyk-91, line 405: thanks for the comment above, that explains this part well

atlasdb-dbkvs-tests/src/test/java/com/palantir/nexus/db/pool/HikariCpConnectionManagerTest.java
jeremyk-91: yeah, in sync here that trying to setPassword concurrently (as long as we're not doing it) shouldn't be allowed. Maybe we could stick a `@NotThreadSafe` or some javadocs there noting that that's a bad idea?

commons-db/src/main/java/com/palantir/nexus/db/pool/ConnectionManager.java
```diff
@@ -51,4 +51,12 @@
     void initUnchecked();
 
     DBType getDbType();
+
+    /**
+     * Change the password which will be used for new connections. Existing connections are not modified in any way.
+     *
+     * <p>Note that if the password is changed at runtime, there is a slight negative impact on pool performance
+     * because Hikari will copy the JDBC parameters every time a new connection is made.
```
jeremyk-91, line 59: yeah let's say concurrent calls to this method = UB

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/HikariCPConnectionManager.java
jeremyk-91: I'd hope that the perf of this method isn't that critical (though now that you've written this more efficient impl it makes sense, of course)!

jeremyk-91 (approved):
👍 
Broadly looks good. I'd suggest some minor docs changes, mainly to tell people more explicitly to avoid calling `setPassword()` concurrently, but the impl looks solid.

PR #5311, opened 2021-03-11
[PDS-146088] Set default timeout to give up on evicting C* connections

Jolyon-S:
**Goals (and why)**:
* Closing C* connections fail when certain C* nodes are rolled; these threads hang forever and pollute the pool.
* We analysed the p999 for how long it takes to close a connection and it is <5ms; therefore defaulting the timeout to 10 seconds should not affect anything but the broken connections.

* Set default to 10 seconds. Config field no longer Optional because it sometimes complains if you have a default optional.

N/A.

-

CassandraKeyValueServiceConfig, probably

ASAP


jeremyk-91 (approved):
🕐 0:03 🏁 ????

PR #5314, opened 2021-03-12
"If migration has not run before" is confusing

jeremyk-91:
**Goals (and why)**:
- Have clear log messages.

- replace "Starting migration for namespace and use case {} if migration has not run before." with log lines chosen to match the context

none 🔥 

- is this too noisy?


gmaretic (approved):
LGTM

sudiksha27 (approved):

PR #5316, opened 2021-03-15
[CCB] | Fix endpoint names

sudiksha27:
**Goals (and why)**:
We introduced multi client endpoints with the same name as single namespace TimeLock endpoints - getCommitTimestamps, startTransactions.
This has caused metrics published for the two different endpoints to be mixed up.

timelock-api

yesterday 🔥

### DIFF
--- a/timelock-api/src/main/conjure/timelock-api.yml
+++ b/timelock-api/src/main/conjure/timelock-api.yml
@@ -199,6 +199,8 @@ services:
           Version of ConjureTimelockService#leaderTime endpoint for acquiring leaderTimes for a set of namespaces.
       startTransactions:
         http: POST /sts
+        deprecated: |
+          This endpoint is deprecated. Please use {@link #startTransactionsForClients} to start transactions for multiple clients.
         args:
           requests: map<Namespace, ConjureStartTransactionsRequest>
         returns: map<Namespace, ConjureStartTransactionsResponse>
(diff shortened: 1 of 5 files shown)


Jolyon-S (commented):
This looks fine on the surface, but are there any scary things we should be aware of given that we're changing endpoint? It looks fine, but just wondering if there is something that we're missing (because this seems too easy!)

sudiksha27:
This PR just introduces the endpoints - we'll have to bump the TL product dep when we update the client to use these. That will be a separate PR.

timelock-api/src/main/conjure/timelock-api.yml
```diff
@@ -199,13 +199,31 @@ services:
...
+          This endpoint is deprecated. Please use {@link #startTransactionsForClients} to start transactions for multiple clients.
         args:
           requests: map<Namespace, ConjureStartTransactionsRequest>
         returns: map<Namespace, ConjureStartTransactionsResponse>
         docs: |
           Version of ConjureTimelockService#startTransactions that starts transactions for multiple namespaces.
       getCommitTimestamps:
         http: POST /gcts
+        deprecated: |
+          This endpoint is deprecated. Please use {@link #getCommitTimestampsForClients} to get commitTimestamps for multiple clients.
```
jeremyk-91, line 212: nit: commit timestamps should be two words?

PR #5318, opened 2021-03-17
[getSortedColumns] | snapshot transactions

sudiksha27:
Ref - #5092
**Goals (and why)**:
Implements getSortedColumns endpoint
- This endpoint returns an iterator over cell-value pairs within specified table for the specified rows, where columns fall within the provided BatchColumnRangeSelection
- The returned iterator is guaranteed to contain cells sorted first in order of column, then in order of input rows

- row wise RowColumnRangeIterator's are fetched and merged - column wise first, then preserving input row order (if stable ordering)

- Lock validations are carried batch-wise on sorted and merged remote entries
- local entries fetched for each row and merged - column wise first, then preserving input row order (if stable ordering)
- iterator over merged remote and local entries is returned

Tests added

- Correctness -> Is the sorting right? Also, are we doing the conflict checking correctly?
- Perf 

SnapshotTransaction#getSortedColumns

Before EOD tomorrow please

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -20,6 +20,7 @@ import static java.util.stream.Collectors.toList;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Timer;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Functions;
 import com.google.common.base.MoreObjects;
@@ -465,7 +466,9 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
         Comparator<Cell> cellComparator = columnOrderThenPreserveInputRowOrder(distinctRows);
 
         Iterator<Map.Entry<Cell, Value>> postFilterIterator = getRowColumnRangePostFilteredWithoutSorting(
-                tableRef, mergeByComparator(rawResults.values(), cellComparator), batchColumnRangeSelection.getBatchHint());
+                tableRef,
+                mergeByComparator(rawResults.values(), cellComparator),
+                batchColumnRangeSelection.getBatchHint());
         Iterator<Map.Entry<Cell, byte[]>> remoteWrites = Iterators.transform(
                 postFilterIterator,
                 entry -> Maps.immutableEntry(entry.getKey(), entry.getValue().getContents()));
@@ -500,7 +503,12 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 cellComparator);
     }
 
-    private static Comparator<Cell> columnOrderThenPreserveInputRowOrder(Iterable<byte[]> rows) {
+    /**
+     * Provides comparator to sort cells by columns (sorted lexicographically on byte ordering) and then in the order
+     * of input rows.
+     * */
+    @VisibleForTesting
+    static Comparator<Cell> columnOrderThenPreserveInputRowOrder(Iterable<byte[]> rows) {
         return Cell.COLUMN_COMPARATOR.thenComparing(
                 (Cell cell) -> ByteBuffer.wrap(cell.getRowName()),
                 Ordering.explicit(Streams.stream(rows)
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -439,7 +439,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
 
     @Override
     public Iterator<Map.Entry<Cell, byte[]>> getSortedColumns(
-            TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection columnRangeSelection) {
+            TableReference tableRef, Iterable<byte[]> rows, BatchColumnRangeSelection batchColumnRangeSelection) {
         checkGetPreconditions(tableRef);
         if (Iterables.isEmpty(rows)) {
             return Collections.emptyIterator();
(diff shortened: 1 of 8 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -434,6 +437,96 @@ public void markTableInvolved(TableReference tableRef) {
...
+    private Comparator<Cell> columnOrderThenPreserveInputRowOrder(Iterable<byte[]> rows) {
+        return Cell.columnComparator.thenComparing(
+                (Cell cell) -> ByteBuffer.wrap(cell.getRowName()),
+                Ordering.explicit(Streams.stream(rows)
+                        .map(ByteBuffer::wrap)
+                        .distinct()
+                        .collect(ImmutableList.toImmutableList())));
+    }
+
+    protected <V> Iterator<Map.Entry<Cell, V>> mergeColumnFirst(
```
jeremyk-91, line 504: can this be private?

also, nit: mergeByComparator(...) or something, that's controlled by what you passed in (which may not be column first)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -434,6 +437,96 @@ public void markTableInvolved(TableReference tableRef) {
...
+        Iterator<Map.Entry<Cell, byte[]>> remoteWrites = Iterators.transform(
+                postFilterIterator,
+                entry -> Maps.immutableEntry(entry.getKey(), entry.getValue().getContents()));
+        Iterator<Map.Entry<Cell, byte[]>> localWrites =
+                getSortedColumnsLocalWrites(tableRef, distinctRows, columnRangeSelection, cellComparator);
+        Iterator<Map.Entry<Cell, byte[]>> merged = IteratorUtils.mergeIterators(
+                localWrites,
+                remoteWrites,
+                Comparator.comparing(Map.Entry::getKey, cellComparator),
+                com.palantir.util.Pair::getLhSide);
```
jeremyk-91, line 478: This is a fairly common pattern (it's used twice more in this class: once in getRowsColumnRange and once in mergeInLocalWrites). You'll probably need to allow the user to pass the comparator in, but the logical piece of "when you need to merge, local writes win" is something that can be shared across the iterators.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -525,6 +613,29 @@ public void markTableInvolved(TableReference tableRef) {
...
+                return Collections.emptyIterator();
+            }
+            Map<Cell, Value> postFiltered = ImmutableMap.copyOf(getWithPostFilteringSync(tableRef, raw, x -> x));
+            return postFiltered.entrySet().iterator();
+        }));
+    }
+
+    private Map<Cell, Value> validateBatch(
+            TableReference tableRef,
+            @org.checkerframework.checker.nullness.qual.Nullable List<Map.Entry<Cell, Value>> batch) {
```
jeremyk-91, line 630: nit: just use the normal `@Nullable`

jeremyk-91: ... actually, `batch` shouldn't be null so I think we can remove this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -525,6 +613,29 @@ public void markTableInvolved(TableReference tableRef) {
...
+        ImmutableMap.Builder<Cell, Value> rawBuilder = ImmutableMap.builder();
+        batch.forEach(rawBuilder::put);
+        Map<Cell, Value> raw = rawBuilder.build();
+
+        validatePreCommitRequirementsOnReadIfNecessary(tableRef, getStartTimestamp());
+        return raw;
+    }
```
jeremyk-91, line 637:
```suggestion 
    private Map<Cell, Value> validateBatch(
            TableReference tableRef,
            List<Map.Entry<Cell, Value>> batch) {
        validatePreCommitRequirementsOnReadIfNecessary(tableRef, getStartTimestamp());
        return ImmutableMap.copyOf(batch);
    }
```

I _think_ this is equivalent 😅

jeremyk-91 (commented):
🕐  1:05 🏁  1:05

We paired on this. I think the broad structure is correct, it's clear you understand how the core of the feature is supposed to work. Some stylistic suggestions.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1684,6 +1692,185 @@ public void orphanedSentinelsDoNotGetDeletedAfterDetectionInConservativelySweptT
...
+
+        List<Cell> cells = rows.stream().map(row -> Cell.create(row, COL_A)).collect(Collectors.toList());
+        putCellsInTable(cells, TABLE_SWEPT_THOROUGH);
+
+        verifyPrefetchValidations(rows, cells, 100, 10, 1000);
+        verifyPrefetchValidations(rows, cells, 100, 3, 299);
+        verifyPrefetchValidations(rows, cells, 100, 4, 300);
+        verifyPrefetchValidations(rows, cells, 100, 4, 301);
+    }
+
```
jeremyk-91, line 1832: I think we are still missing one set of test cases, though they are likely to be a little annoying so I would prefer for them to be in a separate PR. Specifically, the logic of how we deal with wide rows is not tested yet. So something like this is missing:

```
    @Test
    public void getSortedColumnsAvoidsVeryLargeReadsInWideRows() {
        List<byte[]> rows = Stream.of("tom", "jerry", "alice", "bob", "cari", "david")
                .map(PtBytes::toBytes)
        List<byte[]> columns = LongStream.range(0, 10_000).mapToObj(PtBytes::toBytes).collect(Collectors.toList());

        // Only provide entries for the tom row
        List<Cell> cells = columns.stream().map(column -> Cell.create(
                rows.get(0), column)).collect(Collectors.toList());
        putCellsInTable(cells, TABLE_SWEPT_THOROUGH);

        verifyPrefetchValidations(
                rows, cells, 10_000, 1, 10_000);
        // verify that any gRCR on the KVS has a batch hint of *strictly less* than 10k
    }
```

Also: is the above result what you expect, and is it correct?

jeremyk-91 (commented):
🕐  1:25 🏁  🔥 

This looks broadly correct, though I believe there is a class of test cases that we haven't added around the handling of wide rows and a highly fragmented batch. In a sense, I think the current tests focus on the happy cases of the Math.max(Math.min(...)) per-row batching expression we have. I'd prefer these tests to be **separate** as 

1. this will only further delay this specific PR,
2. it's nontrivial,
3. even if wrong the error there manifests as inefficiency as opposed to correctness.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1684,6 +1692,195 @@ public void orphanedSentinelsDoNotGetDeletedAfterDetectionInConservativelySweptT
...
+        assertThat(retrievedEntries).hasSameElementsAs(cells.subList(1, batchHint - 1));
+
+        // should throw while fetching the next batch
+        assertThatThrownBy(sortedColumns::next).isInstanceOf(TransactionLockTimeoutException.class);
+    }
+
+    @Test
+    public void getSortedColumnsObeysColumnRangeSelection() {
+        List<byte[]> rows = ImmutableList.of(ROW_FOO, ROW_BAR);
+        List<Cell> col1_cells = ImmutableList.of(Cell.create(ROW_FOO, COL_A), Cell.create(ROW_BAR, COL_A));
```
jeremyk-91, line 1797: nit: `colA_cells`

jeremyk-91 (approved):

PR #5325, opened 2021-03-18
make testRuntimePasswordChange more reliable

berler:
**Goals (and why)**: testRuntimePasswordChange flakes a lot in tests (requires retriggering jobs in circle). This is an attempt to make this test more reliable.

* use a random username for the test postgres user (in case there is some weird interaction with concurrent tests)
* change a test timeout from 10 seconds to 60 seconds when the password is expected to be wrong
* when updating the user password, now executes the update twice (in case for some reason postgres was not correctly applying the password change the first time)

Ran tests locally (note after dozens of tries I still have never reproduced the failures seen in circle when running locally).


jeremyk-91 (approved):
👍  Thanks!

PR #5327, opened 2021-03-19
[SQLite] Part1: Async/Lazy Migration

gmaretic:
**Goals (and why)**:
We have migrated virtually everywhere, but if validation is slow for any reason, the server will block until we verified all namespaces. This alleviates the issue.

Submit a task to migrate(initialise) all namespaces asynchronously (previously was blocking). This is fine because if we need a component that has not been initialised yet, we will initialise it lazily.

Parameterized existing tests.

None, really. Though flagging that this may result in some initial sluggishness due to lazy initialisation -- mostly relevant if someone actually needs to migrate still. The alternative is that the server would not have started instead.

small

today

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponents.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponents.java
@@ -21,6 +21,7 @@ import com.google.common.base.Suppliers;
 import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.timelock.management.DiskNamespaceLoader;
 import com.palantir.atlasdb.timelock.management.PersistentNamespaceLoader;
+import com.palantir.common.concurrent.PTExecutors;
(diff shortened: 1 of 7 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/paxos/LocalPaxosComponents.java
```diff
@@ -100,6 +102,31 @@ public static LocalPaxosComponents createWithBlockingMigration(
...
+                metrics,
+                paxosUseCase,
+                legacyLogDirectory,
+                sqliteDataSource,
+                leaderUuid,
+                canCreateNewClients,
+                timeLockVersion,
+                skipConsistencyCheckAndTruncateOldPaxosLog,
+                sqliteAsyncExecutor);
+        sqliteAsyncExecutor.shutdown();
```
jeremyk-91, line 116: can we do this in a finally block? otherwise if `createWithAsyncMigration` throws this may not run

jeremyk-91 (approved):
👍

PR #5332, opened 2021-03-23
[getSortedColumns] | Serializable conflict checking

sudiksha27:
Ref - #5092
**Goals (and why)**:
Implements getSortedColumns endpoint for serializable conflict checking.

- If the table does not use serializable conflict checking, fall back to SnapshotTransaction#getSortedColumns
- Otherwise, if the request has exactly on distinct row, use `wrapIteratorWithBoundsChecking` as done in serializable getRowsColumnRangeIterator
- if request has more than one distinct rows, we keep track of where we are in the iteration.
- Having read till cell ->  (ri, cj), we ensure all cells with columns lexicographically smaller than cj and rows whose first appearance occurs before ri in request params are read. If that is not the case, a read write conflict is raised. 
- Writes that happened outside of the range read by us are ignored by the conflict checking logic

Added tests

Correctness

SerializationTransaction#getSortedColumns

First pass before EOD today would be great

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -704,7 +704,7 @@ public class SerializableTransaction extends SnapshotTransaction {
                 for (Map.Entry<BatchColumnRangeSelection, byte[]> e : rangeEnds.entrySet()) {
                     BatchColumnRangeSelection range = e.getKey();
                     byte[] rangeEnd = e.getValue();
-                    rangesToRows.put(nextLexicographicalRangeEnd(range, rangeEnd), row);
+                    rangesToRows.put(getBatchColumnRangeSelectionForEntriesReadSoFar(range, rangeEnd), row);
                 }
             }
 
@@ -738,8 +738,8 @@ public class SerializableTransaction extends SnapshotTransaction {
             if (endOfRange == null) {
                 return;
             }
-            BatchColumnRangeSelection range =
-                    nextLexicographicalRangeEnd(request.getColumnRangeSelection(), endOfRange.getColumnName());
+            BatchColumnRangeSelection range = getBatchColumnRangeSelectionForEntriesReadSoFar(
+                    request.getColumnRangeSelection(), endOfRange.getColumnName());
             Iterable<byte[]> rows = request.getRows();
             Comparator<Cell> comparator = columnOrderThenPreserveInputRowOrder(request.getRows());
             Iterator<Map.Entry<Cell, ByteBuffer>> readValues =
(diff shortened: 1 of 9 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -657,6 +731,65 @@ private void verifyColumnRanges(Transaction readOnlyTransaction) {
...
 
+    private void verifyGetSortedColumns(Transaction readOnlyTransaction) {
+        sortedColumnRangeEnds.forEach((request, endOfRangeReference) -> {
+            Cell endOfRange = endOfRangeReference.get();
+            // no checks required if no data has been read so far
+            if (endOfRange == null) {
+                return;
+            }
+            BatchColumnRangeSelection range =
+                    nextLexicographicalRangeEnd(request.getColumnRangeSelection(), endOfRange.getColumnName());
```
jeremyk-91, line 742: A bit concerned about this if the end of range is genuinely the last possible column name, and there exists a value there. To be clear the conceptual operation is correct, just that `nextLexicographicalRangeEnd` itself might be buggy

jeremyk-91 (commented):
🕐  2:15 🏁  3:50 🔥 

The tests are a bit tricky - I see you've been largely consistent with the design already there, but I think we should clean it up. Also, I think we should add tests for two cases:

1. someone who actually has the column `fffff....` (the last one) - checking that we can detect a conflict there
2. a conflict happening on the very last cell an iterator will visit.

If I'm not mistaken, this exposes an old bug in AtlasDB as well. 🔥

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1078,6 +1092,233 @@ public void testMarkTableInvolvedSkipsLockChecksOnCommitIfReadDoneDuringTransact
...
+        List<Cell> cells = Streams.stream(sortedColumns).map(Map.Entry::getKey).collect(Collectors.toList());
+        assertThat(cells).isEmpty();
+        assertThatCode(t1::commit).doesNotThrowAnyException();
+    }
+
+    @Test
+    public void testGetSortedColumnsNoConflict() {
+        List<byte[]> rows = generateRows(5);
+        List<Cell> cellsWrittenOriginally = generateCells(rows, generateColumns(5));
+        writeCells(cellsWrittenOriginally);
```
jeremyk-91, line 1112: ❤️  great, thanks for changing this!

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1078,6 +1084,167 @@ public void testMarkTableInvolvedSkipsLockChecksOnCommitIfReadDoneDuringTransact
...
+
+    @Test
+    public void getSortedColumnsUnloadedBatchIsConsideredUnread() {
+        List<byte[]> rows = writeColumnsForRows(2);
+
+        Transaction t1 = startTransaction();
+        Iterator<Map.Entry<Cell, byte[]>> sortedColumns = t1.getSortedColumns(
+                TEST_TABLE,
+                rows,
+                BatchColumnRangeSelection.create(PtBytes.EMPTY_BYTE_ARRAY, PtBytes.EMPTY_BYTE_ARRAY, 100));
```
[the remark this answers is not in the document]
jeremyk-91: yeah, I think we still need this.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1078,6 +1092,233 @@ public void testMarkTableInvolvedSkipsLockChecksOnCommitIfReadDoneDuringTransact
...
+        List<Cell> cells = Streams.stream(sortedColumns).map(Map.Entry::getKey).collect(Collectors.toList());
+        sanityCheckOnSortedCells(rows, cells, cellsWrittenOriginally);
+
+        // we write to a cell that has been read to introduce conflict
+        Transaction t2 = startTransaction();
+        put(t2, onlyRow, lastColumnName, "v0_0");
+        t2.commit();
+
+        assertThatThrownBy(t1::commit).isInstanceOf(TransactionSerializableConflictException.class);
+    }
```
jeremyk-91, line 1222: Ideally we'd have multiple rows: the reason I asked for this test was mainly to ensure that the case where we stop in the middle of `(r1, c), (r2, c) ...` where `c` _is_ the last column is handled correctly, while this goes through the "finished iterating" case if I'm reading it correctly. Otherwise looks good.

jeremyk-91 (commented):
discussed offline: this review is a 2nd pass on tests only

jeremyk-91 (approved):
🚀

PR #5333, opened 2021-03-23
[PDS-146088] Aggressive profiling when a Cassandra client close times out

jeremyk-91:
**Goals (and why)**:
PDS-146088.

- In the event of a timeout, up to once per hour log the thread state (basically a programmatic jstack).

**Testing (What was existing testing like?  What have you done to improve it?)**: Added unit tests covering all added classes.

- Might the volume be too large, even if this is only once every hour?
- Is the rate limiter actually applied?

**Where should we start reviewing?**: RateLimitingFailureHandler


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/RateLimitingFailureHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.common.base.Suppliers;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * Wraps a potentially expensive failure-handling computation in a memoizing supplier, so that it is only executed
+ * once per fixed duration.
+ */
+public class RateLimitingFailureHandler implements Runnable {
```
Jolyon-S, line 28:
```suggestion
public final class RateLimitingFailureHandler implements Runnable {
```

Jolyon-S: Also, on the one hand this is "rate limiting", but I feel like the pattern is more some form of expiring or caching failure handler (yes, it limits the rate to once per hour, but still).

Jolyon-S: And also also, this doesn't have to be a failure thing - this is more `RateLimitingRunnable` or `CachedRunnable` or `TimeExpiringMemoizedRunnable` or something.

Jolyon-S (approved):
1. The handler appears to be correct, from my read;
2. Assuming we have say 10k threads (which would be a _very_ high estimate!), this equates to 10k log lines an hour, right? That's roughly 3 per second in the absolute crazy case, and I think that's fairly reasonable, especially given that this is only triggered after we hit the "issue".

PR #5334, opened 2021-03-23
[PDS-169034] Background Sweep configuration

jeremyk-91:
**Goals (and why)**:
- Don't waste resources scheduling threads for background sweep when it's disabled and targeted sweep does the work anyway

- Add a config flag for whether to run the process at all. Defaults to true. I need to speak with large internal product team before making it false.

- Not much, existing tests should validate I haven't blown anything completely apart
- Will be tested when I turn this on in atlasdb proxy internally.

- Is this overly conservative?


sudiksha27 (approved):
Makes sense!

PR #5337, opened 2021-03-24
Use gradle-external-publish-plugin to avoid bintray issues

CRogers:
Bintray is being decommissioned and starting on Sunday will not be accepting any more release versions: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

This PR changes the publishing code to use the new `gradle-external-publish-plugin` that publishes to maven central.


jeremyk-91:
Change looks fine, from what I can tell. We also need to remove mandating `deploy` to be needed, so I'll do that.

jeremyk-91 (approved):
👍  thanks a lot for this!

PR #5339, opened 2021-03-25
[LW/AC] Keep events around for active transactions

Jolyon-S:
**Goals (and why)**:
Caching will be significantly more complex and less useful if we do not guarantee that we will keep events around for active transactions.

Note that this is a partial reversal of this PR: https://github.com/palantir/atlasdb/pull/5012/

* `retentionEvents` now takes the earliest timestamp to keep, and only retentions events if they are before that (as well as if the size exceeds 1000).
* Changed to a `TreeMap` in `VersionedEventStore` to allow us to get the lowest timestamp.

* Unit tests updated
* LWECIT also updated (and is the reason behind most of the diff).

It's LW, so correctness and performance, of course! Namely:
1. Does this violate any of our correctness guarantees? From what I can understand, we're just being more conservative on the retention side of things, so we're only going to keep more events, and we actually don't make any assumptions on what events exist anyway.
2. Does this cause a regression in terms of speed or memory? TreeMap over HashMap is now a log cost, but I think we use tree maps elsewhere and this is likely negligible. As for memory, we can now technically keep an arbitrary number of events around for long running transactions (at most one hour in AtlasDB). I suppose for _super_ write heavy users of LW, this could be an issue (since before we had certain guarantees that transactions were no more than 5 minutes).

* `VersionedEventStore`

Sooner rather than later, because we need this for other LW/AC stuff.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
@@ -25,13 +25,13 @@ import com.palantir.lock.watch.LockWatchVersion;
 import com.palantir.lock.watch.TransactionUpdate;
 import com.palantir.logsafe.Preconditions;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.NavigableMap;
 import java.util.Optional;
+import java.util.TreeMap;
 import org.immutables.value.Value;
 
 final class TimestampStateStore {
-    private final Map<Long, MapEntry> timestampMap = new HashMap<>();
+    private final NavigableMap<Long, MapEntry> timestampMap = new TreeMap<>();
 
     void putStartTimestamps(Collection<Long> startTimestamps, LockWatchVersion version) {
         startTimestamps.forEach(startTimestamp -> {
@@ -77,6 +77,14 @@ final class TimestampStateStore {
         return ImmutableTimestampStateStoreState.builder()
                 .timestampMap(timestampMap)
                 .build();
+    }
+
+    public Optional<Long> getEarliestLiveTimestamp() {
+        if (timestampMap.isEmpty()) {
+            return Optional.empty();
+        } else {
+            return Optional.of(timestampMap.firstKey());
+        }
     }
 
     @Value.Immutable
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
@@ -19,6 +19,9 @@ package com.palantir.atlasdb.keyvalue.api.watch;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.SortedSetMultimap;
+import com.google.common.collect.TreeMultimap;
 import com.palantir.atlasdb.transaction.api.TransactionLockWatchFailedException;
 import com.palantir.lock.v2.LockToken;
 import com.palantir.lock.watch.LockWatchVersion;
@@ -32,11 +35,13 @@ import org.immutables.value.Value;
 
 final class TimestampStateStore {
     private final NavigableMap<Long, MapEntry> timestampMap = new TreeMap<>();
+    private final SortedSetMultimap<Long, Long> livingVersions = TreeMultimap.create();
 
     void putStartTimestamps(Collection<Long> startTimestamps, LockWatchVersion version) {
         startTimestamps.forEach(startTimestamp -> {
             MapEntry previous = timestampMap.putIfAbsent(startTimestamp, MapEntry.of(version));
             Preconditions.checkArgument(previous == null, "Start timestamp already present in map");
+            livingVersions.put(version.version(), startTimestamp);
         });
     }
 
@@ -57,11 +62,13 @@ final class TimestampStateStore {
     }
 
     void remove(long startTimestamp) {
-        Optional.ofNullable(timestampMap.remove(startTimestamp));
+        Optional.ofNullable(timestampMap.remove(startTimestamp))
+                .ifPresent(entry -> livingVersions.remove(entry.version().version(), startTimestamp));
     }
 
     void clear() {
         timestampMap.clear();
+        livingVersions.clear();
     }
 
     Optional<LockWatchVersion> getStartVersion(long startTimestamp) {
@@ -76,15 +83,12 @@ final class TimestampStateStore {
     TimestampStateStoreState getStateForTesting() {
         return ImmutableTimestampStateStoreState.builder()
                 .timestampMap(timestampMap)
+                .livingVersions(livingVersions)
                 .build();
     }
 
-    public Optional<Long> getEarliestLiveTimestamp() {
-        if (timestampMap.isEmpty()) {
-            return Optional.empty();
-        } else {
-            return Optional.of(timestampMap.firstKey());
-        }
+    public Optional<Long> getEarliestLiveSequence() {
+        return Optional.ofNullable(Iterables.getFirst(livingVersions.keySet(), null));
     }
 
     @Value.Immutable
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.palantir.lock.watch.LockWatchEvent;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -31,12 +32,19 @@ import java.util.stream.Collectors;
 final class VersionedEventStore {
     private static final boolean INCLUSIVE = true;
 
+    private final int minEvents;
     private final int maxEvents;
     private final NavigableMap<Long, LockWatchEvent> eventMap = new TreeMap<>();
 
-    VersionedEventStore(int maxEvents) {
-        Preconditions.checkArgument(maxEvents > 0, "maxEvents must be positive");
+    VersionedEventStore(int minEvents, int maxEvents) {
+        Preconditions.checkArgument(minEvents > 0, "minEvents must be positive", SafeArg.of("minEvents", minEvents));
+        Preconditions.checkArgument(
+                maxEvents >= minEvents,
+                "maxEvents must be greater than or equal to minEvents",
+                SafeArg.of("minEvents", minEvents),
+                SafeArg.of("maxEvents", maxEvents));
         this.maxEvents = maxEvents;
+        this.minEvents = minEvents;
     }
 
     Collection<LockWatchEvent> getEventsBetweenVersionsInclusive(Optional<Long> maybeStartVersion, long endVersion) {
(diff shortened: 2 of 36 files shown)


atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/cacheClearedOnSnapshotUpdate/event-cache-1.json
```diff
@@ -49,20 +49,24 @@
   },
   "timestampStoreState" : {
     "timestampMap" : {
-      "16" : {
+      "1" : {
```
Jolyon-S, line 52: This is now sorted, thus the reorder.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -48,11 +50,25 @@
...
+        }
+
+        int numToRetention = eventMap.size() - maxEvents;
+        long earliestVersion = earliestSequenceToKeep.orElse(Long.MAX_VALUE);
+
         ImmutableLockWatchEvents.Builder builder = LockWatchEvents.builder();
-        Iterators.consumingIterator(Iterators.limit(eventMap.entrySet().iterator(), numToRetention))
-                .forEachRemaining(entry -> builder.addEvents(entry.getValue()));
+        List<Map.Entry<Long, LockWatchEvent>> eventsToClear = eventMap.entrySet().stream()
+                .limit(numToRetention)
```
gmaretic, line 63: I know it technically does not matter since the treemap will return entries in ascending key order, but still let's swap the filter and limit so that it would also work with a generic map.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -48,11 +50,25 @@
...
+                .filter(entry -> entry.getKey() < earliestVersion)
+                .collect(Collectors.toList());
+
+        eventsToClear.forEach(entry -> {
+            eventMap.remove(entry.getKey(), entry.getValue());
```
gmaretic, line 68: Is it even possible for this to fail because the value does not match? I would certainly hope not, in which case let's just go with `eventMap.remove(entry.getKey())`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -48,11 +50,25 @@
                 .orElseGet(ImmutableList::of);
     }
 
-    LockWatchEvents retentionEvents() {
-        int numToRetention = Math.max(0, eventMap.size() - maxEvents);
+    LockWatchEvents retentionEvents(Optional<Long> earliestSequenceToKeep) {
```
gmaretic, line 53: This overall makes sense, but as you pointed out, it can cause the size of this to grow almost unboundedly. We should probably do two things:
a) come up with some limit on number of events where we retention the old events anyway
b) have a way for transactions to declare they are / are not using LW, so if we have slow transactions that do not actually use them, we can still retention away

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
```diff
@@ -57,11 +62,13 @@ void putCommitUpdates(Collection<TransactionUpdate> transactionUpdates, LockWatc
     }
 
     void remove(long startTimestamp) {
-        Optional.ofNullable(timestampMap.remove(startTimestamp));
+        Optional.ofNullable(timestampMap.remove(startTimestamp))
+                .ifPresent(entry -> livingVersions.remove(entry.version().version(), startTimestamp));
     }
 
     void clear() {
```
gmaretic, line 69: Should this be synchronised with writes? Can it happen that we have a clear, but in the middle there is a new start transactions so we update `livingVersions` before we clear it? What is the damage if that happens, can we live with that?

atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/getCommitUpdateIsInvalidateSomeAsEventsAreRetained/event-cache-1.json
```diff
@@ -0,0 +1,82 @@
...
+          },
+          "commitVersion" : {
+            "id" : "470c855e-f77b-44df-b56a-14d3df085dbc",
+            "version" : 6
+          }
+        }
+      }
+    },
+    "livingVersions" : {
+      "3" : [ 1 ]
```
gmaretic, line 79: This is not intuitive since we processed the commit timestamps update, but I assume that is because it is living until snapshot transaction manager tells the LWEC to remove it?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -273,15 +273,18 @@ public void largerUpdateAfterSmallUpdateOnlyPicksUpNewEvents() {
...
         setupInitialState();
         eventCache.processGetCommitTimestampsUpdate(COMMIT_UPDATE, SUCCESS);
         eventCache.processStartTransactionsUpdate(ImmutableSet.of(), SUCCESS_2);
         verifyStage();
 
         CommitUpdate commitUpdate = eventCache.getCommitUpdate(START_TS_1);
-        assertThat(commitUpdate.accept(new InvalidatedAllVisitor())).isTrue();
+        assertThat(commitUpdate.accept(new CommitUpdateVisitor())).containsExactlyInAnyOrder(DESCRIPTOR, DESCRIPTOR_3);
+
+        eventCache.removeTransactionStateFromCache(START_TS_1);
```
gmaretic, line 286: Ah yes, and this simulates the tm telling us we are done

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -273,15 +273,18 @@ public void largerUpdateAfterSmallUpdateOnlyPicksUpNewEvents() {
...
+    public void getCommitUpdateIsInvalidateSomeAsEventsAreRetained() {
         createEventCache(2);
         setupInitialState();
         eventCache.processGetCommitTimestampsUpdate(COMMIT_UPDATE, SUCCESS);
         eventCache.processStartTransactionsUpdate(ImmutableSet.of(), SUCCESS_2);
         verifyStage();
 
         CommitUpdate commitUpdate = eventCache.getCommitUpdate(START_TS_1);
-        assertThat(commitUpdate.accept(new InvalidatedAllVisitor())).isTrue();
+        assertThat(commitUpdate.accept(new CommitUpdateVisitor())).containsExactlyInAnyOrder(DESCRIPTOR, DESCRIPTOR_3);
```
gmaretic, line 284: What is the semantics of this commit update? I can see these are the descriptors that were locked while we were in the commit stage, so we will use this information to invalidate existing cached entries for those?

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
```diff
@@ -467,18 +470,20 @@ public void newEventsStartingWithTheSameVersionAsCurrentDoesNotThrow() {
...
 
-        assertThatThrownBy(() -> eventCache.getUpdateForTransactions(TIMESTAMPS, Optional.empty()))
-                .isExactlyInstanceOf(TransactionLockWatchFailedException.class)
-                .hasMessage("Events do not enclose the required versions");
+        TransactionsLockWatchUpdate update = eventCache.getUpdateForTransactions(TIMESTAMPS_2, Optional.empty());
 
-        assertThatThrownBy(() -> eventCache.getCommitUpdate(START_TS_1))
-                .isExactlyInstanceOf(TransactionLockWatchFailedException.class)
-                .hasMessage("start or commit info not processed for start timestamp");
+        assertThat(update.clearCache()).isTrue();
```
gmaretic, line 480: Let's add some `.as()` to these assertions, so poor reviewers don't have their head explode figuring out what is going on

gmaretic (commented):
Broadly makes sense, but I do have some concerns

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
Jolyon-S: I disagree; if no version is provided, then swapping the filter and the limit does nothing, and it still depends on the behaviour of the sorted map (which it did before in this class). It would actually just be wrong to have a generic map here (the type is still `NavigableMap` which forces this).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
Jolyon-S: Everything in these classes is not thread safe by itself, and is run via the resilient proxy in synchronised methods. Happy to make that clearer if you want.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheIntegrationTest.java
Jolyon-S: The commit update essentially tells you what events we have seen between the version at start and commit time. This allows you to work out which locks have been taken out during your transaction and thus, if you are a serialisable transaction and you have read a cached value, you can determine whether you need to conflict or not (also, you wrote the commit update originally :P).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
gmaretic: I mean I see what you're trying to say, even though you are not saying it explicitly, which is that we want to guarantee that we always retention from the oldest, so if our `Set` here was not sorted, we'd be in trouble. Also, I guess this ordering of limit and filter does limit the amount of filtering we do, although we could further improve this by having a loop instead of a stream. However, this requires us to jump through mental hoops because it's too smart: we have an `entrySet()` of some map, then we take `numToRetention` elements and do the logic. Wait, it's a set -- better go make sure that `eventMap.entrySet()` returns a set that is sorted in the correct way for this logic to work. Oh, it's a `NavigableMap`, well that still doesn't guarantee it. Oh, it's actually initialized to a `TreeMap()`, ok let's go see the javadoc. Ok, I guess it works.

In fact, the performant way to do this would be to just have  a for loop and break when either condition is met, but that would look ugly, so I can live with adding a comment above this stream reminding us that this entryset is sorted ascending by key

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -63,26 +64,36 @@ LockWatchEvents retentionEvents(Optional<Long> earliestSequenceToKeep) {
             return LockWatchEvents.builder().build();
         }
 
+        // Guarantees that we remove some events while still also potentially performing further retention
         if (eventMap.size() > maxEvents) {
-            return retentionEvents(eventMap.size() - maxEvents, Long.MAX_VALUE);
+            List<LockWatchEvent> overMaxSizeEvents = retentionEvents(eventMap.size() - maxEvents, Long.MAX_VALUE);
+            List<LockWatchEvent> restOfEvents =
```
gmaretic, line 70: This is kinda difficult to parse because it's not immediately obvious that the method call will modify `eventMap`, but can't think of a way to make it better...

gmaretic (approved):
The remaining stuff is mostly nits

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
Jolyon-S: I actually think the for loop would be a bit clearer in what the semantics are, so I think I'll go with that, and also add javadocs.

Jolyon-S: I'll comment here (although `NavigableMap` does say on its javadoc via `SortedMap` that it _does_ guarantee the keys are sorted in `entrySet`).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
Jolyon-S: Commented on this.

Jolyon-S:
merged because trial-publish wasn't working.

PR #5343, opened 2021-03-26
getSortedColumns bug fix

sudiksha27:
**Goals (and why)**:
The ordering of entries was being lost in the process of running validation checks on bat of merge-sorted entries. This PR fixes the bug. 

TBD - we did validate that after this change wrong ordering is not produced on internal test stack

Cannot write a test

🔥


jeremyk-91 (approved):
oof

PR #5347, opened 2021-03-26
Handle socket timeout exception

anjor:
Retries on socket timeout exception.

### DIFF
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
@@ -32,8 +32,6 @@ public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocat
     private static final int MAX_NUM_RETRIES = 5;
     private final Supplier<T> delegate;
 
-    private int numRetries = 0;
-
     private RetryOnSocketTimeoutExceptionProxy(Supplier<T> delegate) {
         this.delegate = delegate;
     }
@@ -47,7 +45,7 @@ public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocat
     @Override
     protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
         ResultOrThrowable attempt = singleInvocation(method, args);
-        numRetries = 0;
+        int numRetries = 0;
         while ((numRetries < MAX_NUM_RETRIES) && !attempt.isSuccessful()) {
             numRetries++;
             Throwable cause = attempt.throwable().get();
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
@@ -32,7 +32,7 @@ public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocat
     private static final int MAX_NUM_RETRIES = 5;
     private final Supplier<T> delegate;
 
-    private static int numRetries = 0;
+    private int numRetries = 0;
 
     private RetryOnSocketTimeoutExceptionProxy(Supplier<T> delegate) {
         this.delegate = delegate;
@@ -47,6 +47,7 @@ public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocat
     @Override
     protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
         ResultOrThrowable attempt = singleInvocation(method, args);
+        numRetries = 0;
         while ((numRetries < MAX_NUM_RETRIES) && !attempt.isSuccessful()) {
             numRetries++;
             Throwable cause = attempt.throwable().get();
@@ -56,7 +57,6 @@ public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocat
             }
             attempt = singleInvocation(method, args);
         }
-        numRetries = 0;
         if (attempt.isSuccessful()) {
             return attempt.result().orElse(null);
         }
--- a/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
+++ b/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
@@ -24,9 +24,7 @@ import com.palantir.conjure.java.api.errors.UnknownRemoteException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
-import java.time.Clock;
-import java.time.Duration;
-import java.time.Instant;
+import java.net.SocketTimeoutException;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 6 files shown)


atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,96 @@
...
+        InvocationTargetException exception = (InvocationTargetException) throwable;
+        Throwable cause = exception.getCause();
+        if (!isCausedBySocketTimeoutException(cause)) {
+            return ResultOrThrowable.failure(cause);
+        }
+        return ResultOrThrowable.success(null);
+    }
+
+    @VisibleForTesting
+    static boolean isCausedBySocketTimeoutException(Throwable throwable) {
```
jeremyk-91, line 86: we should add simple tests

jeremyk-91 (commented):
Looks good! I think we should add some tests, but otherwise we should be close

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,96 @@
...
+            return ResultOrThrowable.failure(th);
+        }
+    }
+
+    private ResultOrThrowable isRetriable(Throwable throwable) {
+        if (!(throwable instanceof InvocationTargetException)) {
+            return ResultOrThrowable.failure(throwable);
+        }
+        InvocationTargetException exception = (InvocationTargetException) throwable;
+        Throwable cause = exception.getCause();
```
Jolyon-S, line 78: Is this necessary? Can you not just get a cause from a generic throwable?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,96 @@
...
+        if (!isCausedBySocketTimeoutException(cause)) {
+            return ResultOrThrowable.failure(cause);
+        }
+        return ResultOrThrowable.success(null);
```
Jolyon-S, line 82: Flip the if, and I'd add the other part in the else block.

```suggestion
        if (isCausedBySocketTimeoutException(cause)) {
            return ResultOrThrowable.success(null);
        } else {
            return ResultOrThrowable.failure(cause);
        }
```

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,96 @@
...
+        return (U) Proxy.newProxyInstance(interfaceClass.getClassLoader(), new Class<?>[] {interfaceClass}, proxy);
+    }
+
+    @Override
+    protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
+        ResultOrThrowable attempt = singleInvocation(method, args);
+        while ((numRetries < MAX_NUM_RETRIES) && !attempt.isSuccessful()) {
+            numRetries++;
+            Throwable cause = attempt.throwable().get();
+            ResultOrThrowable retriableExceptionCheck = isRetriable(cause);
```
Jolyon-S, line 53: I'm really not a fan of how you are using this here - it's never actually a result, but rather it is either retriable, or it is not and there is an exception. You could just return `Optional<Throwable>` instead for the same effect.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,96 @@
...
+
+        return (U) Proxy.newProxyInstance(interfaceClass.getClassLoader(), new Class<?>[] {interfaceClass}, proxy);
+    }
+
+    @Override
+    protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
+        ResultOrThrowable attempt = singleInvocation(method, args);
+        while ((numRetries < MAX_NUM_RETRIES) && !attempt.isSuccessful()) {
+            numRetries++;
+            Throwable cause = attempt.throwable().get();
```
Jolyon-S, line 52: There is no form of backoff between retries; do we want that? I'm a bit worried that we're just going to still fail 6 times in rapid succession.

atlasdb-conjure/src/test/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxyTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.palantir.atlasdb.http.v2;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import org.junit.Test;
+
+public class RetryOnSocketTimeoutExceptionProxyTest {
```
Jolyon-S, line 25: Can you add some tests that mock the supplier to fail say 3 times, then succeed, and verify that it passes? For example:

```java
Supplier<Long> supplier = mock(Supplier.class);
when(supplier.get())
    .thenThrow(new RuntimeException(new SocketTimeoutException()))
    .thenThrow(...)
    .thenThrow(...)
    .thenReturn(11L);

```

Jolyon-S: (And some other tests of this form for no failures, only failures, failures where exceptions don't have the right root cause, etc.)

Jolyon-S (commented):
Wiring looks right, I have a couple of concerns and a few test suggestions.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
Jolyon-S: ugh, FFP does the same.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
Jolyon-S: actually no, FFP.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
Jolyon-S: FFP code...

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.function.Supplier;
+
+/**
+ * This proxy exists to retry in case of a SocketTimeoutException (something that isn't retried natively by Dialogue).
+ */
+public final class RetryOnSocketTimeoutExceptionProxy<T> extends AbstractInvocationHandler {
+    private static final int MAX_NUM_RETRIES = 5;
+    private final Supplier<T> delegate;
+
+    private static int numRetries = 0;
```
Jolyon-S, line 35: ooooh this looks weird

does FFP use a `static` here?!

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
```diff
@@ -0,0 +1,98 @@
...
+        while ((numRetries < MAX_NUM_RETRIES) && !attempt.isSuccessful()) {
+            numRetries++;
+            Throwable cause = attempt.throwable().get();
+            ResultOrThrowable retriableExceptionCheck = isRetriable(cause);
+            if (!retriableExceptionCheck.isSuccessful()) {
+                throw retriableExceptionCheck.throwable().get();
+            }
+            attempt = singleInvocation(method, args);
+        }
+        numRetries = 0;
```
Jolyon-S, line 59: Probably makes more sense to reset this to 0 at the beginning of the loop?

Jolyon-S: Also, this won't be nice with statics - more than one proxy -> they will _all_ have wonky retry numbers (write a test where you have two proxies to convince yourself that this won't work with a static).

Jolyon-S (commented):
Static looks gnarly.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/RetryOnSocketTimeoutExceptionProxy.java
anjor: oh you are right. This shouldn't be static.

PR #5351, opened 2021-03-29
[DB TimeLock, 2021] Part 1: Another Look At TimeLock Migration ETE Tests

jeremyk-91:
**Goals (and why)**:
- For various internal reasons, we expect to have another fairly large swathe of migrations from embedded timelocks to external timelock services. If we're going to do this, I would prefer for us to have working end-to-end tests.

This PR can be viewed as consisting of a few components:
1. Refactoring of TimeLockMigrationEteTest to support more general test contexts.
2. The writing of a test context for DbKVS -> Db-TimeLock.
3. Changes to our Circle configuration to run this in a separate container because the tests are heavyweight.

This PR focuses on tests.

- Spinning up another container for parallelism costs $$$, but I think it's worthwhile.
- I haven't written a fully parameterized test to run it for both because that would make builds much longer unless we did even more janky circle optimisation.

Start with TimeLockMigrationEteTest.


scripts/circle-ci/run-circle-tests.sh
```diff
@@ -55,7 +57,7 @@ JAVA_GC_LOGGING_OPTIONS="${JAVA_GC_LOGGING_OPTIONS} -XX:-TraceClassUnloading"
 JAVA_GC_LOGGING_OPTIONS="${JAVA_GC_LOGGING_OPTIONS} -Xloggc:build-%t-%p.gc.log"
 
 # External builds have a 16gb limit.
-if [ "$CIRCLE_NODE_INDEX" -eq "7" ]; then
+if [ "$CIRCLE_NODE_INDEX" -eq "9" ]; then
```
gmaretic, line 60: Why did we change this from 7 to 9?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/startup/TimeLockMigrator.java
```diff
@@ -71,9 +76,14 @@ protected synchronized void tryInitialize() {
         try {
             destination.ping();
         } catch (Exception e) {
-            throw new AtlasDbDependencyException("Could not contact the Timelock Server.", e);
+            throw new AtlasDbDependencyException("Could not contact the TimeLock Server.", e);
         }
         long currentTimestamp = source.backupAndInvalidate();
+        log.info(
+                "Now fast forwarding the timestamp on TimeLock Server, possibly as part of a TimeLock migration."
+                        + " If you are already using TimeLock, DON'T PANIC if this number is less than what you"
```
gmaretic, line 84: nit : DO NOT

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TimeLockMigrationEteTest.java
```diff
@@ -42,12 +51,55 @@
...
+                        .dockerComposeYmlFile(new File("docker-compose.timelock-migration.cassandra.yml"))
+                        .databaseServiceName("cassandra")
+                        .build();
+                TEST_CONTEXT = ImmutableTimeLockMigrationTestContext.builder()
+                        .eteConfigWithEmbedded(embeddedConfig)
+                        .eteConfigWithTimeLock(new File("docker/conf/atlasdb-ete.timelock.cassandra.yml"))
+                        .dockerClientConfiguration(dockerClientConfiguration)
+                        .build();
+                break;
+            case POSTGRES_DB_TIMELOCK:
```
gmaretic, line 74: So this is basically a placeholder for "we will parameterize on this once we want to run the test on posgres as well"? This whole static block is a bit of a mess, you could have just put it all as separate methods of the enum though not sure if you agree that's cleaner :P

gmaretic (approved):

scripts/circle-ci/run-circle-tests.sh
jeremyk-91: discussed offline: this is the errorprone container, and when we added container 8 we forgot to update this.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/TimeLockMigrationEteTest.java
jeremyk-91: There's an issue with shutting down the cluster on Postgres being unstable. This only applies to the shutdown part (not the assertions), and I don't have the bandwidth to fix that right now.

PR #5355, opened 2021-03-30
[PDS-172208] | Reference tracker for multi-client batchers

sudiksha27:
**Goals (and why)**:
The internal AtlasDb proxy closes TM after fixed amount of time which in turn closes the multi-client autobatcher. Since the autobatcher is being shared among multiple TMs, they should not be closed unless all TMs are down.

This PR introduces a ReferenceTrackingWrapper to resolve this bug;

- Implements ReferenceTrackingWrapper - the wrapper simply maintains a counter and triggers close on delegate if the counter is down to zero
- Wiring

Unit tests added

- Not defensive against more closes than opens - as far as I see, think this should never happen.
- We only maintain a counter instead of record of some sort of ID of clients using the delegate. Is this a problem?
- This was a problem before as well - we do not handle the case where autobatcher is closed. Internal proxy could still be holding a reference to the instance and give out to a new client.

ReferenceTrackingWrapper

🔥 yesterday

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -1268,9 +1268,10 @@ public abstract class TransactionManagers {
             return new LegacyLeaderTimeGetter(namespacedConjureTimelockService);
         }
 
-        ReferenceTrackingWrapper<LeaderTimeCoalescingBatcher> batcher =
+        ReferenceTrackingWrapper<LeaderTimeCoalescingBatcher> referenceTrackingBatcher =
                 timelockRequestBatcherProviders.get().leaderTime().getBatcher(multiClientTimelockServiceSupplier);
-        return new NamespacedCoalescingLeaderTimeGetter(timelockNamespace, batcher);
+        referenceTrackingBatcher.recordReference();
+        return new NamespacedCoalescingLeaderTimeGetter(timelockNamespace, referenceTrackingBatcher);
     }
 
     private static Supplier<InternalMultiClientConjureTimelockService> getMultiClientTimelockServiceSupplier(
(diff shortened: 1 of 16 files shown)


lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -137,7 +137,11 @@ public long currentTimeMillis() {
 
     @Override
     public void close() {
-        transactionStarter.close();
-        commitTimestampGetter.close();
+        try {
+            transactionStarter.close();
+            commitTimestampGetter.close();
+        } catch (Exception e) {
```
gmaretic, line 143: Looks like we don't actually close the `leaderTimeGetter`? Also log something if we actually throw here?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -1267,7 +1268,7 @@ private static LeaderTimeGetter getLeaderTimeGetter(
             return new LegacyLeaderTimeGetter(namespacedConjureTimelockService);
         }
 
-        LeaderTimeCoalescingBatcher batcher =
+        ReferenceTrackingWrapper<LeaderTimeCoalescingBatcher> batcher =
```
gmaretic, line 1271: I think we should increase the reference count here, since we are decreasing it on close. Also the same for the other two batchers above

gmaretic (changes requested):
I have a few suggestions. It is a bit nasty, but I don't see an easy way to make it nicer, so let's go with it

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
sudiksha27: Yes, this looks like another bug :(

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
sudiksha27: I was thinking of adding it to internal proxy where the constructors live but this is fine too

gmaretic (approved):
LGTM!

PR #5357, opened 2021-03-31
Sanitizing the Git Version for Correct Snapshot Publishing

jeremyk-91:
**Goals (and why)**:
- In #5337, we switched our publishing story to use the gradle publishing plugin among other changes. Unfortunately, JGit has a six-year old bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=452189 that means that we often get erroneous reports of the build being dirty.
- We normally get round this with something like `System.env.CIRCLE_TAG ?: gitVersion()`
- But Atlas does do develop snapshot publishing which that doesn't fix.

- If `git status` reports the build is *clean* BUT the plugin says it is of the form `bleh.dirty`, just return `bleh` as the version.

- Did some manual tests. No automated tests, sadly.

- Might this ever be unsafe?

@CRogers for SA / is there any pattern that we might unexpectedly be violating in this way?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ version = System.env.CIRCLE_TAG ?: sanitizeVersion()
 description = 'Transactional distributed database layer'
 
 def sanitizeVersion() {
-    def lastTag = "git describe --tags --always --first-parent".execute().text
+    def lastTag = "git describe --tags --always --first-parent".execute().text.trim()
     def isClean = "git status --porcelain".execute().text.isEmpty()
 
     if (isClean) {
--- a/build.gradle
+++ b/build.gradle
@@ -51,6 +51,14 @@ task printLastVersion {
     }
 }
 
+task isClean {
+    doLast {
+        def diff = "git status --porcelain".execute().text
+        println diff
+        println diff.isEmpty()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -41,9 +41,19 @@ apply plugin: 'com.palantir.external-publish'
...
 
+def sanitizeVersion() {
+    def isClean = "git status --porcelain".execute().text.isEmpty()
+    def pluginReportedVersion = gitVersion()
+
+    if (isClean && pluginReportedVersion.endsWith(".dirty")) {
+        return pluginReportedVersion.dropRight(".dirty".length())
+    }
+    return pluginReportedVersion
+}
```
Jolyon-S, line 55: Is this still prone to semantic merge conflicts which cause develop to be dirty? From my read of it, it is prone to that.

Jolyon-S (approved):
Given that we discussed this offline, this looks reasonable (even if it is actually pretty dirty). However, if you want to wait for @CRogers to have a look before merging then I'd support that decision.

build.gradle
jeremyk-91: It should be fine. The `isClean` check should fail if we actually end up generating something

PR #5361, opened 2021-04-01
Remove usage of unsupported dropRight (only in Groovy 3.0)

jeremyk-91:
**Goals (and why)**:
Have a working build

Replace dropRight with substring

Manual test only

None really

+1-1

yesterday


Jolyon-S (approved):
Could factor out `.dirty` into a variable, but honestly this is fine.

PR #5363, opened 2021-04-06
[DB TimeLock 2021] Parts 3 and 4: Blob Store and Config Change Protection

jeremyk-91:
**Goals (and why)**:
Currently if a user changes to db timelock without following the actual migration steps, we start up and silently corrupt the user. Some direct users of AtlasDB have client-side protections, but not all may have them. Ideally we can find this out after one bad node has restarted and avert an outage altogether.

- Write a blob store in SQLite that allows us to easily store bits of data if timelock needs it
- Persist the timestamp persistence configuration a user provides in the database
- On startup, check if the timestamp persistence is locationally compatible. Currently this is just "if you're using Paxos, are we still on Paxos - if you're using db-timelock, are we still on db-timelock". We _may_ want to extend this to include specific database coordinates in the future, but not right now as it's nontrivial (e.g. if the database password or say some connection parameters change, you _don't_ want to fail).

New classes are tested

- Is this PR too large? I can split if needed
- I don't know if permitting repeated reseeds is correct. This is *inconsistent* with isNewService(), but banning these could lead to some hairy interactions if users change the underlying database of db-timelock at a similar time as when we make the checks stricter (for which there are no concrete plans yet, but this could happen). If we decide to ban these later on I'm fine with users who still wrongly have the flag set to face being unable to restart a node.

**Where should we start reviewing?**: SqliteBlobStore?


sudiksha27 (commented):
Actually looks pretty good. Curious about the motivation behind the change.

Approval pending on fixing the build.

sudiksha27 (approved):

PR #5365, opened 2021-04-06
add dbkvs runtime password config

berler:
**Goals (and why)**: Supports live password changes for products that use the standard atlas install and runtime config. Previously only products that had their own custom config and managed their own hikari pool could update the password at runtime. No behavior is changed when the new dbkvs runtime config is absent.

* adds a new dbkvs runtime config
* when the dbkvs runtime config is absent, behavior is the same as before
* subscribes to the runtime config so that the hikari pool is updated any time the config changes


jeremyk-91 (approved):
👍  Looks reasonable, thanks!

PR #5374, opened 2021-04-09
fix dbkvs runtime config

berler:
**Goals (and why)**: Accidentally did not set a jackson type name for the new dbkvs runtime config, which means that it relied on using the class name instead of the logical name ("relational").

**Implementation Description (bullets)**: Add missing jackson annotation.


gmaretic (approved):
Makes sense

jeremyk-91 (approved):

PR #5376, opened 2021-04-12
LW Caching I: Basic interfaces

Jolyon-S:
**Goals (and why)**:
* Define the main interfaces that the caching flow will use

Define interfaces based off my `silver-snow` branch on what the caching will look like. These may change, but I feel that outlining the main things will help reason with what we need.

Interfaces. No testing needed, yet.

Are these in the right project? These interfaces cannot live in `lock-api` like the rest of LW because they depend on things like `CellReference`.

`LockWatchValueCache`

Monday/Tuesday this week to unblock subsequent work.


gmaretic (approved):
Does this herald a new era where people who have not worked on the workstream can actually understand what is going on?! Stay tuned...

PR #5377, opened 2021-04-12
LW Caching II: Refactor current internal stores to use not raw longs

Jolyon-S:
**Goals (and why)**:
Too many longs in these datastructures makes it hard to reason about what is going on.

Replaced `<Long>` things with either `Sequence` or `StartTimestamp` to make it much clearer what the hell is going. Note that I didn't do this for API things - the scope is mostly where we could very easily confuse which was which.

Just fixed tests. Warning, a lot of json changes (but this is good because it'll become much much clearer what these longs are).

Hopefully this isn't uglier.

diff

Today to unblock things.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
@@ -23,6 +23,7 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.SafeArg;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
@@ -32,10 +33,12 @@ import java.util.stream.Collectors;
 
 final class VersionedEventStore {
     private static final boolean INCLUSIVE = true;
+    private static final Sequence MAX_VERSION = Sequence.of(Long.MAX_VALUE);
 
     private final int minEvents;
     private final int maxEvents;
-    private final NavigableMap<Long, LockWatchEvent> eventMap = new TreeMap<>();
+    private final NavigableMap<Sequence, LockWatchEvent> eventMap =
+            new TreeMap<>(Comparator.comparingLong(Sequence::value));
 
     VersionedEventStore(int minEvents, int maxEvents) {
         Preconditions.checkArgument(minEvents > 0, "minEvents must be positive", SafeArg.of("minEvents", minEvents));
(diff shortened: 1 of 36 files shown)


atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/cacheClearedOnSnapshotUpdate/event-cache-1.json
```diff
@@ -65,8 +65,8 @@
       }
     },
     "livingVersions" : {
-      "3" : [ 1 ],
-      "6" : [ 16 ]
+      "Sequence{value=3}" : [ { } ],
```
gmaretic, line 68: What happened here (and everywhere else)?

gmaretic (changes requested):
Overall, this is a great change. It does look like `livingVersions` json serialisation is broken as the timestamps are serialised as `{ }`. Also a small suggestion to make the new types comparable

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
```diff
@@ -32,10 +33,12 @@
 
 final class VersionedEventStore {
     private static final boolean INCLUSIVE = true;
+    private static final Sequence MAX_VERSION = Sequence.of(Long.MAX_VALUE);
```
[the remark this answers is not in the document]
Jolyon-S: No, because of the class initialisation deadlock (I think it'd complain?)

atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/cacheClearedOnSnapshotUpdate/event-cache-2.json
```diff
@@ -36,7 +36,11 @@
       }
     },
     "livingVersions" : {
-      "7" : [ 123, 1255 ]
+      "Sequence{value=7}" : [ {
+        "start-ts" : 123
+      }, {
+        "start-ts" : 1255
+      } ]
```
Jolyon-S, line 43: Not the most compact, and I don't know why sequence and start timestamp serialise differently, but whatever. It's now clear what's going on internally.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/VersionedEventStore.java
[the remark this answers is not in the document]
gmaretic: fair enough

atlasdb-impl-shared/src/test/resources/lockwatch-event-cache-output/cacheClearedOnSnapshotUpdate/event-cache-2.json
gmaretic: They do serialise the same, but both `TreeMap` and `TreeMultimap` do some weird stuff. Since this is only for tests anyways, it's fine

gmaretic (approved):

PR #5378, opened 2021-04-12
LW Cache III, B: Centralised Cache

Jolyon-S:
**Goals (and why)**:
* Create a central cache (per transaction manager).
* Handle necessary cases

* `LockWatchValueCache` contains the values in a vavr cache (which is great at giving slightly modified views of a map).
* Start transactions goes through this and updates events
* Commits publish the updated values to the central cache

* ValueStore and SnapshotStore both have tests
* LockWatchValueCache has a placeholder for testing later - the currently implemented methods make it hard to test as-is.
* ValueCacheSnapshot does _not_ have tests, but it's basically a map.

The usual - correctness and performance.

`LockWatchValueCacheImpl` is probably the best place to start.

This week - subsequent work is partially blocked on this (PR chain incoming!)


atlasdb-impl-shared/src/main/java/com/palantir/lock/watch/LockWatchReferencesVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.lock.watch.LockWatchReferences.EntireTable;
+import com.palantir.lock.watch.LockWatchReferences.ExactCell;
+import com.palantir.lock.watch.LockWatchReferences.ExactRow;
+import com.palantir.lock.watch.LockWatchReferences.RowPrefix;
+import com.palantir.lock.watch.LockWatchReferences.RowRange;
+import java.util.Optional;
+
+public final class LockWatchReferencesVisitor implements LockWatchReferences.Visitor<Optional<TableReference>> {
```
Jolyon-S, line 27: Because the visitor is package private otherwise....

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/StructureHolder.java
```diff
@@ -0,0 +1,53 @@
...
+package com.palantir.atlasdb.keyvalue.api.cache;
+
+import java.util.function.Function;
+
+/**
+ * Utility class to make manipulating {@link io.vavr.collection.Map} and {@link io.vavr.collection.Set} less
+ * error-prone as all methods that modify state return a *new* instance in vavr.
+ */
+final class StructureHolder<V> {
+    private final V initialValue;
```
jeremyk-91, line 26: nit/unactionable?: not wrong, though it does feel more conceptually intuitive to me to have a supplier of the initial value

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,163 @@
...
+    public synchronized void processStartTransactions(Set<Long> startTimestamps) {
+        TransactionsLockWatchUpdate updateForTransactions =
+                eventCache.getUpdateForTransactions(startTimestamps, currentVersion);
+        updateCurrentVersion(updateForTransactions);
+        updateStores(updateForTransactions);
+    }
+
+    // TODO(jshah): This needs to be *very* carefully wired to ensure that the synchronised aspect here is not an
+    //  issue. Chances are that this may need to be re-jigged to take a batch, and be connected to the batched commit
+    //  timestamp call.
```
jeremyk-91, line 65: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,163 @@
...
+                Set<CellReference> invalidatedCells = invalidatedLocks.stream()
+                        .flatMap(LockWatchValueCacheImpl::extractTableAndCell)
+                        .collect(Collectors.toSet());
+                KeyedStream.stream(digest.loadedValues())
+                        .filterKeys(cellReference -> !invalidatedCells.contains(cellReference))
+                        .forEach(valueStore::putValue);
+                return null;
+            }
+        });
+        snapshotStore.removeTimestamp(StartTimestamp.of(startTs));
```
jeremyk-91, line 91: This seems like it should be in a finally block? (Unless it's removed elsewhere)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,163 @@
...
+     *     (without this last guarantee, we'd need to store snapshots for all sequences).
+     */
+    private void updateStores(TransactionsLockWatchUpdate updateForTransactions) {
+        Multimap<Sequence, StartTimestamp> reversedMap = createSequenceTimestampMultimap(updateForTransactions);
+        updateForTransactions.events().forEach(event -> {
+            valueStore.applyEvent(event);
+            Sequence sequence = Sequence.of(event.sequence());
+            reversedMap
+                    .get(sequence)
+                    .forEach(timestamp -> snapshotStore.storeSnapshot(sequence, timestamp, valueStore.getSnapshot()));
```
jeremyk-91, line 118: The potential allocations here look scary; I _think_ this is OK because we have vavr, but we should confirm this is fine.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/AtlasLockDescriptorUtils.java
```diff
@@ -40,7 +40,7 @@ private AtlasLockDescriptorUtils() {
         TableReference tableRef = tableRefAndRemainder.get().tableRef();
         ByteString remainingBytes = tableRefAndRemainder.get().remainder();
 
-        return IntStream.range(1, remainingBytes.size() - 2)
+        return IntStream.range(1, remainingBytes.size() - 1)
                 .filter(index -> isZeroDelimiterIndex(remainingBytes, index))
                 .mapToObj(index -> createCellFromByteString(remainingBytes, index))
                 .map(cell -> CellReference.of(tableRef, cell))
```
jeremyk-91, line 46: I realise this was always an issue (and maybe it's fine we don't fix it initially), but I'm kind of worried about what happens if people write blob keys/columns that have a lot of zero bytes in them. The blow-up factor is bounded by 3001 I think. 

Probably not for this PR, but it may be worth having some warning logs and/or requiring users to turn off a safety flag here.

atlasdb-impl-shared/src/main/java/com/palantir/lock/watch/LockWatchReferencesVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public Optional<TableReference> visit(ExactRow reference) {
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<TableReference> visit(ExactCell reference) {
+        return Optional.empty();
+    }
```
jeremyk-91, line 53: What's the plan for the other types? They _do_ have table references too. I don't know exactly how this is going to be used so leaving this up to you

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+                        TABLE.getQualifiedName(), CELL_1.getRowName(), CELL_1.getColumnName())))
+                .build(1L);
+    }
+
+    private static CacheValue createValue(int value) {
+        return CacheValue.of(createBytes(value));
+    }
+
+    private static Cell createCell(int value) {
+        return Cell.create(createBytes(value), createBytes(value + 100));
```
jeremyk-91, line 139: non-actionable: I wonder when someone will call `createBytes(28)`...

jeremyk-91 (commented):
🕐 1h15

I think the broad components are what I'd expect, though would defer to @gmaretic who knows things more precisely here.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/AtlasLockDescriptorUtils.java
```diff
@@ -40,7 +40,7 @@ private AtlasLockDescriptorUtils() {
         TableReference tableRef = tableRefAndRemainder.get().tableRef();
         ByteString remainingBytes = tableRefAndRemainder.get().remainder();
 
-        return IntStream.range(1, remainingBytes.size() - 2)
+        return IntStream.range(1, remainingBytes.size() - 1)
```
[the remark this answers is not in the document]
Jolyon-S: Actually, I'm just going to spin out a separate PR that fixes this and adds the test. However, you might be curious to know that the actual issue here was nothing to do with two zero bytes. Instead, imagine you have a descriptor which is:
```
XXXX 0 1 0 1
```
where `XXXX` is the table part, and row and column are just one byte. Then, your remaining bytes are 101. Now you try and form `IntStream.range(1, 1)` except that the endpoint is exclusive! Regardless, I'll add a test for this, and for two zero bytes on the end.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import com.palantir.atlasdb.keyvalue.api.watch.Sequence;
+import com.palantir.atlasdb.keyvalue.api.watch.StartTimestamp;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public final class SnapshotStoreImpl implements SnapshotStore {
```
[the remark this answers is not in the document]
Jolyon-S: I didn't use the `(Not)ThreadSafe` annotations internally, but I think that it is definitely better for clarity.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    private final Map<Sequence, ValueCacheSnapshot> snapshotMap;
+    private final Multimap<Sequence, StartTimestamp> liveSequences;
```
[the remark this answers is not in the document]
Jolyon-S: I don't think it is essential for correctness - adding the same timestamp twice will be weird, but if you then don't remove it twice, you'll just end up with sequences and snapshots lying around, which is a leak but not a correctness issue.

Still, I don't disagree with making it a set multimap as that is the intended behaviour, but just pointing out that I don't think it is essential.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    private final Map<StartTimestamp, Sequence> timestampMap;
+
+    public SnapshotStoreImpl() {
+        snapshotMap = new HashMap<>();
+        timestampMap = new HashMap<>();
+        liveSequences = HashMultimap.create();
```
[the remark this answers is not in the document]
Jolyon-S: Huh, didn't know this. Will change this, thanks!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/StructureHolder.java
Jolyon-S: I agree, actually. This was copied from internal code.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: That does make sense, although to be honest I expect it to throw when there has been a leader election, and so chances are it'd be gone anyway. I'll add it in a finally block since I think that makes the most sense code-wise, though.

Jolyon-S: In fact, `eventCache.getCommitUpdate` can throw if there has been an election, so definitely makes sense to wrap this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: Internally, we tend to have often >100 snapshots of the cache at any one time. I expect that this is fine, but I'll just dig in vavr's docs to confirm.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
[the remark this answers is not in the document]
jeremyk-91: Fair enough. Yeah, `correctness` here is a bit strong. Having a resource leak is bad though 😅

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: Had a look, it definitely states that this is the kind of pattern that can be handled. Also, the difference in interface between vavr's map and java's basic map means that it should be quite hard for us to accidentally switch these.

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/AtlasLockDescriptorUtils.java
Jolyon-S: That's a good call. For now, I've added a TODO, but I think we should make sure users are aware.

atlasdb-impl-shared/src/main/java/com/palantir/lock/watch/LockWatchReferencesVisitor.java
Jolyon-S: Technically speaking, they do have table references, but I don't want to think that I'm watching a whole table when actually it is just an `ExactCell`. I think, for now, that these should just throw, and one day we'll implement more specific watching.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImplTest.java
Jolyon-S: It's actually fine - before I was using `value + 999`, and it just removes the excess. I know +28 is within a byte's abilities but is negative, but I believe cells are unsigned, right? (Or that the sign doesn't really matter).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                }
+
+                @Override
+                public Void invalidateSome(Set<LockDescriptor> invalidatedLocks) {
+                    Set<CellReference> invalidatedCells = invalidatedLocks.stream()
+                            .flatMap(LockWatchValueCacheImpl::extractTableAndCell)
+                            .collect(Collectors.toSet());
+                    KeyedStream.stream(digest.loadedValues())
+                            .filterKeys(cellReference -> !invalidatedCells.contains(cellReference))
+                            .forEach(valueStore::putValue);
```
gmaretic, line 88: So this again seems like it's ripe for race conditions.

What if this is processed first, and then we process a start transactions request that actually saw fewer events than us? Seems that we should also update the `currentVersion` here to protect against that.

Also, if any of the puts throws, we ignore the rest which is a bit wasteful but probably fine.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/StructureHolder.java
```diff
@@ -0,0 +1,54 @@
...
+    private StructureHolder(Supplier<V> initialValueSupplier) {
+        this.initialValueSupplier = initialValueSupplier;
+        this.structure = initialValueSupplier.get();
+    }
+
+    synchronized void with(Function<V, V> mutator) {
+        structure = mutator.apply(structure);
+    }
+
+    synchronized V getStructure() {
```
gmaretic, line 39:
```suggestion
    synchronized V getSnapshot() {
```

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+        snapshotStore.storeSnapshot(SEQUENCE_2, TIMESTAMP_4, SNAPSHOT_2);
+
+        assertSnapshotsEqualForTimestamp(SNAPSHOT_1, TIMESTAMP_1, TIMESTAMP_2, TIMESTAMP_3);
+        assertSnapshotsEqualForTimestamp(SNAPSHOT_2, TIMESTAMP_4);
+    }
+
+    @Test
+    public void snapshotsNotOverwrittenForSameSequence() {
+        snapshotStore.storeSnapshot(SEQUENCE_1, TIMESTAMP_1, SNAPSHOT_1);
+        snapshotStore.storeSnapshot(SEQUENCE_1, TIMESTAMP_2, SNAPSHOT_2);
```
gmaretic, line 69: Should this even fail silently or throw?

gmaretic (changes requested):
Ok, so went through this in detail and have a bunch of comments. I don't think any of these are super difficult to resolve, and it's possible I misunderstood things so ping me if you think I did

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+import com.palantir.lock.watch.LockWatchVersion;
+import com.palantir.lock.watch.TransactionsLockWatchUpdate;
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * The operations on this class *must* be synchronised to protect against concurrent modification of the cache.
```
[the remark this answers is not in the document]
Jolyon-S: ... uh, yes!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public LockWatchValueCacheImpl(LockWatchEventCache eventCache) {
+        this.eventCache = eventCache;
+        this.valueStore = new ValueStoreImpl();
+        this.snapshotStore = new SnapshotStoreImpl();
+    }
+
+    @Override
+    public synchronized void processStartTransactions(Set<Long> startTimestamps) {
+        TransactionsLockWatchUpdate updateForTransactions =
```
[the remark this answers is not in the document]
Jolyon-S: It doesn't? What happens if two calls are made at the same time? One might get the update, GC, while the other gets the update for a different batch where there is a leader election. The second one goes through, then the first comes back and now you have an inconsistent state.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStore.java
```diff
@@ -0,0 +1,36 @@
...
+import com.palantir.atlasdb.keyvalue.api.watch.StartTimestamp;
+import java.util.Optional;
+
+/**
+ * Stores snapshots of the cache taken at the sequence corresponding to the provided start timestamp. De-duplicates in
+ * the case of multiple start timestamps corresponding to a single sequence (which is likely as they are batched),
+ * and and removes snapshots when they are no longer referenced by any live start timestamps.
+ */
+public interface SnapshotStore {
+    void storeSnapshot(Sequence sequence, StartTimestamp timestamp, ValueCacheSnapshot snapshot);
```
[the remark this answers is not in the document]
Jolyon-S: Reasonable.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    //  issue. Chances are that this may need to be re-jigged to take a batch, and be connected to the batched commit
+    //  timestamp call.
+    @Override
+    public synchronized void updateCache(TransactionDigest digest, long startTs) {
+        try {
+            CommitUpdate commitUpdate = eventCache.getCommitUpdate(startTs);
+            commitUpdate.accept(new Visitor<Void>() {
+                @Override
+                public Void invalidateAll() {
+                    // If this is an election, we should throw. If not and we are not a serialisable transaction, we are
```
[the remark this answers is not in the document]
Jolyon-S: We may still have cache hits, but that really only matters for serialisable transactions - I suppose for snapshot transactions, you may have read legitimately and you don't care about R-W conflicts.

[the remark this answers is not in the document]
Jolyon-S: In reflection, I wonder if we need to throw at all - we could just not publish the values, and any throwing that happens can occur in the `PreCommitCondition` for serialisable transactions.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private void clearCache() {
+        valueStore.reset();
+        snapshotStore.reset();
+    }
+
+    private static Stream<CellReference> extractTableAndCell(LockDescriptor descriptor) {
+        // TODO(jshah): this has potentially large blow-up for cells with lots of zero bytes. We should probably make
+        //  users opt-in or warn when this is the case.
+        return AtlasLockDescriptorUtils.candidateCells(descriptor).stream();
```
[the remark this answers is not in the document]
Jolyon-S: I don't think so, as ALDU still assumes only a single table - all of those cells are for a single table only.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     *  1. For each new event, we apply it to the cache. The effects of this application is described in
+     *     {@link LockWatchValueCache}.
+     *  2. For each transaction, we must ensure that we store a snapshot of the cache at the sequence corresponding
+     *     to the transaction's start timestamp. Note that not every sequence will have a corresponding timestamp, so we
+     *     don't bother storing a snapshot for those sequences. Also note that we know that each call here will only
+     *     ever have new events, and that consecutive calls to this method will *always* have increasing sequences
+     *     (without this last guarantee, we'd need to store snapshots for all sequences).
+     */
+    private void updateStores(TransactionsLockWatchUpdate updateForTransactions) {
```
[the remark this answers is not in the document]
Jolyon-S: I will be extra defensive against this, but it turns out that this is already defended against in LWEC - if you try and get updates for transactions that have a latest version V, and you are on C > V, it will throw. I'm also pretty certain that it only returns updates between C+1 and V inclusive when C < V, so I think we're actually safe here because of the defensiveness of LWEC.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImpl.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.stream.Stream;
+
+public final class ValueStoreImpl implements ValueStore {
+    // TODO(jshah): implement cache eviction based on cache size
+    private final StructureHolder<Map<CellReference, CacheEntry>> values;
+    private final StructureHolder<Set<TableReference>> watchedTables;
+    private final LockWatchVisitor visitor = new LockWatchVisitor();
+
+    public ValueStoreImpl() {
+        values = StructureHolder.create(HashMap::empty);
```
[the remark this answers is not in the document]
Jolyon-S: I actually have a counter-proposal:

```java
private final StructureHolder<io.vavr.collection.Map<CellReference, CacheEntry>> values;
```

Then we don't need to express that for the `HashMap`, since it must be an implementation of vavr's Map.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                eventCache.getUpdateForTransactions(startTimestamps, currentVersion);
+        updateCurrentVersion(updateForTransactions);
+        updateStores(updateForTransactions);
+    }
+
+    // TODO(jshah): This needs to be *very* carefully wired to ensure that the synchronised aspect here is not an
+    //  issue. Chances are that this may need to be re-jigged to take a batch, and be connected to the batched commit
+    //  timestamp call.
+    @Override
+    public synchronized void updateCache(TransactionDigest digest, long startTs) {
```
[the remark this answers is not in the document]
Jolyon-S: Now we no longer throw (also, that technically wasn't a conflict - the conflicts are in `invalidateSome`, and that filters out the writes, not throwing). Either way, I've renamed to `updateCacheOnCommit` to reflect the fact that this should happen at commit time.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: This is actually not an issue since this does _not_ show us all events. The commit update _only_ shows you lock events, and so you cannot even apply the events at commit time.

It doesn't matter if you saw 1000 events in this commit update, then the next start transactions request only showed you one, as those events would _just_ filter out the reads from that transaction (in other words, it would just be more pessimistic). It doesn't affect the correctness or state of the cache.

Put differently, if we decided to randomly not flush X reads to the central cache, it wouldn't make it incorrect, just inefficient. This is the same, except it's not random and hopefully not inefficient :P

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/StructureHolder.java
Jolyon-S: Again, copied from internal. Sure.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImplTest.java
Jolyon-S: You're right, probably should fail loudly.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheEntry.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Represents either:
+ *  1. A value that cannot be currently cached due to a lock being taken out. This has an empty cache value (which
+ *     should never actually need to be read), and a status of LOCKED.
+ *  2. A value that is cached because the last seen event for it was an unlock event (or there was never a lock event in
+ *     the first place). This has a status of UNLOCKED and a value which may be present or empty.
+ */
+@Value.Immutable
+public interface CacheEntry {
```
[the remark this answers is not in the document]
Jolyon-S: That is very fancy! Sure, I'll give it a go.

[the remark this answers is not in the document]
Jolyon-S: Tried it, hated it. 🔥

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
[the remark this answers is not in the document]
gmaretic: Not sure I follow -- you can just filter out any candidates where the tableref is not a watched table

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
gmaretic: So, because the methods are synchronized, and presumably the method that calls `processStartTransactions` is also synchronised (since transaction starter), it is guaranteed that any transaction calling it must have started after the transaction calling update cache on commit. Therefore, anything we put in will have been visible by that other transaction since it was also visible at the start of our transaction. If we remove something, it may have been visible, but whatever

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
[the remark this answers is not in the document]
gmaretic: You should be able to verify the thing is sensible, but yeah, this is just simpler

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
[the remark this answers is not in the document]
Jolyon-S: All candidates for a given descriptor have the same table reference - that part doesn't change, only the divider between row and column.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: Yes - this is doubly guaranteed by the sequence. Either the new transactions have a later or equal sequence, or they don't (and we throw).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -113,53 +119,84 @@ public TransactionScopedCache createTransactionScopedCache(long startTs) {
      */
     private void updateStores(TransactionsLockWatchUpdate updateForTransactions) {
         Multimap<Sequence, StartTimestamp> reversedMap = createSequenceTimestampMultimap(updateForTransactions);
-        updateForTransactions.events().forEach(event -> {
+
+        // There is an edge case where transactions on the current version will not have a snapshot stored.
```
gmaretic, line 123: what is the edge case

gmaretic (approved):
Yay, now it's even more complicated, but it does make sense!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: Updated comment - basically, if there are no events, we don't store a snapshot (that's without the code below, though; made this much clearer).

PR #5381, opened 2021-04-14
LW Cache IV: Per-Transaction Caching

Jolyon-S:
**Goals (and why)**:
* Each transaction must have a view of the global cache
* These per-transaction caches need to also track locally read values, and flush them back to the central cache.

* Transaction cache has a snapshot from the central cache that contains cached values and also locked values.
* When reads hit, this is registered (for serialisable conflict detection later)
* When reads miss, or a write is performed, the value is stored locally
* At commit time, read values should be flushed to the main cache.

Implemented tests for new classes, as well as for LWVCI.

The usual: correctness and performance.

`TransactionScopedCacheImpl`

Week beginning 19th.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheEntry.java
```diff
@@ -45,12 +45,16 @@ static CacheEntry unlocked(CacheValue value) {
...
+    }
+
     enum Status {
         LOCKED,
         UNLOCKED;
 
-        public boolean isUnlocked() {
-            return this == Status.UNLOCKED;
+        boolean isUnlocked() {
+            return this.equals(Status.UNLOCKED);
```
Jolyon-S, line 57: I know the underlying equality check for an enum is `==` but it just feels wrong in hindsight.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
```diff
@@ -0,0 +1,152 @@
...
+                .map(snapshot::getValue)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .filter(CacheEntry::isUnlocked)
+                .map(CacheEntry::value)
+                .collectToMap();
+    }
+
+    @Value.Immutable
+    public interface LocalCacheEntry {
```
Jolyon-S, line 120: No, I don't want to use `derive4j` because I think it's annoying.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionScopedCacheImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+    // TODO(jshah): figure out how we can improve perf given that this is now synchronised. Maybe its fine, but maybe
+    //  we should instead autobatch?
+    @Override
+    public synchronized void write(TableReference tableReference, Cell cell, CacheValue value) {
+        valueStore.cacheLocalWrite(tableReference, cell, value);
+    }
+
+    // TODO(jshah): as above. Equally, we should probably use the async value loader, then have it get afterwards,
+    //  thus reducing the time spent in this method.
```
Jolyon-S, line 45: I quite frankly couldn't motivate myself to wire the async stuff in this PR as I think that'd just make the complexity insane when taken in all at once.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
```diff
@@ -0,0 +1,152 @@
...
+                .filter(Objects::nonNull)
+                .map(LocalCacheEntry::value)
+                .collectToMap();
+    }
+
+    private Map<CellReference, CacheValue> getSnapshotValues(Set<CellReference> cellReferences) {
+        return KeyedStream.of(cellReferences)
+                .map(snapshot::getValue)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
Jolyon-S, line 113: If only we had java 11...

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
```diff
@@ -31,6 +84,175 @@ public void before() {
...
+
+        // Throws this message because the leader election cleared the info entirely (as opposed to us knowing that
+        // there was an election)
+        assertThatThrownBy(() -> valueCache.updateCacheOnCommit(scopedCache1.getDigest(), TIMESTAMP_1))
+                .isExactlyInstanceOf(TransactionLockWatchFailedException.class)
+                .hasMessage("start or commit info not processed for start timestamp");
+    }
+
+    @Test
+    public void lockUpdatesPreventCachingAndUnlockUpdatesAllowItAgain() {
```
gmaretic, line 140: Very nice test

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStore.java
```diff
@@ -0,0 +1,46 @@
...
+
+    void cacheRemoteReads(TableReference tableReference, Map<Cell, byte[]> remoteReadValues);
+
+    void cacheEmptyReads(TableReference tableReference, Set<CellReference> emptyCells);
+
+    Map<CellReference, CacheValue> getCachedValues(Set<CellReference> cellReferences);
+
+    /**
+     * Contains a map of all the values that were read remotely and stored locally (filtering out those that were
+     * unable to be cached due to values being locked). Also note that writes do not appear in the digest.
```
gmaretic, line 43: We need to exclude writes because they would otherwise become visible to other transactions that started in the same batch as us?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        static LocalCacheEntry hit(CacheValue value) {
+            return ImmutableLocalCacheEntry.builder()
+                    .status(Status.HIT)
+                    .value(value)
+                    .build();
+        }
+
+        enum Status {
+            READ,
+            WRITE,
```
gmaretic, line 150: The distinction between WRITE and HIT here is just for cleaner code, we only really care about whether it was a remote read or not?

gmaretic (approved):
A bunch of comments and some questions to sanity check my understanding, but overall looks quite good! Test coverage is pretty decent :)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
```diff
@@ -0,0 +1,154 @@
...
+            localUpdates.put(cellReference, LocalCacheEntry.write(value));
+        }
+    }
+
+    @Override
+    public void cacheRemoteReads(TableReference tableReference, Map<Cell, byte[]> remoteReadValues) {
+        if (snapshot.isWatched(tableReference)) {
+            KeyedStream.stream(remoteReadValues)
+                    .mapKeys(cell -> CellReference.of(tableReference, cell))
+                    .filterKeys(snapshot::isUnlocked)
```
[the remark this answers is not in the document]
Jolyon-S: No, this means that a lock has been taken out on this cell. If I understand correctly, that's supposed to mean that we cannot read this value from the cache because it is currently being changed (because some other transaction holds the lock here)?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionScopedCacheImpl.java
```diff
@@ -0,0 +1,95 @@
...
+        return TransactionDigest.of(valueStore.getTransactionDigest());
+    }
+
+    private static Map<Cell, byte[]> filterEmptyValues(Map<CellReference, CacheValue> snapshotCachedValues) {
+        return KeyedStream.stream(snapshotCachedValues)
+                .filter(value -> value.value().isPresent())
+                .map(value -> value.value().get())
+                .mapKeys(CellReference::cell)
+                .collectToMap();
+    }
```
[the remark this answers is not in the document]
Jolyon-S: Reasonable - I'll split it up along the lines you've suggested

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueCacheSnapshotImpl.java
```diff
@@ -30,13 +30,18 @@
...
-        return enabledTables().contains(tableReference);
+    default Optional<CacheEntry> getValue(CellReference tableAndCell) {
+        return values().get(tableAndCell).toJavaOptional();
     }
 
     @Override
-    default Optional<CacheEntry> getValue(CellReference tableAndCell) {
-        return values().get(tableAndCell).toJavaOptional();
+    default boolean isUnlocked(CellReference tableAndCell) {
+        return getValue(tableAndCell).map(CacheEntry::isUnlocked).orElse(true);
```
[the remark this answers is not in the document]
Jolyon-S: No, this is that the default assumes the value is unlocked unless proven otherwise (nothing to do with the table being watched IIRC).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStore.java
Jolyon-S: As much as anything, because the taking out of locks will invalidate those values anyway

Jolyon-S: (and to be as conservative as possible)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
Jolyon-S: No - the distinction is for later, when we only want `HIT` values for conflict checking (we don't need to do serialisable conflict checking for `READ` values as those are read from the DB, whereas we do need to check the `HIT` values as they have not touched the conflict checking flow at all).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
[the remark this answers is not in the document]
gmaretic: Yeah I was asking more around why this being locked means we cannot cache. But now that I think about it some more, since we have succeeded on a transactional read, the transaction that held the lock is already done by this point, but we do not know if we have seen the old or the new version, so we do not want to cache

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueCacheSnapshotImpl.java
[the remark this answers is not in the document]
gmaretic: But that is incorrect. It is only unlocked because the semantics of being watched means it's unlocked unless explicitly locked. We do not have such a guarantee for unwatched tables. I know I'm splitting hairs though since it literally does not matter for unwatched tables, but we should just be defensive and return false as the default

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
gmaretic: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueCacheSnapshotImpl.java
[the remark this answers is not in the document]
Jolyon-S: Ah, I think I see your point. I think we need to do this:

1. If table is unwatched, return false;
2. If table is watched, return false if the value is locked, else return true (even if the value is absent)

I wanted to ensure that we return true if no value is there (because that means we just don't have a cached value or a lock taken out), but didn't check whether the table was watched.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStoreImpl.java
Jolyon-S: Yeah, I agree that in those cases, derive4j might be nicer, but here I think not.

PR #5383, opened 2021-04-14
[PDS-176064] burning smoke and bullet fire | getSortedColumns now avoids local writes correctly

jeremyk-91:
**Goals (and why)**:
- Fix a bug in `getSortedColumns`: when checking our reads against the read-only transaction, we should only check cells for which there were no _local_ writes. This brings this in line with the behaviour of `getRowsColumnRange`.

- When checking our reads against the read-only transaction, we should only check cells for which there were no _local_ writes.
- I wrote two tests that would have previously failed but now pass. (The another-table test is just there because it's something that could accidentally go wrong as well.)

- Added new tests that are evidence of the previous bug.

- Are there any other edge cases I missed out?

**Where should we start reviewing?**: `SerializableTransactionTest`


atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1201,6 +1204,101 @@ public void testGetSortedColumnsReadWriteConflictAtEndOfBatch() {
...
+        List<Map.Entry<Cell, byte[]>> sortedColumnValues =
+                Streams.stream(sortedColumns).collect(Collectors.toList());
+        List<Cell> sortedCellsRead =
+                sortedColumnValues.stream().map(Entry::getKey).collect(Collectors.toList());
+        sanityCheckOnSortedCells(rows, sortedCellsRead, cellsWrittenOriginally);
+        assertThat(sortedColumnValues)
+                .contains(
+                        Maps.immutableEntry(cellsWrittenOriginally.get(0), newValue1),
+                        Maps.immutableEntry(cellsWrittenOriginally.get(17), newValue2));
+        t1.commit();
```
sudiksha27, line 1233: nit: given the name of the test - 
```suggestion
        assertThatCode(t1::commit).doesNotThrowAnyException();
```

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1201,6 +1204,101 @@ public void testGetSortedColumnsReadWriteConflictAtEndOfBatch() {
...
+                Streams.stream(sortedColumns).collect(Collectors.toList());
+        List<Cell> sortedCellsRead =
+                sortedColumnValues.stream().map(Entry::getKey).collect(Collectors.toList());
+        sanityCheckOnSortedCells(rows, sortedCellsRead, cellsWrittenOriginally);
+        assertThat(sortedColumnValues)
+                .doesNotContain(
+                        Maps.immutableEntry(cellsWrittenOriginally.get(0), newValue1),
+                        Maps.immutableEntry(cellsWrittenOriginally.get(17), newValue2));
+        t1.commit();
+    }
```
sudiksha27, line 1263: same as above.

sudiksha27 (approved):
Makes sense!

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1201,6 +1204,101 @@ public void testGetSortedColumnsReadWriteConflictAtEndOfBatch() {
...
+        List<Map.Entry<Cell, byte[]>> sortedColumnValues =
+                Streams.stream(sortedColumns).collect(Collectors.toList());
+        List<Cell> sortedCellsRead =
+                sortedColumnValues.stream().map(Entry::getKey).collect(Collectors.toList());
+        sanityCheckOnSortedCells(rows, sortedCellsRead, cellsWrittenOriginally);
+        assertThat(sortedColumnValues)
+                .doesNotContain(
+                        Maps.immutableEntry(cellsWrittenOriginally.get(0), newValue1),
+                        Maps.immutableEntry(cellsWrittenOriginally.get(17), newValue2));
+        t1.commit();
```
jeremyk-91, line 1262:
```suggestion
       assertThatCode(t1::commit).doesNotThrowAnyException();
```
following @sudiksha27's comment after this one so I can batch them in the ui

PR #5395, opened 2021-04-22
[LW] CodeGen for caching and register watches

gmaretic:
**Goals (and why)**:
Allow users to specify gets in a table should be cached

We currently only support caching with serializable cell conflict handling. TMs.create checks your schemas and if you have any tables that require caching, it schedules a job to register the appropriate lock watches every 2 minutes.
All the actual caching logic is going to live in the transaction impl, and users will automatically cache all gets on tables that are cached (as opposed to specifying if a get is cached or not)

Just some sanity checks

This will, of course, clash with any other registration of lock watches. Luckily the only user of lock watches will not have a schema with a caching table, so we should be fine as in that case no job is scheduled.

This is actually really small

Soon™️


atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -197,6 +200,78 @@ public void v2SchemaTablesCanBeGeneratedWithAllRowAndColumnComponents() {
...
+    @Test
+    public void cannotAddTableDefinitionWithCachingAndIncompatibleConflictDetection() {
+        Schema schema = new Schema("Table", TEST_PACKAGE, Namespace.EMPTY_NAMESPACE);
+        TableDefinition tableDef = new TableDefinition() {
+            {
+                allSafeForLoggingByDefault();
+                sweepStrategy(TableMetadataPersistence.SweepStrategy.THOROUGH);
+
+                rowName();
+                hashFirstRowComponent();
```
Jolyon-S, line 212: Stuff like this is fine by me, but for a min repro you can surely remove it?

Jolyon-S (changes requested):
I'm uncertain as to whether we need the lock watch registrar at all - see comments, and we can discuss offline.

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
```diff
@@ -197,6 +200,58 @@ public void v2SchemaTablesCanBeGeneratedWithAllRowAndColumnComponents() {
...
+        TableDefinition tableDef = new TableDefinition() {
+            {
+                rowName();
+                rowComponent("key", ValueType.BLOB);
+                noColumns();
+                enableCaching();
+            }
+        };
+        assertThatThrownBy(() -> schema.addTableDefinition(TEST_TABLE_NAME, tableDef))
+                .as("Jolyon loves this stuff")
```
Jolyon-S, line 215: I do?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -522,6 +517,12 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
                 .<ConflictTracer>map(Function.identity())
                 .orElse(ConflictTracer.NO_OP);
 
+        Callback<TransactionManager> callbacks = new Callback.CallChain<>(
+                timelockConsistencyCheckCallback(config(), runtime.get(), lockAndTimestampServices),
+                targetedSweep.singleAttemptCallback(),
+                asyncInitializationCallback(),
+                createClearsTable());
+
```
Jolyon-S, line 525: Just revert this change.

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
```diff
@@ -591,6 +592,15 @@ private TransactionManager serializableInternal(@Output List<AutoCloseable> clos
...
                 closeables);
 
+        Set<LockWatchReference> lockWatchesToRegister = schemas().stream()
+                .map(Schema::getLockWatches)
+                .flatMap(Set::stream)
+                .collect(Collectors.toSet());
+
+        if (!lockWatchesToRegister.isEmpty()) {
+            lockAndTimestampServices.lockWatcher().registerPreciselyWatches(lockWatchesToRegister);
+        }
```
Jolyon-S, line 602: Please add javadoc either here, or preferably on the method in LWM, that this will be overwritten if anyone else calls the `registerPreciselyWatches` method. It might be obvious, but I'd prefer that we document it so we don't shoot ourselves in the foot.

Jolyon-S (commented):
Couple of comments / nits, otherwise looks good.

I don't know schema / generation stuff too well in Atlas - I assume we are not persisting this fact? So the fact that we're adding a new field is fine?

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
gmaretic: 👍

atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
gmaretic: I actually think this is the better place for it

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaTest.java
gmaretic: It's your favourite nit!

Jolyon-S (approved):
Nice! That's definitely the best solution.

PR #5396, opened 2021-04-23
[PDS-146088] Time out when borrowing objects from the Cassandra object pool

Jolyon-S:
**Goals (and why)**:
Turns out that we fail to return objects to the Cassandra object pool if the function hangs.

* Refactored the `SafeShutdownRunner` to be more like how it used to be, and split out a separate `TimedRunner` to actually just run things in a timed manner.
* Run functions on Cassandra connections wrapped in a timed runner.

* Added unit tests for the `TimedRunner`. No tests for CCPC.

* I've set a default of 10 minutes to time out, but I don't actually know how long we expect requests to Cassandra to take. Maybe we should be more conservative and set to one hour, and add a metric to analyse?
* I've removed some old metrics that we don't really need - is that OK?

* `CassandraClientPoolingContainer`

Probably before EoD Tuesday next week (27th April).


atlasdb-commons/src/main/java/com/palantir/util/TimedRunner.java
```diff
@@ -0,0 +1,101 @@
...
+    public interface TaskContext<T> {
+        /**
+         * Executed to shut down an object.
+         */
+        Callable<T> task();
+
+        /**
+         * Executed only if the shutdownCallback times out or otherwise has an exception.
+         */
+        Runnable taskFailureHandler();
```
jeremyk-91, line 82: nit: javadocs here need updating (it's now just a task as opposed to shutdownCallback, and the task is general I think)

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableManipulationIntegrationTest.java
```diff
@@ -175,16 +175,16 @@ public void tableCreationAppearsToSucceedButIsNoop() {
...
         assertThatThrownBy(() -> kvs.put(LOWER_LOWER, ImmutableMap.of(CELL, SECOND_BYTE_ARRAY), 1))
-                .isInstanceOf(RetryLimitReachedException.class);
+                .hasCauseInstanceOf(RetryLimitReachedException.class);
     }
 
     private void createTablesIgnoringException() {
         assertThatThrownBy(() -> kvs.createTables(TABLES.stream()
                         .collect(Collectors.toMap(x -> x, no -> AtlasDbConstants.GENERIC_TABLE_METADATA))))
-                .isInstanceOf(IllegalStateException.class);
+                .hasCauseInstanceOf(IllegalStateException.class);
```
jeremyk-91, line 187: I assume there was a decision here that got changed - these tests should still verify isInstanceOf if I'm not mistaken

jeremyk-91 (approved):
looks good, modulo comments

atlasdb-cassandra-integration-tests/src/test/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraKeyValueServiceTableManipulationIntegrationTest.java
Jolyon-S: No - these go through `TimedRunner`, which was throwing a SafeRuntimeException, not the original exception. However, I think I'm going to refactor that because it is less invasive.

PR #5400, opened 2021-04-26
[LW] Resolve TODOs

Jolyon-S:
**Goals (and why)**:
Now that the main piece of work is through, I wanted to resolve the TODOs sooner rather than later.

Resolve TODOs. This is several small features, like changing things to be batched, adding necessary things for later, etc.

Updated tests accordingly.

Not going to affect the wiring stuff.

`LockWatchValueCache`

This week.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
@@ -31,15 +31,23 @@ import com.palantir.lock.watch.LockWatchEvent;
 import com.palantir.lock.watch.LockWatchEventCache;
 import com.palantir.lock.watch.LockWatchVersion;
 import com.palantir.lock.watch.TransactionsLockWatchUpdate;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.concurrent.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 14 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImpl.java
```diff
@@ -32,13 +32,13 @@
 import io.vavr.collection.HashSet;
 import java.util.stream.Stream;
 
-public final class ValueStoreImpl implements ValueStore {
+final class ValueStoreImpl implements ValueStore {
     // TODO(jshah): implement cache eviction based on cache size
```
Jolyon-S, line 36: I think this should be a different PR.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -202,8 +210,17 @@ private void clearCache() {
     }
 
     private static Stream<CellReference> extractTableAndCell(LockDescriptor descriptor) {
-        // TODO(jshah): this has potentially large blow-up for cells with lots of zero bytes. We should probably make
-        //  users opt-in or warn when this is the case.
-        return AtlasLockDescriptorUtils.candidateCells(descriptor).stream();
+        List<CellReference> candidateCells = AtlasLockDescriptorUtils.candidateCells(descriptor);
```
gmaretic, line 213: We can still pass in the actual schema and only try to match with the cached tables from the schema if this is something that we actually hit

gmaretic (approved):
LGTM

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
Jolyon-S: Let's leave that optimisation for now, but possibly something we can visit later.

PR #5402, opened 2021-04-26
[LW] Wire cache into transactions

gmaretic:
**Goals (and why)**:


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1569,12 +1581,14 @@ protected boolean shouldDeleteAndRollback() {
     @Override
     public final void delete(TableReference tableRef, Set<Cell> cells) {
         putInternal(tableRef, Cells.constantValueMap(cells, PtBytes.EMPTY_BYTE_ARRAY));
+        cells.forEach(cell -> cache.write(tableRef, cell, CacheValue.empty()));
```
Jolyon-S, line 1584: I can add a method for `cache.delete`, would make this nicer.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1569,12 +1581,14 @@ protected boolean shouldDeleteAndRollback() {
...
     }
 
     @Override
     public void put(TableReference tableRef, Map<Cell, byte[]> values) {
         ensureNoEmptyValues(values);
         putInternal(tableRef, values);
+        values.forEach((cell, value) -> cache.write(tableRef, cell, CacheValue.of(value)));
```
Jolyon-S, line 1591: didn't realise this was done in a map, I can make this just accept a map instead of individual values.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1814,6 +1828,8 @@ private void commitWrites(TransactionService transactionService) {
                         "commitPutCommitTs",
                         () -> putCommitTimestamp(commitTimestamp, commitLocksToken, transactionService));
 
+                lockWatchValueCache.updateCacheOnCommit(cache.getDigest(), getStartTimestamp());
+
```
Jolyon-S, line 1832: This might live better in the same place that `LockWatchEventCache` does its update on commit, because then we can batch (this is the central cache you're updating, which means this could slow things done if we're not careful).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -184,6 +192,10 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
                 timelockService.startIdentifiedAtlasDbTransactionBatch(conditions.size());
         Preconditions.checkState(conditions.size() == responses.size(), "Different number of responses and conditions");
         try {
+            lockWatchValueCache.processStartTransactions(responses.stream()
```
Jolyon-S, line 195: again, this should probably live in the same place as `LockWatchEventCache` and its processing.

Jolyon-S: there may even be an argument of pairing the two of them in some handy interface.

Jolyon-S (commented):
There may be some improvements we can make, comments below.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: Yup, that would make sense 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: 👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
gmaretic: Yeah, there is a comment, this should live in transaction starter, but I need to do more annoying wiring first

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: You mean in the CommitTimestampGetters? That probably makes sense, yeah. Just to double check -- we only actually update with the reads, so even if the transaction fails after we do this update, it's still correct, right?

Jolyon-S: the former: yes (I think that's what it is called these days).

the latter: only update reads, and that _should_ happen after we verify the PCC, so if there were R-W conflicts, those reads should not end up in the cache either (we'll have to be careful about which part of the lifecycle that happens in).

It is worth noting that, internally, we actually only update the cache once the transaction has fully committed successfully, but I think it is technically fine to do it during commit here (with the caveat above that the PCC is done at the correct time).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCache.java
```diff
@@ -31,10 +32,16 @@
  * the central cache (taking in to account the since-locked descriptors), as well as checking for conflicts for
  * serializable transactions by adding a check in the {@link com.palantir.atlasdb.transaction.api.PreCommitCondition}.
  */
-public interface LockWatchValueCache {
+public interface LockWatchValueCache extends ValueCacheUpdater {
+    @Override
```
gmaretic, line 36: I hate the atlas pattern of hiding inherited methods

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
gmaretic: Did the wiring, will look into unifying the caches to make code slightly nicer

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -439,77 +438,6 @@ public void testLockAfterGet() throws Exception {
         mockery.assertIsSatisfied();
     }
 
-    @Ignore("Was ignored long ago, and now we need to fix the mocking logic.")
```
gmaretic, line 442: drive by

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCache.java
```diff
@@ -31,10 +32,16 @@
...
     void processStartTransactions(Set<Long> startTimestamps);
 
-    void updateCacheOnCommit(ValueDigest digest, long startTs);
+    @Override
+    void updateCacheOnCommit(Set<Long> startTimestamps);
```
Jolyon-S, line 40: You can merge in https://github.com/palantir/atlasdb/pull/5408/files for this change (ideally, we get that merged before this).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCache.java
```diff
@@ -31,10 +32,16 @@
...
+
+    default void processFailedTransactions(Set<Long> startTimestamp) {
```
Jolyon-S, line 42: hmm, I suppose we need to remove the state if the transaction failed? I can implement this in a subsequent PR as you mentioned.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -304,15 +305,7 @@ private void updateRangeEnd(Cell end) {
...
-        try {
-            return getWithLoader(
-                            tableRef,
-                            cells,
-                            (tableReference, toRead) -> Futures.immediateFuture(super.get(tableRef, toRead)))
-                    .get();
-        } catch (InterruptedException | ExecutionException e) {
-            throw Throwables.rewrapAndThrowUncheckedException(e.getCause());
-        }
+        return super.get(tableRef, cells);
```
Jolyon-S, line 308: Why aren't we getting with loader? We'll surely still need to do R-W conflicts for normal remote reads?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/ShouldNotDeleteAndRollbackTransaction.java
```diff
@@ -85,6 +87,7 @@ public ShouldNotDeleteAndRollbackTransaction(
                 keyValueService,
                 null,
                 null,
+                new LockWatchValueCacheImpl(NoOpLockWatchEventCache.create()),
```
Jolyon-S, line 90: I think we should create a no-op value cache.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -771,13 +778,17 @@ private void extractLocalWritesForRow(
...
+                cells,
+                (table, uncached) ->
+                        getInternal("get", tableRef, cells, immediateKeyValueService, immediateTransactionService));
     }
 
     @Override
     @Idempotent
     public ListenableFuture<Map<Cell, byte[]>> getAsync(TableReference tableRef, Set<Cell> cells) {
+        // todo(gmaretic): make this use cache as well?
         return getInternal("getAsync", tableRef, cells, keyValueService, defaultTransactionService);
```
Jolyon-S, line 792: the cache's `get` method is literally just a `AtlasFutures.getUnchecked` on the `getInternal` - not against just renaming and exposing that so that we have async as well.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -321,6 +333,8 @@ protected Transaction createTransaction(
                 keyValueService,
                 timelockService,
                 lockWatchManager,
+                // cannot use caching if we do not talk to timelock
+                new LockWatchValueCacheImpl(NoOpLockWatchEventCache.create()),
```
Jolyon-S, line 337: Again, a no-op cache would be nice.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
```diff
@@ -78,7 +78,7 @@
             ImmutableSet.of(LockWatchReferences.entireTable(TABLE.getQualifiedName())));
 
     private LockWatchEventCache eventCache;
-    private LockWatchValueCache valueCache;
+    private LockWatchValueCacheImpl valueCache;
```
Jolyon-S, line 81: Why is this necessary?

lock-api/src/main/java/com/palantir/lock/cache/ValueCacheUpdater.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.lock.cache;
+
+import java.util.Set;
+
+public interface ValueCacheUpdater {
```
Jolyon-S, line 21: annoying that we need this, but I buy it.

Jolyon-S: Although actually there might be a decomp where we pair the caches _and_ don't need this.

lock-api/src/main/java/com/palantir/lock/cache/ValueCacheUpdater.java
```diff
@@ -0,0 +1,37 @@
...
+    ValueCacheUpdater NO_OP = new ValueCacheUpdater() {
```
Jolyon-S, line 22: Ugly, but I think this will do _for now_. Can we add a TODO to maybe revisit this?

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -62,7 +63,8 @@ public static RemoteTimelockServiceAdapter create(
                 rpcClient,
                 conjureClient,
                 new LegacyLeaderTimeGetter(conjureClient),
-                RequestBatchersFactory.create(lockWatchEventCache, namespace, Optional.empty()));
+                RequestBatchersFactory.create(
+                        lockWatchEventCache, ValueCacheUpdater.NO_OP, namespace, Optional.empty()));
```
Jolyon-S, line 67: Checked that everywhere that calls this has a no-op event cache. In that case we may as well surely just inline the no op event cache here too?

lock-api/src/test/java/com/palantir/lock/client/TransactionStarterTest.java
```diff
@@ -55,6 +57,7 @@
 
     private final StartTransactionsLockWatchEventCache lockWatchEventCache =
             spy(StartTransactionsLockWatchEventCache.createForTests());
+    private final ValueCacheUpdater valueCache = mock(ValueCacheUpdater.class);
```
Jolyon-S, line 60: we do have a handy no-op implementation we could use, but this is fine, don't really care.

Jolyon-S (changes requested):
I think we could technically merge like this or something similar, but I wonder if that'd actually be more overhead for us than upside: we'd have to revisit and re-wire things. I honestly think that maybe we should just couple the caches now instead of later (happy to do that in a PR if you want).

lock-api/src/main/java/com/palantir/lock/cache/ValueCacheUpdater.java
Jolyon-S: I've made this better in https://github.com/palantir/atlasdb/pull/5409

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: That's probably outiside the scope here

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/NoOpLockWatchManager.java
```diff
@@ -16,23 +16,28 @@
...
 import com.palantir.lock.watch.LockWatchVersion;
 import com.palantir.lock.watch.TransactionsLockWatchUpdate;
 import java.util.Optional;
 import java.util.Set;
 
-public final class NoOpLockWatchManager extends LockWatchManager {
-    private final LockWatchEventCache eventCache;
+public final class NoOpLockWatchManager extends LockWatchManagerInternal {
+    private final LockWatchCache cache;
+    private final TransactionScopedCache transactionScopedCache;
```
Jolyon-S, line 32: I expect I'll find out why this is happening - the no op value cache should happily hand out no op transaction caches, so this feels weird.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -36,40 +43,54 @@
...
-public final class LockWatchManagerImpl extends LockWatchManager implements AutoCloseable {
+public final class LockWatchManagerImpl extends LockWatchManagerInternal {
 
     private static final Logger log = LoggerFactory.getLogger(LockWatchManagerImpl.class);
 
     private final Set<LockWatchReferences.LockWatchReference> referencesFromSchema;
     private final Set<LockWatchReferences.LockWatchReference> lockWatchReferences = ConcurrentHashMap.newKeySet();
-    private final LockWatchEventCache lockWatchEventCache;
+    private final LockWatchCache lockWatchCache;
+    private final LockWatchValueScopingCache valueScopingCache;
```
Jolyon-S, line 53: There should be a method on LWC to get LWVSC, so no need to keep as a field.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
Jolyon-S: https://github.com/palantir/atlasdb/pull/5413 implements `getAsync`, but we can merge that in the next PR (probably too painful to do _yet another_ rebase here :P)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -240,7 +237,7 @@ public Transaction delegate() {
             try {
                 result = runTaskThrowOnConflict(wrappedTask, tx);
             } finally {
-                lockWatchEventCache.removeTransactionStateFromCache(getTimestamp());
+                lockWatchManager.removeTransactionStateFromCache(getTimestamp());
```
Jolyon-S, line 240: It's worth noting a discrepancy in the implementation that we might need to be careful of: LWEC requires this to be called to clear state, whereas LWVC will clear the state on commit (and so this is only for use in an exceptional context, although it doesn't matter if it is called also). Just need to make sure that we definitely do the commit method before this, although I expect we probably do.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImplTest.java
```diff
@@ -40,6 +41,9 @@
     @Mock
     private LockWatchEventCache lockWatchEventCache;
 
+    @Mock
+    private LockWatchValueScopingCache valueScopingCache;
```
Jolyon-S, line 45: Probably doing a couple of really basic tests confirming that the methods pass through to the right caches

lock-api-objects/src/main/java/com/palantir/lock/watch/LockWatchStateUpdate.java
```diff
@@ -33,6 +34,7 @@
     @JsonSubTypes.Type(value = LockWatchStateUpdate.Success.class, name = LockWatchStateUpdate.Success.TYPE),
     @JsonSubTypes.Type(value = LockWatchStateUpdate.Snapshot.class, name = LockWatchStateUpdate.Snapshot.TYPE)
 })
+@DoNotMock
```
Jolyon-S, line 37: I've never seen this before - why do we need it?

lock-api/src/main/java/com/palantir/lock/watch/LockWatchCacheImpl.java
```diff
@@ -34,6 +29,10 @@ public LockWatchCacheImpl(LockWatchEventCache eventCache, LockWatchValueCache va
         this.valueCache = valueCache;
     }
 
+    public static LockWatchCache noop() {
```
Jolyon-S, line 32: Since we generally call classes `NoOp`,  that implies this should be `noOp()`

Jolyon-S (commented):
Nearly there, just a few small style things.

I think the reason I'm being so pedantic on this PR is that this actually just wires production code to use the value cache, which I'm probably being more scared than I need to be when it comes to enabling!

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/NoOpLockWatchManager.java
Jolyon-S: (Update: didn't work out why you did this - why not just create no op caches on the fly?)

gmaretic: We can inline, sure

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
gmaretic: There cannot be one, for the same reason that LWVSC exists in the first place

Jolyon-S: ah damn

Jolyon-S: (Might be worth noting that or commenting that we do _not_ need to update the transaction state in both LWC and LWVSC).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
gmaretic: The only time we do not do the commit method before is if the transaction fails, and we will never do it

lock-api-objects/src/main/java/com/palantir/lock/watch/LockWatchStateUpdate.java
Jolyon-S: Discussed - fine, I buy it.

gmaretic: Because mocking this makes tests explode since even the noop implementations depend on being able to read stuff

lock-api/src/main/java/com/palantir/lock/watch/LockWatchCacheImpl.java
gmaretic: I don't even know why I do these things when I know you will complain about it 🗡️

Jolyon-S: 🔥

Jolyon-S (approved):
As far as I can tell, for the average user, _nothing_ should change here:
1. LWEC works as before
2. Reads will now go through the cache - there is overhead in creating one, but then it will just read immediately because no tables are watched.

PR #5403, opened 2021-04-27
[LW] Bound Cache by size

Jolyon-S:
**Goals (and why)**:
We should evict values based on size and LRU.

Implement size-based eviction via a caffeine cache.

Added extra tests.

Should be fine.

`ValueStoreImpl`

This week (when it is done).


gmaretic (approved):
Surprisingly non-janky 🥇

PR #5405, opened 2021-04-27
[LW] Validate reads from the cache

Jolyon-S:
**Goals (and why)**:
To avoid data corruption and give ourselves confidence that our code works, all reads through the cache (and after a while, only a percentage of reads) should be validated against the remote result.

* Implemented another transaction scoped cache that, on get, reads from the remote, and then remote reads performed by the delegate just read from the results already retrieved (to avoid multiple reads).
* Throw on failure - when the resilient layer is made, this should cause the cache to fall back to a no-op implementation.

Added unit tests.

Did I miss anything?

`ValidatingTransactionScopedCache`

This week.


atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/AtlasLockDescriptorUtils.java
```diff
@@ -40,11 +47,22 @@ private AtlasLockDescriptorUtils() {
...
         ByteString remainingBytes = tableRefAndRemainder.get().remainder();
 
-        return IntStream.range(1, remainingBytes.size() - 1)
+        List<CellReference> candidateCells = IntStream.range(1, remainingBytes.size() - 1)
                 .filter(index -> isZeroDelimiterIndex(remainingBytes, index))
                 .mapToObj(index -> createCellFromByteString(remainingBytes, index))
                 .map(cell -> CellReference.of(tableRef, cell))
                 .collect(Collectors.toList());
+
+        if (candidateCells.size() > CELL_BLOWUP_THRESHOLD) {
```
Jolyon-S, line 56: Moved here since we actually do this in multiple places.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -31,35 +31,26 @@
...
-
     private final LockWatchEventCache eventCache;
     private final ValueStore valueStore;
     private final SnapshotStore snapshotStore;
+    private final double validationProbability;
 
     private volatile Optional<LockWatchVersion> currentVersion = Optional.empty();
 
-    public LockWatchValueCacheImpl(LockWatchEventCache eventCache) {
+    public LockWatchValueCacheImpl(LockWatchEventCache eventCache, long maxCacheSize, double validationProbability) {
```
Jolyon-S, line 50: We probably need to introduce a config block for this.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImpl.java
```diff
@@ -132,7 +125,7 @@ public TransactionScopedCache createTransactionScopedCache(long startTs) {
      *     ever have new events, and that consecutive calls to this method will *always* have increasing sequences
      *     (without this last guarantee, we'd need to store snapshots for all sequences).
      */
-    private void updateStores(TransactionsLockWatchUpdate updateForTransactions) {
+    private synchronized void updateStores(TransactionsLockWatchUpdate updateForTransactions) {
```
Jolyon-S, line 128: Synchronised to be safe - these should be executed within synchronised methods anyway, but since not every public method on this class is synchronised, I wanted to be doubly certain.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCacheTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    public void validationDoesNotReadFromRemoteWhenItShouldNotValidate() {
+        TransactionScopedCache delegate = TransactionScopedCacheImpl.create(snapshotWithSingleValue());
+        TransactionScopedCache validatingCache = new ValidatingTransactionScopedCache(delegate, () -> false);
+
+        validatingCache.get(TABLE, ImmutableSet.of(CELL_1), valueLoader);
+        verify(valueLoader, never()).apply(any(), any());
+    }
+
```
gmaretic, line 83: Add a test that we still read from remote when our cache does not have the entry

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
```diff
@@ -83,7 +83,18 @@
     @Before
     public void before() {
         eventCache = LockWatchEventCacheImpl.create(MetricsManagers.createForTests());
-        valueCache = new LockWatchValueCacheImpl(eventCache);
+        valueCache = new LockWatchValueCacheImpl(eventCache, 20_000, 0.0);
+    }
+
+    @Test
+    public void valueCacheCreatesValidatingTransactionCaches() {
```
gmaretic, line 90: This test is testing that it caches reads, not that it does validation

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class ValidatingTransactionScopedCache implements TransactionScopedCache {
+    private static final Logger log = LoggerFactory.getLogger(ValidatingTransactionScopedCache.class);
+
+    private final TransactionScopedCache delegate;
+    private final Supplier<Boolean> validationSupplier;
```
gmaretic, line 41: Rename to something more descriptive, `shouldValidate`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
```diff
@@ -0,0 +1,110 @@
...
+    public Map<Cell, byte[]> get(
+            TableReference tableReference,
+            Set<Cell> cell,
+            BiFunction<TableReference, Set<Cell>, ListenableFuture<Map<Cell, byte[]>>> valueLoader) {
+        if (validationSupplier.get()) {
+            Map<Cell, byte[]> remoteReads = AtlasFutures.getUnchecked(valueLoader.apply(tableReference, cell));
+            Map<Cell, byte[]> cacheReads =
+                    delegate.get(tableReference, cell, (table, cells) -> getCells(remoteReads, cells));
+
+            validateCacheReads(tableReference, remoteReads, cacheReads);
```
gmaretic, line 74: Ah I wanted to complain that throwing is not ideal, but I see in the comment that that is to come -- is the idea to then disable caching for all transactions until a bounce happens?

gmaretic (commented):
Mostly looks good, some comments on tests and wondering about how we deal with failures eventually, rather than just failing the transaction

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
Jolyon-S: IIRC this test confirms that we read from remote twice, whereas we wouldn't if this was not the validating cache.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
Jolyon-S: Yes

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
Jolyon-S: (it's a bit indirect, but this would fail if this is not a validating cache).

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCacheTest.java
Jolyon-S: That's actually already tested - `validationCausesAllCellsToBeReadRemotely` includes a cell that was not cached at the start too

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueCacheImplTest.java
gmaretic: Ah right

gmaretic (approved):

PR #5408, opened 2021-04-28
[LW] Rewire Transaction Caches

Jolyon-S:
**Goals (and why)**:
Hiding the digest details so that only the value cache cares about it makes it significantly easier to wire things through.

* Change interface to accept a set of start timestamps only.
* Add a concurrent hash map of start timestamp to transaction cache
* Hand out the locally stored cache when a transaction wants it
* On commit, just looks up the right cache in the map internally.

Current tests should catch any deviations in behaviour, plus added a test.

This presumably makes the wiring much easier, right?

`LockWatchValueCacheImpl`

Today/tomorrow to unblock work.


gmaretic (approved):
Seems reasonable :)

PR #5409, opened 2021-04-28
[LW] Pair the caches together

Jolyon-S:
**Goals (and why)**:
Pair the event and value caches together to make wiring them significantly easier.

* Define an interface in lock-api for LWVC and rename the old one to LWVSC.
* Define `LockWatchCache` with associated implementation that delegates to both caches.

Don't think this necessarily needs testing, but happy to be wrong here.

* Concurrency should be handled by the delegates, although could that open us up to race conditions? I don't think it's a massive issue (the internals of each cache are designed to be resilient / designed such that arbitrary events can happen to the other between updates).
* Resilient caches: should be fine for LWEC, not yet implemented for LWVC.

`LockWatchCache`.

ASAP to unblock work.


gmaretic (approved):
Is nice

PR #5412, opened 2021-04-29
Cassandra calls use tracing tags.

jkozlowski:
@carterkozak for SA

**Goals (and why)**:

Dynamic and long operation names are problematic for indexing infra.

I copied the FunctionalTagTranslator approach we had somewhere else, and extended it with a small interface that has some useful tags (like adding tags for TableReference, sizes of collections).

* The old codepath allocates an Object[] for the vararg regardless of whether tracing is enabled; new codepath mostly allocates lambdas, which should be rather compact, so should be an overall win.


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlQuery.java
```diff
@@ -51,20 +51,13 @@ public void logSlowResult(KvsProfilingLogger.LoggingFunction log, Stopwatch time
...
-
-                return safeQueryFormat() + ": " + argsString;
-            }
-        };
+    public String getSafeLog() {
+        String argsString = args().stream()
+                .filter(Arg::isSafeForLogging)
+                .map(arg -> String.format("%s = %s", arg.getName(), arg.getValue()))
+                .collect(Collectors.joining(", "));
+        return safeQueryFormat() + ": " + argsString;
```
carterkozak, line 61: Nit: if this is a relatively hot path, I'd avoid String.format at all costs, and try to avoid streams. Something along these lines should substantially reduce allocations:

```suggestion
        StringBuilder buffer = new StringBuilder()
            .append(safeQueryFormat())
            .append(": ");
        boolean first = true;
        for (Arg<?> arg : args()) {
            if (arg.isSafeForLogging()) {
                if (!first) buffer.append(", ");
                first = false;
                buffer.append(arg.getName()).append(" = ").append(arg.getValue());
            }
        }
        return buffer.toString();
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
```diff
@@ -180,13 +186,19 @@ public CASResult cas(
...
 
-    private static CloseableTrace startLocalTrace(CharSequence operationFormat, Object... formatArguments) {
-        return CloseableTrace.startLocalTrace(SERVICE_NAME, operationFormat, formatArguments);
+    private static CloseableTracer startLocalTrace(
+            @CompileTimeConstant String operation, Consumer<TagConsumer> tagTranslator) {
+        return CloseableTracer.startSpan(getOperation(operation), FunctionalTagTranslator.INSTANCE, tagTranslator);
+    }
+
+    private static String getOperation(@CompileTimeConstant String operation) {
+        return SERVICE_NAME + "." + operation;
```
carterkozak, line 202: If the inputs are compile-time-constants, should we inline the service name into those constants rather than re-concatenating each call? This runs even for unsampled traces.

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
```diff
@@ -429,41 +442,107 @@ public boolean shouldTriggerCompactions() {
...
+        public <T> void translate(TagAdapter<T> _adapter, T _target, Object _data) {}
+
+        @Override
+        public boolean isEmpty(Object _data) {
+            return true;
+        }
+    }
+
+    private static String getOperation(@CompileTimeConstant String operation) {
+        return SERVICE_NAME + "." + operation;
```
carterkozak, line 546: ideally here too

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
```diff
@@ -0,0 +1,70 @@
...
+        default void tableRef(TableReference tableReference) {
+            accept("table", LoggingArgs.safeTableOrPlaceholder(tableReference).toString());
+        }
+
+        default void timestamp(long ts) {
+            accept("ts", Long.toString(ts));
+        }
+
+        default void size(@CompileTimeConstant String name, Iterable<?> iterable) {
+            integer(name, Iterables.size(iterable));
```
carterkozak, line 42: Is this used directly? Not sure I'd want to spin through a potentially large iterable to collect this data for tracing, would most callers be able to eagerly produce a collection instead?

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
jkozlowski: I'd hope the compiler is smart enough to inline these into constants?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CqlQuery.java
jkozlowski: I don't think we use Cql in hot code, mostly for creating tables and startup things. @jeremyk-91 ?

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
jkozlowski: This one is weird, the API to the things that call this ask for an iterable, but generally people will put in a Collection, and Iterables.size does the right thing (instanceof Collection and call .size directly).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
jkozlowski: I am happy to do that, if the compiler is not smart enough. Lemme push a commit

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
jkozlowski: Done

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
jkozlowski: Done

carterkozak (approved):

atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
```diff
@@ -429,41 +443,90 @@ public boolean shouldTriggerCompactions() {
...
+        DetachedSpan detachedSpan = DetachedSpan.start("atlasdb-kvs.getAsync");
         ListenableFuture<Map<Cell, Value>> future = delegate().getAsync(tableRef, timestampByCell);
-        return attachDetachedSpanCompletion(detachedSpan, future, tracingExecutorService);
+        return attachDetachedSpanCompletion(detachedSpan, future, tracingExecutorService, sink -> {
+            sink.tableRef(tableRef);
+            sink.size("cells", timestampByCell);
+        });
+    }
+
+    @MustBeClosed
```
jeremyk-91, line 464: nice: wasn't aware this was a thing

jeremyk-91 (approved):
Great, this is much cleaner than the original. Thanks for doing this!

I'd probably take @carterkozak's suggestion for the CqlQuery (we use it on every transaction commit), but other than that looks good

anjor (approved):

sudiksha27 (approved):
+1

PR #5413, opened 2021-04-29
[LW] Fix bugs and be protective, I

Jolyon-S:
**Goals (and why)**:
* There were a couple of bugs
* Implement `getAsync`
* Be a bit more protective around usage of the transaction cache.

* Lots of things!

Updated tests, however, for the most part, the fact that the tests continue to pass is good.

I haven't broken anything, have I?

LWVSCI probably a good place to start.

Whenever


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -62,12 +62,17 @@ public LockWatchValueScopingCacheImpl(
     public synchronized void processStartTransactions(Set<Long> startTimestamps) {
         TransactionsLockWatchUpdate updateForTransactions =
                 eventCache.getUpdateForTransactions(startTimestamps, currentVersion);
+
+        if (updateForTransactions.clearCache()) {
+            clearCache();
+        }
```
Jolyon-S, line 68: Oops! I think I was technically doing this by calculating it myself, except that I was doing it _after_ updating the stores, which means that I was making it worse for myself.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -177,37 +183,20 @@ private synchronized void updateCurrentVersion(TransactionsLockWatchUpdate updat
...
-                currentVersion = maybeUpdateVersion;
-            } else {
-                LockWatchVersion current = currentVersion.get();
-                if (leaderElection(current, updateVersion)) {
-                    clearCache();
-                    currentVersion = maybeUpdateVersion;
-                } else if (newUpdate(current, updateVersion)) {
-                    currentVersion = maybeUpdateVersion;
-                }
-            }
```
Jolyon-S, line 191: We don't need this logic, since we just clear the version on `clearCache` being true from the update anyway.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/NoOpTransactionScopedCache.java
```diff
@@ -43,11 +43,22 @@ public void delete(TableReference tableReference, Set<Cell> cells) {}
...
+    @Override
+    public ListenableFuture<Map<Cell, byte[]>> getAsync(
+            TableReference tableReference,
+            Set<Cell> cells,
+            BiFunction<TableReference, Set<Cell>, ListenableFuture<Map<Cell, byte[]>>> valueLoader) {
+        return valueLoader.apply(tableReference, cells);
+    }
+
+    @Override
+    public void close() {}
```
Jolyon-S, line 60: We _could_ make this throw like the other caches, but I don't think we need to (since it doesn't actually affect semantics if it does or not as nothing is being cached anyway.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
```diff
@@ -27,28 +28,28 @@
...
     private static final Logger log = LoggerFactory.getLogger(ValidatingTransactionScopedCache.class);
 
     private final TransactionScopedCache delegate;
-    private final Supplier<Boolean> shouldValidate;
+    private double validationProbability;
+    private final Random random;
 
     @VisibleForTesting
-    ValidatingTransactionScopedCache(TransactionScopedCache delegate, Supplier<Boolean> shouldValidate) {
+    ValidatingTransactionScopedCache(TransactionScopedCache delegate, double validationProbability) {
```
Jolyon-S, line 45: Since I made it possible to specify the probability instead of making it static here, no need for the janky supplier.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
```diff
@@ -64,20 +65,45 @@ public void delete(TableReference tableReference, Set<Cell> cells) {
...
             BiFunction<TableReference, Set<Cell>, ListenableFuture<Map<Cell, byte[]>>> valueLoader) {
-        if (shouldValidate.get()) {
-            Map<Cell, byte[]> remoteReads = AtlasFutures.getUnchecked(valueLoader.apply(tableReference, cell));
-            Map<Cell, byte[]> cacheReads =
-                    delegate.get(tableReference, cell, (table, cells) -> getCells(remoteReads, cells));
+        return AtlasFutures.getUnchecked(getAsync(tableReference, cells, valueLoader));
+    }
+
+    @Override
+    public ListenableFuture<Map<Cell, byte[]>> getAsync(
```
Jolyon-S, line 74: Exposing this actually turned out to be fairly gnarly. I'm pretty convinced this is correct, but a really tight sanity check or suggestions for further tests would be welcome!

gmaretic (commented):
Looks correct, but will need some rejigging when the wiring PR merges

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValidatingTransactionScopedCache.java
```diff
@@ -64,20 +65,41 @@ public void delete(TableReference tableReference, Set<Cell> cells) {
...
+            TableReference tableReference,
+            Set<Cell> cells,
+            BiFunction<TableReference, Set<Cell>, ListenableFuture<Map<Cell, byte[]>>> valueLoader) {
+        if (shouldValidate()) {
+            ListenableFuture<Map<Cell, byte[]>> remoteReads = valueLoader.apply(tableReference, cells);
+            ListenableFuture<Map<Cell, byte[]>> cacheReads = delegate.getAsync(
+                    tableReference,
+                    cells,
+                    (table, cellsToRead) -> Futures.transform(
```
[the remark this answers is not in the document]
Jolyon-S: Yes, that should be it - we get everything remotely, then what we don't get from the cache, we get from those reads (and indeed, filter out non-nulls). Finally, we validate the result when they both return.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -780,14 +780,18 @@ private void extractLocalWritesForRow(
                         tableRef,
                         cells,
                         (table, uncached) -> getInternal(
-                                "get", tableRef, cells, immediateKeyValueService, immediateTransactionService));
+                                "get", tableRef, uncached, immediateKeyValueService, immediateTransactionService));
```
Jolyon-S, line 783: This is bad! We don't need to get all cells remotely, only the uncached ones!

gmaretic: nice catch 🍡

gmaretic (approved):
LGTM

PR #5414, opened 2021-04-29
[PDS-180370] Targeted Sweep Migration Docs Should Comment On When To Change Transaction Types

jeremyk-91:
**Goals (and why)**:
- Avoid liveness issues seen in e.g. PDS-180730
- It is non-intuitive that switching to read-only transactions MUST currently happen in a distinct upgrade from the upgrade to THOROUGH_MIGRATION.

- Outline this

- Docs

- Is the reverse process correct?

### DIFF
--- a/docs/source/cluster_management/sweep/targeted-sweep.rst
+++ b/docs/source/cluster_management/sweep/targeted-sweep.rst
@@ -86,15 +86,26 @@ the immutable timestamp lock.
 CONSERVATIVE to THOROUGH
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. warning::
+
+   You should not change user transactions from read-only to read-write and the sweep strategy from conservative to
+   thorough in the same release. In this case, read-only transactions on old nodes may fail during any state where
+   both old and new nodes are simultaneously operating (though there are no correctness implications).
+
 Thus, a way of changing sweep strategy from CONSERVATIVE to THOROUGH while avoiding downtime is as follows:
-  1. Roll service nodes from a version with CONSERVATIVE sweep strategy to one with THOROUGH_MIGRATION.
-  2. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with THOROUGH.
+  1. Roll service nodes from a version which uses read-only transactions (e.g. ``runTaskReadOnly``) to one that uses
+     only read-write transactions (e.g. ``runTaskReadWrite``). Both versions should still use the CONSERVATIVE sweep
+     strategy.
+  2. Roll service nodes from a version that uses CONSERVATIVE sweep strategy AND only uses read-write transactions to
+     one with THOROUGH_MIGRATION.
+  3. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with THOROUGH.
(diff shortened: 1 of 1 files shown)


docs/source/cluster_management/sweep/targeted-sweep.rst
```diff
@@ -104,7 +115,8 @@ The process here needs to account for the existence of old entries written to th
      This is safe; see step 2 above.
   2. Wait until targeted sweep for strategy THOROUGH has caught up to after the upgrade. This can be verified by
      consulting the ``millisSinceLastSweptTs`` targeted sweep metric.
-  3. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with CONSERVATIVE.
+  3. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with CONSERVATIVE. If desired,
+     the CONSERVATIVE product version may immediately begin using read-only transactions.
```
Jolyon-S, line 119: This is fine because CONSERVATIVE is more permissive than THOROUGH(_MIGRATION) in the sense that it allows both types, whereas the others only allow one type.

docs/source/cluster_management/sweep/targeted-sweep.rst
```diff
@@ -86,15 +86,26 @@ the immutable timestamp lock.
...
+
 Thus, a way of changing sweep strategy from CONSERVATIVE to THOROUGH while avoiding downtime is as follows:
-  1. Roll service nodes from a version with CONSERVATIVE sweep strategy to one with THOROUGH_MIGRATION.
-  2. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with THOROUGH.
+  1. Roll service nodes from a version which uses read-only transactions (e.g. ``runTaskReadOnly``) to one that uses
+     only read-write transactions (e.g. ``runTaskReadWrite``). Both versions should still use the CONSERVATIVE sweep
+     strategy.
+  2. Roll service nodes from a version that uses CONSERVATIVE sweep strategy AND only uses read-write transactions to
+     one with THOROUGH_MIGRATION.
+  3. Roll service nodes from a version with THOROUGH_MIGRATION sweep strategy to one with THOROUGH.
```
Jolyon-S, line 101: I know that the wording implies that the service _must_ go through these versions, but I think it is worth being _really_ explicit about the fact that it must go through (e.g. via some form of schema version thing).

docs/source/cluster_management/sweep/targeted-sweep.rst
jeremyk-91: yep!

Jolyon-S (approved):
Looks good, just worth actioning the one comment.

docs/source/cluster_management/sweep/targeted-sweep.rst
```diff
@@ -86,15 +86,26 @@ the immutable timestamp lock.
...
      During this roll, the queue to which cell references are enqueued will vary depending on the individual node.
```
sudiksha27, line 108: You could mention schema bumps here as well?

jeremyk-91: discussed offline: this is how it could be used, but at this level we don't really have the concept of schema version I think

PR #5416, opened 2021-04-29
[WIP][LW] Resilient Cache for values

Jolyon-S:
**Goals (and why)**:
* We want to be able to fall back to a no-op cache for the value cache too.

* Modify the `ResilientLockWatchEventCache` to be a more generic proxy appropriate for both event and value cache.
* Since not all methods are synchronised on the value cache, do some magic to not always run the methods as synchronised.

WIP

Is this just worse than having two very similar proxies? Maybe a common proxy and two variants?

`ResilientLockWatchProxy`

This week or early next week.


Jolyon-S:
Closing in favour of https://github.com/palantir/atlasdb/pull/5418

PR #5417, opened 2021-04-30
[LW] Fix rare edge case

Jolyon-S:
**Goals (and why)**:
In the last month, we've had a single instance where a success event has been received with no events but a later version than the one known in LWEC. I don't know how, and my read of the code didn't yield how, but I think we should just retry in this case.

* Retry on success event edge case.

Added test to LWECIT.

I still don't know how we hit this, and I am a little concerned, but at least we throw instead of silently accepting it. Understanding how we hit this / reproing would be great, but the fact that it is so rare makes this probably quite hard.

diff

Sooner rather than later!


gmaretic (approved):
The famous
`// idk`

PR #5418, opened 2021-04-30
[LW] Resilient Proxy for the value cache

Jolyon-S:
**Goals (and why)**:
* We want to be able to fall back to a no-op cache for the value cache too.

* Modify the `ResilientLockWatchEventCache` to be a more generic proxy appropriate for both event and value cache.
* Since not all methods are synchronised on the value cache, remove the synchronised aspect of the proxy, and make LWEC thread safe.

Fix up the proxy tests. Note that there are _no_ concurrency tests for LWEC.

Have I missed a synchronised method? Are we exposing ourselves to race conditions here on the volatile read of the delegate?

`ResilientLockWatchProxy`

This week or early next week.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-package com.palantir.atlasdb.keyvalue.api.watch;
+package com.palantir.atlasdb.keyvalue.api;
 
 import com.codahale.metrics.Counter;
 import com.google.common.reflect.AbstractInvocationHandler;
-import com.google.errorprone.annotations.concurrent.GuardedBy;
+import com.palantir.atlasdb.keyvalue.api.cache.LockWatchValueScopingCache;
 import com.palantir.atlasdb.transaction.api.TransactionLockWatchFailedException;
 import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.lock.watch.LockWatchEventCache;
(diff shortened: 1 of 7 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -33,31 +36,48 @@
...
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
 @ThreadSafe
 public final class LockWatchValueScopingCacheImpl implements LockWatchValueScopingCache {
     private final LockWatchEventCache eventCache;
+    private final double validationProbability;
+    private final CacheStore cacheStore;
```
Jolyon-S, line 50: Not guarded by as the cache store declares thread safety itself (so that we can concurrently get transaction caches).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -33,31 +36,42 @@
...
     private final ValueStore valueStore;
     private final SnapshotStore snapshotStore;
-    private final CacheStore cacheStore;
-    private final double validationProbability;
```
Jolyon-S, line 48: I don't use `@GuardedBy` because it doesn't work in lambdas, even if they are in synchronised methods. Ugh.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
```diff
@@ -14,58 +14,66 @@
...
+    protected Object handleInvocation(Object proxy, Method method, Object[] args) throws IllegalAccessException {
         try {
             return method.invoke(delegate, args);
         } catch (InvocationTargetException e) {
             throw handleException(e);
         }
     }
 
-    synchronized RuntimeException handleException(InvocationTargetException rethrow) {
+    private synchronized RuntimeException handleException(InvocationTargetException rethrow) {
```
Jolyon-S, line 76: I don't think that this is an issue being synchronised, as we probably won't be throwing that many exceptions, and when we do, it should be fine. However, we could probably drop this, and at worst we'd try and fallback twice.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
```diff
@@ -14,58 +14,66 @@
...
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import javax.annotation.concurrent.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class ResilientLockWatchEventCache extends AbstractInvocationHandler {
+@ThreadSafe
+public final class ResilientLockWatchProxy<T> extends AbstractInvocationHandler {
```
Jolyon-S, line 34: The reason I didn't just wrap the merged cache with one of these is that it is conceivable that the value cache falls back (due to say, validation errors), but the event cache does not, still serving atlasdb-proxy.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
```diff
@@ -14,58 +14,66 @@
...
+    private volatile T delegate;
 
-    private ResilientLockWatchEventCache(
-            LockWatchEventCache defaultCache, LockWatchEventCache fallbackCache, MetricsManager metricsManager) {
+    private ResilientLockWatchProxy(T defaultCache, T fallbackCache, MetricsManager metricsManager) {
         this.delegate = defaultCache;
         this.fallbackCache = fallbackCache;
         this.fallbackCacheSelectedCounter =
-                metricsManager.registerOrGetCounter(ResilientLockWatchEventCache.class, "fallbackCacheSelectedCounter");
+                metricsManager.registerOrGetCounter(ResilientLockWatchProxy.class, "fallbackCacheSelectedCounter");
```
gmaretic, line 64: Do we want separate metrics for event and value?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/ResilientLockWatchProxy.java
gmaretic: It seems fine to keep it synchronized

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/watch/ResilientLockWatchProxyTest.java
```diff
@@ -18,58 +18,90 @@
...
+        proxyCache.updateCacheOnCommit(timestamps);
+        verify(defaultCache).updateCacheOnCommit(timestamps);
+        verify(fallbackCache, never()).updateCacheOnCommit(any());
+
+        // Failure
+        when(defaultCache.createTransactionScopedCache(timestamp))
+                .thenThrow(new TransactionFailedNonRetriableException(""));
+        assertThatThrownBy(() -> proxyCache.createTransactionScopedCache(timestamp))
+                .isExactlyInstanceOf(TransactionLockWatchFailedException.class);
+        verify(defaultCache).createTransactionScopedCache(timestamp);
```
gmaretic, line 81: I guess this test is slightly redundant, but no reason not to add it

gmaretic (approved):
A metrics nit, but otherwise seems good

PR #5421, opened 2021-05-04
Transaction onSuccess(Runnable) / Callback Registration

jeremyk-91:
**Goals (and why)**:
- Support `onSuccess(...)` which is needed for an internal workflow

- Store a list of registered callbacks that is thread safe
- Run these callbacks when the transaction successfully commits.

- Tests that cover most cases in `SnapshotTransactionTest`

- Does this accidentally break anything?
- Is my choice of list bad? It needs to be threadsafe.

**Where should we start reviewing?**: Transaction.java

**Priority (whenever / two weeks / yesterday)**: this week; on critical path

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -2494,7 +2494,6 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
 
     @Override
     public void onSuccess(Runnable callback) {
-        ensureUncommitted();
         Preconditions.checkNotNull(callback, "Callback cannot be null");
         successCallbackManager.registerCallback(callback);
     }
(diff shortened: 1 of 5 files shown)


atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1867,6 +1871,101 @@ public void getSortedColumnsDistributesLoadWithMinBound() {
...
+                    });
+                    return null;
+                }))
+                .isInstanceOf(RuntimeException.class)
+                .hasMessageContaining("boom");
+        txManager.runTaskReadOnly(txn -> {
+            assertThat(txn.get(TABLE, ImmutableSet.of(TEST_CELL)))
+                    .containsExactly(Maps.immutableEntry(TEST_CELL, PtBytes.toBytes("tom")));
+            return null;
+        });
```
sudiksha27, line 1966: Neat!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2571,4 +2590,21 @@ private Counter getCounter(String name, TableReference tableRef) {
...
+        }
+
+        public void runCallbacks() {
+            Preconditions.checkState(
+                    isDefinitivelyCommitted(),
+                    "Callbacks must not be run if it is not known that the transaction has definitively committed! "
+                            + "This is likely a bug in AtlasDB transaction code.");
+            callbacks.forEach(Runnable::run);
+        }
+    }
```
sudiksha27, line 2609: Neat design

sudiksha27 (approved):
Looks pretty good!

PR #5422, opened 2021-05-04
[LW] Only push to central cache at commit time

Jolyon-S:
**Goals (and why)**:
We must only update the central cache once a commit has completed successfully - it is unsafe to do so earlier (particularly since, as written, it will do so before even checking the pre commit condition!)

* Unbatch the update on commit endpoint, and change where it is called - this requires some rewiring in the parent interfaces of LWC and LWM too.

Fixed tests, but didn't add any new ones. This is a bug in the integration, mostly, so I think we need ETE tests to properly catch this.

The above change means that _all_ transactions must go through a single, synchronised method (in a non-batched manner); is this going to be a problem? 

Let's say that we have N concurrent commits; these are autobatched when retrieving the commit timestamp, and then in that same batched consumer, they have to do some work on the event cache in a synchronised way - so all N have to wait for `N*e` (where e is the time taken to do whatever work is necessary in the event cache). Adding another synchronised method there versus at the end in a non-batched way is probably roughly equivalent - at worst, a transaction waits `N*v`, but at best it doesn't have to wait (since it isn't batched). Therefore I'd argue it is fine, but I just want to confirm, since we went through the effort of batching the commit timestamp endpoint for LW in the first place.

LWVSCI probably.

This week - this is a bug, but no one is using Atlas LW caching yet.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -120,10 +112,17 @@ public TransactionScopedCache createTransactionScopedCache(long startTs) {
     }
 
     private synchronized void processCommitUpdate(long startTimestamp) {
-        CommitUpdate commitUpdate = eventCache.getCommitUpdate(startTimestamp);
         Optional<TransactionScopedCache> cache = cacheStore.getCache(StartTimestamp.of(startTimestamp));
+        cache.ifPresent(TransactionScopedCache::close);
```
Jolyon-S, line 116: I moved close back here. Reason being that I think we want to ensure the cache is closed before processing commit updates, as that means that the values will not change from this point onwards.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/TransactionCacheValueStore.java
```diff
@@ -43,6 +43,4 @@
      * unable to be cached due to values being locked). Also note that writes do not appear in the digest.
      */
     Map<CellReference, CacheValue> getValueDigest();
-
-    Set<CellReference> getHitDigest();
```
Jolyon-S, line 47: Removed all this because we don't need it.

timelock-server/src/integTest/resources/paxosSingleServer.ftl
```diff
@@ -30,7 +30,7 @@ server:
   minThreads: 1
   maxThreads: 200
   applicationConnectors:
-    - type: h2
+    - type: https
```
Jolyon-S, line 33: apparently this makes it work locally, idk if we want to necessarily push to develop. @sudiksha27

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
gmaretic: Then I would argue it shouldn't be called `close()`? Maybe call it `freeze()` or something? The design of close then call methods is suspicious

gmaretic (commented):
Discussed offline:
This does not seem necessary -- we can update the central cache with READS as soon as we get the commit ts as we will also get all the lock watches with it and can figure out if it is safe to do so, regardless of the transaction actually committing

PR #5423, opened 2021-05-04
Upgrade GBJ

jeremyk-91:
**Goals (and why)**:
- upgrade gradle baseline java

- bump the version...
- and register an ObjectMapper on the server side that is consistent with internal framework.

- ~BUT disable the `-parameters` compile flag. After this bump it seems that there are consistently Jackson failures of the form~

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid definition for property 'jsonValue' (of type `com.palantir.paxos.ImmutableClient`): Could not find creator property with name 'jsonValue' (known Creator properties: [original, value])
```

Most likely this is some strange interaction between Immutables, the @JsonValue  flag and this argument…

existing stuff will do

It's kind of terrifying that our Jackson config is dependent on compiler flags. But at least it fails and we can't compile as opposed to silent breaks

**Where should we start reviewing?**: build.gradle

**Priority (whenever / two weeks / yesterday)**: soon please

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         classpath 'com.netflix.nebula:gradle-info-plugin:9.3.0'
         classpath 'com.netflix.nebula:nebula-publishing-plugin:17.3.2'
         classpath 'com.palantir.gradle.externalpublish:gradle-external-publish-plugin:1.2.1'
-        classpath 'com.palantir.baseline:gradle-baseline-java:3.75.0'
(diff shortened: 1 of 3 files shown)


Jolyon-S (approved):
Looks fine. Don't think I didn't notice that branch name though!

jeremyk-91:
@Jolyon-S note that from an internal thread @carterkozak and @pkoenig10 had a better solution, which I've switched to.

carterkozak (approved):

PR #5426, opened 2021-05-06
[LW] Fix some small things

Jolyon-S:
**Goals (and why)**:
Fixing a couple of things.

Rename close to finalise and tweak semantics.

Fix tests, add a couple more.

Useful?

TSCI.

This week.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
@@ -29,7 +29,6 @@ import com.palantir.atlasdb.transaction.api.TransactionLockWatchFailedException;
 import com.palantir.atlasdb.util.MetricsManager;
 import com.palantir.common.streams.KeyedStream;
 import com.palantir.lock.LockDescriptor;
-import com.palantir.lock.watch.CommitUpdate;
 import com.palantir.lock.watch.CommitUpdate.Visitor;
 import com.palantir.lock.watch.LockWatchEvent;
 import com.palantir.lock.watch.LockWatchEventCache;
@@ -37,6 +36,7 @@ import com.palantir.lock.watch.LockWatchVersion;
 import com.palantir.lock.watch.TransactionsLockWatchUpdate;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -120,10 +120,18 @@ public TransactionScopedCache createTransactionScopedCache(long startTs) {
...
+        cache.ifPresent(TransactionScopedCache::finalise);
 
-        commitUpdate.accept(new Visitor<Void>() {
+        Map<CellReference, CacheValue> cachedValues = cache.map(TransactionScopedCache::getValueDigest)
+                .map(ValueDigest::loadedValues)
+                .orElseGet(ImmutableMap::of);
+
+        if (cachedValues.isEmpty()) {
+            return;
+        }
```
Jolyon-S, line 132: This shortcut is important, as it allows us to avoid throwing in case there has been a leader election and we are not a serializable transaction (because we can then just not push our values centrally, instead of needing to retry).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchManagerImpl.java
```diff
@@ -78,7 +78,8 @@ public static LockWatchManagerInternal create(
             NamespacedConjureLockWatchingService lockWatchingService) {
         LockWatchEventCache eventCache = LockWatchEventCacheImpl.create(metricsManager);
         // todo(gmaretic): params should come from config
-        LockWatchValueScopingCache valueCache = new LockWatchValueScopingCacheImpl(eventCache, 100000, 1.0);
+        LockWatchValueScopingCache valueCache =
+                LockWatchValueScopingCacheImpl.create(eventCache, metricsManager, 100000, 1.0);
```
Jolyon-S, line 82: semantic merge conflict? idk how this went through in develop.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
gmaretic: Where do we throw otherwise, `getCommitUpdate()`? But either way, having a shortcut makes sense

gmaretic (approved):
👍

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
Jolyon-S: Yes

PR #5427, opened 2021-05-06
[LW] Ensure Serializable Transactions use the cache correctly

Jolyon-S:
**Goals (and why)**:
Serializable transactions create a special snapshot transaction at commit time. This transaction does not go through the transaction starter, and basically just makes up a commit timestamp (1 + the actual transaction's commit timestamp). This means that it does not go through the various layers that are typically required for the lock watch value and event cache.

The aim here is for serializable transactions to be able to validate R-W conflicts correctly while still reading as much as possible from the cache, and making the least  invasive changes to do so.

* Rewire the constructor of snapshot transaction to take the supplier for a transaction cache directly. This isn't great as all-but-one of these is the same (or effectively so). An alternative would be to have a method on snapshot transaction which is `getCache()`, and we override that in the one-off serializable transaction; I don't mind either way.
* Expose a new method on the `LockWatchValueScopingCache` (and therefore all the random interfaces/classes that use it) that allows us to get a throwaway cache to use. We need to get a cache that: 1. is not persisted internally (as the throwaway transaction won't commit), and 2. is the same type as the original one (as we _probably_ don't want to read cached values if the current transaction has a no-op cache and has been reading remotely). This second point may not be necessary - the point of the cache is that it should be equivalent to the remote, so we might actually get away without needing it.
* Pass in this throwaway cache to the one-off transaction.

WIP - probably need some tests wherever serializable transaction is tested, along with ETE / integration tests when we get there (we _could_ block this PR on that - at the moment, we'll always get no-op caches which means we just read remotely 100% of the time for this transaction at the moment, which is bad for perf, but not a correctness issue or memory leak, and we have 100% validation now anyway).

* Correctness - we're not associating this commit timestamp with a given sequence. However, I'd argue that the cache is or should always be an accurate representation of the KVS at any given time, and thus we don't need a sequence as we'll never actually push back to it.
* Design - do we need to worry about type? I have initially but I'm starting to think we're fine without worrying.

`SerializableTransaction` is a good start.

Before release, certainly.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -859,6 +860,7 @@ private Transaction getReadOnlyTransaction(final long commitTs) {
                 keyValueService,
                 timelockService,
                 lockWatchManager,
+                () -> lockWatchManager.createThrowawayCache(cache.get()),
```
gmaretic, line 863: So this is kinda janky, but looks like it works: we get another cache at start ts, but it will be created from the updated value store that has processed the commit update

gmaretic (commented):
I mean surely we can just have a method on `TransactionScopedCacheImpl` to give a read only version of itself or something similar? The only reason why we can't just use the existing one is that `finalize()` prevents reads

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
Jolyon-S: This cache isn't really at _any_ timestamp - it is just the latest from the value store (the only thing that passing in the cache does is tell us which type it is).

Jolyon-S:
> I mean surely we can just have a method on `TransactionScopedCacheImpl` to give a read only version of itself or something similar? The only reason why we can't just use the existing one is that `finalize()` prevents reads

That, but also we'd need to ensure that the read-only cache takes into account the commit update (the whole point of this stage!).

PR #5430, opened 2021-05-06
[WIP] Commit version of cache for serializable transactions

Jolyon-S:
**Goals (and why)**:
WIP

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/FilteringValueCacheSnapshot.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/FilteringValueCacheSnapshot.java
@@ -1,0 +1,101 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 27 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/FilteringValueCacheSnapshot.java
```diff
@@ -0,0 +1,101 @@
...
+import com.palantir.lock.LockDescriptor;
+import com.palantir.lock.watch.CommitUpdate;
+import com.palantir.lock.watch.CommitUpdate.Visitor;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.immutables.value.Value;
+
+final class FilteringValueCacheSnapshot implements ValueCacheSnapshot {
```
Jolyon-S, line 31: This is annoyingly necessary due to the presence of invalidate all.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/FilteringValueCacheSnapshot.java
```diff
@@ -0,0 +1,101 @@
...
+                return LockedCells.invalidateSome(invalidatedLocks);
+            }
+        });
+    }
+
+    @Value.Immutable
+    interface LockedCells {
+        boolean allLocked();
+
+        Set<CellReference> lockedCells();
```
Jolyon-S, line 81: does `Set.universe` exist? Would simplify the logic here nicely.

PR #5432, opened 2021-05-07
[WIP] [LW] getOrCreateCache

Jolyon-S:
**Goals (and why)**:
I need to modify the cache for the one-off snapshot transaction in `SerializableTransaction`. The first step will be to make the create cache method first check whether it exists or not, and get if it does, create if it does not.

* Refactor the create cache method to first check if the cache is present
* Refactor the cache store to always provide a cache, instead of forcing the caller to handle optionals.

Fixed old tests, added new.

Standard things - is this sensible, etc. etc.

LWVSCI.

Soon to unblock further work.


gmaretic (approved):
👍

PR #5433, opened 2021-05-07
[LW] Serializable Transactions Part II: updateCacheOnCommit

Jolyon-S:
**Goals (and why)**:
It is important that we do _not_ remove the state from the cache at commit time, as we'll need to retrieve information just after the commit timestamp update, but before we enter the `finally` block.

* Removed the state removal at commit timestamp time.
* Renamed usages to reflect

Added tests to confirm that the state still exists after calling update cache on commit.

We didn't solely rely on this for cleanup, so this should be fine - we still call `removeTimestampState` in the `finally` block.

LWVSCI.

Again, sooner rather than later to unblock subsequent work.

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCache.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCache.java
@@ -36,13 +36,13 @@ public interface LockWatchValueScopingCache extends LockWatchValueCache {
     @Override
     void processStartTransactions(Set<Long> startTimestamps);
 
-    @Override
-    void updateCacheAndRemoveTransactionState(Set<Long> startTimestamps);
-
     /**
-     * This method should only be used when the transaction has failed and the state needs to be removed from the
-     * cache without updating the central value store.
+     * This does *not* remove state from the cache - {@link #removeTransactionState(long)} must be called at the end
+     * of the transaction to do so, or else there will be a memory leak.
      */
+    @Override
+    void updateCacheOnCommit(Set<Long> startTimestamps);
+
     @Override
     void removeTransactionState(long startTimestamp);
 
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStore.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStore.java
@@ -30,8 +30,8 @@ public interface SnapshotStore {
     /**
      * Stores a snapshot for a given sequence and set of timestamps. Calls with the same sequence but a different
      * snapshot will overwrite the snapshot - this means that multiple calls to {@link #getSnapshot(StartTimestamp)}
-     * may return different snapshots over time for the same start timestamp. However, these newer snapshots should
-     * just be more complete versions of past snapshots, and thus maintains correctness (the timestamp could
+     * may return different snapshots over time for the same start timestamp. However, these newer snapshots *should*
+     * just be more complete versions of past snapshots, and thus maintain correctness (the timestamp could
      * correctly receive any one of the snapshots, and the only factor that changes is how much is cached, not the
      * values themselves).
      */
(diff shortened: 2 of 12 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
```diff
@@ -99,20 +99,6 @@ public void valueCacheCreatesValidatingTransactionCaches() {
         assertThat(getRemotelyReadCells(scopedCache, TABLE, CELL_1)).containsExactlyInAnyOrder(CELL_1);
     }
 
-    @Test
-    public void updateCacheOnCommitFlushesValuesToLatestSequenceCache() {
```
Jolyon-S, line 103: The semantics have changed:
Before: if transactions A, B start at the same time, but A finishes before B even gets its transaction cache, then B will see the reads from A.
After: A and B start at the same time; if A finishes, then B will still have the same cache that A started with, _unless_ C starts after A finishes but with the same sequence, in which case B and C will have the same cache.

Effectively it is a very minor difference in when a snapshot for a fixed sequence is re-updated, and is only really relevant in cases where there are _very_ few updates (and hence the sequence isn't changing much at all). In these cases, the new semantics are marginally less efficient.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStore.java
```diff
@@ -27,15 +27,21 @@
  * and and removes snapshots when they are no longer referenced by any live start timestamps.
  */
 public interface SnapshotStore {
+    /**
+     * Stores a snapshot for a given sequence and set of timestamps. Calls with the same sequence but a different
+     * snapshot will overwrite the snapshot - this means that multiple calls to {@link #getSnapshot(StartTimestamp)}
+     * may return different snapshots over time for the same start timestamp. However, these newer snapshots should
+     * just be more complete versions of past snapshots, and thus maintains correctness (the timestamp could
+     * correctly receive any one of the snapshots, and the only factor that changes is how much is cached, not the
+     * values themselves).
```
Jolyon-S, line 36: It _is_ possible for me to write a check that every overwriting snapshot contains the same locked cells as the old one, but I think that this is probably more expensive than it is worth (O(N) in number of values cached, which is probably more trouble than it is worth).

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCache.java
```diff
@@ -36,13 +36,13 @@
     @Override
     void processStartTransactions(Set<Long> startTimestamps);
 
-    @Override
-    void updateCacheAndRemoveTransactionState(Set<Long> startTimestamps);
-
     /**
-     * This method should only be used when the transaction has failed and the state needs to be removed from the
-     * cache without updating the central value store.
+     * This does *not* remove state from the cache - {@link #removeTransactionState(long)} must be called at the end
```
gmaretic, line 40: Almost makes it sensible to make `TransactionScopedCache` closeable, with a hook to remove itself from the cache when the transaction commits/aborts. It's ugly, but does not rely on future users remembering to use this correctly

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStore.java
gmaretic: This is fine, maybe wording should reflect that the implementation should satisfy this rather than claiming that the interface magically does it :P

gmaretic: Or rather that the usage must be such to preserve correctness

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
gmaretic: Yeah, that makes sense, it's a corner case anyway and the difference is minor

gmaretic (approved):
Took 10 min to review, some minor nits

atlasdb-api/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCache.java
Jolyon-S: I'll leave that as a future refactor, for now, since I think that will complicate the next PRs. Maybe in the cleanup pass I'll do this.

PR #5434, opened 2021-05-07
Fix the Leaky Roof: Fail Fast on Post-Transaction Iterator Usage

jeremyk-91:
**Goals (and why)**:
- Prevent users from using Iterators, Streams, Visitables and Async operations after a transaction has ended.
- Ensure that AtlasDB devs do not accidentally allow users to do this in the future

- Add tests.
- In the cases we have not handled correctly, fix the transaction code.
- Add information to the javadoc indicating our decision.

- Tests added that exceptions are thrown if one tries to use the forbidden incantations.

- Did I miss any key tests?

**Where should we start reviewing?**: SnapshotTransactionTest

**Priority (whenever / two weeks / yesterday)**: 1 week

### DIFF
--- a/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
+++ b/atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
@@ -78,6 +78,8 @@ public interface Transaction {
      * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
      * input {@code rows}, even if there are rows where no columns match the predicate.
      *
+     * Batching visitables must be used strictly within the scope of the transaction.
+     *
      * @param tableRef table to load values from
      * @param rows unique rows to apply the column range selection to
      * @param columnRangeSelection range of columns and batch size to load for each of the rows provided
@@ -97,6 +99,8 @@ public interface Transaction {
      * If the {@link Iterable} does not have a stable ordering (i.e. iteration order can change across iterators
      * returned) then the returned iterator is sorted lexicographically with columns sorted on byte ordering, but
      * the ordering of rows is undefined.
+     *
+     * Iterators must be used strictly within the scope of the transaction.
      *
      * @param tableRef table to load values from
      * @param rows unique rows to apply the column range selection to
@@ -120,6 +124,8 @@ public interface Transaction {
      * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
      * input {@code rows}, even if there are rows where no columns match the predicate.
      *
+     * Iterators must be used strictly within the scope of the transaction.
+     *
      * @param tableRef table to load values from
      * @param rows unique rows to apply the column range selection to
      * @param columnRangeSelection range of columns and batch size to load for each of the rows provided
@@ -142,6 +148,8 @@ public interface Transaction {
      * returned) then the returned iterator is sorted lexicographically with columns sorted on byte ordering, but the
      * ordering of rows is undefined. In case of duplicate rows, the ordering is based on the first occurrence of
      * the row.
+     *
+     * Iterators must be used strictly within the scope of the transaction.
      *
      * @param tableRef table to load values from
      * @param rows unique rows to apply column range selection to
@@ -175,6 +183,8 @@ public interface Transaction {
     /**
      * Creates a visitable that scans the provided range.
      *
+     * Batching visitables must be used strictly within the scope of the transaction.
+     *
      * @param tableRef the table to scan
      * @param rangeRequest the range of rows and columns to scan
      * @return an array of <code>RowResult</code> objects representing the range
@@ -189,6 +199,8 @@ public interface Transaction {
      * the batch hint in each RangeRequest. If this isn't done then this method may do more work
      * than you need and will be slower than it needs to be.  If the batchHint isn't specified it
      * will default to 1 for the first page in each range.
+     *
+     * Batching visitables must be used strictly within the scope of the transaction.
      *
      * @deprecated Should use either {@link #getRanges(TableReference, Iterable, int, BiFunction)} or
      * {@link #getRangesLazy(TableReference, Iterable)} to ensure you are using an appropriate level
@@ -205,6 +217,8 @@ public interface Transaction {
      *
      * It is guaranteed that the range requests seen by the provided visitable processor are equal to the provided
      * iterable of range requests, though no guarantees are made on the order they are encountered in.
+     *
+     * Streams must be read within the scope of the transaction.
      */
     @Idempotent
     <T> Stream<T> getRanges(
@@ -233,6 +247,8 @@ public interface Transaction {
     /**
      * Returns visitibles that scan the provided ranges. This does no pre-fetching so visiting the resulting
      * visitibles will incur database reads on first access.
+     *
+     * Streams and visitables must be read within the scope of this transaction.
      */
     @Idempotent
     Stream<BatchingVisitable<RowResult<byte[]>>> getRangesLazy(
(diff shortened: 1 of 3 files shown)


atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
```diff
@@ -78,6 +78,8 @@
      * It is guaranteed that the {@link Map#keySet()} of the returned map has a corresponding element for each of the
      * input {@code rows}, even if there are rows where no columns match the predicate.
      *
+     * Batching visitables must be used strictly within the scope of the transaction.
```
Jolyon-S, line 81: This is copy-and-pasted throughout, but I wonder if you maybe need to be a bit more explicit with what you mean at least once (maybe at the top):
```
Batching visitables, iterables, streams, and futures _must_ be used strictly within the scope of the transaction. Concretely, this means that results must be retrieved before exiting the transaction - in the case of streams and iterators, this means that you must collect the results before exiting; in the case of futures, you must await all asynchronous calls before returning from the transaction
```
Or something along those lines. I know what you mean here, but I have a feeling that users may not.
```

File: atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java [hunk 2]
```diff
@@ -143,6 +149,8 @@
      * ordering of rows is undefined. In case of duplicate rows, the ordering is based on the first occurrence of
      * the row.
      *
+     * Iterators must be used strictly within the scope of the transaction.
```
Jolyon-S, line 152: Also, you haven't added a similar comment on `getAsync`!

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1891,6 +1898,108 @@ public void transactionStillCommittedEvenIfCallbackThrows() {
...
     }
 
+    @Test
+    public void cannotCompleteFutureAfterTransactionCommit() {
+        SettableFuture<Object> blocker = SettableFuture.create();
+        ListenableFuture<Map<Cell, byte[]>> getFuture = txManager.runTaskThrowOnConflict(txn -> {
+            txn.put(TABLE, ImmutableMap.of(TEST_CELL, PtBytes.toBytes("jeremy")));
+            return Futures.transformAsync(
+                    blocker,
+                    _unused -> txn.getAsync(TABLE, ImmutableSet.of(TEST_CELL)),
```
Jolyon-S, line 1908: This is testing something subtly different if I am not mistaken: this is testing that you cannot complete a transaction, then call a method on that transaction. I think you ideally want a test where you start `getAsync` well within the bounds of the transaction, but that you return that (unfinished) future and use that outside. I don't think this matters except in SerializableTransactions, and I appreciate that this is probably hard to test, but that's the behaviour here that was at fault before, I believe.

Jolyon-S (commented):
I like the solution, but I think you're missing some details in the javadoc, and there's a test that I'd like to see (in an ideal world, although it may be hard to write).

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: Makes sense - yeah I've added this at the top. What you've written is very clear, so I've used that :)

atlasdb-api/src/main/java/com/palantir/atlasdb/transaction/api/Transaction.java
jeremyk-91: good spot - added!

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91: You're right: yeah the getAsync only happens after the transaction has completed here. I added one that handles usage of a future that began within (and completed within) but is used outside.

There isn't really a great way of injecting this, though, without considerably rewiring the SnapshotTransaction :(

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -788,12 +788,18 @@ private void extractLocalWritesForRow(
...
+                cache.get()
+                        .getAsync(
+                                tableRef,
+                                cells,
+                                (table, uncached) -> getInternal(
+                                        "getAsync", tableRef, uncached, keyValueService, defaultTransactionService)),
+                input -> {
+                    ensureStillRunning();
+                    return input;
+                },
```
Jolyon-S, line 801: technically, this is the output / result, not the input

Jolyon-S: Also, let's extract a `scopeToTransaction` method for futures too, then it is consistent and a bit cleaner.

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1147,9 +1147,6 @@ public void getSortedColumnsOnlyReadCellsAreCheckedForConflicts() {
         t2.commit();
 
         assertThatCode(t1::commit).doesNotThrowAnyException();
-
-        List<Cell> cells = Streams.stream(sortedColumns).map(Map.Entry::getKey).collect(Collectors.toList());
-        sanityCheckOnSortedCells(rows, cells, cellsWrittenOriginally);
```
Jolyon-S, line 1152: Can we not just move this before the commit? But still, nice catch!

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1891,6 +1897,121 @@ public void transactionStillCommittedEvenIfCallbackThrows() {
         });
     }
 
+    @Test
+    public void cannotUseCompletedFutureAfterTransactionCommit() throws Exception {
+        ListenableFuture<Map<Cell, byte[]>> getFuture = txManager.runTaskThrowOnConflict(txn -> {
+            txn.put(TABLE, ImmutableMap.of(TEST_CELL, PtBytes.toBytes("jolyon")));
```
Jolyon-S, line 1903: this is the second time my name features in a test!

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
```diff
@@ -1891,6 +1897,121 @@ public void transactionStillCommittedEvenIfCallbackThrows() {
         });
     }
 
+    @Test
+    public void cannotUseCompletedFutureAfterTransactionCommit() throws Exception {
```
Jolyon-S, line 1901: [Sanity check]: Do we have tests for normal getAsync usage?

Jolyon-S (approved):
Looks good - glad we spotted this weird edge case. A couple of nits, then good to go.

atlasdb-tests-shared/src/test/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionTest.java
jeremyk-91: ... I found the first one!

```
        assertThatThrownBy(() -> schema.addTableDefinition(TEST_TABLE_NAME, tableDef))
                .as("Jolyon loves this stuff")
                .isInstanceOf(SafeIllegalArgumentException.class)
                .hasMessage("Caching can only be enabled with the SERIALIZABLE_CELL conflict handler.");
```

PR #5437, opened 2021-05-07
Enforce the baseline checkImplicitDependencies check

carterkozak:
**Goals (and why)**:

This should prevent atlas from breaking when it's consumed by products that use newer versions of shared dependencies.

Builds must be updated

This adds a test.

yesterday: This would have prevented several failures that made it through testing.


pkoenig10:
Probably also want to enable `checkUnusedDependencies`?

Jolyon-S (approved):
Looks sensible - although we should probably enable `checkUnusedDependencies` as @pkoenig10 notes to ensure that we aren't depending on more than we need.

commons-executors-api/build.gradle
```diff
@@ -5,3 +5,6 @@ libsDirName = file('build/artifacts')
 // Do not add dependencies to this project. This project is used by
 // api projects outside of atlasdb that try very hard to minimize
 // their dependency footprint.
+dependencies {
+    implementation 'com.boundary:high-scale-lib'
+}
```
jeremyk-91, line 10: given the comment above, are we sure this is ok?

jeremyk-91 (commented):
seems good, one question though

commons-executors-api/build.gradle
jeremyk-91: I guess it's fine as this was implicitly added anyway, but is there a way to avoid this?

pkoenig10:
Are we still planning to merge this PR?

PR #5438, opened 2021-05-07
[PDS-183571] Maximum Lock Tenuring, Part 1: Forced Release

jeremyk-91:
**Goals (and why)**:
- See PDS-183571.
- In general, if a user holds a commit lock for a long period of time (for whatever reason: intractable serializable check, badly written pre-commit condition, etc) other transactions that read the row or cell (depending on lock level) that that user wrote to will all fail.

- Introduce a maximum tenure for a lock on the client side, after which we will cease refreshing it.
- Allow a callback to be provided.
- Configure transactions to hold their locks for 5 minutes, and thereafter log and release them.

- Added unit tests for the new bits.
- SnapshotTransaction integration tests run with the expiry mechanism as well.

- Should expiration actively unlock the lock as opposed to merely stopping it from being refreshed?
- Are there any *correctness* properties this might accidentally break?
- Is there too much perf overhead?

**Where should we start reviewing?**: ClientLockingOptions maybe

**Priority (whenever / two weeks / yesterday)**: next week

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.palantir.atlasdb.AtlasDbConstants;
 import com.palantir.atlasdb.transaction.impl.TransactionRetryStrategy;
+import com.palantir.conjure.java.api.config.service.HumanReadableDuration;
 import org.immutables.value.Value;
 
 @JsonDeserialize(as = ImmutableTransactionConfig.class)
@@ -49,6 +50,15 @@ public abstract class TransactionConfig {
     @Value.Default
     public boolean lockImmutableTsOnReadOnlyTransactions() {
         return false;
+    }
+
+    /**
+     * Indicates how long user transactions are allowed to take to commit, in terms of how long we'll refresh the
+     * commit locks for.
+     */
+    @Value.Default
+    public HumanReadableDuration commitLockTenure() {
+        return HumanReadableDuration.minutes(5);
     }
 
     /**
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
@@ -54,7 +54,8 @@ public abstract class TransactionConfig {
 
     /**
      * Indicates how long user transactions are allowed to take to commit, in terms of how long we'll refresh the
-     * commit locks for.
+     * commit locks for. Note that locks may still require more time before TimeLock realises that they are no longer
+     * actively held and releases them.
      */
     @Value.Default
     public HumanReadableDuration commitLockTenure() {
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
@@ -59,7 +59,7 @@ public abstract class TransactionConfig {
      */
     @Value.Default
     public HumanReadableDuration commitLockTenure() {
-        return HumanReadableDuration.minutes(5);
+        return HumanReadableDuration.minutes(15);
     }
 
     /**
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -2215,6 +2215,8 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                 lockDescriptors,
                 lockAcquireTimeoutMillis,
                 Optional.ofNullable(getStartTimestampAsClientDescription(currentTransactionConfig)));
+
+        RuntimeException stackTraceSnapshot = new RuntimeException("I exist to show you the stack trace");
         LockResponse lockResponse = timelockService.lock(
                 request,
                 ClientLockingOptions.builder()
@@ -2226,7 +2228,7 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
                                         + " other transactions to proceed.",
                                 SafeArg.of("commitLockTenure", currentTransactionConfig.commitLockTenure()),
                                 UnsafeArg.of("firstTenLockDescriptors", Iterables.limit(lockDescriptors, 10)),
-                                new RuntimeException("I exist to show you the stack trace")))
+                                stackTraceSnapshot))
                         .build());
         if (!lockResponse.wasSuccessful()) {
             log.error(
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -137,7 +137,6 @@ import com.palantir.tracing.CloseableTracer;
 import com.palantir.util.AssertUtils;
 import com.palantir.util.paging.TokenBackedBasicResultsPage;
 import java.nio.ByteBuffer;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -2219,11 +2218,13 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
         LockResponse lockResponse = timelockService.lock(
                 request,
                 ClientLockingOptions.builder()
-                        .maximumLockTenure(Duration.ofMinutes(5))
+                        .maximumLockTenure(
+                                currentTransactionConfig.commitLockTenure().toJavaDuration())
                         .tenureExpirationCallback(() -> log.warn(
-                                "This transaction held on to its commit locks for five minutes, which is"
+                                "This transaction held on to its commit locks for longer than its tenure, which is"
                                         + " suspicious. In the interest of liveness we will unlock this lock and allow"
                                         + " other transactions to proceed.",
+                                SafeArg.of("commitLockTenure", currentTransactionConfig.commitLockTenure()),
                                 UnsafeArg.of("firstTenLockDescriptors", Iterables.limit(lockDescriptors, 10)),
                                 new RuntimeException("I exist to show you the stack trace")))
                         .build());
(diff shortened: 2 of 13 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
```diff
@@ -51,6 +52,15 @@ public boolean lockImmutableTsOnReadOnlyTransactions() {
...
     }
 
+    /**
+     * Indicates how long user transactions are allowed to take to commit, in terms of how long we'll refresh the
+     * commit locks for.
+     */
+    @Value.Default
+    public HumanReadableDuration commitLockTenure() {
+        return HumanReadableDuration.minutes(5);
+    }
```
sudiksha27, line 62: The server can still hold the lock for 10 more minutes (maxAllowedLockTimeout) , right? Is so, maybe worth documenting that as well.
Actually I think releasing the locks sooner is also not a bad idea.

sudiksha27 (commented):
Broadly looks correct, one small suggestion.

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
```diff
@@ -36,14 +41,24 @@
...
+    private final Clock clock;
 
     private ScheduledFuture<?> task;
 
     public LockRefresher(
             ScheduledExecutorService executor, TimelockService timelockService, long refreshIntervalMillis) {
+        this(executor, timelockService, refreshIntervalMillis, Clock.systemUTC());
+    }
+
+    public LockRefresher(
```
gmaretic, line 54:
```suggestion
    @VisibleForTesting
    public LockRefresher(
```

lock-api-objects/src/main/java/com/palantir/lock/v2/ClientLockingOptions.java
```diff
@@ -0,0 +1,52 @@
...
+     * After {@link ClientLockingOptions#maximumLockTenure()} has passed since the client acquired the lock, we will
+     * not try to refresh the lock any longer. This is infinite by default.
+     */
+    Optional<Duration> maximumLockTenure();
+
+    /**
+     * Invoked if the lock tenure expired on the client. Must be cheap to execute.
+     */
+    @Value.Default
+    default Runnable tenureExpirationCallback() {
```
gmaretic, line 41: Yeah, I know you said must be cheap, but I can already see people putting the unlock here and then wondering why their other locks are expiring when they are serially unlocking who knows how many locks one by one.

Maybe add another parameter which lets you specify that it should be unlocked on timeout, so we can then batch all of the timed our locks in `LockRefresher` if that's what the user wants

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
```diff
@@ -79,12 +94,49 @@ private void refreshLocks() {
...
+            Instant deadline = candidate.getValue().lockRefreshDeadline();
+            if (now.isAfter(deadline)) {
+                log.info(
+                        "A lock token has expired on the client, because it has exceeded its tenure: we will stop"
+                                + " refreshing it automatically. Some time may still be required (20 seconds by"
+                                + " default) before the server releases the associated lock grants.",
+                        SafeArg.of("lockToken", candidate.getKey()),
+                        SafeArg.of("expiryDeadline", deadline),
+                        SafeArg.of("now", now));
+                candidate.getValue().clientExpiryCallback().run();
```
gmaretic, line 114: Yeah, I'm just worried people can mangle this by have unlock as the callback

lock-api/src/main/java/com/palantir/lock/client/RemoteTimelockServiceAdapter.java
```diff
@@ -110,6 +115,15 @@ public LockResponse lock(LockRequest request) {
         return lockLeaseService.lock(request);
     }
 
+    @Override
+    public LockResponse lock(LockRequest lockRequest, ClientLockingOptions options) {
+        log.warn(
```
gmaretic, line 120: nice!

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
```diff
@@ -40,6 +41,8 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.stream.Collectors;
 
+// TODO (jkong): This class almost shouldn't be a plain TimelockService, but the amount of effort that would be
```
gmaretic, line 44: Are you actually going to do this?

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
```diff
@@ -133,6 +136,15 @@ public LockResponse lock(LockRequest request) {
         return response;
     }
 
+    @Override
+    public LockResponse lock(LockRequest lockRequest, ClientLockingOptions options) {
```
gmaretic, line 140: nit: here and elsewhere where that is the correct behaviour feels like `lock(request)` should delegate to `lock(request, ClientLockingOptions.default())` or some such

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -2214,7 +2215,21 @@ protected LockToken acquireLocksForCommit() {
                 lockDescriptors,
                 lockAcquireTimeoutMillis,
                 Optional.ofNullable(getStartTimestampAsClientDescription(currentTransactionConfig)));
-        LockResponse lockResponse = timelockService.lock(request);
+
+        RuntimeException stackTraceSnapshot = new RuntimeException("I exist to show you the stack trace");
```
gmaretic, line 2219: Creating a RE every time for the possible timeout is pretty smelly, but I don't really see a great way around it, because debugging this without a stacktrace would probably be a nightmare

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
```diff
@@ -51,6 +52,15 @@ public boolean lockImmutableTsOnReadOnlyTransactions() {
         return false;
     }
 
+    /**
+     * Indicates how long user transactions are allowed to take to commit, in terms of how long we'll refresh the
+     * commit locks for.
+     */
+    @Value.Default
+    public HumanReadableDuration commitLockTenure() {
+        return HumanReadableDuration.minutes(5);
```
gmaretic, line 61: Are we sure this default is ok for all users of atlasdb?

lock-api/src/test/java/com/palantir/lock/client/LockRefresherTest.java
```diff
@@ -95,6 +107,51 @@ public void doesNotFailIfDelegateThrows() {
...
+
+        when(clock.instant()).thenReturn(Instant.EPOCH.plus(Duration.ofMinutes(3)));
+        tick();
+        verify(timelock, times(2)).refreshLockLeases(TOKENS);
+        verify(callback, never()).run();
+
+        when(clock.instant()).thenReturn(Instant.EPOCH.plus(Duration.ofMinutes(7)));
+        tick();
+        verify(timelock, times(2)).refreshLockLeases(TOKENS);
+        // Callback ran once per expired token
```
gmaretic, line 151: tick once more to make sure callbacks have run exactly once?

gmaretic (approved):
A bunch of nits, but broadly looks good

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
jeremyk-91: The 10 minutes is for the v1 lock service. These commit locks should be v2 ones, which I think by default expire in 20 seconds (LockLeaseContract). It's fair to note that more time may be needed - I've added a line to the comment suggesting that.

lock-api/src/main/java/com/palantir/lock/client/LockRefresher.java
jeremyk-91: good call!

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
jeremyk-91: I removed the TODO and left my lamentation about the bad design as a plain comment.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
jeremyk-91: Yeah. I don't know: we could deal with some `StackTraceElement[]` objects instead, but I figured having an RE as a container for the stack trace was easier to use.

(Also, random thought: I'm guessing we pronounce RE differently, since I pronounce it as having a vowel sound as in arr-ih and I noticed you have an `a` as the article)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
jeremyk-91: It is potentially a bit strict, but for the most part should be fine. What do you think would be good?

jeremyk-91: My main worry is if someone runs a huge serializable read-only transaction

lock-api/src/test/java/com/palantir/lock/client/LockRefresherTest.java
jeremyk-91: ... and looks like you caught a bug here! 🚀

lock-api/src/main/java/com/palantir/lock/client/TimeLockClient.java
jeremyk-91: Makes sense here, yes. For the delegating ones I'm a bit less certain that the default logic should live there as well

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: That's because I don't read it as an acronym, just write it that way due to general laziness

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/TransactionConfig.java
gmaretic: Exactly, I think 5 minutes should be plenty for 99% of the cases, but there may be outliers

lock-api-objects/src/main/java/com/palantir/lock/v2/ClientLockingOptions.java
jeremyk-91: Hmm, this isn't intended to be exposed to users of AtlasDB in general though. I added a bit more warning but really the only people who need to know are atlas devs I think.

gmaretic:
LGTM

PR #5439, opened 2021-05-10
[PDS-146088] Add even more logging

Jolyon-S:
**Goals (and why)**:
We don't have much visibility as to when timed execution is failing.

Add logs

N/A

Should we also add a metric? Might make it easier to debug.

`TimedRunner`

ASAP, so we can get this through the pipeline.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/TimedRunner.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/TimedRunner.java
@@ -17,6 +17,7 @@
 package com.palantir.util;
 
 import com.palantir.common.concurrent.PTExecutors;
+import com.palantir.logsafe.SafeArg;
 import java.time.Duration;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


atlasdb-commons/src/main/java/com/palantir/util/TimedRunner.java
```diff
@@ -49,15 +50,18 @@ public static TimedRunner create(Duration timeoutDuration) {
...
             failure = e;
         } catch (ExecutionException e) {
+            log.warn("Execution failed with exception", e);
             if (e.getCause() instanceof Error) {
                 throw (Error) e.getCause();
             } else {
                 failure = (Exception) e.getCause();
             }
         } catch (TimeoutException e) {
+            log.warn("Execution timed out", SafeArg.of("timeoutDuration", timeoutDuration));
```
jeremyk-91, line 64: Probably still want `e` here for the stacktrace?

jeremyk-91 (approved):

PR #5441, opened 2021-05-10
[LW] Integration Tests

Jolyon-S:
**Goals (and why)**:
Fix equality checks, and add integration tests.

* Integration tests
* Extract utility method for comparing `Map<Cell, byte[]`s
* fix cache value

This PR is primarily about testing

Fixes the issues?

Diff

Yesterday, because of the fixes 🔥


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -0,0 +1,280 @@
...
+                .isExactlyInstanceOf(TransactionSerializableConflictException.class)
+                .hasMessageContaining("There was a read-write conflict on table");
+    }
+
+    @Test
+    public void serializableTransactionsDoNotThrowWhenOverwritingAPreviouslyCachedValue() {
+        putValue();
+
+        assertThatCode(() -> txnManager.runTaskThrowOnConflict(txn -> {
+                    txn.get(TABLE_REF, ImmutableSet.of(CELL_1));
```
gmaretic, line 161: I guess the first read wasn't actually cached here, we need another transaction first to get it into the cache

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -0,0 +1,280 @@
...
+                    txn.put(TABLE_REF, ImmutableMap.of(CELL_1, DATA_3));
+                    return null;
+                }))
+                .doesNotThrowAnyException();
+    }
+
+    @Test
+    public void serializableTransactionsReadViaTheCacheWherePossible() {
```
gmaretic, line 169: Maybe make more obvious that you mean the check reads from the cache

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -0,0 +1,280 @@
...
+        putValue();
+
+        txnManager.runTaskThrowOnConflict(txn -> {
```
gmaretic, line 172: Maybe add a test that we actually invalidate a cached entry once we write to it for the subsequent transactions?

gmaretic (commented):
Tests seem pretty good, didn't check out the rest yet

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
Jolyon-S: Yeah, I suppose so. It'd be locally cached after the get, but definitely worth actually making it be centrally cached.

PR #5443, opened 2021-05-10
[LW] Serializable Transactions IV: Wiring

Jolyon-S:
**Goals (and why)**:
We have this fancy read-only cache, but we still aren't using it in the one-off snapshot transaction created by a serializable transaction.

* Change the way in which snapshot transactions get the cache (use an overridable method instead of a memoized supplier)
* Override this method in SerializableTransaction's one-off snapshot transaction
* Note that the cache _is_ effectively memoized because of the implementation - it is persisted inside LWM, so each call should just be a map lookup.

**None** - I think this is hard to test at the unit level. Will test at the integration level - I don't mind leaving this until we have integration tests before merging.

The usual AtlasDB things - correctness, performance.

`SerializableTransaction`

This week!

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
@@ -51,6 +51,7 @@ import com.palantir.atlasdb.keyvalue.api.RangeRequest;
 import com.palantir.atlasdb.keyvalue.api.RangeRequests;
 import com.palantir.atlasdb.keyvalue.api.RowResult;
 import com.palantir.atlasdb.keyvalue.api.TableReference;
+import com.palantir.atlasdb.keyvalue.api.cache.TransactionScopedCache;
 import com.palantir.atlasdb.keyvalue.api.watch.LockWatchManagerInternal;
 import com.palantir.atlasdb.keyvalue.impl.Cells;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -880,6 +881,11 @@ private Transaction getReadOnlyTransaction(final long commitTs) {
                 transactionConfig,
                 conflictTracer,
                 tableLevelMetricsController) {
+            @Override
+            protected TransactionScopedCache getCache() {
+                return lockWatchManager.getReadOnlyTransactionScopedCache(SerializableTransaction.this.getTimestamp());
```
gmaretic, line 886:
```suggestion
                return lockWatchManager.getReadOnlyTransactionScopedCache(this.getTimestamp());
```

gmaretic (approved):
(3 min) surprisingly non-awful

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
Jolyon-S: I'm actually not going to do this: you'll notice that this pattern is used in `getCommitTimestamps` too, and I think it makes it _really_ clear which method we mean to call.

PR #5447, opened 2021-05-11
[PDS-184114] Interrupts look sus

jeremyk-91:
**Goals (and why)**: PDS-184114

**Implementation Description (bullets)**: Don't propagate interrupts too much. We can get an interrupted exception when a blocking timeout runs out - in that case we want to cancel that request, but not cancel the entire server's operation.

Not much. I tried to add tests, but they don't fail in the old model (in spite of this change clearly improving things on internal cow deployment); most likely this is a result of different threading models in the internal server framework and dropwizard.

- does this fix it?

**Where should we start reviewing?**: ConjureRequestExceptionHandler.java


gmaretic (approved):
🔥

PR #5448, opened 2021-05-11
[LW] Leader Election Tests

Jolyon-S:
**Goals (and why)**:
Add testing around how the value cache handles leader elections.

Add a test with a leader election. Note that there probably should be more tests, but I cannot think of any interesting ones right now - once this is merged, it should be fairly easy to add more.

This PR is about tests.

Any useful tests to add?

`LockWatchValueIntegrationTest`

This week


gmaretic (approved):
Noice

PR #5449, opened 2021-05-12
[LW] Metrics

gmaretic:
**Goals (and why)**:
Add metrics for cache hits and misses

Super simple

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -99,4 +99,8 @@ public final class AtlasDbMetricNames {
 
     public static final String CACHE_HITS = "lockWatchCacheHits";
     public static final String CACHE_MISSES = "lockWatchCacheMisses";
+    public static final String CACHE_SIZE = "lockWatchCacheSize";
+    public static final String CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
+    public static final String EVENT_CACHE_VALIDATION_FAILURES = "eventCacheFallbackCount";
+    public static final String VALUE_CACHE_VALIDATION_FAILURES = "valueCacheFallbackCount";
 }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -97,10 +97,10 @@ public final class AtlasDbMetricNames {
     public static final String LEGACY_READ = "legacyRead";
     public static final String LEGACY_WRITE = "legacyWrite";
 
-    public static final String CACHE_HITS = "lockWatchCacheHits";
-    public static final String CACHE_MISSES = "lockWatchCacheMisses";
-    public static final String CACHE_SIZE = "lockWatchCacheSize";
-    public static final String CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
-    public static final String EVENT_CACHE_VALIDATION_FAILURES = "eventCacheFallbackCount";
-    public static final String VALUE_CACHE_VALIDATION_FAILURES = "valueCacheFallbackCount";
+    public static final String LW_CACHE_HITS = "lockWatchCacheHits";
+    public static final String LW_CACHE_MISSES = "lockWatchCacheMisses";
+    public static final String LW_CACHE_SIZE = "lockWatchCacheSize";
+    public static final String LW_CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
+    public static final String LW_EVENT_CACHE_FALLBACK_COUNT = "lockWatchEventCacheFallbackCount";
+    public static final String LW_VALUE_CACHE_FALLBACK_COUNT = "lockWatchValueCacheFallbackCount";
 }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -96,4 +96,7 @@ public final class AtlasDbMetricNames {
 
     public static final String LEGACY_READ = "legacyRead";
     public static final String LEGACY_WRITE = "legacyWrite";
+
+    public static final String CACHE_HITS = "lockWatchCacheHits";
+    public static final String CACHE_MISSES = "lockWatchCacheMisses";
 }
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
@@ -18,15 +18,30 @@ package com.palantir.atlasdb.keyvalue.api.cache;
 
 import com.palantir.atlasdb.AtlasDbMetricNames;
 import com.palantir.atlasdb.util.AccumulatingValueMetric;
+import com.palantir.atlasdb.util.CurrentValueMetric;
 import com.palantir.atlasdb.util.MetricsManager;
 
 public final class CacheMetrics {
     private final AccumulatingValueMetric hits;
     private final AccumulatingValueMetric misses;
+    private final AccumulatingValueMetric cacheSize;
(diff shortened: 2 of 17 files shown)


atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
```diff
@@ -98,19 +104,29 @@ public void tableNotWatchedInSchemaDoesNotCache() {
         scopedCache.finalise();
         assertThat(scopedCache.getHitDigest().hitCells()).isEmpty();
         assertThat(scopedCache.getValueDigest().loadedValues()).isEmpty();
+
+        verifyNoMoreInteractions(metrics);
```
Jolyon-S, line 108: `verifyNoInteractions`?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -75,7 +76,12 @@ public static LockWatchValueScopingCache create(
                 ResilientLockWatchProxy.newValueCacheProxyFactory(
                         NoOpLockWatchValueScopingCache.create(), metricsManager);
         LockWatchValueScopingCache defaultCache = new LockWatchValueScopingCacheImpl(
-                eventCache, maxCacheSize, validationProbability, watchedTablesFromSchema, proxyFactory::fallback);
+                eventCache,
+                maxCacheSize,
+                validationProbability,
+                watchedTablesFromSchema,
+                proxyFactory::fallback,
+                CacheMetrics.create(metricsManager));
```
Jolyon-S, line 84: We could probably add the RLWP metrics to `CacheMetrics`, in which case we could just pass in `CacheMetrics` to this level directly.

Jolyon-S (commented):
1. We should probably add metrics for failed validations (so that we can easily alert on them)
2. We have some other metrics internally - what do you think about adding metrics around the current size of the cache, as well as the % utilisation?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+                metricsManager);
+    }
+
+    public void registerHits(long number) {
+        hits.accumulateValue(number);
+    }
+
+    public void registerMisses(long number) {
+        misses.accumulateValue(number);
+    }
```
jeremyk-91, line 72: The `AccumulatingValueMetric` is backed by an `AtomicLong`. How much cross-thread interaction do we expect? At some point it could make sense for us to use `LongAdder` if this is high

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+    public void decreaseCacheSize(long removed) {
+        cacheSize.accumulateValue(-removed);
+    }
+
+    public void registerEventCacheValidationFailure() {
+        eventCacheValidationFailures.setValue(1);
+    }
+
+    public void registerValueCacheValidationFailure() {
+        valueCacheValidationFailures.setValue(1);
```
jeremyk-91, line 87: Interesting. For these two, why did we choose to use a null/1 value as opposed to an accumulating value metric?

Might want to rename this to `hasFailures` or something (to me at least the intuitive reading of a 1 value is "there was one failure", as opposed to "we're in a failed state").

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public void setMaximumCacheSize(long maximumCacheSize) {
+        metricsManager.registerOrGetGauge(
+                CacheMetrics.class,
+                AtlasDbMetricNames.CACHE_RATIO_USED,
+                () -> () -> cacheSize.getValue() / (double) maximumCacheSize);
```
jeremyk-91, line 94: sanity check: This method is only called once, right? If it's not then there may be some unexpected consequences (the `MetricSupplier` isn't invoked if the metric already exists).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
jeremyk-91: (I see this is a straight port of the previous system. Feel free to skip for this PR.)

atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
```diff
@@ -96,4 +96,11 @@ private CellFilterMetrics() {
 
     public static final String LEGACY_READ = "legacyRead";
     public static final String LEGACY_WRITE = "legacyWrite";
+
+    public static final String CACHE_HITS = "lockWatchCacheHits";
+    public static final String CACHE_MISSES = "lockWatchCacheMisses";
+    public static final String CACHE_SIZE = "lockWatchCacheSize";
+    public static final String CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
+    public static final String EVENT_CACHE_VALIDATION_FAILURES = "eventCacheFallbackCount";
+    public static final String VALUE_CACHE_VALIDATION_FAILURES = "valueCacheFallbackCount";
```
jeremyk-91, line 105: I think the constant name should match the metric name (which based on later usage is what this metric actually is).

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImplTest.java
```diff
@@ -103,28 +113,45 @@ public void watchEventUpdatesWatchableTables() {
...
+
+    @Test
+    public void metricsCorrectlyCountOverlappingPuts() {
+        valueStore = new ValueStoreImpl(ImmutableSet.of(TABLE), 300, metrics);
+        valueStore.applyEvent(WATCH_EVENT);
+
+        valueStore.putValue(CellReference.of(TABLE, CELL_2), VALUE_2);
+        valueStore.putValue(CellReference.of(TABLE, CELL_2), VALUE_2);
+        verify(metrics, times(2)).increaseCacheSize(anyLong());
+        verify(metrics).decreaseCacheSize(anyLong());
```
jeremyk-91, line 154: Throughout this test: would it make sense to assign values rather than using `anyLong()` throughout?

jeremyk-91 (approved):
🕐 0:30

👍  broadly looks good, DNM for comments

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
gmaretic: It's because as soon as you have a failure you fall back to no cache, so you will never have more

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
gmaretic: It should only be called once per TM, so we should be fine. I'm fine seeing if it breaks and fixing if we need to

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
gmaretic: Replaced with `Counter`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImplTest.java
gmaretic: Done, although it leaks implementation details a bit

PR #5460, opened 2021-05-19
[LW] Bug Hunter

Jolyon-S:
**Goals (and why)**:
* We have a resource leak! Best fix it.

* Transaction scoped caches are now only created at start transaction time, to bring them inline with all other state. This makes it easier for us to guarantee that it gets removed at the correct time.
* Read only transactions now actually only use the no-op cache.
* No op caches are not stored anywhere.

* Added test for read only transactions
* Added relevant tests for changed behaviour.

* Did I miss anything? I think the defensiveness should fix all cases that we didn't consider (because now caches are only made _exactly_ on that start transactions call, and thus all go through the right place.

yesterday 🔥

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
@@ -20,6 +20,7 @@ import com.palantir.atlasdb.keyvalue.api.watch.StartTimestamp;
 import com.palantir.atlasdb.transaction.api.TransactionFailedRetriableException;
 import com.palantir.lock.watch.CommitUpdate;
 import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -76,7 +77,13 @@ final class CacheStoreImpl implements CacheStore {
 
     @Override
     public void removeCache(StartTimestamp timestamp) {
-        cacheMap.remove(timestamp);
+        Optional<Caches> cache = Optional.ofNullable(cacheMap.remove(timestamp));
+
+        if (!cache.isPresent()) {
+            log.warn(
+                    "Attempted to remove cache state, but no cache was present for timestamp",
+                    SafeArg.of("timestamp", timestamp));
+        }
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
```diff
@@ -78,13 +76,7 @@ public TransactionScopedCache getCache(StartTimestamp timestamp) {
...
     @Override
     public void removeCache(StartTimestamp timestamp) {
-        Optional<Caches> cache = Optional.ofNullable(cacheMap.remove(timestamp));
-
-        if (!cache.isPresent()) {
-            log.warn(
-                    "Attempted to remove cache state, but no cache was present for timestamp",
-                    SafeArg.of("timestamp", timestamp));
-        }
+        cacheMap.remove(timestamp);
```
sudiksha27, line 79: Just to be 100% sure that this fixes the initial issue, can we keep this log line? Actually, if would great if we add another log line when we successfully remove from cache, keeping ts in params, so we know if the logs we are seeing on the dev stack are due to double removal attempts

Jolyon-S: We could do. I don't have strong feelings about keeping/removing this log line (at least for now), but adding the second logline should maybe go on an RC

Jolyon-S: re-added the old logs

sudiksha27 (commented):
While I get why this would fix the issues with read-only txns, I don't understand fully why it would fix the original issue, which wasn't for read-only txns?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
gmaretic: Could log removal at debug

sudiksha27: I don't think this log itself is very useful

gmaretic (approved):
Yup, this design kinda makes more sense anyway!

gmaretic:
> While I get why this would fix the issues with read-only txns, I don't understand fully why it would fix the original issue, which wasn't for read-only txns?

Do we know that?

sudiksha27:
Ah no you're right, we don't have stack traces for those, the stack traces we do have are for cache misses(these include readWrites). So it _may_ be possible that all those transactions were read-only.

Jolyon-S:
> Ah no you're right, we don't have stack traces for those, the stack traces we do have are for cache misses(these include readWrites). So it _may_ be possible that all those transactions were read-only.

AFAIK the actual errors were only for read only transactions. The cache misses will include RWs because the read only ones were never actually removing from the cache! Still, not sure what caused those misses.

PR #5461, opened 2021-05-19
[PDS-146088] Logging (yet again)

Jolyon-S:
**Goals (and why)**:
Log around some more possible edge cases in the object pool borrowing code.

Logging

Is there anything else that could be useful to log at the same time here?

Diff

Sooner is better - we'll miss Thursday this week but should get it in time for next Thursday

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
@@ -180,6 +180,8 @@ public class CassandraClientPoolingContainer implements PoolingContainer<Cassand
                 } else {
                     invalidateQuietly(resource);
                 }
+            } else {
+                log.warn("Failed to acquire Cassandra resource from object pool");
             }
         }
     }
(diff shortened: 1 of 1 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -217,6 +219,7 @@ private void invalidateQuietly(CassandraClient resource) {
             log.debug("Discarding resource of host {}", SafeArg.of("host", CassandraLogHelper.host(host)));
             clientPool.invalidateObject(resource);
         } catch (Exception e) {
+            log.warn("Attempted to invalidate a non-reusable Cassandra resource, but failed to due an exception", e);
```
Jolyon-S, line 222: This case looks like we just never invalidate the object - that seems like an issue to me.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -180,6 +180,8 @@ public int getPoolSize() {
                 } else {
                     invalidateQuietly(resource);
                 }
+            } else {
+                log.warn("Failed to acquire Cassandra resource from object pool");
```
Jolyon-S, line 184: Probably not a problematic case, but why not also log it?

jeremyk-91 (approved):

PR #5465, opened 2021-05-24
[DbTL 2021] Part 2: Oracle Container Integration (ETEs)

jeremyk-91:
**Goals (and why)**:
- Gather confidence in the Oracle part of DbKVS before Atlas releases are cut.

- Wire up the Oracle docker containers

- Add ETE test to bring end-to-end test parity between Postgres and Oracle, at least for the KVS part.

- Is it okay that I did a 1-node user system? I don't think adding more user nodes / using leader block really tests much meaningful.

**Where should we start reviewing?**: OracleDbKvsEteTestSuite.java

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           paths: [ . ]
 
   test:
-    parallelism: 10
+    parallelism: 11
     machine:
       image: circleci/classic:edge
     resource_class: xlarge
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
@@ -117,8 +117,7 @@ public abstract class EteSetup {
                 .machine(machine)
                 .saveLogsTo(LogDirectory.circleAwareLogDirectory(logDirectory))
                 .shutdownStrategy(ShutdownStrategy.AGGRESSIVE_WITH_NETWORK_CLEANUP)
-                .nativeServiceHealthCheckTimeout(
-                        org.joda.time.Duration.standardMinutes(5))
+                .nativeServiceHealthCheckTimeout(org.joda.time.Duration.standardMinutes(5))
                 .build();
 
         DockerProxyRule dockerProxyRule = DockerProxyRule.fromProjectName(docker.projectName(), eteClass);
--- a/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
+++ b/atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
@@ -46,6 +46,7 @@ import java.util.stream.Collectors;
 import org.awaitility.Awaitility;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.RuleChain;
+import org.slf4j.LoggerFactory;
 
 // Important: Some internal tests depend on this class.
 // Please recompile Oracle internal tests if any breaking changes are made to the setup.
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           paths: [ . ]
 
   test:
-    parallelism: 10
+    parallelism: 11
     machine:
```
sudiksha27, line 36: We are doing this because we re adding another container to run Oracle tests?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
```diff
@@ -116,6 +116,7 @@ public static RuleChain setup(
                 .machine(machine)
                 .saveLogsTo(LogDirectory.circleAwareLogDirectory(logDirectory))
                 .shutdownStrategy(ShutdownStrategy.AGGRESSIVE_WITH_NETWORK_CLEANUP)
+                .nativeServiceHealthCheckTimeout(org.joda.time.Duration.standardMinutes(5))
                 .build();
```
sudiksha27, line 120: Would it make sense to extract out the 5min duration to keep this and `CREATE_TRANSACTION_MANAGER_MAX_WAIT_TIME_SECS` in sync?

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/OracleDbKvsEteTestSuite.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.runners.Suite;
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+    TodoEteTest.class,
+    TimestampManagementEteTest.class,
+    CoordinationEteTest.class,
+    LockWithoutTimelockEteTest.class
+})
+public class OracleDbKvsEteTestSuite extends EteSetup {
```
sudiksha27, line 34: Why do we not run `LockWatchEteTest`? Only asking because we seem to run those on Postgres

sudiksha27 (commented):
The docker-compose is in sync with our internal Oracle test setup. The tests are taking a while to start up on my local but I think there is not much we can do there?
Have a few questions for better understanding but otherwise looks pretty good!

.circleci/config.yml
jeremyk-91: Yep, exactly - that's right. That container by itself is already the slowest one - some of this is because the Oracle image is *massive* I think :(

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/EteSetup.java
jeremyk-91: Yeah, that seems fair.

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/OracleDbKvsEteTestSuite.java
jeremyk-91: Actually, I remember why now: the LW one is used only where you have a timelock-server (which is required for LW). I was having difficulty getting Oracle-backed DB-timelock to run properly here (too many containers, perhaps)? so ended up not doing that. That Oracle-backed DB-timelock kind of works is tested internally, so chose not to replicate it here.

sudiksha27: I see, this is fine in that case

sudiksha27 (approved):
Looks good

PR #5466, opened 2021-05-24
Add traces for kvs-get

anjor:
**Goals (and why)**:
Improved visibility into kvs-get performance.

Added a few spans to CellLoader#loadWithTs.

A bit unsure about changes in TaskRunner

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
@@ -37,7 +37,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.Callable;
@@ -102,7 +101,7 @@ final class CellLoader {
                         "cells",
                         String.valueOf(cells.size()),
                         "tableRef",
-                        Objects.requireNonNull(LoggingArgs.tableRef(tableRef).getValue()),
+                        LoggingArgs.safeInternalTableNameOrPlaceholder(tableRef.toString()),
                         "timestampClause",
                         loadAllTs ? "for all timestamps " : "",
                         "startTs",
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
@@ -29,6 +29,7 @@ import com.palantir.atlasdb.util.AnnotatedCallable;
 import com.palantir.atlasdb.util.AnnotationType;
 import com.palantir.common.base.FunctionCheckedException;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.tracing.CloseableTracer;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
```diff
@@ -92,10 +93,13 @@ void loadWithTs(
             boolean loadAllTs,
             CassandraKeyValueServices.ThreadSafeResultVisitor visitor,
             ConsistencyLevel consistency) {
-        Map<InetSocketAddress, List<Cell>> hostsAndCells =
-                HostPartitioner.partitionByHost(clientPool, cells, Cell::getRowName);
+        Map<InetSocketAddress, List<Cell>> hostsAndCells;
+        try(CloseableTracer tracer = CloseableTracer.startSpan("partitionByHost")) {
```
jeremyk-91, line 97: nit: checkstyle will probably want a space after 

Also we should probably include a bit of metadata (the table ref and cells.size(), maybe)?

jeremyk-91 (commented):
Broadly looks good, we discussed the listening-executor changes.

I think you'll need changelogs and probably some formatting fixes (you can run `./gradlew spotlessApply` to get the latter in place, usually).

This should give us a good start: I'd say the next step would be to start considering spans _within_ the tasks, though fine with doing that separately/later depending on your capacity.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
```diff
@@ -92,8 +95,20 @@ void loadWithTs(
...
-        Map<InetSocketAddress, List<Cell>> hostsAndCells =
-                HostPartitioner.partitionByHost(clientPool, cells, Cell::getRowName);
+        Map<InetSocketAddress, List<Cell>> hostsAndCells;
+        try (CloseableTracer tracer = CloseableTracer.startSpan(
+                "partitionByHost",
+                ImmutableMap.of(
+                        "cells",
+                        String.valueOf(cells.size()),
+                        "tableRef",
+                        Objects.requireNonNull(LoggingArgs.tableRef(tableRef).getValue()),
```
jeremyk-91, line 105: Hmm, I was expecting something like `safeInternalTableNameOrPlaceholder` actually. If I'm not mistaken that would actually spit out the table's name regardless.

(Internally there's the assumption that table names in this KVS are safe, but still...)

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TaskRunner.java
```diff
@@ -44,22 +50,43 @@
             }
         }
 
-        List<Future<V>> futures = new ArrayList<>(tasks.size());
+        List<ListenableFuture<V>> futures = new ArrayList<>(tasks.size());
         for (Callable<V> task : tasks) {
-            futures.add(executor.submit(task));
+            DetachedSpan detachedSpan = DetachedSpan.start("task");
```
jeremyk-91, line 55: not actionable now: Yep, this is a reasonable first step. Later on we probably want a bit more information about what the tasks are doing (i.e. maybe change the method's signature or also allow users to provide a List<TaskContext> where that has metadata about th task), but this would give us some useful signal.

jeremyk-91 (commented):
Broadly looks good, I think there's one snag around the LoggingArgs

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TaskRunner.java
anjor: agreed

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CellLoader.java
anjor: Fixed.

jeremyk-91 (approved):
👍  LGTM

PR #5471, opened 2021-05-27
[PDS-146088] Force-clear the pool if no evictions are happening

Jolyon-S:
**Goals (and why)**:
One symptom of PDS-146088 is that the client pool stops evicting. Force-clearing the pool if the eviction task has not even considered anything for eviction after some period of time is one approach that might work.

Clear the client pool if the last successful eviction run was over 15 minutes ago.

No tests, which is perhaps dangerous, however I'm not sure what testing would or should look like here.

As it stands, this PR is really bad - we need to also reset the last evicted time in the evictor, and the decomp is pretty wacky. I just wanted to get the concept down before I forgot.

CCPC.

ASAP - we should get an RC out soon for internal testing.


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -339,6 +356,32 @@ private void registerPoolMetric(CassandraClientPoolHostLevelMetric metric, Gauge
         poolMetrics.registerPoolMetric(metric, gauge, poolNumber);
     }
 
+    private static final class TimingOutEvictionPolicy<T> implements EvictionPolicy<T> {
+        private final EvictionPolicy<T> delegate;
+        private final Clock clock;
+        private volatile Instant lastEviction;
```
jeremyk-91, line 362: Can we use `AtomicReference` and have the update operation take the latest one? Different threads writing their values to `lastEviction` don't actually read it, and so unless there is some other synchronisation elsewhere I think this might actually go backwards

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -78,8 +81,20 @@ public CassandraClientPoolingContainer(
...
+                new NonEvictionLoggingEvictionPolicy<>(new DefaultEvictionPolicy<>()), clock);
+        this.clientPool = createClientPool(evictionPolicy);
         this.timedRunner = TimedRunner.create(config.timeoutOnConnectionBorrow().toJavaDuration());
+
+        evictionBasedPoolClearer = () -> {
+            Instant now = clock.instant();
+            if (Duration.between(evictionPolicy.getLastEviction(), now).abs().getSeconds()
+                    > config.timeoutOnPoolEvictionFailure().toSeconds()) {
+                evictionPolicy.resetLastEviction();
+                clientPool.clear();
```
jeremyk-91, line 95: If we want to roll this out more broadly, it seems like we should add a condition that the pool is full (or at least 90% full) - otherwise we create a lot of churn for stacks that are not super busy. I'm fine with testing in this configuration though.

changelog/@unreleased/pr-5471.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: fix
+fix:
+  description: Force-evict the Cassandra client pool if no eviction has occurred for
+    a while (by default, 60 * `timeBetweenConnectionEvictionRunsSeconds`, which makes
+    this duration 20 minutes).
```
jeremyk-91, line 5: I think we just went for 15 mins here?

jeremyk-91 (commented):
Broadly makes sense. Two small suggested changes, and another that's more for if we want to productionise this. With those I'm not opposed to an RC, we can cut that tomorrow.

changelog/@unreleased/pr-5471.v2.yml
Jolyon-S: I might not have pushed

Jolyon-S: pushing

PR #5475, opened 2021-06-01
[LW] Fix getRows Caching

Jolyon-S:
**Goals (and why)**:
`RowResult<byte[]>` doesn't do equality properly.

Fix `RowResult` comparison as done in the validation.

Added a test for `getRows` validation.

Is this just stupid copy-paste? Can we make it better without compromising current code?

Diff

yesterday 🔥

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/util/ByteArrayUtilities.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/util/ByteArrayUtilities.java
@@ -16,11 +16,13 @@
 
 package com.palantir.atlasdb.util;
 
+import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.primitives.UnsignedBytes;
 import com.palantir.atlasdb.keyvalue.api.Cell;
 import com.palantir.atlasdb.keyvalue.api.RowResult;
 import java.util.Map;
 import java.util.NavigableMap;
+import java.util.Optional;
 
 public final class ByteArrayUtilities {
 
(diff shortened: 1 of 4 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/util/ByteArrayUtilities.java
```diff
@@ -43,6 +45,21 @@ public static boolean areMapsEqual(Map<Cell, byte[]> map1, Map<Cell, byte[]> map
         return true;
     }
 
+    private static boolean areByteMapsEqual(Map<byte[], byte[]> map1, Map<byte[], byte[]> map2) {
```
Jolyon-S, line 48: we _could_ arguably merge this with the one for cell.

Jolyon-S: Another thing: do maps actually work with `byte[]` as keys? Can we guarantee that the key check will work too?

Jolyon-S: as in, will `map2.containsKey(e.getKey())` ever work?

Jolyon-S: Just checked in a test, yes it does work.

gmaretic:
```suggestion
    private static boolean areByteMapsEqual(SortedMap<byte[], byte[]> map1, SortedMap<byte[], byte[]> map2) {
```
Then it does work for sure. Otherwise you need to have the same reference for the key

gmaretic: This is why I wouldn't merge it with the one for cell, because it doesn't work for arbitrary maps

gmaretic (approved):
One nit

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/util/ByteArrayUtilities.java
Jolyon-S: ^ that is fair - I wanted to be abundantly careful here

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -275,6 +279,52 @@ public void failedValidationCausesNoOpCacheToBeUsed() {
...
+                            TABLE_CELL_2,
+                            CacheValue.of(DATA_2),
+                            TABLE_CELL_3,
+                            CacheValue.of(DATA_3),
+                            TABLE_CELL_4,
+                            CacheValue.of(DATA_4)));
+            return read;
+        });
+
+        // New set of rows and columns to force new references (to test every part of the equality checking)
```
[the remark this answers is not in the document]
Jolyon-S: Not that I could see, but why not test everything?

PR #5484, opened 2021-06-03
fix operation name

anjor:
**Goals (and why)**:
We use the same operation name for multiget_slice and multiget_multislice
Fix operation name for multiget_multislice.


jeremyk-91 (approved):
whoops

PR #5487, opened 2021-06-04
[RC Only] add logging for trace

anjor:
**Goals (and why)**:
Logging to debug traces


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
```diff
@@ -71,6 +75,10 @@ public CassandraClient delegate() {
             sink.accept("consistency", consistency_level.name());
```
jkozlowski, line 75: Maybe log something here as well? This will only get logged IFF the trace is observable, if it's missing we know they're basically not getting published.

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
```diff
@@ -21,16 +21,26 @@
...
+    Logger log = LoggerFactory.getLogger(Tracing.class);
 
     static CloseableTracer startLocalTrace(@CompileTimeConstant String operation, Consumer<TagConsumer> tagTranslator) {
+        CloseableTracer tracer = CloseableTracer.startSpan(operation, FunctionalTagTranslator.INSTANCE, tagTranslator);
+        log.info(
+                "startLocalTrace",
+                SafeArg.of("operation", operation),
+                SafeArg.of("tracer class", tracer.getClass()),
+                SafeArg.of("trace", tracer.toString()));
         return CloseableTracer.startSpan(operation, FunctionalTagTranslator.INSTANCE, tagTranslator);
```
jkozlowski, line 44: It would be cool to basically figure out if traceId when open == traceId on close? CloseableTracer is final, so can't override that, maybe you can log in a finally block? You'd need to check the ordering between finally and calling close on the try-with-resources.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
anjor: The trace is not initialised at this point, so not sure what to log.

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
```diff
@@ -21,16 +21,26 @@
...
 public interface Tracing {
+    Logger log = LoggerFactory.getLogger(Tracing.class);
 
     static CloseableTracer startLocalTrace(@CompileTimeConstant String operation, Consumer<TagConsumer> tagTranslator) {
+        CloseableTracer tracer = CloseableTracer.startSpan(operation, FunctionalTagTranslator.INSTANCE, tagTranslator);
+        log.info(
+                "startLocalTrace",
+                SafeArg.of("operation", operation),
+                SafeArg.of("tracer class", tracer.getClass()),
+                SafeArg.of("trace", tracer.toString()));
```
jkozlowski, line 43: tracer.toString() doesn't do anything interesting. I think a better thing is Tracer#maybeGetTraceMetadata or something from that class that will get you the traceId so you can log it. 

Actually, the logline itself will contain the traceId, so probably just remove tracer.toString()

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
jkozlowski: Log anything, just a constant message is fine. You'll know whether the trace was published to consumers (i.e. logged) or not. If this is logged, and we don't see the trace, we know something really weird is going on.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
```diff
@@ -90,6 +98,10 @@ public CassandraClient delegate() {
...
             sink.accept("kvs", kvsMethodName);
         })) {
+            log.info(
+                    "multiget_multislice trace ",
+                    SafeArg.of("class", trace.getClass()),
+                    SafeArg.of("trace", trace.toString()));
             return client.multiget_multislice(kvsMethodName, tableRef, keyPredicates, consistency_level);
```
jkozlowski, line 105: Might also be useful to log after this, which will also provide clues as to whether the traceId somehow changed.

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
jkozlowski: Actually: possibly the actual important information is to log the whole TraceMetadata: just traceId possibly isn't the whole picture, if the parentSpanId is incorrectly parented. So yeah, I think we want to log the TraceMetadata obtained from Tracer#maybeGetTraceMetadata in all cases.

atlasdb-client/src/main/java/com/palantir/atlasdb/tracing/Tracing.java
anjor: `A try-with-resources statement can have catch and finally blocks just like an ordinary try statement. In a try-with-resources statement, any catch or finally block is run after the resources declared have been closed.`
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
anjor: added

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingCassandraClient.java
```diff
@@ -70,8 +75,18 @@ public CassandraClient delegate() {
             sink.integer("columns", numberOfColumns);
             sink.accept("consistency", consistency_level.name());
             sink.accept("kvs", kvsMethodName);
+            log.info("multiget_slice TagConsumer");
         })) {
+            log.info(
+                    "multiget_slice trace ",
+                    SafeArg.of("class", trace.getClass()),
+                    SafeArg.of("metadata", Tracer.maybeGetTraceMetadata()));
             return client.multiget_slice(kvsMethodName, tableRef, keys, predicate, consistency_level);
```
jkozlowski, line 84: one more logline after this piece?

PR #5488, opened 2021-06-06
Verbose sql exception logging to capture bind var

CVdV-au:
Goals (and why):
SQLExceptions currently provide details of all bind variables, however for binary arrays which are used for all columns except timestamps, we just print the array object number. This isn't actually useful.

Implementation Description (bullets):

We'll now print first 64 bytes in hex notation (with ... indicating truncation)
Oracle's HEXTORAW can use this representation (quoted)
Postgres' notation '\x'
Testing (What was existing testing like? What have you done to improve it?):
N/A, N/A

Concerns (what feedback would you like?):

Where should we start reviewing?:

Priority (whenever / two weeks / yesterday):
Low


CVdV-au:
Abandoned - now I have working membership of Palantir org

PR #5489, opened 2021-06-07
Improve utility of SQLException logging

CVdV-au:
**Goals:**

SQLExceptions currently provide details of all bind variables, however for binary arrays which are used for all columns except timestamps, we just print the array object number. This isn't actually useful.

**Implementation Description:**

- We'll now print first 64 bytes in hex notation (with ... indicating truncation)
- Oracle's HEXTORAW can use this representation (quoted)
- Postgres' notation '\x'

** Testing (What was existing testing like? What have you done to improve it?): **
N/A, N/A

Priority (whenever / two weeks / yesterday):
Low

### DIFF
--- a/changelog/@unreleased/pr-5489.v2.yml
+++ b/changelog/@unreleased/pr-5489.v2.yml
@@ -1,0 +1,22 @@
+type: improvement
+improvement:
+  description: |-
+    Improve utility of SQLException logging
+
+    **Goals:**
+
+    SQLExceptions currently provide details of all bind variables, however for binary arrays which are used for all columns except timestamps, we just print the array object number. This isn't actually useful.
+
+    **Implementation Description:**
+
+    - We'll now print first 64 bytes in hex notation (with ... indicating truncation)
+    - Oracle's HEXTORAW can use this representation (quoted)
+    - Postgres' notation '\x'
+
+    ** Testing (What was existing testing like? What have you done to improve it?): **
+    N/A, N/A
+
+    Priority (whenever / two weeks / yesterday):
+    Low
+  links:
+  - https://github.com/palantir/atlasdb/pull/5489
--- a/changelog/@unreleased/pr-5489.v2.yml
+++ b/changelog/@unreleased/pr-5489.v2.yml
@@ -1,22 +1,5 @@
 type: improvement
 improvement:
-  description: |-
-    Improve utility of SQLException logging
-
-    **Goals:**
-
-    SQLExceptions currently provide details of all bind variables, however for binary arrays which are used for all columns except timestamps, we just print the array object number. This isn't actually useful.
-
-    **Implementation Description:**
-
-    - We'll now print first 64 bytes in hex notation (with ... indicating truncation)
-    - Oracle's HEXTORAW can use this representation (quoted)
-    - Postgres' notation '\x'
-
-    ** Testing (What was existing testing like? What have you done to improve it?): **
-    N/A, N/A
-
-    Priority (whenever / two weeks / yesterday):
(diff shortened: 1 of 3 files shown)


commons-db/src/main/java/com/palantir/nexus/db/sql/BasicSQLUtils.java
```diff
@@ -329,20 +328,36 @@ public static void toStringSqlArgs(final StringBuilder sb, Object[] args) {
...
             sb.append(")\n"); // $NON-NLS-1$
         }
     }
 
+    private static final int MAX_ARRAY_PRINT_BYTES = 64;
+
+    // Oracle use HEXTORAW('0f1f') to create a 2 byte RAW literal
+    // Postgres use '\\x0f1f' to create a 2 byte bytea literal
+    private static void prettyPrintByteArray(final StringBuilder sb, byte[] args) {
+        int printLength = Math.min(args.length, MAX_ARRAY_PRINT_BYTES);
```
sudiksha27, line 352: can we change the second argument to arg?

commons-db/src/main/java/com/palantir/nexus/db/sql/BasicSQLUtils.java
```diff
@@ -329,20 +328,36 @@ public static void toStringSqlArgs(final StringBuilder sb, Object[] args) {
...
+        for (int i = 0; i < printLength; i++) {
+            sb.append(String.format("%02x", args[i] & 0xff));
+        }
+        if (args.length > MAX_ARRAY_PRINT_BYTES) {
+            sb.append("...");
+        }
+    }
+
     public static void toStringSqlArgs(final StringBuilder sb, Iterable<Object[]> args) {
```
sudiksha27, line 361: can we please add tests for this util?

commons-db/src/main/java/com/palantir/nexus/db/sql/BasicSQLUtils.java
```diff
@@ -329,20 +328,36 @@ public static void toStringSqlArgs(final StringBuilder sb, Object[] args) {
...
+                        prettyPrintByteArray(sb, (byte[]) arg);
                     } else {
-                        sb.append(arg.toString());
+                        sb.append(arg);
                     }
                     sb.append("'"); // $NON-NLS-1$
                 }
-                if (i != args.length - 1) {
+                if (i != 0 && i != args.length - 1) {
                     sb.append(", ");
```
sudiksha27, line 340: I think what we had previously was correct - can we add a test please?

changelog/@unreleased/pr-5489.v2.yml
```diff
@@ -0,0 +1,22 @@
+type: improvement
+improvement:
+  description: |-
+    Improve utility of SQLException logging
```
sudiksha27, line 4: changelog needs to be shortened - usually it is a one-liner explaining what the change does.

sudiksha27 (commented):
Broadly makes sense - would be good to have a sanity test.

commons-db/src/main/java/com/palantir/nexus/db/sql/BasicSQLUtils.java
CVdV-au: Done

changelog/@unreleased/pr-5489.v2.yml
CVdV-au: Done

sudiksha27 (approved):
Thanks for addressing the comments 👍🏼

PR #5492, opened 2021-06-07
Make maxConnectionAge default 888 seconds (a little less than 15min)

CVdV-au:
**Goals (and why)**:
This is required so products don't need to override this value to achieve compliance with STIG/Fedramp O121-C2-016500
which requires idle timeout at 15 min. Hikari cautions that this value should be set "a few seconds less
than any server side enforcement" hence 888.

None - the drop of this value should have no real consequence at either client or DB server, as it only applies to connections that are fulfilling the minConnections constraint (which we set to 8). The vast majority of our connections are actually closed as a results of the maxIdleTime configuration (600).

### DIFF
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
@@ -79,7 +79,7 @@ public abstract class ConnectionConfig {
 
     @Value.Default
     public Integer getMaxConnectionAge() {
-        return 1800;
+        return 888;
     }
 
     @Value.Default
--- a/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
+++ b/atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
@@ -77,6 +77,12 @@ public abstract class ConnectionConfig {
         return 256;
     }
 
+    /**
+     * STIG O121-C2-016500 and Fedramp requires idle connections are closed within 15 mins.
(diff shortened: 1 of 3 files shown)


berler (approved):

atlasdb-dbkvs-hikari/src/main/java/com/palantir/nexus/db/pool/config/ConnectionConfig.java
```diff
@@ -79,7 +79,7 @@ public int getMaxConnections() {
 
     @Value.Default
     public Integer getMaxConnectionAge() {
-        return 1800;
+        return 888;
```
j-baker, line 82: ideally we'd write this as

```
return SERVER_SIDE_EXPECTED_TIMEOUT - A_FEW_SECONDS
```

so someone understands how the number was achieved

berler: I would prefer a comment over adding more constants. Breaking it out into multiple variables makes it harder for someone looking at the code to immediately see what the actual value is, and you probably would want a comment anyway so that people don't increase it later thinking that this is safe.

The important thing which is not obvious is that hikari does not enforce the idle timeout (at all) when the number of idle connections is equal or less than the min pool size. Also when a connection is returned to the pool, hikari does not know how long the connection was idle prior to being returned to the pool. For this reason, this field is effectively the actual idle timeout.

changelog/@unreleased/pr-5492.v2.yml
```diff
@@ -0,0 +1,15 @@
+type: improvement
+improvement:
+  description: |-
+    Make maxConnectionAge default 888 seconds (a little less than 15min)
```
sudiksha27, line 4: we should make the changelog a one-liner.

sudiksha27 (approved):
Should fix changelog but otherwise LGTM!

jeremyk-91 (approved):

PR #5493, opened 2021-06-07
Delete the (dangerous) SingleCallProxy

carterkozak:
Previously, this proxy would fail if the object was logged or
hashed prior to the expected methods being called.
The only user (verified internal as well) is BatchingVisitableFromIterable
where I've implemented a more reasonable check.


atlasdb-commons/src/main/java/com/palantir/common/proxy/SingleCallProxy.java
```diff
@@ -1,59 +0,0 @@
...
-import com.palantir.logsafe.exceptions.SafeIllegalStateException;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Proxy;
-import java.util.concurrent.atomic.AtomicBoolean;
-
-/**
- * Ensure that only a single method is ever called.  Throw {@link IllegalStateException} on all subsequent calls.
- */
-public final class SingleCallProxy implements DelegatingInvocationHandler {
```
carterkozak, line 27: Originally I planned to deprecate this, however it's not actually used outside of atlas.

atlasdb-commons/src/main/java/com/palantir/common/base/BatchingVisitableFromIterable.java
```diff
@@ -59,4 +59,34 @@ public BatchingVisitableFromIterable(Iterable<? extends T> iterable) {
...
+
+        SingleCallBatchingVisitable(BatchingVisitable<T> delegate) {
+            this.delegate = Preconditions.checkNotNull(delegate, "Delegate BatchingVisitable is required");
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public <K extends Exception> boolean batchAccept(int batchSize, AbortingVisitor<? super List<T>, K> visitor)
+                throws K {
+            BatchingVisitable<T> current = (BatchingVisitable<T>) delegateUpdater.getAndSet(this, null);
```
jeremyk-91, line 80: nice!

jeremyk-91 (approved):
👍  thanks! yeah this is much better than the original

PR #5495, opened 2021-06-08
[DO NOT MERGE] [PDS-146088] Test infinite loop hypothesis

jeremyk-91:
**Goals (and why)**:
- PDS-146088

- Copy the generic object pool code to Atlas.
- Introduce debug logs at key places within the client pool.

**Testing (What was existing testing like?  What have you done to improve it?)**: existing tests should be enough?

- Does this break anything?
- Is licensing done correctly?

**Where should we start reviewing?**: Commit by commit. Commit 2 is a bit large - just confirm it's a copy and then the CHANGELOG bits in there.


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
```diff
@@ -348,6 +349,11 @@ private NonEvictionLoggingEvictionPolicy(EvictionPolicy<T> delegate) {
 
         @Override
         public boolean evict(EvictionConfig config, PooledObject<T> underTest, int idleCount) {
+            if (log.isDebugEnabled()) {
+                log.debug("Beginning an eviction attempt from the Cassandra client pool",
+                        SafeArg.of("underTestState", underTest.getState()),
+                        SafeArg.of("idleState", underTest.getIdleTimeMillis()));
+            }
             boolean delegateResult = delegate.evict(config, underTest, idleCount);
```
Jolyon-S, line 357: Do we want to wrap with `TimedRunner`?

atlasdb-cassandra/src/main/java/org/apache/commons/pool2/impl/CommonsPoolGenericObjectPool.java
```diff
@@ -774,14 +777,31 @@ public void evict() throws Exception {
                     } catch (final NoSuchElementException nsee) {
                         // Object was borrowed in another thread
                         // Don't count this as an eviction test so reduce i;
+                        // CHANGELOG: Additional logging
+                        if (log.isDebugEnabled()) {
+                            log.debug("Not counting an eviction candidate, because we believe it was borrowed in "
```
Jolyon-S, line 784: I suggest that we also log:
* `i`
* `m`
* some sort of UUID per eviction run. Then we can group logs by eviction run to confirm that it is indeed the same run that is looping indefinitely.

atlasdb-cassandra/src/main/java/org/apache/commons/pool2/impl/CommonsPoolGenericObjectPool.java
```diff
@@ -774,14 +777,31 @@ public void evict() throws Exception {
...
 
+                    // CHANGELOG: Declaration of a variable to check the object's state before starting eviction.
+                    PooledObjectState preTestState = underTest.getState();
+
                     if (!underTest.startEvictionTest()) {
                         // Object was borrowed in another thread
                         // Don't count this as an eviction test so reduce i;
+                        // CHANGELOG: Additional logging
+                        if (log.isDebugEnabled()) {
+                            log.debug("Not counting an eviction candidate, because we weren't able to start an "
```
Jolyon-S, line 800: As above

atlasdb-cassandra/src/main/java/org/apache/commons/pool2/impl/CommonsPoolGenericObjectPool.java
jeremyk-91: Makes sense.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientPoolingContainer.java
jeremyk-91: We could, but I think it's probably not needed. The delegate impl looks like this:

```
    @Override
    public boolean evict(final EvictionConfig config, final PooledObject<T> underTest,
            final int idleCount) {

        if ((config.getIdleSoftEvictTime() < underTest.getIdleTimeMillis() &&
                config.getMinIdle() < idleCount) ||
                config.getIdleEvictTime() < underTest.getIdleTimeMillis()) {
            return true;
        }
        return false;
    }
```

so this should be a bunch of memory reads. Actually this starting log block isn't probably necessary in fact.

jeremyk-91:
closing in favour of https://github.com/palantir/atlasdb/pull/5503 - this turned out to be correct! but there's a simpler fix

PR #5496, opened 2021-06-08
fix oracle ete tests

berler:
Removes unused docker environment variables and an unused port from the
oracle docker config. Also fixes a connection leak in the test setup
code (connections must be closed).

Also update to the latest patch version of oracle.

Note that the old container version somehow barely managed to work under
the circle 10 minute no-output timeout. This newer patched version of oracle
is a slightly larger image and would consistently fail due to a circle timeout.
Adding a workaround to pull the image ahead of time instead of pulling during
testing.


jeremyk-91 (approved):
👍  looks good, thanks!

PR #5497, opened 2021-06-09
Make TransactionConflictExceptions actually useful

Jolyon-S:
**Goals (and why)**:
Conflict exceptions have a load of info that is not really useful to us as-is. Aim is to change that by recording the table - it's not logged anywhere, just accessible so that internally we can use it.

Make TCE and TSCE track the conflicting table.

N/A.

Should be safe.

diff

ASAP.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SerializableTransaction.java
```diff
@@ -924,11 +924,13 @@ protected TransactionScopedCache getCache() {
...
-                                    + tableRef + ".  "
-                                    + "This case can cause deadlock and is very likely to be a "
-                                    + "read write conflict.");
+                            throw new TransactionSerializableConflictException(
+                                    "An uncommitted conflicting read was "
+                                            + "written after our start timestamp for table "
+                                            + tableRef + ".  "
+                                            + "This case can cause deadlock and is very likely to be a "
+                                            + "read write conflict.",
+                                    tableRef);
```
Jolyon-S, line 933: Spotless apparently.

gmaretic (approved):
🪂

PR #5498, opened 2021-06-09
Reduce flakes in the SingleLeaderPaxosSuite

Jolyon-S:
**Goals (and why)**:
Reduce flakes as part of our termly goals.

* `currentLeaderFor` can throw `NoSuchElementException` if no leader was successfully elected yet. I'm not 100% certain as to how this can be the case (given that we wait until a leader is elected first), but I feel that we may as well defend against the case, since we clearly tried but failed.

I'll run it a few times and see if it flakes.

Should or can we fix `waitUntilLeaderIsElected` instead?

Diff

Whenever, but this should make builds less painful.


gmaretic (approved):
This looks fine

PR #5501, opened 2021-06-10
Rise of the Coffee Elf (Java 11 Adoption)

jeremyk-91:
**Goals (and why)**:
- Move to Java 11! Large internal product has moved, so we can go ahead.

- Update source compatibility to 11.
- Remove references to Java-8 only GC options where they are used.
- Introduce a single JDK 9 only feature to confirm that we are building with 11.

- Existing tests should suffice

- Some of the Circle containers will now run with G1 instead of CMS, which might affect flake rates?

- shared.gradle


changelog/@unreleased/pr-5501.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: break
```
Jolyon-S, line 1: Version 1.0.0? Time to break all of our APIs?!?!

anjor:
lgtm!

anjor (approved):

PR #5503, opened 2021-06-11
[PDS-146088] Validations must not throw an exception or the commons-pool implodes

jeremyk-91:
**Goals (and why)**:
PDS-146088: don't enter a death spiral when Cassandra bounces

- Ensure calls to validateObject() don't throw exceptions.

- Added a very basic test confirming `validateObject()` doesn't throw. Note that `destroyObject()` still does, but that is allowed.

- Do we need any further logging or rate limiting for future work we might do here?

**Where should we start reviewing?**: CassandraClientFactory

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
@@ -189,7 +189,15 @@ public class CassandraClientFactory extends BasePooledObjectFactory<CassandraCli
 
     @Override
     public boolean validateObject(PooledObject<CassandraClient> client) {
-        return client.getObject().getOutputProtocol().getTransport().isOpen();
+        try {
+            return client.getObject().getOutputProtocol().getTransport().isOpen();
+        } catch (Throwable t) {
+            log.info(
+                    "Failed when attempting to validate a Cassandra client in the Cassandra client pool."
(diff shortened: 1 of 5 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
```diff
@@ -189,7 +189,15 @@ private static void login(Client client, CassandraCredentialsConfig config) thro
 
     @Override
     public boolean validateObject(PooledObject<CassandraClient> client) {
-        return client.getObject().getOutputProtocol().getTransport().isOpen();
+        try {
+            return client.getObject().getOutputProtocol().getTransport().isOpen();
+        } catch (Throwable t) {
+            log.info(
```
Jolyon-S, line 195: Should we warn? Is there any safe information we can log about the client?

Jolyon-S (approved):
LGTM

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/CassandraClientFactory.java
jeremyk-91: I think info is OK, as it doesn't indicate anything bad about this specific user node / that the health is trending badly.

Client information: yeah, we can add the host (as we do in `destroyObject()`) - I've added this.

PR #5506, opened 2021-06-14
Change semantics of priority tables to be exclusive

jeremyk-91:
**Goals (and why)**:
- When priority tables are specified, background sweep will previously begin one iteration of sweep for another table, realise it shouldn't do that, and then begin one iteration for another table and so on. This behaviour does not make sense, and furthermore could mean that we can't "make background sweep run ONLY on table xyz", because it might select other tables that are eligible in a larger deployment or if there are more threads.

- Only select tables to be considered for sweeping when priority tables are non-empty IF there is a priority table.

- Updated the one test that matters.

- This is a semantic change, would it cause problems for anyone?

**Priority (whenever / two weeks / yesterday)**: Tuesday would be good


sudiksha27 (approved):
LGTM!

jeremyk-91:
Thanks @sudiksha27 !

PR #5507, opened 2021-06-15
Exclusion Hunter IV: CatchBlockLogException

Jolyon-S:
**Goals (and why)**:
Hunt down errorprone exclusions.

Add exceptions to log messages in catch clauses.

N/A.

In at least one place, explicit care was taken to log the exception unsafely. Is this necessary these days - that was done in 2017.

Diff

Whenever

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
@@ -462,25 +462,25 @@ public final class TracingKeyValueService extends ForwardingObject implements Ke
     }
 
     @MustBeClosed
-    private static CloseableTracer startLocalTrace(@CompileTimeConstant String operation) {
+    private static CloseableTracer startLocalTrace(@CompileTimeConstant final String operation) {
         return CloseableTracer.startSpan(operation);
     }
 
     @MustBeClosed
     private static CloseableTracer startLocalTrace(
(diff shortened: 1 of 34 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingKeyValueService.java
```diff
@@ -462,25 +462,25 @@ public boolean shouldTriggerCompactions() {
     }
 
     @MustBeClosed
-    private static CloseableTracer startLocalTrace(@CompileTimeConstant String operation) {
+    private static CloseableTracer startLocalTrace(@CompileTimeConstant final String operation) {
```
Jolyon-S, line 465: for some reason, non compile time constant errors started popping up too.

lock-impl/src/main/java/com/palantir/lock/impl/LockServiceImpl.java
```diff
@@ -460,10 +460,7 @@ public LockResponse lockWithFullLockResponse(LockClient client, LockRequest requ
                     entry.getKey().get(client, entry.getValue()).unlock();
                 }
             } catch (Throwable e) { // (authorized)
-                log.error(
-                        "Internal lock server error: state has been corrupted!!",
-                        UnsafeArg.of("exception", e),
-                        SafeArg.of("stacktrace", e.getStackTrace()));
```
Jolyon-S, line 466: Is this really necessary these days???

gmaretic (approved):

PR #5508, opened 2021-06-15
fix docs publishing

anjor:
**Goals (and why)**:
Publishing is currently broken due to API breaks in the docs infrastructure.
API breaks fixed in docs.


jeremyk-91 (approved):
Approving to unblock builds. It's kind of awkward that we have Markdown files being rendered by an RST renderer (in the ADRs), so I think that's something we could improve longer term, but in any case getting the build to green is higher priority. Thanks for going through this!

PR #5513, opened 2021-06-17
[DO NOT MERGE] [PDS-193911] Expose values - be careful about usage.

jeremyk-91:


jeremyk-91:
Can close now as the underlying ticket was fixed

PR #5518, opened 2021-06-23
[Exclusions] Missing fail is a fail

jeremyk-91:
**Goals (and why)**:
Not have too many errorprone exclusions. This one is very low risk - in general in a test if you have
```
try {
  bla();
} catch (ExceptionT e) {
  // expected
}
```

this will pass even if no exception is thrown. *However*, there are cases where not throwing is legit (see the suppression).

- Add fail() where it should be
- Add suppression

Not much.

- Is the suppression correct?

**Where should we start reviewing?**: eh, it's small


Jolyon-S (approved):
4:1

PR #5519, opened 2021-06-23
[DB TimeLock] 3A: AtlasDB Runtime Configuration for TimeLock

jeremyk-91:
**Goals (and why)**:
- Support live reloading of KVS configuration when one is using timelock. This is important at deployments using db-timelock, where the password might get rotated

- Wiring, plus a bunch of tests

- Some unit tests were added.

- Is the design a bit too heavily engineered?

**Where should we start reviewing?**: eh

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/timelock/config/TsBoundPersisterRuntimeConfiguration.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/config/TsBoundPersisterRuntimeConfiguration.java
@@ -1,0 +1,24 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


timelock-agent/src/main/java/com/palantir/timelock/config/TsBoundPersisterRuntimeConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.timelock.config;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonSubTypes(@JsonSubTypes.Type(value = DatabaseTsBoundPersisterRuntimeConfiguration.class, name = "database"))
```
gmaretic, line 23: Why the subtyping?

gmaretic (approved):
This seems reasonable, if a bit overly complicated. We don't have a test that the kvs actually reloads the config but that should be covered by existing unit tests

timelock-agent/src/main/java/com/palantir/timelock/config/TsBoundPersisterRuntimeConfiguration.java
jeremyk-91: we discussed this offline:

>  re: subtyping, I didn’t want to have an optional block for kvs runtime config specifically (because we might end up needing a different optional block for runtime configs for other persistence modes in the future)

jeremyk-91:
Yeah, I didn't test anything once we reached the AtlasDbFactory as that should be tested elsewhere, yeah

PR #5525, opened 2021-06-29
[HD] Better(?) KVS calls for getSortedColumns when desired number of results is known

gmaretic:
**Goals (and why)**:
`getSortedColumns` returns an iterator that will lazily read more from the KVS as needed. If we know in advance how many results we need, the reads can be better optimised.

The existing method's implementation reads `batchHint`/`numRows` entries from each row. The user would need to know (and rely on) this implementation detail to optimise their usage of `getSortedColumns`. The new method allows the user to specify the desired number of results, and the read batch size is calculated to not need additional lookups in 99.5% of cases.

Need to add tests.

Is this even useful? I was thinking we could in general have an A/B testing framework that we could use in cases like these where we could call either method with some probability and then compare metrics? Sounds like another hack(day/week) project, but it would probably be super useful as this is difficult to measure otherwise. Also note that we would probably want to measure the number of KVS calls since the lazy nature of the iterator makes time-based metrics useless.

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -472,11 +472,39 @@ public class SnapshotTransaction extends AbstractTransaction implements Constrai
             return Collections.emptyIterator();
         }
         Iterable<byte[]> distinctRows = getDistinctRows(rows);
+        return getSortedColumnsInternal(
+                tableRef,
+                distinctRows,
+                batchColumnRangeSelection,
+                getPerRowBatchSize(batchColumnRangeSelection, Iterables.size(distinctRows)));
+    }
+
+    @Override
+    public Iterator<Map.Entry<Cell, byte[]>> getSortedColumns(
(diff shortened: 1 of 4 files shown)


jeremyk-91 (commented):
Had a look through here. The code for this seems fine, though agree that I would be in favour of some kind of A/B testing before we roll this out more broadly.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -472,11 +472,39 @@ public void markTableInvolved(TableReference tableRef) {
...
+    public Iterator<Map.Entry<Cell, byte[]>> getSortedColumns(
+            TableReference tableRef, Iterable<byte[]> rows, ColumnRangeSelection columnRangeSelection, int numResults) {
+        checkGetPreconditions(tableRef);
+        if (Iterables.isEmpty(rows)) {
+            return Collections.emptyIterator();
+        }
+        Preconditions.checkArgument(numResults <= 1_000_000, "Requested number of results is too large.");
+        Iterable<byte[]> distinctRows = getDistinctRows(rows);
+        int numRows = Iterables.size(distinctRows);
+        int batchSize = getPerRowBatchSizeForFixedSize(numResults, numRows);
```
jeremyk-91, line 492: let's special case 1 row case to not multiply

PR #5529, opened 2021-07-05
[PDS-197811] Autobatcher timeout protection

jeremyk-91:
**Goals (and why)**:
- PDS-197811
- Suppose an autobatcher operation takes a long time for whatever reason, this shouldn't as far as possible block up all of AtlasDB.

- Wrap the autobatcher functions in a simple time-limiter

- Will add.

- This creates 2 threads per autobatcher instead of 1, but I think is useful enough. There are maybe O(10) autobatchers at a single atlasdb. Atlasdb proxy has cross client batching.

**Where should we start reviewing?**: Autobatchers


atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -79,25 +94,89 @@ private Autobatchers() {}
...
+                throw new RuntimeException(e);
+            } catch (TimeoutException e) {
+                log.info(
+                        "Autobatcher timed out and request had to be interrupted",
+                        SafeArg.of("safeLoggablePurpose", safeLoggablePurpose),
+                        e);
+                throw new RuntimeException(e);
+            } catch (UncheckedExecutionException e) {
+                Throwables.throwIfUnchecked(e.getCause());
+                throw new RuntimeException(e.getCause());
```
gmaretic, line 139: This seems like something that would be worth having in our `Throwables` class

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -79,25 +94,89 @@ private Autobatchers() {}
...
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            } catch (TimeoutException e) {
+                log.info(
+                        "Autobatcher timed out and request had to be interrupted",
+                        SafeArg.of("safeLoggablePurpose", safeLoggablePurpose),
+                        e);
+                throw new RuntimeException(e);
+            } catch (ExecutionException | UncheckedExecutionException e) {
+                Throwables.throwIfUnchecked(e.getCause());
```
gmaretic, line 160: Will an `ExecutionException` even have an unchecked cause? See why you would still do it this way, but curious

atlasdb-autobatch/src/test/java/com/palantir/atlasdb/autobatch/AutobatchersTests.java
```diff
@@ -0,0 +1,75 @@
...
+                    list.forEach(element -> element.result().set(new Object()));
+                })
+                .batchFunctionTimeout(Duration.ofSeconds(1))
+                .safeLoggablePurpose("testing")
+                .build();
+
+        ListenableFuture<Object> response = autobatcher.apply(new Object());
+
+        // Ensure that the second object is not put in the first batch
+        Uninterruptibles.sleepUninterruptibly(500, TimeUnit.MILLISECONDS);
```
gmaretic, line 62: Can we make this deterministic with a latch?

gmaretic (commented):
Overall design looks good, but have some questions and a nit in the test

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -79,25 +94,89 @@ private Autobatchers() {}
...
+
+    private static <I, O> Consumer<List<BatchElement<I, O>>> maybeWrapWithTimeout(
+            Consumer<List<BatchElement<I, O>>> batchFunction, EventHandlerParameters parameters) {
+        return parameters
+                .batchFunctionTimeoutContext()
+                .map(context -> wrapWithTimeout(batchFunction, parameters.safeLoggablePurpose(), context))
+                .orElse(batchFunction);
+    }
+
+    private static <I, O> CoalescingRequestFunction<I, O> maybeWrapWithTimeout(
```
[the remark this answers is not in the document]
jeremyk-91: Yeah, I tried a bunch of things but couldn't find a nice way to organise it

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
```diff
@@ -126,19 +205,63 @@ private AutobatcherBuilder(Function<Integer, EventHandler<BatchElement<I, O>>> h
...
             Preconditions.checkArgument(purpose != null, "purpose must be provided");
 
-            int bufferSizeValue = bufferSize.orElse(DEFAULT_BUFFER_SIZE);
+            ImmutableEventHandlerParameters.Builder parametersBuilder = ImmutableEventHandlerParameters.builder();
+            bufferSize.ifPresent(parametersBuilder::batchSize);
+            parametersBuilder.safeLoggablePurpose(purpose);
+            Optional<TimeoutOrchestrationContext> timeoutOrchestrationContext =
+                    batchFunctionTimeout.map(timeout -> ImmutableTimeoutOrchestrationContext.builder()
+                            .batchFunctionTimeout(timeout)
+                            .exclusiveExecutor(PTExecutors.newCachedThreadPool("autobatcher." + purpose + "-timeout"))
```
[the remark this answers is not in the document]
jeremyk-91: Yep, that's right. I'll answer 2 first as that's easier

> If we are not into sharing, why don't we use a single threaded executor -- aren't we guaranteed to only ever have one request in progress per autobatcher?

Strictly speaking, this is not necessarily true after this PR: if the batch function is _uninterruptible_, it is possible for us to have multiple requests in progress per autobatcher. (I've had a look in the Dialogue client and blocking clients are interruptible). In practice for most of our functions I'd expect one thread. What do you think?

> Why don't we share one across all autobatchers?

This allows better executor metrics (e.g. the safe loggable purpose) to be available in the event of a runaway executor. Other than that I didn't really have a reason

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
jeremyk-91: will add

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
jeremyk-91: In this context no, as `SimpleTimeLimiter`'s contract is that it is used for checked exceptions (though in general, yes). I found this preferable to handling the conditional logic here explicitly, though happy to change if you prefer the alternative.

atlasdb-autobatch/src/test/java/com/palantir/atlasdb/autobatch/AutobatchersTests.java
jeremyk-91: yeah let's do that

atlasdb-autobatch/src/main/java/com/palantir/atlasdb/autobatch/Autobatchers.java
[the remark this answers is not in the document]
jeremyk-91: (also, discussed offline: the current method actually shares them underneath the hood in terms of execution, because they use the Atlas view executor.)

gmaretic (approved):

PR #5530, opened 2021-07-05
reduce backoff

anjor:
**Goals (and why)**:
A lot of calls seem to be requiring upto 3 retries due to KVS slowness, which quickly pile on latency. This will reduce the time between those retries.


changelog/@unreleased/pr-5530.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: |-
+    Reduces backoff upon receiving an exception from KVS.
```
jeremyk-91, line 4: let's say Cassandra here, to be clear that postgres/oracle users aren't affected

jeremyk-91 (approved):

changelog/@unreleased/pr-5530.v2.yml
anjor: :+1:

PR #5512, opened 2021-06-16
Upgrade dependencies

schlosna:
**Goals (and why)**: Unblock excavators as https://github.com/palantir/atlasdb/pull/5407 seems stuck

**Implementation Description (bullets)**:

**Testing (What was existing testing like?  What have you done to improve it?)**:

**Concerns (what feedback would you like?)**:

**Where should we start reviewing?**:

**Priority (whenever / two weeks / yesterday)**:

PR #5531, opened 2021-07-05
Upgrade dependencies

schlosna:
**Goals (and why)**: 
Unblock excavators as #5407 seems stuck, replaces https://github.com/palantir/atlasdb/pull/5512

sort versions.props, bump dependency per commit

versions.props, versions.lock, commit per bump


Jolyon-S (approved):
LGTM! Reviewed each commit in turn, and it seems reasonable at each stage.

Thank you very much!

PR #5533, opened 2021-07-05
[LW] Reduce flakes on ETE test

Jolyon-S:
**Goals (and why)**:
Reduce test flakes

Use flake retrying rule

This is about tests

The real issue is probably due to some random other transaction starting in the background

Diff

Whenever


schlosna (approved):

jeremyk-91 (approved):
👍

PR #5535, opened 2021-07-05
Fix baseline errors

schlosna:
**Goals (and why)**:
Fix errors flagged by https://github.com/palantir/atlasdb/pull/5517

Suppressed many `BadAssert` that are on performance sensitive paths and where we might not want to throw at runtime, only under test.

Did not migrate `ProxyNonConstantType` to [proxy processor](https://github.com/palantir/proxy-processor) yet, worth doing in a follow up, but may require a bit more work across consumers.


jeremyk-91 (approved):
sweet, took a pass and it makes sense (in particular, agree with _not_ changing the asserts in the prod code, at least not now, _but_ yes to changing those in test)!

PR #5539, opened 2021-07-06
[PDS-198280] Banishing Corrupted Banishment: Prevent TimeLock Self-Confusion

jeremyk-91:
**Goals (and why)**:
- See PDS-198280 - the timestamp corroboration check was failing on occasion at a number of installations. It was noted that this correlates with (1) leader elections, AND (2) very high timestamp load (in the region of 100K QPS).
- Not having corruption is, well, important.

- TimeLock previously had an edge case where it was possible for it to break its standard guarantees of giving out numbers that are monotonically increasing. Specifically, this occurs as a race condition when a timestamp service from a previous term may still be live if request threads have acquired a reference to it and then slept, and needs to refresh its bound (the fact that it should be invalidated by the leadership proxy is irrelevant, because timestamp services generally aren't meaningfully closed). This and a newly created timestamp service can race to propose a new value, and if they happen to propose the *same* value, both instances of the service will think their Paxos succeeded, meaning that the timestamps used to service the request prompting a refresh of the old bound could potentially be the same as timestamps later handed out by the newly created service.
- See following diagram for thought process and pathing.
ALP-path

- This change prevents this case, because a closing of the old delegate when gaining leadership happens before the new delegate is created, and also a read to confirm the old delegate has not been closed happens before a response is returned to the user. Thus, in the previous case, even if the old delegate falls into the same race condition, any timestamps it tries to return will not actually be returned to the user, because it will be caught in the `LeadershipGuardedClientAwareManagedTimestampService`.
 
- Added a simple test. I tried to write a stress test but wasn't able to reproduce the original issue.

- I believe this is the minimal fix. There will still be a bit of noise when we run into this race condition, and it may be possible to mitigate some of that by judiciously adding closes and or close-checks at the correct places in the timestamp service code, but I think this addresses the core issue.

**Where should we start reviewing?**: PaxosResourcesFactory.java

**Priority (whenever / two weeks / yesterday)**: yesterday 🔥  or since 2016, perhaps?

### DIFF
--- a/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
+++ b/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
@@ -1,0 +1,98 @@
+/*
+ * (c) Copyright 2021 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.atlasdb.timelock.transaction.timestamp;
+
+import com.palantir.leader.NotCurrentLeaderException;
+import com.palantir.lock.v2.PartitionedTimestamps;
+import com.palantir.timestamp.TimestampRange;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link ClientAwareManagedTimestampService} that also verifies that it is not known to not be the leader before
+ * returning results to the user.
+ */
+public class LeadershipGuardedClientAwareManagedTimestampService
+        implements ClientAwareManagedTimestampService, AutoCloseable {
+    private static final Logger log =
+            LoggerFactory.getLogger(LeadershipGuardedClientAwareManagedTimestampService.class);
+
+    private final ClientAwareManagedTimestampService delegate;
+    private final AtomicBoolean isClosed = new AtomicBoolean(false);
+
+    public LeadershipGuardedClientAwareManagedTimestampService(ClientAwareManagedTimestampService delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public PartitionedTimestamps getFreshTimestampsForClient(UUID clientIdentifier, int numTimestampsRequested) {
+        return ensureStillOpen(() -> delegate.getFreshTimestampsForClient(clientIdentifier, numTimestampsRequested));
+    }
+
+    @Override
+    public void fastForwardTimestamp(long currentTimestamp) {
+        ensureStillOpen(() -> {
+            delegate.fastForwardTimestamp(currentTimestamp);
+            return null;
+        });
+    }
+
+    @Override
+    public String ping() {
+        return ensureStillOpen(delegate::ping);
+    }
+
+    @Override
+    public boolean isInitialized() {
+        return ensureStillOpen(delegate::isInitialized);
+    }
+
+    @Override
+    public long getFreshTimestamp() {
+        return ensureStillOpen(delegate::getFreshTimestamp);
+    }
+
+    @Override
+    public TimestampRange getFreshTimestamps(int numTimestampsRequested) {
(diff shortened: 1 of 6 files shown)


timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link ClientAwareManagedTimestampService} that also verifies that it is not known to not be the leader before
+ * returning results to the user.
+ */
+public class LeadershipGuardedClientAwareManagedTimestampService
```
gmaretic, line 32: `@AutoDelegate`? 🔥

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
```diff
@@ -0,0 +1,98 @@
...
+        implements ClientAwareManagedTimestampService, AutoCloseable {
+    private static final Logger log =
+            LoggerFactory.getLogger(LeadershipGuardedClientAwareManagedTimestampService.class);
+
+    private final ClientAwareManagedTimestampService delegate;
+    private final AtomicBoolean isClosed = new AtomicBoolean(false);
```
gmaretic, line 38: Why is this preferable over a volatile boolean? Isn't this a super hot path?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public PartitionedTimestamps getFreshTimestampsForClient(UUID clientIdentifier, int numTimestampsRequested) {
+        return ensureStillOpen(() -> delegate.getFreshTimestampsForClient(clientIdentifier, numTimestampsRequested));
+    }
+
+    @Override
+    public void fastForwardTimestamp(long currentTimestamp) {
+        ensureStillOpen(() -> {
```
gmaretic, line 51: This is kinda pointless, as it won't prevent the operation. As long as fast forwarding is implemented defensively, it is not a problem though. I *think* we should be good, but need to check

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+    public void doesNotReturnResultsAfterClosing() {
+        when(delegate.getFreshTimestamp()).thenReturn(88L);
+        delegatingService.close();
+        assertThatThrownBy(delegatingService::getFreshTimestamp)
+                .isInstanceOf(NotCurrentLeaderException.class)
+                .hasMessage("Lost leadership elsewhere");
+
+        // Maybe excessive, but I think reasonable
+        verify(delegate).getFreshTimestamp();
+    }
```
gmaretic, line 50: Add a test that verifies that if we close immediately after calling delegate's get timestamp, we don't return?

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
jeremyk-91: Agree - you're right, a volatile boolean is preferable. I didn't actually think about perf _that_ much as I was spending a lot of time worrying about correctness

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
jeremyk-91: I think we'd have to override everything anyway 😅

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampService.java
jeremyk-91: Discussed offline: indeed it won't prevent the operation, but it will stop users from thinking a potentially racy _successful_ operation actually succeeded

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/transaction/timestamp/LeadershipGuardedClientAwareManagedTimestampServiceTest.java
jeremyk-91: agree (also I do like the new test! 🌶️ )

gmaretic (approved):

PR #5544, opened 2021-07-07
[do not merge] [PDS-198544] Allow configuration mode with slow async lock reaping

jeremyk-91:
DNM in current form. But could be useful

**Goals (and why)**:
- allow users in more connection strained environments to still use atlasdb

- reap overdue stuff every 10 mins as opposed to 10 seconds

- not much :(

- this of course means that if a client dies it holds locks it had for 10 minutes.

**Priority (whenever / two weeks / yesterday)**: it depends


jeremyk-91:
this was just an idea that turns out to not be immediately needed.

PR #5552, opened 2021-07-13
Exclusion Hunter V: DefaultCharset

Jolyon-S:
**Goals (and why)**:
Continue the hunt.

Use `StandardCharsets.UTF_8`.

diff

whenever


lock-impl/src/main/java/com/palantir/lock/logger/LockStateYamlWriter.java
```diff
@@ -43,7 +44,8 @@
      * Creates a LockStateYamlWriter for the given file in append mode.
      */
     public static LockStateYamlWriter create(File file) throws IOException {
-        return new LockStateYamlWriter(new BufferedWriter(new FileWriter(file, true)));
+        return new LockStateYamlWriter(Files.newBufferedWriter(
+                file.toPath(), StandardCharsets.UTF_8, StandardOpenOption.CREATE, StandardOpenOption.APPEND));
```
Jolyon-S, line 48: This is probably the only change we want to sanity check

sudiksha27: Is this file ever read? Can you write a test and see if this works locally?

Jolyon-S: I have literally no idea. Could do

Jolyon-S: I believe `LockServiceTest` tests this code path

sudiksha27 (approved):

PR #5554, opened 2021-07-13
Exclusion Hunter VII: HidingField

Jolyon-S:
**Goals (and why)**:
The hunt continues.

Remove field - the class this extends already contains that field and is instantiated in the exact same way, so _probably_ isn't necessary.

Did I miss some crazy obscure behaviour here?

Diff

Whenever


sudiksha27 (approved):
👍🏼

PR #5555, opened 2021-07-13
Move isNewService check to TimelockAgent

leonz:
**Goals (and why)**:
We are looking to introduce a dynamic check for `is-new-service` which is not based on the static configuration supplied to Timelock.  To do this, we intend to override the value supplied by configuration.

So, we need to disable the invariant check that happens on immutable creation since this may not yet be the "final" state.

- Move the invariant check to the TimelockAgent create function.  I believe this is still the critical path.
- Expose the PaxosInstallConfiguration and TimelockInstallConfiguration builders so that these objects can be re-created externally.

Existing unit testing should be sufficient.  Unfortunately any of the testing that checks for failure needs to be moved into TimelockAgentTest.

ASAP

### DIFF
--- a/changelog/@unreleased/pr-5555.v2.yml
+++ b/changelog/@unreleased/pr-5555.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Move isNewService check to TimelockAgent
+  links:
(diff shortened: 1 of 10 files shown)


jeremyk-91 (approved):
👍

PR #5558, opened 2021-07-14
AutoDelegateProcessor supports gradle incremental builds

carterkozak:
Documentation: https://docs.gradle.org/6.9/userguide/java_plugin.html#sec:incremental_annotation_processing

This allows us to avoid full recompilation when sources change
in modules using the annotation processor. Faster builds.


jeremyk-91 (approved):
Makes sense, thanks!

PR #5559, opened 2021-07-14
Format generated code using Goethe (palantir-java-format for javapoet)

carterkozak:
I suspect we'll need to update some generated sources to reflect the new formatting, but I'm not familiar enough with this codebase to find those outputs without running the test suite.

**Goals (and why)**:

Standardize generated source formatting, should produce code that's easier to read.

https://github.com/palantir/goethe


atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaHotspottingTest.java
```diff
@@ -49,7 +49,7 @@ private static Schema getHotspottingSchema() {
     }
 
     private static Schema getIgnoredHotspottingSchema() {
-        Schema ignoredHotspottingSchema = new Schema(SCHEMA_NAME, "valid.package", Namespace.DEFAULT_NAMESPACE);
+        Schema ignoredHotspottingSchema = new Schema(SCHEMA_NAME, "valid.pkg", Namespace.DEFAULT_NAMESPACE);
```
carterkozak, line 52: I got a chuckle out of this.

jeremyk-91: Wasn't expecting this... it's fine, though is this a thing we expect from the formatter?

jeremyk-91 (approved):
Great! Yeah, I think this will be nice :)

atlasdb-client/src/test/java/com/palantir/atlasdb/table/description/SchemaHotspottingTest.java
carterkozak: Yep, it uses the jdk ast parser, which does some validation.

PR #5561, opened 2021-07-14
Config: Allow connecting to different timelock namespace than KVS

blacksmithgu:
There is currently no way to connect to a different timelock namespace than the configured KVS namespace. When using postgres/oracle as a DB-based KVS, the KVS namespace is set to their database name (such as 'db' or 'opgid' or something), which is undesirable as a timelock namespace.

This simply extends an existing config option to allow for the timelock namespace to differ.


blacksmithgu:
I am open to adding a new config option instead of co-opting an existing one, if necessary.

jeremyk-91 (approved):
👍

jeremyk-91 (commented):
This is fine - could we update the failing tests to reflect the new behaviour as well? Thanks!

blacksmithgu:
Test fixed - merging now.

PR #5562, opened 2021-07-15
[Exclusions] NonAtomicVolatileUpdate

jeremyk-91:
**Goals (and why)**:
- Not have errorprone exclusions

- Justify the one instance of this thing

- No actual code change

- Is there something trickier I'm missing here?
- Strictly speaking one can rely on an implementation detail of `ExpiringMemoizingSupplier`, that the run on one thread in JMM terms _happens-before_ the run on another (because of the volatile read/write of the time) to have the version be a non-volatile variable, but since I didn't see an explicit guarantee of this in the API I'd prefer not to do that.

**Where should we start reviewing?**: +3-1

### DIFF
--- a/timestamp-impl/src/test/java/com/palantir/timestamp/InMemoryTimestampBoundStore.java
+++ b/timestamp-impl/src/test/java/com/palantir/timestamp/InMemoryTimestampBoundStore.java
@@ -30,7 +30,6 @@ public class InMemoryTimestampBoundStore implements TimestampBoundStore {
         return upperLimit;
     }
 
-    @SuppressWarnings("NonAtomicVolatileUpdate")
     @Override
     public void storeUpperLimit(long newLimit) throws MultipleRunningTimestampServiceError {
         if (shouldThrowErrorMultipleServerError) {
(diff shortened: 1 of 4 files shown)


timestamp-impl/src/test/java/com/palantir/timestamp/InMemoryTimestampBoundStore.java
```diff
@@ -28,6 +30,7 @@ public long getUpperLimit() {
         return upperLimit;
     }
 
+    @SuppressWarnings("NonAtomicVolatileUpdate")
```
Jolyon-S, line 33: Why not make `numberOfAllocations` atomic too?

Jolyon-S (commented):
Can you leave a comment as to why you suppressed this in the other place? Thanks

timestamp-impl/src/test/java/com/palantir/timestamp/InMemoryTimestampBoundStore.java
jeremyk-91: it is! but you're right, this line shouldn't be there

Jolyon-S (approved):

PR #5567, opened 2021-07-21
[Exclusions] Missing summaries

anjor:
**Goals (and why)**:
Add summaries to a bunch of classes/interfaces.

Could some of the summaries be better?

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/KeyedWeakReference.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/KeyedWeakReference.java
@@ -19,7 +19,7 @@ import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 
 /**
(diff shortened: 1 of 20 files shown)


atlasdb-commons/src/main/java/com/palantir/util/KeyedWeakReference.java
```diff
@@ -19,7 +19,7 @@
 import java.lang.ref.WeakReference;
 
 /**
- *
+ * Keyed Weak Reference.
```
Jolyon-S, line 22: lol really? this is a totally pointless javadoc, why do we even have it?

Jolyon-S (commented):
There's a couple of suspiciously sparse javadocs - do we really need all of them?

anjor:
going to start over and do this properly, actually.

PR #5572, opened 2021-07-26
Manually run the SafeLogger migration

carterkozak:
Excavator is timing out.

**Goals (and why)**:


carterkozak:
I reverted classes which didn't immediately work out of the box, and any class matching `.*Logger(s)?.java`, most of which expose public logger instances. Those may be handled separately.

j-baker (approved):

PR #5578, opened 2021-07-28
Exclusions 4: Slf4jThrowable

jeremyk-91:
**Goals (and why)**: Remove exclusions

**Implementation Description (bullets)**: Fix offending instances

- In cleaning up some of these I marginally changed some log lines, could this be worrying?


atlasdb-client/src/main/java/com/palantir/atlasdb/keyvalue/impl/TracingPrefsConfig.java
```diff
@@ -80,7 +80,7 @@ public void run() {
                 loadedConfig = false;
             }
         } catch (Throwable t) {
-            log.error("Error occurred while refreshing {}: {}", TRACING_PREF_FILENAME, t, t);
+            log.error("Error occurred while refreshing {}", TRACING_PREF_FILENAME, t);
```
Jolyon-S, line 83: You've already changed the logline - why not make this and the other one `unsafe` args?

PR #5581, opened 2021-07-31
Remove okhttp dependency

pkoenig10:


schlosna (approved):
Nice!

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -60,7 +59,10 @@ private ConjureJavaRuntimeTargetFactory() {
                 Optional.empty(),
                 trustContext.orElseThrow(() -> new IllegalStateException("CJR requires a trust context")));
         T client = JaxRsClient.create(
-                type, addAtlasDbRemotingAgent(parameters.userAgent()), HOST_METRICS_REGISTRY, clientConfiguration);
+                type,
+                addAtlasDbRemotingAgent(parameters.userAgent()),
+                NoOpHostEventsSink.INSTANCE,
```
Jolyon-S, line 64: My understanding is that this is fine because we never actually retrieve the metrics from `HOST_METRICS_REGISTRY`, correct?

Jolyon-S (approved):
This looks fine - just want your sanity check on my comment, but then good to go from my end.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
pkoenig10: Correct

PR #5582, opened 2021-08-02
remove background thread for tracing conf

berler:
**Goals (and why)**:

Currently all services that use atlas have a background thread which every minute checks on disk for a file for tracing prefs that only actually apply to cassandra. This moves this config to the actual KVS runtime config and removes this background thread entirely.

**Implementation Description (bullets)**: Remove unnecessary background thread and create an equivalent config structure in the cassandra kvs runtime config.


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
```diff
@@ -17,20 +17,23 @@
...
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 
 public class TracingQueryRunner {
     private final Logger log;
-    private final TracingPrefsConfig tracingPrefs;
+    private final Supplier<CassandraTracingConfig> tracingPrefs;
+    private final Random random = new Random();
```
schlosna, line 34:
```suggestion
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
```diff
@@ -75,7 +78,7 @@ public TracingQueryRunner(Logger log, TracingPrefsConfig tracingPrefs) {
 
     private boolean shouldTraceQuery(Set<TableReference> tableRefs) {
         for (TableReference tableRef : tableRefs) {
-            if (tracingPrefs.shouldTraceQuery(tableRef.getQualifiedName())) {
+            if (tracingPrefs.get().shouldTraceQuery(tableRef.getQualifiedName(), random)) {
```
schlosna, line 81: Can we just use `ThreadLocalRandom`?
```suggestion
            if (tracingPrefs.get().shouldTraceQuery(tableRef.getQualifiedName(), ThreadLocalRandom.current())) {
```

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+                return true;
+            } else {
+                if (random.nextDouble() <= traceProbability()) {
+                    return true;
+                }
+            }
+        }
+        if (tablesToTrace().isEmpty()) {
+            return true; // accept enabled = true but no tables specified to mean trace all tables
+        }
```
Jolyon-S, line 74: I know that this was the previous behaviour, but it's really... weird. If my read is correct, it's basically saying that if you leave the tables to trace as empty, you also want to ignore the tracing probability, which I'm fairly certain is not what was intended. It seems to me that the above contains check should probably have been:

```java
if (tablesToTrace().isEmpty() || tablesToTrace().contains(tablename)) {
```

which makes more sense. And given that users will be forced to change their config to their Cassandra KVS config anyway, I'd push in that direction (so long as this is clearly reflected in the docs).

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+        return false;
+    }
+
+    @Value.Check
+    protected void check() {
+        // This does not actually validate anything but mimics the old behavior where enabling tracing
+        // would log a warning.
+        if (enabled()) {
+            if (loggedWarning.compareAndSet(false, true)) {
```
Jolyon-S, line 83: Do we really need this atomic boolean to mimic the old behaviour? It made sense before because it was periodically reloading the file, but is this object periodically recreated?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
Jolyon-S: From my end, no strong preferences, although as noted I'd much rather we do the random code in here (and thus don't need to pass a `Random` around).

Jolyon-S (commented):
While I think this is definitely an improvement (in truth, I had no idea that such a bizarre piece of code existed in the first place), I'm a little concerned about what this means for those tracing in the wild. I suppose that anyone who does have tracing enabled will just no longer have it enabled. Is that the case anywhere? Can we validate this?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @JsonProperty("trace-probability")
+    @Value.Default
+    public double traceProbability() {
+        return 1.0;
+    }
+
+    @JsonProperty("tables-to-trace")
+    public abstract Set<String> tablesToTrace();
```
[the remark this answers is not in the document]
berler: Immutables implicitly does this already.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
berler: can we make changes to behavior in a separate change?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
berler: This code will trigger every time the runtime config is parsed, which by default in witchcraft is every 5 seconds. This will be a ton more spammy than before if we removed this.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
berler: I'm currently just preserving existing behavior where a single instance of Random was used for all threads. Can we make behavior changes separately?

docs/source/configuration/enabling_cassandra_tracing.rst
```diff
@@ -25,36 +24,51 @@ When a query is traced, you will see the following line your log file:
...
+     runtime:
+       atlas:
+         keyValueService:
+           type: cassandra
+           tracing:
+             enabled: true
+             trace-probability: 1.0
+             min-duration-to-log-ms: 0
+             tables-to-trace:
+               - "_transactions"
```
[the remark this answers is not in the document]
berler: Just an example, and I was trying to match the exact previous example.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @JsonProperty("min-duration-to-log-ms")
+    @Value.Default
+    public int minDurationToLogMs() {
+        return 0;
+    }
+
+    public boolean shouldTraceQuery(String tablename, Random random) {
```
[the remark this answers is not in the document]
berler: Done

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+    public double traceProbability() {
+        return 1.0;
+    }
+
+    @JsonProperty("tables-to-trace")
+    public abstract Set<String> tablesToTrace();
+
+    @JsonProperty("min-duration-to-log-ms")
+    @Value.Default
+    public int minDurationToLogMs() {
```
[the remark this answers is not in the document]
berler: Done

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
berler: Done

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
```diff
@@ -0,0 +1,94 @@
...
+    @JsonProperty("enabled")
+    @Value.Default
+    public boolean enabled() {
+        return false;
+    }
+
+    @JsonProperty("trace-probability")
+    @Value.Default
+    public double traceProbability() {
+        return 1.0;
```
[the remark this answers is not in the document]
berler: Added precondition check

berler:
> While I think this is definitely an improvement (in truth, I had no idea that such a bizarre piece of code existed in the first place), I'm a little concerned about what this means for those tracing in the wild. I suppose that anyone who does have tracing enabled will just no longer have it enabled. Is that the case anywhere? Can we validate this?

A full search internally shows that this is never enabled. None of our products will ever produce the current config file. I would be surprised if this was ever enabled anywhere in the last ~4 years or more.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
Jolyon-S: Interesting. I wasn't aware that it actually creates this object that frequently.

What happens if:
1. there are multiple transaction managers in the same JVM (and this is configured for some but not all)?
2. the tracer is enabled, disabled, then re-enabled?

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/TracingQueryRunner.java
```diff
@@ -74,9 +76,18 @@ public TracingQueryRunner(Logger log, TracingPrefsConfig tracingPrefs) {
...
+            return false;
+        }
+        if (prefs.tablesToTrace().isEmpty()
+                || tableRefs.stream().map(TableReference::getQualifiedName).anyMatch(prefs.tablesToTrace()::contains)) {
+            if (prefs.traceProbability() >= 1.0) {
                 return true;
+            } else {
+                if (ThreadLocalRandom.current().nextDouble() <= prefs.traceProbability()) {
+                    return true;
+                }
```
Jolyon-S, line 90: nit
```suggestion
            } else if (ThreadLocalRandom.current().nextDouble() <= prefs.traceProbability()) {
                    return true;
                }
```

Jolyon-S (commented):
One nit and then I think we're good. Thanks!

berler:
> One nit and then I think we're good. Thanks!
> 
> Edit: and an answer to the question posted above.

@Jolyon-S As I said above:

A full search internally shows that this is never enabled. None of our products will ever produce the current config file.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/cassandra/CassandraTracingConfig.java
berler: When there are multiple in the same JVM it will only log once.

If enabled, disabled, then re-enabled it still would have only logged the first time.

It will also log an additional time whenever the service restarts.

berler: I'm also not aware of any internal products that would create multiple Cassandra KVSes in the same JVM outside of testing.

berler: Also looked at the witchcraft code again, it actually defaults to reparse config every 1 second.

Jolyon-S:
> > One nit and then I think we're good. Thanks!
> > Edit: and an answer to the question posted above.
> 
> @Jolyon-S As I said above:
> 
> A full search internally shows that this is never enabled. None of our products will ever produce the current config file.

I was referring to the one about multiple instances in the same JVM. There _is_ one such service that can have >80 transaction managers in the same JVM.

Jolyon-S (approved):

PR #5584, opened 2021-08-05
AtlasDb "dynamic snitch" -- do not route to very slow hosts 

Sam-Kramer:
**Goals (and why)**:
Remove hosts from host list if they're 10x slower than every other node. This is useful for SND/AZ outages, as the client will purposefully not route to a slow coordinator node. 

- Track all (successful) requests in an **ExponentiallyDecayingReservoir**, which is associated with a single host 
  - Entries have a maximum life of 5 minutes
- When fetching hosts, determine the average p99 latency for all hosts, and compare it against each of our hosts
- If any p99 latency is N times greater (recommendation: 10) than the _average excluding itself_ , then we remove the host from the host list 
- All hosts that have not been queried (value = 0) have been assigned a default value of 1; this is done in-order to ensure we query all hosts at least once every 5 minutes
- In the case of coordinator SND in a large cluster (9 nodes, 3 nodes in AZs A/B/C), we still prefer to route to AZ local hosts instead, unless they are as well N times slower.  

- Unit tests
- Will try on IL? Maybe Oregano. 

- Overhead per call is now higher; probably worth memoizing this
- Code change could have refactored and done a more proper "generic" pattern for creating weighting patterns
- This will increase costs ($$$) for cross-AZ traffic, even if there is no degradation. We've to query all hosts in order to determine our latencies. However, this cost _should_ be small, as it should be a small number of requests every 5 minutes (default window for ExponentiallyDecayingReservoir). 

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -15,7 +15,7 @@
  */
 package com.palantir.atlasdb.keyvalue.cassandra.pool;
 
-import com.codahale.metrics.Histogram;
+import com.codahale.metrics.ExponentiallyDecayingReservoir;
 import com.codahale.metrics.Snapshot;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
@@ -320,12 +320,12 @@ public class CassandraService implements AutoCloseable {
         Map<InetSocketAddress, Snapshot> latencies = StreamEx.of(desiredHosts)
                 .mapToEntry(currentPools::get)
                 .mapValues(CassandraClientPoolingContainer::getLatency)
-                .mapValues(Histogram::getSnapshot)
+                .mapValues(ExponentiallyDecayingReservoir::getSnapshot)
                 .toMap();
 
         OptionalDouble averageLatency = EntryStream.of(latencies)
                 .values()
-                .mapToDouble(Snapshot::get99thPercentile)
+                .mapToDouble(Snapshot::getMedian)
                 .average();
         if (!averageLatency.isPresent()) {
             return desiredHosts;
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
@@ -56,7 +56,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.OptionalDouble;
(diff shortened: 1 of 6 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java
```diff
@@ -310,9 +314,51 @@ private String getRingViewDescription() {
         return hosts;
     }
 
+    private static double getAverageWithoutSelf(double sum, double self, int count) {
+        return (sum - self) / (count - 1);
+    }
+
+    @VisibleForTesting
+    Set<InetSocketAddress> maybeFilterSlowHosts(Set<InetSocketAddress> desiredHosts) {
```
leonz, line 322: I think that by adding another way to "weight" hosts, we've made it harder to reason about the overall hostselection.  What happens if the slowest host is the only available host?  Especially in combination with the weighted local filtering?  Combining _this_, "local hosts", _and_ active connections, I'm curious if we come up with a combination that is actually worse overall.

Active connections creates a `WeightedHosts` pool - is that something we can potentially leverage here to provide calling functions more options?

There's probably some re-factoring here where at the top level we pass in a list of hosts, and a list of "weighting functions", and our system reasonably attempts to give a "best overall result" when the weighting functions may conflict.

Sam-Kramer: `What happens if the slowest host is the only available host?` I've pushed a test case which proves that we do not filter out the host, and we just select it. This makes sense, as the average is calculated without itself, so if its n = 1, the slowness indicator will always be 0, which will always pass. 

Yeah I agree, there is a refactor here which consolidates this logic much more clearly. One argument for my reason why I pushed for _this_ rather than doing a refactor, is that it is stupid simple. In the sense that, the slow hosts are essentially being filtered just like the exclude list logic (i.e. exceptions). So the filter tree goes: Exclude list > slow hosts > Active connections

I also want to highlight the subtle difference here in that, active connections is arguably _preferred_ during execution, while slow hosts are _required_ to be removed (to use Kubernete's terms). Local host filtering is also required as well. Of course, unless the slow host is the only that's left, then we've no choice. 

So the weighting logic comes a bit more complex, as it needs to be some factor of `locality + # of connections + p99`. Which then, we should really test aggressively, as it will be a major change.

Sam-Kramer: Another way to see the PR, is that we're simply handling the extreme edge case that `# of active connections` fails to consider, which is that the host may have actual degradation, which is only detectable via latency.

Sam-Kramer: Made a prettier version here: https://github.com/palantir/atlasdb/pull/5586

PR #5585, opened 2021-08-05
Use SafeLoggable.getLogMessage() when available during rewrap

henryptung:
Don't risk leaking unsafe information from getMessage() into a
SafeLoggable constructor (which is likely declared @CompileTimeConstant).

### DIFF
--- a/atlasdb-commons/src/test/java/com/palantir/common/base/ThrowablesTest.java
+++ b/atlasdb-commons/src/test/java/com/palantir/common/base/ThrowablesTest.java
@@ -19,9 +19,17 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.fail;
 
+import com.google.errorprone.annotations.CompileTimeConstant;
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.exceptions.SafeExceptions;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import java.io.IOException;
 import java.sql.SQLException;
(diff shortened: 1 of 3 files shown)


pustaczek:
This code path seemingly doesn't end up being hit in the AtlasDB codebase, because if [constructors don't fall into this pattern](https://github.com/palantir/atlasdb/blob/d35ec651cd6150e814661e027285bc908b63b837/atlasdb-commons/src/main/java/com/palantir/common/base/Throwables.java#L207-L215), the rewrap gives up and returns the original exception. No safe exceptions I could find have such constructors, though the change would still make sense just out of caution. Is this intended?

henryptung:
Yep, mostly just a safety fix for downstream consumers using the Throwables lib (and especially if they implement their own SafeLoggable exceptions, which may hit the relevant constructor signatures).

atlasdb-commons/src/test/java/com/palantir/common/base/ThrowablesTest.java
```diff
@@ -90,6 +98,25 @@ public void testThrowCauseAsUnchecked() {
...
+        assertThat(rewrapped1.getLogMessage()).doesNotContain("secret");
+
+        SafeRewrappableException rewrapped2 = null;
+        try {
+            Throwables.rewrapAndThrowIfInstance(rewrapped1, SafeRewrappableException.class);
+        } catch (SafeRewrappableException ex) {
+            rewrapped2 = ex;
+        }
+        assertThat(rewrapped2).isNotNull();
+        assertThat(rewrapped2.getLogMessage()).doesNotContain("secret");
```
jeremyk-91, line 117: For this section, it probably suffices to use `assertThatThrownBy(...)`?

jeremyk-91 (commented):
Thanks for adding the test. I think it can be written a bit more simply/clearly, but otherwise looks good.

jeremyk-91 (approved):
👍

PR #5587, opened 2021-08-06
Fix Runaway Retries When HikariConnectionPool is Closed

bmarcaur:
**Goals (and why)**:
Some of our testing infrastructure for an internal project was experiencing a considerable amount of INFO logs due to `RetriableTransactions` quickly spinning when the underlying connection was closed. We are working to track down why the connection pool was closed, but the `AsyncPuncher` was not closed. However, since a closed connection pool throws a SQLException the `RetriableTransactions` believes that it needs to retry despite this scenario never recovering. This produced logs in excess of 5GB for a single test run.

### DIFF
--- a/commons-db/src/main/java/com/palantir/nexus/db/pool/RetriableTransactions.java
+++ b/commons-db/src/main/java/com/palantir/nexus/db/pool/RetriableTransactions.java
@@ -172,7 +172,7 @@ public final class RetriableTransactions {
                                     e);
                         }
                         if (cm.isClosed()) {
-                            log.warn("Aborting tx retry, underlying connection is closed", e);
+                            log.error("Aborting transaction retry, underlying connection manager is closed", e);
                             return TransactionResult.failure(e);
                         }
                         if (shouldStillRetry(startTimeMs, attemptTimeMs)) {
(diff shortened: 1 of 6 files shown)


commons-db/src/main/java/com/palantir/nexus/db/pool/RetriableTransactions.java
```diff
@@ -171,6 +171,10 @@ public Throwable getError() {
                                     now,
                                     e);
                         }
+                        if (cm.isClosed()) {
+                            log.warn("Aborting tx retry, underlying connection is closed", e);
```
schlosna, line 175: Nit as message should likely be an error and clarify the connection manager itself is closed

```suggestion
                            log.error("Aborting transaction retry, underlying connection manager is closed", e);
```

schlosna (commented):
Overall looks good, one nit on the logging

bmarcaur:
Thanks for the review @schlosna

schlosna:
Hmm, something is up with CircleCI not running, may need to poke at that on Monday

sudiksha27 (approved):
LGTM!

schlosna (approved):

jeremyk-91 (approved):

PR #5599, opened 2021-08-16
[Correctness Logging] Part 1: Explicit Timestamp Violation Metrics

jeremyk-91:
**Goals (and why)**:
- Re-express timestamp violation metrics in a form where automation can pick up
- Ensure that this is cost-effective.

- Add a callback when a timestamp violation is detected that increments a counter metric
- Invoke the callback when said violation is detected

- Added specific tests around callback behaviour

- If a violation is detected and we then quickly reset the AtlasDB-using service, we might not see it. I don't know if this is avoidable though.
- I don't really like the passing in of an Optional namespace, but this handles the general case with timelock or namespacing more easily. (This information will be very useful for services that use multiple transaction managers, such as AtlasDB Proxy.)

**Where should we start reviewing?**: TimestampCorroboratingTimelockService

### DIFF
--- a/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
+++ b/atlasdb-config/src/main/java/com/palantir/atlasdb/factory/TransactionManagers.java
@@ -1036,21 +1036,17 @@ public abstract class TransactionManagers {
                 timelockRequestBatcherProviders,
                 schemas);
         return withMetrics(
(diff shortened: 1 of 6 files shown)


sudiksha27 (approved):
Neat!
I find the concerns valid but I cannot think of a better way around it.

jeremyk-91:
Thanks @sudiksha27 !

PR #5600, opened 2021-08-17
[PDS-205737] Invalidation Train

jeremyk-91:
**Goals (and why)**:
- Backup and restore workflows need to account for the possibility that, during a restore, a running service will read data that has been written before the clean-transactions-range operation has run (which is uncommitted). Typically this is done by locking down KVS access. However, this is clearly insufficient where lock watches may be used.
- Note that this currently isn't an issue just yet because validation of the Atlas cache is at 100 percent, but it's best we have this in place now, rather than find ourselves in corruption land later on.

- Add a new endpoint that removes service implementations from timelock's tracking map. Future user requests will re-create these, but they have a different leadership ID, different lock watch ID, and all locks will be invalidated through this.

Added fairly standard tests for the new endpoint and all associated functionality.

- Is there some bizarre race condition I might have missed?
- Is there a resource leak I might have missed?

**Where should we start reviewing?**: TimelockNamespaces

@leonz / @j-baker for SA


timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimelockNamespacesTest.java
```diff
@@ -154,6 +155,29 @@ public void pathsForTimelockAndLockWatchServicesAreNotValid() {
...
+                .thenReturn(mock(TimeLockServices.class));
+
+        String client = uniqueClient();
+        TimeLockServices services = namespaces.get(client);
+        namespaces.invalidateResourcesForClient(client);
+        verify(services).close();
+
+        services = namespaces.get(client);
+        namespaces.invalidateResourcesForClient(client);
+        verify(services).close();
```
sudiksha27, line 172: can we add an assertion that 2 service objects were different?

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimelockNamespacesTest.java
```diff
@@ -154,6 +155,29 @@ public void pathsForTimelockAndLockWatchServicesAreNotValid() {
...
+    }
+
+    @Test
+    public void handlesInvalidationOfNonexistentClients() {
+        assertThatCode(() -> namespaces.invalidateResourcesForClient("somethingUnknown"))
+                .doesNotThrowAnyException();
```
sudiksha27, line 178: 💯

sudiksha27 (approved):
Solid PR - I like the final solution! Tests are good - cannot think of a race condition we may be missing so LGTM!

timelock-impl/src/test/java/com/palantir/atlasdb/timelock/TimelockNamespacesTest.java
jeremyk-91: Yep, will do.

PR #5602, opened 2021-08-18
PDS-202988

sudiksha27:
**Goals (and why)**:
Make the Sqlite connection checkout timeout configurable. This config is used by TL to connect to Sqlite on local disk.

Before end of week is ideal


jeremyk-91 (approved):
👍

PR #5605, opened 2021-08-20
Fix codegen of reusable persisters for dynamic column values

pustaczek:
**Goals (and why)**: Fix compilation errors in generated code related to a missing static identifier, when the user tries to create a schema with dynamic columns, and the value type is specified as a reusable persister.

- There are two code generation context which need code generating a reusable persister, NamedColumnValueRenderer and DynamicColumnValueRenderer.
- Generating the declaration of the persister variable has been extracted into a getInstantiateReusablePersisterCode method in the ColumnValueDescription class.
- Unfortunately, this declaration goes either into a definition of an anonymous class implementing Hydrator (named columns case), or directly into a definition of the value class (dynamic columns case).
- The first case doesn't require static because the anonymous class definition is written into a static itself, but the second case requires static because the variable is used from a static method. Also, the first case doesn't allow static, because Java 11 does not allow statics in anonymous classes.

**Testing (What was existing testing like?  What have you done to improve it?)**: There exist both snapshot tests and tests which check whether the generated code contains certain substrings, but they do not cover the combination of dynamic columns (1 of 2) and reusable persisters (1 of 5). I added the second kind of test, to match an existing test for named columns with reusable persisters.

**Concerns (what feedback would you like?)**: It would be nice to share more code between the named and the dynamic case, as well as use snapshot tests more, though this would probably involve slightly bigger changes.

**Priority (whenever / two weeks / yesterday)**: Whenever.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ColumnValueDescription.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/table/description/ColumnValueDescription.java
@@ -349,8 +349,9 @@ public final class ColumnValueDescription {
                 + ", com.palantir.atlasdb.table.description.ColumnValueDescription.Compression." + compression + ")";
     }
 
-    public String getInstantiateReusablePersisterCode() {
-        return "private final " + canonicalClassName + " REUSABLE_PERSISTER = " + "new " + canonicalClassName + "();";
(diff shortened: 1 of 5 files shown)


jeremyk-91 (approved):
👍

PR #5612, opened 2021-08-26
Exclusion Hunter II: ArrayAsKeyOfSetOrMap

sudiksha27:
**Goals (and why)**:
Hunt ArrayAsKeyOfSetOrMap exclusion - `Arrays do not override equals() or hashCode, so comparisons will be done on reference equality only. If neither deduplication nor lookup are needed, consider using a List instead. Otherwise, use IdentityHashMap/Set, a Map from a library that handles object arrays, or an Iterable/List of pairs.`

Changed Sets and Maps keyed with byte arrays to IdentityHashMap/ Set.

Within a week is great.


jeremyk-91 (approved):
well done, you have proven me wrong about this one 💯

PR #5617, opened 2021-08-31
Hunt exclusions V

sudiksha27:
:(

Before end of week is ideal

### DIFF
--- a/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
+++ b/leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
@@ -55,7 +55,7 @@ public class PaxosStateLogImpl<V extends Persistable & Versionable> implements P
     private static final String TMP_FILE_SUFFIX = ".tmp";
     private static final SafeLogger log = SafeLoggerFactory.get(PaxosStateLogImpl.class);
 
-    private static boolean nameIsALongPredicate(File file) {
+    private static boolean checkNameIsALong(File file) {
         if (file == null) {
             return false;
         }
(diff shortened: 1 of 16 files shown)


leader-election-impl/src/main/java/com/palantir/paxos/PaxosStateLogImpl.java
```diff
@@ -56,18 +55,16 @@
...
+    private static boolean checkNameIsALong(File file) {
+        if (file == null) {
+            return false;
+        }
+        try {
+            getSeqFromFilename(file);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
```
jeremyk-91, line 67: I realise the transform is mechanical, and this is correct, but this still gives me chills in a bad way 😅

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TimeLockServerHolder.java
```diff
@@ -168,4 +159,13 @@ TimeLockInstallConfiguration installConfig() {
             throw new RuntimeException(e);
         }
     }
+
+    private static WireMockConfiguration wireMockConfiguration(int port) {
```
jeremyk-91, line 163: Here's a fun one: I think this should take an `Integer` as opposed to an `int` (why?)

jeremyk-91 (approved):
👍

timelock-server/src/testCommon/java/com/palantir/atlasdb/timelock/TimeLockServerHolder.java
sudiksha27: because httpsport takes an `Integer`?

jeremyk-91 (approved):

PR #5618, opened 2021-08-31
Fix wrong write lock status reported by the getLockState

pustaczek:
**Goals (and why)**: Fix wrong write lock status reported by the getLockState debugging endpoint.

- The old implementation assumed a lock is write-locked when it exists in cache, but has no read holders.
- This can be incorrect when a lock is both read-locked and write-locked due to reentrancy, which [can happen](https://github.com/palantir/atlasdb/blob/4ec93785191c82dbc3aca27f0145e137cf485a51/lock-impl/src/test/java/com/palantir/lock/LockServiceTest.java#L1250-L1253).
- This can also be incorrect when a lock was recently unlocked, but wasn't evicted from [a weakValues cache](https://github.com/palantir/atlasdb/blob/4ec93785191c82dbc3aca27f0145e137cf485a51/lock-impl/src/main/java/com/palantir/lock/impl/LockServiceImpl.java#L198-L205) yet.
- New implementation get the owner of the write lock directly from the underlying lock implementation.

**Testing (What was existing testing like?  What have you done to improve it?)**: This endpoint is not tested very thoroughly, as it is a debugging endpoint. I added two tests for the specific issues (they fail without the first commit), but I haven't improved the testing situation more than that.

**Concerns (what feedback would you like?)**: Tests could be made more thorough.

**Priority (whenever / two weeks / yesterday)**: Whenever.

### DIFF
--- a/lock-impl/src/main/java/com/palantir/lock/impl/LockServerSync.java
+++ b/lock-impl/src/main/java/com/palantir/lock/impl/LockServerSync.java
@@ -21,6 +21,7 @@ import com.palantir.logsafe.Preconditions;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.locks.AbstractQueuedSynchronizer;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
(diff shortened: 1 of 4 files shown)


lock-impl/src/main/java/com/palantir/lock/impl/LockServerSync.java
```diff
@@ -267,4 +268,12 @@ private synchronized void decrementReadCount(int clientIndex) {
         return Collections.unmodifiableList(
                 readLockHolders.keysView().collect(clients::fromIndex, new ArrayList<>(readLockHolders.size())));
     }
+
+    synchronized Optional<LockClient> getWriteClient() {
+        if (getState() > 0) {
+            return Optional.of(clients.fromIndex(writeLockHolder));
+        } else {
+            return Optional.empty();
+        }
```
jeremyk-91, line 277: extreme nit-pickyness: you'll notice that the style here is generally not to have the else clause after a return, so that should be followed (I know this is different in AtlasDB Proxy)

jeremyk-91 (approved):
I think this looks good! One minor point

PR #5620, opened 2021-09-02
[PDS-202945] Only load 1 column from the stream store when checking index validity

jeremyk-91:
**Goals (and why)**:
- Reduce load on Cassandra when sweeping tables that have highly referenced streams.

- Use a getRowsColumnRange to retrieve at most one cell per row to check whether a stream is referenced.

- Externally validated across the fleet
- There are some existing tests (e.g. TargetedSweepEteTest).

Did I accidentally break something? I verified that the erroneous log message was _never_ printed across the fleet for the past 10 days.

**Where should we start reviewing?**: `StreamTableRenderer`. The rest is generated code.


gmaretic (approved):
I forgot this was already a thing! It looks good to me

PR #5624, opened 2021-09-06
[k8s TL] | Runtime clusterConfig

sudiksha27:
**Goals (and why)**:
Move ClusterConfiguration from install to runtime. This is required for setting TimeLock service in the Rubix world. Note that the config will still not be live re-loaded.

Move ClusterConfiguration from TimeLockInstallConfig to TimeLockRuntimeConfig. 

The clusterConfig with `enableNonstandardAndPossiblyDangerousTopology` is still part of install config as this config has been overridden in several places.

Modified existing tests.

1. Initializing TimeLockAgent with a snapshot of clusterConfiguration right now. Is this reasonable? If not, at what point can we reliably take this snapshot?
2. Did I break something accidentally?

TimeLockInstallConfiguration/ TimeLockAgent

Before EOD 7 September

### DIFF
--- a/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
@@ -32,10 +32,7 @@ import org.immutables.value.Value;
 @Value.Immutable
 @JsonIgnoreProperties(value = "asyncLock")
 public interface TimeLockInstallConfiguration {
-
     PaxosInstallConfiguration paxos();
-
-    ClusterConfiguration cluster();
 
     @Value.Default
     default boolean iAmOnThePersistenceTeamAndKnowWhatImDoingSkipSqliteConsistencyCheckAndTruncateFileBasedLog() {
@@ -61,9 +58,8 @@ public interface TimeLockInstallConfiguration {
     }
 
     @Value.Derived
-    default boolean isNewServiceNode() {
-        return paxos().isNewService()
-                || cluster().knownNewServers().contains(cluster().localServer());
+    default boolean isNewService() {
+        return paxos().isNewService();
     }
 
     static Builder builder() {
--- a/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
@@ -36,7 +36,11 @@ public interface TimeLockInstallConfiguration {
     PaxosInstallConfiguration paxos();
 
     @Deprecated
-    ClusterInstallConfiguration cluster();
+    @Value.Default
+    default ClusterInstallConfiguration cluster() {
+        return ImmutableClusterInstallConfiguration.builder().build();
+    }
+    ;
 
     @Value.Default
     default boolean iAmOnThePersistenceTeamAndKnowWhatImDoingSkipSqliteConsistencyCheckAndTruncateFileBasedLog() {
--- a/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
+++ b/timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
@@ -32,7 +32,11 @@ import org.immutables.value.Value;
 @Value.Immutable
 @JsonIgnoreProperties(value = "asyncLock")
 public interface TimeLockInstallConfiguration {
+
     PaxosInstallConfiguration paxos();
+
+    @Deprecated
+    ClusterInstallConfiguration cluster();
 
     @Value.Default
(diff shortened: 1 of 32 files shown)


timelock-agent/src/main/java/com/palantir/timelock/config/TimeLockInstallConfiguration.java
```diff
@@ -35,7 +35,11 @@
 
     PaxosInstallConfiguration paxos();
 
-    ClusterConfiguration cluster();
+    @Deprecated
```
jeremyk-91, line 38: Normally we tell people what to do instead when marking something as deprecated.

timelock-agent/src/test/java/com/palantir/timelock/config/TimeLockRuntimeConfigurationTest.java
```diff
@@ -15,29 +15,60 @@
...
 
+import com.google.common.collect.ImmutableList;
+import com.palantir.conjure.java.api.config.service.PartialServiceConfiguration;
+import java.util.Optional;
 import org.junit.Test;
 
 @SuppressWarnings("CheckReturnValue")
 public class TimeLockRuntimeConfigurationTest {
+    private static final String SERVER_A = "horses-for-courses:1234";
+    public static final String SERVER_B = "paddock-and-chips:2345";
```
jeremyk-91, line 29: nit: `private`

timelock-agent/src/test/java/com/palantir/timelock/config/ClusterConfigurationTest.java
```diff
@@ -42,11 +43,14 @@ public void shouldThrowIfLocalServerNotInServers() {
         assertThatIllegalArgumentException().isThrownBy(builder::build);
     }
 
+    // Todo(snanda): Remove ignore when ClusterInstallConfig is killed
```
jeremyk-91, line 46: nit: I think these are normally `TODO` (all caps).

Also, would prefer if we explain that the check happens at a different level (to not give the impression that this is ignored because we are allowing such configurations without the disclaimer).

timelock-agent/src/test/java/com/palantir/timelock/paxos/TimeLockAgentTest.java
```diff
@@ -229,10 +237,11 @@ private File getMockFileWith(boolean isDirectory, boolean canCreateDirectory) th
...
-                        .paxos(installConfiguration)
-                        .build()))
+        assertThatThrownBy(() -> TimeLockAgent.verifyIsNewServiceInvariant(
+                        TimeLockInstallConfiguration.builder()
+                                .paxos(installConfiguration)
+                                .build(),
+                        CLUSTER_CONFIG))
                 .isInstanceOf(IllegalArgumentException.class);
     }
 }
```
jeremyk-91, line 247: I might have missed this, but somewhere here we probably want a test that verifies that the old mechanism works (i.e. a test where install configuration says to allow a nonstandard topology but the cluster config is 1 or 2 nodes and does not - this should not fail).

timelock-agent/src/main/java/com/palantir/timelock/paxos/TimeLockAgent.java
```diff
@@ -128,20 +130,22 @@ public static TimeLockAgent create(
             MetricsManager metricsManager,
             TimeLockInstallConfiguration install,
             Refreshable<TimeLockRuntimeConfiguration> runtime,
```
jeremyk-91, line 132: It's a bit of a shame this does _also_ contain ClusterConfigurations that may be wrong. I'm thinking as a defense mechanism, should we map any user-provided runtime config into a layer that rejects any attempts to access the ClusterConfiguration through this Refreshable?

jeremyk-91 (commented):
Broadly seems correct; I'm not entirely happy with some of the changes / the config is a little more complex than I'd like it to be, but I don't really have a better solution.

timelock-agent/src/test/java/com/palantir/timelock/paxos/TimeLockAgentTest.java
```diff
@@ -172,6 +174,52 @@ public void newServiceSetNoDataDirectoryExistsDoesNotThrowWhenIgnoreFlagSet() {
...
+        assertThatThrownBy(() -> TimeLockAgent.verifyTopologyOffersHighAvailability(
+                        TimeLockInstallConfiguration.builder()
+                                .paxos(createPaxosInstall(false, true, false))
+                                .build(),
+                        dangerousTopologyConfig))
+                .isInstanceOf(SafeIllegalArgumentException.class);
+    }
+
+    @Test
+    public void doNotIfStartedWithLessThanThreeServers_DangerousTopologyEnabledInInstallConfig() {
```
jeremyk-91, line 192: nit: let's just name this and the next one `doesNotThrow...` (in isolation it's not super obvious what's being tested I think)

jeremyk-91 (approved):

PR #5625, opened 2021-09-07
Increase default max fetch size for AgnosticLightResultSetImpl

jrfitzsimmons:
**Goals (and why)**:

The default was lowered back in 2009 and hadn't been revisited since. PG-139709 increased the default fetch size for other connections but AgnosticLightResultSetImpl overrides that on the result set. On PDS-210269 we observed a drop from 5 minutes to 13 seconds of fetching overhead when iterating over a result set with millions of rows.

Two weeks.

### DIFF
--- a/changelog/@unreleased/agnostic-light-result-set-fetch-size.v2.yml
+++ b/changelog/@unreleased/agnostic-light-result-set-fetch-size.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Increase default max fetch size for AgnosticLightResultSetImpl
+  links:
+    - https://github.com/palantir/atlasdb/pull/5625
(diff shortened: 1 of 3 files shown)


CVdV-au:
The original change @afchang made was in pgdev 496cad6707dab2ed5d452321ff79e9247f075904 in 2009.
In Oracle 12, JDBC memory handling was significantly improved https://www.oracle.com/technetwork/database/application-development/jdbc-memory-management-12c-1964666.pdf, and less and less non-Atlas code actually uses these classes as JOOQ permeates the PG code base. Atlas code already overrides the default in most pathways, so overall this is a low risk change IMHO.

jrfitzsimmons:
Closed in favour of https://github.com/palantir/atlasdb/pull/5626

PR #5628, opened 2021-09-07
Leaky sweep

gmaretic:
**Goals (and why)**:
Be able to sweep away sentinels from tables that were conservatively swept and are now thoroughly swept without destroying Cassandra.

The issue that we want to avoid is having 100k+ consecutive tombstones, this adds a new endpoint to the sweep endpoint which adds a 1% chance of skipping sweeping any one entry. This is correct because:

1) it is only allowed on thoroughly swept tables (hence no read only transactions would see an old version)
2) if any version is skipped, we keep the freshest version so that old writes do not come back to life

Added a bunch of tests

Thought carefully about this, but at first missed point 2) above -- have I missed anything else?

Tests?

This week?

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepTaskRunner.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepTaskRunner.java
@@ -51,6 +51,7 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 import java.util.function.LongSupplier;
+import java.util.stream.Collectors;
 
 /**
  * Sweeps one individual table.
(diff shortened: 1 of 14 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SweepTaskRunner.java
```diff
@@ -254,31 +273,70 @@ private int sweepBatch(TableReference tableRef, List<CellToSweep> batch, RunType
...
-                    if (runType == RunType.FULL) {
+                    safeToDeleteLast = safeToDeleteLast
+                            && addCurrentCellTimestamps(
+                                    currentBatch,
+                                    cell.cell(),
+                                    currentCellTimestamps.subList(0, numberOfTimestampsForThisBatch),
+                                    runType);
+
+                    if (runType != RunType.DRY) {
+                        if (!safeToDeleteLast && numberOfTimestampsForThisBatch == currentCellTimestamps.size()) {
```
jeremyk-91, line 294: can we have a comment on why the second clause is actually a safe thing to do?

jeremyk-91 (approved):
Talked through this, convinced it's correct. The background sweep code is pretty rough, but it's not frequently touched and the new behaviour is well-tested.

jeremyk-91 (approved):

PR #5629, opened 2021-09-07
Don't Overexpose Codegen Dependencies

bmarcaur:
**Goals (and why)**:
This is causing pain in downstream consumers of the atlas-client and processors as shading Jackson transitively can lead to duplicate classes on the classpath.

### DIFF
--- a/atlasdb-client/build.gradle
+++ b/atlasdb-client/build.gradle
@@ -66,12 +66,14 @@ dependencies {
     exclude (group: 'io.dropwizard.metrics', module: 'metrics-core')
     exclude (group: 'org.hdrhistogram', module: 'HdrHistogram')
   }
-  compile group: 'com.squareup', name:'javapoet'
+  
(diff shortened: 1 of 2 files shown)


atlasdb-client/build.gradle
```diff
@@ -72,7 +72,7 @@ dependencies {
   compile group: 'com.palantir.common', name: 'streams'
   compile group: 'com.github.rholder', name: 'guava-retrying'
   compile group: 'org.rocksdb', name: 'rocksdbjni'
-  compile group: 'com.palantir.goethe', name: 'goethe'
+  implementation group: 'com.palantir.goethe', name: 'goethe'
```
carterkozak, line 75: We might as well move javapoet to `implementation` while we're here

bmarcaur: Done

carterkozak (approved):
Thanks! lgtm

Jolyon-S (approved):

PR #5630, opened 2021-09-09
Track open transactions

Jolyon-S:
**Goals (and why)**:
We don't have a good way of knowing how many open transactions we have. This not only helps to solve some internal issues, but this may well tell us if we are leaking transaction state over time (if transaction count increases over time, this means we're missing a place to decrement, which in turn likely means we are missing some places to remove lock watch state, for instance).

Add a counter to `SnapshotTransactionManager` that increments when transactions are started, and decrements when they finish. 

None

Reasonable?

Diff

Today, for internal issues please.

### DIFF
--- a/changelog/@unreleased/pr-5630.v2.yml
+++ b/changelog/@unreleased/pr-5630.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: The number of open transactions are now tracked in a metric, `openTransactionCounter`.
+  links:
+  - https://github.com/palantir/atlasdb/pull/5630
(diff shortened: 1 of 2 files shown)


changelog/@unreleased/pr-5630.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: The number of open transactions are now tracked in a metric, `openTransactionCounter`.
```
jeremyk-91: nit: the origin is a part of the metric too

jeremyk-91 (approved):

PR #5631, opened 2021-09-13
Add metrics for cache instance count

pustaczek:
**Goals (and why)**: Start tracking the count of existing caches in metrics, to help diagnosing memory leaks such as PDS-185998 and (currently) PDS-209612.

- Simple gauge which tracks the map size.

**Testing (What was existing testing like?  What have you done to improve it?)**: Metrics are contained in a CacheMetrics class, which is usually mocked and receives some simple testing. I added a minimal assertion to one of the existing tests.

**Concerns (what feedback would you like?)**: A single CacheMetrics instance seems to be exclusive to a single CacheStoreImpl instance, but metric names are strings registered in MetricsManager - is it certain individual gauges will be tagged appropriately?

**Priority (whenever / two weeks / yesterday)**: Two weeks.

### DIFF
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -106,5 +106,5 @@ public final class AtlasDbMetricNames {
     public static final String LW_CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
     public static final String LW_EVENT_CACHE_FALLBACK_COUNT = "lockWatchEventCacheFallbackCount";
     public static final String LW_VALUE_CACHE_FALLBACK_COUNT = "lockWatchValueCacheFallbackCount";
-    public static final String LW_CACHE_INSTANCE_COUNT = "lockWatchCacheInstanceCount";
+    public static final String LW_TRANSACTION_CACHE_INSTANCE_COUNT = "lockWatchTransactionCacheInstanceCount";
 }
--- a/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
+++ b/atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
@@ -106,4 +106,5 @@ public final class AtlasDbMetricNames {
     public static final String LW_CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
     public static final String LW_EVENT_CACHE_FALLBACK_COUNT = "lockWatchEventCacheFallbackCount";
     public static final String LW_VALUE_CACHE_FALLBACK_COUNT = "lockWatchValueCacheFallbackCount";
+    public static final String LW_CACHE_INSTANCE_COUNT = "lockWatchCacheInstanceCount";
 }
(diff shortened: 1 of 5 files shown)


atlasdb-client/src/main/java/com/palantir/atlasdb/AtlasDbMetricNames.java
```diff
@@ -106,4 +106,5 @@ private CellFilterMetrics() {
     public static final String LW_CACHE_RATIO_USED = "lockWatchCacheRatioUsed";
     public static final String LW_EVENT_CACHE_FALLBACK_COUNT = "lockWatchEventCacheFallbackCount";
     public static final String LW_VALUE_CACHE_FALLBACK_COUNT = "lockWatchValueCacheFallbackCount";
+    public static final String LW_CACHE_INSTANCE_COUNT = "lockWatchCacheInstanceCount";
```
Jolyon-S, line 109: Which cache is this? I'd suggest `lockWatchTransactionCacheInstanceCount`

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
```diff
@@ -112,4 +113,9 @@ public void setMaximumCacheSize(long maximumCacheSize) {
                 AtlasDbMetricNames.LW_CACHE_RATIO_USED,
                 () -> () -> cacheSize.getCount() / (double) maximumCacheSize);
     }
+
+    public void setCacheInstanceCountGauge(Gauge<Integer> getCacheMapCount) {
```
Jolyon-S, line 117: again, rename perhaps to `setTransactionCacheInstanceCountGauge`

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImplTest.java
```diff
@@ -66,6 +70,10 @@ public void multipleCallsToGetReturnsTheSameCache() {
         TransactionScopedCache cache2 = cacheStore.getCache(TIMESTAMP_2);
 
         assertThat(cacheStore.getCache(TIMESTAMP_1)).isEqualTo(cache1).isNotEqualTo(cache2);
+
+        ArgumentCaptor<Gauge<Integer>> cacheInstanceCount = ArgumentCaptor.forClass(Gauge.class);
+        verify(metrics).setCacheInstanceCountGauge(cacheInstanceCount.capture());
+        assertThat(cacheInstanceCount.getValue()).extracting(Gauge::getValue).isEqualTo(2);
```
Jolyon-S, line 76: Can you extract out a method to do this, then add this check throughout all of these tests? It seems like a nice idea, to be able to confirm how many caches we have stored in memory.

changelog/@unreleased/pr-5631.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add metrics for cache instance count
```
Jolyon-S, line 3: This is too sparse

Jolyon-S (approved):

PR #5632, opened 2021-09-13
[PDS-210062] Fix SnapshotTransaction/Delete Executor Resource Leak

jeremyk-91:
**Goals (and why)**:
- See PDS-210062
- Resource leaks are bad, we don't want them.

- #4656 was an attempt at reducing the memory in the delete executor's work queue. It wasn't a complete fix: 

```
deleteExecutor.execute(() -> deleteCells(keyValueService, tableRef, keysToDelete));
```

still reads the value of `keyValueService` _at the time the lambda is called_, and so that must maintain a reference to `this`. It would suffice to read it first

```
KeyValueService theKvs = keyValueService;
deleteExecutor.execute(() -> deleteCells(theKvs, tableRef, keysToDelete));
```

- But this PR adds a bit more abstraction so it's harder to misuse the delete executor.
- Possibly worthwhile reading for a broader coverage of the subject: https://cr.openjdk.java.net/~briangoetz/lambda/lambda-state-final.html

Manual testing with a debugger to verify whether or not references to `this` were present in the lambda. 

- Is the additional layer of abstraction worthwhile?
- Did I mess anything up when fixing?

**Where should we start reviewing?**: SnapshotTransaction

---

@Chrisbattarbee for SA (and also h/t for excellent internal sleuthing, without which this would have been _really_ hard to find).

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
@@ -167,6 +167,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1462,14 +1463,25 @@ private void deleteOrphanedSentinelsAsync(TableReference table, Set<Cell> actual
...
             }
         });
     }
 
+    /**
+     * This method exists to ensure that tasks run on the delete executor do not have references to the
+     * SnapshotTransaction that spawned the task (so that it can be garbage collected, which may be significant as
+     * large write or serializable read transactions may have allocated a lot of memory).
+     */
+    private void runTaskOnDeleteExecutor(Consumer<KeyValueService> task) {
```
gmaretic, line 1479: Pretty smart :)
Very nice find, this must have been a nightmare to figure out 🗡️

gmaretic (approved):

gmaretic:
Relevant note:

> References to this -- including implicit references through unqualified field references or method invocations -- are, essentially, references to a final local variable. Lambda bodies that contain such references capture the appropriate instance of this. In other cases, no reference to this is retained by the object.
> 
> This has a beneficial implication for memory management: while inner class instances always hold a strong reference to their enclosing instance, lambdas that do not capture members from the enclosing instance do not hold a reference to it. This characteristic of inner class instances can often be a source of memory leaks.

PR #5635, opened 2021-09-14
[LW] Additional filtering for values flushed to central cache

Jolyon-S:
**Goals (and why)**:
When a transaction is committed, it will get a commit timestamp, along with lock watch information at that point. These are stored in the event cache. Then, when the values are to be flushed to the central cache, the events between start and commit timestamp are retrieved, unlock events are filtered out, and the resultant set of lock descriptors represents the set of locks taken out from start to end of that transaction.

These are used to filter the values loaded within the transaction, and the remaining values are flushed to the central cache. However, the time at which values are flushed to the central cache is later than the commit timestamp: the transaction needs to actually be committed (and thus conflict checking etc. needs to happen) before we can use those values. Thus, the central cache will likely have progressed its own version by then, including having new lock descriptors taken out. The result is that we might attempt to write values to the central cache that are out of date or no longer valid.

Atlasdb-proxy doesn't have this issue as it did the correct thing simply because it couldn't do the incorrect thing: it retrieved events from the start of the transaction up until _the present_ (at value flushing time) - this is guaranteed to only ever be more conservative (as well as being correct, of course). AtlasDB just needs to do the same thing here.

Added a `SpanningCommitUpdate` that contains two commit updates: one which includes start timestamp to commit timestamp events, and one that contains events from start to latest. The reason for two is that serializiable conflict checking _should_ only be using the former, since it is only evaluating against writes that could have happened within the scope of the transaction, and using the latter one would result in more cache misses than is strictly necessary.

WIP

* Is it worth splitting the commit update like this? Should we just be more conservative and invalidate more when it comes to conflict checking?

`LockWatchValueScopingCacheImpl`

This week

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
@@ -81,22 +81,6 @@ interface ClientLogEvents {
                 .build();
     }
 
-    default CommitUpdate toCommitUpdate(LockWatchVersion startVersion, CommitInfo commitInfo) {
-        if (clearCache()) {
-            return CommitUpdate.invalidateAll();
-        }
-
-        // We want to ensure that we do not miss any versions, but we do not care about the event with the same version
-        // as the start version.
-        verifyReturnedEventsEnclosesTransactionVersions(
-                startVersion.version() + 1, commitInfo.commitVersion().version());
-
-        LockEventVisitor eventVisitor = new LockEventVisitor(commitInfo.commitLockToken());
-        Set<LockDescriptor> locksTakenOut = new HashSet<>();
-        events().events().forEach(event -> locksTakenOut.addAll(event.accept(eventVisitor)));
-        return CommitUpdate.invalidateSome(locksTakenOut);
-    }
-
     default SpanningCommitUpdate toSpanningCommitUpdate(
             LockWatchVersion startVersion, CommitInfo commitInfo, LockWatchVersion endVersion) {
         if (clearCache()) {
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
@@ -29,6 +29,7 @@ import com.palantir.lock.watch.LockEvent;
 import com.palantir.lock.watch.LockWatchCreatedEvent;
 import com.palantir.lock.watch.LockWatchEvent;
 import com.palantir.lock.watch.LockWatchVersion;
+import com.palantir.lock.watch.SpanningCommitUpdate;
 import com.palantir.lock.watch.TransactionsLockWatchUpdate;
 import com.palantir.lock.watch.UnlockEvent;
 import com.palantir.logsafe.SafeArg;
@@ -94,6 +95,33 @@ interface ClientLogEvents {
         Set<LockDescriptor> locksTakenOut = new HashSet<>();
         events().events().forEach(event -> locksTakenOut.addAll(event.accept(eventVisitor)));
         return CommitUpdate.invalidateSome(locksTakenOut);
+    }
+
+    default SpanningCommitUpdate toSpanningCommitUpdate(
+            LockWatchVersion startVersion, CommitInfo commitInfo, LockWatchVersion endVersion) {
+        if (clearCache()) {
+            return SpanningCommitUpdate.invalidateAll();
+        }
+
+        // We want to ensure that we do not miss any versions, but we do not care about the event with the same version
+        // as the start version.
+        verifyReturnedEventsEnclosesTransactionVersions(startVersion.version() + 1, endVersion.version());
+
+        LockEventVisitor eventVisitor = new LockEventVisitor(commitInfo.commitLockToken());
+        Set<LockDescriptor> transactionLocks = new HashSet<>();
+        Set<LockDescriptor> spanningLocks = new HashSet<>();
+        events().events().forEach(event -> {
+            Set<LockDescriptor> descriptors = event.accept(eventVisitor);
+            spanningLocks.addAll(descriptors);
+
+            if (event.sequence() <= commitInfo.commitVersion().version()) {
+                transactionLocks.addAll(descriptors);
+            }
+        });
+        return SpanningCommitUpdate.builder()
+                .transactionCommitUpdate(CommitUpdate.invalidateSome(transactionLocks))
+                .spanningCommitUpdate(CommitUpdate.invalidateSome(spanningLocks))
+                .build();
     }
 
     default void verifyReturnedEventsEnclosesTransactionVersions(long lowerBound, long upperBound) {
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
@@ -81,21 +81,7 @@ public final class LockWatchEventCacheImpl implements LockWatchEventCache {
 
     @Override
     public synchronized CommitUpdate getCommitUpdate(long startTs) {
-        Optional<LockWatchVersion> startVersion = timestampStateStore.getStartVersion(startTs);
-        Optional<CommitInfo> maybeCommitInfo = timestampStateStore.getCommitInfo(startTs);
-
-        assertTrue(
-                maybeCommitInfo.isPresent() && startVersion.isPresent(),
-                "start or commit info not processed for start timestamp");
-
-        CommitInfo commitInfo = maybeCommitInfo.get();
-
-        VersionBounds versionBounds = VersionBounds.builder()
-                .startVersion(startVersion)
-                .endVersion(commitInfo.commitVersion())
-                .build();
-
-        return eventLog.getEventsBetweenVersions(versionBounds).toCommitUpdate(startVersion.get(), commitInfo);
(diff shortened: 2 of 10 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
```diff
@@ -96,6 +97,33 @@ default CommitUpdate toCommitUpdate(LockWatchVersion startVersion, CommitInfo co
         return CommitUpdate.invalidateSome(locksTakenOut);
     }
 
+    default SpanningCommitUpdate toSpanningCommitUpdate(
```
Jolyon-S, line 100: Is there a nice way to de-duplicate code here?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -97,6 +98,28 @@ public synchronized CommitUpdate getCommitUpdate(long startTs) {
         return eventLog.getEventsBetweenVersions(versionBounds).toCommitUpdate(startVersion.get(), commitInfo);
     }
 
+    @Override
+    public SpanningCommitUpdate getSpanningCommitUpdate(long startTs) {
```
Jolyon-S, line 102: Again, somewhat duplicated; can we deduplicate this? Not sure.

Also, I don't want `getCommitUpdate` to necessarily call this as it does more work than it needs to.

lock-api-objects/src/main/java/com/palantir/lock/watch/SpanningCommitUpdate.java
```diff
@@ -0,0 +1,40 @@
...
+import com.palantir.common.annotations.ImmutablesStyles.StagedBuilderStyle;
+import com.palantir.lock.watch.ImmutableSpanningCommitUpdate.TransactionCommitUpdateBuildStage;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@StagedBuilderStyle
+public interface SpanningCommitUpdate {
+    CommitUpdate transactionCommitUpdate();
+
+    CommitUpdate spanningCommitUpdate();
```
Jolyon-S, line 28: Staged to prevent crossing wires!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
gmaretic: I think both of the below suggestions are reasonable

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
```diff
@@ -96,6 +97,33 @@ default CommitUpdate toCommitUpdate(LockWatchVersion startVersion, CommitInfo co
...
+        if (clearCache()) {
+            return SpanningCommitUpdate.invalidateAll();
+        }
+
+        // We want to ensure that we do not miss any versions, but we do not care about the event with the same version
+        // as the start version.
+        verifyReturnedEventsEnclosesTransactionVersions(startVersion.version() + 1, endVersion.version());
+
+        LockEventVisitor eventVisitor = new LockEventVisitor(commitInfo.commitLockToken());
+        Set<LockDescriptor> transactionLocks = new HashSet<>();
```
gmaretic, line 111: We can have a common method that returns a `Set` of something like `MaybeSpanningLockDescriptor`, from which you can easily construct one or the other so both of the methods can just call that. Takes an additional traversal for spanning but that's prob fine

gmaretic: Alternatively, you can just have toCommitUpdate call toSpanning and take the spanning update (or the other one, really)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
gmaretic: If you want to deduplicate: you should probably just have a single method on `ClientLogEvents` that takes all three arguments, then call it with `commitInfo.commitVersion().version()` above and get the `CommitUpdate` from the returned spanning update.
Not entirely sure without seeing it if it's a bit too magical, though

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
```diff
@@ -343,9 +343,48 @@ public void failedValidationCausesCacheToFallback() {
         assertThat(scopedCache2).isExactlyInstanceOf(NoOpTransactionScopedCache.class);
     }
 
+    @Test
+    public void locksAfterCommitTimeAreNotMissedWhenFlushingValues() {
```
gmaretic, line 347: Reads reasonably well, which is a great success all things considered

changelog/@unreleased/pr-5635.v2.yml
```diff
@@ -0,0 +1,7 @@
+type: fix
+fix:
+  description: Resolved an edge case in the lock watch value cache whereby loaded
```
gmaretic, line 3: It's actually surprising there were so few validation failures due to this

lock-api-objects/src/main/java/com/palantir/lock/watch/SpanningCommitUpdate.java
gmaretic: I don't think staged is really necessary -- they are already not parameters, so it's pretty difficult to cross wires.
One option here is, if you go with the `MaybeSpanningLockDescriptor` approach, to just have a factory that takes a set of those and then constructs the correct thing -- they can then even actually be `@Value.Lazy` so you prevent some of the extra work for the case where you do not care about spanning update. (That last one reminds me too much of timelock partitioning so probably best not to)

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
```diff
@@ -42,14 +42,26 @@ void processGetCommitTimestampsUpdate(
...
     /**
-     * Updates the cache with the update, and calculates the {@link CommitUpdate} taking into account all changes to
-     * lock watch state since the start of the transaction, excluding the transaction's own commit locks.
+     * Calculates the {@link CommitUpdate} taking into account all changes to lock watch state since the start of the
+     * transaction, excluding the transaction's own commit locks.
      *
      * @param startTs start timestamp of the transaction
      * @return the commit update for this transaction's precommit condition
      */
     CommitUpdate getCommitUpdate(long startTs);
```
gmaretic, line 51: When are we actually going to call this and not the other one? Is it just for atlas-proxy?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
Jolyon-S: Gone for the most simple, which is to just use spanning commit updates and take the relevant parts only

lock-api-objects/src/main/java/com/palantir/lock/watch/SpanningCommitUpdate.java
Jolyon-S: Eh, I like staged builders anyway so I'll stick with it (and no no no to `@Value.Lazy`!)

lock-api/src/main/java/com/palantir/lock/watch/LockWatchEventCache.java
Jolyon-S: Yes!

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -107,7 +93,7 @@ public SpanningCommitUpdate getSpanningCommitUpdate(long startTs) {
                 maybeCommitInfo.isPresent()
                         && startVersion.isPresent()
                         && eventLog.getLatestKnownVersion().isPresent(),
```
gmaretic, line 95: I guess you could extract a method to avoid throwing here for `getCommitUpdate` if latest known version is not present

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
```diff
@@ -354,7 +354,8 @@ public void locksAfterCommitTimeAreNotMissedWhenFlushingValues() {
         verify(metrics, times(1)).registerHits(0);
         verify(metrics, times(1)).registerMisses(2);
 
-        // Simulate the case where the transaction gets a commit timestamp but has not yet committed
+        // Simulate the case where the transaction gets a commit timestamp but has not yet committed. This start
+        // transactions update introduces the lock taken out on CELL_1
```
gmaretic, line 358: :noice:

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
Jolyon-S: I could, but actually I don't want to, since if you don't have a current version, you most definitely do not have any of the other information present.

gmaretic (approved):
LGTM

PR #5637, opened 2021-09-14
Fix Potential Missed Serializable Transaction Conflicts

jeremyk-91:
**Goals (and why)**:
- Catch transaction conflicts

- Don't read at commitTs + 1 but instead at commitTs
- Tests

- Added a test that exposes the bug

- Is this actually safe

**Where should we start reviewing?**: SerializableTransaction

**Priority (whenever / two weeks / yesterday)**: five years ago 🔥


atlasdb-processors/build.gradle
```diff
@@ -7,8 +7,8 @@ dependencies {
     compileOnly group: 'com.google.auto.service', name: 'auto-service'
 
     compile group: 'com.google.guava', name: 'guava'
-    implementation group: 'com.squareup', name: 'javapoet'
-    implementation group: 'com.palantir.goethe', name: 'goethe'
+    compile group: 'com.squareup', name: 'javapoet'
+    compile group: 'com.palantir.goethe', name: 'goethe'
```
Jolyon-S, line 11: do we want to commit this change?

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
```diff
@@ -1408,6 +1414,47 @@ public void testGetSortedColumnsConflictForReadBoundaryEntry() {
...
+
+        assertThatThrownBy(() -> manager.runTaskThrowOnConflict(tx -> {
+            long myTs = tx.getTimestamp();
+            tx.get(TEST_TABLE_SERIALIZABLE, ImmutableSet.of(CELL_ONE));
+
+            // This write is done to ensure that the transaction actually has writes that need to commit
+            tx.put(TEST_TABLE_SERIALIZABLE, ImmutableMap.of(CELL_TWO, BYTES_TWO));
+
+            keyValueService.put(TEST_TABLE_SERIALIZABLE, ImmutableMap.of(CELL_ONE, BYTES_TWO), myTs + 1);
+            transactionService.putUnlessExists(myTs + 1, myTs + 2);
```
Jolyon-S, line 1434: we should comment that this is effectively orchestrating a "B" transaction (if this is "ABA").

atlasdb-processors/build.gradle
gmaretic: 🔥

atlasdb-tests-shared/src/main/java/com/palantir/atlasdb/transaction/impl/AbstractSerializableTransactionTest.java
gmaretic: Yeah maybe a bit more comment on what we are doing

gmaretic (approved):
Safe as far as I can tell

atlasdb-processors/build.gradle
jeremyk-91: Nope, will remove! (At least for now)

PR #5643, opened 2021-09-16
[LW] Add testing around edge cases when committing transactions

Jolyon-S:
**Goals (and why)**:
https://github.com/palantir/atlasdb/pull/5637 was causing validation failures - not the ABA case, but rather AB where the B write occurred at commit timestamp plus one. Aim here is to add testing around this case to confirm that this is also no longer an issue with validation here.

Add a test that fakes a transaction committing at `commitTimestamp + 1`.

Is this test reasonable to understand?

Diff

This week

### DIFF
--- a/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
+++ b/timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
@@ -379,20 +379,17 @@ public final class LockWatchValueIntegrationTest {
         // truncate the table to verify we are really using the cached values
         truncateTable();
 
-        assertThatCode(() -> txnManager.runTaskThrowOnConflict(txn -> {
-                    NavigableMap<byte[], RowResult<byte[]>> read = txn.getRows(TABLE_REF, rows, columns);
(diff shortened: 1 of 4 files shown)


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -374,12 +386,85 @@ public void getRowsUsesCacheAsExpected() {
                     txn,
                     ImmutableSet.of(
                             CellReference.of(TABLE_REF, CELL_2), CellReference.of(TABLE_REF, CELL_3), TABLE_CELL_4));
-            // we weren't able to cache our own write so we look this up
+            // we weren't able to cache our own write, so we look this up
```
Jolyon-S, line 389: grammar

jeremyk-91 (commented):
Thanks for opening this PR! While this test does cover what we want to do (it builds on the previous tests by integrating the LW layer), there is an issue. Specifically, pre-commit conditions are run _after_ the serializable conflict check, and so this test still passes even if I undo the commit + 1 fix. I think we need to do the writes slightly earlier.

jeremyk-91 (approved):

PR #5646, opened 2021-09-17
[LW] Resolve possible resource leak

Jolyon-S:
**Goals (and why)**:
Some amount of state is added during the start timestamp flow, but the relevant state is _not_ removed when a segment of the batch fails.

* Add methods to remove state for failed start transaction attempts, and wire those through the start transaction batchers.

Multi namespace batched ones only

This fixes these issues, but not the general pattern. A followup PR will either make these weak/soft references, or just kill the cache store in general (although note that this does not make the changes in this PR irrelevant).

Diff

Sooner rather than later


lock-api/src/main/java/com/palantir/lock/client/BatchingCommitTimestampGetter.java
```diff
@@ -73,7 +73,6 @@ public long getCommitTimestamp(long startTs, LockToken commitLocksToken) {
                 GetCommitTimestampsResponse response =
                         leaseService.getCommitTimestamps(requestedVersion, count - commitTimestamps.size());
                 commitTimestamps.addAll(process(batch.subList(commitTimestamps.size(), count), response, cache));
-                LockWatchLogUtility.logTransactionEvents(requestedVersion, response.getLockWatchUpdate());
```
Jolyon-S, line 76: Drive-by

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
```diff
@@ -135,6 +137,7 @@ public void servesRequestsAsSoonAsResponseIsReceived() {
         LockCleanupService relevantLockCleanupService = LOCK_CLEANUP_SERVICE_MAP.get(namespace);
         verify(relevantLockCleanupService).refreshLockLeases(any());
         verify(relevantLockCleanupService).unlock(any());
+        verify(NAMESPACE_CACHE_MAP.get(namespace), times(4)).removeTransactionStateFromCache(anyLong());
```
Jolyon-S, line 140: These numbers are magical to me; @sudiksha27 mind giving me some context so I can at least comment on why this is 4 (or 5 below)?

jeremyk-91 (approved):
should be good - but agree, would like comments on the magic 4 and 5

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
sudiksha27: Right, so, for one namespace, we accumulate all requests for it in ResponseHandler#pendingFutures i.e. this queue is like [1, 25] for this test. Then we try to fetch as many startTs as possible from the server. For each batch of startTxnReponses, we try to server as many futures as we can. In one go our limit is 5 for these tests. So basically for one namespace i can fetch 5 responses.
If you see ResponseHandler#serveRequests, you will find, we greedily server requests from the pool of responses and the rest of the responses are transient i.e. they may be served if we have enough responses from future server calls. 
In this test, we can serve 1st request that needs 1 response. The next req needs 25 so we cannot serve it and therefore, there are 4 requests that need clearing up (server is configured to throw after one call)

lock-api/src/test/java/com/palantir/lock/client/MultiClientTransactionStarterTest.java
```diff
@@ -166,6 +169,8 @@ public void shouldNotFreeResourcesIfRequestIsServed() {
         verify(LOCK_CLEANUP_SERVICE_MAP.get(alpha), never()).unlock(any());
         verify(LOCK_CLEANUP_SERVICE_MAP.get(beta)).refreshLockLeases(any());
         verify(LOCK_CLEANUP_SERVICE_MAP.get(beta)).unlock(any());
+        verify(NAMESPACE_CACHE_MAP.get(alpha), never()).removeTransactionStateFromCache(anyLong());
+        verify(NAMESPACE_CACHE_MAP.get(beta), times(5)).removeTransactionStateFromCache(anyLong());
     }
```
sudiksha27, line 174: From above, the queue for beta here is [25] which means we cannot server any request. So the 5 responses we have from the first server call are transient and will be cleared.

sudiksha27 (commented):
LGTM!

PR #5649, opened 2021-09-20
Switch Cluster client to use DCAwareRR

leonz:
**Goals (and why)**:
In order to perform a datacenter switch, we need to be able to isolate Cassandra clients to a specific datacenter ("local").  The DCAwareRR client ensures that we do not try to route requests to the "remote" datacenter.

(Copied from comment below): I believe that the equivalent of this for Thrift is already practically implemented by the local host filtering in the CassandraService: https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java#L303

This change is not required for correctness, especially since we will disable the client interfaces on the remote DC. Instead, it's meant to gently urge the clients to behave correctly the first time, to minimize churn and make it easier to reason about queries.

Use the Datastax DCAwareRR policy.

I don't believe that we rely on the existing implementation for anything important, please correct me if I am wrong.

### DIFF
--- a/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
+++ b/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
@@ -138,11 +138,11 @@ public class DefaultCqlClientFactory implements CqlClientFactory {
         // Refuse to talk to nodes twice as (latency-wise) slow as the best one, over a timescale of 100ms,
         // and every 10s try to re-evaluate ignored nodes performance by giving them queries again.
         //
-        // We are using the DCAware RR policy specifically to support "datacenter switch"-style Cassandra,
-        // migrations (see https://thelastpickle.com/blog/2019/02/26/data-center-switch.html),
-        // and don't expect the policy to be helpful for latency improvements.
+        // The DCAware RR policy prevents auto-discovery of remote datacenter contact points in a multi-DC setup,
+        // which we are using to orchestrate migrations across datacenters.  We don't expect the policy to be helpful
+        // for latency improvements.
         //
-        // We don't specify the local DC, so it will get set based on the contact points we provide.
+        // Since the local DC isn't specified, it will get set based on the contact points we provide.
         LoadBalancingPolicy policy = LatencyAwarePolicy.builder(
                         DCAwareRoundRobinPolicy.builder().build())
                 .build();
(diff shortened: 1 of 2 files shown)


atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
```diff
@@ -137,12 +137,15 @@ public DefaultCqlClientFactory() {
...
-        LoadBalancingPolicy policy =
-                LatencyAwarePolicy.builder(new RoundRobinPolicy()).build();
+        //
+        // The DCAware RR policy prevents auto-discovery of remote datacenter contact points in a multi-DC setup,
+        // which we are using to orchestrate migrations across datacenters.  We don't expect the policy to be helpful
+        // for latency improvements.
+        //
+        // Since the local DC isn't specified, it will get set based on the contact points we provide.
+        LoadBalancingPolicy policy = LatencyAwarePolicy.builder(
+                        DCAwareRoundRobinPolicy.builder().build())
```
j-baker, line 147: I don't think this will work. The DCAwareRoundRobinPolicy needs to know the _name_ of the local datacentre in order to work properly, which isn't accessible to us here.

leonz: The DCRR policy will auto-discover the datacenter based on the first host to which it connects: https://github.com/datastax/java-driver/blob/3.7.2/driver-core/src/main/java/com/datastax/driver/core/policies/DCAwareRoundRobinPolicy.java#L111

leonz:
I believe that the equivalent of this for Thrift is already practically implemented by the local host filtering in the _CassandraService_: https://github.com/palantir/atlasdb/blob/develop/atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/pool/CassandraService.java#L303

To be extra clear, this change is not required for correctness, especially since we will disable the client interfaces on the remote DC.  Instead, it's meant to gently urge the clients to behave correctly the first time, to minimize churn and make it easier to reason about queries.

atlasdb-cassandra/src/main/java/com/palantir/atlasdb/keyvalue/cassandra/async/client/creation/DefaultCqlClientFactory.java
jeremyk-91: Given the autodiscovery code I think this is correct (though admittedly have less Cassandra experience than either of you!)

jeremyk-91 (approved):
👍

PR #5651, opened 2021-09-21
Do not recreate configs unnecessarily

gmaretic:
**Goals (and why)**:
Do not keep calling the config supplier

This is pretty awful

Existing

Should we instead move this to use dialogue? Can we?


gmaretic:
Changelog bot is confused, we can survive without it

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -76,19 +76,16 @@ private ConjureJavaRuntimeTargetFactory() {
...
-        Supplier<T> clientSupplier = new CachedTransformingSupplier<>(
-                serverListConfigSupplier,
-                serverListConfig -> JaxRsClient.create(
-                        type,
-                        addAtlasDbRemotingAgent(parameters.userAgent()),
-                        NoOpHostEventsSink.INSTANCE,
-                        options.serverListToClient(serverListConfig)));
-
-        return decorateFailoverProxy(type, clientSupplier);
+        Refreshable<T> refreshableClient = serverListConfigRefreshable.map(serverListConfig -> JaxRsClient.create(
```
sudiksha27, line 83: Why do we want to keep `CachedTransformingSupplier` around? Seems like we should always take the Refreshable path?

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
```diff
@@ -76,19 +76,16 @@ private ConjureJavaRuntimeTargetFactory() {
...
+                type,
+                addAtlasDbRemotingAgent(parameters.userAgent()),
+                NoOpHostEventsSink.INSTANCE,
+                options.serverListToClient(serverListConfig)));
+        return decorateFailoverProxy(type, refreshableClient);
     }
 
     public <T> InstanceAndVersion<T> createProxyWithQuickFailoverForTesting(
```
sudiksha27, line 91: this method is not being used?

timelock-agent/src/main/java/com/palantir/timelock/invariants/TimeLockActivityCheckerFactory.java
```diff
@@ -50,7 +51,10 @@ public TimeLockActivityCheckerFactory(
 
     private ServiceCreator createServiceCreatorForRemote(String remoteUrl) {
         return ServiceCreator.withPayloadLimiter(
-                metricsManager, () -> getServerListConfig(remoteUrl), userAgent, () -> RemotingClientConfigs.DEFAULT);
+                metricsManager,
+                Refreshable.only(getServerListConfig(remoteUrl)),
```
sudiksha27, line 55: This is not actually refreshable, right? maybe we should leave a comment.

sudiksha27 (commented):
Looks good to me - have some minor questions for my understanding.

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
gmaretic: 🎉

atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ConjureJavaRuntimeTargetFactory.java
gmaretic: I didn't want to break the public API, but given the name and fact that it is not used in large internal product, should be fine

timelock-agent/src/main/java/com/palantir/timelock/invariants/TimeLockActivityCheckerFactory.java
gmaretic: Is `ClusterConfiguration` mutable? Otherwise it is not really a change

sudiksha27: I do not think so. It is a snapshot not an actual refreshable IIRC.

sudiksha27 (approved):
LGTM!

PR #5652, opened 2021-09-22
[LW] Rework value cache commit flow

Jolyon-S:
**Goals (and why)**:
There are four important parts of a cache's lifecycle:

```
------- creation
   |
   |
------- commit timestamp
   |
------- serializable conflict checking (done at commit timestamp)
   |
------- put unless exists
```

Creation happens at start timestamp, and currently works. The commit timestamp contains some lock watch updates, but importantly there may be other commits between this and the final put unless exists. Additionally, the transaction can fail at any point in this period.

Currently, we put values at commit timestamp time, which is definitely wrong, as values aren't (effectively) committed to the KVS until put unless exists time. Thus, we need to refactor the flow such that pushing values to the central cache only happens after put unless exists, and using the new `onSuccess` method is the natural way to do so.

Serializable conflict checking happens after the commit timestamp, but uses the commit timestamp as the point in time in the KVS, and thus creating the read-only cache at that time remains correct.

* Revert non-test changes from https://github.com/palantir/atlasdb/pull/5635
* Instead of adding a spanning commit update, just add a method that gets all updates from a start timestamp's version to latest (the difference is that we never need both commit update and spanning update together, and thus that PR no longer really makes sense).
* Add _another_ method on the whole cache series of interfaces for `onTransactionCommit`, and register this at the very creation of the transaction.
* Remove obtaining a commit timestamp for (effectively) read-only transactions with cached reads, since the on success method makes this redundant.

Fixed up current tests, and added additional tests at the unit and integration level for LWVC. Note that there are no good tests for LWEC or `ClientLogEvents`, but that will come in a followup PR.

* We're now adding additional places where all transactions will block on a single object. Do we think this will affect performance? Should we reconsider the data structures used to enhance parallelism?

`LockWatchValueScopingCacheImpl`

The sooner the better, as this may well be the last failing validation bug.


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -81,29 +80,39 @@ public synchronized void processGetCommitTimestampsUpdate(
...
+        CommitInfo commitInfo = maybeCommitInfo.get();
+
+        VersionBounds versionBounds = VersionBounds.builder()
+                .startVersion(startVersion)
+                .endVersion(commitInfo.commitVersion())
+                .build();
+
+        return eventLog.getEventsBetweenVersions(versionBounds)
+                .toCommitUpdate(startVersion.get(), commitInfo.commitVersion(), maybeCommitInfo);
     }
```
Jolyon-S, line 99: Reverted to how it was before (with the difference in `toCommitUpdate` call)

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
```diff
@@ -202,6 +202,7 @@ protected boolean shouldStopRetrying(int numTimesFailed) {
 
                                 Transaction transaction = createTransaction(
                                         immutableTs, startTimestampSupplier, immutableTsLock, condition);
+                                transaction.onSuccess(() -> lockWatchManager.onSuccess(transaction.getTimestamp()));
```
Jolyon-S, line 205: According to the contract of `onSuccess`, this should execute first (unless there are other calls registered while the transaction is being created).

However, if this fails, we'll still clean up the state, but won't flush the values to the central cache, which is acceptable.

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
```diff
@@ -311,14 +306,6 @@ public void leaderElectionCausesCacheToBeCleared() {
...
         TransactionScopedCache scopedCache3 = valueCache.getTransactionScopedCache(TIMESTAMP_3);
         assertThat(getRemotelyReadCells(scopedCache3, TABLE, CELL_1)).containsExactlyInAnyOrder(CELL_1);
-        eventCache.processGetCommitTimestampsUpdate(
-                ImmutableList.of(TransactionUpdate.builder()
-                        .startTs(TIMESTAMP_3)
-                        .commitTs(999_999_999)
-                        .writesToken(LockToken.of(UUID.randomUUID()))
-                        .build()),
-                LockWatchStateUpdate.success(newLeader, -1L, ImmutableList.of()));
-        valueCache.updateCacheOnCommit(ImmutableSet.of(TIMESTAMP_3));
```
Jolyon-S, line 321: Not entirely sure what this code was aiming to do

lock-api/src/main/java/com/palantir/lock/watch/LockWatchCacheImpl.java
```diff
@@ -43,14 +43,20 @@ public void processStartTransactionsUpdate(Set<Long> startTimestamps, LockWatchS
...
-        valueCache.removeTransactionState(startTimestamp);
+    public void removeTransactionStateFromCache(long startTs) {
+        eventCache.removeTransactionStateFromCache(startTs);
+        valueCache.ensureStateRemoved(startTs);
+    }
+
+    @Override
+    public void onSuccess(long startTs) {
+        valueCache.onSuccessfulCommit(startTs);
+        eventCache.removeTransactionStateFromCache(startTs);
```
Jolyon-S, line 59: Note the reversed order here - it matters for `onSuccess`, but actually doesn't matter for `removeTransactionStateFromCache`

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -423,6 +423,55 @@ public void nearbyCommitsDoNotAffectResultsPresentInCache() {
...
+                throw new RuntimeException("Transaction failed at commit time");
+            }
+        };
+        try {
+            txnManager.runTaskWithConditionThrowOnConflict(commitFailingCondition, (txn, _unused) -> {
+                startTimestamp.set(txn.getTimestamp());
+                txn.put(TABLE_REF, ImmutableMap.of(CELL_4, DATA_4));
+                txn.get(TABLE_REF, ImmutableSet.of(CELL_1, CELL_2, CELL_3));
+                return null;
+            });
```
jeremyk-91, line 447: please insert a `fail();` after this line: we don't want the test to pass if this transaction actually commits

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
jeremyk-91: I agree this is _acceptable_. If I'm not mistaken we will end up, for successful transactions, cleaning up the state twice (`OpenTransactionImpl#finish()` also removes the transaction state from the cache) - do we need to update that?

jeremyk-91 (commented):
This took more than an hour to digest, but that's probably fair given the difficulty of the code, and it looks like this works to me. Some questions; would also be interested in @gmaretic's review, as he's worked somewhat more closely with this in the past.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransactionManager.java
Jolyon-S: I think calling this twice is perfectly fine. However, if you disagree, we can rewire so it does not.

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
Jolyon-S: Refactored to just use `assertThatThrownBy` since that's just better.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
```diff
@@ -130,13 +118,15 @@ default void verifyReturnedEventsEnclosesTransactionVersions(long lowerBound, lo
...
-            }
+        private LockEventVisitor(Optional<LockToken> commitLocksToken) {
+            commitRequestId = commitLocksToken.flatMap(lockToken -> {
+                if (lockToken instanceof LeasedLockToken) {
+                    return Optional.of(
+                            ((LeasedLockToken) lockToken).serverToken().getRequestId());
+                } else {
+                    return Optional.empty();
+                }
+            });
```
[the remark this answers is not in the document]
Jolyon-S: Not quite sure what you mean here - can you elaborate?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
```diff
@@ -1815,15 +1797,6 @@ private void commitWrites(TransactionService transactionService) {
                 if (validationNecessaryForInvolvedTablesOnCommit()) {
                     throwIfImmutableTsOrCommitLocksExpired(null);
                 }
-
-                // if the cache has been used, we must work out which values can be flushed to the central cache by
-                // obtaining a commit update, which is obtained via the get commit timestamp request.
-                if (getCache().hasUpdates()) {
```
Jolyon-S, line 1821: We no longer require this given that we have `onTransactionCommit`, and that we don't need to construct the read-only cache for serialisable read-only transactions.

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImpl.java
```diff
@@ -155,6 +143,31 @@ public Void invalidateSome(Set<LockDescriptor> invalidatedLocks) {
                 return null;
             }
         });
+        ensureStateRemoved(startTimestamp);
+    }
+
+    /**
+     * Retrieval of transaction scoped caches (read-only or otherwise) does not need to be synchronised. The main
+     * reason for this is that, the only race condition that could conceivably occur is for the state to not exist here
```
gmaretic, line 151: nit
```suggestion
     * reason for this is that the only race condition that could conceivably occur is for the state to not exist here
```

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/ClientLogEvents.java
gmaretic, line 129:
```suggestion
            commitRequestId = commitLocksToken
                    .filter(lockToken -> lockToken instanceof LeasedLockToken)
                    .map(lockToken ->
                            ((LeasedLockToken) lockToken).serverToken().getRequestId());
```

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/transaction/impl/SnapshotTransaction.java
gmaretic: 🌮

atlasdb-impl-shared/src/test/java/com/palantir/atlasdb/keyvalue/api/cache/LockWatchValueScopingCacheImplTest.java
gmaretic: encouraging 🔪

gmaretic (approved):
Generally looks good, though would have been easier to review if you separated out the revert from the other changes 😭

PR #5653, opened 2021-09-22
TL server killer

sudiksha27:
**Goals (and why)**:
The ultimate TL server killer endpoint. This is required to orchestrate automated cross-infra TL server cluster migrations.

Should this endpoint have any additional safety checks to prevent destruction of cluster?

By EOD would be great 🏃🏼‍♀️

### DIFF
--- a/changelog/@unreleased/pr-5653.v2.yml
+++ b/changelog/@unreleased/pr-5653.v2.yml
@@ -1,5 +1,6 @@
 type: feature
 feature:
-  description: We now expose an endpoint that kills TimeLock server.
+  description: There is now the option to shut down the TimeLock server i.e. stop
+    the server from talking to peer nodes or serve requests.
   links:
   - https://github.com/palantir/atlasdb/pull/5653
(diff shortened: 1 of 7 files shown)


changelog/@unreleased/pr-5653.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: feature
+feature:
+  description: We now expose an endpoint that kills TimeLock server.
```
jeremyk-91, line 3: This is true. I don't know if we want to call it out as loudly in the release notes though!

jeremyk-91 (dismissed):
🔥

jeremyk-91 (commented):
actually: before we merge this, can we do a nicer shutdown? internal server API lets you stop the lifecycle I think, so if we can pass a callback into the timelock agent that might work better

timelock-impl/src/main/java/com/palantir/atlasdb/timelock/management/TimeLockManagementResource.java
```diff
@@ -102,6 +111,12 @@ public static TimeLockManagementService jersey(
         });
     }
 
+    @Override
+    public ListenableFuture<Void> forceKillTimeLockServer(AuthHeader authHeader) {
+        serviceStopper.run();
```
jeremyk-91, line 116: can we log something just to make it obvious that this is running?

jeremyk-91 (approved):
👍

PR #5654, opened 2021-09-22
Nuke Persistent Lock

gmaretic:
**Goals (and why)**:
Simplify code, stop failing on specific legacy sweep use cases due to leaking the lock

Scorched earth policy

Existing tests, I guess

Does someone still rely on this? Even if so, they will have noticed the endpoint disappearing so should be super low risk. Now there is no deduplication for multiple parallel legacy sweep processes -- do we care (no)?

idk

Today


docs/source/cluster_management/backup-restore.rst
```diff
@@ -1,106 +0,0 @@
-.. _backup-restore:
-
-==================
-Backup and Restore
-==================
-
-Overview
-========
-
-Backup and restore for AtlasDB are non-trivial operations that are not satisfied by simply backing up and restoring the underyling key value store you are using to store your data in.  This page describes how to take advantage of built in AtlasDB utilities, namely the Atlas CLI, in order to perform a live consistent point in time backup and subsequent restore of that backup from total data loss.
```
jeremyk-91, line 10: should probably have a follow up to actually preserve some of the information on how one would perform a backup.

jeremyk-91 (commented):
Thanks for doing this! Some notes:

- We will need to patch large internal product appropriately, I'd suggest we check with the maintainers if this is safe to remove completely. I can't see a reason why it _wouldn't_ be safe, but we should check.
- There really should exist some documentation on the theory of backup and restore, but that's outside the scope of this change, and it's definitely a positive to eliminate the current documentation, which is _actively wrong_ and may cause *SEVERE DATA CORRUPTION* (TM).

jeremyk-91 (approved):
we've given a day's notice, and can add this back later on if desired. 👍

gmaretic:
👍

PR #5655, opened 2021-09-22
Retry Legacy Sweep Iterations

gmaretic:
**Goals (and why)**:
Do not fail if a RuntimeException happens (such as insufficient consistency or too many retries)

Super straightforward, just block for 5 seconds, then retry

None

Went with the min effort approach, should we do a fancier exponential backoff?

Today

### DIFF
--- a/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SpecificTableSweeper.java
+++ b/atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SpecificTableSweeper.java
@@ -16,8 +16,10 @@
 package com.palantir.atlasdb.sweep;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.palantir.atlasdb.encoding.PtBytes;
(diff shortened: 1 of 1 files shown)


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/sweep/SpecificTableSweeper.java
```diff
@@ -126,15 +126,21 @@ void runOnceAndSaveResults(TableToSweep tableToSweep, SweepBatchConfig batchConf
 
     SweepResults runOneIteration(
             TableReference tableRef, byte[] startRow, SweepBatchConfig batchConfig, SweepTaskRunner.RunType runType) {
-        try {
```
jeremyk-91, line 129: Could we have a finite number of retries, maybe 50 or something?

The problem here is that there is an edge case where if the network link between an Atlas node A1 and a Cassandra node C1 fails even though there exists another Atlas node A2 that can talk to C1, then we deadlock (because this is run holding the sweep locks) and this could starve background sweep.

jeremyk-91 (approved):
👍

PR #5656, opened 2021-09-22
Make CachedTransformingSupplier exception safe

blacksmithgu:
If `outputFunction.apply` throws an exception, the supplier will be stuck in a broken state until the input changes again. Swapping the order of assignment will only update the input if the output computation succeeds, forcing the supplier to recompute the output every call until it does not error.

This makes it so that errors in `outputFunction` show up in error stacktraces, instead of getting a mysterious `NullPointerException` after the first failure when calling `CachedTransformingSupplier#get`.

### DIFF
--- a/atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
+++ b/atlasdb-commons/src/main/java/com/palantir/util/CachedTransformingSupplier.java
@@ -49,8 +49,8 @@ public class CachedTransformingSupplier<T, U> implements Supplier<U> {
             return cachedOutput;
         }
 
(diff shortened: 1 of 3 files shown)


jeremyk-91 (commented):
Thanks for the PR! This is legit, could we have a test to demonstrate why this fixes the issue?

jeremyk-91:
Actually looks like this isn't used any more after #5651 merged, and refreshables should handle this correctly, so I think we can close this out.

blacksmithgu:
I can see it was removed - I'll verify that fixes the use case where I stumbled across this, but I don't think this is needed anymore either.

PR #5658, opened 2021-09-23
[LW] Add additional fail-safe mechanisms around stored state

Jolyon-S:
**Goals (and why)**:
We've hit several memory leaks in the past within the lock watch code path. The aim here is to not only prevent these from taking down a service, but also to automatically surface them. Instead of adding metrics for each of these, they instead cause the cache to fall back. This in turn increments the metric for cache fallbacks, which we can then monitor or create alerts from.

Throw `SafeIllegalStateException` if internal structures exceed a constant threshold. In practice, these structures should be approximately proportional in size to the number of open transactions, which will almost never exceed 5k, and thus 20k is a _very_ safe limit.

Note that any runtime exception is caught in the resilient cache and rewrapped as a retryable exception after the fallback has been selected.

No tests for these bounds.

Should we just take the $$$ hit and add metrics? My instincts say no, since we likely won't need the metrics unless there is a real issue.

Diff

This/next week


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
```diff
@@ -98,6 +96,18 @@ public TransactionScopedCache getReadOnlyCache(StartTimestamp timestamp) {
                 .orElseGet(() -> NoOpTransactionScopedCache.create().createReadOnlyCache(CommitUpdate.invalidateAll()));
     }
 
+    private void validateStateSize() {
+        if (cacheMap.size() > maxCacheCount) {
+            log.warn(
+                    "Transaction cache store has exceeded maximum concurrent caches. This likely indicates a memory"
+                            + " leak",
+                    SafeArg.of("cacheMapSize", cacheMap.size()),
```
gmaretic, line 104: Let's also log the lowest timestamp in the map? This should let us figure out how old the oldest entry is

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
```diff
@@ -75,4 +84,18 @@ public void reset() {
     public Optional<ValueCacheSnapshot> getSnapshotForSequence(Sequence sequence) {
         return Optional.ofNullable(snapshotMap.get(sequence));
     }
+
+    private void validateStateSize() {
+        if (snapshotMap.size() > MAXIMUM_SIZE
+                || liveSequences.size() > MAXIMUM_SIZE
+                || timestampMap.size() > MAXIMUM_SIZE) {
+            log.warn(
+                    "Snapshot store has exceeded its maximum size. This likely indicates a memory leak.",
```
gmaretic, line 93: Again, maybe log the minimum start timestamp in timestamp map?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/LockWatchEventCacheImpl.java
```diff
@@ -36,7 +36,7 @@
 
 @ThreadSafe
 public final class LockWatchEventCacheImpl implements LockWatchEventCache {
-    // The minimum number of events should be the same as Timelocks' LockEventLogImpl.
+    // The minimum number of events should be the same as Timelock's LockEventLogImpl.
```
gmaretic, line 39: 🎉

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/watch/TimestampStateStore.java
```diff
@@ -93,10 +101,21 @@ TimestampStateStoreState getStateForTesting() {
...
 
-    public Optional<Sequence> getEarliestLiveSequence() {
+    Optional<Sequence> getEarliestLiveSequence() {
         return Optional.ofNullable(Iterables.getFirst(livingVersions.keySet(), null));
     }
 
+    private void validateStateSize() {
+        if (timestampMap.size() > MAXIMUM_SIZE || livingVersions.size() > MAXIMUM_SIZE) {
+            log.warn(
+                    "Timestamp state store has exceeded its maximum size. This likely indicates a memory leak",
```
gmaretic, line 111: same as above

gmaretic (approved):
Looks good, but I'd add a bit more info to the log lines

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheStoreImpl.java
Jolyon-S: It's not a navigable map in any way, so that's hard to do :(

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/SnapshotStoreImpl.java
Jolyon-S: And neither are these

PR #5659, opened 2021-09-23
Reduce MNPTLSIT flakes

Jolyon-S:
**Goals (and why)**:
This test flakes an incredible amount, and this deeply impacts our ability to roll out features on AtlasDB.

Force the test to run each case sequentially. This will hopefully prevent errors around the SQLite DB being locked (which is one such issue).

This will slow down tests. Is this acceptable?

Diff

ASAP if you want tests to pass (provided this works...)

### DIFF
--- a/atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
+++ b/atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
@@ -79,9 +79,8 @@ public class SafeShutdownRunnerTest {
         runner.shutdownSafely(blockingUninterruptibleRunnable);
         runner.shutdownSafely(blockingUninterruptibleRunnable);
 
-        verify(blockingUninterruptibleRunnable, times(2)).run();
-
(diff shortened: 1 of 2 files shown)


atlasdb-commons/src/test/java/com/palantir/util/SafeShutdownRunnerTest.java
```diff
@@ -79,9 +79,8 @@ public void slowTasksTimeOutWithoutThrowing() {
         runner.shutdownSafely(blockingUninterruptibleRunnable);
         runner.shutdownSafely(blockingUninterruptibleRunnable);
 
-        verify(blockingUninterruptibleRunnable, times(2)).run();
-
         closeAndAssertNumberOfTimeouts(runner, 2);
+        verify(blockingUninterruptibleRunnable, times(2)).run();
```
Jolyon-S, line 83: This also flaked. Re-ordering should enforce that the shutdown runner has closed and thus has definitely run both tasks.

gmaretic (approved):
Nice!

PR #5660, opened 2021-09-23
[LW] Fix cache size metrics

Jolyon-S:
**Goals (and why)**:
Fix cache metrics - they aren't reset when the cache is reset.

Reset metrics on reset.

Is `cleanUp()` too heavy-handed?

Diff

Whenever


atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
```diff
@@ -87,6 +87,10 @@ public void decreaseCacheSize(long removed) {
         cacheSize.dec(removed);
     }
 
+    public void resetCacheSize() {
+        cacheSize.dec(cacheSize.getCount());
+    }
```
Jolyon-S, line 92: why there isn't a reset method is beyond me

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -406,7 +406,7 @@ public void nearbyCommitsDoNotAffectResultsPresentInCache() {
         AtomicReference<LockWatchCache> lwCache = new AtomicReference<>(null);
         PreCommitCondition condition = timestamp -> {
             if (timestamp != startTs.get()) {
-                simulateOverlappingWriteTransaction(lwCache, timestamp);
+                simulateOverlappingWriteTransaction(lwCache, startTs.get(), timestamp);
```
Jolyon-S, line 409: Changes in this class are a drive-by to reduce flakes (there was a weird race condition here, because I think other transactions were being started in the background).

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/CacheMetrics.java
gmaretic: Probably they expect you to throw it away and create a new counter?

atlasdb-impl-shared/src/main/java/com/palantir/atlasdb/keyvalue/api/cache/ValueStoreImpl.java
```diff
@@ -78,8 +78,12 @@
     @Override
     public void reset() {
         values.resetToInitialValue();
-        loadedValues.invalidateAll();
         watchedTables.resetToInitialValue();
+        loadedValues.invalidateAll();
+
+        // Forcing the cache to run cleanup here guarantees that the metrics are not affected after they have been reset
+        loadedValues.cleanUp();
+        metrics.resetCacheSize();
```
gmaretic, line 86: Kind of jank that invalidation does not trigger removal listeners

gmaretic (approved):
LGTM

PR #5666, opened 2021-09-27
[LW] Add Value Stress Test

Jolyon-S:
**Goals (and why)**:
In the past, we've often found that stress tests can allow us to catch edge cases that conventional unit or integration tests do not find. Thus, adding a totally random stress test to help catch any unseen exceptions.

This should not necessarily be the final state: we may want to try and orchestrate a deterministic stress test, but given how difficult that currently will be within AtlasDB's transaction flow, this won't be attempting in this PR.

Add a stress test with 10k transactions and 200 threads.

This needs to give us decent signal while not impacting test flakiness or speed _too_ much.

Diff

This week


timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -96,14 +104,22 @@
     private static final TestableTimelockCluster CLUSTER = new TestableTimelockCluster(
             "non-batched timestamp paxos single leader",
             "paxosMultiServer.ftl",
-            generateThreeNodeTimelockCluster(9090, builder -> builder.clientPaxosBuilder(
+            generateThreeNodeTimelockCluster(9096, builder -> builder.clientPaxosBuilder(
```
Jolyon-S, line 107: I have a suspicion that 9090 was conflicting (I assume it spins up 9090, 9091, 9092, and we use 9092 elsewhere too).

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/LockWatchEteTest.java
```diff
@@ -114,6 +114,7 @@ public void multipleTransactionVersionsReturnsSnapshotAndOnlyRelevantRecentEvent
     }
 
     @Test
+    @ShouldRetry
```
Jolyon-S, line 117: was flaking

gmaretic: If this test is retried, we could miss legit nondeterministic bugs -- what was the flake and can we fix it?

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
gmaretic: fml

timelock-server/src/integTest/java/com/palantir/atlasdb/timelock/LockWatchValueIntegrationTest.java
```diff
@@ -471,6 +487,61 @@ public void lateAbortingTransactionDoesNotFlushValuesToCentralCache() {
...
+
+        try {
+            executor.shutdown();
+            executor.awaitTermination(60, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+        }
+
+        assertThat(failedTransactions).isEmpty();
```
gmaretic, line 520: Slightly different semantics, but I think it reads better:
```suggestion
        List<Future<?>> transactions = IntStream.range(0, numTransactions)
                .mapToObj(_num -> executor.submit(this::randomTransactionTask))
                .collect(Collectors.toList());

        for (Future<?> transaction : transactions) {
            try {
                transaction.get(60, TimeUnit.SECONDS);
            } catch (TransactionFailedRetriableException e) {
                // this is ok
            } catch (InterruptedException | TimeoutException e) {
                fail("Transaction took too long", e);
            } catch (ExecutionException e) {
                fail("Encountered nonretriable exception", e);
            }
        }

        executor.shutdown();
```

gmaretic (approved):
Generally looks good, though not sure about that retry

atlasdb-ete-tests/src/test/java/com/palantir/atlasdb/ete/LockWatchEteTest.java
Jolyon-S: It's a genuine flake. I'm going to probably refactor these ETEs anyway soon enough, so I'm going to keep this rule here.

PR #5668, opened 2021-09-28
[WIP] [LW] Add Unit Tests for Missing Codepaths

Jolyon-S:
**Goals (and why)**:
There are parts of the code that are not tested - this was largely discovered by running all lock watch unit tests with coverage, and observing missing lines. Thus, it is entirely possible for there to be gaps (seeing as multiple edge cases could go through the same code), but this is still a step in the right direction.

* Fix up a couple of odd pieces of behaviour
* Tests!

Are there any obvious places that are missing tests? Do any of these look flaky?

Diff

This week, ideally.


Jolyon-S:
closing in favour of https://github.com/palantir/atlasdb/pull/5672 and https://github.com/palantir/atlasdb/pull/5671

